PR #2369, opened 2021-06-20
Cytoplasm nucleus bug

### LINKED ISSUE #1961: Unplacing and replacing nucleus crashes the game.
Steps to reproduce:
1. Place the nucleus over the original cytoplasm
2. Undo the placement
3. Place it in the same place again
```
E 0:00:24.462   IntPtr Godot.Object.GetPtr(Godot.Object ): System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'NucleusMesh'.
  <C++ Error>   Unhandled exception
  <C++ Source>  /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Core/Object.base.cs:37 @ IntPtr Godot.Object.GetPtr(Godot.Object )()
  <Stack Trace> Object.base.cs:37 @ IntPtr Godot.Object.GetPtr(Godot.Object )()
                Node.cs:537 @ void Godot.Node.RemoveChild(Godot.Node )()
                SceneDisplayer.cs:28 @ void SceneDisplayer.LoadNewScene()()
                SceneDisplayer.cs:20 @ void SceneDisplayer.set_Scene(System.String )()
                MicrobeEditor.cs:1788 @ void MicrobeEditor.UpdateOrganellePlaceHolderScene(SceneDisplayer , System.String )()
                MicrobeEditor.cs:1764 @ void MicrobeEditor.UpdateAlreadyPlacedVisuals()()
                MicrobeEditor.cs:1679 @ void MicrobeEditor.OnOrganellesChanged()()
                MicrobeEditor.cs:1199 @ void MicrobeEditor.UpdateEditor(Single )()
                MicrobeEditor.cs:530 @ void MicrobeEditor._Process(Single )()
```

https://user-images.githubusercontent.com/59157446/102834405-c7c3d980-43a8-11eb-9180-f77811ea5454.mp4

### LINKED ISSUE #2340: Crash when deleting other organelles and then moving the nucleus
I seem to sometimes be able to get a crash when trying to move the nucleus after deleting (and moving) other organelles to be left with less than 5 MP. Callstack seems to be the same as: https://github.com/Revolutionary-Games/Thrive/issues/1961
so fixing that might also fix this

Kemikal1:

Fixes #2340  and Fixes #1961

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
You'll probably want to override your local master branch from the main repo and rebase these changes on that, and finally force push to clear this out.

Btw would this also fix: https://github.com/Revolutionary-Games/Thrive/issues/1961 ?

Kemikal1:
It fixes everyhting related to nucleus crash for me.
"You'll probably want to override your local master branch from the main repo and rebase these changes on that, and finally force push to clear this out."
IS this a must?Ill pull origin to my main and delete this fork if it gets accepted.

hhyyrylainen:
I guess as there aren't conflicts that is not necessary as this will be squashed to a single commit anyway so it doesn't matter that there's a bunch of useless commits in this PR.

src/general/SceneDisplayer.cs
```diff
@@ -23,7 +23,7 @@ public string Scene
 
     private void LoadNewScene()
     {
-        if (currentlyShown != null)
+        if (currentlyShown != null && IsInstanceValid(currentlyShown))
```
hhyyrylainen, line 26: I'll just make an observation regarding this as I don't have time to investigate now: so this now asks Godot "hey have we accidentally deleted this currentlyShown scene"? I think a better approach would be investigating where currentlyShown is is freed but not set to null. 

Would adding a line after 29 like `currentlyShown = nul;l`, fix this?

Kemikal1: Yes and Yes.

Kemikal1: Is cytoplasm.tscn supposed to be loaded in organel.Definition.Display Scene??Cause its not.

hhyyrylainen: It's not because the DisplayScene is what's shown in the microbe for that organelle. Cytoplasm only has a scene for when it is used as a corpse chunk.

Kemikal1: Im saying because that also seems to fix the problem.

hhyyrylainen: I suppose that also *masks* the problem. I want to fix the actual problem which is that `currentlyShown` is sometimes freed without setting it to null. That's the change that actually addresses the underlying problem, and one I want added to the codebase.

Kemikal1: Godot doesnt set it to null.it marks it as unused.Doesnt remove all references to the removed object.thats why the null check doesnt work.

hhyyrylainen: That's why the correct thing is in our code to set the variable to null after deleting that. That solves all the problems and is valid usage. It's completely nonsensical to me to not do things properly here and instead rely on asking the engine if we did something stupid with an object. I have experience in C++ and to me this seems as silly as deleting a pointer in C++ without setting it to null, the difference being there that you can't ask the runtime "did I delete this already?", instead you need to handle it yourself.

Kemikal1: "I have experience in C++ and to me this seems as silly as deleting a pointer in C++ without setting it to null"
Godot has its own type of null.And that is "unused."And aparently null here means uninstantaiated and "unused" is the actuall "null".And thats the check.I dont think were doing any kind of stupid things??
So?How do you wnat it boss?Set to null?

hhyyrylainen: Set the C# variable to null. There's a null check on this *exact* line which will prevent the problem of freeing a Godot object multiple times.

hhyyrylainen (approved):
Works perfectly for the nucleus redo bug.

hhyyrylainen:
Does this still fix #2340 ?

Kemikal1:
Yes

hhyyrylainen:
Excellent

Issue #1645, opened 2020-09-23
Make cells rotate around their center of mass (fractional offset)


hhyyrylainen:
instead of always having the hex at 0,0 be the center of the cell in the stage.
Instead could compute average position of all hexes and use that as offset all the organelles when rendering and in physics.
This will prevent weirdness if the player doesn't build their cell around the middle of the editor grid.

closes #1678

Kasterisk:
This should be already fixed by #1695

hhyyrylainen:
I left this open because that only moves the center based on the hexes.
This issue is for making it be possible to fractionally move the origin offset, to make it exactly perfect.

hhyyrylainen:
I'm starting to think that the current approach is probably good enough and this can be just closed if no one cares enough to implement this slight improvement.

hhyyrylainen:
Looks like no one cares, so I'll close this as an unnecessary complication to the offset code that isn't needed.

PR #2379, opened 2021-06-23
Flagella orientation

### LINKED ISSUE #1297: Flagella orientation calculation fails at 0,0
https://community.revolutionarygamesstudio.com/t/general-0-5-0-feedback/2645/5?u=hhyyrylainen

related issue: https://github.com/Revolutionary-Games/Thrive/issues/1500

### LINKED ISSUE #1500: Flagella orientation is sometimes broken
reversed_flagellum

even when not placed at origin. ~~Might be related to loading a save or reproduction progress.~~ Seems to also be based on the positioning of the organelles.

### LINKED ISSUE #2255: Flagella orientation not automatically points outwards from cell origin after exiting the editor
Example cell with all external parts oriented inwards/sideways:

Once exiting the editor, the pili automatically corrects their orientation while the flagella doesn't seem to do that:

Weirdly enough, all seems to be oriented correctly outwards while seen through the editor's microbe preview.

Kemikal1:

Fixes #2255 ,fixes #1297 
Fixes #1500 
Some engine bug most likely, checking the transform with a condition seems to refresh it.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
> Id apreciate if someone told me where the first call to organelle updates (Update method) in the code is after editor exit?I couldnt figure it out.Thx

It's called from Microbe._Process
https://github.com/Revolutionary-Games/Thrive/blob/9ad02d03cc9668c09670e7cb63e083d50d408202/src/microbe_stage/Microbe.cs#L1259

Pretty weird that it would take exactly 5 updates... The membrane takes exactly 1 frame update to be generated properly, after that the positions the organelles get from it should stay constant.

I originally thought that maybe the rotation math fails when calculating on a zero vector? Which would come from calculating the angle between the middle hex and itself.

Kemikal1:
The math works ok,since the editor was showing it right but the game right after editor exit wasnt and when i changed the angle to always be 180 only the editor changed not the flagella in game,very weird indeed.

hhyyrylainen:
For the pilus the logic is so that the "needs position update" should cause one position change. Does it truly require 5 position updates for this to work? Or would one work similarly to the pilus?

Kemikal1:
If you are refering to the NeedUpdateAnyway , it didnt work for me with any number of updates under 5.If i change the angle in Update method the flagella orientation only changes in the microbe editor preview not after i exit the editor.

hhyyrylainen:
I suggest you use a debugger / print statements to check why the flagella visual positioning math only comes out correctly after 5 times. That will give us more insight into where that weirdness comes from. As I said the membrane positions *should* be computed after one frame, so it's interesting why 5 frames are needed here for the position calculations to come up with the right numbers...

Kemikal1:
I debugged the memebrane coords when updating for both pilus and flagella after exiting editor,basically the same,ill show them here today.

hhyyrylainen:
If the positions come out the same, then why does it help to update the position 5 times to be the same? If they are truly the same, this could be a Godot bug / incorrect usage problem.

Kemikal1:
Untitled23
Untitled22
I printed the arguments of onPosistionChanged

hhyyrylainen:
Are you then saying that Godot does  something different on the 5th time the exact same position is set on a node?
Sounds like an engine issue.

Kemikal1:
Yea engine issue again i think.The math only fails when the pilus fails to orientate i guess.

Zireael07:
If Godot does something different only on the 5th time, then yes, I suspect it's an issue to be raised on the engine repo.

Kemikal1:
Would it be okay if i tried to fix the current external components orientation without changing their rotation
 calculation method?

hhyyrylainen:
I'm not sure what you are asking. If you are going to tweak the math, it should be in the code that tries to calculate the angle for visual display. That's the whole point of that computation so doing a separate step in a different method to try to fix it afterwards, seems like a non-optimal design to me.

Kemikal1:
Im trying to calculate the true position of an organelle according to other organelles so as not to fail at hex(0,0).Or maybe modify the update method to include the true middle of a microbe?Im just asking.

hhyyrylainen:
I think it would be good to just hardcode 0,0 to return a specific rotation value.

Kemikal1:
So it shouldnt depend on the other organelles?

hhyyrylainen:
It could if you want, but I'll accept even a very simple fix.

Kemikal1:
After 3-4 hours of testing,the only bug about orientation i had was when ,me and my divided cell had 2 flagellas attached and one rotated wrongly after i exited the editor ,it corrected itself after i hit a rock.Must be the update function again not working as intended ,maybe engine prob.

hhyyrylainen:
Could someone from the team playtest this?

FuJa0815:
I have managed to create a microbe without a membrane: 
grafik

What I have done:

1. Create a microbe with ridiculus amounts of flagela
2. remove everything and just have a single flagella as the microbe

The membrane reappeared after dying.

Microbes with only a single flagella seem very broken. Single pili are fine. After some more testing it crashed: 
[log.txt](https://github.com/Revolutionary-Games/Thrive/files/6795064/log.txt)

This does not happen on 0.5.4.

Kemikal1:
Did you enter between those steps in the game after editing or did you do all the steps in the editor?Also when you remove everything did u use the button for new microbe??

FuJa0815:
I went ingame in between.
I did not use the new microbe button, I've just deleted everything.

It is consistently buggy if you have only a single flagellum and nothing else. A single pili works

Kemikal1:
For me it also happens with 0.5.4.
Just press new microbe then youll see no membrane.
[Toomany memb.zip](https://github.com/Revolutionary-Games/Thrive/files/6795400/Toomany.memb.zip)

FuJa0815:
Huh, seems like there are more steps required to reproduce this bug. But, another orientation bug:
0.5.4.0:
grafik
grafik

This PR:
grafik
grafik

FuJa0815:
Another orientation bug:
grafik
grafik

Kemikal1:
How did you create those bugs?

FuJa0815:
I have done nothing special. I just built the cells shown in the screenshots and went ingame.

Kemikal1:
This works for me.Are you sure you have my latest PR?

Kemikal1:
https://user-images.githubusercontent.com/11059623/125198166-7e19de00-e269-11eb-979b-975c5442a908.mp4


https://user-images.githubusercontent.com/11059623/125198171-8245fb80-e269-11eb-8803-8fe014bb6207.mp4

FuJa0815:
I think I've got it!
This only bugs out if you have any other language than english selected!

src/microbe_stage/Microbe.cs
```diff
@@ -1373,6 +1373,35 @@ public void UnbindAll()
...
     }
 
+    /// <summary>
+    /// Recalculates exterior organelles according to the true middle of the cell
+    /// </summary>
+    public void RecalculateExteriorOrganellesPosition()
+    {
+        foreach (var organelle in organelles.Organelles)
+        {
+            if ((organelle.Definition.Name == "Flagellum" || organelle.Definition.Name == "Pilus") &&
```
FuJa0815, line 1383: this is the troublemaker with languages

Kemikal1: LOOOOOOOOOL

Kemikal1: Is there a variable that hold the name for those ?Like aconstant?

FuJa0815: use `InternalName`

FuJa0815: or even better:
`organelle.HasComponentFactory<ExternallyPositionedComponent>()`

Kemikal1:
Anyway i found another bug that if you enter then exit editor repeateadly with a single flagellum the microbe bugs completely  until you die or exit to main menu.

src/microbe_stage/Microbe.cs
hhyyrylainen: Definitely should use the actual info on what this *should* run on.
If exact names should be checked, which probably isn't needed here, but I'll comment for the future: use `var rusticyanin = simulation.GetOrganelleType("rusticyanin");` and then compare for example `if(organelle == rusticyanin)`

hhyyrylainen: Is this some kind of special loop to process external organelles? Why not use the normal organelle Update callback for whatever these need to do?

Kemikal1: I need it to get the other organelles position.SO that even if you place the external components inside the cell they will be placed outside the cell at editor exit/preview

hhyyrylainen: Why is that being done?
It's been this way forever:
- External organelles always place their visual model at the membrane edge, no matter if they are an edge organelle or not.

Kemikal1: Their orientation would fail,if i didnt.

hhyyrylainen: How, why? Why not just fix the external organelle visual positioning math, like I suggested?

Kemikal1: The math in the `Update` doesnt have access to others organelles position,and i would have to modify the` update `function of the organelles to take the middle of the cell into account,which i thought wouldnt be that nice.

hhyyrylainen: The middle of the cell is always 0,0. at least once https://github.com/Revolutionary-Games/Thrive/issues/1645 is complete.
I would much prefer to have the calculation be:
- just normal when the flagella is not at 0,0
- If the flagella ends up being at 0,0 *then* special code kicks in that looks to see if the flagellum should point backwards or forwards determined by looking at the Microbe the organelle is attached to
Would that approach have a problem?

Kemikal1: Its a problem if the flagella is palced "inside" the cell,being surrounded by other organelles.The rotation is based on the middle of the cell and` organellePos`.So  the `membranecords` dont match with actuall membrane coordonates.So rotation fails Even if its not at 0,0,because i couldve made an exception with that.

hhyyrylainen: Why not then either disconnect the visual rotation from the thrust calculation *or* also change the thrust calculation to re-calculate the orientation like for the visuals, so that way the visuals and the computed thrust would match?

Kemikal1: Sorry,i didnt understand,I dont think the visual rotation and thrust calculation are connected?

hhyyrylainen: They aren't currently, I was just trying to explain what could be done to address this issue *without* having to modify the organelle layout. I just looked at the code and it is hugely complicated... it needs a ton of explaining comments if it is to be kept as is. That's why I even more strongly want just the calculations done on the flagella to be changed, otherwise this causes code complications in many other places.

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -59,11 +66,13 @@ protected override void CustomAttach()
     ///   Calculate the momentum of the movement organelle based on
     ///   angle towards middle of cell
     /// </summary>
-    private static Vector3 CalculateForce(Hex pos, float momentum)
+    private static Vector3 CalculateForce(Hex pos, float momentum, Vector3 defaultPos)
```
Maxonovien, line 69: Is defaultPos intended to define the behavior of an organelle placed in the middle of the cell, as mimicking an organelle at this position?

I think it should be clarified in the function documentation. Especially as it does not seem obvious that this is the only right way to go

Kemikal1: Thats how it was ,i didnt make that change.

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -10,9 +10,9 @@ public class MovementComponent : ExternallyPositionedComponent
 
     private readonly Compound atp = SimulationParameters.Instance.GetCompound("atp");
 
+    private int countInitialization = 6;
```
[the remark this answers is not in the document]
Kemikal1: The flagella needs 5 inital updates to correctly orient itself after exiting the editor.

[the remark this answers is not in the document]
Maxonovien: I got that from the discussion, but not from the name, which I think is a pretty good sign this variable should be renamed.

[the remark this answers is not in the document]
hhyyrylainen: I still would like to find out why this hack is needed, and if there's **any** way at all to avoid this...

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -59,11 +66,13 @@ protected override void CustomAttach()
     ///   Calculate the momentum of the movement organelle based on
     ///   angle towards middle of cell
     /// </summary>
-    private static Vector3 CalculateForce(Hex pos, float momentum)
+    private static Vector3 CalculateForce(Hex pos, float momentum, Vector3 defaultPos)
     {
-        Vector3 organelle = Hex.AxialToCartesian(pos);
+        Vector3 organellePoz = Hex.AxialToCartesian(pos);
```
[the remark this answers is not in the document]
Kemikal1: NO its pos,because poz is already used

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -1841,6 +1841,11 @@ private bool AddOrganelle(OrganelleTemplate organelle)
 
         organelle.PlacedThisSession = true;
 
+        if (organelle.Position == new Hex(0, 0))
+            organelle.FalsePosition = organelle.Position;
+        else
+            organelle.FalsePosition = new Hex(1, 1);
```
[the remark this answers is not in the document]
Kemikal1: WHat do you mean?

src/microbe_stage/organelle_components/MovementComponent.cs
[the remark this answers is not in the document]
hhyyrylainen: "poz" is not an approved abbreviation. You'll need to name your variable something else.

src/microbe_stage/organelle_components/MovementComponent.cs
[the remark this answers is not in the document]
Kemikal1: Believe me id also like to know why...Checked the spawn ,and all,and everything is fine.At least what ive seen

src/microbe_stage/organelle_components/ExternallyPositionedComponent.cs
```diff
@@ -21,7 +26,7 @@ public void OnAttachToCell(PlacedOrganelle organelle)
     {
         this.organelle = organelle;
         organellePos = Hex.AxialToCartesian(organelle.Position);
-
+        defaultVisualPos = new Vector3(0, 0, -1);
```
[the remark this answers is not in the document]
Kemikal1: Whats the difference?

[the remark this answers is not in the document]
hhyyrylainen: It's a proper constant instead of guessing that -1 in Z axis is forward.
I've learned not to trust values like this by changing graphics engines a bunch of times (and definitely the forward direction is *not* standard)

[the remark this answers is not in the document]
Kemikal1: Its default as ive defined it, it could by any pf the 4 Directions.

src/microbe_stage/organelle_components/MovementComponent.cs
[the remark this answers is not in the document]
Kemikal1: Are those in the style guide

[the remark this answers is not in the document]
hhyyrylainen: Not yet... https://github.com/Revolutionary-Games/Thrive/issues/2392
it seems that all documentation issues only get done after many months after I finally pick them up.

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -58,12 +65,16 @@ protected override void CustomAttach()
     /// <summary>
     ///   Calculate the momentum of the movement organelle based on
     ///   angle towards middle of cell
+    ///   If the flagella is  placed in the microbes center ,hence delta equals 0,
```
84634E1A607A, line 68: Typos.

src/microbe_stage/organelle_components/ExternallyPositionedComponent.cs
```diff
@@ -21,7 +26,7 @@ public void OnAttachToCell(PlacedOrganelle organelle)
     {
         this.organelle = organelle;
         organellePos = Hex.AxialToCartesian(organelle.Position);
-
+        defaultVisualPos = Vector3.Forward;
```
84634E1A607A, line 29: Assign a pos with a "Forward" is a little bit strange. Maybe add a comment or use new instead?

84634E1A607A (changes requested):
Problem solved, code looks good, but two small suggestions.

src/microbe_stage/organelle_components/ExternallyPositionedComponent.cs
Kemikal1: No reason to use new, vector forward already exists creating a new vector takes more time.

src/microbe_stage/organelle_components/MovementComponent.cs
84634E1A607A: Space after comma?

84634E1A607A (approved):

src/microbe_stage/PlacedOrganelle.cs
```diff
@@ -45,7 +45,6 @@ public PlacedOrganelle()
     }
 
     public OrganelleDefinition Definition { get; set; }
-
```
Maxonovien, line 48: I don't see any reason to just remove line breaks from this file, especially as the rest of properties are separated by it.

src/microbe_stage/PlacedOrganelle.cs
```diff
@@ -198,7 +197,6 @@ public void OnAddedToMicrobe(Microbe microbe)
     {
         if (Definition == null)
             throw new Exception("PlacedOrganelle has no definition set");
-
```
Maxonovien, line 201: Same here, from what I recall the rest of the code separates such `if`s by a line break.

src/microbe_stage/organelle_components/ExternallyPositionedComponent.cs
```diff
@@ -39,13 +44,15 @@ public virtual void Update(float elapsed)
 
         Vector3 middle = Hex.AxialToCartesian(new Hex(0, 0));
         var delta = middle - organellePos;
+        if (delta == new Vector3(0, 0, 0))
+            delta = defaultVisualPos;
         Vector3 exit = middle - delta;
         var membraneCoords = organelle.ParentMicrobe.Membrane.GetVectorTowardsNearestPointOfMembrane(exit.x,
             exit.z);
-
```
Maxonovien, line 45: Line breaks before `if`s are the norm in the code, so I'd argue against removing them.

src/microbe_stage/organelle_components/ExternallyPositionedComponent.cs
```diff
@@ -39,13 +44,15 @@ public virtual void Update(float elapsed)
 
         Vector3 middle = Hex.AxialToCartesian(new Hex(0, 0));
         var delta = middle - organellePos;
+        if (delta == new Vector3(0, 0, 0))
```
Maxonovien, line 47: As said before, there is already `Vector3.Zero` for this.

src/microbe_stage/organelle_components/ExternallyPositionedComponent.cs
```diff
@@ -39,13 +44,15 @@ public virtual void Update(float elapsed)
 
         Vector3 middle = Hex.AxialToCartesian(new Hex(0, 0));
         var delta = middle - organellePos;
```
Maxonovien, line 46: Also, now that I come back to this after a long time, I'm a bit confused by the name of this variable. Perhaps it could be worth renaming it to better reflect what this is.

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -58,12 +65,16 @@ protected override void CustomAttach()
...
-    private static Vector3 CalculateForce(Hex pos, float momentum)
+    private static Vector3 CalculateForce(Hex pos, float momentum, Vector3 defaultPos)
     {
-        Vector3 organelle = Hex.AxialToCartesian(pos);
+        Vector3 organellePosition = Hex.AxialToCartesian(pos);
         Vector3 middle = Hex.AxialToCartesian(new Hex(0, 0));
-        var delta = middle - organelle;
+        var delta = middle - organellePosition;
+        if (delta == Vector3.Zero)
+            delta = defaultPos;
```
Maxonovien, line 77: Same here. By the way, the codes are very similar so it could be worth having some math util, here making a difference unless it is zero.

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -58,12 +65,16 @@ protected override void CustomAttach()
...
     ///   angle towards middle of cell
+    ///   If the flagella is  placed in the microbe's center, hence delta equals 0,
+    ///   consider defaultPos as the organelles "false" position.
     /// </summary>
-    private static Vector3 CalculateForce(Hex pos, float momentum)
+    private static Vector3 CalculateForce(Hex pos, float momentum, Vector3 defaultPos)
     {
-        Vector3 organelle = Hex.AxialToCartesian(pos);
+        Vector3 organellePosition = Hex.AxialToCartesian(pos);
         Vector3 middle = Hex.AxialToCartesian(new Hex(0, 0));
```
Maxonovien, line 74: It might be worth having a constant for Hex(0, 0) in the dedicated class. There are some already, but none for this very one.

Maxonovien (changes requested):
Some issues regarding clarity, efficiency, and style. File `PlacedOrganelle.cs` is modified only to remove line breaks, so I don't think it should be changed at all.

src/microbe_stage/organelle_components/MovementComponent.cs
Kemikal1: I thinkt that should be in a diff PR, this PR's goal is different.

src/microbe_stage/organelle_components/MovementComponent.cs
Kemikal1: I dont think thats worth doing for  2 lines of code.

src/microbe_stage/organelle_components/ExternallyPositionedComponent.cs
Kemikal1: I dont understand what that delta is as well, so any suggestion?

FuJa0815: relativeOrganelleVector?

FuJa0815 (dismissed):

Kemikal1:
So i was testing today and i found that without updating the flagella 6 times, if you print the `organelle.OrganelleGraphics.Transform`  in the `OnPositionChanged` the flagella is ok. Smells like engine bug.

src/microbe_stage/organelle_components/ExternallyPositionedComponent.cs
```diff
@@ -15,13 +20,12 @@ public abstract class ExternallyPositionedComponent : IOrganelleComponent
...
     ///   Last calculated position, Used to not have to recreate the physics all the time
     /// </summary>
-    protected Vector3 lastCalculatedPosition = new Vector3(0, 0, 0);
+    protected Vector3 lastCalculatedPosition = Vector3.Zero;
 
     public void OnAttachToCell(PlacedOrganelle organelle)
     {
         this.organelle = organelle;
         organellePos = Hex.AxialToCartesian(organelle.Position);
-
```
hhyyrylainen, line 24: unnecessary change

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -58,12 +66,16 @@ protected override void CustomAttach()
     /// <summary>
     ///   Calculate the momentum of the movement organelle based on
     ///   angle towards middle of cell
+    ///   If the flagella is placed in the microbe's center, hence delta equals 0,
+    ///   consider defaultPos as the organelles "false" position.
```
hhyyrylainen, line 70:
```suggestion
    ///   consider defaultPos as the organelle's "false" position.
```

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -49,6 +48,15 @@ protected override void CustomAttach()
         SetSpeedFactor(0.25f);
     }
 
+    protected override bool NeedsUpdateAnyway()
+    {
+        // The basis of the transform represents the rotation, as long as the rotation is not modified,
+        // the organnel needs to be updated.
+        // TO DO:
```
hhyyrylainen, line 55: two typos here, and TODO: needs to have the TODO comment immediately following it, not on the next line

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -38,7 +38,6 @@ public override void Update(float elapsed)
     protected override void CustomAttach()
     {
         force = CalculateForce(organelle.Position, Momentum);
-
```
hhyyrylainen, line 41: unnecessary line removal which puts unrelated stuff in the same code "block", which I dislike

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -49,6 +48,15 @@ protected override void CustomAttach()
         SetSpeedFactor(0.25f);
     }
 
+    protected override bool NeedsUpdateAnyway()
+    {
+        // The basis of the transform represents the rotation, as long as the rotation is not modified,
+        // the organnel needs to be updated.
+        // TO DO:
+        // Calculated rotations should never equal the identity, it should be kept an eye on if it does
```
hhyyrylainen, line 56: Also this TODO needs a github issue opened for this and linked here

Kemikal1: So the issue is?

hhyyrylainen: This needs to be like this:
```
TODO: Calculated rotations should never equal the identity, it should be kept an eye on if it does
See: CREATE A NEW ISSUE AND LINK IT HERE
```
This is presumably here due to an engine bug? Could you clarify that a bit here in this comment?

Kemikal1: Is there a reason to link it if its Godot related tho?

hhyyrylainen: I've created issues in our repo to track issues that are Godot problems.

hhyyrylainen (approved):
I guess no one else is going to check this, so I'm going to merge this now to give this a few days time to be in master before 0.5.6.1

PR #1521, opened 2020-08-12
Mod support

I-Have-No-Idea-What-IAmDoing:
Not Really Fixes But Improves??? #584 

I made a mod loader with a priority/load order 
image

it loads pck files which has it own caveats like:
- 2gb size limit (this probably won't be a problem but might be with giant mods)
- one file can only be changed by one mod (so 2 mods can't edit the main menu, that where the load order comes in)
- no way to unload mods
- mods are not persistent (any changes made by mods are lost when the game is closed, this is either a good or bad thing)
- they are harder to make than just editing files

you should be able to load c# code by loading the dll file but i have no idea how to test it, so mods should use gdscript for scripting. i am also unsure if you can modify the base game code since it compiled into the dll file.

mods are detected by a mod_info.json file which holds info like the name, author, version, a description of the mod, and the location of a dll file (if it needed). the actual mod file is the mod.pck file. each mod has it own folder but the name of that folder does not matter.

there 3 problems which i don't know how to fix which is why this is a draft right now:
- When replacing existing asset, you have to change scene for this to take effect(go into the game and go back) the loader should already reload the main menu but the changes still don't take effect until you change scene
(Edit: This is not true, the asset only changes when you load a save... weird)
- Json files are not able to be changed since they are readonly and are assigned when the game is starting (this prevent mods that add new music,or add new organelles, or anything in the simulation parameters)
- i unsure on how you change/add c# file, i know you load a dll but i am unsure, how to do it, if that possible or even if my implementation actually works right now 

there also some changes that i want to make like:
- a reset button (since you can't unload mods this will just load the pck file of the release, this will reset any file but it won't get rid of any new ones) ✅
- a way to easily move the load order beside unloading and reloading (either with a drag and drop or just buttons) ✅

it should be simple to disable this feature as this is fairly self-contained, you just have to disable the button that opens the mod loader

Here is 4 mods that i made that you can use to test with:
https://we.tl/t-yafn4CjqIi

- Mod Template - This is a template for modders to use , it only contain the mod_info.json
- Alternate Main Menu - This mod show adding new assets and changing scene layout
- Thriv 2 (Better Asset Flip) - This shows changing existing assets but this is the one with the bug
- Quick Reset - This show scripting (including calling c# methods) and sound (This one has the most compatibility with other mods)

all of them can be seen and tested in the main menu

PR #2381, opened 2021-06-28
New Mod Loader

### LINKED ISSUE #584: Better modding experience
Not a huge priority but in the future the modding experience (i.e. no more "just unzip these files on top of the game" installations) needs to be improved.

Currently potential approaches:
- https://mod.io (they have their own content policy and terms of service that I partly dislike)
- roll our own (we'd need a website to handle uploading and downloading mod files and then the game or the launcher needs the ability to download them)

I-Have-No-Idea-What-IAmDoing:
**Long Description of What This PR Does**
image

This PR adds a mod loader with a load order/priority list, this is a rebuilt version of my previous mod loader #1521.
It still loads pck files and mods are still detected by a mod_info.json file, which holds information about the mod. 

The new features in this PR are:
### General
* The description of mods use rich text labels, so they can be formatted 
* Mods can now have icons associated with them that will display next to them
* They can also have a preview image that displayed next to the description
* The list of mod have actual style to them now
* The Mod list is split into 3 tabs, the actual mod loader, the autoloaded mod (shows the mod that is being autoloaded [which is determined by the mod maker]), and Failed to Load mod (which shows why some mods fail to load).
* The split between the mod info panel and the loader uses a split container and can be manipulated show show more of either panel
image

### Mod Info
image

* Mods can now specify which version of thrive they are compatible or not compatible with
  * The version they are compatible with are shown in the mod loader 
  * If they specify version of mods they are compatible with but it does not include the current version then it will show up yellow and a tooltip shows why
    * It they are not compatible with the current version then they are shown up as red and show a similar tooltip 
* Mods can also define dependencies
  * These are also shown in the main menu through a button which opens a popup
  * Dependencies are defined through the name of the mod
    * the version of the mod does not matter for dependencies
* Mods can also define mods they are incompatible with, they work the same way as dependencies
* Mods can also specify which mods they want to load before and after
  * They work a little differently from dependencies since they do not cause errors if they are not loaded with it
*  All of these are checked by mod loader before loading the mod and can alert the player if they find anything wrong with the mods they are trying to load
*  There is a check button so the player can check the list before trying to load them
*  There is also a safe mode checkbox which if off, turns off checking before loading
image
* Mods can have a display name attach to it, which will be used instead of the mod name
### Autoloaded Mods
image
* Autoloaded mods are not displayed in the mod loader but are now displayed in this tab
  * Since these are loaded automatically this make more sense this way
 ### Failed to load Mods
image
* This tab show mods that failed to load
* It shows the reason on the side
* The name should probably be changed but I am not sure what else to call this tab
______________________________________
This is still a draft because i still have some issues to work out:
* I still need to rework everything to be translatable
* Mods that modify c# script don't load and crash the game right now
  * Might be a Godot issue? 
* Still need to work on documenting and cleaning up the code

**Related Issues**

replaces #1521 
resolves #584

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
> Mods that modify c# script don't load and crash the game right now

How are you handling that right now? Mods that contain code, need to compile into their own .dll file that is included in a .pck file and then loaded as an extra assembly when loading the mod.

I-Have-No-Idea-What-IAmDoing:
Right now they are loaded as separate files, it the same way I was trying to do it in the older version

Though whenever i try doing that Godot and the editor just crashes

hhyyrylainen:
Do you have like a sample mod on hand that has C# code in it? Does it crash if you make a dll from just like the following code:
```
public static class RandomMod {
}
```
Maybe it doesn't like conflicting class names or something (if you modified the source code and compiled that entirely as a .dll). I think the way code mods need to work is that they *only* contain code for that specific mod. Then maybe reflection or linking against the main codebase in .dll form is how the code can reference the existing code. That means that we need to probably add a bunch of code hooks that mods can use to hook into existing code in the game.

I-Have-No-Idea-What-IAmDoing:
Yeah, it worked, though I don't think it's conflicting class name because when i tried changing the class name to something is already being used, it does not crash

hhyyrylainen:
Well anyway, the mod making instructions need to contain the bit about not compiling the entire game source code with changes into a mod. That would also mean that only a single code changing mod could be enabled at a time, which is also bad...

I-Have-No-Idea-What-IAmDoing:
I think i am going to abandon this branch again and just reimplement everything on a different one as Godot is now just freezing on start-up whenever i tried loading the project

hhyyrylainen:
Sad day.
I kinda want to get mod support in in time for the Steam release, which I'm still hoping can work out to happen this year. For that I'll probably start with the how to make code and asset mods, and the mod loader backend systems. After those are done I would appreciate any help with the GUI for the mod loader. Though, that's still not very high on my todo list so this is not going to happen yet.

PR #2382, opened 2021-06-28
Reset player's microbe state to Normal after selecting a cell to unbind

### LINKED ISSUE #2360: Manually unbinding last bound cell does not reset the player's microbe state
When manually selecting the only bound cell in unbinding mode it doesn't return the player's microbe state back to normal, so as a result the player is stuck on the unbinding state (and need to manually toggle the unbind action or unbind all to leave it). The microbe state should automatically reset due to all cells in the colony being disbanded.

I assume this is different from #2342 because this is specific to the player's microbe.

Kasterisk:

Added player cell state reset in `AcceptUnbind` to fix a problem where the player is stuck in unbinding mode after unbinding the only one bonded cell.

**Related Issues**

Fixes #2360 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/microbe_stage/PlayerMicrobeInput.cs
```diff
@@ -131,6 +131,10 @@ public bool AcceptUnbind()
         RemoveCellFromColony(target);
 
         stage.HUD.HintText = string.Empty;
+
+        // In case this is the last colony member being unbonded, return the player's microbe state to normal
+        stage.Player.State = Microbe.MicrobeState.Normal;
```
hhyyrylainen, line 136: Shouldn't this line only run when the last member was removed?
But then shouldn't the logic really go into the microbe class / colony?
@FuJa0815

FuJa0815: The AcceptUnbind method is for removing microbes by clicking on them after pressing Ctrl+U. Therefore this logic is wrong here, because the player might want to remove multiple members at once before exiting the unbind mode. I'm on my phone right now, therefore I can't really look into the code right now, but there has to be a better place.

Kasterisk: As of the latest version, the player automatically exits the unbind mode after clicking a cell. What's weird was that it doesn't seem to do that if you're unbinding the only one bound cell, so this logic essentially forces the player's microbe state reset.

hhyyrylainen: So there is something here to fix anyway, I'm still pretty busy with other stuff, so if you two could figure out a good fix / change regarding this that'd be great.

FuJa0815 (approved):
This fix works by resetting the state earlier. This works because setting State works colony-wise. This of course does not work when you remove the cell from the colony before resetting the state

hhyyrylainen:
Does this fully work as intended? Doesn't this now end binding mode on unbinding the first cell? Didn't it before allow you to unbind as many cells as you wanted?

FuJa0815:
> 
> 
> Does this fully work as intended? Doesn't this now end binding mode on unbinding the first cell? Didn't it before allow you to unbind as many cells as you wanted?

https://github.com/Revolutionary-Games/Thrive/pull/2382#discussion_r661607825

Issue #2236, opened 2021-05-16
RichTextLabel not updating its height when rect size is changed with FitContentHeight property enabled.


Kasterisk:
This is to track a workaround needed to fix bug mentioned in the title, which will possibly be irrelevant in future Godot versions.
https://github.com/Revolutionary-Games/Thrive/blob/5ce5b11baf6f569fa4085f838b4cfc1b4aa16011/src/gui_common/tooltip/microbe_editor/SelectionMenuToolTip.cs#L165

hhyyrylainen:
I found that the constant _Draw calls sap a *ton* of performance in the editor on the report tab, so I'll try to make it so that this workaround only needs to run once when the text content is changed.

hhyyrylainen:
This should be checked again with Godot 4 now being used.

PR #2383, opened 2021-06-28
Help menu enhancements

### LINKED ISSUE #1913: Make tooltips and help text use actual keybindings
Now with rebindable keys the help and tooltip texts probably are going to end up being wrong if the player rebinds keys.
So the text strings for those should be updated with placeholders `{0}` and code hooked up to replace those with the actual keybindings for things when loading the things (this is probably a bit tricky to figure out the optimal place to do this regarding many of the tooltips). Perhaps on game start and exiting the options menu would be the most bulletproof for static text, and for all text that is dynamically generated just lookup the keybindings whenever the text is generated.

### LINKED ISSUE #2250: Overhaul the help menu to also use bbcode markup
Now that we have the fancy bbcode for organelle process, the same treatment should be given to the help menu texts. We can now have the compound icons in there as well to make it visually more interesting and faster to associate the organelles with the compounds they handle. Also the input keys can be made to react to the key bindings.

### LINKED ISSUE #2303: Change help screen to use the old approach where help texts are manually placed on either side
Currently we have no control where the help text boxes goes on which side so it results with the help screen looking uneven (due to length difference) and possibly breaking text flow (sometimes separate boxes/sections relate to each other so they should be placed in the same side).

Kasterisk:

Moved back to the old way of manually sorting the help screen layout for flexibility, which will be kept that way until Thrivepedia eventually comes along to warrant a total overhaul. The help menu texts now use the custom rich text label so compound icons and keybinds are now supported in the texts which should them look more interesting.

As an addition, this includes improvement to the #2236 workaround by having the `CustomRichTextLabel` class handle that directly to make it more reliable and easily used anywhere without some unexpected bug happening.

**Related Issues**

Closes #2303 
Closes #2250 
Closes #1913 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


Kasterisk:
Ah, just noticed the localization removed a big portion/all of the help texts. Will look into this and see what's up.

hhyyrylainen:
Is that problem with the translations fixed?

Kasterisk:
Not yet, looks like this requires an update to the thrive pybabel specifically the json extractor script. Could you maybe update it to add `Message` to the keyword extraction? Since that will likely needs version updating, etc that I'm not fully familiar with.

hhyyrylainen:
That part is in pybabel-godot-thrive/json_extractor.py on line 124 it specifies the keys in the json that have translatable content as their value.

Kasterisk:
Yeah I figured, I could try tweaking the keys myself and open a pull request in the thrive pybabel repo I guess, but I believe that's as far as I could do.

hhyyrylainen:
You can locally install the new version with pip, so you can locally test that it works completely.
Btw, why does the format need to be like this? Why not go back to specifying the left and right hand side texts as an array? That is already supported by the text extractor.

Kasterisk:
Already opened a PR.

> Btw, why does the format need to be like this? Why not go back to specifying the left and right hand side texts as an array? That is already supported by the text extractor.

The help texts json also includes easter eggs and tips which doesn't use left and right format so I figured this would make it more general-use.

hhyyrylainen:
Thrive babel version 1.9 is now up on pypi. The install script that specifies the version should be updated in this PR, also, once I'm done making them, the CIConfiguration.yml format check step will need to be updated to use the new image.

FuJa0815 (approved):
Have not found any bugs

Kasterisk:
Updated to match the new compound bbcode syntax.

src/general/HelpScreen.cs
```diff
@@ -94,20 +94,18 @@ private void BuildHelpTexts(string category)
...
-            if (i < middleIndex)
+            if (text.Column == HelpText.TextColumn.Left)
             {
                 leftColumn.AddChild(helpBox);
             }
-            else
+            else if (text.Column == HelpText.TextColumn.Right)
             {
                 rightColumn.AddChild(helpBox);
             }
```
hhyyrylainen, line 111: I think there should be a GD.PrintErr as an else here warning that a help text doesn't have the column set.

hhyyrylainen (changes requested):
Playtesting works, and code also looks good, just one tiny comment and this can be merged after that.

src/general/HelpScreen.cs
Kasterisk: Added.

hhyyrylainen (approved):

PR #2385, opened 2021-07-01
Set Chunk Absorption offset to be absolute to avoid spinning behavior on unusually shaped cells.

### LINKED ISSUE #2378: Reduce the AI spinning around while stationary
It seems the new AI spins around while not moving, way too much, it should be reduced.

adQuid:

Changes chunk AI behavior to use an absolute offset instead of one based on engulf size, so that extremely wide cells don't miss the target.

**Related Issues**

fixes #2378 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


buckly90:
The AI still seems to spin around a lot while engulfing chunks.

I feel like it might be chasing the object while it is in the dissolving state, and only stops when the object is finally removed entirely.

hhyyrylainen:
I wonder how the old AI worked with that? Maybe it actually stopped moving when it reached the chunk position it tried to engulf?

adQuid:
I've gone ahead and set the AI to do that. It still turns around one time while engulfing, but at least it's not spinning like an idiot now.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -305,8 +305,15 @@ private void PursueAndConsumeChunks(Vector3 chunk, Random random)
             MoveWithRandomTurn(0.1f, 0.2f, random);
         }
 
-        // Always set target Position, for use later in AI
-        microbe.MovementDirection = new Vector3(0.0f, 0.0f, -Constants.AI_BASE_MOVEMENT);
+        // If this Microbe is right on top of the chunk, stop instead of spinning
+        if (DistanceFromMe(chunk) < 1.0f)
```
Kasterisk, line 309: Instead of an arbitrary number would this be better if this compares against the microbe's `Radius` property? I think that would lead to a behavior where cells would stop immediately once a chunk enters its membrane area.

adQuid: I'll play around with it, but I suspect that that would have the same problem that caused this bug in the first place: radius pretends that the microbe is a circle, but absorption is based on shape collision.

adQuid: ok, it worked.

hhyyrylainen: Well the radius is the *maximum* distance the membrane is from the microbe center. I guess it is possible to make a shape that this logic would break for. Perhaps the AI could slow down but still approach the chunk location?

adQuid: I would expect any non-zero speed to result in the spinning behavior (although maybe we could slow it down). I would also expect issues with very slow cells now looking like idiots slowing to a crawl right before grabbing a chunk.

I've tested using radious with a microbe that's just a horizontal line and it does work, which leads me to think that radious might be what's used to determine engulfing. Setting it back to 1.0f would be safest, but would result in that one-time 180 degree turn every time a microbe eats a chunk.

hhyyrylainen: I'm still wondering why that happens. Only if the chunk doesn't actually exist can there be a coordinate an AI cell wants to orbit. If the chunk actually exists, the cell would push it, causing the target position to also drift away...
I don't think using Radius here is good because otherwise the players can find a way to cheese the AI by making a really wide cell that stops before reaching a chunk.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -305,8 +305,15 @@ private void PursueAndConsumeChunks(Vector3 chunk, Random random)
             MoveWithRandomTurn(0.1f, 0.2f, random);
         }
 
-        // Always set target Position, for use later in AI
-        microbe.MovementDirection = new Vector3(0.0f, 0.0f, -Constants.AI_BASE_MOVEMENT);
+        // If this Microbe is right on top of the chunk, stop instead of spinning
+        if (DistanceFromMe(chunk) < microbe.Radius)
```
hhyyrylainen, line 309: I think `microbe.Radius` should be replaced with a constant defined in Constants, something like `MICROBE_ENGULF_CHUNK_STOP_DISTANCE = 0.8f`
My reasoning is that as most chunks have radius of 1, that's mostly fine, but single hex bacteria, will have a radious of 0.5, I'm a little scared that those cells might stop too soon, and 0.8 probably is low enough that the cells do get within that range during an AI update and can stop moving.
Other than that I'd like to see playtesting on this PR to ensure the AI no longer is seen spinning around.

hhyyrylainen (changes requested):

src/microbe_stage/MicrobeAI.cs
adQuid: True, but keep in mind that a microbe can't engulf anything until it is at least engulf size 2. That said, with a fixed stop distance I think any positive value will be the same, landing on the chunk and flipping once.

hhyyrylainen (approved):
This is fine by me code-wise, just needs playtesting to verify behaviour.

Nunez2196:
So I decided to test this PR using both a save file provided by adQuid on Discord and my own quick scenarios using free build. I did 15 runs of the provided save file on a clean build of master and then I did the same amount of runs with this PR. After 15 runs I can say not once did the spinning microbe problem occur on this PR where as it occurred a majority of the time on master using really wide cells. I didn't have any glitches or crashes that were new while playing this PR.

hhyyrylainen:
So I guess this has now been tested fully. I'll merge this.

PR #2353, opened 2021-06-19
Added menu shortcuts

84634E1A607A:
**Brief Description of What This PR Does**

Added ESC shortcut to go back to the upper one

**Related Issues**

close #2352 

**Progress Checklist**

Note: before starting this checklist the issue should be marked as non-draft.

- [x] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [ ] Initial code review passed
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author.

Issue #2326, opened 2021-06-14
Cell Process panel (and tutorial popups) is on top of menu


84634E1A607A:
Cell Process panel is on top of menu, making menu buttons unclickable.

To reproduce, open Cell Process menu and drag it to the middle of the screen. Then press escape, you will find the buttons unclickable.

hhyyrylainen:
![2021-06-14_08 40 19 7171](https://user-images.githubusercontent.com/3796411/121844135-2c267c80-ccec-11eb-90f8-61403c08aab7.png)

Can confirm, the process panel being open prevents mouse click events from going to the pause menu buttons.

84634E1A607A:
I think ProcessPanel, CheatMenu and EditorCheatMenu should be Panel instead of WindowDialog, and maybe add custom dragging handler to it.

hhyyrylainen:
> add custom dragging handler to it.

That's the hard part as resizing is also a feature in addition to dragging the panel around.

PR #2208, opened 2021-04-25
Add confirmation popup on exit and return to menu

ender-titan1:
closes #2156 
Popup doesn't show unless the last save was more than 10 seconds ago. (idk why the edits of PauseMenu.tscn are so weird, because all i did was add the timer and popup)

Issue #2020, opened 2021-01-09
The Tutorial Boxes Should Respect The Screen Borders


RacerBG:
**Describe the bug**
When moving the tutorial boxes with the mouse they are moving outside of the screen edges (borders).

**To Reproduce**
1. Drag the tutorial boxes over the edges (borders) of the screen.

**Expected behavior**
They should respect the screen edges (borders) and don't allow the mouse to move them outside of it.

**Screenshots**
![screen borders - tutorial](https://user-images.githubusercontent.com/68707872/104091801-85473b80-5288-11eb-8024-3fb818a2e5eb.png)

**System info (please complete the following information):**
- Thrive version [0.5.3.0]

hhyyrylainen:
They use Godot WindowDialog Control type. All the mouse drag and resizing logic is in Godot engine. So we can't do anything about that, this needs to be reported to Godot developers.

PR #2387, opened 2021-07-02
Added Dialog Input Helper Class

### LINKED ISSUE #1857: Create proper method to detect exclusive dialog open
the pause menu wants to know if an exclusive dialog is open. Right now it only detects an exclusive tutorial popup being open through a special method.

better solution to allowing exclusive dialogs work with the pause menu: https://github.com/Revolutionary-Games/Thrive/pull/1696#discussion_r503372192

Would be the best if there was a way to ask godot if there's an exclusive dialog open.

### LINKED ISSUE #2384: Make load game incompatibility popup prevent exiting with ESC
Exiting by pressing ESC when a popup is open from the load menu can leave the popup on screen:
https://github.com/Revolutionary-Games/Thrive/pull/2353#issuecomment-869483147

should be changed to be handled similarly to the options menu when there's unsaved changes to automatically close the popup when pressing ESC (and maybe it would be nice to also prevent the menu closing, so pressing just ESC could dismiss the incompatible save version popup, but this is an optional extra quality of life feature).

84634E1A607A:

Fixes #2384 

Added a helper class to handle all dialog escapes. Maybe fixes #1857 too, because now when a dialog is open the first escape closes it instead of opening the pause menu.

> > Also I found that closing the modified options menu with Esc doesn't close the menu completely and instead returning to the pause menu after confirming the unsaved changes prompt, not a problem but thought it's worth mentioning since you want the menu to be closed entirely by pressing escape and whether that should be fine.
> 
> I found that case as well, but as it would need a bunch of extra state tracking logic to close the underlying menu when the "unsaved changes" prompt is closed. Due to that extra complexity, I don't think it's that worth it to solve this.
> 
> _Originally posted by @hhyyrylainen in https://github.com/Revolutionary-Games/Thrive/issues/2353#issuecomment-871408091_

This is not solved due to ~~the complexity~~ my laziness. 

image

The handler doesn't apply to this dialog because when the key binding is in progress InputManager somehow doesn't work. ESC doesn't cause problem though.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


Kasterisk:
Overall works great, just one missing handling for the "Reset Inputs" warning popup.

84634E1A607A:
Okay, anyone look at this? I add a helper class which handles all probable dialog escapes. I think that it works fine with me at least for now.

simulation_parameters/Constants.cs
```diff
@@ -449,6 +449,8 @@ public static class Constants
 
     public const int SUBMENU_CANCEL_PRIORITY = -1;
 
+    public const int DIALOG_WINDOW_CANCEL_PRIORITY = 1;
```
FuJa0815, line 452: Can you add a summary here? State that this is the priority used for the input system.

src/general/DialogInputHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+                        || asDialog.Name == "OrganellePopup")
+                    {
+                        continue;
+                    }
+
+                    dialogs.Add(asDialog);
+                }
+
+                nodes.AddRange(currentNode.GetChildren().OfType<Node>());
+            }
```
FuJa0815, line 70: I think recursion and linq could clean up and shorten that code a lot

src/general/MainMenu.tscn
```diff
@@ -10,6 +10,7 @@
 [ext_resource path="res://assets/textures/gui/BG_Menu01.png" type="Texture" id=8]
 [ext_resource path="res://assets/textures/gui/BG_Menu03.png" type="Texture" id=9]
 [ext_resource path="res://assets/textures/gui/logo.png" type="Texture" id=10]
+[ext_resource path="res://src/general/DialogInputHelper.cs" type="Script" id=11]
```
FuJa0815, line 13: Instead of adding this Helper on various scenes, add this helper as an AutoLoad (Godot: Project>Project Settings...>AutoLoad).

This makes sure that there is only one instance of your class, which also allows you to remove all those isDuplicate checks

FuJa0815 (changes requested):

src/general/DialogInputHelper.cs
84634E1A607A: I don't prefer deep first search because it includes too much system-handled stack operations. As for linq, I haven't really learned how to use that...

src/general/DialogInputHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+                var currentNode = nodes[0];
+
+                nodes.RemoveAt(0);
+
+                if (currentNode is WindowDialog asDialog)
+                {
+                    // Skip ProcessPanel and CheatMenu and OrganellePopup
+                    if (asDialog.Name == "ProcessPanel"
+                        || asDialog.Name == "CheatMenu"
+                        || asDialog.Name == "OrganellePopup")
```
hhyyrylainen, line 61: I kinda don't like this code, feels dirty to me to walk the scene tree hoping to find some dialogs. And then especially those explicit name checks are very prone to breaking in the future.
I think a better design would be a new custom dialog base type that should be used everywhere (we can write in our style guide that plain Godot dialogs should not be used).

Kasterisk: I could add the universal custom dialog type (and perhaps some kind of exclusive dialog detection as well(?)) in https://github.com/Revolutionary-Games/Thrive/pull/2208 if possible or maybe in a separate dedicated branch instead. So this could just utilize that for the dialog escape handling.

84634E1A607A: > I think a better design would be a new custom dialog base type that should be used everywhere (we can write in our style guide that plain Godot dialogs should not be used).

I've tried this, we can add a WindowDialogWithInput base class, but many classes such as ConfirmationDialog in Godot are directly based from WindowDialog, which is unmodifiable. If we want to add a universal dialog class we must also add all derived classes as well. That's a lot of work.

hhyyrylainen: It's less work than to edit each place where a dialog is used, much easier to make custom dialog types for every Godot class and use those.

Kasterisk: With unified custom dialog type we can also specify some kind of UI animation behaviour that comes with it, for example a fade or scale tweening for when a popup is shown or hidden.

84634E1A607A:
I'm thinking of using an attribute instead

84634E1A607A:
Well I messed up. I'll close this and maybe open another.

hhyyrylainen:

84634E1A607A:
image
I cannot push because ruby finds "duplicate" codes... Anyone help?

hhyyrylainen:
Look at the duplicate fragments and the file locations they are at. Then refactor that code so that it uses an unified implementation so the duplicated code goes away, instead replaced by calls to the unified code.

hhyyrylainen:
For classes sharing functionality, I *think* partial classes might be the solution here.

84634E1A607A:
Of no use... And I don't think override methods can be put into a helper class...

hhyyrylainen:
That's basically as simple as it can get, but other methods that probably have code, you could do something like
```
public override void _Ready()
{
    SomeHelperClass.DoStuff(this);
}
```
That way the overall duplication amount goes down and the checker should let that _EnterTree be like that.

84634E1A607A:
Well, 
image
is of the same length...

hhyyrylainen:
If you post the entire complaint from the checker then I or someone else can probably help.

84634E1A607A:
```
Run check formatting.....................................................Failed
- hook id: check_formatting
- duration: 12.01s
- exit code: 2

[0;97;49mStarting formatting checks with the following checks: ["compile", "files", "inspectcode", "cleanupcode", "duplicatecode", "localization"][0m
JetBrains Inspect Code 2021.1.3
Running on AMD 64 in 64-bit mode, .NET runtime 4.0.30319.42000 under Microsoft Windows NT 10.0.19043.0
JetBrains Duplicates Finder 2021.1.3
Running on AMD 64 in 64-bit mode, .NET runtime 4.0.30319.42000 under Microsoft Windows NT 10.0.19043.0
JetBrains Cleanup Code 2021.1.3
Running on AMD 64 in 64-bit mode, .NET runtime 4.0.30319.42000 under Microsoft Windows NT 10.0.19043.0
Using toolset version 16.0 from ��C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin��
Using toolset version 16.0 from ��C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin��
Configuration: Debug, Platform: Any CPU
Collecting files to analyze:
  from project 'E:\source\repos\Thrive\Thrive.csproj'.
Using toolset version 16.0 from ��C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin��
278 files found to analyze.
Configuration: Debug, Platform: Any CPU
Duplicates report was written to E:\source\repos\Thrive\duplicate_results.xml
[0;31;49mFound duplicate with cost 109[0m
Fragment in file src\general\WindowDialogWithDefaultFunctions.cs
Lines 3--25
Fragment code:
/// <summary>
///   This is the WindowDialogs with custom functions.
/// </summary>
public class WindowDialogWithDefaultFunctions : WindowDialog
{
    public override void _EnterTree()
    {
        InputManager.RegisterReceiver(this);
        base._EnterTree();
    }

    public override void _ExitTree()
    {
        InputManager.UnregisterReceiver(this);
        base._ExitTree();
    }

    [RunOnKeyDown("ui_cancel", Priority = Constants.WINDOW_DIALOG_CANCEL_PRIORITY)]
    public bool OnEscapePressed()
    {
        return DialogHelper.DefaultDialogEscapeHandler(this);
    }
}
End of fragment
Fragment in file src\general\WindowDialogWithDefaultFunctions.cs
Lines 27--46
Fragment code:
public class AcceptDialogWithDefaultFunctions : AcceptDialog
{
    public override void _EnterTree()
    {
        InputManager.RegisterReceiver(this);
        base._EnterTree();
    }

    public override void _ExitTree()
    {
        InputManager.UnregisterReceiver(this);
        base._ExitTree();
    }

    [RunOnKeyDown("ui_cancel", Priority = Constants.WINDOW_DIALOG_CANCEL_PRIORITY)]
    public bool OnEscapePressed()
    {
        return DialogHelper.DefaultDialogEscapeHandler(this);
    }
}
End of fragment
Fragment in file src\general\WindowDialogWithDefaultFunctions.cs
Lines 48--67
Fragment code:
public class ConfirmationDialogWithDefaultFunctions : ConfirmationDialog
{
    public override void _EnterTree()
    {
        InputManager.RegisterReceiver(this);
        base._EnterTree();
    }

    public override void _ExitTree()
    {
        InputManager.UnregisterReceiver(this);
        base._ExitTree();
    }

    [RunOnKeyDown("ui_cancel", Priority = Constants.WINDOW_DIALOG_CANCEL_PRIORITY)]
    public bool OnEscapePressed()
    {
        return DialogHelper.DefaultDialogEscapeHandler(this);
    }
}
End of fragment
End of duplicate
[0;31;49mDuplicate finder found duplicates, see duplicate_results.xml[0m
```

I don't think they are duplicates though. Any suggentions?

hhyyrylainen:
I wonder if it is possible to use the same script for multiple scenes in Godot, that would solve the issue.
Or maybe I don't know some C# trick that is usable here.

The overall duplicate cost is only 109 so it's pretty close to the limit what duplicate code is allowed, so if nothing else is found a suppressing comment could be added to the top of all of these files to disable duplicate checks in them.

84634E1A607A:
Wait. 
image
In Godot these Dialog nodes can only use built-in types. If we want to use custom types we have to attach script... That's a lot of work, I don't like it...

src/general/MainMenu.tscn
84634E1A607A: Then the _Ready() only gets called once, making it miss all other dialogs.

hhyyrylainen:
We need to anyway. Each custom component we have *already* created requires a scene for easily placing in other scenes. This feature is no exception to that.

84634E1A607A:
Okay, all handlers are added. However the checkformatting.rb keeps saying that there're duplicates. 

> suppressing comment could be added 

Don't know how... Help needed...

hhyyrylainen:
A comment in the beginning of a file with `NO_DUPLICATE_CHECK` should work. KeyNames.cs already has that.

84634E1A607A:
Changing many dlgs to CustomedDialog changed its margin in editor, making it very strange. But in the game nothing seemed to happened. Hope it doesn't break things.

84634E1A607A:
When Tutorial dialogs close a button sound is played by default in its OnClosed function, and there's nothing I can do.

84634E1A607A:
Button RESET_INPUTS doesn't have a click sound.

hhyyrylainen:
> When Tutorial dialogs close a button sound is played by default in its OnClosed function, and there's nothing I can do.

That wasn't the case before, right? I'm sure there's some way to fix that...

84634E1A607A:
> > When Tutorial dialogs close a button sound is played by default in its OnClosed function, and there's nothing I can do.
> 
> That wasn't the case before, right? I'm sure there's some way to fix that...

Tutorial dialogs are not handled by esc system before. Other dialogs don't have a exit sound actually, but tutorial dialogs have.

84634E1A607A:
Maybe that sound can be canceled.

hhyyrylainen:
> Tutorial dialogs are not handled by esc system before. Other dialogs don't have a exit sound actually, but tutorial dialogs have.

Tutorial dialogs are meant to play the exit sound when the "X" in their corner is pressed with the mouse. In other cases they shouldn't make a sound. I think I got it working that way in the end and I'm not misremembering that...

84634E1A607A:
Well... You're right.
image
Now if the tutorial is closed automatically the sound isn't played; otherwise it is. The problem is that if you press escape you are not closing it "automatically".

hhyyrylainen:
Pressing escape *shouldn't* close the tutorial. There should be a separate tutorial dismiss button. Otherwise controller people would have problems, as they couldn't access the menu while a tutorial is ongoing that they might not want to dismiss.

84634E1A607A:
As you say.

hhyyrylainen:
There's a separate issue about the tutorials: https://github.com/Revolutionary-Games/Thrive/issues/1944

src/general/DialogHelper.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <returns>True if handled, false if not.</returns>
+    public static bool DefaultDialogEscapeHandler(WindowDialog dialog)
+    {
+        if (!dialog.Visible)
+        {
+            return false;
+        }
+
+        dialog.Hide();
+        return true;
```
Kasterisk, line 20: The code here can just be directly put inside the custom popup class. Also since normal popups already closes by themselves if you press any key (if I recall correctly) I'd amend the check on top by adding two more conditions that only exclusive _and_ escapable-allowed dialogs can be closed with escape, meaning there would be a bool like `AllowExclusiveCloseWithEscape` in `CustomDialog`.

Kasterisk (changes requested):
Warning, long post ahead! Lot to explain but I'll try making it as understandable as possible 🙂

I see there are multiple separate dialog classes with similar implementations, my suggestion is to consolidate them into one class named `CustomDialog` that inherits `AcceptDialog`, in it should be the escape handling etc. `AcceptDialog` is chosen instead of `WindowDialog` to take advantage of its auto-resize functionality (unless you're able to reimplement that functionality and have the custom class inherits `WindowDialog` instead for a maximum customizability down the line).

The `CustomDialog` class would later be a base class for more customized dialogs with their own dedicated node scenes (like modified default confirmation buttons or buttons added/removed to better match a given dialog context). The result of this design would be various types of custom accept dialog derived nodes with their own distinct purposes that can be organized into one folder inside `gui_common`.

Also, for the "is any exclusive dialog open" flag, best I can come up with is to have it inside the `CustomDialog` class as a static property with a private setter. The way it works is to have the bool set according to the dialog's visibility and whether it's exclusive by using the built-in `PopupExclusive` property, this could be done by using show/hide related signals connected to methods inside the class. One downside with this approach is that it doesn't support multiple visible exclusive popups, but I don't think there should be more than one exclusive dialog open anyway.

84634E1A607A:
Good thought, but I cannot delete that OK button even if I Hide() it, RemoveChild() and QueueFree(), so each dialog must have an OK button if I do that.

Kasterisk:
Have you tried `AcceptDialog.GetOk()` to access the button? In OptionsMenu there's example of an accept dialog that has its default Ok button made not visible through code.

84634E1A607A:
Works!

84634E1A607A:
But what's autoresize and how to auto center?

Kasterisk:
By auto-resize I mean the AcceptDialog's capability to automatically adjust its minimum rect size based on its contents min size and for the contents to automatically scale to the AcceptDialog's rect size. A bit confusing but you can try experimenting with WindowDialog by adding Container control nodes inside it and calling `PopupCenteredShrink` on the dialog to see the difference.

Also could you clarify what to auto-center? Doesn't Label have the Align property? Or if you mean other types of control nodes, there's CenterContainer which you could use to center nodes (relative to rect size inside the AcceptDialog node.

84634E1A607A:
Okay, thank you. Found that.

84634E1A607A:
Don't know if I should change all occurrences such as `private WindowDialog microbeWelcomeMessage;` to `CustomWindowDialog`. It isn't necessary now, but maybe afterwards.

Kasterisk:
So I've been looking a bit more into this and apparently I miss to see that the are `Viewport.GetModalStackTop` and `Viewport.GuiHasModelStack` available, this should have made things easier... Regardless, this means that active exclusive popup can be handled by GUICommon instead of static variables in individual custom dialogs.

Still, it is very useful to have a custom dialog implementation. If you don't mind, I took the liberty to overhaul the design a bit, so `CustomAcceptDialog` is now generalized as `CustomDialog` which is basically the reimplementation of `WindowDialog`, it functions similarly but with a couple of extra additions. I initially suggested using `AcceptDialog` as a base class but later thought it will be too limiting, instead with the new custom dialog it should be more suitable in the long run for general custom use and as a base class for any derived types. Along with that, this should address couple of issues we had with the built in WindowDialog such as #2020.

Took quite a bit, but I'm satisfied with how it looks (notice especially the titlebar):
2021-08-04_15 36 19 4098_cropped
To compare this is how it looks with the default WindowDialog which is a bit bland:

84634E1A607A:
Let me try on it.

84634E1A607A:
image
Here, 
image
Maybe should use RectSize (current size)

84634E1A607A:
And also can window size change affect popup?

Kasterisk:
That's a good find, should be fixed now. Thanks!

> And also can window size change affect popup?

I guess yeah, the Control's anchor setting still determines that automatically. The only manually handled case is when the popup goes out of the window border when dragged along the screen.

I'll do another pass of testing sometime soon, will mark this as a draft until I feel it's ready.

hhyyrylainen:
If possible it would be nice to fix the case where the tutorial window goes offscreen if you do something like:
- maximise the game
- wait for a tutorial to open
- Shrink the window back to 720p
- And now the tutorial popup right edge is off screen

If doing the opposite then the tutorial right edge no longer aligns with the screen.
I think the tutorial anchors are right, so something else strange goes on with that.

84634E1A607A:
I think this can be done by connect ViewPort's size_changed to a custom function

Kasterisk:
> If doing the opposite then the tutorial right edge no longer aligns with the screen.
I think the tutorial anchors are right, so something else strange goes on with that.

Hmm yeah, tried adjusting the anchors and that doesn't solve it. The dialogs being a top-level control is probably the main reason why it's able to go anywhere, the mentioned viewport signal approach to force the popups to stay onscreen is probably a good solution to try.

Kasterisk:
#1624 is now "partially" solved, popups still moves around but at least it doesn't go offscreen. However I don't think we can do much more about it, any solution to force the window dialogs to easily stay in place on window resize would certainly be very hacky in my opinion @hhyyrylainen.

84634E1A607A:
As for adjusting to window, sometimes not only size but also position is needed to be changed. I prefer
```C#
    private void OnWindowSizeChanged()
    {
        if (BoundToScreenArea)
        {
            var screenSize = GetViewport().GetVisibleRect().Size;
            var minSize = GetMinimumSize();
            var adjustedSize = new Vector2(RectSize);
            var adjustedPosition = new Vector2(RectPosition);

            var titleBarHeight = GetConstant("custom_titlebar_height", "WindowDialog");

            // First adjust position
            if (RectPosition.x + RectSize.x > screenSize.x)
            {
                adjustedPosition.x = Mathf.Max(screenSize.x - RectSize.x, 0);
            }

            // Then adjust size
            if (screenSize.x < RectSize.x)
            {
                adjustedSize.x = Mathf.Max(screenSize.x, minSize.x);
            }

            if (RectPosition.y + RectSize.y > screenSize.y)
            {
                adjustedPosition.y = Mathf.Max(screenSize.y - RectSize.y, titleBarHeight);
            }

            if (screenSize.y < RectSize.y + titleBarHeight)
            {
                adjustedSize.y = Mathf.Max(screenSize.y - titleBarHeight, minSize.y);
            }

            RectSize = adjustedSize;
            RectPosition = adjustedPosition;
        }
    }
```

Kasterisk:
Implemented a simplified version of it, the popup window's size should now clamped within viewport if enabled.

hhyyrylainen:
Is this now complete besides the merge conflicts?

Kasterisk:
Yeah I feel satisfied with how this is currently, probably still missing something but reviews are welcome of course.

84634E1A607A:
@Kasterisk  After merging the dialogs seems to have some margin problems. Will you maybe change it? Is it because I missed some font conflicts when merging?

Kasterisk:
Looks like there are incorrect resource IDs being set in thrive_theme.tres, I will fix these and double check merged files to see if anything is off.

84634E1A607A:
That merge nearly killed me. I really want to know what to do when a sequentially numbered resource file has conflicts...

Kasterisk:
There aren't much really, I guess you gotta be familiar enough with the resource files to know which resource IDs and node order should go where and which should be kept. Basically you have to go through and correct every conflicted parts _manually_, at least that's how I done it, I'm open for better ways though lol.

Fortunately Godot will have some kind of unique ID for resources in 4.0 which should at least address the sequentially numbered IDs being a major pain for version control.

hhyyrylainen:
With my merge conflicts I've had a txt file open where I record the conflicting ids and what they are on each side of the merge conflict. Then I go through the entire file fixing merge conflicts one by one and finally I refer back to my list of ids I saved to that text file to make sure everything is right. Usually it's just a matter of flipping some ids around. One time I did have to start using a new id number for the changes from one conflict. 
But yeah I don't know a better way, that's very error prone and tedious, but it can be done.

sirdanforth:
Tested to make sure that the escape key still works with all the dialogue and I did not encounter any bugs pertaining to this. However, I did come across two bugs involving the tutorial pop-ups but I'm don't know if they are related or not.

1.) If the tutorial pop-up is behind the pause menu, I am unable to click on any of the options the pop-up is directly behind. This happened in both the editor mode and while playing.

2.) You can interact with the tutorial pop-up when paused (move it around and resize it). Happened in both the editor mode and while playing.

I am able to interact with menu here


Tutorial pop-up behind menu


Tutorial pop-up behind menu while playing

84634E1A607A:
> 1.) If the tutorial pop-up is behind the pause menu, I am unable to click on any of the options the pop-up is directly behind. This happened in both the editor mode and while playing.

Known Issue: #2326 (Apply to all dialogs (popups))

> 2.) You can interact with the tutorial pop-up when paused (move it around and resize it). Happened in both the editor mode and while playing.

Maybe a new one? Can it be reproduced in Master?

Kasterisk:
I think the tutorial popup being interactable in pause menu is interlinked with issue #2326 with the underlying problem being popups seem to ignoring input order.

> I can try to test this bug on the master branch but first I need to know.... The master branch is this right: https://github.com/Revolutionary-Games/Thrive

Yeah, that's the master branch.

sirdanforth:
> I think the tutorial popup being interactable in pause menu is interlinked with issue #2326 with the underlying problem being popups seem to ignoring input order.
> 
> > I can try to test this bug on the master branch but first I need to know.... The master branch is this right: https://github.com/Revolutionary-Games/Thrive
> 
> Yeah, that's the master branch.

I was able to interact with the tutorial pop-up in the master branch as well. Hope this helps!

Kasterisk:
@hhyyrylainen Feel free to take a look, just committed the latest changes and I'm happy with how this is currently.

The dialogs GUI input order issue can probably be solved in another PR.

hhyyrylainen:
@Kasterisk could you make an approving review on this so that it makes your previous "changes requested" to go away?

src/tutorial/microbe_stage/MicrobeTutorialGUI.cs
```diff
@@ -81,7 +81,7 @@ public bool MicrobeWelcomeVisible
             }
             else
             {
-                microbeWelcomeMessage.Visible = false;
+                microbeWelcomeMessage.ClosePopup();
```
hhyyrylainen, line 84: Why don't these work with the old approach anymore?
Even if the Visible = false was modified to call this.
Anyway if there's no way to make that modification, then what can be we do to reduce the risk that someone accidentally uses `Visible = false`?

src/tutorial/TutorialPhase.cs
```diff
@@ -22,12 +22,6 @@ public abstract class TutorialPhase
     [JsonProperty]
     public bool HandlesEvents { get; protected set; } = true;
 
-    /// <summary>
-    ///   True when this tutorial uses an exclusive popup preventing all other GUI from being interacted with
-    /// </summary>
-    [JsonProperty]
-    public bool Exclusive { get; protected set; }
```
hhyyrylainen, line 29: Related to my earlier comment, I guess removing this makes sense of the dialog type already handles being exclusive and reporting that...

src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -1233,8 +1233,6 @@ internal void OnFinishEditingClicked()
 
     internal void ConfirmFinishEditingPressed()
     {
-        GUICommon.Instance.PlayButtonPressSound();
-
```
hhyyrylainen, line 1237: Why was this removed?

src/gui_common/QuickLoadHandler.tscn
```diff
@@ -10,10 +11,12 @@ script = ExtResource( 1 )
 __meta__ = {
 "_edit_use_anchors_": false
 }
-DifferentVersionDialogPath = NodePath("DifferentVersionDialog")
+DifferentVersionDialogPath = NodePath("CustomConfirmationDialog")
```
hhyyrylainen, line 14: Why was this node renamed? It used to be specific to what it actually did, instead of just being the class name.

src/gui_common/CustomDialog.cs
```diff
@@ -0,0 +1,494 @@
...
+                Mathf.Max(childMinSize.x, contentSize.x),
+                Mathf.Max(childMinSize.y, contentSize.y));
+        }
+
+        // Redecide whether the largest rect is the default elements' or the contents'
+        return new Vector2(Mathf.Max(2 * buttonArea.GetValueOrDefault() + titleWidth,
+            contentSize.x + margin * 2), contentSize.y + margin * 2);
+    }
+
+    public void ClosePopup()
```
hhyyrylainen, line 252: So currently this does nothing?
But if/when something is done here, it would be super awesome to automatically be able to trigger that from just Hide or `Visible = false` call

hhyyrylainen (changes requested):
didn't have energy to look at all this yet, but here's some of the first things I noticed

Kasterisk (approved):

src/tutorial/TutorialPhase.cs
Kasterisk: Yeah, I think the exclusive check in GUICommon is already inclusive.

src/gui_common/CustomDialog.cs
Kasterisk: This method still needs to be implemented as I haven't got around with animating popup closing. Overriding the `Visible` property won't hurt, I guess it's worth to try.

src/gui_common/QuickLoadHandler.tscn
Kasterisk: Must have forgot to rename this...

src/microbe_stage/editor/MicrobeEditorGUI.cs
Kasterisk: Readded.

doc/style_guide.md
```diff
@@ -303,6 +303,11 @@ Godot usage
 - You should follow general GUI standards in designing UI. Use widgets
   that are meant for whatever kind of interaction you are designing.
 
+- When adding window dialogs to the game, consider using the
+  `CustomDialog` type rather than the built-in `WindowDialog` to ensure
+  consistency across the GUIs. This is because the custom implementation
```
hhyyrylainen, line 308:
```suggestion
  consistency across the GUI. This is because the custom implementation
```

simulation_parameters/Constants.cs
```diff
@@ -481,6 +481,11 @@ public static class Constants
 
     public const int SUBMENU_CANCEL_PRIORITY = -1;
 
+    /// <summary>
+    ///   Popups have a highest priority to ensure it can react first.
```
hhyyrylainen, line 485:
```suggestion
    ///   Popups have a highest priority to ensure they can react first.
```

src/gui_common/dialogs/TutorialDialog.cs
```diff
@@ -0,0 +1,32 @@
...
+/// </summary>
+public class TutorialDialog : CustomDialog
+{
+    private Tween tween;
+
+    /// <summary>
+    ///   Tweakable delay to make tutorial sequences flow more naturally.
+    /// </summary>
+    [Export]
+    public float ShowDelay { get; set; } = 1.0f;
```
hhyyrylainen, line 14: I think this should be reduced to 0.7 because with this much delay I think the editor tutorials don't popup fast enough when going through the editor tabs.

src/gui_common/GUICommon.cs
```diff
@@ -69,6 +74,37 @@ public static void SmoothlyUpdateBar(TextureProgress bar, float target, float de
         bar.Value = MathUtils.Lerp((float)bar.Value, target, weight, 1.0f / (float)bar.MaxValue);
     }
 
+    /// <summary>
+    ///   Closes any currently active exclusive modal popup.
```
hhyyrylainen, line 78:
```suggestion
    ///   Closes any currently active exclusive modal popups.
```

src/gui_common/GUICommon.cs
```diff
@@ -42,6 +42,11 @@ private GUICommon()
     /// </summary>
     public Tween Tween { get; }
 
+    /// <summary>
+    ///   Checks whether the top-most modal on stack is an exclusive popup.
```
hhyyrylainen, line 46:
```suggestion
    ///   Checks whether the top-most modal on the modal stack is an exclusive popup.
```

src/gui_common/CustomDialog.cs
```diff
@@ -0,0 +1,489 @@
...
+    public bool Resizable { get; set; }
+
+    /// <summary>
+    ///   If true, the window's position is clamped inside the screen so it doesn't go out of bounds.
+    /// </summary>
+    [Export]
+    public bool BoundToScreenArea { get; set; } = true;
+
+    [Export]
+    public bool ExclusiveAllowCloseOnEscape { get; private set; } = true;
```
hhyyrylainen, line 68: Shouldn't the set be public here as this is an export property (meant to be set from the outside)?

src/gui_common/CustomDialog.cs
```diff
@@ -0,0 +1,489 @@
...
+
+        DrawStyleBox(titleBarPanel, new Rect2(
+            new Vector2(3, -titleBarHeight + 3), new Vector2(RectSize.x - 6, titleBarHeight - 3)));
+
+        // Draw title in the title bar
+        var titleFont = GetFont("custom_title_font", "WindowDialog");
+        var titleHeight = GetConstant("custom_title_height", "WindowDialog");
+        var titleColor = GetColor("custom_title_color", "WindowDialog");
+
+        var translated = TranslationServer.Translate(WindowTitle);
```
hhyyrylainen, line 137: I think this should be cached when the window title is updated and not called on each draw.

src/gui_common/CustomDialog.cs
```diff
@@ -0,0 +1,489 @@
...
+    /// <summary>
+    ///   The text displayed in the window's title bar.
+    /// </summary>
+    [Export]
+    public string WindowTitle
+    {
+        get => windowTitle;
+        set
+        {
+            windowTitle = value;
```
hhyyrylainen, line 49: I think it would be good to have an early return if `windowTitle == value`

src/gui_common/CustomDialog.cs
```diff
@@ -0,0 +1,489 @@
...
+                break;
+            }
+        }
+    }
+
+    public override void _Draw()
+    {
+        var panel = GetStylebox("custom_panel", "WindowDialog");
+        var titleBarPanel = GetStylebox("custom_titlebar", "WindowDialog");
+        var titleBarHeight = GetConstant("custom_titlebar_height", "WindowDialog");
```
hhyyrylainen, line 123: Should we maybe cache these constant retrieves? Each one of them causes a call into the native Godot code, which is probably many times slower than calling C# code.

src/gui_common/CustomDialog.cs
```diff
@@ -0,0 +1,489 @@
...
+
+    /// <summary>
+    ///   This is overriden so mouse position could take the titlebar into account due to it being drawn
+    ///   outside of the normal Control's rect bounds.
+    /// </summary>
+    public override bool HasPoint(Vector2 point)
+    {
+        var rect = new Rect2(Vector2.Zero, RectSize);
+
+        var titleBarHeight = GetConstant("custom_titlebar_height", "WindowDialog");
```
hhyyrylainen, line 198: Especially here as I think this HasPoint might get called for each single mouse movement...

src/gui_common/CustomDialog.cs
```diff
@@ -0,0 +1,489 @@
...
+
+        // Enlarge upwards for title bar
+        var adjustedRect = new Rect2(
+            new Vector2(rect.Position.x, rect.Position.y - titleBarHeight),
+            new Vector2(rect.Size.x, rect.Size.y + titleBarHeight));
+
+        // Inflate by the resizable border thickness
+        if (Resizable)
+        {
+            var scaleborderSize = GetConstant("custom_scaleborder_size", "WindowDialog");
```
hhyyrylainen, line 208:
```suggestion
            var scaleBorderSize = GetConstant("custom_scaleborder_size", "WindowDialog");
```

hhyyrylainen: Can we rename the string in GetConstant? If we can't then scaleborder needs to be added to the list of words (and my change suggestion here can be ignored).

src/gui_common/CustomDialog.cs
```diff
@@ -0,0 +1,489 @@
...
+                continue;
+
+            var childMinSize = child.GetCombinedMinimumSize();
+
+            contentSize = new Vector2(
+                Mathf.Max(childMinSize.x, contentSize.x),
+                Mathf.Max(childMinSize.y, contentSize.y));
+        }
+
+        // Redecide whether the largest rect is the default elements' or the contents'
```
hhyyrylainen, line 247:
```suggestion
        // Re-decide whether the largest rect is the default elements' or the contents'
```

src/gui_common/CustomDialog.cs
```diff
@@ -0,0 +1,489 @@
...
+    /// <summary>
+    ///   Evaluates what kind of drag type is being done on the window based on the current mouse position.
+    /// </summary>
+    private DragType DragHitTest(Vector2 position)
+    {
+        var result = DragType.None;
+
+        if (Resizable)
+        {
+            var scaleborderSize = GetConstant("custom_scaleborder_size", "WindowDialog");
```
hhyyrylainen, line 278: Rider also reports typos here...

src/gui_common/CustomDialog.cs
```diff
@@ -0,0 +1,489 @@
...
+                result = DragType.ResizeTop;
+            }
+            else if (position.y >= (RectSize.y - scaleborderSize))
+            {
+                result = DragType.ResizeBottom;
+            }
+
+            if (position.x < scaleborderSize)
+            {
+                result |= DragType.ResizeLeft;
```
hhyyrylainen, line 292: Is this code derived from Godot code or our own reimplementation?
I'm slightly getting the former vibe with how much complex logic there is in this class...
If that's the case then we need to more explicitly put the copyright notice in this file from Godot and say that this is a modified version.

src/saving/SaveStatusOverlay.cs
```diff
@@ -86,13 +73,7 @@ public void ShowMessage(string message, float visibleTime = 0.7f)
...
         Action onClosed = null)
     {
-        errorDialog.WindowTitle = title;
-        extraDescriptionLabel.Text = message;
-        exceptionLabel.Text = exception;
-        errorDialog.PopupCenteredShrink();
-
-        onDialogDismissReturnToMenu = returnToMenu;
-        onDialogCloseCallback = onClosed;
+        errorDialog.ShowError(title, message, exception, returnToMenu, onClosed);
```
hhyyrylainen, line 76: This'll have a merge conflict with the save disallowed when extinct PR I opened today (there needs to be a feature to hide the copy exception button in the error dialog).

hhyyrylainen (changes requested):
Finally got around to doing basically a full review. All the basics are good here, though I have some concerns. First is that when playing through the editor tutorial I got a massive lag spike and this error:
```
Failed method: VScrollBar:_update_callback target ID: 32711
TOTAL BYTES: 4194288
NULL count: 0
CALL : 174758
CALL _update_callback: 1
CALL _update_minimum_size: 2
CALL _sort_children: 1
ERROR: push_call: Message queue out of memory. Try increasing 'memory/limits/message_queue/max_size_kb' in project settings.
   At: core/message_queue.cpp:55.
Failed method: HScrollBar:_update_callback target ID: 32710
TOTAL BYTES: 4194288
NULL count: 0
CALL : 174758
CALL _update_callback: 1
CALL _update_minimum_size: 2
CALL _sort_children: 1
ERROR: push_call: Message queue out of memory. Try increasing 'memory/limits/message_queue/max_size_kb' in project settings.
   At: core/message_queue.cpp:55.
Failed method: Label:_update_callback target ID: 32712
TOTAL BYTES: 4194288
NULL count: 0
CALL : 174760
CALL _update_minimum_size: 1
CALL _sort_children: 1
ERROR: push_call: Message queue out of memory. Try increasing 'memory/limits/message_queue/max_size_kb' in project settings.
   At: core/message_queue.cpp:55.
Failed method: ScrollContainer:_update_callback target ID: 32709
TOTAL BYTES: 4194288
NULL count: 0
CALL : 174760
CALL _update_minimum_size: 1
CALL _sort_children: 1
ERROR: push_call: Message queue out of memory. Try increasing 'memory/limits/message_queue/max_size_kb' in project settings.
   At: core/message_queue.cpp:55.
Failed method: ScrollContainer:_update_minimum_size target ID: 32709
TOTAL BYTES: 4194288
NULL count: 0
CALL : 174761
CALL _update_minimum_size: 1
ERROR: push_call: Message queue out of memory. Try increasing 'memory/limits/message_queue/max_size_kb' in project settings.
   At: core/message_queue.cpp:55.
Failed method: ScrollContainer:_sort_children target ID: 32709
TOTAL BYTES: 4194288
NULL count: 0
CALL : 174761
CALL _update_minimum_size: 1
ERROR: push_call: Message queue out of memory. Try increasing 'memory/limits/message_queue/max_size_kb' in project settings.
   At: core/message_queue.cpp:55.
Jukebox: starting track: res://assets/sounds/microbe-editor-theme-5.ogg position: 0
Jukebox: starting track: res://assets/sounds/microbe-editor-theme-3.ogg position: 0
Jukebox: starting track: res://assets/sounds/microbe-editor-theme-1.ogg position: 0
```

Also I find that the dialog buttons are slightly smaller font, which I don't like for readability. Also I find that the bold (?) titles for the windows are actually a bit harder to read, especially the error dialogs where you would want to actually read them.

Compare:


With how it used to be:


Also the button order doesn't seem to be OS specific anymore, which is also a regression here (Godot seems to automatically know which way around the OK and Cancel should be on a specific platform).

src/gui_common/CustomDialog.cs
Kasterisk: Yes, though a heavily modified one. Will add it.

src/saving/SaveStatusOverlay.cs
Kasterisk: Fixed the merge conflict.

Kasterisk:
Did all the requested changes. Still don't know if that error still occurs, further testing may be needed.

84634E1A607A:
For me it just crashed when I tried to start a new game.

hhyyrylainen (changes requested):
The system specific ordering for Ok and cancel buttons is still not working.

PR #2388, opened 2021-07-05
Ai tweaks over run & tumble

Maxonovien:

This PR tweaks and improves the behavior of AI for Run & Tumble algorithm (used whenever no non-rivaled chunk/prey/predator are nearby), with following changes:
1. It considers iron similarly to glucose (if they are of any use), thus looking for ATP related compound if one of them is low.
2. Low amount defined as 1/2 capacity, to avoid cells only looking for vital (ATP-producing) compounds when dying already
3. Fixed gradient computation from absorbed compounds, as only those absorbed during this tick were considered instead of the difference with those absorbed the tick before.
4. Added priority weight for compound search to prioritize collect of lacking components (1-stored/capacity).
5. Detection threshold to avoid being stuck on traces deposit (**no effect** as set to 0, but may be changed to a relevant value/formula (e.g. storage-related?))

**Related Issues**

This PR is not related to any established issue.

The current AI is oddly behaving, leading to death while resources are present. This PR first intends to fix this (especially for iron/sulfide hydrogen builds which keep dying) for both immersion and gameplay. (Bullet points 1; 2; 4)

Additionally, it fixes a likely bug (not-reported) in the gradient computation algorithm (which was previously computed as the mere quantity of absorbed compounds, instead of a difference). (3) As a result, the AI tends to stay more within clouds to absorb compounds.

It also replaces the comparison of the gradient to 0 by a threshold (5), to ease readability and change to this threshold - should the need arise (e.g. massive cells wandering around a tiny cloud and dying from this). The threshold is still set to zero, and thus does not change game experience.

Overall, this PR is believed to improve game experience (by intended game lines) through minor fixes.
Note: this does not fix all issues related to run & tumble (resting rather than moving, prioritizing compounds depending on their usage...) ; such changes are more fundamentally different and thus left to further discussion.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/microbe_stage/MicrobeAI.cs
```diff
@@ -407,6 +422,40 @@ private void RunAndTumble(Random random)
...
+
+        // If this microbe lacks vital compounds don't bother with ammonia and phosphorous
+        // This algorithm doesn't try to determine if iron and sulfuric acid is useful to this microbe
+        foreach (var compound in usefulVitalCompounds)
+        {
+            if (microbe.Compounds.GetCompoundAmount(compound.Key) < 0.5f * microbe.Compounds.Capacity)
+            {
+                usefulCompounds = usefulCompounds.Where(x => x.Key != ammonia && x.Key != phosphates).ToList();
+            }
+        }
```
adQuid, line 444: Using a .Any statement would simplify this and increase clarity. Roughly:

if ( usefulVitalCompounds.Any(microbe.Compounds.GetCompoundAmount(compound.Key) < 0.5f * microbe.Compounds.Capacity)){
  usefulCompounds = usefulCompounds.Where(x => x.Key != ammonia && x.Key != phosphates).ToList();
}

src/microbe_stage/MicrobeAI.cs
```diff
@@ -65,7 +70,8 @@ public void Think(float delta, Random random, MicrobeAICommonData data)
 
         ChooseActions(random, data);
 
-        // Clear the absorbed compounds for run and rumble
+        // Store & clear the absorbed compounds for run and rumble
+        previouslyAbsorbedCompounds = new Dictionary<Compound, float>(microbe.TotalAbsorbedCompounds);
```
adQuid, line 74: Since this is always set as a clone of TotalAbsorbedCompounds, is this nessicary? I see no place where this dictionary is modified.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -370,34 +376,43 @@ private void RunAndTumble(Random random)
...
-        }
+        var compoundsPriority = ComputeCompoundsPriorities(usefulCompounds.ToList());
 
         float compoundDifference = 0.0f;
-        foreach (var compound in usefulCompounds)
+        foreach (var compoundPriority in compoundsPriority.ToList())
         {
-            compoundDifference += compound.Value;
+            // Note this is about absorbed quantities (which is all microbe has access to), not real ones
+            float quantityDifference = 0.0f;
```
adQuid, line 385: I took a while to pick up on the difference between quantityDifference and compoundDifference, which is a good sign these need to be renamed more clearly.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -38,13 +39,17 @@ public class MicrobeAI
     [JsonProperty]
     private float pursuitThreshold;
 
+    private Dictionary<Compound, float> previouslyAbsorbedCompounds;
```
adQuid, line 42: I just found out that this doesn't NEED a [JsonProperty] annotation for serialization, but to avoid confusing a future dev I would like to see these all having the annotation, or having them all removed (which maybe leads to more complications?) if you want to earmark spring cleaning into this PR.

adQuid (changes requested):
I see a couple clarity issues, and with such a small behavior change I'll need to do a lot of rounds of testing to see if it really changes behavior in a positive way, but overall I like it. The algorithm seems smart, and it's a light change that knows what it wants to do.

src/microbe_stage/MicrobeAI.cs
Maxonovien: It is, actually, because TotalAbsorbedCompound is reset right afterwards (l.75).
This clone is here to keep the information in memory for one decision loop, because it will be used for next iteration to computation.

src/microbe_stage/MicrobeAI.cs
Maxonovien: I wondered about the annotation, but I must confess I don't know these at all in C#. I thus left it out, in order not to break anything. Should I put a [JsonProperty] or a [JsonIgnore]?

From what I get it seems to affect saves (made as JSON files), therefore a [JsonIgnore] could make the gradient wrong for one execution at file load. Yet this is likely not a major issue as overall movement depends on many executions. I'm really unsure about this one.

src/microbe_stage/MicrobeAI.cs
Maxonovien: Agreed. I kept compoundDifference name legacy, but it seems to do more harm than good, because I don't see how to make quantityDifference clearer without renaming compondDifference aswell.

src/microbe_stage/MicrobeAI.cs
adQuid: Hmm, ok. I vaguely recall that; probably a future refactor candidate...

src/microbe_stage/MicrobeAI.cs
adQuid: No name is sacred. Code without mercy.

src/microbe_stage/MicrobeAI.cs
adQuid: In this case, you will want JsonProperty.

Maxonovien: I did so, and tested to be sure. It seems to all work well with saves!

src/microbe_stage/MicrobeAI.cs
```diff
@@ -370,34 +377,43 @@ private void RunAndTumble(Random random)
         microbe.State = Microbe.MicrobeState.Normal;
 
         var usefulCompounds = microbe.TotalAbsorbedCompounds.Where(x => microbe.Compounds.IsUseful(x.Key));
+        var compoundsPriority = ComputeCompoundsPriorities(usefulCompounds.ToList());
```
adQuid, line 380: Could we call this compoundWeights instead of compoundsPriority? Ultimately this is more of a weights-based algorithm than any kind of sort.

Maxonovien: compoundWeights makes more sense for the var here (because context) indeed, but I feel like the function name ComputeCompoundsPriorities would not describe well its computations. 

Changing one but not the other is setting me off a bit, I have mixed feelings over this one.

adQuid: Personally I feel like changing both names solves the problem. Take your time coming up with something you're happy with. I just read "priority" and thought "sorting", which isn't being done here.

Maxonovien: I get the confusion point here, I think you are right on the issue overall. I'll see what I can find after a good rest.

src/microbe_stage/MicrobeAI.cs
hhyyrylainen: As this is a private *field* this does need the JsonProperty annotation to be saved. Only public properties with both public set and get, are automatically assumed to be JsonProperty, all other stuff that needs saving, needs that explicit property.  Also conversely, if there's a public property that shouldn't be saved, it needs the JsonIgnore property.

src/microbe_stage/MicrobeAI.cs
hhyyrylainen: Would it be possible to modify the behaviour so that a copy of the data is not needed? The AI can decide when it wants to clear the absorbed compounds data. If this is kept as is, maybe this should be modified to not allocate new dictionaries all the time, instead clearing previouslyAbsorbedCompounds and adding the data from TotalAbsorbedCompounds into that existing object.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -370,43 +377,86 @@ private void RunAndTumble(Random random)
...
+    /// </summary>
+    /// <returns> A dictionary of weights for each (useful) compound, to be used in search algorithms.</returns>
+    private Dictionary<Compound, float> ComputeCompoundsSearchWeights(
+        List<KeyValuePair<Compound, float>> usefulCompounds)
+    {
+        // Vital compounds are *direct* ATP producers
+        var usefulVitalCompounds = usefulCompounds.Where(x => x.Key == glucose || x.Key == iron);
+
+        // If this microbe lacks vital compounds don't bother with ammonia and phosphorous
+        // This algorithm doesn't try to determine if iron and sulfuric acid is useful to this microbe
```
hhyyrylainen, line 440: I don't think sulfuric acid is a thing in the game...
Maybe this should say "hydrogen sulfide"?

src/microbe_stage/MicrobeAI.cs
Maxonovien: I don't really get this point : "The AI can decide when it wants to clear the absorbed compounds data."
If the AI is to compute a gradient, they need to have access to value from two ticks. I think a buffer structure is required then (at least I do not see how to circumvene this).

Removing reallocation seems definitely possible however, I'll dive onto it. Is there an interest other than memory usage?

hhyyrylainen:
> Fixed gradient computation from absorbed compounds, as only those absorbed during this tick were considered instead of the difference with those absorbed the tick before.

Is this actually fixed?
Because the previous logic (as far as I remember) is as follows:
- The AI uses the total absorbed compounds to calculate how much stuff it has picked up
- It does the AI stuff
- Then it resets the absorbed compounds
- During the next update the Microbe absorb update step gathers compounds and updates total absorbed compounds

So from that logic it should follow that the total absorbed compounds already tracks how much the microbe has absorbed since the last AI update.

src/microbe_stage/MicrobeAI.cs
hhyyrylainen: See my comment I posted on this PR itself, but basically, as the absorbed compounds is cleared each AI update, it *already* contains the delta relative to the previous update (as only the newly absorbed compound amounts are stored in that dictionary).
So the information on how much the AI absorbed during the last AI update cycle should already be computed in that variable and be ready to use.

Maxonovien:
> 
> 
> > Fixed gradient computation from absorbed compounds, as only those absorbed during this tick were considered instead of the difference with those absorbed the tick before.
> 
> Is this actually fixed?
> Because the previous logic (as far as I remember) is as follows:
> 
>     * The AI uses the total absorbed compounds to calculate how much stuff it has picked up
> 
>     * It does the AI stuff
> 
>     * Then it resets the absorbed compounds
> 
>     * During the next update the Microbe absorb update step gathers compounds and updates total absorbed compounds
> 
> 
> So from that logic it should follow that the total absorbed compounds already tracks how much the microbe has absorbed since the last AI update.

Indeed, but it does not compare it to previous absorptions, which is required to know if you are moving to places where you absorb more and more (along the gradient).
In terms of maths, absorption between two ticks gives you the integral of the compound density function, not the gradient!

hhyyrylainen:
So you want *in addition to* knowing the absorption during the last update, to know the one previous to that?
I think I get it now. What about using just a single variable to store that oldest absorption amount? That way it is much cheaper to update it and takes less memory to store.

Maxonovien:
> 
> 
> So you want _in addition to_ knowing the absorption during the last update, to know the one previous to that?
> I think I get it now. What about using just a single variable to store that oldest absorption amount? That way it is much cheaper to update it and takes less memory to store.

Yes, I want and need both of them! I started at first by using a single memory value indeed, but as variable weights were added to the computation, I decided to store the absorbed value by compound to not have computations wronged by weight changes (it seemed to help the AI behavior in-game as well). 

There might be a way to compute correction from the memory value , but it would require knowing previous weights (hence previously stored amounts, hence absorbed and _used_ amounts last step). Also, this correction would apply previous priorities to the current computed value, while it should be the contrary. So, while there might be solutions, they seem quite complex for little gain (as they will also use memory & time, although perhaps less - I don't know enough about C# & Godot operations cost to tell).

hhyyrylainen:
I'm not the most experienced with C# either, but I think allocations are cheap until they aren't, as when the garbage collector needs to kick in it will eat some performance. So as far as possible we should try to keep down the amount of garbage objects generated during each game update.

Maxonovien:
Agreed, although here it should have been changed to no longer allocate new dictionaries. I'll add a remark about the use of a compound-value dictionary rather than a single value though, because, it is still a bit more performance consuming. 

As stated before, I see no better alternative though, and the (small) cost seems reasonable to me given the feature. But it should definitely be improved if possible!

adQuid (approved):

adQuid:
I know you had discussed further tweeks you would like to experiment with, but if those don't end up happening I'm happy to see the code merge as it stands.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -38,13 +39,20 @@ public class MicrobeAI
     [JsonProperty]
     private float pursuitThreshold;
 
+    // Memory of the previous absorption step is required to compute gradient
+    // Values dictionary rather than single value as they will be combined with variable weights.
```
hhyyrylainen, line 43: This should be an xml comment.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -65,7 +73,13 @@ public void Think(float delta, Random random, MicrobeAICommonData data)
 
         ChooseActions(random, data);
 
-        // Clear the absorbed compounds for run and rumble
+        // Store & clear the absorbed compounds for run and rumble
+        previouslyAbsorbedCompounds.Clear();
+        foreach (var compound in microbe.TotalAbsorbedCompounds.ToList())
```
hhyyrylainen, line 78: Why is this ToList call here? This unnecessary allocates a duplicate of data. You should work directly on the keypairs in the dictionary.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -370,43 +384,86 @@ private void RunAndTumble(Random random)
...
 
-        // If this microbe lacks glucose, don't bother with ammonia and phosphorous
-        // This algorithm doesn't try to determine if iron and sulfuric acid is useful to this microbe
-        if (microbe.Compounds.GetCompoundAmount(glucose) < 0.5f)
+        float gradientValue = 0.0f;
+        foreach (var compoundWeight in compoundsWeights.ToList())
         {
-            usefulCompounds = usefulCompounds.Where(x => x.Key != ammonia && x.Key != phosphates);
-        }
+            // Note this is about absorbed quantities (which is all microbe has access to), not real ones
```
hhyyrylainen, line 392: I'd flip this comment around, because the storage in microbes is the "real" compound amounts processes etc. use these amounts. The amounts in clouds are pretty arbitrary in contrast.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -370,43 +384,86 @@ private void RunAndTumble(Random random)
...
+            float compoundDifference = 0.0f;
 
-        float compoundDifference = 0.0f;
-        foreach (var compound in usefulCompounds)
-        {
-            compoundDifference += compound.Value;
+            // No need to add a difference if compound was not absorbed
+            if (microbe.TotalAbsorbedCompounds.ContainsKey(compoundWeight.Key))
```
hhyyrylainen, line 396: Use try get value instead of checking for key existance and then looking up the value separately.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -370,43 +384,86 @@ private void RunAndTumble(Random random)
...
+                compoundDifference += microbe.TotalAbsorbedCompounds[compoundWeight.Key];
+
+            // Idem if not absorbed before
+            if (previouslyAbsorbedCompounds.ContainsKey(compoundWeight.Key))
```
hhyyrylainen, line 400: The same thing here.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -370,43 +384,86 @@ private void RunAndTumble(Random random)
...
+                compoundDifference -= previouslyAbsorbedCompounds[compoundWeight.Key];
+
+            compoundDifference *= compoundWeight.Value;
+            gradientValue += compoundDifference;
         }
 
+        // Implement a detection threshold to possibly rule out too tiny variations
+        float differenceDetectionThreshold = 0;
```
hhyyrylainen, line 408: Why is this a variable if this is never changed? Shouldn't this be a constant defined in Constants?

src/microbe_stage/MicrobeAI.cs
```diff
@@ -370,43 +384,86 @@ private void RunAndTumble(Random random)
...
+    /// <summary>
+    ///   Prioritizing compounds that are stored in lesser quantities.
+    ///   If ATP-producing compounds are low (less than half storage capacities),
+    ///   non ATP-related compounds are discarded.
+    /// </summary>
+    /// <returns> A dictionary of weights for each (useful) compound, to be used in search algorithms.</returns>
+    private Dictionary<Compound, float> ComputeCompoundsSearchWeights(
+        List<KeyValuePair<Compound, float>> usefulCompounds)
+    {
+        // Vital compounds are *direct* ATP producers
```
hhyyrylainen, line 443: I think a TODO comment should be added here to explain that this is a *slight* shortcut, as in theory we don't know here if iron is truly useful  for ATP production or if some future process uses it as input but doesn't make ATP.
The logic is fine enough now but I want that clarifying comment.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -370,43 +384,86 @@ private void RunAndTumble(Random random)
...
+        var usefulVitalCompounds = usefulCompounds.Where(x => x.Key == glucose || x.Key == iron);
+
+        // If this microbe lacks vital compounds don't bother with ammonia and phosphorous
```
hhyyrylainen, line 446: Should this say "phosphate"?

src/microbe_stage/MicrobeAI.cs
```diff
@@ -370,43 +384,86 @@ private void RunAndTumble(Random random)
...
+        // This algorithm doesn't try to determine if iron and hydrogen sulfide is useful to this microbe
+        if (usefulVitalCompounds.Any(
+            compound => microbe.Compounds.GetCompoundAmount(compound.Key) < 0.5f * microbe.Compounds.Capacity))
+        {
+            usefulCompounds = usefulCompounds.Where(x => x.Key != ammonia && x.Key != phosphates).ToList();
+        }
+
+        var compoundsPriority = new Dictionary<Compound, float>();
```
hhyyrylainen, line 454: I think this dictionary should be stored in the microbe AI class (doesn't need saving) so that new dictionaries don't need to be allocated each update.

hhyyrylainen (changes requested):

src/microbe_stage/MicrobeAI.cs
Maxonovien: At first thought I intended to size it with cell capacity (because 'small' variations depend on your size, after all. But that could indeed be at least related to a constant.

src/microbe_stage/MicrobeAI.cs
Maxonovien: Indeed. I also figured out that the code was a bit weak if further compounds were added (e.g. a third component for reproduction would still be searched, as such). I thought of having some kind of categories (Vital, ATP-related, reproduction...), which would allow to check if a Compound var is part of it or not. That'd rather be refactoring work though I think, so I kept it a bit aside for the moment. 

These do not cover multiple usages anyways.

src/microbe_stage/MicrobeAI.cs
Maxonovien: Yes, I guess. This was (along with the sulfuric acid comment) an old comment that I did not update. Perhaps from a time when names were different (or some tired commenter haha). I'll fix this anyways

src/microbe_stage/MicrobeAI.cs
```diff
@@ -38,13 +39,26 @@ public class MicrobeAI
...
+    private Dictionary<Compound, float> compoundsSearchWeights;
+
     public MicrobeAI(Microbe microbe)
     {
         this.microbe = microbe ?? throw new ArgumentException("no microbe given", nameof(microbe));
         glucose = SimulationParameters.Instance.GetCompound("glucose");
         oxytoxy = SimulationParameters.Instance.GetCompound("oxytoxy");
         ammonia = SimulationParameters.Instance.GetCompound("ammonia");
         phosphates = SimulationParameters.Instance.GetCompound("phosphates");
+        iron = SimulationParameters.Instance.GetCompound("iron");
```
FuJa0815, line 59: I'd move this line under the glucose definition to match the variable declaration above.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -38,13 +39,26 @@ public class MicrobeAI
     [JsonProperty]
     private float pursuitThreshold;
 
+    /// <remarks>
+    ///   Memory of the previous absorption step is required to compute gradient.
+    ///   Values dictionary rather than single value as they will be combined with variable weights.
+    /// </remarks>
+    [JsonProperty]
+    private Dictionary<Compound, float> previouslyAbsorbedCompounds;
```
FuJa0815, line 47: Am I missing something obvious?

This variable gets declared here.
It gets initialized on line 60 with the values of `TotalAbsorbedCompounds`.
Lines 82 to 88 also apply the values of `TotalAbsorbedCompounds` to this variable.
Line 404 reads the value.

Why don't you just access `TotalAbsorbedCompounds` directly on line 404?

FuJa0815 (changes requested):

src/microbe_stage/MicrobeAI.cs
Maxonovien: I guess this is a tricky part that definitely seems to need some more comments (which I've just written): lines 82->88 store the value from `TotalAbsorbedCompounds`, and line 89 clears `TotalAbsorbedCompounds` just afterwards (to free it the new absorption values for next time step). So previouslyAbsorbedCompounds serves as a storage (which is why it is always getting the value of `TotalAbsorbedCompounds`, but from the previous step).

src/microbe_stage/MicrobeAI.cs
```diff
@@ -65,7 +85,14 @@ public void Think(float delta, Random random, MicrobeAICommonData data)
 
         ChooseActions(random, data);
 
-        // Clear the absorbed compounds for run and rumble
+        // Store the absorbed compounds for run and rumble
+        previouslyAbsorbedCompounds.Clear();
+        foreach (var compound in microbe.TotalAbsorbedCompounds)
+        {
+            previouslyAbsorbedCompounds[compound.Key] = compound.Value;
+        }
```
FuJa0815, line 93:
```suggestion
        previouslyAbsorbedCompounds = new Dictionary<Compound, float>(microbe.TotalAbsorbedCompounds);
```
This also makes a copy of the dictionary

src/microbe_stage/MicrobeAI.cs
```diff
@@ -370,43 +397,85 @@ private void RunAndTumble(Random random)
         microbe.State = Microbe.MicrobeState.Normal;
 
         var usefulCompounds = microbe.TotalAbsorbedCompounds.Where(x => microbe.Compounds.IsUseful(x.Key));
+        ComputeCompoundsSearchWeights(usefulCompounds.ToList());
```
FuJa0815, line 400: It would be much better to make this `.ToList()` one line above, or, even better in my opinion, use `.ToDictionary(i => i.Key, i => i.Value)` there.

FuJa0815 (changes requested):

src/microbe_stage/MicrobeAI.cs
Maxonovien: This was specifically not done to not create a new dictionnary instance.

src/microbe_stage/MicrobeAI.cs
Maxonovien: I get for the call to `.ToList()` above (I only use it right below anyways), but I don't get the benefit of having a dictionary here, as I never use `usefulCompounds` as such.

If this is for futur usages which may use a dictionary, why not keeping it as a generic IEnumerable, as it could be used as a list as well?

FuJa0815: I saw that this is a collection of a key value pair, which looks odd to me becasue there is a special datatype in c# for that and `TotalAbsorbedCompounds` is a Dictionary anyways.
`Dictionary<Compound, float>` looks much cleaner to me than `IEnumerable<KeyValuePair<Compound,float>>`.

But looking though the code, it seems that an arraylist should be more more efficient than a hashmap in this case, so you can keep it as a list.

Maxonovien: So the `IEnumerable<KeyValuePair<Compound,float>>` was the original type of `usefulCompounds` and I did not change it. But given that it's only used as a list at some point I just went ahead and directly converted to List in the var instanciation.

hhyyrylainen: Useful compounds purposefully uses a general interface, as it doesn't have a real dictionary to return (at least I think so) because it gets the ammonia etc. added there, so that's why the return type is as is.
ToList should *only* be used if the list is iterated multiple times, otherwise it is a waste.

Maxonovien: Okay so I had a doubt and tried again, but in fact I tried keeping IEnumerable, but jetbrains did not let me do so. The up to date version (commit 540b708) is using IEnumerables and I got this from auto check:
src\microbe_stage\MicrobeAI.cs:458 Possible multiple enumeration type: PossibleMultipleEnumeration

(for all lines where usefulCompounds is used)

hhyyrylainen: If there *are* multiple enumerations, *then* use ToList, if it is enumerated just once (and for example the method call makes the checker confused) then add a suppression.

Maxonovien: Yeah, I think the function call is what makes the checker confused. I went ahead and used another solution: I actually moved the computation of useful compounds to the search weights computation itself as only it was using this compounds list.

Maxonovien:
Hm, that may very well be the case, I think this was old code, so it was perhaps not removed at the time. Two possibilities:
1. It's a defensive measure, which is not very costly (the list isn't long) so we keep it
2. We get rid of it because compounds are already filtered (and just check -but I can't right now - that it works fine, to be sure).

Maxonovien:
Hm, that may very well be the case, I think this was old code, so it was perhaps not removed at the time. Two possibilities:
1. It's a defensive measure, which is not very costly (the list isn't long) so we keep it
2. We get rid of it because compounds are already filtered (and just check -but I can't right now - that it works fine, to be sure).

hhyyrylainen:
If it was that way beforehand, it was probably protection code written when the compound absorbing *could* absorb non-useful compounds, but now it can't. There's literally a call to IsUseful for each compound type before it is absorbed by a cell. So a cell cannot absorb something it doesn't consider useful.

https://github.com/Revolutionary-Games/Thrive/blob/79f34d3e678c4d47bc664489ce1c2debf56d5263/src/microbe_stage/CompoundCloudPlane.cs#L486

Maxonovien:
Wouldn't this be a problem with the `.Any` though? I think it iterates once but does not return a new IEnumerable to `.Where`...

hhyyrylainen:
I guess, maybe. I'm still looking over other PRs so I can't dedicate too much brainpower for this...

src/microbe_stage/MicrobeAI.cs
```diff
@@ -369,44 +396,94 @@ private void RunAndTumble(Random random)
...
+    private void ComputeCompoundsSearchWeights()
+    {
+        IEnumerable<Compound> usefulCompounds = microbe.TotalAbsorbedCompounds.Keys;
+
+        // If this microbe lacks vital compounds don't bother with ammonia and phosphate
+        if (microbe.TotalAbsorbedCompounds.Keys.Any(
+            compound => IsVitalCompound(compound) &&
+                microbe.Compounds.GetCompoundAmount(compound) < 0.5f * microbe.Compounds.Capacity))
+        {
+            usefulCompounds = usefulCompounds.Where(x => x != ammonia && x != phosphates);
```
FuJa0815, line 461: I'd use `usefulCompounds = usefulCompounds.Except(new[] { ammonia, phosphates });`

src/microbe_stage/MicrobeAI.cs
```diff
@@ -369,44 +396,94 @@ private void RunAndTumble(Random random)
...
+    /// <summary>
+    ///   Tells if a compound is vital to this microbe.
+    ///   Vital compounds are* direct* ATP producers
+    ///   TODO: what is used here is a shortcut linked to the current game state:
+    ///     such compounds could be used for other processes in future versions
+    ///   TODO: currently, vital compounds are not necessarily useful (e.g. no use of iron)!
+    /// </summary>
+    private bool IsVitalCompound(Compound compound)
+    {
+        return compound == glucose || compound == iron;
```
FuJa0815, line 484: why didn't you write ` || (compound == iron && microbe.Compounds.IsUseful(iron))` here?

FuJa0815 (changes requested):

src/microbe_stage/MicrobeAI.cs
FuJa0815: just noticed that that isn't really your code but adopted :D
But still, I'd prefer an `Except`-usage here

Maxonovien: Yes it was adopted, but I'm definitely in favor of using `Except` as well, now that I know of it. :)

hhyyrylainen: Well the Except syntax shown above, probably allocates a new two element array. Just to remind everyone that we are working on a realtime game and not https://github.com/Revolutionary-Games/ThriveDevCenter where I would allow this kind of code in 0 milliseconds.

src/microbe_stage/MicrobeAI.cs
Maxonovien: Hm, the thing is, at this point I don't need to know whether it is useful to the microbe because it was filtered out already if not. So I just kept vital independant from useful, although in practice, I agree there is little reason for a compound to be vital but not useful. 

I'm a bit torn about changing it. As of now, this feels like a static method that wasn't defined as such, so I think I'll add some usefulness check indeed.

Note that this isn't limited to iron (glucose can be useless as well), by the way.

buckly90:
Was testing this as small chemo bacteria. I found a space between a glucose cloud and a hydrogen cloud, with some ammonia and phosphates in the bottom right corner. From there I turned on godmode and duplicated myself to create starving clones.

These clones seemed to mostly go to the right, and... never made it to the sulfides. They would seriously go right to the edge of the cloud and then just, turn around. Most agonizing experience ever, watching that. They seemed to ignore the glucose cloud.

Looking closer I think it might be something to do with their run and tumble behavior kicking in too early or something, because they begin to "dance" on the edge of the clouds before continuing on in another direction. I'm not really sure.

Maxonovien:
The reason why they leave is because they have filled their glucose storage already. So they probably started venturing to find phosphate and ammonia to reproduce.

buckly90:
If that was the case, they wouldn't be starving to death. They didn't have anything at all, so I would assume they would want to prioritize the glucose or sulfides.

Maxonovien:
Well, single-hex molecules fill out their storage pretty quick. Do they starve to death while leaving the cloud, or only before/after?

buckly90:
They starve to death after. But that's only because they never make it to the clouds, if they happen to go to the cloud they just turn around once they reach it, without actually seemingly making contact with it.

Maxonovien:
If they turn this is because they spot some compounds difference, normally. My best bet is that there is some residual glucose affecting their behavior. There is a constant controlling how sensitive they are (`AI_GRADIENT_DETECTION_THRESHOLD`), which may solve the issue.

Maxonovien:
So, I just added some margin to the gradient detection with the aforementionned constant, and it seems pretty fine on your save. 

Another factor however is that your cells are very fast, so it may affect their ability to detect thin clouds. I could make them slow down when reaching a cloud, so that they are more able to explore it, but I am not sure about this choice, so I leave it up for discussion (and probably another PR). Let me know if you encounter any further issue!

src/microbe_stage/MicrobeAI.cs
```diff
@@ -384,44 +411,94 @@ private void RunAndTumble(Random random)
...
-        }
+            // Note this is about absorbed quantities (which is all microbe has access to) not the ones in the clouds.
+            // Gradient computation is therefore cell-centered, and might be different for different cells.
+            float compoundDifference = 0.0f;
 
-        float compoundDifference = 0.0f;
-        foreach (var compound in usefulCompounds)
-        {
-            compoundDifference += compound.Value;
+            float quantityAbsorbedThisStep, quantityAbsorbedPreviousStep;
```
hhyyrylainen, line 423: These variables can be inlined in the out declarations on the next two lines

src/microbe_stage/MicrobeAI.cs
```diff
@@ -384,44 +411,94 @@ private void RunAndTumble(Random random)
...
+                MoveWithRandomTurn(0.0f, 1.5f, random);
             }
         }
     }
 
+    /// <summary>
+    ///   Prioritizing compounds that are stored in lesser quantities.
+    ///   If ATP-producing compounds are low (less than half storage capacities),
+    ///   non ATP-related compounds are discarded.
+    ///   Updates compoundsSearchWeights class dictionary.
```
hhyyrylainen, line 465: It's not static, so it is not a class variable. Instead should be mentioned "instance field"

src/microbe_stage/MicrobeAI.cs
```diff
@@ -384,44 +411,94 @@ private void RunAndTumble(Random random)
...
+            // Should be tweaked with consumption
+            var compoundPriority = 1 - microbe.Compounds.GetCompoundAmount(compound) / microbe.Compounds.Capacity;
+
+            compoundsSearchWeights.Add(compound, compoundPriority);
+        }
+    }
+
+    /// <summary>
+    ///   Tells if a compound is vital to this microbe.
+    ///   Vital compounds are* direct* ATP producers
```
hhyyrylainen, line 492: ` *direct* ` ?

hhyyrylainen (changes requested):
A few small style comments, other than that I'm happy with this. Seems to work also fine with older upgraded saves.

hhyyrylainen (approved):
My comments were addressed, so this can be merged once my PR is approved and merged (to fix the CI error): https://github.com/Revolutionary-Games/Thrive/pull/2482

PR #2138, opened 2021-03-21
Cloud spawns happen on regular intervals, and conform to the compound densities of a patch. (closes 1560)

Gwen-Benzschawel:
This replaces the old system for one that's like a bag. Compounds are added to the bag and shuffled. When the spawn er needs a cloud, it draws one from the bag. When the bag is empty, the bag is refilled and shuffled. This ensures that you can get a steady rate of each compound, but the order of spawn is still random. 

The compound bag is set up so that it will empty the bag at a predefined constant rate that can be changed for game-play balance.

closes #1560

PR #2389, opened 2021-07-06
1541 - Fixing lack of iron chunks in cave

### LINKED ISSUE #1541: Investigate why there is no iron spawning in the cave
It should be one of the biomes with the most iron.

Edit: it seems that the new patch spawn conditions aren't properly being set when moving to another patch.

### LINKED ISSUE #1626: Investigate why chunk spawn rate is way down
I just tested the game for like 6 minutes, with 2 of those being in the ice shelf (which has a high iron concentration), without seeing any iron chunks. This is highly not good, though this could have been just very bad RNG.
It should be investigated what's the problem.

Potentially related to: https://github.com/Revolutionary-Games/Thrive/issues/1541

### LINKED ISSUE #1777: Check if hydrogen sulfide spawns in the cave
Got one report that it doesn't.
Related to: https://github.com/Revolutionary-Games/Thrive/issues/1541

Maxonovien:

This PR adress the issue of lacking iron in the cave, by two fixes:
1. Increasing small chunks density in the cave, to avoid high proportion displayed on patch map hiding sparse, very rich deposits (only affects new games as these densities are saved)
2. Reworking spawning algorithm by running an attempt per spawner, up to max, rather than up to max attempts per spawner, to favor diversity (affects all patches).

**Related Issues**

Resolves #1541 and resolves #1626 and resolves #1777 .

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/microbe_stage/SpawnSystem.cs
+++ b/src/microbe_stage/SpawnSystem.cs
@@ -68,6 +68,8 @@ public class SpawnSystem
     /// </summary>
     private int estimateEntityCount;
 
+    private Dictionary<Spawner, int> attemptsPerSpawnType;
+
     public SpawnSystem(Node root)
     {
         worldRoot = root;
@@ -232,8 +234,7 @@ public class SpawnSystem
 
         int spawned = 0;
 
-        // TODO ugly
-        Dictionary<Spawner, int> attemptsPerSpawnType = new Dictionary<Spawner, int>();
+        attemptsPerSpawnType.Clear();
         int maxAttempts = -1;
         foreach (var spawnType in spawnTypes)
         {
@@ -265,6 +266,8 @@ public class SpawnSystem
             foreach (var spawnType in spawnTypes)
             {
                 int numAttempts = attemptsPerSpawnType[spawnType];
+
+                // Already did max attempts for this type
                 if (i > numAttempts)
                     continue;
 
--- a/src/microbe_stage/SpawnSystem.cs
+++ b/src/microbe_stage/SpawnSystem.cs
@@ -232,7 +232,19 @@ public class SpawnSystem
 
         int spawned = 0;
 
+        // TODO ugly
+        Dictionary<Spawner, int> attemptsPerSpawnType = new Dictionary<Spawner, int>();
+        int maxAttempts = -1;
         foreach (var spawnType in spawnTypes)
+        {
+            int numAttempts = Math.Min(Math.Max(spawnType.SpawnFrequency * 2, 1),
+                maxTriesPerSpawner);
+            attemptsPerSpawnType[spawnType] = numAttempts;
+            if (numAttempts > maxAttempts)
+                maxAttempts = numAttempts;
+        }
+
+        for (int i = 0; i < maxAttempts; i++)
         {
             /*
             To actually spawn a given entity for a given attempt, two
@@ -249,11 +261,13 @@ public class SpawnSystem
             numAttempts stores how many times the SpawnSystem attempts
             to spawn the given entity.
             */
-            int numAttempts = Math.Min(Math.Max(spawnType.SpawnFrequency * 2, 1),
-                maxTriesPerSpawner);
 
-            for (int i = 0; i < numAttempts; i++)
+            foreach (var spawnType in spawnTypes)
             {
+                int numAttempts = attemptsPerSpawnType[spawnType];
+                if (i > numAttempts)
+                    continue;
+
                 if (random.Next(0, numAttempts + 1) < spawnType.SpawnFrequency)
                 {
                     /*
--- a/src/microbe_stage/SpawnSystem.cs
+++ b/src/microbe_stage/SpawnSystem.cs
@@ -239,12 +239,12 @@ public class SpawnSystem
         int maxAttempts = -1;
         foreach (var spawnType in spawnTypes)
         {
-            int numAttempts = Math.Min(Math.Max(spawnType.SpawnFrequency * 2, 1),
+            int attempts = Math.Min(Math.Max(spawnType.SpawnFrequency * 2, 1),
                 maxTriesPerSpawner);
-            attemptsPerSpawnType[spawnType] = numAttempts;
+            attemptsPerSpawnType[spawnType] = attempts;
 
-            if (numAttempts > maxAttempts)
-                maxAttempts = numAttempts;
+            if (attempts > maxAttempts)
+                maxAttempts = attempts;
         }
 
         for (int i = 0; i < maxAttempts; i++)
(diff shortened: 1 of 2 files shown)


src/microbe_stage/SpawnSystem.cs
```diff
@@ -232,7 +235,18 @@ private void SpawnEntities(Vector3 playerPosition, Vector3 playerRotation, int e
...
         int spawned = 0;
 
+        attemptsPerSpawnType.Clear();
+        int maxAttempts = -1;
         foreach (var spawnType in spawnTypes)
+        {
+            int numAttempts = Math.Min(Math.Max(spawnType.SpawnFrequency * 2, 1),
+                maxTriesPerSpawner);
+            attemptsPerSpawnType[spawnType] = numAttempts;
+            if (numAttempts > maxAttempts)
```
adQuid, line 245:
```suggestion
            
            if (numAttempts > maxAttempts)
```

src/microbe_stage/SpawnSystem.cs
```diff
@@ -247,13 +261,19 @@ private void SpawnEntities(Vector3 playerPosition, Vector3 playerRotation, int e
             spawn cycle, the SpawnSystem attempts to spawn each given
             entity multiple times depending on the spawnFrequency.
             numAttempts stores how many times the SpawnSystem attempts
-            to spawn the given entity.
+            to spawn the given entity. Furthermore, each attempt gives a spawn chance
+            for each spawner (up to their type limit) to still favor diversity,
+            instead of one spawner taking up all the free spawning slots.
```
adQuid, line 266: Was the issue that the spawners were making so much of one thing that they were reaching the maximum of living entities, which kept iron from spawning in the cave? From testing this change clearly works, but I don't understand why.

hhyyrylainen: That could certainly be a thing. For a while now, I've wanted per entity spawn caps (that are more reasonable, for example I think if the 1000 entity limit fills up with cells the game will lag, a lot).

Maxonovien: That was not even the total cap, but the per-frame cap (toxyns were taking all slots, that's why decreasing their density led to iron spawning)

adQuid: Per-frame? As in the spawner has been adding 500 toxyns per frame? I certainly haven't observed that in game...

Maxonovien: No, the limit was much smaller (2, I think), so at each frame, 2 entities would be spawned (which makes sense, both to not have long spawning frame and to be able to spawn new things to adapt - although it could be raised shortly after new patch spawn). 

Yet, given that toxins had such a high spawning rate, they always spawned up to the frame spawn limit (they roughly had ~100 chances out of 500 (max cap) so 20% chances, for 50 spawn trials or so). And thus, spawners later in the list were not tested in this frame, and the list would be reset at the next frame.

I guess that another fix could have been to save the remaining spawner list for next frame to cycle it all.

hhyyrylainen: > For a while now, I've wanted per entity spawn caps

Capping toxins to 100, and having a separate cap for iron and marine snow chunks, is what I meant in my comment. This doesn't have anything to do with per-frame caps, which also probably could be problematic if the toxins take up all the per-frame spawns... this is addressed by: https://github.com/Revolutionary-Games/Thrive/pull/2138 (which is an incomplete PR to change the spawn system) which uses a round robin approach to spawn from a randomized item list.

Maxonovien: Indeed a per-entity cap is a fully independant question, related to the total cap which is not the reason for the issues fixed here (but I agree this could prove useful overall).

This PR was really about identifying their cause and fixing them within the current system as they really harm game experience in patches.

Maxonovien: I'm also going to take a look at the incomplete spawn change PR though, see if I feel like taking it up.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -232,7 +235,19 @@ private void SpawnEntities(Vector3 playerPosition, Vector3 playerRotation, int e
 
         int spawned = 0;
 
+        attemptsPerSpawnType.Clear();
+        int maxAttempts = -1;
         foreach (var spawnType in spawnTypes)
+        {
+            int numAttempts = Math.Min(Math.Max(spawnType.SpawnFrequency * 2, 1),
```
hhyyrylainen, line 242: I know you used existing code as a basis, but as this is being rewritten anyway, I'd like to see this variable to be renamed to just `attempts`

Maxonovien: I also tend to use `num` prefix for number variables, as it makes clear, in my opinion, that we are counting things. I can change it if this fits the project coding style better, though.

hhyyrylainen: Yes please. Abbreviations (except a few) are banned in the styleguide. Also this is the only source file that has "num" prefix in a name.

Maxonovien: Indeed, I use it so often I did not even realize it was an abbreviation. My bad, it's fixed!

FuJa0815 (changes requested):

src/microbe_stage/SpawnSystem.cs
```diff
@@ -232,7 +235,19 @@ private void SpawnEntities(Vector3 playerPosition, Vector3 playerRotation, int e
 
         int spawned = 0;
 
+        attemptsPerSpawnType.Clear();
+        int maxAttempts = -1;
         foreach (var spawnType in spawnTypes)
+        {
+            int attempts = Math.Min(Math.Max(spawnType.SpawnFrequency * 2, 1),
```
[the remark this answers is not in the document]
FuJa0815: As I was clicking `Submit Review`, I've noticed that `Math.Clamp` is a .net core method, which is not avaiable here.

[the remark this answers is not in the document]
hhyyrylainen: Godot has `Mathf.Clamp`...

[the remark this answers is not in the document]
FuJa0815: Oh, I thought Mathf was godot stuff for floating point stuff... I've just checked and it is also for integer calculations. I didn't know this. In that case, this still applies :D

src/microbe_stage/SpawnSystem.cs
```diff
@@ -68,9 +68,12 @@ public class SpawnSystem
     /// </summary>
     private int estimateEntityCount;
 
+    private Dictionary<Spawner, int> attemptsPerSpawnType;
```
FuJa0815, line 71: wouldn't it be better to have the attempts counter in the `Spawner` class itself? Or is there somthing prohibiting this I don't see?

Maxonovien: No, you are right, the spawner-dependent part could be stored in `Spawner` indeed for memory saving.

However, given that I think I'm finally succeeding with a more efficient rework with PR #2397 (which also solves this spawn order thing), I think I'll discontinue this one for a bit, and only resume if the other PR actually does not work / takes longer and a hotfix is needed.

hhyyrylainen:
So this PR will be superceded by https://github.com/Revolutionary-Games/Thrive/pull/2397 if that gets done first.

PR #2390, opened 2021-07-06
2370 - Editor access in colony

### LINKED ISSUE #2370: Editor button needs to lock if the player joins a colony
Got a report that if you are ready to go to the editor and you join a colony, the editor button stays enabled (and pressing it crashes the game).

Maxonovien:

This PR does disable editor upon creating colony and reenables it upon unbinding (only if it was possible already).

**Related Issues**

Fixes #2370 .

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/microbe_stage/Microbe.cs
```diff
@@ -1382,6 +1382,12 @@ internal void OnColonyMemberRemoved(Microbe microbe)
             RevertNodeParent();
             ai?.ResetAI();
 
+            // Colony reset at the end, to allow for previous operations on it.
+            Colony = null;
+
+            // Checks if the cell is ready to reproduce (allows editor access on unbind).
+            if (allOrganellesDivided)
+                OnReproductionStatus?.Invoke(this, true);
```
Kasterisk, line 1390: A little nitpick but it's nice to have a whitespace separating the `return` and this.

Kasterisk (approved):
Seems to work great overall.

src/microbe_stage/Microbe.cs
```diff
@@ -1382,6 +1382,13 @@ internal void OnColonyMemberRemoved(Microbe microbe)
             RevertNodeParent();
             ai?.ResetAI();
 
+            // Colony reset at the end, to allow for previous operations on it.
+            Colony = null;
```
hhyyrylainen, line 1386: Why was this added here, as there's already:
https://github.com/Revolutionary-Games/Thrive/blob/cf56b89a03d8dbf5067a260d14e319170fc6d1b4/src/microbe_stage/MicrobeColony.cs#L61
?
I think a better design would be to reset allOrganellesDivided to false when joining a colony (and then make sure that after unbind the reproduction progress checks will properly enable the editor if needed).

hhyyrylainen (changes requested):

src/microbe_stage/Microbe.cs
Maxonovien: Hm, that's indeed a good point. But I added this because I realised that, at this point, the colony was not set to null (commenting my line restores the issue).

After investigating, this line in facts sets the microbe Colony to null, but after Microbe-related methods (those which allow to reproduce). I tried moving it, but it disables full unbinding of colony, because, well, the colony to unbind is null. It would make sense, however to delete the `microbe.Colony = null;` line in MicrobeColony.

hhyyrylainen: Did you try using the allOrganellesDivided variable approach? I think it might remove the need to fiddle with where the Colony variable is set to null.

Maxonovien: It would only work for disabling editor, not reenabling it : to activate the editor button from the HUD, I have to use MicrobeHUD.ShowReproductionDialog(); through MicrobeStage.OnPlayerReproductionStatusChanged, which checks player.Colony == null. 

So I need to nullify the Colony beforehand, or change/circumvene other files safety check (which is not advisable in general).

hhyyrylainen: Then setting allOrganellesDivided to false here, might work? I think the logic would go like this:
- microbe runs update
- microbe leaves a colony
- microbe runs update (reproduction progress gets re-checked here, and Colony is now null)

Maxonovien: I wanted to rely on such logic at first but did not consider allOrganellesDivided shortcuts. It may work, in fact. I'll take a look at it after reworking AI code

Maxonovien: You were fully right on this it seems to work perfectly well, including when saving & loading. It seems much nicer as well. Thanks!

src/microbe_stage/Microbe.cs
```diff
@@ -1420,25 +1420,29 @@ internal void SuccessfulKill()
...
     {
-        var oldRotation = Rotation;
-        var vectorToParent = GlobalTransform.origin - ColonyParent.GlobalTransform.origin;
-        ChangeNodeParent(ColonyParent);
-
         State = MicrobeState.Normal;
 
-        var vectorToParentRotated = vectorToParent.Rotated(Vector3.Down, Rotation.y);
-        var vectorToMembrane = Membrane.GetExternalOrganelle(vectorToParentRotated.x, vectorToParentRotated.y);
+        if (ColonyParent != null)
```
FuJa0815, line 1425: I think an early return would be better. Something like
```cs
if (ColonyParent == null)
{
    UnreadyToReproduce();
    return;
}
// code...
```
would be more readable

FuJa0815 (changes requested):

FuJa0815 (approved):

hhyyrylainen:
I opened a separate issue about the unbinding mode being exited: https://github.com/Revolutionary-Games/Thrive/issues/2409

hhyyrylainen (approved):
If @FuJa0815 tested this, that's good enough for me.

PR #2064, opened 2021-01-26
Make Converter classes also catch ArgumentNullException

Apsum:
closes #1863

PR #2394, opened 2021-07-10
1863 catch argumentnullexception

### LINKED ISSUE #1863: Make Converter classes also catch ArgumentNullException
as I noticed that the null pointer catch doesn't actually trigger, instead the non existing key read from JObject seems to result in argument null exception instead.

FuJa0815:

`GodotBasisConverter.ReadJson` now also catches `ArgumentNullException`s

**Related Issues**

closes #1863
closes #2064 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
I think there was another converter where this change should be made...

src/saving/serializers/RandomConverter.cs
```diff
@@ -73,11 +73,9 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
 
             return random;
         }
-        catch (ArgumentException e)
-        {
-            throw new JsonException("Can't read Random (missing property)", e);
-        }
-        catch (NullReferenceException e)
+        catch (Exception e) when (
+            e is ArgumentException ||
```
FuJa0815, line 77: I didn't add `ArgumentNullException` here, because rider complained that `this is always false`

hhyyrylainen (approved):

Issue #2095, opened 2021-02-13
Rework editor speed display to use actual physics simulation


hhyyrylainen:
Godot supports multiple parallel scene trees that have physics and separate rendering, we could use this in the editor, to run a simulation sandbox with the player cell in it, to measure how fast it is able to go. This should permanently fix all the problems related to the manually coded algorithm for speed not matching reality, but this is pretty hard to implement, and will introduce a slight delay before new speed value can be shown.

hhyyrylainen:
Or another approach that results in the right value. At least the external organelle positioning needs to be taken into account (which currently isn't, causing at least some inaccuracy, as well as the cell origin to hex mass center moving is not taken into account).

hhyyrylainen:
It was noticed that flagellum upgrades are especially badly shown by the speed display currently.

hhyyrylainen:
I'll note that the multicellular speed calculation also depends on the cell speed calculation and in addition does some math to arrive at the overall colony calculation, which I think is also an estimate (which is the root cause of this whole issue; estimating the speed based on some math formulas).

Accidental-Explorer:
I think the code contains comments saying that an estimate is deliberately used for performance reasons, is that not an issue?

Or does that only apply to the auto-evo side?

hhyyrylainen:
Using the formula approach will *definitely* need to be used for auto-evo, but we should have a more accurate display to use for the player species in the editor as players keep reporting modifying flagellum speed as a bug. 

So we just have to accept the performance impact of providing more accurate numbers for the player. Theoretically we could simulate hundreds of different cell layouts to determine their speeds and use that data to give better estimates, but that would be even more work than the base idea of using an actual simulation to determine the movement speed.

PR #2395, opened 2021-07-10
Food Source Specialization in Auto-Evo

adQuid:

  Incrementally improves auto-evo by adding classes for various food sources, to allow for more specialization of microbes and extension of code in the future. When testing you can always add this code to the OnReturnFromEditor function from MicrobeStage.cs:

```
var i = 0;
		foreach (MicrobeSpecies toSpawn in CurrentGame.GameWorld.Map.CurrentPatch.SpeciesInPatch.Keys)
		{
			if (toSpawn != Player.Species)
			{
				SpawnHelpers.SpawnMicrobe(toSpawn, Player.Translation + new Vector3(4.0f * ++i, 0.0f, 0.5f * (float)Math.Pow(CurrentGame.GameWorld.Map.CurrentPatch.GetSpeciesPopulation(toSpawn), 0.35)),
				rootOfDynamicallySpawned, SpawnHelpers.LoadMicrobeScene(), true, Clouds,
				CurrentGame);
			}
		}
```

  This isn't great at predator-prey relationships, since right now larger cells can't realistically catch smaller ones to eat them, but it at least makes species that COULD eat others, and kind of threaten the player. It's also not great at making the jump to Flagella, but it can accomplish it. It makes progress though, and isn't absurdly huge.

Some scientific assumptions:
* There exist five ways to generate energy:
    * Sunlight
    * Hygrogen Sulfide
    * Iron
    * Free Glucose 
    * Eating another species of microbe
 * There is a cap of energy available for each food source, determined by the concentrations in the patch and the population of species present.
 * The population of a species is limited to how well that species can gather energy from the sources, as compared to other species. This is calculated by sum of squares to softly enforce the exclusion principle.
 * Predator/prey relationships only matter for the predator. The ability of prey to avoid consumption is not modeled in this iteration of the algorithm.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


adQuid:
@Maxonovien This is the auto-evo branch we were talking about. Feel free to contribute.

src/auto-evo/simulation/foodsource/HeterotrophicFoodSource.cs
```diff
@@ -0,0 +1,78 @@
...
+        predatorSpeed += ProcessSystem
+            .ComputeEnergyBalance(microbeSpecies.Organelles.Organelles.Select(x => x.Definition), patch.Biome,
+                microbeSpecies.MembraneType).FinalBalance;
+
+        // It's great if you can engulf this prey, but only if you can catch it
+        var engulfScore = predatorSize / preySize > Constants.ENGULF_SIZE_RATIO_REQ
+            && !microbeSpecies.MembraneType.CellWall ?
+                Constants.AUTO_EVO_ENGULF_PREDATION_SCORE :
+                0.0f;
+        engulfScore *= predatorSpeed > preySpeed ? 1.0f : 0.1f;
```
Maxonovien, line 46: Why is the value for a faster prey 0.1 and not 0? Is it to simulate that, by chance predators could catch prey?

If yes, 
1. Is it really a good approximation? It's not like a cell would trip. Of course, they could be impeded by chunks or cells (but the predator would as well, and given they are small they would not be really a problem). The most likely case are other predators, but that would be very populations depending then.
2. Shouldn't this apply to the predator as well, as they may miss their prey? (doing like 0.9f : 0.1f).
3. Ultimately, unless there is a formula coming from this, this could use to be defined in Constants.

Maxonovien (commented):
While the proposed system is simple and interesting, there are some flaws that should likely be fixed, unmotivated large size encouragement and no actual incentive to use storage/ammonia synthesis.

Related to this, and perhaps more crucially, this system only focuses on food source, i.e. survivalist fit. **But evolution is first and foremost about reproduction**: if you are short-lived, yet reproduces a lot, you might be fitter than longer-lived species. So gathering of phosphate & ammonia should be included. To be fair, the proposed system and its implementation could be rather easily extended onto these lines.

But I really feel like some parts are missing a bit too much right now.

src/auto-evo/simulation/foodsource/HeterotrophicFoodSource.cs
```diff
@@ -0,0 +1,78 @@
...
+                pilusScore += Constants.AUTO_EVO_PILUS_PREDATION_SCORE;
+                continue;
+            }
+
+            foreach (var process in organelle.Definition.RunnableProcesses)
+            {
+                if (process.Process.Outputs.ContainsKey(oxytoxy))
+                {
+                    oxytoxyScore += Constants.AUTO_EVO_TOXIN_PREDATION_SCORE;
+                }
```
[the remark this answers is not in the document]
hhyyrylainen: It would be interesting to take the output compound amounts of the processes these organelles run into account when determining their score, indeed.

simulation_parameters/Constants.cs
```diff
@@ -337,8 +337,9 @@ public static class Constants
 
     // Mutation Variables
     public const float MUTATION_BACTERIA_TO_EUKARYOTE = 0.01f;
-    public const float MUTATION_CREATION_RATE = 0.15f;
-    public const float MUTATION_DELETION_RATE = 0.03f;
+    public const float MUTATION_CREATION_RATE = 0.25f;
+    public const float MUTATION_NEW_ORGANELLE_CHANCE = 0.25f;
+    public const float MUTATION_DELETION_RATE = 0.05f;
```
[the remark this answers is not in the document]
adQuid: This actually reaches an equilibrium due to a quirk in our mutations code: https://github.com/Revolutionary-Games/Thrive/blob/master/src/general/Mutations.cs#L201

I'm actually unsure if there's any scientific bias in the random mutations process, but due to the initial population of single-celled microbes, and the quick rate at which most players add organelles, the other species currently need a little push to keep from immediately being out-sized by the player.

src/auto-evo/simulation/foodsource/HeterotrophicFoodSource.cs
adQuid: It does simulate the chance of catching prey, in a very inaccurate and un-tested way. There's a few ways to catch something faster than you:
- prey is out of ATP and predator is not
- prey is flanked by two predators
- prey AI makes bad choices 
It also helps a predator not have to run from other things. Not precise, but representative.

The real motive was really to make sure that slightly larger heterotrophs would be encouraged to provide some more challenge in the way of predation. I admit I haven't tested this particular variable in isolation, so maybe it's unnessicary. This is also why I'm reluctant to name a constant here, but I'm not against it.

src/auto-evo/simulation/foodsource/HeterotrophicFoodSource.cs
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+
+        // Piluses are much more usefull if the microbe can close to melee
+        pilusScore *= predatorSpeed;
```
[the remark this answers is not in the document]
adQuid: If it were direct arithmetic (subtract or divide prey speed from predator speed), the relative values would not change between prospective predators for the same prey. This would only lower the weight of the pilus.

If there was a requirement to be faster than prey for the pilus to matter, I predict the result would be the pilus never evolving in the majority of game. The reality is that with the low amount of generations, it's hard to get dependent adaptations (like having a spike and being fast enough to stab anyone) to happen spontaneously.

src/auto-evo/simulation/foodsource/HeterotrophicFoodSource.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        // Intentionally don't penalize for osmoregulation cost to encourage larger monsters
+        return behaviorScore * (pilusScore + engulfScore + predatorSize + oxytoxyScore);
```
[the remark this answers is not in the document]
adQuid: I guess my main justification for larger predators is to be able to be "blocked" less often by other threatening microbes that can cut you out of chases. But as with other things in this PR, this is first a matter of gameplay. The auto-evo is comparing itself to the last generation of the player, and I expect the player to consistently add a couple organelles every generation by pure intelligent design. By pure evolutionary forces, we should expect the player to always be an apex predator unless they actively choose not to be.

As for endurance-chasing, I was planning to have metabolism factor in to this originally, but I put that to the side when working on sprint. I think I'll add this back in.

Maxonovien:
Just thought of something, I think at the moment a species with a lot of predators does not have fitness penalties, which could be a nice improvement (e.g. leading to larger photosynthetizers).

adQuid:
Sounds like a good feature to add in a future PR :)

Maxonovien:
Indeed, unless the balance ends up too off (hence the post here). But sure, I think it likely can wait.

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -134,139 +131,76 @@ private static void RunSimulationStep(SimulationConfiguration parameters, List<S
...
-                        {
-                            compoundUseScore += process.Process.Outputs[ATP]
-                                / process.Process.Inputs[compound] / Constants.AUTO_EVO_ATP_USE_SCORE_DIVISOR;
-                        }
-                    }
+                // Severely penalize a species that can't osmoregulate
+                if (ProcessSystem.ComputeEnergyBalance(currentSpecies.Organelles.Select(i => i.Definition),
+                    patch.Biome, currentSpecies.MembraneType).FinalBalance < 0)
+                {
+                    newPopulation /= 10;
```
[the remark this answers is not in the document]
adQuid: I've experimented with having this work off FinalBalanceStationary, and my results in 2/2 trials were species that start taking damage immediately after a couple generations. You could add a second if statement to do some kind of heavier penalty for the stationary cost and a lighter one for moving cost, but at the end of the day the AI doesn't have any semi-mobile behavior patterns. I think it's smarter to build auto-evo for the AI of today, then revisit when that assumption changes, than build it for something it doesn't do now and hope we get around to making the computer know how to play them later.

simulation_parameters/Constants.cs
```diff
@@ -397,10 +398,11 @@ public static class Constants
...
     public const float AUTO_EVO_GLUCOSE_USE_SCORE_DIVISOR = 1;
-    public const float AUTO_EVO_PILUS_PREDATION_SCORE = 1;
-    public const float AUTO_EVO_TOXIN_PREDATION_SCORE = 1;
-    public const float AUTO_EVO_PREDATION_ENERGY_MULTIPLIER = 0.5f;
-    public const float AUTO_EVO_SUNLIGHT_ENERGY_AMOUNT = 6000;
+    public const float AUTO_EVO_ENGULF_PREDATION_SCORE = 100;
+    public const float AUTO_EVO_PILUS_PREDATION_SCORE = 20;
+    public const float AUTO_EVO_TOXIN_PREDATION_SCORE = 10;
+    public const float AUTO_EVO_PREDATION_ENERGY_MULTIPLIER = 0.4f;
+    public const float AUTO_EVO_SUNLIGHT_ENERGY_AMOUNT = 600000;
```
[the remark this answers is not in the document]
adQuid: This number really represents a population capacity by another name (since you don't know how much energy a single organism will need to survive). Or at least a conversion of sunlight amount in the patch to population density. In the little arena the player swims around in, I don't know how that would reflect, if at all.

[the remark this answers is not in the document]
Maxonovien: Well, I'd say it would be great to be able to make a bridge between the two worlds, but I get it might be pretty specific to the current game state.

src/auto-evo/simulation/foodsource/HeterotrophicFoodSource.cs
Maxonovien: I'd say a constant is always useful to get a better grasp at what things are represented, and they may be tweaked easily from the Constants.cs file. But I won't say that's a necessary change either...

src/auto-evo/simulation/foodsource/PhotosyntheticFoodSource.cs
```diff
@@ -0,0 +1,51 @@
...
+                        photosynthesisingScore += process.Process.Outputs[atp]
+                            / process.Process.Inputs[sunlight] / Constants.AUTO_EVO_ATP_USE_SCORE_DIVISOR;
+                    }
+                }
+            }
+        }
+
+        // Moving too much can be harmfull
+        var energyCost = microbeSpecies.BaseOsmoregulationCost();
+        energyCost *= 1 + (microbeSpecies.Activity / Constants.MAX_SPECIES_ACTIVITY);
```
Maxonovien, line 42: Note Activity only determines sessility or not atm (below or above 40). But I think this is more of an issue regarding the current AI than your algorithm

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -965,38 +970,9 @@ public void RemoveOrganelle(Hex hex)
...
-
-        float baseMovementForce = Constants.CELL_BASE_THRUST *
-            (Membrane.MovementFactor - Rigidity * Constants.MEMBRANE_RIGIDITY_MOBILITY_MODIFIER);
-
-        float finalSpeed = (baseMovementForce + organelleMovementForce) / microbeMass;
-
-        return finalSpeed;
+        // TODO: Have this occur in a more central place
+        ApplyEditsToSpecies();
+        return ((MicrobeSpecies)editedSpecies.Clone()).BaseSpeed();
```
Maxonovien, line 975: I don't get the reason for cloning here

src/microbe_stage/MicrobeSpecies.cs
```diff
@@ -121,4 +121,39 @@ public override object Clone()
 
         return result;
     }
+
+    public float BaseOsmoregulationCost()
+    {
+        return Organelles.Sum(organelle => organelle.Definition.HexCount)
+            * MembraneType.OsmoregulationFactor
+            * Constants.ATP_COST_FOR_OSMOREGULATION;
```
Maxonovien, line 129: If I recall well, the editor uses a pretty similar method to compute osmoregulation, and thus it might be good to unify this in case further factors are added.

src/auto-evo/simulation/foodsource/PhotosyntheticFoodSource.cs
```diff
@@ -0,0 +1,51 @@
...
+                    {
+                        photosynthesisingScore += process.Process.Outputs[glucose]
+                            / process.Process.Inputs[sunlight] / Constants.AUTO_EVO_GLUCOSE_USE_SCORE_DIVISOR;
+                    }
+
+                    if (process.Process.Outputs.ContainsKey(atp))
+                    {
+                        photosynthesisingScore += process.Process.Outputs[atp]
+                            / process.Process.Inputs[sunlight] / Constants.AUTO_EVO_ATP_USE_SCORE_DIVISOR;
+                    }
```
Maxonovien, line 35: I don't remember any process producing directly atp from light, so this case is no useful yet, is it (seems like a copy-paste of chemosynthetic)?

Untrustedlife:
One minor suggestion, you could also specialize towards eating marine snow in the places where that exists.

src/auto-evo/simulation/foodsource/PhotosyntheticFoodSource.cs
adQuid: Normally I would be in the "don't futureproof" camp and want to delete this, but since a sunlight to ATP organelle _could_ exist just by changing a JSON file, I think this currently useless code is justified.

src/auto-evo/simulation/foodsource/PhotosyntheticFoodSource.cs
adQuid: After thinking about this, I decided to delete this altogether. This as an attempt to try and "push" autotrophs towards being less active in hopes that linages would slowly move towards being low enough to be sessile, but in practice I'm not seeing that happen. I think that will need to wait for a future PR.

src/microbe_stage/editor/MicrobeEditor.cs
adQuid: This was part of a bigger problem. In my latest push I just moved the functionality to the ProcessSystem class, which seems to be our de-facto calculation class.

src/microbe_stage/MicrobeSpecies.cs
adQuid: I went ahead and refactored the crap out of this area of the code; feel free to re-evaluate.

adQuid:
> One minor suggestion, you could also specialize towards eating marine snow in the places where that exists. That and just scavenging in general (basically detritovores)

My first thought would be to file that under "new feature for a new PR", but I do see how the mesopelagic and abyssopelagic zones will end up with no energy at all. That said, this is the case with the current auto-evo now that glucose decays, so moving forward without that addition shouldn't be making anything worse.

src/auto-evo/simulation/foodsource/PhotosyntheticFoodSource.cs
Maxonovien: I'd think so at first, but this organelle might use sunlight + other compound, which you do not account for. So I'd argue in favor of either making it stronger or scraping it all the like.

adQuid: You drive a hard bargain. Consider it gone.

adQuid: Besides, I think it's technically not photosynthesis by definition if it doesn't make Glucose.

hhyyrylainen: What if all of these were generalized so that the input compounds for all processes making ATP were taken into account, instead of keeping these hardcoded categories?

src/auto-evo/simulation/foodsource/PhotosyntheticFoodSource.cs
hhyyrylainen: Well species splitting is not implemented, which might help with life diversifying.

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -134,139 +131,78 @@ private static void RunSimulationStep(SimulationConfiguration parameters, List<S
...
+            var niches = new List<IFoodSource>
             {
-                var currentSpeciesEnergy = 0.0f;
-
-                currentSpeciesEnergy += sunlightInPatch
-                    * GetCompoundUseScore(currentSpecies, Sunlight) / totalPhotosynthesisScore;
-
-                currentSpeciesEnergy += hydrogenSulfideInPatch
-                    * GetCompoundUseScore(currentSpecies, HydrogenSulfide) / totalChemosynthesisScore;
+                new PhotosyntheticFoodSource(patch),
```
hhyyrylainen, line 142: I think better names for these would be like environmental food source, and limited compound food source. Because chemosynthesis is performed by a few specific organelles, so the naming in the code is confusing for me at least because terms that are used during gameplay are used differently here.

src/microbe_stage/MicrobeSpecies.cs
```diff
@@ -121,4 +121,9 @@ public override object Clone()
 
         return result;
     }
+
+    public float BaseSpeed()
```
hhyyrylainen, line 125: Use a property

adQuid: The value of BaseSpeed relies on the organelles, which are a mutable property. That leaves us with the choices of refactoring organelle layout to not allow modification without modifying speed, trusting that all future devs will remember you can't change one without the other, or making some redundant calculations for the dependent value.

src/auto-evo/simulation/foodsource/PhotosyntheticFoodSource.cs
adQuid: The danger of that is losing control of weights for other factors. We had already discussed earlier in this PR things like rust-eaters that sporadically move, and I would like to give the "less active plants are better" rule that I just took out. At this point, the photosynthesis are coming closer and closer together, so maybe they should merge, but I have a hunch in the longer term we will want to nudge the two to evolve differently in other ways.

src/auto-evo/simulation/foodsource/PhotosyntheticFoodSource.cs
adQuid: ^I can't believe I got this far along and only learned about that in the forum post lol

src/auto-evo/simulation/PopulationSimulation.cs
adQuid: See, both of those names sound too generic to me. All food sources are from the "environment", and prey species have "limited compounds"

hhyyrylainen: Wouldn't the names be *better*, then? As they apply to future compound cases as well.
If you want slap in environmental in the limited food source, and maybe name the other "patch level constant food source".

adQuid: But they're all constant by the patch. The difference is that photosynthesis calculates total energy slightly differently (https://github.com/Revolutionary-Games/Thrive/pull/2395/files#diff-5ef2a2af7e7495a66a2d7e97f255f818f642906b88ce01b1935197cdd8d975f8R15), and microbes don't need to move around to find the food, while chemosythesis requires that a microbe actively hunts compound clouds.

src/microbe_stage/MicrobeSpecies.cs
hhyyrylainen: I meant that literally convert this calculation to a get only property.
These kinds of methods that just compute or retrieve one value should be made as properties in C# and not methods.
I guess I've been coding so much C# lately that I didn't even consider that my comment could be misinterpreted...

src/auto-evo/simulation/foodsource/PhotosyntheticFoodSource.cs
hhyyrylainen: Well, then what about having the general step before that with manually coded tweak values for the specific compound types? (These could perhaps be defined in compounds.json for easy experimenting)

src/auto-evo/simulation/PopulationSimulation.cs
hhyyrylainen: I meant in the sense that constant = environmental. So meaning that those are available uniformly everywhere in the patch, in contrast to the limited compounds that could in theory be depleted (and that could be simulated) and also the microbes need to move (or rely on currents) to happen upon that resource.

adQuid: I've now gone halfway: I renamed chemosynthetic to compound, but I kept photosynthetic the same, since right now that class isn't generic enough to be used for any other uniform sources of energy. When those energy sources enter the game, we can always refactor and rename.

hhyyrylainen: I'm not super confident on that as I've seen too many PRs that go the minimal architectural, hacky route, and not doing proper refactoring, so I fear it won't get done later if it isn't done now...

adQuid: Naming a class to pretend to do something is very different then actually refactoring code. If there was another environmental energy source, we could have a conversation about altering the class to be able to do that to, but I can't write code to functionality that doesn't exist yet. Maybe sunlight and this other thing will be moved to some dictionary and the class can have a param like compounds. Maybe the two will be very different and I need to pass in a whole lambda. Maybe they will end up being so different we both agree there should be a new class. Maybe there won't be any other environmental food sources until this whole algorithm gets refactored away.

If there was a regression in functionality with this, I could see your case, as if this was some weird code that would need to be torn up to be extended. But right now this is just a class name.

hhyyrylainen: Maybe I should read through this code properly, but to me it seems simple, you'll have stuff like:
```
new UniformEnvironmentalFoodSource("sunlight", Constants.SUNLIGHT_ENERGY_CONTENT);
new UniformEnvironmentalFoodSource("heat", Constants.HEAT_ENERGY_CONTENT);
```
Something like that, so the difference is just what compound amount is looked up to give energy to the species that can use it. But if you absolutely hate that, I can read through this PR properly before commenting again.

hhyyrylainen: Having read the code properly now, it seems to be that I'm right. The only non-general part of PhotosyntheticFoodSource is that currently it's not possible to configure the compound it looks for and the modifier value for it's amount. So I don't see why the class needs to be so specifically named as it could with very small change actually work in general for any dissolved compound used in processes as input to get energy.

hhyyrylainen (changes requested):
Overall I'm fine with the overall direction this has taken (as long as playtesting this comes back positive). Though, I have quite a few smallish code related tweaks I'd like to see.

src/microbe_stage/ProcessSystem.cs
```diff
@@ -160,6 +160,35 @@ void MakeSureResultExists(Compound compound)
         return result;
     }
 
+    public static float CalculateSpeed(IEnumerable<OrganelleTemplate> organelles, MembraneType membraneType,
```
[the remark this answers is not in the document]
Kasterisk: IMO maybe more suitable to be put in the Microbe class as this method pertains to microbe's functioning.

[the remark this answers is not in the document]
hhyyrylainen: Well the Microbe class is already the longest class in the game, so if at all possible, I'd consider putting it somewhere else...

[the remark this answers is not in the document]
Maxonovien: Sounds like the `MicrobeSpeciesInfo` class suggested and implemented in #2427 could be used here. It could shorten the `Microbe.cs` file significantly if we switched some info to it. Although, speed depends on duplicated organelles so it can't be moved here.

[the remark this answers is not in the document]
hhyyrylainen: Well this specific calculation *could* as this is the normal movement speed of a species (without any duplicated organelles etc.)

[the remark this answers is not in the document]
adQuid: This was placed here in order to have the functionality be in a static library. Since the microbe editor makes this calculation with a detached set of organelles, I don't see a way to attach this to the microbe or microbe species class, or any property thereof. I don't love it here, nor am I a big fan of statics, so I'm open to suggestions.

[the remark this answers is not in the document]
hhyyrylainen: I think the suggestion was to still keep it static, but move it to a different class like Microbe or maybe MicrobeSpecies would make a tiny bit more sense (as the MicrobeSpecies uses the method to calculate the speed that it has).

src/auto-evo/simulation/PopulationSimulation.cs
adQuid: Would you like it to take in a string and throw an exception on any string other than "sunlight"?

hhyyrylainen: No... Like in that example code, I'd say it should either take in the string or a Compound instance *and* the resource multiplier. That way the class will work with any Compound. And the only limit is that it should be used on Compounds where it makes sense. No need to make a general class on purpose act in a non-general way.

Thrive.csproj
```diff
@@ -305,7 +309,6 @@
     <Folder Include="shaders\" />
     <Folder Include="src\microbe_stage\organelle_components\" />
     <Folder Include="src\auto-evo\steps\" />
-    <Folder Include="src\auto-evo\simulation\" />
```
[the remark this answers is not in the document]
adQuid: I have no idea. This is an auto-generated file: I ran build, it chose to re-write things this way, and the code still runs. I'm certainly not going to dive into a csproj file and pretend I know better than the compiler.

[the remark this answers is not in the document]
hhyyrylainen: Could you then reset this change? If there's no reason for it.

simulation_parameters/Constants.cs
```diff
@@ -393,15 +394,15 @@ public static class Constants
     public const int AUTO_EVO_MINIMUM_MOVE_POPULATION = 250;
     public const float AUTO_EVO_MINIMUM_MOVE_POPULATION_FRACTION = 0.1f;
     public const float AUTO_EVO_MAXIMUM_MOVE_POPULATION_FRACTION = 0.8f;
-
-    // Auto evo population algorithm tweak variables
```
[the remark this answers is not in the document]
adQuid: It's accurate, but it conveys no information not already found in variable names. I'm happy to put it back if you think it's critical for future developers to know that these variables are the "tweak" variables, as opposed to the non-tweaky ones.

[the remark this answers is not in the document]
hhyyrylainen: I think it is at least useful, as someone who wants to experiment tweaking the auto-evo results can start here. Whereas some other constants in the file will blow up the game if you touch them...

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -134,139 +131,78 @@ private static void RunSimulationStep(SimulationConfiguration parameters, List<S
...
+            foreach (var currentSpecies in species)
             {
-                foreach (var process in organelle.Definition.RunnableProcesses)
+                // Modify populations based on energy
+                var newPopulation = (long)(energyBySpecies[currentSpecies]
+                    / ProcessSystem.ComputeEnergyBalance(
+                        currentSpecies.Organelles.Organelles.Select(organelle => organelle.Definition),
+                        patch.Biome, currentSpecies.MembraneType).FinalBalanceStationary);
+
+                // Severely penalize a species that can't osmoregulate
```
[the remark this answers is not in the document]
adQuid: is that like a spellchecker or something?

src/microbe_stage/ProcessSystem.cs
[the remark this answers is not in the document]
adQuid: You want a static method in the microbespecies class, that takes in a list of organelle definitions unrelated to the organelles the species has, and calculates a base speed for that? That seems like a clear violation of Single Responsibility Principle.

Thrive.csproj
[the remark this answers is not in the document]
adQuid: Anything I change by hand in this file should be overwritten by the next VS build, and I'm not sure there's no reason for it. It's an auto-generated file.

src/auto-evo/simulation/PopulationSimulation.cs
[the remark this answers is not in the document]
hhyyrylainen: Yes, which sadly seems to only work in Rider, and the jetbrains CLI tools don't seem to enforce spelling for some reason I can't figure out. 
I'll open an issue to remember to investigate: https://github.com/Revolutionary-Games/Thrive/issues/2453

src/microbe_stage/ProcessSystem.cs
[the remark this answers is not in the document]
hhyyrylainen: Well if you want to make a new helper class with a good name go ahead, but I can personally say that this makes more sense to be in MicrobeSpecies *than* the process system as the movement organelles have *no* relation to processes. So this currently makes even less sense than where I suggested to put this.

Thrive.csproj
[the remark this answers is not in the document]
hhyyrylainen: It might be auto-generated but Rider will act differently depending on which folders are listed here, so I've added many of these through Rider manually.

src/microbe_stage/ProcessSystem.cs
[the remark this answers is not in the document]
Maxonovien: I feel it would make sense in the `MicrobeSpecies` class as well. But I don't see the necessity for static methods depending on the organelles of the species. Granted, you'd need to create a new species for every mutated case, but as you use it within an encapsulated framework, you could use some garbage collection method C# has to prevent memory overuse. Not optimal, but it makes sense at least.

Other possibility, you could create that method, but now you'd have to pass it organelles changes, that is how many of each type was added/deleted (e.g. as a `Dictionary<OrganelleTemplate, int>`, and then compute with this correction. It could even still give you the base speed of the species if you pass it an empty dict!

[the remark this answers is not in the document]
adQuid: In auto-evo new species are already being made for every mutation, so that's not the issue. The issue is reusing it with the code in the microbe editor, which has no species to work with at all. Building a new species with every change was what I originally tried to work with in my take on the dynamic MP PR, but you can review that PR for why that became a mess in its own right.

src/auto-evo/simulation/PopulationSimulation.cs
adQuid: Ok, that makes a lot of sense. I failed to see the second param.

hhyyrylainen:
CI is right, this doesn't even compile: `Source file '/home/hhyyrylainen/Projects/thrive/src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs' could not be found.`

the entire food_source folder is missing when I checkout this branch (and pulled in the latest changes).

adQuid:
I forgot to add the  renamed file in git. Sorry for wasting your time.

Thrive.csproj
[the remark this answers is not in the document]
Maxonovien: From the file diff, I don't see it added back  in the `Folder` `ItemGroup`. Am I missing something?

Thrive.csproj
```diff
@@ -305,7 +310,7 @@
...
+	<Folder Include="src\auto-evo\simulation\" />
```
adQuid, line 313:
```suggestion
    <Folder Include="src\auto-evo\simulation\" />
```

src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
```diff
@@ -0,0 +1,49 @@
...
+        biomeConditions = patch.Biome;
+        this.compound = SimulationParameters.Instance.GetCompound(compound);
+        totalEnvironmentalEnergySource = patch.Biome.Compounds[this.compound].Dissolved * constant;
+    }
+
+    public float FitnessScore(Species species)
+    {
+        var microbeSpecies = (MicrobeSpecies)species;
+
+        var photosynthesisingScore = 0.0f;
```
Maxonovien, line 22: This variable name is much more specific than the one of the class (environmental source), so you may want to change this.

src/auto-evo/simulation/food_source/CompoundFoodSource.cs
```diff
@@ -0,0 +1,61 @@
...
+                        compoundUseScore += process.Process.Outputs[atp]
+                            / process.Process.Inputs[compound] / Constants.AUTO_EVO_ATP_USE_SCORE_DIVISOR;
+                    }
+                }
+            }
+        }
+
+        var energyCost = ProcessSystem.ComputeEnergyBalance(
+            microbeSpecies.Organelles.Organelles,
+            biomeConditions, microbeSpecies.MembraneType).FinalBalanceStationary;
```
Maxonovien, line 52: You have the very same code in EnvirontalFoodSource, minus ATP (which could be added back in Environmental food source as the class is now more generic, btw). You might want to factor it somehow. 

Perhaps an intermediate interface for these two classes, helpers, or even rely on the `MicrobeSpecies.cs` class for some parts?

(EDIT: After further reading it seems like `MicrobeInternalCalculation` could be of use, although it seems linked to microbes rather than species)

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -0,0 +1,78 @@
...
+        predatorSpeed += ProcessSystem
+            .ComputeEnergyBalance(microbeSpecies.Organelles.Organelles, patch.Biome,
+                microbeSpecies.MembraneType).FinalBalance;
+
+        // It's great if you can engulf this prey, but only if you can catch it
+        var engulfScore = predatorSize / preySize > Constants.ENGULF_SIZE_RATIO_REQ
+            && !microbeSpecies.MembraneType.CellWall ?
+                Constants.AUTO_EVO_ENGULF_PREDATION_SCORE :
+                0.0f;
+        engulfScore *= predatorSpeed > preySpeed ? 1.0f : Constants.AUTO_EVO_SLOW_ENGULF_MULTIPLIER;
```
Maxonovien, line 46: Although I knew the code already, I did not guess that the constant was to be used there when I saw it defined. 

It made me think of a multiplier for a slow engulf, rather than slow "engulfer"/predator. Perhaps there is a more suitable name for the constant?

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -0,0 +1,78 @@
...
+                pilusScore += Constants.AUTO_EVO_PILUS_PREDATION_SCORE;
+                continue;
+            }
+
+            foreach (var process in organelle.Definition.RunnableProcesses)
+            {
+                if (process.Process.Outputs.ContainsKey(oxytoxy))
+                {
+                    oxytoxyScore += Constants.AUTO_EVO_TOXIN_PREDATION_SCORE;
+                }
```
Maxonovien, line 63: This does not take into account the output quantity of the oxytoxy production process. Given that their osmoregulation cost is higher, it penalizes the use of eukaryote organelles, even though they produce more, and at a lower overall cost.

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        float organelleMovementForce = 0;
+
+        foreach (var organelle in organelles)
+        {
+            microbeMass += organelle.Definition.Mass;
+
+            if (organelle.Definition.HasComponentFactory<MovementComponentFactory>())
+            {
+                if (organelle.Orientation == 3)
```
Maxonovien, line 18: This is likely due to the implementation of organelles, but checking if an orientation equals 3 utterly confuses me.

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -0,0 +1,33 @@
...
+                {
+                    organelleMovementForce += Constants.FLAGELLA_BASE_FORCE
+                        * organelle.Definition.Components.Movement.Momentum / 100.0f;
```
Maxonovien, line 21: Why divide it by 100.0f?

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -0,0 +1,33 @@
...
+                }
+            }
+        }
+
+        float baseMovementForce = Constants.CELL_BASE_THRUST *
+            (membraneType.MovementFactor - membraneRigidity * Constants.MEMBRANE_RIGIDITY_MOBILITY_MODIFIER);
+
+        float finalSpeed = (baseMovementForce + organelleMovementForce) / microbeMass;
```
Maxonovien, line 29: I guess the approximation is fine, but I truly find it disturbing that a speed is computed as a force over a mass (which should be a speed/s, that is an acceleration).

src/microbe_stage/ProcessSystem.cs
```diff
@@ -43,6 +43,15 @@ public ProcessSystem(Node worldRoot)
         return result;
     }
 
+    /// <summary>
+    ///   Computes the energy balance for the given organelles in biome
+    /// </summary>
+    public static EnergyBalanceInfo ComputeEnergyBalance(IEnumerable<OrganelleTemplate> organelles,
+        BiomeConditions biome, MembraneType membrane)
+    {
+        return ComputeEnergyBalance(organelles.Select(i => i.Definition), biome, membrane);
```
Maxonovien, line 52: You used `organelles.Select(organelle => organelle.Definition)` in `MicrobeEditor.cs`, so I'd advocate for consistency in one way or another.

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -1248,14 +1217,15 @@ private void CalculateOrganelleEffectivenessInPatch(Patch patch = null)
...
+            ProcessSystem.ComputeEnergyBalance(organelles, patch.Biome, membrane));
     }
 
     private void CalculateCompoundBalanceInPatch(List<OrganelleTemplate> organelles, Patch patch = null)
     {
         patch ??= CurrentPatch;
 
-        var result = ProcessSystem.ComputeCompoundBalance(organelles.Select(i => i.Definition), patch.Biome);
+        var result = ProcessSystem
+            .ComputeCompoundBalance(organelles.Select(organelle => organelle.Definition), patch.Biome);
```
Maxonovien, line 1228: There are two calls to ProcessSystem.ComputeXBalance functions here, but one of them takes an `OrganelleTemplate` list, and the other the list of their definitions. I think some consistent choice should be made here.

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
adQuid: upon re-read I totally agree. Any suggestions?

src/auto-evo/simulation/food_source/CompoundFoodSource.cs
adQuid: This gets into a more esoteric argument about software design (which I'm happy to get in if you really want to), but I would rather let this stand.  In short, code is duplicated in the literal sense, but these two algorithms are responsible for different things, and I expect them to evolve in separate directions over time.

src/microbe_stage/MicrobeInternalCalculations.cs
adQuid: inherited code. I assume that we can take that off and just make the constant smaller, but that's going to increase testing scope on what's already a big PR.

src/microbe_stage/MicrobeInternalCalculations.cs
adQuid: "fine" is good enough for me right now :)

(this is also code from before the PR)

src/microbe_stage/ProcessSystem.cs
adQuid: I've converted everything doing that select to use this overload instead. Good catch.

src/auto-evo/simulation/food_source/CompoundFoodSource.cs
Maxonovien: I agree but for this part.
> I expect them to evolve in separate directions over time
Still, I can't affirm that I am right in my expectations so I do not have further objection to keeping them duplicated.

Nonetheless, I'd think that compound use score could be useful in general and could thus benefit from a perhaps parametric computation (e.g. passing multiplier by output compound). This is really at the edge of the PR's scope, so I'm open to leaving it for a potential future refactory.

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
Maxonovien: To me it really feels like as the probability of a lucky catch. `ENGULF_LUCKY_CATCH_PROBABILITY` perhaps? Or `LUCKY_ENGULF_PROBABILITY`? If you don't want to state it as a proba, you may prefer `ENGULF_CHANCE_FACTOR` or something along these lines?

src/microbe_stage/MicrobeInternalCalculations.cs
adQuid: I added a comment to try to help. It's a sloppy way to ensure that neither the editor nor auto-evo factors in forward-facing flagella for speed calculations.

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
adQuid: I used ENGULF_LUCKY_CATCH_PROBABILITY. Seems accurate enough.

src/microbe_stage/MicrobeInternalCalculations.cs
Maxonovien: I'm still pretty wary of this legacy code now. But I think there would be much to change, so we should perhaps leave it aside for now.

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
Maxonovien: I assumed you'd keep the prefix by the way, although it could perhaps be generalized outside of auto evo later on.

adQuid: no no, you're right again. I'm changing it.

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -0,0 +1,78 @@
...
+        preySize = prey.Organelles.Organelles.Sum(organelle => organelle.Definition.HexCount);
+        patch.SpeciesInPatch.TryGetValue(prey, out long population);
+        totalEnergy = population * prey.Organelles.Count * Constants.AUTO_EVO_PREDATION_ENERGY_MULTIPLIER;
+    }
+
+    public float FitnessScore(Species species)
+    {
+        var microbeSpecies = (MicrobeSpecies)species;
+
+        // No canibalism
```
Maxonovien, line 27:
```suggestion
        // No cannibalism
```

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -0,0 +1,78 @@
...
+        var predatorSpeed = microbeSpecies.BaseSpeed;
+        predatorSpeed += ProcessSystem
+            .ComputeEnergyBalance(microbeSpecies.Organelles.Organelles, patch.Biome,
+                microbeSpecies.MembraneType).FinalBalance;
+
+        // It's great if you can engulf this prey, but only if you can catch it
+        var engulfScore = predatorSize / preySize > Constants.ENGULF_SIZE_RATIO_REQ
+            && !microbeSpecies.MembraneType.CellWall ?
+                Constants.AUTO_EVO_ENGULF_PREDATION_SCORE :
+                0.0f;
```
Maxonovien, line 45: This ternary might be a bit more readable as an if statement

src/microbe_stage/ProcessSystem.cs
Maxonovien, line 52: Aside from the signature you changed already, shouldn't there be some consistency for the LINQ statements, which are sometimes `organelle => organelle.Definition` (see l.149 of this file) and sometimes `i => i.Definition`?

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        float organelleMovementForce = 0;
+
+        foreach (var organelle in organelles)
+        {
+            microbeMass += organelle.Definition.Mass;
+
+            if (organelle.Definition.HasComponentFactory<MovementComponentFactory>())
+            {
+                // only count flagella that face backwards
```
Maxonovien, line 18:
```suggestion
                // Only count flagella that face backwards
```

src/auto-evo/simulation/food_source/IFoodSource.cs
```diff
@@ -0,0 +1,12 @@
+﻿public interface IFoodSource
+{
+    public float TotalEnergyAvailable();
+
+    /// <summary>
+    ///   Provides a fitness metric to determine population adjustments for species in a patch.
+    /// </summary>
+    /// <param name="microbe">The species to be evaluated.</param>
+    /// <returns>A float to represent score. Scores are only compared against other scores from the same FoodSource,
+    /// so different implementations do not need to worry about scale.</returns>
```
Maxonovien, line 10: If this starts to be multiline, then you might as well want a more structured version of this comment, such as the one used for `<summary>`, that is returning to the line and add spaces.

src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    private BiomeConditions biomeConditions;
+    private float totalEnvironmentalEnergySource;
+
+    public EnvironmentalFoodSource(Patch patch, string compound, float constant)
+    {
+        biomeConditions = patch.Biome;
+        this.compound = SimulationParameters.Instance.GetCompound(compound);
+        totalEnvironmentalEnergySource = patch.Biome.Compounds[this.compound].Dissolved * constant;
+    }
```
Maxonovien, line 14: Perhaps the constant might benefit for a more explicit name / a documentation.

Maxonovien (commented):
I'm overall very happy with the code as it stands, but here's some nitpicking anyways. :)

src/microbe_stage/ProcessSystem.cs
adQuid: I'm happy to change them around if you really want, but I think that anonymous functions have that liberty. I don't like "i" as an alias now that I look at it (this is inherited code I moved), but in the future I suggest we not try to tie every LINQ statement together like that. This isn't a data contract; it's intentionally limited to internal scope, specifically so we DONT have to worry about that.

hhyyrylainen: Latest styleguide and what I've done with ThriveDevCenter is to use the first letter of the variable name / class for the LINQ statement. So for example here all of these examples should be changed to be `o => o.Definition`

adQuid: Done, but I sure hope you don't expect me to enact this change across the codebase in this PR...

hhyyrylainen: I don't. I guess on my 100+ items TODO list it goes...

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -0,0 +1,80 @@
...
+        predatorSpeed += ProcessSystem
+            .ComputeEnergyBalance(microbeSpecies.Organelles.Organelles, patch.Biome,
+                microbeSpecies.MembraneType).FinalBalance;
+
+        // It's great if you can engulf this prey, but only if you can catch it
+        var engulfScore = 0.0f;
+        if (predatorSize / preySize > Constants.ENGULF_SIZE_RATIO_REQ && !microbeSpecies.MembraneType.CellWall)
+        {
+            engulfScore = Constants.AUTO_EVO_ENGULF_PREDATION_SCORE;
+        }
```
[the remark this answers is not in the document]
Maxonovien: I think it is alright as it stands. I'm never opposed to more readability, but I think it is more than fine here

Maxonovien (approved):
I have nothing else to add on the code. 

I playtested this for 20-30 generations and found a very lively microbial scenery. Auto-evo will start to take some time as more species appear, but it stays within a reasonable time (tens of seconds, at most) even after a while.

Overall I strongly like the changes brought by this PR. Approved! :)

src/microbe_stage/MicrobeInternalCalculations.cs
hhyyrylainen: Well this is actually entirely incorrect. The flagella thrust is entirely dependent on where it ends up being with the external organelle positioning. The orientation has 0 effect on the actual force it can generate...

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -965,38 +965,7 @@ public void RemoveOrganelle(Hex hex)
...
-        foreach (var organelle in editedMicrobeOrganelles.Organelles)
-        {
-            microbeMass += organelle.Definition.Mass;
-
-            if (organelle.Definition.HasComponentFactory<MovementComponentFactory>())
-            {
-                Vector3 organelleDirection = (Hex.AxialToCartesian(new Hex(0, 0))
-                    - Hex.AxialToCartesian(organelle.Position)).Normalized();
-
-                float directionFactor = organelleDirection.Dot(forwardsDirection);
```
hhyyrylainen, line 983: This is more correct math, but even this isn't 100% exact, but if you move back to this math, I'd consider it good enough (with a TODO comment linking https://github.com/Revolutionary-Games/Thrive/issues/2095)

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -134,139 +131,78 @@ private static void RunSimulationStep(SimulationConfiguration parameters, List<S
...
-
-        private static float GetPredationScore(MicrobeSpecies species)
-        {
-            var predationScore = 0.0f;
 
-            foreach (var organelle in species.Organelles)
+            foreach (var niche in niches)
             {
-                if (organelle.Definition.HasComponentFactory<PilusComponentFactory>())
+                // If this isn't a source of energy here, no need for more calculations
```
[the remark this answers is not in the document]
hhyyrylainen: This was not addressed.

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -134,139 +131,78 @@ private static void RunSimulationStep(SimulationConfiguration parameters, List<S
...
+                if (niche.TotalEnergyAvailable() == 0.0f)
```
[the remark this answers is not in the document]
hhyyrylainen: Neither was this...

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -134,139 +131,78 @@ private static void RunSimulationStep(SimulationConfiguration parameters, List<S
...
+                var totalNicheFitness = 0.0f;
+                foreach (var currentSpecies in species)
                 {
-                    if (process.Process.Outputs.ContainsKey(Oxytoxy))
-                    {
-                        predationScore += Constants.AUTO_EVO_TOXIN_PREDATION_SCORE;
-                    }
+                    // Softly enforces https://en.wikipedia.org/wiki/Competitive_exclusion_principle
+                    // by exaggerating fitness differences
+                    var thisSpeciesFitness = Math.Max((float)Math.Pow(niche.FitnessScore(currentSpecies), 2.5f), 0.0f);
```
hhyyrylainen, line 167: `Mathf` ? if the result is anyway cast to float right away

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -134,139 +131,78 @@ private static void RunSimulationStep(SimulationConfiguration parameters, List<S
...
+                    fitnessBySpecies[currentSpecies] = thisSpeciesFitness;
+                    totalNicheFitness += thisSpeciesFitness;
                 }
-            }
 
-            return predationScore;
-        }
+                // If no species can get energy this way, no need for more calculations
+                if (totalNicheFitness == 0.0f)
```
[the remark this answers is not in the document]
hhyyrylainen: also wasn't addressed...

src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
```diff
@@ -0,0 +1,47 @@
...
+        var microbeSpecies = (MicrobeSpecies)species;
+
+        var energyCreationScore = 0.0f;
+        foreach (var organelle in microbeSpecies.Organelles)
+        {
+            foreach (var process in organelle.Definition.RunnableProcesses)
+            {
+                if (process.Process.Inputs.ContainsKey(compound))
+                {
+                    if (process.Process.Outputs.ContainsKey(glucose))
```
hhyyrylainen, line 27: Wouldn't it be also nice to check for ATP production here?
Someone might be surprised in the future if this doesn't work similarly in checking the process outputs as the compound food source (might be worth it to make a general helper method to calculate the score from process outputs).

src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
```diff
@@ -0,0 +1,47 @@
...
+                    {
+                        energyCreationScore += process.Process.Outputs[glucose]
+                            / process.Process.Inputs[compound] / Constants.AUTO_EVO_GLUCOSE_USE_SCORE_DIVISOR;
```
hhyyrylainen, line 30: Why is `AUTO_EVO_GLUCOSE_USE_SCORE_DIVISOR` a divisor and not a multiplier like `0.5f`? Floating point divides are much more expensive than multiplies, and it is very easy to convert constants to multiply operations.

hhyyrylainen (changes requested):
Here's another round of comments. I've read through all the code here now so these should be the last ones to solve.

src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
adQuid: I'm happy to take either path if you and @Maxonovien can agree. He argued that the atp calculation should be taken out as dead code.

hhyyrylainen: I wouldn't be dead code, right, as it is used by 50% of the implementations.
And if someone makes a mod where an environmental resource makes ATP, they are going to be surprised that auto-evo refuses to evolve that.

Maxonovien: The mod argument is a satisfying reason for me to keep it ATP calculations

adQuid: Moving the code to a helper would fix this, but since I assume that one of the next steps here would be to differentiate the two (most likely by adding in movement factors to the compound source), I was resisting doing that.

Maxonovien: You are right in the sense that it is dead code currently. But it might be ran if someone decided to mod photosynthesis to produce ATP instead of glucose. And with the current changes they'll be able to do so only by modifying the .json files, and not touch the actual code. 

This is why it is satisfying enough to me as an argument, for the sake of mod support. The other approach of removing dead code still makes sense in my view, but I for one am open to any decision here, so I have no issue with hh's suggestion.

hhyyrylainen: Wouldn't the movement score factors be applied *after* the general computation? So a shared code implementation can be used for that shared part.

adQuid: I don't know. I could see it being implemented that way, I could see it being implemented some other way. Ultimately, this section of code just iterations and counts. I don't foresee the "algorithm" changing, and it would be tricky to make a utility function that is referentially transparent and doesn't make several assumptions about the code calling it. Code reuse is nice, but when you're breaking Single Responsibility Principle to do that, that's when I say you're going too far.

I'd invest more time in this if it was more complicated logic (duplicating the movement logic in the first draft of this PR was dumb and I'm ashamed of that), but this is really just a sum calculation with more complicated accessors.

hhyyrylainen: But still, the operation here is the general one "how much energy / useful resource output this process has". According to least astonishment principle that part should not work differently elsewhere. If someone ever adds an environmental energy source that makes ATP directly, they *will* be surprised why auto-evo doesn't want to evolve their cool new organelles.
That's why I want this operation put into a single place, because if there are multiple implementations they may become out of sync (and already has) so they will astonish someone who expects the same operation to work the same in a slightly different context.

hhyyrylainen: Something like:
```
compoundUseScore = 0;
                    if (process.Process.Outputs.ContainsKey(glucose))
                    {
                        compoundUseScore += process.Process.Outputs[glucose]
                            / process.Process.Inputs[compound] * Constants.AUTO_EVO_GLUCOSE_USE_SCORE_MULTIPLIER;
                    }

                    if (process.Process.Outputs.ContainsKey(atp))
                    {
                        compoundUseScore += process.Process.Outputs[atp]
                            / process.Process.Inputs[compound] * Constants.AUTO_EVO_ATP_USE_SCORE_MULTIPLIER;
                    }
return compoundUseScore;
```
then calling code has something like: `compoundUseScore += CalculateProcessOutputScore(process)`
it could be a protected method in the abstract base class for all energy sources (which would handle also looking up glucose and ATP).

adQuid: I was overestimating the difficulty in this refactor. It was actually super easy to make this cleanly.

hhyyrylainen: One thing that I want to say as slight criticism of your latest commit is that it now looks up the energy sources exponential number of times, as each auto-evo step * each patch * each energy source * each organelle * each process,
is worse than it was before: auto-evo step * each patch * each energy source
so I'd like a way to still keep the once per energy source lookup of the Compound instances.

hhyyrylainen (approved):
I fixed the removed line in the .csproj, and changed one thing to protected.
I'm now done trying to look through this PR, so I'll accept this (as long as you are also fine with the changes I just made).

adQuid:
Both of those changes make a lot of sense to me. Thank you for all the patience with this one; I would not have fought so hard for this PR if I didn't really believe it was going to add value.

hhyyrylainen:
Oh no
It should have been squashed...

hhyyrylainen:
I was two seconds away from doing it properly...

hhyyrylainen:
There's way too many dirty commits for me to accept this, I'll rewrite master history...

adQuid:
omg I'm so sorry. Is there anything I can do to help, or should I just not touch anything at this point?

hhyyrylainen:
Don't touch anything. I need to disable branch protection and force-push onto master, then re-merge this from a backup branch I just made.

PR #1739, opened 2020-10-23
1550 reduse compound spawn while still

Straywind:
closes #1550 

I just want some thought on whether or not you're happy with the way I'm going about it. This is still bare bones, and I will change quite a lot to make it work the way we want, but I just want thoughts on the way I'm going about it before I waste too much time on one way, just to do it another way.

PR #2397, opened 2021-07-10
1560 regular spawn

### LINKED ISSUE #1541: Investigate why there is no iron spawning in the cave

### LINKED ISSUE #1550: Reduce compound spawns while the player is stationary
to not make it too OP

I would implement it with a variable tracking how much the player has moved, if the player has not moved much for a while I'd start to reduce the chance that compound clouds are spawned.

### LINKED ISSUE #1560: Make cloud spawns less random and more time based
Instead of the totally chance dependent cloud spawning I think we should move to a timer based approach that would guarantee clouds spawn in proportion with their relative densities and some (configurable) max rate.

A bit similar to: https://github.com/Revolutionary-Games/Thrive/issues/1551

### LINKED ISSUE #1626: Investigate why chunk spawn rate is way down

### LINKED ISSUE #1777: Check if hydrogen sulfide spawns in the cave

Maxonovien:

This PR takes on #2138 (algorithms only) to rework the spawn system. It tries to stick to the old implementation.

It brings two major changes:
1. Spawners are now drawn from a ShuffleBag. This enumerable structure allows to draw items from a list, each item being only drawn once. When empty, the bag refills the list and uniformely shuffles it. This means that over a patch (with no spawner change), each spawner get as many opportunities to start spawning things (with varying success) as the others. This favors spawn diversity (spawners may still fail to spawn).
2. To avoid 'belt-spawning' (see image) when the player is not moving, this PR introduces a cap of 15 entities spawning around an immobile player, 'immobile being defined here as not leaving a small specific zone.  Also, when immobile, spawning is no longer bias in favor of the player's orientation, but uniformely distributed around).

image

Note: further issues are related but could be dealt with another PR:
1. Per-species cap (issues 1261 and 1551, not tagged as not solving) 
2. Contextual encounters (issue 2055)
3. Keeps attempts based system, which uses caps breaking linearity of success with density (related to issue 2055).

**Related Issues**

Fixes #1550 . Closes #1739 
Also resolves #1541 and resolves #1626 and resolves #1777 . Partly overlaps with and closes #2389.

closes #2138 (previous PR)

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/general/ShuffleBag.cs
```diff
@@ -0,0 +1,137 @@
...
+            FillAndShuffle();
+        }
+
+        return currentContent[currentContent.Count - 1];
+    }
+
+    private bool Drop()
+    {
+        if (currentContent.Count == 0)
+            throw new NullReferenceException("Cannot drop from empty bag!");
```
FuJa0815, line 98: use an `InvalidOperationException` here. It isn't really a null reference problem.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -230,6 +251,26 @@ private void SpawnEntities(Vector3 playerPosition, Vector3 playerRotation, int e
         if (existing >= maxAliveEntities)
             return;
 
+        // Here we want to check that the player move to not basically spawn in circle around him.
```
FuJa0815, line 254: moved

FuJa0815 (changes requested):

Thrive.csproj
```diff
@@ -71,6 +71,7 @@
     <Compile Include="src\general\StringUtils.cs" />
     <Compile Include="src\general\TranslationHelper.cs" />
     <Compile Include="src\general\TypeUtils.cs" />
+	<Compile Include="src\general\ShuffleBag.cs" />
```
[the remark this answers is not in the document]
Maxonovien: I did! Otherwise Godot would have been endlessly complaining. I didn't know there was a non manual way to do this.
(Also I wasn't able to open this file in VisualStudio).

It comes from an automatic tab indentation from my text editor, which I've fixed. Still, I'm surprised that this passed the auto-checks.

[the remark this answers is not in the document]
Kasterisk: I usually create a new C# script through the Godot editor so that way I think Godot could register the file and automatically add it into the csproj. I'm assuming you uses an external editor which in this case probably the reason why it didn't get added because it kind of bypass Godot's detection. I personally haven't tried creating new scripts externally (surprisingly) so couldn't really confirm.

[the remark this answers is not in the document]
Maxonovien: Oh, this definitely is the reason. I create my files directly through VisualStudio (as I'm not used to Godot), and indeed it is not detected before I modify the .csproj. I'll do it from Godot next time!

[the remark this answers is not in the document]
hhyyrylainen: I create (most) new script files with Rider, and that properly updates the csproj.

src/general/ShuffleBag.cs
```diff
@@ -0,0 +1,137 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+
+public class ShuffleBag<T> : IEnumerable<T>
```
FuJa0815, line 5: As this looks like a class designed for more general use, this should be documented better

src/general/ShuffleBag.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        object IEnumerator.Current => Current;
+
+        public void Dispose()
+        {
+            // If we have exhausted the bag, fill and shuffle it for the next one.
+            if (sourceBag.currentContent.Count == 0)
+                sourceBag.FillAndShuffle();
+
+            // TODO CLEAR MEMORY
```
FuJa0815, line 123: what is meant by clear memory?

src/microbe_stage/SpawnSystem.cs
```diff
@@ -40,6 +40,13 @@ public class SpawnSystem
     [JsonProperty]
     private int maxAliveEntities = 1000;
 
+    /// <summary>
+    ///   This limits the number of things that can be spawned in a single spawn radius.
+    ///   Used to limit items spawning in one single circle when the player doesn't move.
+    /// </summary>
+    [JsonProperty]
+    private int maxEntitiesInSpawnRadius = 15;
```
FuJa0815, line 48: shouldn't this be a constant in the constants class?

src/microbe_stage/SpawnSystem.cs
```diff
@@ -68,9 +75,23 @@ public class SpawnSystem
...
+    /// <summary>
+    ///   Last recorded position of the player. Positions are recorded upon leaving the immobility zone.
+    /// </summary>
+    private Vector3 lastRecordedPlayerPosition;
+
     public SpawnSystem(Node root)
     {
         worldRoot = root;
+        spawnTypes = new ShuffleBag<Spawner>(random);
+        lastRecordedPlayerPosition = new Vector3(0, 0, 0);
```
FuJa0815, line 93: Vector3.Zero

src/microbe_stage/SpawnSystem.cs
```diff
@@ -230,6 +251,26 @@ private void SpawnEntities(Vector3 playerPosition, Vector3 playerRotation, int e
         if (existing >= maxAliveEntities)
             return;
 
+        // Here we want to check that the player moved to not basically spawn in circle around him.
+        // Solution inspired by gwen is to check if the player
+        // moves out of a square/cycle around his precedent registered position
+        // Not perfect however as going on and off could still break this.
+        float distanceToLastPosition = (playerPosition - lastRecordedPlayerPosition).Length();
+        bool immobilePlayer = distanceToLastPosition < Constants.PLAYER_IMMOBILITY_ZONE_RADIUS;
```
FuJa0815, line 259: Make a `PLAYER_IMMOBILLITY_ZONE_RADIUS_SQUARED` and use `LengthSquared`. This is faster.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -230,6 +251,26 @@ private void SpawnEntities(Vector3 playerPosition, Vector3 playerRotation, int e
...
+
+        if (immobilePlayer)
+        {
+            // If the player is staying inside a circle around he's previous position, only go up to the local spawn cap
```
FuJa0815, line 263:
```suggestion
            // If the player is staying inside a circle around his previous position, only spawn up to the local spawn cap
```

FuJa0815 (changes requested):

src/microbe_stage/SpawnSystem.cs
Maxonovien: I kept the same logic as for `maxAliveEntities`, but I'm definitely open to move both to `Constants.cs`

src/general/ShuffleBag.cs
Maxonovien: Given that I'm not familiar with enumerators (especially in C#), there may be some memory cleaning to do when disposing the enumerator. If not, then nothing is to be done, but I added this comment just in case.

Made it a clearer comment in the method documentation though.

hhyyrylainen: I mean this class doesn't hold any disposable resources, so there is nothing to dispose here.
One thing I do want to criticise here is that this relies on being *disposed* to trigger bag re-fill, which I fear is actually problematic *if* someone gets the enumerator without using a foreach as I think that automatically will call dispose (I *think*).

Maxonovien: You can actually trigger bag refill with the `Reset()` method. Dispose does so as well because foreach does not call `Reset` (writing these lines I realize I should perhaps call `this.Reset()` instead, btw?).

After that, if the coder wants to keep a non-reset enumerator... well, that's possible, as for a list enumerator you do not reset.

hhyyrylainen: I think it would make a ton more sense to reset the sequence when *asking* for an enumerator, instead of this way. C# is not C++ so RAII is not really a thing, ie. object destruction needs to not be an important operation / relied upon happening...

src/microbe_stage/SpawnSystem.cs
hhyyrylainen: I think this is a field because in theory this could change, I think? That's why it is also saved to keep loaded saves consistent.

src/general/ShuffleBag.cs
Maxonovien: That's a fair point. I made the filling occur at the creation of the enumerator if the bag is empty.

src/microbe_stage/SpawnSystem.cs
Maxonovien: Then I'll keep the same approach here. If this really ought to be changed for some reason I'll leave it to another PR.

FuJa0815 (dismissed):

src/general/ShuffleBag.cs
```diff
@@ -0,0 +1,212 @@
...
+    ///   If the bag is empty, it refills it beforehand.
+    /// </summary>
+    private T Pick()
+    {
+        if (currentContent.Count == 0)
+        {
+            FillAndShuffle();
+        }
+
+        return currentContent[currentContent.Count - 1];
```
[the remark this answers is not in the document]
hhyyrylainen: I would guess that it's because the list is popped from the back (as popping from the front is very bad for performance).

src/microbe_stage/SpawnSystem.cs
```diff
@@ -68,9 +75,23 @@ public class SpawnSystem
...
+    /// <summary>
+    ///   Estimate count of existing spawn entities within the current spawn radius of the player;
+    ///   Used to prevent spawn belt when player doesn't move.
+    /// </summary>
+    private int estimateEntityCountInSpawnRadius;
+
+    /// <summary>
+    ///   Last recorded position of the player. Positions are recorded upon leaving the immobility zone.
+    /// </summary>
+    private Vector3 lastRecordedPlayerPosition;
```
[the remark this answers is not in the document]
Maxonovien: I agree with you but I have a faint memory of having a reason to avoid it. But this does not seem necessary from looking at the code, so maybe I'm mistaken. I'll take a look.

src/general/ShuffleBag.cs
```diff
@@ -0,0 +1,212 @@
...
+public class ShuffleBag<T> : IEnumerable<T>
+{
+    private Random random;
+    private List<T> initialContent;
+    private List<T> currentContent;
+
+    public ShuffleBag(Random random)
+    {
+        initialContent = new List<T>();
+        currentContent = new List<T>();
```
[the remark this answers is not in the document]
Maxonovien: There might be some caveat regarding the shuffle algorithm though, which I think is the main performance bottleneck in this class.

There may also be things regarding memory size, but it depends on the size of the objects put into the bag.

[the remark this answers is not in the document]
hhyyrylainen: I don't see any operations in this class that would really benefit from a linked list.
The shuffle operation seems to use a straightforward algorithm (which is probably good enough for our randomness needs), which relies on easily accessing individual elements, so changing to a non-continuous memory stored range would be a big mistake.
Also as these lists will only store references, the storage will be just enough to hold Capacity number of object references.

hhyyrylainen (approved):
I'm happy with the changes that are here, *but* this does *not* actually solve https://github.com/Revolutionary-Games/Thrive/issues/1560 so that remains an issue, and for the future there's a problem as each spawner is not fully tapped before potentially moving onto the next one.
I'm fine with merging this if you are fine with my changes.

sirdanforth:
Here's a list I made to keep track of what to specifically test and what I found on each of them.

1. Check to see if spawners still work correctly = The spawners still work as intended and spawn entities normally.

2. Check to see if spawners spawn compounds and entities randomly but in diverse patterns = It seems a random but diverse pattern pretty much happens in the Hydrothermal Vent, Ocean Floor, Mid Ocean, Ocean Surface, and the Underwater Cave. However when I went to test the Polar Sea the compounds did not seem to spawn in diverse patterns. Only iron would spawn and very very rarely would I see any other compound. However, the Polar Sea is iron rich but I would still expect to see other compounds at least a bit more frequently.

3. Make sure no belt spawning occurs = I'm not entirely sure if this is still a problem or not because I did find that compounds would still be in a row like this and sometimes even in a circle around the player. It wasn't quite as noticeable but I thought I would still include it just in case. 



4. Make sure that there should be at max 15 entities spawned at once (especially when player is stationary) = I did not find any problems pertaining to this. I mark this as a pass.

5. Check to see if spawner spawns entities uniformly around player when stationary = No problems found and works as stated.

6. Check to see if PR #1550 has been fixed = Seems like this issue is fixed and is working correctly.

7. Check to see if PR #1541 has been fixed = Issue has been fixed, iron now spawns in caves.

8. Check to see if PR #1626 has been fixed = Issue has been fixed, iron now spawn and more specifically in the Polar Sea. However, it could be that it spawns iron too frequently now. 

9. Check to see if PR #1777 has been fixed = Issue has been fixed, hydrogen sulfide now spawns in caves.

10. Saving and loading does not break PR branch and works as intended.

11. Loading saves made in a previous version works correctly.

Let me know if I missed anything.

hhyyrylainen:
That's a very comprehensive list. Sorry I was in a hurry and merged this before you were done. I'll link that as a good example of very thorough testing on our wiki on the testing page. https://wiki.revolutionarygamesstudio.com/wiki/Testing
Btw, you can edit the wiki pages if you see something missing or a typo or something like that by logging in with a dev forum account.

Maxonovien:
That's a great testing review, thanks! If I get it well you raise caveats for 2. and 3.

For 3., you could indeed call it a "belt spawning" case, but you can't really avoid it because we make things spawn around the player (approximately on the border ), as we do not want things to close nor far away from the player. The precedent issue was that belt spawning would make contiguous spawning from an edge of the screen to the other, so my way to fix it was to just limit the amount of entity. Such patterns are thus fully part of the algorithm, which I don't think is much of an issue, or at least not one the PR intended to fix.

For 2., it is not a surprise that iron is over-represented. But are other compounds not spawned enough? In other words, do you encounter significantly less of them when compared to a patch with similar values for other compounds? Normally, the spawns should be independant for each compound over the long run (you may have some more local interactions, though).

sirdanforth:
> For 2., it is not a surprise that iron is over-represented. But are other compounds not spawned enough? In other words, do you encounter significantly less of them when compared to a patch with similar values for other compounds? Normally, the spawns should be independant for each compound over the long run (you may have some more local interactions, though).

So yeah the other compounds do spawn but very rarely it seems. It took me probably around 5 minutes before I finally found a patch of phosphate. All I was seeing for 5 minutes was only iron clouds, iron particles, and ice shards. Eventually, I did find more patches of compounds and it was a bit more frequent after I found the phosphate cloud but when I saved and loaded it back it was back to rarely finding any other compounds (but maybe this time it didn't take as long).

PR #2399, opened 2021-07-11
Added equality checks to ActiveMenu

FuJa0815:

This PR stops the user from repeadatly pressing F1 to randomize the easter egg message.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/general/PauseMenu.cs
+++ b/src/general/PauseMenu.cs
@@ -49,58 +49,51 @@ public class PauseMenu : ControlWithInput
     [Signal]
     public delegate void MakeSave(string name);
 
+    private enum MenuEnum
+    {
+        Primary,
+        Help,
+        Load,
+        Options,
+        Save,
+        None,
+    }
+
     /// <summary>
     ///   The GameProperties object holding settings and state for the current game session.
     /// </summary>
     public GameProperties GameProperties { get; set; }
 
-    private string ActiveMenu
+    private MenuEnum ActiveMenu
     {
         get
         {
-            if (primaryMenu.Visible)
-                return "primary";
-            if (helpScreen.Visible)
-                return "help";
-            if (loadMenu.Visible)
-                return "load";
-            if (optionsMenu.Visible)
-                return "options";
-            if (saveMenu.Visible)
-                return "save";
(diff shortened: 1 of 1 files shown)


Kasterisk:
Personally I kind of like being able to randomize the easter eggs on demand but that would defeat their purpose I guess :P.

Anyways a better way for this is to have an enum specifying the menus similarly to how the tabs in MicrobeEditorGUI are handled. So SetActiveMenu method could be left as is and the enum could be used to check what's the current opened menu. See [MicrobeEditorGUI.cs#L1474](https://github.com/Revolutionary-Games/Thrive/blob/master/src/microbe_stage/editor/MicrobeEditorGUI.cs#L1474) and [MicrobeEditorGUI.cs#L409](https://github.com/Revolutionary-Games/Thrive/blob/master/src/microbe_stage/editor/MicrobeEditorGUI.cs#L409)

FuJa0815:
I was thinking about this too. I was wondering why this was implemented with strings and not enums. I'll change this to an enum-based approach.

Kasterisk:
The PauseMenu class predates the enum based tab/page handling in the codebase so it just needs some a bit of refactoring.

src/general/PauseMenu.cs
```diff
@@ -49,11 +49,55 @@ public class PauseMenu : ControlWithInput
     [Signal]
     public delegate void MakeSave(string name);
 
+    private enum MenuEnum
```
hhyyrylainen, line 52:
```suggestion
    private enum ActiveMenuType
```
I don't like using Enum as a suffix...

src/general/PauseMenu.cs
```diff
@@ -49,11 +49,55 @@ public class PauseMenu : ControlWithInput
...
+    }
+
     /// <summary>
     ///   The GameProperties object holding settings and state for the current game session.
     /// </summary>
     public GameProperties GameProperties { get; set; }
 
+    private MenuEnum ActiveMenu
+    {
+        get
```
hhyyrylainen, line 69: I hope this get isn't called very often...

hhyyrylainen (changes requested):
Other than that one naming thing this looks good.

hhyyrylainen (approved):

PR #2400, opened 2021-07-11
Fixed organelle being marked as moved if placing it at its previous position

### LINKED ISSUE #2375: Moving and placing organelle at the same position marked that organelle as moved
Note that this is only true for organelles that were added/edited in previous editor sessions. What happen was that the move action marked the organelle as moved and that causes the organelle's move cost to be 0 even though the position didn't change and this is exploitable in a scenario where the mutation points is depleted and the player is able to do said exploit for a free organelle move.

https://user-images.githubusercontent.com/54026083/122966512-9e195880-d3b3-11eb-90f9-05a8fd658dae.mp4

Kasterisk:

Added a check to `numberOfTimesMoved` in move action to only increment/decrement if the organelle position is changed.

**Related Issues**

Fixes #2375 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


FuJa0815 (approved):

Maxonovien:
While this fixes the exploit in the issue, doesn't this leave the no-move action as an actual action that can be 'undone' by the undo button of the editor (effectively doing nothing, but requiring the player to click once more to undo a previous action)?

Wouldn't it be cleaner to finish the action but discard it in the end?

Kasterisk:
Indeed, this registers a byproduct action that allows undo/redo. I guess it makes more sense to cancel the move action altogether if the final move destination is not changed.

hhyyrylainen (approved):
Code looks fine, seeing as you reported the original issue and this has other approvals, I'll accept that this fixes the bug without spending time testing this myself.

PR #2401, opened 2021-07-11
Repurpose compound bbcode attribute to specify the internal name

Kasterisk:

This changes `text` attribute for compound bbcode to `type` to be used to specify the compound's internal name where it was previously placed between the tags, instead the compound's actual name should now takes its place. For example `[thrive:compound type="glucose"]Glucose[/thrive:compound]` syntax will output a `Glucose` text turned into the compound name format based on the given `glucose` type (this should always be specified, otherwise the formatting fails as it can't find a suitable compound type).

**Related Issues**

To workaround the warnings in Weblate and occasional mistranslation of compound's internal name.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/gui_common/CustomRichTextLabel.cs
```diff
@@ -229,28 +229,27 @@ private string BuildTemplateForTag(string input, ThriveBbCode bbcode, List<strin
...
                     if (!value.StartsAndEndsWith("\""))
                         break;
 
-                    name = value.Substring(1, value.Length - 2);
+                    internalName = value.Substring(1, value.Length - 2);
+                }
+
+                if (!SimulationParameters.Instance.DoesCompoundExist(internalName))
+                {
+                    GD.Print($"Compound: \"{internalName}\" doesn't exist, referenced in bbcode");
```
FuJa0815, line 246: I think you should still use the old algorithm here as a fallback to not break any current translations and to have it as a fallback

FuJa0815 (changes requested):

src/gui_common/CustomRichTextLabel.cs
Kasterisk: Added.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -225,32 +225,43 @@ private string BuildTemplateForTag(string input, ThriveBbCode bbcode, List<strin
...
                 {
                     if (!value.StartsAndEndsWith("\""))
                         break;
 
-                    name = value.Substring(1, value.Length - 2);
+                    internalName = value.Substring(1, value.Length - 2);
                 }
 
-                output = $"[b]{name}[/b] [font=res://src/gui_common/fonts/" +
+                if (!string.IsNullOrEmpty(internalName) && !simulationParameters.DoesCompoundExist(internalName))
```
FuJa0815, line 247: `!simulationParameters.DoesCompoundExist(internalName)` should be enough. A compund with an empty sting should not exist and there is no way that internalName is null

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -225,32 +225,43 @@ private string BuildTemplateForTag(string input, ThriveBbCode bbcode, List<strin
...
+                {
+                    GD.Print($"Compound: \"{internalName}\" doesn't exist, referenced in bbcode");
+                    break;
+                }
+
+                // If type attribute is not specified but input is valid internal name, fallback to the old method
+                if (string.IsNullOrEmpty(internalName) && simulationParameters.DoesCompoundExist(input))
```
FuJa0815, line 254: `if (internalName == string.empty && ...`

FuJa0815 (dismissed):

src/gui_common/CustomRichTextLabel.cs
Kasterisk: I think it could be left as is. `string.IsNullOrEmpty` is already used almost everywhere in the codebase and I don't see real benefit of changing it in favor of direct string comparison.

src/gui_common/CustomRichTextLabel.cs
hhyyrylainen: We need to get the translations updated at some point...
Either need to mass flag them as fuzzy or manually edit each language.

locale/en.po
```diff
@@ -2093,10 +2095,10 @@ msgstr ""
 #: ../src/gui_common/tooltip/ToolTipManager.tscn:1716
 msgid "TOXIN_VACUOLE_PROCESSES_DESCRIPTION"
 msgstr ""
-"Turns [thrive:compound]atp[/thrive:compound] into [thrive:"
-"compound]oxytoxy[/thrive:compound]. Rate scales with concentration of "
-"[thrive:compound]oxygen[/thrive:compound]. Can release toxins by pressing "
-"[thrive:input]g_fire_toxin[/thrive:input]."
+"Turns [thrive:compound type=\"atp\"]ATP[/thrive:compound] into [thrive:"
+"compound type=\"oxytoxy\"]OxyToxy NT[/thrive:compound]. Rate scales with "
```
hhyyrylainen, line 2099: I think with the type already specified, the text inside the bbcode tags should be left empty *unless* it requires to be customized from the normal organelle name.

hhyyrylainen (changes requested):

hhyyrylainen:
Once this is done, it would be nice to explain this on the wiki: https://wiki.revolutionarygamesstudio.com/wiki/Translation_Guidelines

Kasterisk:
Yeah, I've thought about that as well. I could try adding it.

locale/en.po
Kasterisk: So I'm assuming it'd look like this `[thrive:compound type="atp"][/thrive:compound]`, would that be valid and not cause warnings in Weblate?

hhyyrylainen: I think it should be fine, that's valid bbcode syntax with nothing inside the tag.
The current warning is about the tag definition changing between the translation and source language.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -225,32 +225,52 @@ private string BuildTemplateForTag(string input, ThriveBbCode bbcode, List<strin
...
+                if (pairs.TryGetValue("type", out string value))
                 {
                     if (!value.StartsAndEndsWith("\""))
                         break;
 
-                    name = value.Substring(1, value.Length - 2);
+                    internalName = value.Substring(1, value.Length - 2);
+                }
+
+                if (!string.IsNullOrEmpty(internalName) && !simulationParameters.DoesCompoundExist(internalName))
```
hhyyrylainen, line 247: Wouldn't this if-else be simpler to understand if this was structured like this:
```
if(!string.IsNullOrEmpty(internalName){
   // Explicitly defined 
   setup internalName here
} else if(!string.IsNullOrEmpty(input)){
   internalName = input;
} else {
  error
}
// Then later
if(!simulationParameters.DoesCompoundExist(internalName)
  report error
```
That way the logic I think would be a bit simpler to understand and reporting the compound not existing would be unified.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -225,32 +225,52 @@ private string BuildTemplateForTag(string input, ThriveBbCode bbcode, List<strin
...
-
-                var compound = SimulationParameters.Instance.GetCompound(input);
-
-                var name = compound.Name;
-
                 var pairs = StringUtils.ParseKeyValuePairs(attributes);
 
-                // Parse attributes if there are any
-                if (pairs.TryGetValue("text", out string value))
+                var fallback = false;
```
hhyyrylainen, line 236: Could you add a comment and create an issue that this fallback should be removed, within like 6 months to give some time for translations to be updated?

hhyyrylainen (changes requested):

src/gui_common/CustomRichTextLabel.cs
Kasterisk: If the internal name is already explicitly defined in the type attribute, there will be nothing left to setup in the first if block as the actual processing would be after the compound validation. I think this could be coalesced further while preserving the unified invalid compound reporting into
```
// get internalname

if (!string.IsNullOrEmpty(input) && string.IsNullOrEmpty(internalName))
{
    // Handle fallback
}

if(!simulationParameters.DoesCompoundExist(internalName)
  report error and abort

// final step
```

hhyyrylainen: If that does also result in non-duplicate string emptiness checks, I'm fine with it.

Kasterisk: For the fallback handling at least there is no duplicate string empty checks. In the final step there's one check for empty input handling `[tag][/tag]` and no more than that.

hhyyrylainen:
I gave this a spin with one Finnish translation, and this seems to work just fine:

hhyyrylainen (approved):
I guess this is now fine by me. As sorry for including some random other changes, I'll merge this right away.

PR #2404, opened 2021-07-13
Add option to disable flashing effect on editor button

Kasterisk:
**Brief Description of What This PR Does**

Note that this doesn't disable the flashing of the editor button but the lighting effect used on it to workaround the bug where HUD is sometimes disappearing.

**Related Issues**

Addresses #2022 but doesn't completely fixes it.

**Progress Checklist**

Note: before starting this checklist the issue should be marked as non-draft.

- [x] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [ ] Initial code review passed
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author.

PR #2405, opened 2021-07-13
Fix built-in tooltips on the options menu not showing when paused

Kasterisk:

Noticed this while testing #2404, the built-in Godot tooltips used for the flash effect and cloud resolution divisor setting doesn't seem to show while ingame (pause menu). Turns out Godot's built in tooltips uses a timer which is dependent on the root viewport's Process notification, so when the tree is paused so does the root viewport and tooltip.

There's a workaround to this, one of which is to pause the lower node in hierarchy (such as the root node of a scene) but doing so won't be very clean. Instead I opted with using the already established custom tooltip system which is already independent of the SceneTree's pause state.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/gui_common/tooltip/ToolTipHelper.cs
```diff
@@ -45,4 +45,13 @@ public static DefaultToolTip CreateDefaultToolTip()
 
         callbackData.Add(toolTipCallbackData);
     }
+
+    /// <summary>
+    ///   Registers a Control mouse enter/exit event to display a custom tooltip from the given tooltip and group name.
+    /// </summary>
+    public static void RegisterToolTipForControl(this Control control, string tooltipName, string groupName,
+        List<ToolTipCallbackData> callbackDatas)
```
hhyyrylainen, line 53:
```suggestion
        List<ToolTipCallbackData> callbackData)
```
Data is already plural (AFAIK) so you can't really add "s" to it.

src/general/OptionsMenu.cs
```diff
@@ -351,6 +358,9 @@ public override void _Ready()
         // We're only utilizing the AcceptDialog's auto resize functionality,
         // so hide the default Ok button since it's not needed
         backConfirmationBox.GetOk().Hide();
+
+        cloudResolutionTitle.RegisterToolTipForControl("cloudResolution", "options", toolTipCallbacks);
+        guiLightEffectsToggle.RegisterToolTipForControl("guiLightEffects", "options", toolTipCallbacks);
```
hhyyrylainen, line 363: When the data is not needed, would it be nice if it could be just given as null?

hhyyrylainen (changes requested):
Just a few small comments.

src/general/OptionsMenu.cs
Kasterisk: Could be, assuming the programmer has alternative way for the callbacks handling, otherwise the tooltip wouldn't react to anything, but that could be documented in the code.

hhyyrylainen: Well for these tooltips, doesn't the tooltip manager or whatever it is, handle the showing? So the client code that specifies that this should have a tooltip, shouldn't need to worry itself about allocating a dummy callback data.

Kasterisk: Yes the tooltip manager handles that but it's notified by the callbacks. Now that you mention it, I guess all these callback data can just be stored in the tooltip manager class instead of in the client code.

Kasterisk: Okay, I ended up unifying all the tooltip callback storing into ToolTipHelper, this should make the custom tooltip more robust and less of a hassle to setup.

src/gui_common/tooltip/ToolTipCallbackData.cs
```diff
@@ -3,13 +3,16 @@
 /// <summary>
 ///   Helper class to contain the OnMouseEnter and OnMouseExit callback for the custom tooltips
 /// </summary>
-public class ToolTipCallbackData : Reference
+public class ToolTipCallbackData : Object
```
hhyyrylainen, line 6: Why was this base class change done?

hhyyrylainen (approved):
I think I'm happy with this now. Just need to fix the translations being out of date.

FuJa0815:
To resolve the conflicts:
```bash
git checkout master
git pull
git checkout fix_tooltip_on_pause
git checkout master locale
ruby scripts/update_localization.rb
git add .
git commit -m "updated localizations"
git push
```

FuJa0815 (approved):
Works good ingame

src/gui_common/tooltip/ToolTipCallbackData.cs
Kasterisk: Was changed so a callback data instance could be freed manually on tooltip unregister, but that itself is not very safe which is why the call is set to run on next frame. Revisiting this, inheriting Reference here may actually be more convenient and safer as you could just rely on garbage collection so no need to manually freeing the instance after removing its reference from the `ToolTipCallbacks` list, a bonus perk is that it won't get deleted immediately afterwards.

PR #2413, opened 2021-07-15
Rotate microbes by using physics force

### LINKED ISSUE #2112: Make microbe rotation use physics force
instead of directly just changing the rotation of the physics body.

NOTE: to anyone going to take a look at this, all the axes are currently locked, so you'll want to unlock them before starting on this, otherwise you'll wonder why the angular force doesn't do anything (https://github.com/Revolutionary-Games/Thrive/pull/2266)

I tried to do this years ago but I ended up just having the cells oscillate around their wanted direction instead of the rotation stabilizing, so this is a bit difficult to do.

With this we can make piluses hitting work better. And also we can finally make rotation speed be affected by cell mass, so implementing cilia will make more sense (#174).

Kjell291:

microbes now rotate by using physics force

**Related Issues**

closes #2112 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

I could not test if loading old saves works, everything else seems to work fine


src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,82 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+    ///   Adds torque to reach the targeted rotation
+    /// </summary>
+    private void ApplyPhysicsRotation(PhysicsDirectBodyState physicsState)
+    {
+        var transform = physicsState.Transform;
+        var target = transform.LookingAt(LookAtPoint, new Vector3(0, 1, 0));
+        if (transform == target)
+            return;
+
+        var strength = Constants.CELL_BASE_THRUST + Constants.CELL_BASE_THRUST * (HexCount/2);
```
Maxonovien, line 2123: Wouldn't this be cleaner as `Constants.CELL_BASE_THRUST * (1 + HexCount / 2)`?

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,82 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+        var maxVelocity = 5.0f;
+
+        var distance = target.basis.GetEuler().y - transform.basis.GetEuler().y;
+        distance = (float)(distance * (180/Math.PI));
```
Maxonovien, line 2127: You have `MathUtils.DEGREES_TO_RADIANS = Mathf.Pi / 180`, which you could use here instead

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,82 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+        if (distance > 180) {
+            distance = -360 - distance;
+        } else if (distance < -180) {
+            distance = 360 + distance;
+        }
```
Maxonovien, line 2132: I assumed by default that `distance` was a measure in meters. You might want to go for a more specific name if this deals with angles.

Also, why changing to degrees?

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,82 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+    /// </summary>
+    private void ApplyPhysicsRotation(PhysicsDirectBodyState physicsState)
+    {
+        var transform = physicsState.Transform;
+        var target = transform.LookingAt(LookAtPoint, new Vector3(0, 1, 0));
+        if (transform == target)
+            return;
+
+        var strength = Constants.CELL_BASE_THRUST + Constants.CELL_BASE_THRUST * (HexCount/2);
+        var maxVelocity = 5.0f;
```
Maxonovien, line 2124: This is not used before l.2137, and you have `velocity` defined l.2135, so you might want to just move this definition a few lines down to be along side velocity.

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,82 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+        var velocity = physicsState.AngularVelocity.y;
+
+        if (velocity < -maxVelocity) {
+            force = 1;
+        } else if (velocity > maxVelocity) {
+            force = -1;
+        } else {
+            float velocityTo90 = 0.0f;
+            if (velocity != 0)
+                velocityTo90 = ((velocity)/maxVelocity) * 90;
```
Maxonovien, line 2144: Why not simply use `float velocityTo90 = velocity / maxVelocity * 90.0f;` (or pi/4 if you go in rad)?

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,82 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+            
+            if (velocityTo90 < 0.1f && velocityTo90 > 0.0f || velocityTo90 > -0.1f && velocityTo90 < 0.0f ){
```
Maxonovien, line 2146: Couldn't this be simplified as checking if velocityTo90 is [-0.1f, 0.1f]?

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,82 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+                if (distance > 2.0f) {
+                    force = 1;
+                }
+                else if (distance < -2.0f) {
+                    force = -1;
```
Maxonovien, line 2151: What does 2.0f stand for here?

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,82 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+                }
+            } else if (distance >= 90 && velocity < maxVelocity * 0.8f) {
+                force = 1;
+            }
+            else if (distance <= -90 && velocity > -maxVelocity * 0.8f) {
+                force = -1;
```
Maxonovien, line 2157: Same here, why 80% of max velocity?

Kjell291:
I will fix everything in a few hours or tomorrow.

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,82 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+                }              
+            } else if (velocityTo90 < 0) {
+                if (0 > distance && distance > -90 && velocityTo90/distance > 0.7f) {
+                    force = (velocityTo90/distance) * 1.5f;
+                }
+                else if (-2.0f > distance && distance > -90 && velocityTo90/distance < 0.7f) {
+                    force = -(distance/velocityTo90);
+                }
+                else if (2.0f < distance && distance < 90) {
+                    force = -90/velocityTo90;
```
Maxonovien, line 2175: I really don't get what motivate this whole bunch of computations. Are these known formulae? Is this a discretization of computations? By the way, why not using more continuous computations?

A few comments would be very welcome to explain things at this point.

Maxonovien (commented):
Sorry, I should have done a full review from the start. 

Overall, the main issue is that I don't understand where your computations come from, so comments explaining your trail of thought would be very useful.

The rest are minor things to help overall readability and consistency of the code with the rest of the project.

src/microbe_stage/Microbe.cs
Kjell291: What other name for distance would you suggest? I cant think of a good one.
I changed it to degrees, because it was easier for me to work with. But I can change it back if you want.

src/microbe_stage/Microbe.cs
Kjell291: I just thought about what cases would be possibly and what would be the best way to handle them. After that I experimented with the numbers to make the gameplay as smooth as possible.

Kjell291:
I added some comments to explain what I did. Just let me know if I should change anything else.

src/microbe_stage/Microbe.cs
Maxonovien: Well, I think I'd go for `angleToTarget`? But this is just my mind as a reviewer, you should make sure to pick a name you find clear as the very person who coded this PR. Take your time to come up with a name your really like.

Maxonovien: As for degrees, the style doc does not mention this, so I guess I can't really tell for sure you should do this. I just know that, as far as I recall, the rest of the code uses radians, which are furthermore scientific system units. So I think you should use radians indeed.

Kasterisk:
Playtested for a bit, in bigger microbes there seems to be some kind of inertia happening where the cell rotation wobbles by some amount before stabilizing to the target direction, perhaps this could be tweaked in such a way so it snaps instead?

src/microbe_stage/Microbe.cs
Maxonovien: Thrive is generally trying to be as scientifically accurate with computations as possible, so it would be better to have a proper model of what happens. I think you should have a few lines describing the general behavior of your rotation process model, which is not completely obvious from what I read.

Maxonovien: Also, for number you played with and whose values you tweaked, these should be defined in `Constants.cs`.

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,137 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+                // Brake if rotating to fast
+                if (distance > 0.0f && distance < 90.0f && velocityTo90 / distance > 0.7f)
+                {
+                    force = -(velocityTo90 / distance) * 1.5f;
+                }
+
+                // Accelerate if rotating to slow
+                else if (distance > 1.5f && distance < 90.0f && velocityTo90 / distance < 0.7f)
+                {
+                    force = distance / velocityTo90;
```
Maxonovien, line 2200: I'm sorry, but I am very disturbed by the fact that a force is a speed over a distance, or a distance over a speed here. Perhaps in the first case 1.5f actually has units, but then it should appear somehow (e.g. by the constant's name, as it should also be defined as one).

src/microbe_stage/Microbe.cs
```diff
@@ -1352,7 +1352,9 @@ public override void _IntegrateForces(PhysicsDirectBodyState physicsState)
     {
         // TODO: should movement also be applied here?
 
-        physicsState.Transform = GetNewPhysicsRotation(physicsState.Transform);
+        // physicsState.Transform = GetNewPhysicsRotation(physicsState.Transform);
```
Maxonovien, line 1355: If this isn't code to be definitely remove, it would be suited to tell why it is just commented.

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,137 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+        else if (velocity > maxVelocity)
+        {
+            force = -1.0f;
+        }
+        else
+        {
+            // Velocity in a range from -90(-maxVelocity) to 90(maxVelocity)
+            var velocityTo90 = 0.0f;
+            if (velocity != 0.0f)
+                velocityTo90 = velocity / maxVelocity * 90.0f;
```
Maxonovien, line 2161: I still don't see any reason for this not to be in a single line formula?

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,137 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+
+        // Brake if velocity exceeds maxVelocity
+        if (velocity < -maxVelocity)
+        {
+            force = 1.0f;
+        }
+        else if (velocity > maxVelocity)
+        {
+            force = -1.0f;
+        }
```
Maxonovien, line 2155: Here you should be able to make it much smaller using Mathf.Sign() to get the sign of your numbers, as well as Mathf.Abs()

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2110,137 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+        var distance = target.basis.GetEuler().y - transform.basis.GetEuler().y;
+        distance = (float)(distance * (180.0f / Math.PI));
+        if (distance > 180.0f)
+        {
+            distance = -360.0f + distance;
+        }
+        else if (distance < -180.0f)
+        {
+            distance = 360.0f + distance;
+        }
```
Maxonovien, line 2137: Working in absolute values might be clearer (see below).

Maxonovien (commented):
Even though the algorithm is made *ad hoc* I think it can be effectively adapted. But there still are many clarity issues. I think the code is made overcomplicated due to having to deal with signs: most conditions could be collapsed with proper sign handling. An even more efficient could perhaps be to use absolute values all along and only apply sign at the end?

Given that it has many if/else if/... structures, I think this would also largely benefit from force computations being done inside a function and using early returns.

I also repeat here for visibility that *ad hoc* formulae should be marked somehow and *ad hoc* values should be defined as constants, ideally in the dedicated `Constants.cs` file, and that radians should be prefered to angles.

src/microbe_stage/Microbe.cs
hhyyrylainen: Degrees vs radians doesn't really matter as it is just one float multiplication away. Though if you go radians -> degrees -> radians, I think that's bad and you shouldn't add that intermediate conversion instead.
In general I want angles in the game to user quaternions, because euler angles (even if they are radians), have gimbal lock problems. So any serious work should use quaternions. I know that due to microbes being locked to an axis gimbal lock probably can't happen here, but I'd still like to see the `basis` being used in computations *without* stripping away information by converting it to euler angles.

Maxonovien:
Additional remarks on the computation: all this would be much more simpler by using some formula, linking angular momentum to the distance left to rotate.

Maxonovien:
So, at the moment this is still an imbroglio of if/if else, which would be much more readable as a function with early return, but I guess this is still in progress.

I write this comment because I found, while taking a look at the development forum, [this thread](https://forum.revolutionarygamesstudio.com/t/cellular-movement/460/14) which is of interest to this PR, as it comes with several ways to simulate microbe rotation.

Kjell291:
I managed to improve the calculation of maxVelocity and many of the if/else statements are now unnecessary.
The code is now much shorter and cleaner. I kept degrees for now, but can change it. 
I dont think working with quaternions has any advantage in this case, becuase both the x and z axis are locked.

hhyyrylainen:
I suggested quaternions as then this algorithm could be generalized to any kind of physics rotation to a target orientation in the game.

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2108,75 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+            angleToTarget = 360.0f + angleToTarget;
+        }
+
+        // Maximum desired angular velocity
+        // If the value is to high the microbe will not brake in time
+        // If it is to low the microbe is rotating slower than it could
+        var maxVelocity = 1.0f + (strength * 0.3f) / (HexCount * 40.0f);
+
+        // Angular velocity of the microbe in a range from -90(-maxVelocity) to 90(maxVelocity)
+        var velocityTo90 = physicsState.AngularVelocity.y / maxVelocity * 90.0f;
```
Maxonovien, line 2143: I have trouble understanding what is going on here. What are, in the end, the units of the velocity to 90? 
The name makes me think of degrees/s, but then the formula seems more of a proportion (unitless) times degrees...

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2108,75 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+
+    private float RotationalForce(float velocityTo90, float angleToTarget)
+    {
+        // Do nothing if velocity is nearly zero and rotation is nearly correct
+        if (Math.Abs(velocityTo90) < 0.1f && Math.Abs(angleToTarget) < 1.0f)
+        {
+            return 0.0f;
+        }
+
+        // Accelerate if rotating to slow
```
Maxonovien, line 2170:
```suggestion
        // Accelerate if rotating too slow
```

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2108,75 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+        if (velocityTo90 / angleToTarget < 1.0f)
+        {
+            return Mathf.Sign(angleToTarget) * 1.0f;
+        }
+
+        // Brake if rotating to fast
```
Maxonovien, line 2176:
```suggestion
        // Brake if rotating too fast
```

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2108,75 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+        var torque = new Vector3(0, strength * RotationalForce(velocityTo90, angleToTarget) * physicsState.Step, 0);
+        physicsState.AddTorque(torque);
+    }
+
+    private float BaseRotationalStrength()
+    {
+        // More strength is required to rotate larger microbes
+        // Value should be lowered if cilia is implemented
+        // Uses CELL_BASE_THRUST for now, the same as for movement
+        var strength = Constants.CELL_BASE_THRUST * (1.0f + HexCount);
```
Maxonovien, line 2155:
```suggestion
        var strength = Constants.CELL_BASE_THRUST * (1 + HexCount);
```

Hex count is an integer, it makes little sense to add it to a float.

src/microbe_stage/Microbe.cs
```diff
@@ -2108,6 +2108,75 @@ private Transform GetNewPhysicsRotation(Transform transform)
...
+        // Do nothing if velocity is nearly zero and rotation is nearly correct
+        if (Math.Abs(velocityTo90) < 0.1f && Math.Abs(angleToTarget) < 1.0f)
+        {
+            return 0.0f;
+        }
+
+        // Accelerate if rotating to slow
+        if (velocityTo90 / angleToTarget < 1.0f)
+        {
+            return Mathf.Sign(angleToTarget) * 1.0f;
```
Maxonovien, line 2173: Is there a reason for the force to be 1, 0 or -1? This seems more of a directional thing here (ternary choice between increase, decrease and do nothing). If yes, you might want to rename the function.
Else, it would be better to have this base force of 1.0f as a constant in the dedicated file.

Maxonovien (changes requested):
It is easier to understand hows things are running globally, but there are still some unclear formulae.  Small style remark: brackets are not mandatory in one line bodies in if statements, and not having them alleviates reading.

Kjell291:
Just to let you know, I dont have much time at the moment. I will try to work on this next week, but maybe it takes two weeks or more.

Nunez2196:
> Just to let you know, I dont have much time at the moment. I will try to work on this next week, but maybe it takes two weeks or more.

I'll be fooling with this branch now, hang tight. I commend you for your contribution thus far

c-mirai:
Apologies if I'm intruding here. I wanted to mention using a [PD controller](https://www.matthewpeterkelly.com/tutorials/pdControl/index.html) or a PID controller to calculate torque. The current algorithm is a sort of clamped P controller that may result in oscillation, slow turning, and/or imprecise turns.

FuJa0815:
What's the state of this?

GodOfKnowledge:
Nice to also see this being worked on, It felt a bit op to spin around, making a few species go extinct.

Kjell291:
I completely forgot about this, I am closing this for now. But I might reopen a new Pull request in the future.

Issue #2276, opened 2021-05-30
ATP stop producing when maximizing the window


84634E1A607A:
The ATP falls sharply when I maximize the Thrive window. This problem can be reproduced in earlier releases as well. 

Also when I experience a "卡顿" (Sorry I really don't know how to express that in English, during which the framerate is unexpectedly lower than usual, such as when the cell is engulfing and organelles reproducing) the ATP will suddenly fall low (lower than it should be? I know that after organelles reproducing the maximum ATP is higher).

hhyyrylainen:
I was sure that there was already an issue for this, but I couldn't find this. I've known about this for a while.

If the game experiences a lag spike (or the framerate is too low): the ATP production slows down because the compound storage is not increased, meaning that when more ATP is produced during a longer frame simulation ATP can't be produced at the rate it should be. And then on top of that the ATP usage is correctly computed for a longer frame, so unless you have really large storage capacity you'll run out of ATP on that frame because ATP could not be produced the right amount.

I think maybe this will explain things better:
- The game experiences a lag spike, the simulation time between frames is 300ms
- ATP generation tries to run as usual, but is multiplied by the elapsed time, which is usually fine, but now multiplied by 300 the resulting compound amount doesn't fit in the cell, so the cell doesn't get that compound amount
- Now ATP usage runs, and is also multiplied by 300 to take the passed time into account
- The consumption is correctly adjusted to be really huge, but as the cell doesn't have enough storage for that amount of ATP, the cell briefly takes ATP damage as the consumption depletes the ATP of the cell

84634E1A607A:
What if we handle the usage first?
Then we will have more space to fit in.

hhyyrylainen:
The effect would be the same: you take damage and / or can't move at normal speed as you don't have ATP to power movement. 

A slightly better way would be to have the consumption be computed first, processes were ran then, and finally ATP damage would be given (if cell doesn't have any ATP), though this is hard to fit in the current design because the process system is ran on background threads (in order to speed up processing many cells on multicore CPUs). So it's not possible to fit the Microbe class doing some stuff, then processes run, and then doing stuff again.

hhyyrylainen:
One way to limit the consumption per frame, would be to assume that the game runs at least at 15 FPS or so, and cap the maximum elapsed time for the process system and Microbe computations.

hhyyrylainen:
[...]

PR #2418, opened 2021-07-16
Fixed process panel errors with all processes

### LINKED ISSUE #2351: Flooded "Invalid new child position" in process panel with all organelles
**Describe the bug**

```
E 0:01:13.464   move_child: Invalid new child position: 9.
  <C++ Error>   Index p_pos = 9 is out of bounds (data.children.size() + 1 = 8).
  <C++ Source>  scene/main/node.cpp:330 @ move_child()
  <Stack Trace> :0 @ void Godot.NativeCalls.godot_icall_2_583(IntPtr , IntPtr , IntPtr , Int32 )()
                Node.cs:811 @ void Godot.Node.MoveChild(Godot.Node , Int32 )()
                ChildObjectCache.cs:94 @ void ChildObjectCache`2[[IProcessDisplayInfo, Thrive, Version=0.5.4.0, Culture=neutral, PublicKeyToken=null],[ChemicalEquation, Thrive, Version=0.5.4.0, Culture=neutral, PublicKeyToken=null]].ApplyOrder()()
                ProcessList.cs:54 @ void ProcessList._Process(Single )()
```

**To Reproduce**
Currently if you add all organelles (nucleusless) one each and open Process List panel you'll get a flooded exception above.

**Additional Info**
image
image

I'm investigating into this. Currently I can tell that this happens each frame when _Process() gets called.

hhyyrylainen:

now all duplicate cell processes are merged into a single TweakedProcess for processing. Note that after this is merged this improvement is pretty important: https://github.com/Revolutionary-Games/Thrive/issues/2419 (though I'm open to feedback regarding a different approach to solving the process panel issue).

**Related Issues**

closes #2351

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/microbe_stage/Microbe.cs
+++ b/src/microbe_stage/Microbe.cs
@@ -2148,16 +2148,44 @@ public class Microbe : RigidBody, ISpawned, IProcessable, IMicrobeAI, ISaveLoade
     /// </summary>
     private void RefreshProcesses()
     {
-        processes = new List<TweakedProcess>();
-        processesDirty = false;
+        if (processes == null)
+        {
+            processes = new List<TweakedProcess>();
+        }
+        else
+        {
+            processes.Clear();
+        }
 
         if (organelles == null)
             return;
 
         foreach (var entry in organelles.Organelles)
         {
-            processes.AddRange(entry.Definition.RunnableProcesses);
+            // Duplicate processes need to be combined into a single TweakedProcess
+            foreach (var process in entry.Definition.RunnableProcesses)
+            {
+                bool found = false;
+
+                foreach (var existing in processes)
+                {
+                    if (existing.Process == process.Process)
+                    {
+                        existing.Rate += process.Rate;
+                        found = true;
+                        break;
+                    }
+                }
+
+                if (!found)
+                {
+                    // Because we modify the process, we must duplicate the object for each microbe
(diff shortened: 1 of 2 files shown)


hhyyrylainen:
@84634E1A607A could you test it as you reported the original issue.

84634E1A607A:
I personally think that the Cell Process system is a bit messy. We already have the IProcessInfo interface and several classes that implement this interface, and now it is more complicated to add the tweaked class. And we seem to need to traverse all the organelles of the cell several times in each frame. In order to adapt to changing the priority of the cell process, I think it is better to create a new table (CellProcess -->  the number of organelles of the process; or CellProcess --> full speed), and then maintain this table in the AddChild() and RemoveChild() of the cell, the follow-up operations directly obtains information from the table in the order of the process. Will it be better than merging when to be displayed?

The current cell processes are sorted according to the organelles. If the conditions cannot be met when traversing to a certain organelle, the organelle will not work: I think it would be much better to combine all the organelles as is said. It can also be adjusted by integrating the input and output of each cell process, and by the way solve the problem #2276.

hhyyrylainen:
> And we seem to need to traverse all the organelles of the cell several times in each frame

Where is that?
The reason why I was comfortable coding things this way is that the process list is *only* recreated when it is marked dirty. It is marked dirty when the organelles of a cell change.

PR #2422, opened 2021-07-16
Added support for basic load after dropping a save file on the game

### LINKED ISSUE #1427: Use files dropped signal to import a save
files dropped signal to load a save / creature in thrive https://docs.godotengine.org/en/stable/classes/class_scenetree.html#signals

makes it easier to load up a save someone gave you, for save sharing.

hhyyrylainen:
Allows loading a save file by dragging it from a folder onto the game.

Follow up issues to polish this up
- https://github.com/Revolutionary-Games/Thrive/issues/2424
- https://github.com/Revolutionary-Games/Thrive/issues/2423

**Related Issues**

closes #1427 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/engine/FileDropHandler.cs
+++ b/src/engine/FileDropHandler.cs
@@ -8,7 +8,7 @@ public class FileDropHandler : Node
 {
     public override void _Ready()
     {
-        GetTree().Connect("files_dropped", this, "OnFilesDropped");
+        GetTree().Connect("files_dropped", this, nameof(OnFilesDropped));
     }
 
     private void OnFilesDropped(string[] files, int screen)
@@ -64,7 +64,7 @@ public class FileDropHandler : Node
                 return;
             }
 
-            if (info.ThriveVersion == Constants.Version)
+            if (info.ThriveVersion != Constants.Version)
             {
                 // TODO: would be nice to show a confirmation dialog here
                 GD.Print("The dropped save version is not exactly the same as current game version");
--- a/src/engine/FileDropHandler.cs
+++ b/src/engine/FileDropHandler.cs
@@ -1,0 +1,80 @@
+﻿using System;
+using Godot;
+
+/// <summary>
+///   Handles signal from Godot when files are dragged and dropped onto the game window
+/// </summary>
+public class FileDropHandler : Node
+{
+    public override void _Ready()
+    {
+        GetTree().Connect("files_dropped", this, "OnFilesDropped");
+    }
+
+    private void OnFilesDropped(string[] files, int screen)
+    {
+        foreach (var file in files)
+        {
+            GD.Print("Detected file drop \"", file, "\" on screen ", screen);
+            HandleFileDrop(file, screen);
+        }
+    }
+
+    private void HandleFileDrop(string file, int screen)
+    {
+        // Currently nothing depends on this (this might be the game window index, at least on Linux this doesn't seem
+        // to change when the monitor is changed the drag happens on...)
+        _ = screen;
+
+        // For now just the load save functionality is done, but in the future this might be extended to allow
+        // other code to dynamically register listeners here
+
+        if (file.EndsWith(Constants.SAVE_EXTENSION, StringComparison.InvariantCulture))
+        {
+            if (!file.StartsWith(Constants.EXPLICIT_PATH_PREFIX, StringComparison.InvariantCulture))
+                file = Constants.EXPLICIT_PATH_PREFIX + file;
+
+            // TODO: could add a dialog to ask if the save should be loaded now or copied to the saves folder
+            HandleLoadSaveFromDrop(file);
+        }
+        else
+        {
+            GD.Print("Unknown file type to handle on drop");
+        }
+    }
+
+    private void HandleLoadSaveFromDrop(string file)
+    {
+        GD.Print("Trying to load dropped save file...");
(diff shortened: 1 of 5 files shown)


src/engine/FileDropHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+public class FileDropHandler : Node
+{
+    public override void _Ready()
+    {
+        GetTree().Connect("files_dropped", this, "OnFilesDropped");
+    }
+
+    private void OnFilesDropped(string[] files, int screen)
+    {
+        foreach (var file in files)
```
adQuid, line 16: Do we want to attempt to load multiple save files at a time? Is there any use case for supporting a multi-drop at this stage?

hhyyrylainen: I imagined that things like dropping individual creations (if we add export from the editor functionality) would be nice to load at the same time. And if https://github.com/Revolutionary-Games/Thrive/issues/2424 is done then it makes sense for saves. Though currently this crashes, but I think once https://github.com/Revolutionary-Games/Thrive/pull/2421 is merged this wouldn't crash anymore.

adQuid: Fair enough. I'm against this kind of future-proofing multiple PRs ahead, but that's a tactical choice you are entitled to make.

hhyyrylainen: I mean it would require *more* code to explicitly stop after the first save load has happened

adQuid: Nonsense: `HandleFileDrop(files[0], screen);`

hhyyrylainen: There needs to be separate logic to allow the save loading to act differently than say other files dropped. So that would break the overall design (for example adding another file type like an exported creature, wouldn't be good to just pass the first file to it). Only thing that would work is if a drop handler method could return a bool indicating that further files shouldn't be handled, or something (but with mixed file types dropped that wouldn't be the right thing).

adQuid: Well again, if you're future-proofing then taking my suggestion would be a bad idea. I was just saying that my tactic would have been to write a method that only works with what the code is expected to do today, then refactor to support more functionality when it is implemented.

hhyyrylainen: I always over futureproof my code. I don't have time right now but if I did, I would implement the extra features on top right now to end this discussion.

src/engine/FileDropHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+                return;
+            }
+
+            if (SaveHelper.IsKnownIncompatible(info.ThriveVersion))
+            {
+                GD.Print("Dropped save file is known incompatible, not loading it");
+                return;
+            }
+
+            if (info.ThriveVersion == Constants.Version)
```
Kasterisk, line 67: Looking at the message in the debug print, shouldn't this check for inequality instead?

src/engine/FileDropHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+    private void HandleLoadSaveFromDrop(string file)
+    {
+        GD.Print("Trying to load dropped save file...");
+
+        // TODO: would be nice to have some kind of popup box showing the errors from here to the user
+        try
+        {
+            var info = Save.LoadJustInfoFromSave(file);
+
+            if (info.Type == SaveInformation.SaveType.Invalid)
```
Kasterisk, line 55: I attempted to drop corrupted save file but seems like this doesn't get triggered, debu log only shows exception from `LoadJustInfoFromSave`. Not very critical issue though, just thought worth mentioning.

Kasterisk (changes requested):
Works great. Some comments:

src/engine/FileDropHandler.cs
hhyyrylainen: Yes, good catch.

src/engine/FileDropHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+using Godot;
+
+/// <summary>
+///   Handles signal from Godot when files are dragged and dropped onto the game window
+/// </summary>
+public class FileDropHandler : Node
+{
+    public override void _Ready()
+    {
+        GetTree().Connect("files_dropped", this, "OnFilesDropped");
```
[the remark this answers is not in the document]
hhyyrylainen: Whoops, forgot that existed while quickly coding up this feature. I'll switch to that.

src/engine/FileDropHandler.cs
hhyyrylainen: In my testing I found that at one point I got an invalid save type. I think LoadJustInfoFromSave is made to not throw an exception, and instead just returns an invalid save info object, or I may have used a different save load method before that does return invalid save info...

Kasterisk: Just checked again and yeah I must have mistaken `PrintErr` in `LoadJustInfoFromSave`'s catch block for an exception. Actually now after testing this again this seems to work correctly, I may have confused the debug log messages at some point...

hhyyrylainen:
I'll leave further improvements to those issues I linked as follow ups in the PR description.

PR #2426, opened 2021-07-17
Made AI move for Run and Rumble if immobile

### LINKED ISSUE #889: AI cells should decide to move immediately
Currently the AI waits about 3 seconds after a cell spawns before it does anything. It would be better to have the cells immediately think and start moving. This is only really visible when respawning.

Maxonovien:

This PR modifies run and rumble to make immobile microbe move randomly (as they don't have any input to compute gradient from).
This makes newly spawned microbes move immediately (as they would previously only have 50% chance to move per step).
Note that this does not affect sessile organisms.

**Related Issues**

Closes #889 .

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/microbe_stage/MicrobeAI.cs
+++ b/src/microbe_stage/MicrobeAI.cs
@@ -367,7 +367,7 @@ public class MicrobeAI
     // For doing run and tumble
     private void RunAndTumble(Random random)
     {
-        // If this microbe is actually immobile, just initialize by moving in a random direction.
+        // If this microbe is currently stationary, just initialize by moving in a random direction.
         // Used to get newly spawned microbes to move.
         if (microbe.MovementDirection.Length() == 0)
         {
--- a/src/microbe_stage/MicrobeAI.cs
+++ b/src/microbe_stage/MicrobeAI.cs
@@ -367,6 +367,14 @@ public class MicrobeAI
     // For doing run and tumble
     private void RunAndTumble(Random random)
     {
+        // If this microbe is actually immobile, just initialize by moving in a random direction.
+        // Used to get newly spawned microbes to move.
+        if (microbe.MovementDirection.Length() == 0)
+        {
+            MoveWithRandomTurn(0, 2 * Mathf.Pi, random);
+            return;
+        }
+
         // Run and tumble
         // A biased random walk, they turn more if they are picking up less compounds.
         // The scientifically accurate algorithm has been flipped to account for the compound
--- a/src/microbe_stage/MicrobeAI.cs
+++ b/src/microbe_stage/MicrobeAI.cs
@@ -371,7 +371,7 @@ public class MicrobeAI
         // Used to get newly spawned microbes to move.
         if (microbe.MovementDirection.Length() == 0)
         {
-            MoveWithRandomTurn(0, 2 * Mathf.Pi, random);
+            MoveWithRandomTurn(0, Mathf.Pi, random);
             return;
         }
 


Maxonovien:
As a note: while initializing microbes to move in any direction makes sense to me, it comes with microbes sometimes heading towards the player and pushing it for a while. There are 3 possible solutions in my mind:
1. Tweaking the possible orientations (only works if orientation on spawn is not fully random; and I don't think it is?)
2. Making microbes avoid the player in general.
3. Simply not caring, the player can move out of the way if wanted.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -367,6 +367,14 @@ private void EngagePrey(Vector3 target, Random random, bool engulf)
     // For doing run and tumble
     private void RunAndTumble(Random random)
     {
+        // If this microbe is actually immobile, just initialize by moving in a random direction.
```
adQuid, line 370:
```suggestion
        // If this microbe is currently stationary, just initialize by moving in a random direction.
```

Being picky with English, but normally "immobile" means that something CAN'T move, as opposed to not currently moving.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -367,6 +367,14 @@ private void EngagePrey(Vector3 target, Random random, bool engulf)
...
+        // Used to get newly spawned microbes to move.
+        if (microbe.MovementDirection.Length() == 0)
+        {
+            MoveWithRandomTurn(0, 2 * Mathf.Pi, random);
```
adQuid, line 374:
```suggestion
            MoveWithRandomTurn(0, Mathf.Pi, random);
```

MoveWithRandomTurn picks a direction and turns relative to the current heading, so just pi should be enough to be able to turn any direction.

adQuid (changes requested):
Currently unable to test this, but I see a couple things in the code.

src/microbe_stage/MicrobeAI.cs
Maxonovien: If I may say, you really should consider leaving us a memo when you change the meaning of the words you borrow from us. :p Actually was actually intended, but both case works so I'll just leave it like this. Thanks!

adQuid: I'll get back to you with a list of all culture America has ruined, but that might be beyond the scope of this PR.

adQuid (approved):
Can't prove it works, but the code looks good.

adQuid:
Ran playtesting. I'm not seeing any issues with the AI and it does improve behavior. I'd like to see this merged, but I'm not sure if we can say this closes #889 

That issue seems to be describing a delay before even calling the AI code instead of a delay in the AI itself. I've tried to find an example of this happening (I recall observing it), but I wasn't able to in my 15 minutes of testing. On the other hand, this issue is 2 years old, and may have just gone away some other way.

@hhyyrylainen thoughts?

Maxonovien:
I thought similarly at first, but the code actually sets new microbeAI to think at the first think step (because their TimeUntilNextUpdate is set to 0, and only got reset at AI_THINK_INTERVAL upon doing the first move). While it was the first reason for the issue, it is fixed already. I think this is what hh pointed out in the last comment on the issue.

hhyyrylainen:
@adQuid I debugged #889 quite a bit and determined that the AI update for a new cell is ran on the next frame. It's just the AI decides very often to not move. That's why the issue is worded like it is.

adQuid:
In which case I've checked off this ticket as tested.

hhyyrylainen:
Does this still allow really lazy species (with the personality values) to not move when spawned? If so I'm happy with this also.

Maxonovien:
The current changes only affect Run & Rumble algorithm, which only triggers if `SpeciesActivity > Constants.MAX_SPECIES_ACTIVITY / 10`. So sessile organisms, which do not qualify, should not be affected.

No difference is made otherwise, but this is the case for the whole movement algorithm in general.

adQuid:
Well update the branch and let's see if we got a merge on our hands!

hhyyrylainen (approved):

buckly90:
Appears to work great without a hitch. Didn't notice anything unusual as I played, and maybe had a bit too much fun endlessly duplicating my cell to watch the cells quickly float away like balloons.

adQuid:
I admit I really did like the cone effect from before this change. It's like I was blasting a firehose of other cells.

PR #2427, opened 2021-07-18
2131 entirely extinct marker

### LINKED ISSUE #2131: Differentiate graph marker indicating extinct species
Currently an X icon is used to mark any extinction types on the species population graph, and may create the impression that a species have gone extinct entirely even though it's only patch-specific. Would be nice if a unique icon is used to indicate that.

Maxonovien:

This PR displays entirely extinct species with a specific marker (currently a skull from old icons)

image

**Related Issues**

Closes #2131 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -798,15 +798,21 @@ public void UpdateReportTabStatistics(Patch patch)
...
                 // Clamp population number so it doesn't go into the negatives
                 var population = extinctInPatch ? 0 : entry.Value;
 
                 var dataPoint = new DataPoint
                 {
                     Value = new Vector2((float)snapshot.TimePeriod, population),
-                    Size = extinctInPatch ? 12 : 7,
-                    IconType = extinctInPatch ? DataPoint.MarkerIcon.Cross : DataPoint.MarkerIcon.Circle,
+                    Size = extinctInPatch ? extinctEverywhere ? 28 : 12 : 7,
+
```
Kasterisk, line 810: Shouldn't be any whitespace here.

src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -798,15 +798,21 @@ public void UpdateReportTabStatistics(Patch patch)
                 var dataset = speciesPopulationChart.GetDataSet(entry.Key.FormattedName);
 
                 var extinctInPatch = entry.Value <= 0;
+                var extinctEverywhere = snapshot.SpeciesGoneDefinitelyExtinct.Contains(entry.Key);
```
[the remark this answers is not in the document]
hhyyrylainen: As long as external effects are still applied before extinctions are checked it should be fine.

src/microbe_stage/Patch.cs
```diff
@@ -188,15 +214,11 @@ public object Clone()
...
+            SpeciesInPatch = new Dictionary<Species, long>(SpeciesInPatch),
+            SpeciesGoneDefinitelyExtinct = new List<Species>(SpeciesGoneDefinitelyExtinct),
             Biome = (BiomeConditions)Biome.Clone(),
         };
 
-        foreach (var entry in SpeciesInPatch)
-        {
-            result.SpeciesInPatch.Add(entry.Key, entry.Value);
-        }
-
```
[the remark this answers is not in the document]
hhyyrylainen: Cloning species is starting to sound a bit expensive...
I already in some recent PR said that cloning a species object should be avoided, unless there's no other way to do something.

src/microbe_stage/editor/MicrobeEditorGUI.cs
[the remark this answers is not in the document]
Maxonovien: This was one of the possibility, but for this to work we need species cloning as underlined below. And, as hh said, I fear this might get a bit expensive. 

There is another potential issue with the fact that we won't be able to (easily) access current state of the species, although this may not truly be a problem for a snapshot (and actually makes sense).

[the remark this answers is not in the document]
hhyyrylainen: Species global population could be split out to a separate class like WorldWideSpeciesInfo or something.

src/microbe_stage/Patch.cs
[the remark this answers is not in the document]
Maxonovien: As said above, cloning the species is necessary to get a solution with global population evaluation. I think recording the species that wen extinct at this round, although less elegant, is not necessarily worse.

However, this brings the benefit of not having to deal with a list to check, although it should stay rather small in size, given that not too many species will go extinct most times, and that snapshot are used few times anyways...

src/microbe_stage/editor/MicrobeEditorGUI.cs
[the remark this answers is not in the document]
Maxonovien: It could make sense for more complex stuff, but I'm not sure we'd have much gain from doing this, compared to simply having an extinction list?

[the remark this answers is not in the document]
hhyyrylainen: @Kasterisk probably has a better suggestion on how to store this bit of information for the graphs.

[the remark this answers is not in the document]
Maxonovien: You mean, the global population comparison? I'm open to this, but that will requires species cloning to keep the pop count of this time step (not to mark fully extinct pops that only went fully extinct later on).

src/microbe_stage/Patch.cs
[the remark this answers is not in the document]
Kasterisk: Regarding the species cloning problem, I like the separate species data class solution as suggested by hhyyrylainen. Maybe a `MicrobeSpeciesInfo` class defined inside Patch (or on its own) where it will be an instance member of PatchSnapshot of type Dictionary called `RecordedSpeciesInfo`. It could be instantiated and populated from every `SpeciesInPatch` on snapshot creation.

src/microbe_stage/editor/MicrobeEditorGUI.cs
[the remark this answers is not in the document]
Kasterisk: See https://github.com/Revolutionary-Games/Thrive/pull/2427#discussion_r672469159. In general it would be better in the long term, in case there are things in the future that we would like recorded from a species, that data class could come in handy.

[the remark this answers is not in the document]
Maxonovien: I guess this could in fact be of use, if we ever want to record things like species structure at this very moment... although I'm still not sure this would be the case. This might still be a bit overkill, but I'll give it a closer look when I'll get back to coding.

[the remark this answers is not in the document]
hhyyrylainen: It seems people want to see a kind of ancestry tree for species. So in the future stuff like generated screenshots could be maybe saved here? I'm sure we can think of some changing species details that could also be stored here.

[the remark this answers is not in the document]
Maxonovien: That's precisely what I thought of, but this would currently be accessed from a patch, so it feels... weird. I guess we can have the instance created outside the patch and simply be referenced in all the patch snapshot instances (because it will be discarded anyways after a given count).

But where to store it for the tree? A world snapshot list (we won't go deque as we don't want to discard them)? Obviously, from there it starts getting complicated and outside the scope of the PR, which is why I was cautious at first. But I think I can do what I proposed in the first paragraph, and leave the rest for latter.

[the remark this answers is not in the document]
hhyyrylainen: I mean the tree wouldn't be stored separately. I think that Species should store where they split off from (at least the ID of the parent species) so the tree can be generated quickly when the player wants to look at it.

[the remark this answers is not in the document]
Maxonovien: i didn't intend to mean species splitting, but how one species evolve over time (it isn't considered a new species from the game perspective, which could be argued both for and against scientifically). The player's primum thrivum end up being no longer a single-cytoplasm after a while.

src/gui_common/charts/DataPoint.cs
```diff
@@ -94,6 +96,9 @@ public override void _Ready()
         graphMarkerCircle = GD.Load<Texture>("res://assets/textures/gui/bevel/graphMarkerCircle.png");
         graphMarkerCross = GD.Load<Texture>("res://assets/textures/gui/bevel/graphMarkerCross.png");
 
+        // Used an old icon here
+        graphMarkerSkull = GD.Load<Texture>("res://assets/textures/gui/bevel/SuicideIcon.png");
+
```
adQuid, line 101:
```suggestion
        graphMarkerSkull = GD.Load<Texture>("res://assets/textures/gui/bevel/SuicideIcon.png");

```

If we don't want to use the old icon, we'll decide that here and change it before the merge. If we're cool with it, this will just be another icon. Either way, separating it and noting it here will only confuse the future dev.

src/general/Species.cs
```diff
@@ -155,6 +155,22 @@ public void BecomePlayerSpecies()
         PlayerSpecies = true;
     }
 
+    /// <summary>
+    /// Gets info specific to the species for saving.
+    /// Used for patch snapshots, but could be expanded
+    /// </summary>
+    /// <remarks> TODO: Check overlap with ClonePropertiesTo </remarks>
```
Kasterisk, line 162:
```suggestion
    /// <summary>
    ///   Gets info specific to the species for storing into a new container class.
    ///   Used for patch snapshots, but could be expanded
    /// </summary>
    /// <remarks>TODO: Check overlap with ClonePropertiesTo</remarks>
```

src/general/SpeciesInfo.cs
```diff
@@ -0,0 +1,5 @@
+public class SpeciesInfo
```
Kasterisk, line 1: Could have a general explanation what SpeciesInfo is and what it could be used for. Beneficial for future programmers.

src/microbe_stage/Patch.cs
```diff
@@ -180,23 +185,21 @@ public class PatchSnapshot : ICloneable
     public double TimePeriod;
 
     public Dictionary<Species, long> SpeciesInPatch = new Dictionary<Species, long>();
+    public Dictionary<Species, SpeciesInfo> RecordedSpeciesInfo = new Dictionary<Species, SpeciesInfo>();
 
     public BiomeConditions Biome;
 
     public object Clone()
     {
+        // We only do a shallow copy of RecordedSpeciesInfo here, but SpeciesInfo are never modified .
```
Kasterisk, line 194:
```suggestion
        // We only do a shallow copy of RecordedSpeciesInfo here, but SpeciesInfo are never modified.
```

Kasterisk (changes requested):
Forgot to review this, a few suggestions. Everything else looks probably fine.

Kasterisk (approved):
Works fine!

src/general/Species.cs
```diff
@@ -155,6 +155,22 @@ public void BecomePlayerSpecies()
...
+    /// <remarks>TODO: Check overlap with ClonePropertiesTo</remarks>
+    public SpeciesInfo RecordSpeciesInfo()
+    {
+        var info = new SpeciesInfo
+        {
+            ID = ID,
+            Population = Population,
+        };
+
+        return info;
```
hhyyrylainen, line 171:
```suggestion
        return new SpeciesInfo
        {
            ID = ID,
            Population = Population,
        };
```

src/general/SpeciesInfo.cs
```diff
@@ -0,0 +1,11 @@
+/// <summary>
+///   Class that stores species-wide information.
+///   It can be expanded to have any species parameter we might want to keep,
+///   e.g. base structure, behavioral values... and such to draw species history.
+/// </summary>
+/// <remarks> This class does not account for individual variations!</remarks>
```
hhyyrylainen, line 6: What does `does not account for individual variations!` mean? I have no clue. This needs more explanation...

src/general/SpeciesInfo.cs
```diff
@@ -0,0 +1,11 @@
+/// <summary>
+///   Class that stores species-wide information.
+///   It can be expanded to have any species parameter we might want to keep,
+///   e.g. base structure, behavioral values... and such to draw species history.
```
hhyyrylainen, line 4: These two lines feel like a remark to me.
Btw remarks should be formatted like this:
```
27:    /// <remarks>
28-    ///   <para>
29-    ///     There's a small bug where if the tutorials are turned back on and then displayed, that leaves the checkboxes
30-    ///     unchecked, when things become visible all the enable tutorials check boxes would need to read this value
31-    ///     to fix this
32-    ///   </para>
33:    /// </remarks>
```

src/microbe_stage/Patch.cs
```diff
@@ -180,23 +185,21 @@ public class PatchSnapshot : ICloneable
     public double TimePeriod;
 
     public Dictionary<Species, long> SpeciesInPatch = new Dictionary<Species, long>();
+    public Dictionary<Species, SpeciesInfo> RecordedSpeciesInfo = new Dictionary<Species, SpeciesInfo>();
 
     public BiomeConditions Biome;
 
     public object Clone()
     {
+        // We only do a shallow copy of RecordedSpeciesInfo here, but SpeciesInfo are never modified.
```
hhyyrylainen, line 194:
```suggestion
        // We only do a shallow copy of RecordedSpeciesInfo here as SpeciesInfo objects are never modified.
```

src/microbe_stage/Patch.cs
```diff
@@ -180,23 +185,21 @@ public class PatchSnapshot : ICloneable
...
-            SpeciesInPatch = new Dictionary<Species, long>(SpeciesInPatch.Count),
+            SpeciesInPatch = new Dictionary<Species, long>(SpeciesInPatch),
+            RecordedSpeciesInfo = new Dictionary<Species, SpeciesInfo>(RecordedSpeciesInfo),
             Biome = (BiomeConditions)Biome.Clone(),
         };
 
-        foreach (var entry in SpeciesInPatch)
-        {
-            result.SpeciesInPatch.Add(entry.Key, entry.Value);
-        }
```
hhyyrylainen, line 198: This was changed to `SpeciesInPatch = new Dictionary<Species, long>(SpeciesInPatch),` I assume that properly copies the info in the dictionary? For some reason in the past I've written code like this in multiple places in Thrive, I don't remember why I didn't use that shorter code...

src/microbe_stage/PatchMap.cs
```diff
@@ -215,6 +215,7 @@ public List<Species> RemoveExtinctSpecies(bool playerCantGoExtinct = false)
 
         foreach (var patch in Patches)
         {
+            // Looks like player species is selected iff not playercantgoextinct...
```
hhyyrylainen, line 218:
```suggestion
            // Player species can only go extinct when playerCantGoExtinct is true
```

src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -798,15 +798,20 @@ public void UpdateReportTabStatistics(Patch patch)
...
                 var population = extinctInPatch ? 0 : entry.Value;
 
                 var dataPoint = new DataPoint
                 {
                     Value = new Vector2((float)snapshot.TimePeriod, population),
-                    Size = extinctInPatch ? 12 : 7,
-                    IconType = extinctInPatch ? DataPoint.MarkerIcon.Cross : DataPoint.MarkerIcon.Circle,
+                    Size = extinctInPatch ? extinctEverywhere ? 28 : 12 : 7,
+                    IconType = extinctInPatch ?
+                        extinctEverywhere ?
```
hhyyrylainen, line 811: I feel like nested ternary operators are a bad code smell, and this should be refactored out to be an if-block at this point for readability.

hhyyrylainen (changes requested):
Functionality-wise this seems to work. I'd like to see some code cleanup before merging, though.

src/microbe_stage/Patch.cs
Maxonovien: From what I understand it does. I was surprised as well by the code, perhaps it simply went under the radar? I don't see any reason not to use the other method.

hhyyrylainen: If it works then it is good (and copies the data the same way as this explicit .Add call). And I don't see a problem with converting the remaining code also to use this approach, though that's probably best saved to a separate small cleanup PR.

src/microbe_stage/PatchMap.cs
Maxonovien: This was a leftover note to self, because it seemed there was a bug lying here: player species should go extinct only when playerCantGoExtinct is False...

hhyyrylainen: I reworded this so that it makes sense in general. I think this can be kept as an explanation to future programmers to see.

src/microbe_stage/editor/MicrobeEditorGUI.cs
Maxonovien: I usually never use ternary ops. But here I really felt they were intuitive, especially when indented, as you can grasp the whole structure in an eyeblick due to the short code length. That's why I kept the nested structure

However, I'm not against going with if-blocks, so if you feel ternary operators may hamper readability, I'lll just use if-blocks instead.

hhyyrylainen: I have a personal rule that nested ternaries are a no-go. non-nested ternaries are fine to shorten code, but here I think the readability is actually much worse due to the nested ternary usage, so I would not use that here.

src/microbe_stage/PatchMap.cs
Maxonovien: Yeah, but this explanation (matching the code) is contradictory, so this may be confusing. Though, I guess It makes the issue more visible, so we may in fact keep your wording (which can be changed after correction).

hhyyrylainen: Oh, did I accidentally write the opposite of what I should have...

hhyyrylainen: After looking at the code carefully, it's starting to look like the logical condition is actually the wrong way around in the code here? If you don't mind, could you make a separate PR from flipping that condition the right way around so that we can discuss this there?

src/microbe_stage/editor/MicrobeEditorGUI.cs
Maxonovien: I'm fine with having this rule. Could be worth to mention it in the style guide and add it to the style cop though.

hhyyrylainen: I would have enabled a check if there was one (feel free to enable it if I happened to miss one).
https://github.com/Revolutionary-Games/Thrive/issues/2454

Maxonovien: I don't know about the checks, I just know it wanted to format my ternary op nesting. I'll take a look during the week-end.

hhyyrylainen: Only thing I could find in Rider settings is enforcing a different style for nested ternaries, which will make them more noticeable (and harder to manually write right).

src/microbe_stage/PatchMap.cs
Maxonovien: Yes, I think the issue is indeed that it should be !PlayerCantGoExtinct in the patch. I'll open a PR next time I boot on windows, should be either tonight or tomorrow. Made myself a memo for this

hhyyrylainen (approved):

hhyyrylainen:
I just noticed a very critical bug (did no one test loading a save after this?):
```
Error Loading - An exception happened while processing loaded objects exception: System.Collections.Generic.KeyNotFoundException: The given key 'Primum Thrivium (1)' was not present in the dictionary.
  at System.Collections.Generic.Dictionary`2[TKey,TValue].get_Item (TKey key) [0x0001e] in <46038998c7d34a1abad25f54bc1b08e4>:0 
  at MicrobeEditorGUI.UpdateReportTabStatistics (Patch patch) [0x0031a] in /home/hhyyrylainen/Projects/thrive/src/microbe_stage/editor/MicrobeEditorGUI.cs:796 
  at MicrobeEditor.OnLoadedEditorReady () [0x0005e] in /home/hhyyrylainen/Projects/thrive/src/microbe_stage/editor/MicrobeEditor.cs:2313 
  at MicrobeEditor.InitEditorSaved () [0x0000e] in /home/hhyyrylainen/Projects/thrive/src/microbe_stage/editor/MicrobeEditor.cs:1375 
  at MicrobeEditor.InitEditor () [0x0004b] in /home/hhyyrylainen/Projects/thrive/src/microbe_stage/editor/MicrobeEditor.cs:1284 
  at MicrobeEditor.OnEnterEditor () [0x000e0] in /home/hhyyrylainen/Projects/thrive/src/microbe_stage/editor/MicrobeEditor.cs:1172 
  at MicrobeEditor._Ready () [0x000b8] in /home/hhyyrylainen/Projects/thrive/src/microbe_stage/editor/MicrobeEditor.cs:441 
  at (wrapper managed-to-native) Godot.NativeCalls.godot_icall_2_426(intptr,intptr,intptr,bool)
  at Godot.Node.AddChild (Godot.Node node, System.Boolean legibleUniqueName) [0x00001] in /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Generated/GodotObjects/Node.cs:526 
  at SceneManager.SwitchToScene (Godot.Node newSceneRoot, System.Boolean keepOldRoot) [0x00031] in /home/hhyyrylainen/Projects/thrive/src/engine/SceneManager.cs:49 
  at InProgressLoad.Step () [0x00171] in /home/hhyyrylainen/Projects/thrive/src/saving/InProgressLoad.cs:149 
```

PR #2439, opened 2021-07-23
Added some new formatting checks and renamed calcium_carbonate

### LINKED ISSUE #1472: Make script for checking export_presets.cfg has the right version number
and make CI run it to make sure the info in there doesn't get outdated, like the version number and the copyright year is the same as in the license file.

### LINKED ISSUE #2090: Add formatting check to make sure no .mo files exist
as those are binary variants of the .po files, and they are something we don't use

### LINKED ISSUE #2150: Add a formatting check to check scene node names
Add a formatting check for the node names property:
```
[node name="AName"
```

It should make sure that the node name:
- Contains at least one capital
- Doesn't contain spaces
- Doesn't contain underscores

It should additionally mention to use PascalCase for the names.

### LINKED ISSUE #2152: Add save version upgrader / converter system
To support loading older saves that need manual conversion operations to be applied before they work in a newer version.

hhyyrylainen:
Added new checks for version number in export_presets.cfg, no .mo files exist, and that scene nodes follow some naming rules.
This required renaming calcium_carbonate to calciumCarbonate.

Also adds a save converter to not break save compatibility over that rename.

**Related Issues**

closes #1472
closes #2090
closes #2150
closes #2152

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


Kasterisk:
You may want to update the incompatible save dialog as it still mentions that a save upgrader is not implemented.

Haven't tested extensively but from a quick look nothing is gamebreaking yet.

hhyyrylainen:
I suppose so. I guess I'll replace that text with that there is no save upgrader from that version. I'll do that in a bit as I'm uploading a new CI image currently and about to push changes to another branch once that is done.

src/saving/ISaveUpgradeStep.cs
```diff
@@ -0,0 +1,170 @@
...
+        {
+            foreach (var entry in saveData.Properties())
+            {
+                RecursivelyUpdateMembraneValues(entry);
+            }
+        }
+
+        private void RecursivelyUpdateMembraneValues(JProperty property)
+        {
+            if (property.Name.Contains("Membrane") || property.Name.Contains("membrane"))
```
FuJa0815, line 68:
```suggestion
            if (property.Name.ToLower().Contains("membrane"))
```

src/saving/ISaveUpgradeStep.cs
```diff
@@ -0,0 +1,170 @@
...
+            // TODO: would it be useful to specify a range of versions an upgrader can upgrade to make it less error
+            // prone to accidentally miss a version
+            return new Dictionary<string, ISaveUpgradeStep>
+            {
+                { "0.5.4.0", new UpgradeStep054To055() },
+            };
+        }
+    }
+
+    internal class UpgradeStep054To055 : BaseJSONUpgradeStep
```
FuJa0815, line 54: I think it would be better to have a seperate file for the UpgradeSteps

src/saving/ISaveUpgradeStep.cs
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    internal abstract class BaseJSONUpgradeStep : ISaveUpgradeStep
+    {
+        protected abstract string VersionAfter { get; }
+
+        public string PerformUpgrade(SaveInformation saveInfo, string inputSave, string outputSave)
+        {
+            if (VersionUtils.Compare(VersionAfter, saveInfo.ThriveVersion) <= 0)
```
FuJa0815, line 100: what if this comparison failed (it returned `int.MaxValue`)?

src/saving/SaveList.cs
```diff
@@ -226,6 +249,69 @@ private void OnConfirmLoadInvalid()
...
+
+        if (SaveUpgrader.IsSaveABackup(saveToBeLoaded))
+        {
+            saveToBeLoaded = SaveUpgrader.RemoveBackupSuffix(saveToBeLoaded);
+            GD.Print("Selected save is a backup, really going to load after upgrade: ", saveToBeLoaded);
+        }
+
+        // Perform save upgrade (the game will lag here, but I'll leave it to someone else to make a progress bar)
+        // Instead could show a popup with a spinner on it and run the upgrade with TaskExecutor in the background
+        var task = new Task(() => { SaveUpgrader.PerformSaveUpgrade(saveToUpgrade); });
```
FuJa0815, line 267: I think those brackets are not nececery

FuJa0815 (changes requested):
Wasn't able to read through all of it yet or test it.

Does this support using two upgraders on one savefile?
Like if there is a save upgrader from version 0.5.4.0 from 0.5.5.0 and one from 0.5.5.0 to 0.5.6.0 and you want to load a save made in version 0.5.4.0 in version 0.5.6.0?

hhyyrylainen:
> Does this support using two upgraders on one savefile?
Like if there is a save upgrader from version 0.5.4.0 from 0.5.5.0 and one from 0.5.5.0 to 0.5.6.0 and you want to load a save made in version 0.5.4.0 in version 0.5.6.0?

Yes, it won't start upgrading if it can't find a sequence of upgraders that can upgrade the save all the way to the current game version. Once upgrading is started it will run all the upgraders in a row (I tested this by changing my game version to 0.5.5.0 and adding a just version number upgrader for 0.5.5.0-alpha).

src/saving/ISaveUpgradeStep.cs
hhyyrylainen: I did it this way because a file needs to be named the same as the first class in it. So I would have to put like a dummy class in a file to get a sensible name for it...

check_formatting.rb
```diff
@@ -156,6 +179,13 @@ def handle_gd_file(_path)
   true
 end
 
+def handle_mo_file(_path)
+  OUTPUT_MUTEX.synchronize do
+    error '.mo files are not allowed'
```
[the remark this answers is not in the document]
hhyyrylainen: It returns true two lines below this from the method. Similarly to handle_gd_file the error condition is already triggered when entering this method so there's no need to have a separate variable tracking the state.

src/saving/ISaveUpgradeStep.cs
hhyyrylainen: I'm quite torn on this one...
This will need to allocate a new converted string for *each* key in the save json (which can be multiple megabytes large), while also only marginally shortening this. I went with this approach as I only really considered there possibly being these two values possible.

hhyyrylainen:
> You may want to update the incompatible save dialog as it still mentions that a save upgrader is not implemented.

Fixed

src/saving/ISaveUpgradeStep.cs
FuJa0815: > a file needs to be named the same as the first class in it

I don't recall there being this requirement in C#

hhyyrylainen: Jetbrains will complain about that if it isn't the case.

PR #2441, opened 2021-07-23
fixes #2099 When re-entering load menu and pressing load game crashes

### LINKED ISSUE #2099: When re-entering load menu and pressing load game crashes
Steps:
- Go to load game from main menu
- Select a save
- Go back to main menu
- Go to the load game menu again
- Load button is active even though nothing is selected, press it
- Get the following crash:
```
Unhandled Exception:
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'SaveListItem'.
  at Godot.Object.GetPtr (Godot.Object instance) [0x0001b] in /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Core/Object.base.cs:37 
  at Godot.Object.IsConnected (System.String signal, Godot.Object target, System.String method) [0x00001] in /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Generated/GodotObjects/Object.cs:537 
  at TransitionManager.StartTransitions (Godot.Object target, System.String onFinishedMethod) [0x00043] in /home/hhyyrylainen/Projects/thrive/src/gui_common/TransitionManager.cs:117 
  at SaveListItem.LoadThisSave () [0x00094] in /home/hhyyrylainen/Projects/thrive/src/saving/SaveListItem.cs:246 
  at SaveManagerGUI.LoadFirstSelectedSave () [0x00015] in /home/hhyyrylainen/Projects/thrive/src/saving/SaveManagerGUI.cs:180 
  at SaveManagerGUI.LoadFirstSelectedSaveButtonPressed () [0x0000c] in /home/hhyyrylainen/Projects/thrive/src/saving/SaveManagerGUI.cs:187 
```

Looks like a bug that `Selected` in SaveManagerGUI is not cleared when new items are loaded, so that it keeps references to the old objects that get disposed.

TomDev03:

This PR fixes #2099. The game was crashing when selecting a saved game, returning to the main menu, enter again in the load game screen and clicking Load.

The problem that I found is that every time we enter the load menu the saveList it is refreshed and the SaveManagerUI it is not and that cause the variable wich contains the selected save file not to be updated so it have a null object.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/saving/SaveList.cs
+++ b/src/saving/SaveList.cs
@@ -126,7 +126,6 @@ public class SaveList : ScrollContainer
             savesList.AddChild(item);
         }
 
-        EmitSignal(nameof(OnItemsChanged));
         loadingItem.Visible = false;
         refreshing = false;
     }
@@ -153,6 +152,7 @@ public class SaveList : ScrollContainer
         loadingItem.Visible = true;
         readSavesList = new Task<List<string>>(() => SaveHelper.CreateListOfSaves());
         TaskExecutor.Instance.AddTask(readSavesList);
+        EmitSignal(nameof(OnItemsChanged));
     }
 
     private void OnSubItemSelectedChanged()
@@ -181,6 +181,7 @@ public class SaveList : ScrollContainer
         SaveHelper.DeleteSave(saveToBeDeleted);
         saveToBeDeleted = null;
 
+        Refresh();
         EmitSignal(nameof(OnItemsChanged));
     }
 
--- a/src/saving/SaveList.cs
+++ b/src/saving/SaveList.cs
@@ -126,6 +126,7 @@ public class SaveList : ScrollContainer
             savesList.AddChild(item);
         }
 
+        EmitSignal(nameof(OnItemsChanged));
         loadingItem.Visible = false;
         refreshing = false;
     }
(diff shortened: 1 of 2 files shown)


hhyyrylainen:
How about instead refreshing the list when returning to the menu? I think that makes much more sense as now the list refresh operation, which might take some time, saps performance unnecessarily if the user never returns to the load menu.

hhyyrylainen:
Looking at the code I got the idea that wouldn't changing `selectedDirty` work? That way Selected will be recomputed when next needed (for example this could be set where the old items are removed)

TomDev03:
@hhyyrylainen 

> Looking at the code I got the idea that wouldn't changing `selectedDirty` work? That way Selected will be recomputed when next needed (for example this could be set where the old items are removed)

Yes I will check that.

TomDev03:
I found te problem. So the problem is that every time I go to Load game screen the saveList is updated but the SaveMangerGUI not. That's cause the old save to be disposed and created new ones, and finally the GUI not be updated and a null object be in the  GUI Selected property.

FuJa0815 (approved):
No bugs found. Great work!

hhyyrylainen (approved):

Issue #1870, opened 2020-11-28
Texts on the dropdown buttons are not localized


Kasterisk:
Looks like they are not currently translatable.
![2020-11-28_19 27 57 7943](https://user-images.githubusercontent.com/54026083/100515599-ea027700-31af-11eb-9266-505bd33dad2d.png)
(e.g. The `Disabled` and `None` text above should be localized)

hhyyrylainen:
There are two potential things that will determine how hard this is to do:
- Godot dropdowns don't support localizing the items (this will require us to manually do the translations or make a custom dropdown widget)
- The pybabel translation extraction doesn't know about the dropdowns (need changes there to make sure this is supported, this might turn out to be a bit more difficult as this requires different domain knowledge).

PR #2444, opened 2021-07-26
2374 output devices

### LINKED ISSUE #2374: Add an option to select the sound output device
At least Godot let's us list the audio devices: https://docs.godotengine.org/fi/stable/classes/class_os.html?highlight=get_current_video_driver#class-os-method-get-audio-driver-name

would be nice to show that in the options menu and if possible allow the user to select the sound output device in the game.

FuJa0815:
Lets the user select an audio output device.
Closes #2374 

I was unable to test the functionality of this PR because I only have one sound card, but I could hear a little noise when changing between the default output device and my soundcard.

**Progress Checklist**

      break existing features:
      members reviewing this PR)
- [x] Functionality is confirmed working by another person on windows (see above checklist link)
- [ ] Functionality is confirmed working by another person on linux (see above checklist link)

the PR author.


FuJa0815:
@TomDev03 I can't request a review from you, because you aren't a github member, but you can still test the pull request and tell me if it worked.

TomDev03:
@FuJa0815 I ran your code and it worked when I was on the game but after that I close the game and reopen it, the game start but 2 seconds after and the game close and got 2 erros. Now I cannot run the game.
E 0:00:03.326   get_node: (Node not found: "" (relative to "/root/MainMenu/OptionsMenu").)
  <C++ Error>   Condition "!node" is true. Returned: __null
  <C++ Source>  scene/main/node.cpp:1371 @ get_node()
  <Stack Trace> :0 @ Godot.Node Godot.NativeCalls.godot_icall_1_578(IntPtr , IntPtr , IntPtr )()
                Node.cs:618 @ Godot.Node Godot.Node.GetNode(Godot.NodePath )()
                NodeExtensions.cs:7 @ Godot.OptionButton Godot.Node.GetNode<Godot.OptionButton >(Godot.NodePath )()
                OptionsMenu.cs:327 @ void OptionsMenu._Ready()()
                
E 0:00:03.362   void OptionsMenu.LoadOutputDevices(Godot.OptionButton ): System.NullReferenceException: Object reference not set to an instance of an object.
  <C++ Error>   Unhandled exception
  <C++ Source>  C:\Users\tomas\Desktop\Thrive\src\general\OptionsMenu.cs:757 @ void OptionsMenu.LoadOutputDevices(Godot.OptionButton )()
  <Stack Trace> OptionsMenu.cs:757 @ void OptionsMenu.LoadOutputDevices(Godot.OptionButton )()
                OptionsMenu.cs:331 @ void OptionsMenu._Ready()()

TomDev03:
@FuJa0815 I test your code and it worked for me. At the begining the option button got null don't know why but after I reload godot it got normal.

src/general/OptionsMenu.cs
```diff
@@ -732,6 +745,21 @@ private void UpdateResetSaveButtonState()
         saveButton.Disabled = result;
     }
 
+    private void UpdateDefaultOutputDeviceText(OptionButton optionButton)
+    {
+        optionButton.SetItemText(0, TranslationServer.Translate("DEFAULT_fOUTPUT_DEVICE"));
```
84634E1A607A, line 750: Works fine. But why is here an f? Typo?

84634E1A607A:
Linux Mint working fine.

src/general/OptionsMenu.cs
```diff
@@ -732,6 +745,21 @@ private void UpdateResetSaveButtonState()
         saveButton.Disabled = result;
     }
 
+    private void UpdateDefaultOutputDeviceText(OptionButton optionButton)
+    {
+        optionButton.SetItemText(0, TranslationServer.Translate("DEFAULT_OUTPUT_DEVICE"));
```
hhyyrylainen, line 750: Does it actually work if you have the item in the dropdown with this text and then having this text in en.po manually?
I'd be interested to know as this is important to know regarding: https://github.com/Revolutionary-Games/Thrive/issues/1870

hhyyrylainen:
This looks to be missing the english translation:

src/general/OptionsMenu.cs
FuJa0815: the way this works is:
The C# list uses the untranslated names (therefore "Default"). When adding the items to the godot dropdown I add the translated value instead of the real list value.

The updating does not use the text but rather the index of the selected item, therefore I can lookup the index in my C# list and get the untranslated value. You could probably generalize this logic a bit to have translatable drop down menus.

FuJa0815:
God, I hate all those merge conflicts with localization files...
Should be working now. You can test the languages with german.

src/general/OptionsMenu.cs
hhyyrylainen: I suppose this is a special case, yes, as the content is dynamically generated...

hhyyrylainen:
> God, I hate all those merge conflicts with localization files...
Should be working now. You can test the languages with german.

As you only presumably added a couple of translations, I've found that the easiest way is to just checkout the files from master, rerun the localize script, and manually re-apply the translations you want to add.

hhyyrylainen:
Is this now again ready to review? You haven't checked off the first item in the PR checklist...

FuJa0815:
Whoops, thought I've already checked it. I did pull the locale folder from master in the end

src/engine/Settings.cs
```diff
@@ -133,6 +133,8 @@ private Settings()
     /// </summary>
     public SettingValue<bool> VolumeGUIMuted { get; set; } = new SettingValue<bool>(false);
 
+    public SettingValue<string> SelectedOutputDevice { get; set; } = new SettingValue<string>("Default");
```
hhyyrylainen, line 136: I think "Default" should be put in Constants to better explain that it is a built-in value that Godot recognizes.

src/engine/Settings.cs
```diff
@@ -532,6 +535,26 @@ public void ApplyWindowSettings()
         OS.VsyncEnabled = VSync;
     }
 
+    /// <summary>
+    ///   Applies current output device settings to the audio system
+    /// </summary>
+    public void ApplyOutputDeviceSettings()
+    {
+        string outputDevice = SelectedOutputDevice.Value;
```
hhyyrylainen, line 543:
```suggestion
        var outputDevice = SelectedOutputDevice.Value;
```

src/engine/Settings.cs
```diff
@@ -532,6 +535,26 @@ public void ApplyWindowSettings()
...
+        if (string.IsNullOrEmpty(outputDevice))
+        {
+            outputDevice = "Default";
```
hhyyrylainen, line 546: use the constant here as well

src/engine/Settings.cs
```diff
@@ -532,6 +535,26 @@ public void ApplyWindowSettings()
...
+        }
+
+        // If the selected output device is invalid Godot resets AudioServer.Device to Default.
+        // It seems like there is some kind of threading going on. The getter of AudioServer.Device
+        // only returns the new value after some time, therefore we can't check if the output device
+        // got applied successfully.
+        AudioServer.Device = outputDevice;
+
+        GD.Print("Set output device to ", outputDevice);
```
hhyyrylainen, line 555:
```suggestion
        GD.Print("Set audio output device to: ", outputDevice);
```

src/engine/Settings.cs
```diff
@@ -532,6 +535,26 @@ public void ApplyWindowSettings()
         OS.VsyncEnabled = VSync;
     }
 
+    /// <summary>
+    ///   Applies current output device settings to the audio system
+    /// </summary>
+    public void ApplyOutputDeviceSettings()
```
hhyyrylainen, line 541:
```suggestion
    public void ApplyAudioOutputDeviceSettings()
```

src/general/OptionsMenu.cs
```diff
@@ -99,6 +99,15 @@ public class OptionsMenu : ControlWithInput
     [Export]
     public NodePath GUIMutedPath;
 
+    [Export]
+    public NodePath OutputDeviceSelectionPath;
```
hhyyrylainen, line 103: I think all the things related to the audio device changing need renaming to be about the *audio* device. "output device" could be something unrelated to audio...

src/general/OptionsMenu.cs
```diff
@@ -172,6 +175,11 @@ public class OptionsMenu : ControlWithInput
         .OrderBy(i => i, StringComparer.InvariantCulture)
         .ToList();
 
+    private static readonly List<string> OutputDevicesCache = AudioServer
+        .GetDeviceList()
+        .OfType<string>()
+        .ToList();
```
hhyyrylainen, line 181: I think there fit on a single line.

src/general/OptionsMenu.cs
```diff
@@ -732,6 +745,21 @@ private void UpdateResetSaveButtonState()
         saveButton.Disabled = result;
     }
 
+    private void UpdateDefaultOutputDeviceText(OptionButton optionButton)
+    {
+        optionButton.SetItemText(0, TranslationServer.Translate("DEFAULT_OUTPUT_DEVICE"));
+    }
+
+    private void LoadOutputDevices(OptionButton optionButton)
```
hhyyrylainen, line 753: `optionButton` should be renamed to be more specific

src/general/OptionsMenu.cs
hhyyrylainen, line 750:
```suggestion
        optionButton.SetItemText(0, TranslationServer.Translate("DEFAULT_AUDIO_OUTPUT_DEVICE"));
```

src/general/OptionsMenu.cs
```diff
@@ -1167,6 +1205,11 @@ private void OnTranslationSitePressed()
         OS.ShellOpen("https://translate.revolutionarygamesstudio.com/engage/thrive/");
     }
 
+    private void UpdateSelectedOutputDevice(Settings settings)
+    {
+        outputDeviceSelection.Selected = OutputDevices.IndexOf(settings.SelectedOutputDevice.Value ?? "Default");
```
hhyyrylainen, line 1210: Doesn't this actually not work because the text "Default" might not be at position 0 where the translation is forcefully applied there on line 750?

src/general/OptionsMenu.tscn
```diff
@@ -589,11 +590,38 @@ __meta__ = {
...
-[node name="HBoxContainer6" type="HBoxContainer" parent="CenterContainer/VBoxContainer/PanelContainer/Sound/VBoxContainer"]
+[node name="HBoxContainer7" type="HBoxContainer" parent="CenterContainer/VBoxContainer/PanelContainer/Sound/VBoxContainer"]
 margin_top = 348.0
 margin_right = 648.0
 margin_bottom = 370.0
 
+[node name="Label3" type="Label" parent="CenterContainer/VBoxContainer/PanelContainer/Sound/VBoxContainer/HBoxContainer7"]
+margin_right = 66.0
+margin_bottom = 22.0
+text = "DEVICE"
```
hhyyrylainen, line 601: This needs to be way more specific translation key...

hhyyrylainen (changes requested):
Functionality wise this seems to work. I didn't add a Finnish translation so I couldn't like test to try to break this with translations. Code-wise there's quite a few improvements I'd like to see.

src/general/OptionsMenu.cs
FuJa0815: this works because the `OutputDevices` list is not affected by translations at all. Only the `OptionButton`'s list is translated

hhyyrylainen: However, you still apply the default device name translation to item 0, right? Shouldn't that also use the index based lookup, otherwise if the default item is not first (for example in a new Godot version), then the translation is put on the wrong item, right?

FuJa0815: I've read through the source code of godot and it seems like the order is dependent of the audio driver implementations. My best idea to make sure it is on 0 is to search for the default device and put it on index 0 when calling AufioServer.GetDevices

hhyyrylainen: You could do that. 
But I was thinking that you could simply just use the indexOf to find where the default device is and apply the text to that index.

FuJa0815: I actually prefer the sorting solution. This guarantees the user that the Default device is always at the top of the list

hhyyrylainen: Yeah, you can do that, then (if you already did it and I missed the notification about that, sorry, I've gotten so many email from github recently that I haven't read the new commit notifications).

84634E1A607A:
I've got one request. On a linux platform, a selection looks like this:
image
Can this be shortened or beautified?

hhyyrylainen:
It looks the same for me. I don't think we can particularly do much about it, other than request Godot to improve their format, because if for example Godot switches to alsa or pipewire the names likely change, so any kind of regex replace magic we might add will not work forever, so we need to just deal with the quite technical looking names.

FuJa0815:
I think we or godot can't do anything about that, because this is the format that alsa provides (see `aplay -l`). But at least this only affects linux users￼￼. Windows looks better, does it?

src/general/OptionsMenu.cs
```diff
@@ -172,6 +175,10 @@ public class OptionsMenu : ControlWithInput
         .OrderBy(i => i, StringComparer.InvariantCulture)
         .ToList();
 
+    private static readonly List<string> AudioOutputDevicesCache = AudioServer
+        .GetDeviceList().OfType<string>().Where(p => p != Constants.DEFAULT_AUDIO_OUTPUT_DEVICE_NAME)
```
hhyyrylainen, line 179: What does `p` stand for here? I would have personally selected `d` for device in this lambda.

src/general/OptionsMenu.cs
```diff
@@ -1142,12 +1169,21 @@ private void OnCustomUsernameTextChanged(string text)
...
+        UpdateResetSaveButtonState();
+    }
+
     private void OnLanguageSettingSelected(int item)
     {
         Settings.Instance.SelectedLanguage.Value = Languages[item];
         resetLanguageButton.Visible = true;
 
         Settings.Instance.ApplyLanguageSettings();
+        UpdateDefaultAudioOutputDeviceText(audioOutputDeviceSelection);
```
hhyyrylainen, line 1186: There's a bug with this, pressing the reset all settings doesn't apply the new language. Instead of having this call here and in OnResetLanguagePressed, this should be really in _Notification so that it reacts properly to any language change.

hhyyrylainen (changes requested):
Now this is pretty close to perfect for me. I just noticed one bug, I have one question about a variable name and the CI wants one additional line break.

src/general/OptionsMenu.cs
```diff
@@ -732,13 +745,28 @@ private void UpdateResetSaveButtonState()
         saveButton.Disabled = result;
     }
 
-    private void LoadLanguages(OptionButton optionButton)
+    private void UpdateDefaultAudioOutputDeviceText(OptionButton audioOutputDeviceOptionButton)
```
Kasterisk, line 748: Why not just directly access the option button instead of passing it?

Kasterisk:
Works perfectly on Windows, it also displays the audio devices in human readable name.

hhyyrylainen (approved):

src/general/OptionsMenu.cs
hhyyrylainen: Seems like LoadLanguages happens to do the same thing...

Kasterisk: @FuJa0815 Everything else looks good, but it would probably be less confusing if we could change the option buttons to be accessed directly from variables instead of passing them in the methods.

Kasterisk (approved):

hhyyrylainen:
I guess this is now finally only waiting for the translation merge conflicts to be fixed and then this can be merged.

hhyyrylainen:
@FuJa0815 perhaps you also have outdated version of babel-thrive...

locale/en.po
```diff
@@ -1425,234 +1425,242 @@ msgstr ""
...
 
-#: ../src/general/OptionsMenu.tscn:555
+#: ../src/general/OptionsMenu.tscn:556
 msgid "GUI_VOLUME"
 msgstr "GUI Volume"
 
-#: ../src/general/OptionsMenu.tscn:600
+#: ../src/general/OptionsMenu.tscn:601
+msgid "AUDIO_OUTPUT_DEVICE"
+msgstr ""
```
hhyyrylainen, line 1524: Hang on, why is there a blank translation in the english locale?

hhyyrylainen (changes requested):

locale/en.po
FuJa0815: uuh, seems like I forgot to readd it this time? But why is the CI ok with it??

FuJa0815: seems like `handle_po_file` is never executed

hhyyrylainen (approved):

PR #2445, opened 2021-07-27
Dynamic MP calculation

### LINKED ISSUE #1644: Make editor dynamically compute remaining MP
To improve usability the editor should after each action (and undo / redo) recompute the amount of MP remaining by comparing the current modified species properties with what they were when entering the editor.

- [ ] basic implementation
- [ ] new cell button works (this likely needs custom handling)

FuJa0815:

This PR replaces the old logic with a dynamic MP calculation

**Related Issues**

Closes #1644

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


FuJa0815:
Loading is still glitched, fixing...

84634E1A607A (changes requested):
I'm not sure if I've "played" out all problems, but I've already found those.

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -2224,32 +2226,41 @@ private void OnRigidityChanged()
...
 
     /// <summary>
     ///   Perform all actions through this to make undo and redo work
     /// </summary>
-    private void EnqueueAction(MicrobeEditorAction action)
+    /// <returns>True when the action was successful</returns>
+    private bool EnqueueAction(MicrobeEditorAction action)
     {
+        var historyClone = (EditorActionHistory)history.Clone();
+        historyClone.AddAction(action);
```
hhyyrylainen, line 2240: As the new microbe action holds the old organelle list, I think this clone is just as bad as the clone in the previous attempt at dynamic MP was (that cloned a species).

hhyyrylainen (changes requested):
Overall I find this approach to be much more complex than the previous PR on this dynamic MP computation. I think it was a really clean approach to compare the changes to be made to a species with how much MP that would consume. I think this action combining will be harder for new people to maintain and develop the editor further...

src/microbe_stage/editor/MicrobeEditor.cs
84634E1A607A: image
Better that the last one. This can be easily avoided.

FuJa0815:
I think the history should be fixed now. And the clone is gone.

84634E1A607A:
> I think the history should be fixed now. And the clone is gone.

No it isn't. Refer to <https://github.com/Revolutionary-Games/Thrive/pull/2445#discussion_r677936774> , this still causes problem

FuJa0815:
> > I think the history should be fixed now. And the clone is gone.
> 
> 
> 
> No it isn't. Refer to <https://github.com/Revolutionary-Games/Thrive/pull/2445#discussion_r677936774> , this still causes problem
> 
> 
> 
> Happens when you delete an organelle and add back.
> 
> 
> 
> In CombineGuaranteed() you should never modify the two ActionData because they are passed in as *reference*, changing them directly changed history and leads Undo() elsewhere.

The do-action for move also modifies the Organelle position, which is correct, but messes with the history with this system. I'll try to find a solution

84634E1A607A:
> The do-action for move also modifies the Organelle position, which is correct, but messes with the history with this system. I'll try to find a solution

I've given my own solution which you can have a look. It works at least for me.

FuJa0815:
@84634E1A607A The problem is that when I don't update Position the merging code will not work correctly. Checking the mergabillity often relies on the Position.

FuJa0815:
I have found a fix, but personally, I'm not pride of it... It seems like every bug has been fixed, but there has to be a better solution, but I can't find it...

84634E1A607A:
> @84634E1A607A The problem is that when I don't update Position the merging code will not work correctly. Checking the mergabillity often relies on the Position. With your fix the MP calculation fails after moving a new organelle twice

Well, that's true. Sorry that I didn't notice that.

src/microbe_stage/editor/MicrobeEditorActionData.cs
```diff
@@ -0,0 +1,339 @@
...
+
+            return MicrobeActionInterferenceMode.Combinable;
+        }
+
+        return MicrobeActionInterferenceMode.NoInterference;
+    }
+
+    public override int CalculateCost()
+    {
+        return Organelle.Definition.MPCost - (ReplacedCytoplasm?.Sum(p => p.Definition.MPCost) ?? 0);
```
84634E1A607A, line 106: Well, in master, replacing cytoplasm with metabolosomes costs 45, but here it costs 45 - 22 = 23.
Then if you replace 3 cytoplasms with a chloroplast it costs 50 - 3 * 22 = -16...

```suggestion
        return Organelle.Definition.MPCost;
```

FuJa0815: Wait, really? Huh, seems like I've never noticed. But this proposed fix does not work because it would break MP calculation when undoing

src/microbe_stage/editor/MicrobeEditorActionData.cs
```diff
@@ -0,0 +1,345 @@
...
+    {
+        if (other is RigidityChangeActionData rigidityChangeActionData)
+        {
+            if (Math.Abs(NewRigidity - rigidityChangeActionData.PreviousRigidity) < MathUtils.EPSILON &&
+                Math.Abs(rigidityChangeActionData.NewRigidity - PreviousRigidity) < MathUtils.EPSILON)
+                return MicrobeActionInterferenceMode.CancelsOut;
+
+            if (Math.Abs(NewRigidity - rigidityChangeActionData.PreviousRigidity) < MathUtils.EPSILON ||
+                Math.Abs(rigidityChangeActionData.NewRigidity - PreviousRigidity) < MathUtils.EPSILON)
+                return MicrobeActionInterferenceMode.Combinable;
```
adQuid, line 301: Can we safely use if/else here? Or just have it always be combinable, even if to a zero change?

adQuid (changes requested):
It took me a while to figure out what each class is doing, but I like the algorithm. With something so logically complex my questions mostly focus on elegance; I have not yet tested.

src/microbe_stage/editor/MicrobeEditorActionData.cs
```diff
@@ -0,0 +1,333 @@
...
+
+    /// <summary>
+    ///   Combines two actions to one
+    /// </summary>
+    /// <param name="other">The action this should be combined with. Guaranteed to be combinable</param>
+    /// <returns>Returns the combined action</returns>
+    protected virtual MicrobeEditorActionData CombineGuaranteed(MicrobeEditorActionData other)
+    {
+        throw new NotImplementedException();
+    }
```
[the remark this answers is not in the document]
hhyyrylainen: This could be just be marked abstract so then all child classes are required to implement this.

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -1436,6 +1428,22 @@ private void CreateMutatedSpeciesCopy(Species species)
...
+    /// <summary>
+    ///   Calculates the remaining MP from the action history
+    /// </summary>
+    /// <returns>The remaining MP</returns>
+    private int CalculateMutationPointsLeft()
+    {
+        if (FreeBuilding || CheatManager.InfiniteMP)
+            return Constants.BASE_MUTATION_POINTS;
+
+        mutationPointsCache = history.CalculateMutationPointsLeft().Clamp(0, Constants.BASE_MUTATION_POINTS);
```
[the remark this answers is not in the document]
FuJa0815: It doesn't have to be there. I've added it to make sure the result is clamped.

src/microbe_stage/editor/MicrobeEditorActionData.cs
```diff
@@ -0,0 +1,345 @@
...
+    public override MicrobeActionInterferenceMode GetInterferenceModeWith(MicrobeEditorActionData other)
+    {
+        return MicrobeActionInterferenceMode.NoInterference;
+    }
+
+    public override int CalculateCost()
+    {
+        return -Constants.BASE_MUTATION_POINTS;
+    }
+}
```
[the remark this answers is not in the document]
FuJa0815: Because they were before. But own files, maybe even a folder would be better now that the classes pack more weight

src/microbe_stage/editor/MicrobeEditorActionData.cs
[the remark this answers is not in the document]
FuJa0815: Correct, this can be abstract now. In earlier iterations barley any classes had to implement this, therefore I didn't want to make it a requirement, but now most classes implement this

src/microbe_stage/editor/MicrobeEditor.cs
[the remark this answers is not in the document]
adQuid: I would rather fail loudly here. If we end up in a situation where `CalculateMutationPointsLeft()` returns a negative number, something has already gone wrong, and I'd like to know sooner rather than later.

src/microbe_stage/editor/MicrobeEditorActionData.cs
FuJa0815: I'm sorry, I don't fully understand. I've added comments, maybe those are clarifying this code a bit?

adQuid (approved):
This looks good to me. With such complex logic there's more I'd like to think on, but this is something we could go back and forth tweaking for months.

src/microbe_stage/editor/action_data/MoveActionData.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System;
+
+[JSONAlwaysDynamicType]
+public class MoveActionData : MicrobeEditorActionData
+{
+    public OrganelleTemplate Organelle;
+    public Hex OldLocation;
+    public Hex NewLocation;
+    public int OldRotation;
+    public int NewRotation;
```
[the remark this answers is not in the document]
FuJa0815: I can't really use the variables in `OrganelleTemplate` because modifying them breaks the history

Jonathon594:
What's the current state of this? I'm going to look into testing the functionality.

Nunez2196:
just tested this, it feels pretty good. Doesn't horribly break or make my computer scream. Branch needs to be updated but, outside of that i'm not sure what else there is if the code is all written nice.

Jonathon594:
I tested it as well and I couldn't break the MP calculation in the editor. Rigidity slider is perma disabled in this build so how are you handling rigidity slider?

Nunez2196:
> I tested it as well and I couldn't break the MP calculation in the editor. Rigidity slider is perma disabled in this build so how are you handling rigidity slider?

actually i forgot to mention that, i ran into that and figured it was done for some reason

hhyyrylainen:
If the rigidity slider doesn't work on this branch, that's a major issue that needs to be resolved.

84634E1A607A:
I'm still able to crash that. I'll look into what had happened.

src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                switch (cache[compareIndex].GetInterferenceModeWith(cache[compareToIndex]))
+                {
+                    case MicrobeActionInterferenceMode.NoInterference:
+                        break;
+                    case MicrobeActionInterferenceMode.Combinable:
+                        var combinedValue = cache[compareIndex].Combine(cache[compareToIndex]);
+                        cache.RemoveAt(compareIndex);
+                        cache.RemoveAt(compareToIndex);
+                        compareIndex--;
```
84634E1A607A, line 54: compareIndex -= 2 ?

src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -0,0 +1,110 @@
...
+                        break;
+                    case MicrobeActionInterferenceMode.ReplacesOther:
+                        cache.RemoveAt(compareToIndex);
+                        compareIndex = compareToIndex;
+                        copyLength--;
+                        break;
+                    case MicrobeActionInterferenceMode.CancelsOut:
+                        cache.RemoveAt(compareIndex);
+                        cache.RemoveAt(compareToIndex);
+                        compareIndex--;
```
84634E1A607A, line 66: Here too?

84634E1A607A (changes requested):
Don't know if I get this right, but through debugging, I think so.

hhyyrylainen:
I did a quick playtest and for me the rigidity slider seemed to work just fine. I was also pleasantly surprised that this doesn't seem to break save compatibility.

I did notice one warning, which may or may not be related to this PR:
```
WARNING: grab_focus: This control can't grab focus. Use set_focus_mode() to allow a control to get focus.
   At: scene/gui/control.cpp:2185.
```

sirdanforth:
I went through a whole play through and couldn't find any problems pertaining to MP. I checked to make sure that saving and loading worked in and out of the editor which it did with no problems beside having failed save errors. However, I did notice a few bugs that may or may not be related.

1.) I found that after a while of playing (maybe around 30 minutes of play time) the loading screen into the editor takes much longer than usual. It would take it about 2-3 minutes before I finally loaded in.

2.) Groups of enemy cells would just be idle and still. They would only move if I got close to them but would only go in one direction. Although, when I took this screenshot they just stayed still no matter how close I got. I'm not sure if this is a bug or if there's a reason for this behavior but I'm reporting on what I found.




3.) I noticed that there has been a lot of failed saves errors more than usual. I know that it's already a bug reported but it just stuck out to me on how often it failed to save. Also, I'm not sure if these two are connected but it seemed that when the loading screen took longer to load it always saved successfully, but if it loaded quickly the save would fail.

Let me know if this helps.

FuJa0815:
Maybe related? Happened after saving/loading in the editor
```
Unhandled Exception:
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'CustomRichTextLabel'.
  at Godot.Object.GetPtr (Godot.Object instance) [0x0001b] in /root/godot/module
s/mono/glue/GodotSharp/GodotSharp/Core/Object.base.cs:37
  at Godot.RichTextLabel.SetBbcode (System.String text) [0x00001] in /root/godot
/modules/mono/glue/GodotSharp/GodotSharp/Generated/GodotObjects/RichTextLabel.cs
:824
  at Godot.RichTextLabel.set_BbcodeText (System.String value) [0x00001] in /root
/godot/modules/mono/glue/GodotSharp/GodotSharp/Generated/GodotObjects/RichTextLa
bel.cs:111
  at CustomRichTextLabel.ParseCustomTags () [0x00317] in C:\Users\jfuehrer\Docum
ents\Thrive\src\gui_common\CustomRichTextLabel.cs:226
  at Invoke._Process (System.Single delta) [0x0003b] in C:\Users\jfuehrer\Docume
nts\Thrive\src\engine\Invoke.cs:56
```

src/microbe_stage/editor/EditorActionHistory.cs
FuJa0815: nope, check one line below. This also reduces the `compareIndex` after the `Insert` call

84634E1A607A:
Can these MP tooltips be updated? 
image

When I delete this Cytoplasm, I cost "-17" (Delete + Add -> Move in dynamic MP calculation logic) but the tooltip isn't updated.

src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -0,0 +1,110 @@
...
+            }
+        }
+
+        return Constants.BASE_MUTATION_POINTS - cache.Sum(p => p.CalculateCost());
+    }
+
+    public override bool Redo()
+    {
+        var result = base.Redo();
+        cache = GetActionHistorySinceLastNewMicrobePress();
```
84634E1A607A, line 81: Is this necessary? Can we just add an action here?

84634E1A607A (commented):
So far no more bugs found.

src/microbe_stage/editor/EditorActionHistory.cs
FuJa0815: It isn't really possible to just remove the last action because it might already have been reduced down in the `CalculateMutationPointsLeft` method. Adding a "Redo" action might be possible, but this further complicates the already way too complicated code

Kemikal1 (commented):
```
Jukebox now playing from: MicrobeEditor
load finished, success: True message: Load finished elapsed: 00:00:02.1278403
Jukebox: starting track: res://assets/sounds/microbe-editor-theme-4.ogg position: 0
Changing membrane to 'double'
Changing membrane to 'cellulose'
Changing membrane to 'chitin'
Changing membrane to 'silica'

Unhandled Exception:
ICSharpCode.SharpZipLib.ValueOutOfRangeException: Invalid MP amount: -54 out of range
  at MicrobeEditor.CalculateMutationPointsLeft () [0x00063] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:1468
  at MicrobeEditor.get_MutationPoints () [0x00000] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:302
  at MicrobeEditorGUI.UpdateMutationPointsBar () [0x00001] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditorGUI.cs:879
  at MicrobeEditor.DirtyMutationPointsCache () [0x0000d] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:1133
  at MicrobeEditor.OnMembraneChanged () [0x000af] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:2221
  at MicrobeEditor.DoMembraneChangeAction (MicrobeEditorAction action) [0x00041] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:2193
  at MicrobeEditorAction.DoAction () [0x00001] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditorAction.cs:52
  at ReversibleAction.Perform () [0x00016] in C:\Thrive\Thrive\src\general\ReversibleAction.cs:23
  at ActionHistory`1[T].AddAction (T action) [0x00058] in C:\Thrive\Thrive\src\general\ActionHistory.cs:64
  at EditorActionHistory.AddAction (MicrobeEditorAction action) [0x00033] in C:\Thrive\Thrive\src\microbe_stage\editor\EditorActionHistory.cs:113
  at MicrobeEditor.EnqueueAction (MicrobeEditorAction action) [0x00059] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:2275
  at MicrobeEditor.SetMembrane (System.String membraneName) [0x0004a] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:812
  at MicrobeEditorGUI.OnMembraneSelected (System.String membrane) [0x00001] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditorGUI.cs:1320
  at (wrapper managed-to-native) Godot.NativeCalls.godot_icall_2_598(intptr,intptr,string,object[])
  at Godot.Object.EmitSignal (System.String signal, System.Object[] args) [0x00001] in /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Generated/GodotObjects/Object.cs:361
  at MicrobePartSelection.OnSelected () [0x00001] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobePartSelection.cs:154
```
ALSO 
A second kind of crash appears when it crashes after you press the revert button:
```
Changing membrane to 'double'
Changing membrane back to 'single'
Changing membrane to 'double'
Changing membrane to 'single'
Changing membrane back to 'double'
Changing membrane back to 'single'
Changing membrane to 'silica'
Changing membrane to 'cellulose'
Changing membrane to 'chitin'
Changing membrane back to 'cellulose'
Changing membrane back to 'silica'
Changing membrane back to 'single'
Changing membrane to 'calciumCarbonate'
Changing membrane to 'single'
Changing membrane to 'cellulose'
Changing membrane to 'chitin'
Changing membrane to 'silica'

Unhandled Exception:
ICSharpCode.SharpZipLib.ValueOutOfRangeException: Invalid MP amount: -102 out of range
  at MicrobeEditor.CalculateMutationPointsLeft () [0x00063] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:1468
  at MicrobeEditor.get_MutationPoints () [0x00000] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:302
  at MicrobeEditorGUI.UpdateMutationPointsBar () [0x00001] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditorGUI.cs:879
  at MicrobeEditor.DirtyMutationPointsCache () [0x0000d] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:1133
  at MicrobeEditor.OnRigidityChanged () [0x00044] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:2250
  at MicrobeEditor.UndoRigidityChangeAction (MicrobeEditorAction action) [0x0001a] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:2240
  at MicrobeEditorAction.UndoAction () [0x00001] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditorAction.cs:57 
  at ReversibleAction.Undo () [0x00019] in C:\Thrive\Thrive\src\general\ReversibleAction.cs:35
  at ActionHistory`1[T].Undo () [0x0002f] in C:\Thrive\Thrive\src\general\ActionHistory.cs:48
  at EditorActionHistory.Undo () [0x00001] in C:\Thrive\Thrive\src\microbe_stage\editor\EditorActionHistory.cs:101
  at MicrobeEditor.Undo () [0x00010] in C:\Thrive\Thrive\src\microbe_stage\editor\MicrobeEditor.cs:740

```
I dont think the "changed back" should be applicable in the same editor session tho?

I somehow managed to break it, seams the revert action might be the culprit. It very rarely breaks in the organelle editor but it breaks more often with the membrane editor.Theres 2 things which might help you get to the problem:
1.Change the membrane, first thing when you get into teh editor without changing the rigidity then press back. The remaining MP is now 50.Should be 100.
2.IF you fiddle around with the membranes rigidity, changing the membrane types, and revert button, you get a crash simmilar to the ones above.

FuJa0815:
The MP calculation was not  updated correctly for some actions. This should hopefully be fixed now.

84634E1A607A:
Bugs found:
1. Why +10?
  image
2. 55 -> 10 -> 10: Why 10? I replaced a cyto, so move this should cost something?
  image
  image
  image
3. Check "Infinite MP" cheat -> Add something -> Uncheck it -> Crash
  ```
Unhandled Exception:
ICSharpCode.SharpZipLib.ValueOutOfRangeException: Invalid MP amount: -220 out of range
  at MicrobeEditor.CalculateMutationPointsLeft () [0x00063] in /home/ajax/Thrive/src/microbe_stage/editor/MicrobeEditor.cs:1469 
  at MicrobeEditor.get_MutationPoints () [0x00000] in /home/ajax/Thrive/src/microbe_stage/editor/MicrobeEditor.cs:302 
  at MicrobeEditorGUI.UpdateMutationPointsBar () [0x00001] in /home/ajax/Thrive/src/microbe_stage/editor/MicrobeEditorGUI.cs:879 
  at MicrobeEditorGUI.set_MouseHoverHexes (System.Collections.Generic.IEnumerable`1[T] value) [0x0003d] in /home/ajax/Thrive/src/microbe_stage/editor/MicrobeEditorGUI.cs:441 
  at MicrobeEditor.UpdateEditor (System.Single delta) [0x0029c] in /home/ajax/Thrive/src/microbe_stage/editor/MicrobeEditor.cs:1553 
  at MicrobeEditor._Process (System.Single delta) [0x000a8] in /home/ajax/Thrive/src/microbe_stage/editor/MicrobeEditor.cs:613 
  ```
4. https://github.com/Revolutionary-Games/Thrive/pull/2445#pullrequestreview-769347945 Not solved.

FuJa0815:
I'm sorry that this PR became much more complicated than planned, but I think everything should hopefully (:pray:) be working now although I haven't had much time to test all of this in depth.

84634E1A607A:
This is indeed a large one.

More issues found (Unfortunately)

Is this correct?

https://user-images.githubusercontent.com/60734649/135711590-2c618cd5-2ad8-4f0f-904f-226ccbf507f8.mp4

FuJa0815:
> This is indeed a large one.
> 
> More issues found (Unfortunately)
> 
> Is this correct?
> a.mp4
> 
> Should this be 35?

Sorry, but I don't fully see the problem here. It would be really useful if you would:
- show a screenshot of the cell when you enter the editor
- list all the steps (Place cytoplasm at -1,1 (the console shows the coordinates), Move from 2,0 to 1,0, undo, etc...)
- state how much MP you have and how much you should have

84634E1A607A:
The cell 0:00 is the original one, I just kept clicking redo. The 0:04 one is the final one. Steps are as the clip indicates. I'm not quite sure if the final MP should be 35 instead of 45.

FuJa0815:
> This is indeed a large one.
> 
> More issues found (Unfortunately)
> 
> Is this correct?
> a.mp4
> 
> Should this be 35?

I'm pretty sure this is correct because placing the Chloroplast makes the two cytoplasm move operations redundant because they get removed anyways.

Kemikal1:
https://user-images.githubusercontent.com/11059623/135714698-edfa00f3-786c-49d4-a6a9-083f3741d928.mp4

Check this out. -MP achieved

84634E1A607A:
> 1. Why +10?
>    image

So why + here?

And why - then?
image

FuJa0815:
> Check this out. -MP achieved
fixed. Thanks for testing!

>So why + here?
>
>And why - then?
>
>Maybe we should just make "+" means you'll gain MP and "-" means you'll cost.
Currently "+"means this will cost you and minus means you gain MP.

With cytoplasm:
- +10 means the deletion will cost you the normal 10MP
- -22 means that you have placed this cell in this session and therefore deleting this organelle refunds you the price for placing it.

Kemikal1 (approved):
Works good and looks good.

84634E1A607A (approved):
Although I still think that those virtual functions are a bit harder to understand than a whole one, functionality looks good.

84634E1A607A (changes requested):
Bug found:
image
Here MP cost preview says the cost is -5; 
image
But when I place it the cost is actually -10.

*Way to produce*
Original Cell:
image

Removed 3 Cytos (1. Top one; 2. Top-right one; 3. Right one)
image

Place here
image

src/microbe_stage/editor/action_data/MembraneActionData.cs
```diff
@@ -0,0 +1,42 @@
...
+    public override MicrobeActionInterferenceMode GetInterferenceModeWith(MicrobeEditorActionData other)
+    {
+        if (other is MembraneActionData membraneActionData)
+        {
+            // If changed back to the old membrane
+            if (membraneActionData.NewMembrane == OldMembrane && NewMembrane == membraneActionData.OldMembrane)
+                return MicrobeActionInterferenceMode.CancelsOut;
+
+            // If changed membrane twice
+            if (membraneActionData.NewMembrane == OldMembrane || NewMembrane == membraneActionData.OldMembrane)
```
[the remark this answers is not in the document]
FuJa0815: If you do something like
- 0->1
- 1->2
- 2->0
you can go back to an old value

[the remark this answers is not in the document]
84634E1A607A: But if you do 0->1, 1->2 then you get a combined 0->2; Then you do 2->0 it cancels out.

FuJa0815:
Fixed that, hopefully that was it with the bugs :D

src/microbe_stage/editor/action_data/MembraneActionData.cs
```diff
@@ -0,0 +1,42 @@
...
+        return NewMembrane.EditorCost;
+    }
+
+    protected override MicrobeEditorActionData CombineGuaranteed(MicrobeEditorActionData other)
+    {
+        var membraneActionData = (MembraneActionData)other;
+        if (OldMembrane == membraneActionData.NewMembrane)
+            return new MembraneActionData(membraneActionData.OldMembrane, NewMembrane);
+
+        return new MembraneActionData(membraneActionData.NewMembrane, OldMembrane);
```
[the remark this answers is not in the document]
FuJa0815: I'm against merging theese. Not merging them makes this more a "case by case"-check.

84634E1A607A (approved):

project.godot
```diff
@@ -314,6 +314,7 @@ limits/message_queue/max_size_kb=8192
 
 [rendering]
 
+quality/driver/driver_name="GLES2"
```
84634E1A607A, line 317: Change this back pls.

FuJa0815: good find

assets/textures/gui/bevel/buttonDisabled.png.import
```diff
@@ -0,0 +1,34 @@
...
+[params]
+
+compress/mode=0
+compress/lossy_quality=0.7
+compress/hdr_mode=0
+compress/bptc_ldr=0
+compress/normal_map=0
+flags/repeat=0
+flags/filter=true
+flags/mipmaps=false
```
hhyyrylainen, line 24: For GUI images mipmaps have to be on.

src/general/ActionHistory.cs
```diff
@@ -6,18 +6,18 @@
 ///   General implementation of an action history and undo / redo for use by editors
 /// </summary>
 /// <typeparam name="T">Type of actions to hold</typeparam>
-public class ActionHistory<T>
+public abstract class ActionHistory<T>
```
hhyyrylainen: I think instead of dirtying up the action history with sub actions and making this abstract, it'd be better to just make this a normal class, make things virtual, but leave the implementations otherwise as-is.

src/general/ReversibleAction.cs
```diff
@@ -12,6 +12,13 @@ public abstract class ReversibleAction
     [JsonProperty]
     public bool Performed { get; private set; }
 
+    /// <summary>
+    ///   A sub-action is an action that was automatically performed with another action.
+    ///   Sub-actions should not be noticeable by the user.
+    /// </summary>
+    [JsonIgnore]
+    public abstract bool IsSubAction { get; }
```
hhyyrylainen: I kinda disagree with this subaction design (related to my previous point). Instead I think something like action inheritance or composition should be used to make a single action encapsulate more things than it normally would.

src/microbe_stage/OrganelleLayout.cs
```diff
@@ -64,6 +64,14 @@ public Hex CenterOfMass
     /// </summary>
     public T this[int index] => Organelles[index];
 
+    /// <summary>
+    ///   Access organelle by hex
+    /// </summary>
+    public T? GetByHex(Hex hex)
```
hhyyrylainen: Isn't this the same as `GetOrganelleAt`?

src/microbe_stage/OrganelleTemplate.cs
```diff
@@ -13,26 +13,13 @@ public class OrganelleTemplate : IPositionedOrganelle, ICloneable
...
         Definition = definition;
         Position = location;
         Orientation = rotation;
-        NumberOfTimesMoved = numberOfTimesMoved;
     }
 
-    /// <summary>
-    /// Used to flag whether this Organelle was placed during the current editor session.
-    /// </summary>
-    public bool PlacedThisSession { get; set; }
```
hhyyrylainen: Doesn't this flag need to be kept as there is a separate hex colour used for new hexes?

src/microbe_stage/editor/action_data/MicrobeActionInterferenceMode.cs
```diff
@@ -0,0 +1,27 @@
+﻿/// <summary>
+///   Describes how two microbe actions interference with each other.
+/// </summary>
+/// <para>Used for MP calculation</para>
```
hhyyrylainen: Needs to be put in a <remarks> section

src/microbe_stage/editor/action_data/MicrobeEditorActionData.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+
+/// <summary>
+///   This is its own interface to make JSON loading dynamic type more strict
```
hhyyrylainen: This is not an interface, incorrect comment

src/microbe_stage/editor/OrganellePopupMenu.cs
```diff
@@ -66,15 +71,19 @@ public bool ShowPopup
     }
 
     /// <summary>
-    ///   The placed organelle to be shown options of.
+    ///   The main organelle.
+    /// </summary>
+    public OrganelleTemplate? SelectedOrganelle { get; set; }
```
hhyyrylainen: Why is this a separate variable? Why not require the first organelle in the selected list to be the one this returns?

src/microbe_stage/editor/OrganellePopupMenu.cs
```diff
@@ -223,21 +233,41 @@ private void UpdateOrganelleNameLabel()
...
-        selectedOrganelleNameLabel.Text = SelectedOrganelle.Definition.Name;
+        if (SelectedOrganelles == null)
+        {
+            selectedOrganelleNameLabel.Text = null;
+            return;
+        }
+
+        var names = SelectedOrganelles.Select(p => p.Definition.Name).Distinct().ToList();
+
+        if (names.Count == 1)
```
hhyyrylainen: add braces to this if

src/microbe_stage/editor/OrganellePopupMenu.cs
```diff
@@ -114,10 +123,11 @@ public override void _Ready()
         selectedOrganelleNameLabel = GetNode<Label>(SelectedOrganelleNameLabelPath);
         deleteButton = GetNode<Button>(DeleteButtonPath);
         moveButton = GetNode<Button>(MoveButtonPath);
+        microbeEditor = GetNode<MicrobeEditor>(MicrobeEditorPath);
```
hhyyrylainen: Rather than doing this I would 100% more want to see it be so that when this menu is opened the microbe editor needs to be given as a parameter there. This kind of referencing a parent node is not a good design (child object tightly coupled with parent, instead of the parent object injecting the dependency when needed).

src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -1288,7 +1307,7 @@ public void UpdateRigiditySliderState(int mutationPoints)
         if (editor == null)
             throw new InvalidOperationException("GUI not initialized");
 
-        if (mutationPoints >= Constants.MEMBRANE_RIGIDITY_COST_PER_STEP && editor.MovingOrganelle == null)
+        if (mutationPoints >= Constants.MEMBRANE_RIGIDITY_COST_PER_STEP && editor.MovingOrganelles == null)
```
hhyyrylainen: So multiple organelles can be moved at once? How does that work? The way I imagine it in my head is just much more confusing than only allowing to move a single organelle at once...

src/microbe_stage/editor/MicrobeEditorAction.cs
```diff
@@ -27,29 +18,45 @@ public class MicrobeEditorAction : ReversibleAction
     [JsonProperty]
     private readonly MicrobeEditor editor;
 
-    public MicrobeEditorAction(MicrobeEditor editor, int cost,
+    private MicrobeEditorActionData data = null!;
```
hhyyrylainen: Isn't the null forgiving here unnecessary as the constructor initializes this field?
(maybe it should directly write to data if the write through Data is confusing the compiler).

src/general/ActionHistory.cs
FuJa0815: I see your point, but why not make `ActionHistory` abstract? What does an instance of an `ActionHistory` supposed to be?

hhyyrylainen: It could be used for an text editor or another editor in the game. With just specifying a compatible action type.

hhyyrylainen: Actually now that I have had some time to think, why not just leave ActionHistory as is (as a general undo / redo stack) and make a CombinedAction class that allows you to shove multiple Action derived types in it. That way ActionHistory stays general, the actions stay simple, and the cell editor when needed can convert actions into CombinedActions and shove the "sub" steps into them. To me it doesn't make much sense that only partially undoing an action can work, it should be either all or nothing. Or in the rare case that is not true, the actions themselves that are allowed to fail should handle it internally.

FuJa0815: Seems like we both had the same idea, implemented that in the last commits

hhyyrylainen (changes requested):
I've done it! It took two hours but I've now reviewed this entire thing. Overall I think the approach is good and doesn't need major reworks. Mostly just some nullability things and other clearing up the code that is here comments here. I have not tested yet, but once the code is cleaned up and we get some testing to make sure old saves work and there aren't some weird MP calculation bug cases this'll be good to go.

src/microbe_stage/editor/action_data/MoveActionData.cs
```diff
@@ -0,0 +1,78 @@
...
+                return MicrobeActionInterferenceMode.CancelsOut;
+
+            // If this organelle got moved twice
+            if ((moveActionData.NewLocation == OldLocation && moveActionData.NewRotation == OldRotation) ||
+                (NewLocation == moveActionData.OldLocation && NewRotation == moveActionData.OldRotation))
+                return MicrobeActionInterferenceMode.Combinable;
+        }
+
+        // If this organelle got placed in this session
+        if (other is PlacementActionData placementActionData &&
```
hhyyrylainen: There's now really long if's here as well.

src/general/MicrobeActionInterferenceMode.cs
```diff
@@ -0,0 +1,29 @@
+﻿/// <summary>
+///   Describes how two microbe actions interference with each other.
+/// </summary>
+/// <remarks>
+///   <para>Used for MP calculation</para>
+/// </remarks>
+public enum MicrobeActionInterferenceMode
```
hhyyrylainen:
```suggestion
public enum ActionInterferenceMode
```
I think that would be a better name as this is used by the generically named class `CombinableActionData`

src/general/CombinableActionData.cs
```diff
@@ -0,0 +1,36 @@
...
+
+public abstract class CombinableActionData
+{
+    /// <summary>
+    ///   Does this action cancel out with the <paramref name="other"/> action?
+    /// </summary>
+    /// <returns>
+    ///   Returns the interference mode with <paramref name="other"/>
+    /// </returns>
+    /// <para>Do not call with itself</para>
```
hhyyrylainen: Needs to be wrapped in remarks. Also I'd still suggest you implement a body for this method here that checks that this isn't called with self and then calls a protected abstract method.

src/microbe_stage/editor/MicrobeEditorGUI.cs
FuJa0815: The symmetry mode seems to not only get forgotten by the community but also by programmers while implementing moving/removing. I've modified the logic so that the symmetry mode works better. (e.g. moving/removing multiple organelles, placements in symmetry mode is now one action when undoing/redoing, better aborting when not having anough MP while placing with symmetry mode, etc..)

src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -1408,29 +1428,29 @@ public override void _Notification(int what)
         }
     }
 
-    public void ShowOrganelleMenu(OrganelleTemplate selectedOrganelle)
+    public void ShowOrganelleMenu(OrganelleTemplate main, List<OrganelleTemplate?> selectedOrganelles)
```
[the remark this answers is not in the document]
FuJa0815: Removing nulls is good, but this parameter cannot be an `IEnuerable` because this method modifies the list by moving the main organelle to the head of the list.

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -1856,94 +1967,64 @@ private void CopyEditedPropertiesToSpecies(MicrobeSpecies target)
...
+        return organelles
+            .Where(o => o.Definition.InternalName != "cytoplasm")
+            .SelectMany(o => o.Definition.Hexes.Select(hex => hex + o.Position))
+            .Select(hex => editedMicrobeOrganelles.GetOrganelleAt(hex))
+            .Where(o => o?.Definition.InternalName == "cytoplasm")!;
+    }
 
-            default:
+    private IEnumerable<RemoveActionData> GetReplacedCytoplasmRemoveActionData(
+        IEnumerable<OrganelleTemplate> organelles)
```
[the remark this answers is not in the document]
FuJa0815: Barely not

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -2102,63 +2165,43 @@ private void DoOrganelleMoveAction(MicrobeEditorAction action)
...
+        OnMembraneChanged();
     }
 
     /// <summary>
     ///   Finishes an organelle move
     /// </summary>
     /// <returns>True if the organelle move succeeded.</returns>
-    private bool MoveOrganelle(OrganelleTemplate organelle, Hex oldLocation, Hex newLocation, int oldRotation,
-        int newRotation)
+    private bool MoveOrganelle(List<OrganelleTemplate?> organelles, Hex newLocation, int newRotation)
```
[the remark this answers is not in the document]
FuJa0815: No, because later down the call stack the index of the elements is relevant

src/microbe_stage/editor/MicrobeEditorGUI.cs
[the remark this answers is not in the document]
hhyyrylainen: That's very unexpected and unintuitive that the given list would be modified... isn't there some better design that can be used here?

src/microbe_stage/editor/MicrobeEditorGUI.cs
hhyyrylainen: It's not forgotten and for movement I think it's pretty useful to only move one organelle at once, that way you actually can do careful adjustments without having to toggle symmetry mode on and off while working.

src/microbe_stage/editor/MicrobeEditor.cs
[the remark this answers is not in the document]
hhyyrylainen: What are you replying to? If the order is important, the order is still conserved without nulls being inserted to pad things out...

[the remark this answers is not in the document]
hhyyrylainen: If the positions are important then pair of (OrganelleTemplate, Position) should be used as the list's data type.

src/microbe_stage/editor/MicrobeEditorGUI.cs
FuJa0815: I'd say it makes more sense to allow symmetrical movement of organelles with symmetry mode. All hexes symmetric to the hovered mouse hex are also highlighted, therefore to me it makes more sense to apply all actions in symmetry, not only placement. Especially in freebuild mode moving six organelles at a time is much faster than six seperate movements. I guess a third opinion from @Revolutionary-Games/game-design wouldn't hurt.

hhyyrylainen: What about a case where symmetry is hovering over organelles with multiple hexes and some that are just a single one. If you move all at once you end up with a mishmash of different sized organelles to move, and it might be actually impossible to place them all at once.

FuJa0815: It works, you can try it. I have not found a bug with this.

hhyyrylainen: If you really want to ruin this functionality, could we please at least make this an option in the game options to select, and make it the non-default one, if players much prefer the move with symmetry we can make that the default?

src/microbe_stage/editor/MicrobeEditorGUI.cs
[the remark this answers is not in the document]
FuJa0815: I can move the responsibillity to the caller, see https://github.com/Revolutionary-Games/Thrive/pull/2445/commits/bd5378862246a940d72482fc905e0a4243bad39f, do you think this is better?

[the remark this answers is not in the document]
hhyyrylainen: I think it's better. If there will be mutliple places this might be called from it might be better to return the new list from this method so that the caller can then use the return value and put it where it is needed.

src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -1586,28 +1606,37 @@ internal void SendUndoToTutorial(TutorialState tutorial)
...
         UpdateCancelButtonVisibility();
     }
 
     private void OnDeletePressed()
     {
-        editor!.RemoveOrganelle(organelleMenu.SelectedOrganelle.Position);
+        if (organelleMenu.MainOrganelle == null)
+            throw new ArgumentException("No Organelle selected when trying to delete");
+
+        editor!.RemoveOrganelle(organelleMenu.MainOrganelle.Position);
```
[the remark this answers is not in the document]
FuJa0815: I hope this refactoring (https://github.com/Revolutionary-Games/Thrive/pull/2445/commits/9e23eef50961d8dc97bb51f52221030123243d7d) clarifies this.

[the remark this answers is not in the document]
hhyyrylainen: It doesn't actually... I'm still unclear if the symmetry applies to the delete operation and if it does then when is the rule about not deleting the nucleus enforced.

Kemikal1:
Ready for testing? are we going to try to merge this in? its 9 months here in storage

hhyyrylainen:
I don't think it's really possible to get this in for 0.5.8 considering my quite major editor refactoring so fixing merge conflicts would take a long time. If this is a high priority thing this should be fixed up right after 0.5.8 regarding merge conflicts, tested by multiple people, and after those it can be merged as I think the code structure is now about as good as it's going to get.

PR #2449, opened 2021-07-29
Added reset localizations script

FuJa0815:

Just a little script that I wrote. Maybe other people will find a use in it.


84634E1A607A:
As far as I know, .sh files can't be easily opened in Windows. May use .rb instead.

hhyyrylainen:
Yes, all scripts should be ruby in the thrive repo to be crossplatform.

84634E1A607A:
> Yes, all scripts should be ruby in the thrive repo to be crossplatform.

But as for cross platform, why don't we use python instead?

hhyyrylainen:
I dislike python. So all the existing scripts are already ruby. There's already thousands of lines of ruby handling various tasks in the repo.

FuJa0815:
Well, actually, I don't know ruby :D
But, I'll try to rewrite this.

hhyyrylainen:
I recently wrote a very simple cleanup script: https://github.com/Revolutionary-Games/Thrive/blob/master/cleanup.rb

84634E1A607A:
It just doesn't work for me. 
```./scripts/reset_localization.rb:32:in `<main>': undefined method `empty?' for nil:NilClass (NoMethodError)```

**Comment by 84634E1A607A** [comment]
What's more, this script gets really a mess in this PR. I'm not sure where it leads to.
image

**Comment by FuJa0815** [comment]
> It just doesn't work for me.
> `` ./scripts/reset_localization.rb:32:in `<main>': undefined method `empty?' for nil:NilClass (NoMethodError) ``

OS?

**Comment by 84634E1A607A** [comment]
> OS?

Windows 21H1
ruby 3.0.1p64 (2021-04-05 revision 0fb782ee38) [x64-mingw32]

**Comment by FuJa0815** [comment]
You don't have poedit installed, do you?

**Comment by 84634E1A607A** [comment]
No. Not installed.

**Comment by hhyyrylainen** [comment]
Also one more thing, could you run rubocop on this? I have a rubocop config file included in Thrive but I haven't setup CI to run that, so there's no automatic checking with it.

File: scripts/reset_localization.rb [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+# frozen_string_literal: true
+
+require 'os'
+require_relative '../RubySetupSystem/RubyCommon'
+
+def editor
+  return 'poedit' unless which('poedit').nil?
+
+  core_editor = `git config --global core.editor`
+  return core_editor unless core_editor.nil?
```
hhyyrylainen, line 11: Is it possible for the backtick syntax to return nil? I don't think I've seen that happen. I'd use instead something like
```
stuff = `some command`.strip
if stuff == ''
  do stuff...
end
```

scripts/reset_localization.rb
```diff
@@ -0,0 +1,40 @@
...
+current_branch = `git rev-parse --abbrev-ref HEAD`.strip!
+has_changes = system('git diff-index --quiet HEAD --') == false
+runOpen3Checked('git', 'stash') if has_changes
+runOpen3Checked('git', 'checkout', 'master')
+runOpen3Checked('git', 'pull')
+runOpen3Checked('git', 'checkout', current_branch)
+runOpen3Checked('git', 'stash', 'pop') if has_changes
+runOpen3Checked('git', 'checkout', 'master', 'locale/')
+runOpen3Checked('ruby', 'scripts/update_localization.rb')
+system(editor, 'locale/en.po')
```
FuJa0815, line 40: I've used `system` here, because it seems like `runOpen3` does not go directly to the terminal, which breaks `vi`

scripts/reset_localization.rb
```diff
@@ -0,0 +1,40 @@
...
+def editor
+  return 'poedit' unless which('poedit').nil?
+
+  core_editor = `git config --global core.editor`
+  return core_editor.strip unless core_editor.empty?
+
+  visual_editor = `git config --global core.visual`
+  return visual_editor.strip unless visual_editor.empty?
+
+  if OS.windows?
```
FuJa0815, line 16: seems like `OS.win?` breaks on windows?? It complained that it was not found, but on linux it worked? Anyways, this works.

hhyyrylainen: I see, I misremembered the name of the thing to check, it indeed is `windows?`

scripts/reset_localization.rb
hhyyrylainen: It doesn't. I guess that's not immediately obvious. But it (as the name implies) uses the open3 standard ruby part. Which doesn't execute the commands in the current terminal, but instead launches the process with pipes for input and output.

FuJa0815: yeah, and that breaks vi. Using system works fine and we can pretty much guarantee that the execution will work here, so I don't see a problem in using system

hhyyrylainen (approved):
I didn't test this, but this is probably fine.

PR #2455, opened 2021-07-30
Mild refactor of the tooltip system

### LINKED ISSUE #2284: Rename chartLegendPhysConds to chartLegendPhysicalConditions
and remove the warning suppression. I didn't do this in my spelling check enabling PR because these seem to be used from a Godot scene, so this needs a little bit extra care to correctly rename it.

Kasterisk:

This refactors the tooltip system in order to fix some quirky behaviors functionality-wise and code-wise that I found while browsing through tooltip related codes (and some playtesting). List of changes in addition to the ones covered by the commit messages:
- Added signal Disconnect calls on the tooltip signals, this is needed to completely unregister tooltip from a control (due to the actual callback data instance still lingering on for a bit). Forgot to do this in previous tooltip PR after changing `ToolTipCallbackData`'s base class back to Reference.
- Fixed tooltip display timer not resetting on mouse button press when tooltip is not visible.
- Consolidate the tooltip visibility transitioning functionality to make it less error prone and more maintainable, still need further improvements.
- Tweaked the display delay of some editor tooltips.
- Minor general UI and code tweaks.

**Related Issues**

Closes #2284 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -56,7 +67,15 @@ public bool Display
...
-            UpdateToolTipVisibility();
+
+            if (previousToolTip != null)
+                FinalizeToolTipVisibility(previousToolTip, false);
+
+            if (display)
+            {
+                // Set timer
+                displayTimer = MainToolTip.DisplayDelay;
+            }
```
FuJa0815, line 78: is it beneficial to run this code when display was already true?

src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -296,33 +321,56 @@ private void FetchToolTips()
...
-    private void UpdateToolTipVisibility()
+    private void FinalizeToolTipVisibility(ICustomToolTip tooltip, bool visible)
     {
-        // TODO: Fix the current tooltip changing while still fading out
-        // when quickly mousing over multiple closely positioned elements
-        // (Happens when a single tooltip is registered to multiple Controls)
-
-        // Make sure to hide any other tooltips that are still visible
-        HideAllToolTips();
+        if ((tooltip.ToolTipNode.Visible && visible) || (!tooltip.ToolTipNode.Visible && !visible))
```
FuJa0815, line 334:
```suggestion
        if (tooltip.ToolTipNode.Visible == visible)
```

src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -296,33 +321,56 @@ private void FetchToolTips()
...
+                {
+                    GUICommon.Instance.ModulateFadeIn(tooltip.ToolTipNode, Constants.TOOLTIP_FADE_SPEED);
+                }
+                else
+                {
+                    GUICommon.Instance.ModulateFadeOut(tooltip.ToolTipNode, Constants.TOOLTIP_FADE_SPEED);
+                }
+
+                break;
+            }
```
FuJa0815, line 359: use `case ToolTipTransitioning.Fade when visible:` and `case ToolTipTransitioning.Fade when !visible:` here

src/gui_common/tooltip/ToolTipManager.cs
Kasterisk: Yes, consider one of a likely scenario: `MainToolTip` was reassigned to another tooltip from an initial tooltip with `Display` already set true, and assuming the newly assigned tooltip is hidden we could make it visible by setting `Display` to true once more (basically to reset the display timer).

src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -56,7 +67,15 @@ public bool Display
...
             display = value;
-            UpdateToolTipVisibility();
+
+            if (previousToolTip != null)
+                FinalizeToolTipVisibility(previousToolTip, false);
+
+            if (display)
+            {
+                // Set timer
+                displayTimer = MainToolTip.DisplayDelay;
```
FuJa0815, line 77: If I read this correctly setting `Display` to true doesn't mean that the tooltip does get displayed but that the tooltip will be displayed in some time, right? If that's the case, I don't think that `Display` is a very descriptive name anymore

src/microbe_stage/editor/MicrobeEditor.tscn
```diff
@@ -1435,45 +1437,39 @@ SelectionGroup = SubResource( 19 )
 [node name="bindingAgent" parent="MicrobeEditorGUI/CellEditor/LeftPanel/MainPanel/VBoxContainer/MarginContainer/EditingPanel/Structure/ScrollContainer/MarginContainer/PartsSelection/Organelle/Clip/GridContainer" index="1" groups=[
 "PlaceablePartSelectionElement",
 ] instance=ExtResource( 12 )]
-margin_right = 98.0
```
FuJa0815, line 1438: Why did you change the organelle order in the editor? It doesn't change anything ingame, but is this an unintened change?

src/gui_common/tooltip/ToolTipManager.cs
Kasterisk: Well, it still _displays_ the tooltip, if the tooltip has its delay set to zero it will get displayed right away. I think this can be mentioned in the summary that the display time depends on set delay as there's no mention of this currently.

src/microbe_stage/editor/MicrobeEditor.tscn
Kasterisk: Automatic change, will try to see if these are revertable.

Kasterisk: Tried reverting it and seems not possible, Godot is persistent with the changes the second I modified the editor tabs visibility.

hhyyrylainen: Updating the margins, happens pretty often. It seems that very often Godot doesn't realize the changes it wants to make automatically *unless* you restart Godot.

src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -58,8 +58,8 @@ public ICustomToolTip MainToolTip
     }
 
     /// <summary>
-    ///   Displays the current tooltip if set true. It's preferable to set this
-    ///   rather than directly from the tooltip
+    ///   If true displays the current tooltip with a set delay <see cref="ICustomToolTip.DisplayDelay"/>.
```
FuJa0815, line 61:
```suggestion
    ///   If true displays the current tooltip with a set delay of <see cref="ICustomToolTip.DisplayDelay"/>.
```

src/gui_common/tooltip/ICustomToolTip.cs
```diff
@@ -16,57 +16,56 @@ public enum ToolTipPositioning
...
+{
+    /// <summary>
+    ///   Immediately display and hide the tooltip without animation.
+    /// </summary>
+    Immediate,
+
+    /// <summary>
+    ///   Use fading to display and hide the tooltip.
+    /// </summary>
+    Fade,
```
FuJa0815, line 32: Actually, while playtesting I haven't found a tooltip which fades either in this PR or in the master. Did I miss some? Aren't there any? Is this broken?

src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -1,9 +1,9 @@
-﻿using System;
+using System;
```
FuJa0815, line 1: @hhyyrylainen this file has lost the BOM but CI is not complaining.

hhyyrylainen: I noticed that, and a fix for the CI not actually running on the right files is on my branch (it's waiting for the feature to be done before merging): https://github.com/Revolutionary-Games/Thrive/pull/2439

FuJa0815: @Kasterisk
you can add the BOM to the file like [that on unix](https://stackoverflow.com/a/35807131) or like [that on windows with np++](https://stackoverflow.com/a/8432691)

src/gui_common/tooltip/ICustomToolTip.cs
Kasterisk: There currently aren't any.

src/gui_common/tooltip/ICustomToolTip.cs
```diff
@@ -16,57 +16,56 @@ public enum ToolTipPositioning
     FollowMousePosition,
 }
 
+/// <summary>
+///   How a tooltip should transition on display and hide.
```
hhyyrylainen, line 20:
```suggestion
///   How a tooltip should transition on becoming visible and on being hidden.
```

src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -296,33 +321,48 @@ private void FetchToolTips()
...
-            // Set timer
-            displayTimer = MainToolTip.DisplayDelay;
+            case ToolTipTransitioning.Immediate:
+                tooltip.ToolTipNode.Visible = visible;
+                break;
+            case ToolTipTransitioning.Fade when visible:
+                GUICommon.Instance.ModulateFadeIn(tooltip.ToolTipNode, Constants.TOOLTIP_FADE_SPEED);
+                break;
+
+            // ReSharper disable once ConditionIsAlwaysTrueOrFalse
```
hhyyrylainen, line 349: Why's this here?
I'm about to open a PR where the jetbrains version is updated to 2021.2, might that help?

hhyyrylainen: Also first time when I see this `when` syntax. So if the check doesn't pass otherwise how about combining these to a one case with an if inside it?

hhyyrylainen (changes requested):

src/gui_common/tooltip/ToolTipManager.cs
Kasterisk: Just knew about it as well but CI of course just doesn't like it. The if block inside one case is the original form I intended, will just revert to that.

sirdanforth:
I did not find any bugs or errors pertaining to this PR.

src/gui_common/tooltip/ToolTipManager.cs
FuJa0815: Pretty sure the jb complaint was significant. removing the `when !visible` probably should have fixed it. This is because the case above already handles the fade case when visible, so here the when condition is unnecessary.

hhyyrylainen (approved):

PR #2429, opened 2021-07-20
Add translation support for colour picker + tooltip

bieladeka:
**Brief Description of What This PR Does**

Add translation support for colour picker + tooltip

**Related Issues**

closes #2018

**Progress Checklist**

Note: before starting this checklist the issue should be marked as non-draft.

- [x] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [ ] Initial code review passed
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author.

PR #2459, opened 2021-08-02
Refactor ColorPicker logic & theme

### LINKED ISSUE #2018: The Colour Editor Needs Tooltips And Translation Support
**Describe the bug**
The cell colour editor should have tooltips over the colour picker button, the two "on/off" switches and the "+" sign.

**To Reproduce**
1. Go to the Microbe Editor under Appearance.

**Expected behavior**
The colour picker should have a tooltip like "Choose a colour" or "Pick a colour", the "+" sign should have something like "Add a new colour" tooltip and the "on/off" pair of buttons should have a tooltip which explains what they are doing.
On this topic the "RGB", "HSV" and the "on/off" indicators should be translatable.

**Screenshots**
colour picker

**System info (please complete the following information):**
- Thrive version [0.5.3.0]

### LINKED ISSUE #2452: TweakedColorPicker Raw mode maximum value too high
**Describe the bug**

When switched to Raw mode, the maximum value should be 1 not 100; when switching back the current value goes 25000 and maximum 32767 instead of 255. The cell is displayed pure-white.

And A channel is of no use, can it be removed?

**To Reproduce**
1. Go to cell editor -> membrane -> colour
2. Switch to Raw mode

**Screenshots**
image
image
image

### LINKED ISSUE #2456: The "Add current color as preset" tooltip is black
**Describe the bug**

When hovering over the + button on the colour selector the tooltip is barley visible

**Screenshots**
tooltip

84634E1A607A:
Fixes #2452, fixes #2456, fixes #2018
Closes #2429 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
The microbe species made by auto-evo can have alpha in their colour, I think, though, I'm not 100% sure.
Also these customizations should perhaps be in the tweaked colour picker scene to be usable elsewhere as well.

Also please name your PRs and commits with descriptive names.

84634E1A607A:
> made by auto-evo

At least in Editor changing alpha doesn't change anything, and as for those by auto-evo we can't change their color.

> should perhaps be in the tweaked colour picker scene

I think that maybe not every color picker in the game needs this change (and currently color picker is used only once here). I also think that not giving a function to hide / disable that switch is a Godot problem. Maybe we can give Godot a issue?

hhyyrylainen:
> At least in Editor changing alpha doesn't change anything, and as for those by auto-evo we can't change their color.

I was making a point that if the player is prevented from doing something auto-evo can do it is unfair.

84634E1A607A:
> I was making a point that if the player is prevented from doing something auto-evo can do it is unfair.

That's true. Anyway, I think the rendering of the cell discards alpha, so changing this is completely useless. (And if you change the alpha to 0 the cell is expected to be transparent?)

Kasterisk (changes requested):
This is actually a common approach to workaround ColorPicker's limited customizability AFAIK, I think this is fine for now.

>  (And if you change the alpha to 0 the cell is expected to be transparent?)

I don't think that should be allowed currently, having transparent cells will certainly break gameplay (unless we turn it into a feature).

Comments on the changes:

hhyyrylainen:
> I don't think that should be allowed currently, having transparent cells will certainly break gameplay (unless we turn it into a feature).

I think there's an allowed (pretty narrow) alpha range for colours in the mutations code. I think the same range limitation should be placed on the player as well.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,64 @@
...
+
+/// <summary>
+///   Tweaked color picker defines some custom ColorPicker functions.
+/// </summary>
+public class TweakedColourPicker : ColorPicker
+{
+    private CheckButton hsvCheckButton;
+    private CheckButton rawCheckButton;
+
+    public bool HSVCheckButtonDisabled
```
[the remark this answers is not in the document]
84634E1A607A: If edited in the editor, this property gets called before _Ready(), even before _EnterTree(), and the button is automatically re-enabled maybe in EnterTree(). So totally useless.

Thrive.csproj
```diff
@@ -337,4 +338,4 @@
   <ItemGroup>
     <ProjectReference Include="third_party\ThirdParty.csproj" />
   </ItemGroup>
-</Project>
+</Project>
```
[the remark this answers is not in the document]
84634E1A607A: Rider did this, I don't know... I'll try to restore it.

Kasterisk:
I see, I was more thinking of a completely transparent cells which would be really weird to have ingame.

src/gui_common/TweakedColourPicker.cs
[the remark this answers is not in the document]
Kasterisk: That's expected, we have this pattern for custom UI controls where a property value is applied to the control once more on Ready. Following is a common example for this:
```csharp

private field

private Control someNode;

public property
{
    get => field
    set
    {
         // set value...

         UpdateControl()
    }

public override void _Ready()
{
     // retrieve controls...

     // re-update here
     UpdateControl()
}

private void UpdateControl()
{
     // update someNode
}
```

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,64 @@
...
+    // Called when the node enters the scene tree for the first time.
+    public override void _Ready()
+    {
+        try
+        {
+            var baseNode = GetChild(4).GetChild(4);
+            hsvCheckButton = baseNode.GetChild<CheckButton>(0);
+            rawCheckButton = baseNode.GetChild<CheckButton>(1);
+        }
+        catch (Exception e)
```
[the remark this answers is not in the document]
Kasterisk: Alright, well let's just hope the node retrievals are reliable enough.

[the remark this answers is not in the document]
84634E1A607A: I think as long as we still use Godot 3.3 then it's alright.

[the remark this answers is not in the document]
hhyyrylainen: Why are exception catches done here? We have literally hundreds of child node retrievals all over the place and none of them catch exceptions. Why catch exceptions here?
I think this catch should be removed entirely. Even if these are some Godot internal nodes we access, we still have some other accesses like this in the forcing the dialogs to do our bidding.

84634E1A607A:
What happened? Locally check_formatting told me
image
so I ran update_localization and CI told me 
image

hhyyrylainen:
Have you committed all of your changes and pushed them? That's the only thing I can think of. The order of the translation texts depend on where they are defined in the source code the first time, so a difference in source code is what I would investigate first.

84634E1A607A:
Yes. And now after I run update_localization.rb only time is changed.
image

hhyyrylainen:
Are you sure your local pybabel godot is up to date? We updated that version a couple of weeks ago.

84634E1A607A:
What the hell, Line 5 in .po is auto-generated comment!

```po
# Translations template for PROJECT.
# Copyright (C) 2020 ORGANIZATION
# This file is distributed under the same license as the PROJECT project.
# FIRST AUTHOR <EMAIL@ADDRESS>, 2020.
# 
msgid ""
msgstr ""
```

hhyyrylainen:
The .po starting comments include one line that is `# ` though it seems that sometimes the space after `#` is automatically removed. I'm not sure what causes that, so I've just let that be.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,104 @@
...
+    private void OnRawButtonToggled(bool isOn)
+    {
+        if (!isOn && hsvModeDisabled)
+            hsvCheckButton.Disabled = true;
+    }
+
+    private void UpdateControl()
+    {
+        hsvCheckButton.Disabled = hsvModeDisabled;
+        rawCheckButton.Disabled = rawModeDisabled;
```
Kasterisk, line 102: I think it's best/simpler if we just hide the buttons entirely since what we want is just so the player can't interface with the mode toggling. The color mode states is already handled in a certain way internally and I think it's a bit hacky if we force it to fit a custom behavior.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,104 @@
...
+        hsvCheckButton.Connect("toggled", this, nameof(OnHSVButtonToggled));
+        rawCheckButton = baseNode.GetChild<CheckButton>(1);
+        rawCheckButton.Connect("toggled", this, nameof(OnRawButtonToggled));
+
+        addPresetButton = GetChild(7).GetChild<Button>(0);
+        addPresetButton.Connect("mouse_entered", this, nameof(OnMouseEnteredAddPresetButton));
+    }
+
+    [Export]
+    public bool HSVModeDisabled
```
Kasterisk, line 31: These two properties can then just be `ColorModeToggleEnabled` which determines the toggle buttons visibility.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,104 @@
...
+            rawCheckButton.Disabled = true;
+    }
+
+    private void OnRawButtonToggled(bool isOn)
+    {
+        if (!isOn && hsvModeDisabled)
+            hsvCheckButton.Disabled = true;
+    }
+
+    private void UpdateControl()
```
Kasterisk, line 99: `UpdateControl` was an example name, would be clearer if this is instead was named `UpdateToggleButtons`.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    public override void _Ready()
+    {
+        base._Ready();
+        UpdateControl();
+    }
+
+    public void OnMouseEnteredAddPresetButton()
+    {
+        addPresetButton.HintTooltip = TranslationServer.Translate("COLOR_PICKER_ADD_PRESET");
```
Kasterisk, line 84: Can this be set once on setup time or does that will not make the change stick?

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,104 @@
+﻿using Godot;
+
+/// <summary>
+///   Tweaked color picker defines some custom ColorPicker functions.
+/// </summary>
+public class TweakedColourPicker : ColorPicker
+{
+    private readonly CheckButton hsvCheckButton;
+
+    private readonly CheckButton rawCheckButton;
```
Kasterisk, line 10: Variables of same type is preferably shouldn't be separated with whitespaces.

Kasterisk (changes requested):

src/gui_common/TweakedColourPicker.cs
84634E1A607A: If we hide it, the TextBox on the right will fill that place. (Too ugly!!!)

src/gui_common/TweakedColourPicker.cs
84634E1A607A: If changed language in editor, the hint will not change if only set once.

84634E1A607A:
Currently I noticed that there's a hint when hovering on preset: 
image
Which used abbreviations and is not included in translation system. However, it's again Godot logic and I can't find a proper signal to change this.

src/gui_common/TweakedColourPicker.cs
Kasterisk: A necessary trade-off for simplicity :) I actually won't mind that, or maybe you could put a spacer to keep the text box to the right in some way.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: Well, HSV mode doesn't cause issues so can be reserved, while raw mode does cause problem so needs to be removed.

src/gui_common/TweakedColourPicker.cs
Kasterisk: Or alternatively why not let the buttons keep its toggle value while still being disabled? I think it makes more sense because what determines the modes is not necessarily the buttons themselves but rather some internal values in the godot logic. So basically, color modes can't be edited through GUI but is possible programmatically.

84634E1A607A: Okay

84634E1A607A: (How to add a placeholder?

Kasterisk:
Yep it's quite tricky, one of the earlier solution to this was to directly extract from the predefined strings. See [MainMenu.cs#L357](https://github.com/Revolutionary-Games/Thrive/blob/master/src/general/MainMenu.cs#L357). But I think it can't be applied here since there is some string interpolation going on there.

src/gui_common/TweakedColourPicker.cs
Kasterisk: You mean like a translation string placeholder?

84634E1A607A: I just don't know what can act as a spacer

Kasterisk: Ah, you could use a plain Control node for that, and set its size flag to `Fill` and `Expand` horizontally or vertically.

84634E1A607A:
And this tooltip is dynamically loaded so I don't think it will work.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: Solved as 

> Or alternatively why not let the buttons keep their toggle value while still being disabled?

84634E1A607A:
I find that there is still chances that UpdateButtons() still don't work even if I've placed it after base._Ready(). Maybe a whole new color picker is needed?

Kasterisk:
I think it would be much more feasible to reimplement the built-in functionalities as few as possible. Looking at the changes, I take it that the bottom section of color picker (presets, buttons) is the problematic one as they can't be reliably updated? If so then the current approach with custom preset elements is probably sufficient.

84634E1A607A:
I think it is. And I think this PR is now open for review. However, I've found two problems myself, wondering if I should solve them.

The first is that Godot seems don't store preset values, making the preset useless. I wonder if I should add a NodePath based preset memory. But since this involves Options operation (add a preset to local option storage maybe), I don't know how to do.

The second is that adding too many presets now extends the width of the color picker. A simple solution is to limit presets count; a more complex one includes adding a VContainer when the previous one is full and maintain those containers every operation. (What an amount of work)

hhyyrylainen:
Aren't the presets remembered, though, if you exit the editor and come back? I think I saw that they are only cleared if you restart the program.

Kasterisk:
You can of course take inspirations from the original source code for how Godot does things, here is for the color picker: https://github.com/godotengine/godot/blob/3.3/scene/gui/color_picker.cpp. This part `ColorPicker::_update_presets()` is very interesting in relating to the preset container width issue.

84634E1A607A:
> Aren't the presets remembered, though, if you exit the editor and come back? I think I saw that they are only cleared if you restart the program.

I think it isn't, at least in Master.

84634E1A607A:
> You can of course take inspirations from the original source code for how Godot does things

Found GridContainer that automatically does that.

84634E1A607A:
Can be reviewed now. Presets are now stored until program exits.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,310 @@
...
+        pickerButton.HintTooltip = TranslationServer.Translate("COLOR_PICKER_PICK_COLOR");
+        addPresetButton.HintTooltip = TranslationServer.Translate("COLOR_PICKER_ADD_PRESET");
+        TweakedColorPickerPreset.HintTooltipBase = TranslationServer.Translate("COLOR") + ": #{0}\n"
+            + TranslationServer.Translate("LEFT_MOUSE") + ": "
+            + TranslationServer.Translate("COLOR_PICKER_SELECT_PRESET") + "\n"
+            + TranslationServer.Translate("RIGHT_MOUSE") + ": "
+            + TranslationServer.Translate("COLOR_PICKER_DELETE_PRESET");
+    }
+
+    private void AddPresetLocal(Color color)
```
Kasterisk, line 244: Doesn't sound quite obvious to me what `AddPresetLocal` means.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,310 @@
+﻿using System.Collections.Generic;
+using Godot;
+
+/// <summary>
+///   Tweaked color picker defines some custom ColorPicker functions.
+/// </summary>
+public class TweakedColourPicker : ColorPicker
+{
+    private static readonly List<KeyValuePair<string, Color[]>> PresetsStorage
```
Kasterisk, line 9: An explanation for what the dictionary key is/does in the summary could go a long way.

Kasterisk (changes requested):
Reviewed the changes and commented on things that could be improved. An additional note, in some parts of the code it's a bit harder to understand what's going on so it would help a lot if some non-obvious and logic intensive parts be documented more.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,310 @@
...
+
+    private void Translate()
+    {
+        pickerButton.HintTooltip = TranslationServer.Translate("COLOR_PICKER_PICK_COLOR");
+        addPresetButton.HintTooltip = TranslationServer.Translate("COLOR_PICKER_ADD_PRESET");
+        TweakedColorPickerPreset.HintTooltipBase = TranslationServer.Translate("COLOR") + ": #{0}\n"
+            + TranslationServer.Translate("LEFT_MOUSE") + ": "
+            + TranslationServer.Translate("COLOR_PICKER_SELECT_PRESET") + "\n"
+            + TranslationServer.Translate("RIGHT_MOUSE") + ": "
+            + TranslationServer.Translate("COLOR_PICKER_DELETE_PRESET");
```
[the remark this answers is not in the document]
84634E1A607A: > key the actual string templating is then done in the .po

Don't know how to do that...

src/gui_common/TweakedColourPicker.cs
84634E1A607A: Once I thought I may call local add preset logic instead of call a whole logic (without affecting ColorPicker), but later I removed that code.

This has been removed.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,310 @@
...
+            }
+        }
+
+        if (hexColorEdit.Text != color)
+        {
+            hexColorEdit.Text = color;
+            hexColorEdit.CaretPosition = caretPosition - 1;
+        }
+
+        if (color.Length == 6 || color.Length == 8)
```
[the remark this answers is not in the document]
84634E1A607A: We can't. I'll add some comments here.

src/gui_common/TweakedColorPickerPreset.cs
```diff
@@ -0,0 +1,48 @@
...
+
+public class TweakedColorPickerPreset : ColorRect
+{
+    [Signal]
+    public delegate void OnPresetSelected(Color color);
+
+    [Signal]
+    public delegate void OnPresetDeleted(TweakedColorPickerPreset preset);
+
+    public static string HintTooltipBase { get; set; }
```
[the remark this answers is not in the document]
84634E1A607A: For performance reasons. If this is set by its owner, then we don't need to translate in each preset for many times.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,340 @@
...
+        GetChild(4).GetChild<Control>(4).Hide();
+        GetChild<Control>(5).Hide();
+        GetChild<Control>(6).Hide();
+        GetChild<Control>(7).Hide();
+
+        pickerButton = GetChild(1).GetChild<ToolButton>(1);
+        hsvCheckButton = GetNode<CheckButton>("ButtonsContainer/HSVCheckButton");
+        rawCheckButton = GetNode<CheckButton>("ButtonsContainer/RawCheckButton");
+        hexColorEdit = GetNode<LineEdit>("ButtonsContainer/HexColorEdit");
+        hexColorEdit.Text = Color.ToHtml();
```
Kasterisk, line 140: Use the already existing `OnColorChanged` method and pass in the current color. And this should be moved somewhere in the bottom as this particular section is for node retrievals.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    /// <summary>
+    ///   Called when HexColorEdit's text has changed (because of user input / color select). In this function,
+    ///   the input has been confirmed, so if we find something wrong (such as illegal input), we should delete it
+    ///   and set the text to a legal modified one. And because caret position is set to 0 after a '.Text = ""', we
+    ///   need to manually adjust the caret position to avoid strange behavior.
+    /// </summary>
+    /// <param name="color">HexColorEdit's text content</param>
+    private void OnHexColorChanged(string color)
```
Kasterisk, line 256: It'd actually be more intuitive if the color is only applied after the text edit is entered in my opinion (`text_entered`), the default color picker behavior follows the same behavior. This will also allows for simple html validity check approach and no need to manually adjusting the caret, if the color code is invalid we can just force a default value like `00000000` to be applied to the color picker and text edit.

Note that with this apply-on-enter approach, if you focus out of the text edit (selecting other element while editing the textbox), the color may not be applied and kind of just stuck. In order to solve this you need to manually apply the color on focus exit using signal, see [color_picker.cpp#L660](https://github.com/godotengine/godot/blob/3.3/scene/gui/color_picker.cpp#L660).

Kasterisk (changes requested):
The bottom part looks a bit off as there's no spacing between the default elements and the custom one. The color textbox should have size expand flag turned on so there won't be that little gap on the right.

Also the add preset button is a bit thin to press, as for its positioning, I'd prefer it placed below the color presets as in the default color picker. It's gonna look weird if the presets fill the second row.

84634E1A607A:
> The color textbox should have size expand flag turned on so there won't be that little gap on the right.
> 
> It's gonna look weird if the presets fill the second row.

Actually I prefer the current design, because completely appressed color rects look even more strange.
And if you place the add button in a second row, once you added the first preset, it seems that the Add Preset button disappeared.

As for the button width, I think I may need help about how to adjust that.

Kasterisk:
> Actually I prefer the current design, because completely appressed color rects look even more strange.

I'm talking about the textbox, it's rather small horizontally. And setting it to expand automatically would help.

> And if you place the add button in a second row, once you added the first preset, it seems that the Add Preset button disappeared.

That's odd, shouldn't the button move down to make room for the added preset instead?

> As for the button width, I think I may need help about how to adjust that.

Set the button's rect min size to the desired size.

Also, if you don't mind, would you be able to add #2018? Basically adding additional tooltips at least to the custom controls, this PR will then supersedes #2429.

locale/en.po
```diff
@@ -1726,6 +1726,32 @@ msgstr ""
 "version.\n"
 "Please load the save manually through the menu."
 
+#: ../src/gui_common/TweakedColourPicker.cs:197
+msgid "COLOR_PICKER_PICK_COLOR"
+msgstr "Pick a color from window"
```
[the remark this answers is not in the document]
84634E1A607A: Well, that editor actually means "Godot editor" not cell editor... So I removed that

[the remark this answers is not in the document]
Kasterisk: Ah, I see.

84634E1A607A:
> shouldn't the button move down to make room for the added preset instead

Yes, and you can't see it then, not before you scroll down. For many times my first thought is that you can only add a single preset.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: But you can't let any non-hex characters exist there. It's against the edit's usage (as a hex input). Anyway I think that it is necessary to ensure there isn't any illegal characters.

Kasterisk:
Well then that's _not_ really a concrete issue, but one that had more to do with the membrane tab panel height at a particular window size, and as you said scrolling down is possible. Since we are aiming for a general TweakedColorPicker it shouldn't be knitted exclusively to a quirk in the membrane tab.

src/gui_common/TweakedColourPicker.cs
Kasterisk: The text edit still doesn't allow non-hex characters but _only_ after the text edit is entered using <kbd>Enter</kbd>. If there is any illegal characters or the HTML hexadecimal color string is invalid on enter using the validity check, the text in hex input field and applied color is then replaced with the mentioned default fallback value.

84634E1A607A:
I think check_formatting.rb has some cross-platform issues. I didn't pass pre-commit test locally. 

> Code cleanup performed changes, please stage / check them before committing

hhyyrylainen:
> I think check_formatting.rb has some cross-platform issues. I didn't pass pre-commit test locally.

Do you know what those are?
It already skips the LF line ending check on windows.

If that message is printed, then in most cases what you need to do is just to stage all the files again, and try to commit again. That time the checks should pass as it already should have made the changes it wants.

84634E1A607A:
> Do you know what those are?

I don't. After running it, my working tree is still **clean**.

```
Code cleanup performed changes, please stage / check them before committing
PS E:\source\repos\Thrive> git add .
PS E:\source\repos\Thrive> git status
On branch 2452_disable_raw_mode
Your branch is up to date with 'github-desktop-84634e1a607a/2452_disable_raw_mode'.

nothing to commit, working tree clean
PS E:\source\repos\Thrive> git commit
On branch 2452_disable_raw_mode
Your branch is up to date with 'github-desktop-84634e1a607a/2452_disable_raw_mode'.
```

hhyyrylainen:
The formatting detects changes by running `git diff --stat` first, then running the code reformatter, and then running that again. If the results are different, it assumes that the reformatter made changes.

84634E1A607A:
And sometimes pre-commit gives out no issue but running check_formatting.rb independently fails with the given reason.
(the following is what displayed after I ran update_localization.rb to update translation locations)

```
PS E:\source\repos\Thrive> pre-commit
Run check formatting.....................................................Passed
- hook id: check_formatting
- duration: 6.12s

Starting formatting checks with the following checks: ["compile", "files", "inspectcode", "cleanupcode", "duplicatecode", "localization"]
No changes to be checked for .cs files
No changes to be checked for .cs files
No changes to be checked for .cs files
No code format issues found
```

```
PS E:\source\repos\Thrive> ruby .\check_formatting.rb
Starting formatting checks with the following checks: ["compile", "files", "inspectcode", "cleanupcode", "duplicatecode", "localization"]
JetBrains Cleanup Code 2021.1.3
Running on AMD 64 in 64-bit mode, .NET runtime 4.0.30319.42000 under Microsoft Windows NT 10.0.19043.0
JetBrains Duplicates Finder 2021.1.3
JetBrains Inspect Code 2021.1.3
Running on AMD 64 in 64-bit mode, .NET runtime 4.0.30319.42000 under Microsoft Windows NT 10.0.19043.0
Running on AMD 64 in 64-bit mode, .NET runtime 4.0.30319.42000 under Microsoft Windows NT 10.0.19043.0
Using toolset version 16.0 from ��C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin� �
Collecting files to analyze:
  from project 'E:\source\repos\Thrive\Thrive.csproj'.
Using toolset version 16.0 from ��C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin� �
Configuration: Debug, Platform: Any CPU
Using toolset version 16.0 from ��C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin� �

...... (There are too many lines)


Code cleanup performed changes, please stage / check them before committing
```

hhyyrylainen:
That doesn't help, you should run that git diff --stat to see if it persistently says something has changed.
Also you could edit check_formatting.rb line 672 to `puts` the content of the new_diff and old_diff variables.

84634E1A607A:
Well, I've got this strange thing.
pid will never be the same, and...

```
------ Below Old Diff ------
pid 3404 exit 0


------ Above Old Diff ------

------ Below New Diff ------
pid 15640 exit 0
warning: LF will be replaced by CRLF in simulation_parameters/common/auto-evo_parameters.json.
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in simulation_parameters/common/gallery.json.
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in simulation_parameters/common/help_texts.json.
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in simulation_parameters/common/input_options.json.
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in simulation_parameters/common/music_tracks.json.
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in simulation_parameters/microbe_stage/backgrounds.json.
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in simulation_parameters/microbe_stage/bio_processes.json.
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in simulation_parameters/microbe_stage/biomes.json.
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in simulation_parameters/microbe_stage/compounds.json.
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in simulation_parameters/microbe_stage/membranes.json.
The file will have its original line endings in your working directory

------ Above New Diff ------
```

hhyyrylainen:
So there needs to be a preprocessing step for the output to strip out any files containing `The file will have its original line endings`, `warning: .* will be replaced by .* in` and `pid 3404 exit 0` (and also probably blank lines)? Then the comparison should be accurate on Windows.

84634E1A607A:
Maybe you can use `old_diff = runOpen3CaptureOutput 'git', 'diff', '--name-status', '-z'` so that all authentic outputs will be divided by `\u0000` while warnings by `\n`. Then you can split it with `\u0000` and drop the first element if started with 'warning'

hhyyrylainen:
I don't think ignoring *every* warning is a good idea...
I opened an issue: https://github.com/Revolutionary-Games/Thrive/issues/2515

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,345 @@
...
+    /// </summary>
+    /// <param name="color">Current hexColorEditor text</param>
+    private void OnHexColorChanged(string color = null)
+    {
+        color ??= hexColorEdit.Text;
+        if (color.IsValidHtmlColor())
+            Color = new Color(color);
+        else
+            hexColorEdit.Text = Color.ToHtml();
+    }
```
Kasterisk, line 291: We usually use brackets for if else blocks.
```csharp
color ??= hexColorEdit.Text;

if (color.IsValidHtmlColor())
{
    Color = new Color(color);
}
else
{
    hexColorEdit.Text = Color.ToHtml();
}
```

Kasterisk (changes requested):
Overall this is starting to look good, still a few things needs improving.

locale/en.po
```diff
@@ -1726,6 +1726,73 @@ msgstr ""
...
+
+#: ../src/gui_common/TweakedColourPicker.cs:212
+msgid "COLOR_PICKER_ADD_PRESET"
+msgstr "Add current color as a preset"
+
+#: ../src/gui_common/TweakedColourPicker.cs:213
+msgid "COLOR_PICKER_HSV_BUTTON_TOOLTIP"
+msgstr "Switch HSV (Hue, Saturation, Value) mode on or off.\n"
+"Can't be truned on when in Raw mode."
```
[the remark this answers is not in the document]
hhyyrylainen: If "Raw" is capitalized, shouldn't maybe mode be as well? Usually if you have something like that the full name is like "Raw Mode" so the second part of the name is also capitalized.
But maybe "raw" shouldn't be capitalized either? I don't really see the need why it would need to be capitalized, "raw" input is a pretty common term, I think, so it isn't a name that needs to be capitalized.

locale/en.po
```diff
@@ -1726,6 +1726,73 @@ msgstr ""
...
+
+#: ../src/gui_common/TweakedColourPicker.cs:214
+msgid "COLOR_PICKER_RAW_BUTTON_TOOLTIP"
+msgstr "Switch Raw mode on or off.\n"
+"In Raw mode you can make the color\n"
+"R, G, B component values to go beyond 1.0.\n"
```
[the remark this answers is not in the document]
hhyyrylainen: British spelling: `colour`

src/gui_common/TweakedColourPicker.cs
hhyyrylainen: Even though Godot requires many uses of Colour, our variables should be named colour as per the style guide saying to use british spelling.

src/gui_common/TweakedColourPicker.tscn
```diff
@@ -1,443 +1,95 @@
...
+custom_constants/margin_top = 5
+
+[node name="AddPresetButton" type="Button" parent="PresetButtonContainer"]
+margin_top = 5.0
+margin_right = 30.0
+margin_bottom = 37.0
+rect_min_size = Vector2( 30, 0 )
+size_flags_horizontal = 0
+size_flags_vertical = 4
+icon = ExtResource( 3 )
```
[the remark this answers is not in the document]
84634E1A607A: A disabled texture is missing...

src/gui_common/TweakedColourPicker.tscn
```diff
@@ -1,443 +1,95 @@
...
+margin_bottom = 541.0
+custom_constants/margin_top = 5
+
+[node name="ButtonsContainer" type="HBoxContainer" parent="MarginButtonsContainer"]
+margin_top = 5.0
+margin_right = 317.0
+margin_bottom = 47.0
+size_flags_horizontal = 3
+
+[node name="HSVCheckButton" type="CheckButton" parent="MarginButtonsContainer/ButtonsContainer"]
```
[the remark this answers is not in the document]
84634E1A607A: Actually it's because I forgot to use thrive theme on those two buttons

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,345 @@
...
+        hexColorEdit.Text = color.a8 == 255 ? color.ToHtml(false) : color.ToHtml(true);
+    }
+
+    /// <summary>
+    ///   Called when (keyboard) entered in hexColorEdit.
+    ///   Set Color when text is valid; reset if not.
+    ///   Because text_entered signal requires a string parameter but focus_exited zero, wrote lite this.
+    /// </summary>
+    /// <param name="color">Current hexColorEditor text</param>
+    private void OnHexColorChanged(string color = null)
```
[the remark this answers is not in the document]
84634E1A607A: I find that if you click at a container, the LineEdit doesn't lose focus.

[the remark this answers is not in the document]
Kasterisk: It might be due to that there are some extra steps needed to fully defocus the line edit. I looked up Godot's color picker implementation and there's an extra `select(0, 0)` call that I assume is used to completely deselect the line edit on focus exit. Maybe this can be tried out to see if the line edit will behave correctly.

```c++
void ColorPicker::_focus_exit() {
	for (int i = 0; i < 4; i++) {
		if (!values[i]->get_line_edit()->get_menu()->is_visible())
			values[i]->get_line_edit()->select(0, 0);
	}
	c_text->select(0, 0);
}

void ColorPicker::_html_focus_exit() {
	if (c_text->get_menu()->is_visible())
		return;
	_html_entered(c_text->get_text());
	_focus_exit();
}
```
Not sure why they specifically chose [select](https://docs.godotengine.org/en/stable/classes/class_lineedit.html#class-lineedit-method-select) method instead of [deselect](https://docs.godotengine.org/en/stable/classes/class_lineedit.html#class-lineedit-method-deselect) as it seem to be a perfect fit there though. But I think both works exactly the same and for our use `deselect` should be used for its clarity.

src/gui_common/TweakedColourPicker.tscn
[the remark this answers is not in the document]
Kasterisk: Retraced the normal texture, here's the result (hopefully github doesn't compress PNGs):
plusButtonDisabled

src/gui_common/TweakedColourPicker.cs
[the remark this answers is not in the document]
Kasterisk: Also if this works, you may want to connect the text_entered and focus_exited signal to separate callbacks as otherwise it would probably be a bit annoying if the line edit completely deselects every time you press enter.

src/gui_common/TweakedColourPicker.tscn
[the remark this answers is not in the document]
84634E1A607A: Fine.

src/gui_common/TweakedColourPicker.cs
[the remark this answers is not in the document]
84634E1A607A: Well I think that when you click at a container or something it **can't get focus**, so the color picker **can't** lose focus. That's the problem. I'll try using _Input().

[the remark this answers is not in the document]
Kasterisk: I don't entirely understand what the problem here is, so the LineEdit doesn't lose focus when you click Container controls? To my knowledge that's totally expected as those aren't even interactable and they don't grab focus now that reread what you said. And when LineEdit does lose focus to other control, it deselects by itself.

I think the new _Input approach is even more unnecessarily more lines of code now than the one I suggested and I don't see any change with the focus behavior from testing?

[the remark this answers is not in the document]
84634E1A607A: I know that it is easier to just check if the text is valid when exiting, but I do think that it is crazy if you can even key in Chinese or other characters in a hex-only line edit. So anyway I choose to filter the input.
image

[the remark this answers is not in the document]
84634E1A607A: I even think that it is better to just make it read-only. The HTML color edit doesn't worth that much attention. It's too professional for most players.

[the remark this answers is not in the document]
Kasterisk: It can still be useful for some people.

> I know that it is easier to just check if the text is valid when exiting, but I do think that it is crazy if you can even key in Chinese or other characters in a hex-only line edit. So anyway I choose to filter the input.

I was suggesting a more simpler solution but if you think this is a better way then I guess we could try going with this. Though it should be at least as robust as possible and handles edge cases pretty well, as for example I tried testing a bit and looks like there's a bug where if you tried to copy paste a string with non-hex characters to the line edit, it'll crash the game, likely due to a detection failing.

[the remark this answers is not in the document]
84634E1A607A: I find solving that difficult. And I find that if you agree with the input filtering work, then the original way is the simplest way (and it also solves the paste problem too).

By the way, I still don't quite understand what's the problem of defocus.

And if you still prefer the enter way, we can actually add a tooltip saying that the color will only be applied when you <kbd>enter</kbd> to ignore the focus-exit problem (although it's a little irresponsible).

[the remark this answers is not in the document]
Kasterisk: So as I understand it, the focus exit problem is where the line edit doesn't lose focus when you click at a general area outside of the html color edit area unless it's a button or other line edits, etc. and as a result the color can't be applied automatically? This is fine, the player should already by intuition presses <kbd>Enter</kbd> every time they finished inputting a html color code. The automatic color apply if the html edit loses focus is just an additional safeguard. If you look at Godot's default color picker behavior, that's what it does.

I still prefer the enter way as it's the simplest implementation (doesn't require manual input filtering and else), but if you want to go otherwise that's fine.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,393 @@
+﻿using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using Godot;
+
+/// <summary>
+///   Tweaked colour picker defines some custom ColorPicker functions.
```
Kasterisk, line 7:
```suggestion
///   Tweaked colour picker allows for tooltip texts translations and has a customized theming.
```

assets/textures/gui/bevel/plusButtonDisabled.png.import
```diff
@@ -0,0 +1,34 @@
...
+[params]
+
+compress/mode=0
+compress/lossy_quality=0.7
+compress/hdr_mode=0
+compress/bptc_ldr=0
+compress/normal_map=0
+flags/repeat=0
+flags/filter=true
+flags/mipmaps=false
```
Kasterisk, line 24: Mipmaps should be enabled to avoid potential quality loss, most other gui textures have this on.

locale/en.po
```diff
@@ -1726,6 +1726,75 @@ msgstr ""
...
+msgid "HSV"
+msgstr "HSV"
+
+#: ../src/gui_common/TweakedColourPicker.tscn:47
+msgid "RAW"
+msgstr "Raw"
+
+#: ../src/gui_common/TweakedColourPicker.tscn:54
+msgid "#"
+msgstr "#"
```
Kasterisk, line 1796: Could be wrong but I don't think this is necessary be translated.

Kasterisk (changes requested):

locale/en.po
84634E1A607A: I don't want it to be translated, but update_localization.rb does. I don't know how to tell it to ignore that.

Kasterisk: Put `PLACEHOLDER` in the node's editor description so the translation system ignores it. See [documentation](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/working_with_translations.md#working-in-scene-files).

84634E1A607A: Okay. I misunderstood that for setting the string displayed as "PLACEHOLDER" earlier.

84634E1A607A: Okay. I misunderstood that for setting the string displayed as "PLACEHOLDER" earlier.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+    {
+        htmlColourEdit.Text = colour.a8 == 255 ? colour.ToHtml(false) : colour.ToHtml(true);
+    }
+
+    /// <summary>
+    ///   Called when (keyboard) entered in HtmlColourEdit or from OnHtmlColorEditFocusExited.
+    ///   Set Color when text is valid; reset if not.
+    /// </summary>
+    /// <param name="colour">Current htmlColourEditor text</param>
+    private void OnHtmlColourChanged(string colour)
```
Kasterisk, line 284: To reflect that this called on text enter, should be named `OnHtmlColourEntered`.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+        else
+        {
+            htmlColourEdit.Text = Color.ToHtml();
+        }
+    }
+
+    /// <summary>
+    ///   Called when focus exited HtmlColorEdit.
+    /// </summary>
+    private void OnHtmlColorEditFocusExited()
```
Kasterisk, line 299: Hhyyrylainen wouldn't like this name :)
```suggestion
    private void OnHtmlColourEditFocusExited()
```

src/gui_common/TweakedColourPicker.tscn
```diff
@@ -1,443 +1,111 @@
...
+texture_normal = ExtResource( 3 )
+texture_pressed = ExtResource( 4 )
+texture_hover = ExtResource( 4 )
+texture_disabled = ExtResource( 5 )
+expand = true
+
+[connection signal="color_changed" from="." to="." method="OnColourChanged"]
+[connection signal="toggled" from="MarginButtonsContainer/ButtonsContainer/HSVCheckButton" to="." method="OnHSVButtonToggled"]
+[connection signal="toggled" from="MarginButtonsContainer/ButtonsContainer/RawCheckButton" to="." method="OnRawButtonToggled"]
+[connection signal="focus_exited" from="MarginButtonsContainer/ButtonsContainer/HtmlColourEdit" to="." method="OnHtmlColorEditFocusExited" binds= [ "-" ]]
```
Kasterisk, line 109: There's still binds set to the signal, should be removed as otherwise the signal can't connect to the method.

84634E1A607A:
Found that if you want to keep the two buttons' widths the same, the text on the button can't be translated.

On the one hand, HSV is a abbreviation so can't be easily translated (in China we also use HSV or maybe HSV颜色模型 but basically the same), on the other hand, "raw mode" is a Godot-only color mode, I can't even find a Chinese translation. So keep them and place all the explanations in the tooltip is a better way.

Kasterisk (changes requested):
Some nitpicks and one critical issue, Everything else looks quite fine to me.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: Hard to change language habit...

84634E1A607A:
> One critical issue

So it proves that I shouldn't have coded while attending net-class (

src/gui_common/TweakedColourPicker.tscn
```diff
@@ -1,443 +1,115 @@
...
+[node name="HtmlColourEdit" type="LineEdit" parent="MarginButtonsContainer/ButtonsContainer"]
+margin_left = 227.0
+margin_right = 299.0
+margin_bottom = 42.0
+size_flags_horizontal = 3
+theme = ExtResource( 1 )
+align = 1
+max_length = 8
+context_menu_enabled = false
+shortcut_keys_enabled = false
```
Kasterisk, line 77: Should be enabled so copy pasting is possible.

Kasterisk (commented):
I guess this is now good enough, works fine in-game. Just one more thing that I overlooked.

Kasterisk (approved):

src/gui_common/thrive_theme.tres
```diff
@@ -363,7 +365,7 @@ corner_radius_bottom_left = 10
 expand_margin_top = 20.0
 
 [resource]
-default_font = ExtResource( 8 )
+default_font = ExtResource( 9 )
```
hhyyrylainen, line 368: Did @Kasterisk approve on changing the *default* font to a different one? I think this may have pretty wide consequences...

src/gui_common/TweakedColourPicker.tscn
```diff
@@ -1,443 +1,114 @@
...
+margin_left = 105.0
+margin_right = 207.0
+margin_bottom = 42.0
+theme = ExtResource( 1 )
+shortcut_in_tooltip = false
+action_mode = 0
+keep_pressed_outside = true
+text = "Raw"
+__meta__ = {
+"_editor_description_": "PLACEHOLDER"
```
hhyyrylainen, line 55: Are these the actual buttons? Shouldn't these be translatable. I would bet that there is at least one language where it makes sense to translate these.

hhyyrylainen: Also using "raw" to denote a special kind of input, colour value, photo mode, is actually very common, and languages might have a convention on those.

src/gui_common/thrive_theme.tres
84634E1A607A: That may be a mistake.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+public class TweakedColourPicker : ColorPicker
+{
+    /// <summary>
+    ///   This is where presets are stored after a colour picker exited scene tree.
+    /// <remarks>
+    ///   KeyValuePair&lt;GetPath(), GetPresets()&gt;:
+    ///   <ul>
+    ///     <li> First parameter (string) is the string format of NodePath; </li>
+    ///     <li> Second parameter (Color[]) is the preset colours to be stored. </li>
+    ///   </ul>
```
hhyyrylainen, line 18: remarks should be wrapped in a <para> in addition to being inside the remarks.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+///   Tweaked colour picker allows for tooltip texts translations and has a customized theming.
+/// </summary>
+public class TweakedColourPicker : ColorPicker
+{
+    /// <summary>
+    ///   This is where presets are stored after a colour picker exited scene tree.
+    /// <remarks>
+    ///   KeyValuePair&lt;GetPath(), GetPresets()&gt;:
+    ///   <ul>
+    ///     <li> First parameter (string) is the string format of NodePath; </li>
```
hhyyrylainen, line 16: This comment line confuses me, I can't decipher what it is trying to say.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+    ///     <li> Second parameter (Color[]) is the preset colours to be stored. </li>
+    ///   </ul>
+    /// </remarks>
+    /// </summary>
+    private static readonly List<KeyValuePair<string, Color[]>> PresetsStorage
```
hhyyrylainen, line 21: Why is this a list and not a dictionary? The value is already a list of colours, so why not make this a dictionary as values with the same string could be combined to a single list of colours in a dictionary.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+        get => rawButtonEnabled;
+        set
+        {
+            rawButtonEnabled = value;
+            UpdateButtonsState();
+        }
+    }
+
+    [Export]
+    public new bool HsvMode
```
hhyyrylainen, line 72: Shouldn't this be named `HSVMode` to be consistent with `HSVButtonEnabled`?

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+            if (value && HsvMode)
+                return;
+
+            base.RawMode = value;
+            UpdateButtonsState();
+        }
+    }
+
+    [Export]
+    public new bool PresetsEnabled
```
hhyyrylainen, line 103: So this is about being able to add new presets? As there's separately a presets visible property. If so I think this should maybe be named `PresetsEditable`

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+        base._Ready();
+
+        // Hide replaced native controls. Can't delete them because it will crash Godot.
+        GetChild(4).GetChild<Control>(4).Hide();
+        GetChild<Control>(5).Hide();
+        GetChild<Control>(6).Hide();
+        GetChild<Control>(7).Hide();
+
+        // Get controls
+        var baseControl = GetChild(4);
```
hhyyrylainen, line 144: This was already fetched on line 138, so this variable could be moved to line 136 to avoid that duplicate lookup.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+        addPresetButton = GetNode<TextureButton>("PresetButtonContainer/AddPresetButton");
+
+        // Update control state.
+        UpdateButtonsState();
+        PresetsEnabled = presetsEnabled;
+        PresetsVisible = presetsVisible;
+        OnColourChanged(Color);
+
+        // Load presets.
+        if (PresetsStorage.Exists(p => p.Key == GetPath()))
```
hhyyrylainen, line 164: Wouldn't it make sense if *all* of the colour pickers shared their presets? That way for example in the future the player could make their spaceship the same colour as a building.
Or if you want to keep the scoped presets, there could be a "PresetGroup" property that is by default `"default"` that is used as the key here.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+        {
+            var presetsStored = PresetsStorage.First(p => p.Key == GetPath());
+            foreach (var colour in presetsStored.Value)
+                AddPreset(colour);
+            PresetsStorage.Remove(presetsStored);
```
hhyyrylainen, line 169: I think the presets shouldn't be cleared. Only overwritten when this exits properly. This would guard against exiting in a weird way that loses the presets.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+        // Update control state.
+        UpdateButtonsState();
+        PresetsEnabled = presetsEnabled;
+        PresetsVisible = presetsVisible;
+        OnColourChanged(Color);
+
+        // Load presets.
+        if (PresetsStorage.Exists(p => p.Key == GetPath()))
+        {
+            var presetsStored = PresetsStorage.First(p => p.Key == GetPath());
```
hhyyrylainen, line 166: Using Exists and then First seems to be useless, much better to use FirstOrDefault?
Though, if the presets is changed to a dictionary this can be replaced with a try get value

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    private void OnRawButtonToggled(bool isOn)
+    {
+        RawMode = isOn;
+    }
+
+    private void OnColourChanged(Color colour)
+    {
+        htmlColourEdit.Text = colour.a8 == 255 ? colour.ToHtml(false) : colour.ToHtml(true);
```
hhyyrylainen, line 276: Comment explaining why the comparison to 255 is needed would be nice

locale/en.po
```diff
@@ -2900,7 +2961,8 @@ msgstr "Selected save is upgradeable"
 #: ../src/saving/SaveList.tscn:192
 msgid "SAVE_IS_UPGRADEABLE_DESCRIPTION"
 msgstr ""
-"The selected save is from an older version of Thrive, but can be upgraded.\n"
+"The selected save is from an older version of Thrive, but can be "
+"upgraded.\n"
```
hhyyrylainen, line 2965: This seems like an invalid change, we use 77 columns wide text, and this is not near that, so if you could revert this, that would be nice.

hhyyrylainen (changes requested):
Here's some hopefully pretty quick to fix code style feedback.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: This overrides Godot property HsvMode to avoid direct change to Godot's properties. So I have to name like this.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: This also overrides a Godot property. Can't rename it because we can't rename Godot's properties or disable them.

At most we an add a Obsolete Attribute and rename the functional one to our own name, but we still need to override this.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: Added an else actually.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: Removed for consistency. (Other places I just give it a colour.ToHtml(true) without checking if a8 equals 255)

src/gui_common/thrive_theme.tres
Kasterisk: Didn't saw that one... Well, thanks for noticing.

locale/en.po
84634E1A607A: The last character is on column 79...

hhyyrylainen: I suppose if you count the " and \n part then that is true...

src/gui_common/TweakedColourPicker.cs
hhyyrylainen: Well maybe the other property that can be renamed, should be? To improve consistency.

src/gui_common/TweakedColourPicker.cs
hhyyrylainen: I think it's fine to keep then. Could add a comment explaining with remark why this is named like this.

84634E1A607A:
I think that in case there's more than one picker in tree, I'll try add delegates to sync grouped pickers' presets. This may take some time.

84634E1A607A:
Now we need a way to store these presets in options after the game exits.

84634E1A607A:
Another thing is that we should adopt proper method to show transparency. Now if you set the alpha channel to 0, you'll get a "blank" preset.

hhyyrylainen:
> Now we need a way to store these presets in options after the game exits, but I really don't understand that part of logic. However, I think it will be easy for one who knows. (Maybe we can add a issue and solve that afterwards.)

That can be left for later.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,404 @@
...
+        {
+            rawButtonEnabled = value;
+            UpdateButtonsState();
+        }
+    }
+
+    /// <summary>
+    ///   Change the picker's HSV mode.
+    ///   <remarks>
+    ///     This is named not HSVMode because this hides a Godot property to avoid break custom functions.
```
hhyyrylainen, line 80:
```suggestion
    ///     This is not named HSVMode because this hides a Godot property to avoid breaking custom functions.
```
Even with that grammar fix I'm still not sure what "avoid breaking custom functions" means here?

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,404 @@
...
+        get => hsvButtonEnabled;
+        set
+        {
+            hsvButtonEnabled = value;
+            UpdateButtonsState();
+        }
+    }
+
+    /// <summary>
+    ///   Decide if user can toggle Raw CheckButton to switch Raw mode.
```
hhyyrylainen, line 64:
```suggestion
    ///   Decide if user can toggle raw CheckButton to switch raw mode.
```

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,404 @@
...
+
+            base.RawMode = value;
+            UpdateButtonsState();
+        }
+    }
+
+    /// <summary>
+    ///   Decide if user can edit the presets.
+    ///   <remarks>
+    ///     This is not named to PresetsEditable because it also hides a Godot property.
```
hhyyrylainen, line 120:
```suggestion
    ///     This is not named to PresetsEditable because this also hides a Godot property.
```

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,404 @@
...
+
+    public new void AddPreset(Color colour)
+    {
+        if (groupStorage.Colours.Contains(colour))
+            return;
+
+        groupStorage.Colours.Add(colour);
+
+        // Broadcast to all group numbers.
+        groupStorage.AddPreset(colour);
```
hhyyrylainen, line 231: Instead of having these be separate. I think there should be one method in groupStorage that does this both. That way someone doesn't modify Colours accidentally without also broadcasting the change. I'll comment more in that class.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,404 @@
...
+        private readonly TweakedColourPicker owner;
+
+        public TweakedColourPickerPreset(TweakedColourPicker owner, Color colour)
+        {
+            this.owner = owner;
+            Color = colour;
+            MarginTop = MarginBottom = MarginLeft = MarginRight = 6.0f;
+            RectMinSize = new Vector2(20, 20);
+            SizeFlagsHorizontal = 8;
+            SizeFlagsVertical = 4;
```
hhyyrylainen, line 356: Aren't there named enums for these?
Looks like there are, `SizeFlags.ShrinkCenter` is a thing that exists. So those should be used instead of magic numbers.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,404 @@
...
+                switch ((ButtonList)mouseEvent.ButtonIndex)
+                {
+                    case ButtonList.Left:
+                        owner.Color = Color;
+                        break;
+                    case ButtonList.Right:
+                        if (!owner.PresetsEnabled)
+                            break;
+
+                        owner.ErasePreset(Color);
```
hhyyrylainen, line 381: Shouldn't these call the Godot consume input `GetTree().SetInputAsHandled()` and return because these are active actions that are performed? The mouse input shouldn't be able to cause any further actions, that would be a bit sily.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,404 @@
...
+        }
+
+        private void UpdateTooltip()
+        {
+            HintTooltip = string.Format(CultureInfo.CurrentCulture,
+                TranslationServer.Translate("COLOUR_PICKER_PRESET_TOOLTIP"), Color.ToHtml());
+        }
+    }
+
+    private class PresetGroupStorage
```
hhyyrylainen, line 396: So for this class I'd make this implement IEnumerable<Color>, making the Colours a private list and adding two methods: Add and Remove for the colour, which would handle the Colours list, and also notifying the AddPreset and ErasePreset callbacks.

Also one more thing: usually callback calls look like `AddPreset?.Invoke(...)` why that form is not used in this class?

hhyyrylainen (changes requested):
A few code comments, it's mostly fine now, mostly just spelling and one minor architecture tweak.
I did some playtesting now and have some bug findings:
The html value for the colour doesn't fit at the default resolution. That needs fixing. Perhaps putting the `#` character in the input entry would fix that space problem? or using a slightly smaller font? The colour parsing can then ignore the leading # character.

Another problem: when clicking on a preset or using the html entry, the colour of the preview cell doesn't change:



Also noticed possibly unrelated bug (maybe @Kasterisk knows what's up), the tooltip is under some buttons not over them like it should be:

Kasterisk:
> The html value for the colour doesn't fit at the default resolution. That needs fixing. Perhaps putting the # character in the input entry would fix that space problem? or using a slightly smaller font? The colour parsing can then ignore the leading # character.

I think that would over-complicate things, I prefer just using the slightly smaller font solution or maybe the # can be removed entirely (though that'll not make it clear what the html value line edit is supposed to be).

> Also noticed possibly unrelated bug (maybe @Kasterisk knows what's up), the tooltip is under some buttons not over them like it should be

I also noticed that with the built-in color picker, not sure what's the problem since the tooltips used are Godot's own tooltips. Of course a possible fix could always be switching to using the custom tooltip system with the tradeoff taking a bit more effort to setup.

src/gui_common/TweakedColourPicker.cs
Kasterisk: Wouldn't [ObservableCollection](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.observablecollection-1?view=net-5.0) be a perfect fit for this?

hhyyrylainen: Probably yes, I'd say.

hhyyrylainen:
> I think that would over-complicate things, I prefer just using the slightly smaller font solution or maybe the # can be removed entirely (though that'll not make it clear what the html value line edit is supposed to be).

To me the much bigger and different looking "#" in front of the number, didn't immediately scream as to what it is. People like me are more likely to realize they belong together if the "#" is actually at the start of the text box.

> I also noticed that with the built-in color picker, not sure what's the problem since the tooltips used are Godot's own tooltips. Of course a possible fix could always be switching to using the custom tooltip system with the tradeoff taking a bit more effort to setup.

Ah so they are the inbuilt tooltips where this is a problem?
Would be nice to open an issue to track this.

Kasterisk:
> To me the much bigger and different looking "#" in front of the number, didn't immediately scream as to what it is. People like me are more likely to realize they belong together if the "#" is actually at the start of the text box.

Not sure if there's an easy way (at least not that I know of) to put a one read-only character at the start of a line edit, closest I could think of is probably a placeholder text that's visible when nothing is inputted to the field. So maybe there'll be some fiddling to get it to work.

> Ah so they are the inbuilt tooltips where this is a problem?
Would be nice to open an issue to track this.

Yeah, I'll open an issue.

hhyyrylainen:
> Not sure if there's an easy way (at least not that I know of) to put a one read-only character at the start of a line edit, closest I could think of is probably a placeholder text that's visible when nothing is inputted to the field. So maybe there'll be some fiddling to get it to work.

It won't be a problem if the presence of "#" doesn't change the parsing. So it is just there when set from the code, but the user can not type it in, or even remove it when typing in a new value.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: Well, we have custom buttons. If directly access Godot HsvMode, custom buttons will not be updated.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: I'll try (and learn).

src/gui_common/TweakedColourPicker.cs
hhyyrylainen: Could you try to explain the problem in the form of a comment?

84634E1A607A:
Or we can use CheckBox instead of CheckButton, which will give out a lot of place.

image

image

hhyyrylainen:
Maybe that's a fine tradeoff? @Kasterisk ? At least the text in the textbox should be left-aligned to make it closer to the # to visually associate it with that more.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,473 @@
...
+                        GetTree().SetInputAsHandled();
+                        owner.Color = Color;
+                        break;
+                    case ButtonList.Right:
+                        if (!owner.PresetsEditable)
+                            break;
+
+                        GetTree().SetInputAsHandled();
+                        owner.ErasePreset(Color);
+                        break;
```
hhyyrylainen, line 419:
```suggestion
                    case ButtonList.Left:
                    {
                        GetTree().SetInputAsHandled();
                        owner.Color = Color;
                        return;
                    }

                    case ButtonList.Right:
                    {
                        if (!owner.PresetsEditable)
                            break;

                        GetTree().SetInputAsHandled();
                        owner.ErasePreset(Color);
                        return;
                    }
```

src/gui_common/TweakedColourPicker.cs
```diff
@@ -0,0 +1,473 @@
...
+        public DeletePresetDelegate ErasePresetDelegate { get; set; }
+
+        public void AddPreset(Color colour)
+        {
+            if (colours.Contains(colour))
+                return;
+
+            colours.Add(colour);
+
+            // ?. is no used because we can make sure that at least one delegate is added when this get called.
```
hhyyrylainen, line 453: I would still prever to use the "proper" way here.

84634E1A607A:
Here are more designs:

image

hhyyrylainen (changes requested):
This time I didn't find any gameplay bugs.
2 small code comments.
And I'd like to see that text box sizing thing (that we started discussing) resolved.

hhyyrylainen:
I don't think the slider fits, because it is not a slider. A dropdown would be suitable here, but I think maybe 2 *radio* buttons (instead of check boxes) might be the best control to use here.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: Okay

84634E1A607A:
> maybe 2 _radio_ buttons

Then we need three (HSV, Raw and RGB)

hhyyrylainen:
> Then we need three (HSV, Raw and RGB)
Ah yeah that's true. Then maybe the checkboxes to give more space for the text box is the best design we have thought of so far.

Kasterisk:
> Maybe that's a fine tradeoff? @Kasterisk ? At least the text in the textbox should be left-aligned to make it closer to the # to visually associate it with that more.

Yeah I'm fine with it. I also think checkboxes design is the best one to use to keep the current toggle behavior while freeing up more space.

Kasterisk:
Only one more thing I noticed is the html color edit on raw mode, a bit confusing at first why it didn't update but obviously it's because _raw mode_ is on. The default color picker automatically hides the textbox if any of the slider values are changed, I'm fine if you also want to make the custom one hide or just leave it visible.

Also just one minor thing, maybe the add preset button's size can be reduced a bit to 25x25.

src/general/MainMenu.tscn
```diff
@@ -365,6 +366,12 @@ __meta__ = {
 "_edit_use_anchors_": false
 }
 
+[node name="TweakedColourPicker" parent="." instance=ExtResource( 11 )]
```
Kasterisk, line 369: Debug leftover

src/gui_common/TweakedColourPicker.cs
```diff
@@ -378,6 +401,21 @@ private void OnRawButtonToggled(bool isOn)
...
+            {
+                htmlColourStart.Visible = false;
+                htmlColourEdit.Visible = false;
+            }
+            else
+            {
+                htmlColourStart.Visible = true;
+                htmlColourEdit.Visible = true;
+            }
+        }
```
Kasterisk, line 417:
```c#
htmlColourStart.Visible = !(mode == PickerMode.Raw && colour.IsRaw());
htmlColourEdit.Visible = !(mode == PickerMode.Raw && colour.IsRaw());
```

Kasterisk: Would this be logically correct?

84634E1A607A: htmlColourStart.Visible = htmlColourEdit.Visible = !(mode == PickerMode.Raw && colour.IsRaw());

84634E1A607A: I think it's right.

84634E1A607A: And also solves a potential bug

Kasterisk (approved):

hhyyrylainen (approved):

PR #2461, opened 2021-08-04
Corrected rotation in colonies

### LINKED ISSUE #2343: Rotation is wrong when you have a level 3 member in a colony
Imagine the following colony where A is the colony member:
(A)-(B) (C)

When the colony now starts binding with cell C via B, the rotation of C gets messed up and cell C might get attached to B on the wrong position.
Using `GlobalTransform` in `Microbe.OnIGotAddedToColony` when acessing the parent might fix it.

FuJa0815:

This PR fixes the broken rotation when you have a "level 3" colony member

**Related Issues**

Closes #2343

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


FuJa0815:
That should make it clearer

adQuid:
> That should make it clearer

Still confused WHY all these rotations are happening. It seems like it's trying to point vectors from the center of the child and parent to the respective membranes. Is this to determine offset to have them stick membrane to membrane? What was the problem before, and what exactly is a "level 3" colony member?

FuJa0815:
Pretty much. The problem is that collision can happen before the microbes touch visually. The ExternalOrganelle method was used for knowing where to attach external organelles to, which I have reused for the binding code. The problem is that the ExternalOrganelle assumes that the organelle is not rotated, thats why I rotate the vectors opposite to the microbe rotation to cancel out the microbe rotation. I need to rotate back the resulting vector. I call the ExternalOrganelle method twice, once to get the vector from this microbe towards the parent microbe and once to get the vector from the parent towards me stopping at the membrane. I sum those vectors up (negating one of them) and have a vector which describes the wanted offset to the parent. Due tue the rotation of the parent, which automatically gets applied to the new children, I need to cancel out this rotation too, which leads to another rotation call.

FuJa0815:
A "level 3" member is when you aren't connected to the master directly but though a proxy-cell. See issue.

src/microbe_stage/Microbe.cs
```diff
@@ -1427,22 +1427,39 @@ private void OnIGotAddedToColony()
...
+        // This vector gets as if I had no rotation.
+        // This is important, because GetExternalOrganelle only works with non-rotated microbes
+        var vectorToParentRotated = vectorFromParent.Rotated(Vector3.Up, Rotation.y);
 
-        vectorToParentRotated = (-vectorToParent).Rotated(Vector3.Down, ColonyParent.Rotation.y);
-        var parentVectorToItsMembrane =
-            ColonyParent.Membrane.GetExternalOrganelle(vectorToParentRotated.x, vectorToParentRotated.y);
+        // This vector gets rotated as if the parent had no rotation.
+        var vectorFromParentRotated =
+            vectorFromParent.Rotated(Vector3.Down, globalParentRotation.y);
```
adQuid, line 1443: why does this rotate on down, while the other one rotates on up?

src/microbe_stage/Microbe.cs
```diff
@@ -1427,22 +1427,39 @@ private void OnIGotAddedToColony()
...
 
-        var requiredDistance = vectorToMembrane.Length() + parentVectorToItsMembrane.Length();
+        // Calculates the vector from my center to my membrane towards the parent.
+        // This vector gets rotated back, because I've rotated it two calls above.
```
adQuid, line 1446: why not just use the vectorFromParent variable again?

src/microbe_stage/Microbe.cs
```diff
@@ -1427,22 +1427,39 @@ private void OnIGotAddedToColony()
...
+        var correctedVectorToParent = Membrane.GetExternalOrganelle(vectorToParentRotated.x, vectorToParentRotated.z)
+            .Rotated(Vector3.Down, Rotation.y);
+
+        // Calculates the vector from the parents' center to his membrane towards me.
+        // This gets added to the vector calculated one call before.
+        correctedVectorToParent += ColonyParent.Membrane
+            .GetExternalOrganelle(vectorFromParentRotated.x, vectorFromParentRotated.z)
+            .Rotated(Vector3.Up, globalParentRotation.y);
+
+        ChangeNodeParent(ColonyParent);
```
adQuid, line 1456: If this isn't important to the rotation logic, I would move this to the top or the bottom of the method, so that people don't think it's a part of this logic.

src/microbe_stage/Microbe.cs
```diff
@@ -1427,22 +1427,39 @@ private void OnIGotAddedToColony()
...
-        var vectorToMembrane = Membrane.GetExternalOrganelle(vectorToParentRotated.x, vectorToParentRotated.y);
+        // Gets the global rotation of the parent
+        var globalParentRotation = ColonyParent.GlobalTransform.basis.GetEuler();
+
+        // A vector from the parent to me
+        var vectorFromParent = GlobalTransform.origin - ColonyParent.GlobalTransform.origin;
+
+        // This vector gets as if I had no rotation.
+        // This is important, because GetExternalOrganelle only works with non-rotated microbes
+        var vectorToParentRotated = vectorFromParent.Rotated(Vector3.Up, Rotation.y);
```
adQuid, line 1439: If I rotate "vectorFromParent", it becomes "vectorToParentRotated"? Is this a typo, or a logical error?

adQuid (dismissed):
I haven't tested if this actually solves the problem, but from a code standpoint I can't follow what's going on, and wouldn't be able to debug if this broke. I'm happy to cede to a more seasoned reviewer if someone else is familiar enough to be comfortable working in this area, but otherwise my suggestions are pretty generic since I lack understanding to be more detailed:

- There seem to be a lot of variables being derived from others, which is a good use case to establish helper methods, and name those methods after the logic
- The rotation seems to be isolated logic, so it would be clearer if everything else was in a different method so a future developer can tell that none of those variables are going to come up in this vector logic.

-

src/microbe_stage/Microbe.cs
FuJa0815: to negate the rotation value

src/microbe_stage/Microbe.cs
FuJa0815: then the variable name would be wrong

Kemikal1 (approved):
Hard to test properly bc of the drift ,but the rotation bug is fixed.

src/microbe_stage/Microbe.cs
```diff
@@ -1427,22 +1427,39 @@ private void OnIGotAddedToColony()
...
-            ColonyParent.Membrane.GetExternalOrganelle(vectorToParentRotated.x, vectorToParentRotated.y);
+        // This vector gets rotated as if the parent had no rotation.
+        var vectorFromParentRotated =
+            vectorFromParent.Rotated(Vector3.Down, globalParentRotation.y);
 
-        var requiredDistance = vectorToMembrane.Length() + parentVectorToItsMembrane.Length();
+        // Calculates the vector from my center to my membrane towards the parent.
+        // This vector gets rotated back, because I've rotated it two calls above.
+        var correctedVectorToParent = Membrane.GetExternalOrganelle(vectorToParentRotated.x, vectorToParentRotated.z)
+            .Rotated(Vector3.Up, Rotation.y);
```
Maxonovien, line 1448: So, here, the correction applied is that the original `vectorToParent` pointed towards the parent's center, but you wanted it to point an actual external organelle instead (picking the closest), right? 

Either way I think the nature of this correction should be specified somewhere in the comment above or/and in the variable names, perhaps using intermediate variables if necessary.

src/microbe_stage/Microbe.cs
```diff
@@ -1427,22 +1427,39 @@ private void OnIGotAddedToColony()
...
 
-        var offset = vectorToParent.Normalized() * requiredDistance;
+        // Calculates the vector from the parents' center to his membrane towards me.
+        // This gets added to the vector calculated one call before.
+        correctedVectorToParent -= ColonyParent.Membrane
+            .GetExternalOrganelle(vectorFromParentRotated.x, vectorFromParentRotated.z)
+            .Rotated(Vector3.Up, globalParentRotation.y);
```
Maxonovien, line 1454: The comment here makes me think that there is some underlying intermediate variable that could be made explicit for readability. Also, the resulting vector seems to be the translation to apply so that contact by external organelles is correct? I don't think it points toward parent anyways, so unless I'm mistaken this should go into a dedicated variable.

Also, you say "added" but use operator `-=`. Looking at the code it seems you add the opposite of the vector from parent's center to the membrane? 

Given that we work with vectors, I for one prefer to read vector1+(-vector2) when dealing with opposites, especially when vector2's origin is closer than the center. I find it easier to visualize, but that may not be a shared opinion.

src/microbe_stage/Microbe.cs
```diff
@@ -1427,22 +1427,39 @@ private void OnIGotAddedToColony()
...
 
-        Rotation = oldRotation - ColonyParent.Rotation;
-        Translation = offset.Rotated(Vector3.Down, ColonyParent.Rotation.y);
+        // Reset the rotation so that it is the same before entering the colony.
+        Rotation = oldRotation - globalParentRotation;
```
Maxonovien, line 1457: When is the rotation modified upon entering the colony? I see no line changing the value of Rotation before this one.

Maxonovien (changes requested):
I think I vaguely grasp how and why it works, although there was quite a bunch of things I had to reconsider after my initial reading. So I think some clarification in truly needed, through comments, naming, and, perhaps, additional variables.

From an algorithmic point of view, there is a lot of turning around associated, but I'm not sure we can avoid it. Aside from that the proposed solution seems right (and, it seems to work empirically by Kemikal's word).

src/microbe_stage/Microbe.cs
```diff
@@ -1427,22 +1427,39 @@ private void OnIGotAddedToColony()
...
-        var vectorToMembrane = Membrane.GetExternalOrganelle(vectorToParentRotated.x, vectorToParentRotated.y);
+        // Gets the global rotation of the parent
+        var globalParentRotation = ColonyParent.GlobalTransform.basis.GetEuler();
+
+        // A vector from the parent to me
+        var vectorFromParent = GlobalTransform.origin - ColonyParent.GlobalTransform.origin;
+
+        // This vector gets as if I had no rotation.
+        // This is important, because GetExternalOrganelle only works with non-rotated microbes
+        var vectorToParentRotated = (-vectorFromParent).Rotated(Vector3.Down, Rotation.y);
```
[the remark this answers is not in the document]
FuJa0815: `vectorFromParent` is simply the vector from the parents' center towards mine.
`-vectorFromParent` is the vector from my center towards the parents'

the problem is that `GetExternalOrganelle` does not take the microbe rotation in effect, which means I have to do it. `vectorToParentRotated` is the vector from my center towards the parent's if I had no rotation (pointing upwards).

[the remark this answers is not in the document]
Maxonovien: So, this is what I understood. But I find the name `vectorToParentRotated` misleading, because it is the vector to parent *when not rotated*. I think that speaking of referential or internal representation may be more enlightening over what is going on here.

src/microbe_stage/Microbe.cs
FuJa0815: The method for the external organelle actually just gives me a point on the membrane nearest to the provided point. I have renamed the method to avoid confusion.

Maxonovien: I found this method rather confusing as well. It is much more clearer now I must say. Thanks!

src/microbe_stage/Microbe.cs
Maxonovien: I assumed there was something of this kind. I think this should be explicited. More generally, I think this code does no emphasize enough when referentials are changed, which is however the heart of the issue there.

src/microbe_stage/Microbe.cs
```diff
@@ -1426,23 +1426,61 @@ private void OnIGotAddedToColony()
...
         ChangeNodeParent(ColonyParent);
+    }
+
+    /// <summary>
+    ///   This method calculates the relative rotation and translation this microbe should have to his microbe parent.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Storing the old global translation and rotation, re-parenting and then reapplying the stored values is not
+    ///     worse than this code because this code utilizes GetVectorTowardsNearestPointOfMembrane. This reduces the
```
hhyyrylainen, line 1443: This "worse than this" really confuses me. Is this actually better than before? If so that should be explicitly stated.

src/microbe_stage/Microbe.cs
```diff
@@ -1426,23 +1426,61 @@ private void OnIGotAddedToColony()
...
+        var newTransform = GetNewRelativeTransform();
+
+        Rotation = newTransform.rotation;
+        Translation = newTransform.translation;
+
         ChangeNodeParent(ColonyParent);
+    }
+
+    /// <summary>
+    ///   This method calculates the relative rotation and translation this microbe should have to his microbe parent.
```
hhyyrylainen, line 1438: Would it be possible to reword this so that the microbe isn't referred to by a gendered pronoun. I've been careful to not do this elsewhere when commenting so this looks odd to me.

src/microbe_stage/Microbe.cs
```diff
@@ -1426,23 +1426,61 @@ private void OnIGotAddedToColony()
...
+        // A vector from the parent to me
+        var vectorFromParent = GlobalTransform.origin - ColonyParent.GlobalTransform.origin;
+
+        // A vector from me to the parent
+        var vectorToParent = -vectorFromParent;
+
+        // This vector represents the vectorToParent as if I had no rotation.
+        // This works by rotating vectorToParent by the negative value (therefore Down) of my current rotation
+        // This is important, because GetVectorTowardsNearestPointOfMembrane only works with non-rotated microbes
+        var vectorToParentWithoutRotation = vectorToParent.Rotated(Vector3.Down, Rotation.y);
```
hhyyrylainen, line 1462: Why not use the quaternion rotation, instead decomposing (and losing information) first and losing information. then requiring to specify the direction here.
Then again I'm having trouble understanding what goes here?
What does this actually do?

If we have a direction vector in local coordinates towards the parent, shouldn't it be enough to just discard the Y value and then use x and z as the x and y values to get to organelle position coordinates to use with the membrane code?

hhyyrylainen: Perhaps drawing a picture (I can upload one somewhere to link here) would greatly improve the understandability of this code.

hhyyrylainen (changes requested):

src/microbe_stage/Microbe.cs
FuJa0815: I've made some scrambles on a piece of paper, but it isn't really that useful. I could try to visualize this bettet once I'm back

FuJa0815: Maybe this helps a little bit... (There is a reason why I'm not in the graphics team :p)
VectorExpl

Raw Gimp file: [VectorExpl.zip](https://github.com/Revolutionary-Games/Thrive/files/6987151/VectorExpl.zip)

hhyyrylainen: Uploaded that image to storage that github won't delete at some point: https://randomthrivefiles.b-cdn.net/documentation/fixed_colony_rotation_explanation_image.png

hhyyrylainen (approved):
I did some comment improvements, so the code is fine enough for me now. I'll leave playtesting for when I'm trying to merge the Fixes colony physics PR.

PR #2464, opened 2021-08-05
Style guide update abbreviations and ternaries

### LINKED ISSUE #2392: Amend allowed list of abbreviations to styleguide
Right now the styleguide prohibits abbreviations, but in fact we use a few in Thrive, these should be mentioned as the only allowed abbreviations:
- min
- max
- pos
- rect (when variable holds a class that has Rect in the name, also allowed in class names to make such classes)

And maybe some others if they are commonly already in use in the source code.

### LINKED ISSUE #2454: Add to styleguide that nested ternaries are disallowed
non-nested ternaries are fine as long as they are readable, nested ternaries are always banned and need to be re-written as if-blocks.

Maxonovien:

This PR updates the style guide in the docs to account for abbreviations and nested ternaries.

**Related Issues**

Closes #2454 closes #2392

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


Maxonovien:
As mentioned in #2392 's comments, there is a `Sqr` abbreviation used inconsistently in `SpawnSystem.cs`. I'd vote to rule it out as an abbreviation because "Squared" is not that long. But the use is inconsistent so it would require minor rewriting anyways, I guess.

doc/style_guide.md
```diff
@@ -31,6 +31,16 @@ Code style rules
...
+  
+- Similarly, some very common abbreviations are used in the code,
+  and can (and should) thus be used when naming variables. These are
+  however *rare* exceptions, not the rule. The allowed abbreviations 
+  are listed below. No other abbreviation should be used without prior
+  discussion (and good reasons).
+  - `min`
+  - `max`
+  - `pos`
+  - `rect` (when related to class names and instances)
```
hhyyrylainen, line 43:
```suggestion
  - `rect` (when related to class names and variables holding instances of those classes)
```

hhyyrylainen (changes requested):
Also if you don't mind could you update the existing code to conform to the new styleguide (as you mentioned you know somep laces that don't follow the rules)?

Maxonovien:
Sure, I'll change what I spotted. If anyone has spotted further things related to abbreviations/style guide, please let me know.

hhyyrylainen:
@Maxonovien I edited the description here to fix it. You need to place the "closes" keyword before each linked issue. You can't specify a list of issues.

Kasterisk:
I think we can also add `rot` to the list for rotation to complement `pos`.  MicrobeEditor has `organelleRot` variable that pretty much uses it for example, though I believe it's the only place where the abbreviation is used.

hhyyrylainen (approved):

hhyyrylainen:
I just noticed that this broke save compatibility due to the renamed property. I'll make a save upgrader to remedy this situation.

PR #2475, opened 2021-08-13
Add note about camelCase internal registry item names

### LINKED ISSUE #2473: Add internal registry type naming rule to styleguide
name internal names in registries with camelCase and not snake_case, otherwise that will break stuff as Godot Node naming can't match the internal names (due to naming rules regarding that).

hhyyrylainen:
**Related Issues**

closes #2473

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/doc/style_guide.md
+++ b/doc/style_guide.md
@@ -342,9 +342,10 @@ Other files
 - New JSON files should prefer PascalCase keys. Existing JSON files
   should stick to what other parts of that file use.
 
-- Registry items (for example organelle internal names) should use camelCase
-  keys, and not snake_case. Otherwise other names that follow the internal
-  names will violate other naming conventions.
+- Registry items (for example organelles) should use camelCase for their
+  internal names (IRegistryType.InternalName), and not snake_case.
+  Otherwise other names that follow the internal names will violate other
+  naming conventions.
 
 - Do not use `<br>` in markdown unless it is a table where line breaks
   need to be tightly controlled. Use blank lines instead of
--- a/doc/style_guide.md
+++ b/doc/style_guide.md
@@ -342,6 +342,10 @@ Other files
 - New JSON files should prefer PascalCase keys. Existing JSON files
   should stick to what other parts of that file use.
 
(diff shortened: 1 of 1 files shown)


doc/style_guide.md
```diff
@@ -342,6 +342,10 @@ Other files
 - New JSON files should prefer PascalCase keys. Existing JSON files
   should stick to what other parts of that file use.
 
+- Registry items (for example organelle internal names) should use camelCase
```
Maxonovien, line 345: `Registry items (for example organelle internal names)`

As someone who's not necessarily used to all the terminology, I find this a little bit unclear. By internal names, you mean the ones defined in their JSON file right?

hhyyrylainen: Yes, for example OrganelleDefinition.InternalName is a property and gets loaded from json dictionary key.

Maxonovien: Oh, so you mean the `"name"` key in the organelles.json should be (and is) in camelCase? Reading the examples I though it was what was following this key (e.g. "PREDATORY_PILUS"). I think the example is more misleading as it is right now...

hhyyrylainen: I don't see things that way as I'm fully aware that all of the registry types have internal names that have nothing to do with translations. In fact if translated names are accidentally used where internal names should be used, everything will be broken.
Personally for you, what should be written as documentation https://github.com/Revolutionary-Games/Thrive/blob/33d5740bd33d9151c38553b629352e6b42eaa064/simulation_parameters/IRegistryType.cs#L6 to make this less confusing
I'll at least mention in this example to refer to IRegistryType.InternalName

Maxonovien: Yeah, I think I got it more clearly here. It was not much about translations, but I got confused with the actual value instead of the key.

Anyways, I think that it is less confusing now that there is the reference.

Maxonovien (approved):
I find it clear enough with the reference to `IRegistryType`, so I'd be happy with having this brought to the style guide as such.

PR #2486, opened 2021-08-14
Microbe HUD patch name display enhancement

Kasterisk:

Improved the plain patch name label into a fancy overlay title with background shadow dark souls style. Prompted every time the patch changes in the editor. A flashing effect is preferable but no idea how to do that currently, could be left for later PRs.

https://user-images.githubusercontent.com/54026083/129440864-81962943-5718-4f28-b2bc-2540ec685af8.mp4

Also removed a double microbe stage opening tutorial `SendEvent` calls in MicrobeStage.cs. The effect will be the tutorial to popup after the fade transition finishes instead of instantly.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/tutorial/TutorialEventArgs.cs
```diff
@@ -87,3 +87,13 @@ public PatchEventArgs(Patch patch)
...
+
+public class ActionEventArgs : TutorialEventArgs
+{
+    public ActionEventArgs(Action data)
+    {
+        Data = data;
+    }
+
+    public Action Data { get; }
+}
```
Kasterisk, line 99: I'll leave this in since this probably could be useful in the future.

hhyyrylainen: Why is it Action. Data is like static data, but Action is like a basic callback / delegate type, right?

Kasterisk: Renamed.

Kasterisk:
The patch name not popping up from an editor save problem is a bit more tricky than I thought, I don't think my earlier ideas from Discord are feasible. I guess a more proper fix would require some access to a saved `previousPatch` by saving MicrobeStage's PatchManager somehow to make the patch change detection effective. But then again, as far as I understand previousPatch will likely be overriden to current patch by [MicrobeStage.cs#L273](https://github.com/Revolutionary-Games/Thrive/blob/master/src/microbe_stage/MicrobeStage.cs#L273) the second the editor save is loaded so no change will be detected afterwards. At this point I think it's probably best to leave this be and not risk save compatibility. Any design advise is welcome.

> One note/opinion I have is that the patch name disappears too quickly. I don't know could just be me.

Probably worth to increase the patch name visible duration a bit.

hhyyrylainen:
So I just took a look and it seems like there's a bug even without this: the patch entities are not despawned if you load a save from the editor and move to a different patch.

I put a breakpoint here:
https://github.com/Revolutionary-Games/Thrive/blob/71fcf56153c5888a3f923177164b1e715f768f8d/src/microbe_stage/PatchManager.cs#L46
And that didn't trigger. Only when I moved to a new patch in the editor that was not loaded from save things worked.
One really quick idea regarding this is that we could make this:
https://github.com/Revolutionary-Games/Thrive/blob/71fcf56153c5888a3f923177164b1e715f768f8d/src/microbe_stage/PatchManager.cs#L18

a json saved property. Probably could do it by using the JSON set only child properties attribute, so that other code regarding the PatchManager object creation doesn't need to change, at least I *think* that's possible. If no one picks this up before then I can look into this later this week.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -546,7 +545,8 @@ public void OnReturnFromEditor()
     public void OnFinishTransitioning()
     {
         TransitionFinished = true;
-        TutorialState.SendEvent(TutorialEventType.EnteredMicrobeStage, EventArgs.Empty, this);
+        TutorialState.SendEvent(
+            TutorialEventType.EnteredMicrobeStage, new CallbackEventArgs(HUD.PopupPatchInfo), this);
```
hhyyrylainen, line 549: This callback won't be set when loading a save. I'll let this slide as it is very unlikely that someone will make a save (or maybe it's impossible to get to the menu yet) while the welcome message is displayed.

hhyyrylainen (approved):
I think this is good enough, I can do the save loading fixes after merging this.
Just need to resolve the translation conflicts.
One potential thing that could be slightly improved is when the auto-save happens, as right now it happens when the patch name is fading in. I think it would look better if the auto-save was triggered when the patch name was at the maximum opaqueness.

hhyyrylainen:
I ended making that saving fixes PR which this should be able to use: https://github.com/Revolutionary-Games/Thrive/pull/2564

src/microbe_stage/MicrobeStage.cs
Kasterisk: Would it be sufficient by just adding json property attribute to the action variable in `patchNamePopup`?

hhyyrylainen: I don't think so. This is because the callback targets need the specific allow attribute (which is easy to add) but the harder part is that it needs to refer to the HUD object, which is not loaded from JSON, so that would need a special kind of handling to somehow pre-register the HUD instance in the json $id resolver to work.

Kasterisk: I see.

locale/zh_CN.po
```diff
@@ -1522,24 +1522,26 @@ msgid "RUN_AUTO_EVO_DURING_GAMEPLAY"
 msgstr "在游戏中启用auto-evo"
 
 #: ../src/general/OptionsMenu.tscn:810
+#, fuzzy
 msgid "DETECTED_CPU_COUNT"
-msgstr "检测到的CPU数量："
+msgstr "选中的："
```
hhyyrylainen, line 1527: It seems this PR has a bunch of translation reverts?
These need to be fixed, we don't want to lose the translators work.

hhyyrylainen (changes requested):
Code and functionality seems fine, but I noticed some serious problems with some translations getting removed and marked as fuzzy.

locale/zh_CN.po
Kasterisk: I think it should be fixed now, messed it up during the previous merge due to my master branch not being up to date.

hhyyrylainen (approved):

Issue #2295, opened 2021-06-05
ATP consumption greater when the cell is static


84634E1A607A:
**Describe the bug**
ATP consumption is greater when the cell is not moving. Strange.

**Screenshots**
![Static](https://user-images.githubusercontent.com/60734649/120881131-a1fe5a00-c601-11eb-95cd-4451ce014c84.png)
Static

![Moving](https://user-images.githubusercontent.com/60734649/120881134-a6c30e00-c601-11eb-88d5-1785c247ffa2.png)
Moving

**System info (please complete the following information):**
0.5.4.0-alpha

Cell:
![image](https://user-images.githubusercontent.com/60734649/120881156-d40fbc00-c601-11eb-99c2-53915803e8d6.png)

hhyyrylainen:
Are you referring to the ATP number being red? That just means it is the limiting factor in the process speed. The fact that it is a process result means that when it is the limiting factor, there's not enough storage. So in this case the ATP being red says "you don't have enough storage for this process to run fast", which in turn means that you aren't using as much ATP when stationary than when moving. So I think everything is working correctly.

hhyyrylainen:
Or are you referring to the glucose numbers (just noticed those)? I think what's happening is that as the process can't run fullspeed it shows the numbers as-if it could run, which has glucose higher, but the lower one shows the actual glucose consumption. I suppose that is a big or at least unoptimal display.

84634E1A607A:
Actually what I want to say is the amount of glucose consumed. When the cell is static, it's 0.124; when the cell moves forward, it gets down to 0.063. When I move backward or elsewhere the glucose consumption rate is between them. Does that mean when the cell is moving it costs less glucose?

hhyyrylainen:
Probably not. See my later comment: I think it's just a display difference: when a process can't run (the output is red) the input amounts are computed from the full speed process, but when the process can run the input amounts are the actual amounts consumed (which, if the process can't run at full speed, is less than the theoretical maximum). It's pretty hard to display this in a consistent manner.
The reason I coded it this way is that if a process can't run, you couldn't see which compounds you need (and how much) as all the inputs would be just zeroes indicating nothing is consumed...

84634E1A607A:
But when the cell is moving forward glucose compound (left bottom corner) accumulates faster. That's pretty strange. (I'll post a video soon.)

84634E1A607A:
[...]

Issue #2085, opened 2021-02-07
Investigate if multiple organelles doing the same process don't show correctly in process panel


hhyyrylainen:
multiple organelles doing the same process don't show in the process panel correctly?

hhyyrylainen:
I think I may have fixed this with the PR where now each process is represented by a single TweakedProcess, so this probably is no longer a problem.

hhyyrylainen:
Seeing no one complain about this anymore, I'll assume this is fixed.

Issue #1982, opened 2020-12-31
Check if the process system should report amounts as 0 for statistics


hhyyrylainen:
when a process can't run. Might make the process panel numbers better?

I already left most of this code in the process system but commented out, so it'll be easy to try this out by swapping which `AddInputAmount` and `AddOutputAmount` are commented out.

PR #2488, opened 2021-08-14
Allow processes to run at lower than maximum speed rate

### LINKED ISSUE #2419: Allow process system to run a process at lower rate
Once processes are properly combined to single objects (https://github.com/Revolutionary-Games/Thrive/pull/2418) the process system needs to be able to run them at partial speed due to storage limits. Maybe it already works but someone should confirm? I'm also fine with if the process statistics make more sense to combine, but as they have quite a lot of data, I think this'll be a better way.

hhyyrylainen:

Reworks the process system to allow processes due to output or input constraints to run at partial speed
bumps version to RC1

I'll make the RC1 from this branch

**Related Issues**

closes  #2419

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -454,6 +454,8 @@ public static class Constants
     public const float EDITOR_ARROW_OFFSET = 3.5f;
     public const float EDITOR_ARROW_INTERPOLATE_SPEED = 0.5f;
 
+    public const float MINIMUM_RUNNABLE_PROCESS_FRACTION = 0.0005f;
+
     public const float DEFAULT_PROCESS_SPINNER_SPEED = 365.0f;
     public const float DEFAULT_PROCESS_STATISTICS_AVERAGE_INTERVAL = 0.4f;
 
@@ -574,6 +576,9 @@ public static class Constants
             AUTO_EVO_MINIMUM_VIABLE_POPULATION) ?
             0 :
             -42;
+
+    private const uint MinimumRunnableProcessFractionIsAboveEpsilon =
+        (MINIMUM_RUNNABLE_PROCESS_FRACTION > MathUtils.EPSILON) ? 0 : -42;
 
     // ReSharper restore UnreachableCode
 #pragma warning restore CA1823
--- a/src/microbe_stage/ProcessSystem.cs
+++ b/src/microbe_stage/ProcessSystem.cs
@@ -275,6 +275,8 @@ public class ProcessSystem
 
         speedFactor *= process.Rate;
 
+        // Note that we don't consider storage constraints here so we don't use spaceConstraintModifier calculations
+
         // So that the speed factor is available here
         foreach (var entry in process.Process.Inputs)
         {
@@ -348,17 +350,19 @@ public class ProcessSystem
         // Can your cell do the process
         bool canDoProcess = true;
 
-        // Loop through to make sure you can follow through with your
-        // whole process so nothing gets wasted as that would be
-        // frustrating.
         float environmentModifier = 1.0f;
+
+        // This modifies the process overall speed to allow really fast processes to run, for example if there are
+        // a ton of one organelle it might consume 100 glucose per go, which might be unlikely for the cell to have
+        // so if there is *some* but not enough space for results (and also inputs) this can run the process as
+        // fraction of the speed to allow the cell to still function well
+        float spaceConstraintModifier = 1.0f;
 
         // First check the environmental compounds so that we can build the right environment modifier for accurate
         // check of normal compound input amounts
         foreach (var entry in processData.Inputs)
         {
-            // Set used compounds to be useful, we dont want to purge
-            // those
+            // Set used compounds to be useful, we dont want to purge those
             bag.SetUseful(entry.Key);
 
             if (!entry.Key.IsEnvironmental)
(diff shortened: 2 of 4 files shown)


src/microbe_stage/ProcessSystem.cs
```diff
@@ -377,81 +381,115 @@ private void ProcessNode(IProcessable processor, float delta, float inverseDelta
...
 
-            var inputRemoved = entry.Value * process.Rate * delta;
-
-            currentProcessStatistics?.AddInputAmount(entry.Key, inputRemoved * inverseDelta);
+            var inputRemoved = entry.Value * process.Rate * delta * environmentModifier * spaceConstraintModifier;
 
             // currentProcessStatistics?.AddInputAmount(entry.Key, 0);
+            // We don't multiply by delta here because we report the per-second values anyway. In the actual process
+            // output numbers (computed after testing the speed), we need to multiply by inverse delta
+            currentProcessStatistics?.AddInputAmount(entry.Key, entry.Value * process.Rate * environmentModifier);
```
adQuid, line 397: You could divide inputRemoved by the delta here and avoid the formulae going out of sync.

src/microbe_stage/ProcessSystem.cs
```diff
@@ -377,81 +381,115 @@ private void ProcessNode(IProcessable processor, float delta, float inverseDelta
...
         {
             if (entry.Key.IsEnvironmental)
                 continue;
 
-            var inputRemoved = entry.Value * process.Rate * delta;
-
-            currentProcessStatistics?.AddInputAmount(entry.Key, inputRemoved * inverseDelta);
+            var inputRemoved = entry.Value * process.Rate * delta * environmentModifier * spaceConstraintModifier;
 
             // currentProcessStatistics?.AddInputAmount(entry.Key, 0);
```
adQuid, line 394:
```suggestion
```
pretty clearly looks like code from a past age.

src/microbe_stage/ProcessSystem.cs
```diff
@@ -377,81 +381,115 @@ private void ProcessNode(IProcessable processor, float delta, float inverseDelta
...
+                        // Due to rounding errors there can be very small disparity here between the amount available
+                        // and what we will take with the modifiers. See the comment in outputs for more details
+                    }
+                }
+
+                if (!canRun)
+                {
+                    canDoProcess = false;
+                    currentProcessStatistics?.AddLimitingFactor(entry.Key);
+                }
```
adQuid, line 423:
```suggestion
                    var neededModifier = availableAmount / inputRemoved;

                    if (neededModifier > Constants.MINIMUM_RUNNABLE_PROCESS_FRACTION)
                    {
                        spaceConstraintModifier = neededModifier;
                        canRun = true;

                        // Due to rounding errors there can be very small disparity here between the amount available
                        // and what we will take with the modifiers. See the comment in outputs for more details
                    }
                    else
                    {
                        canDoProcess = false;
                        currentProcessStatistics?.AddLimitingFactor(entry.Key);
                    }
                }
```

I think you can dispense with the flag and one of the conditionals here.  If availableAmount is zero, neededModifier will be as well. This won't handle a negative MINIMUM_RUNNABLE_PROCESS_FRACTION, but I think it's safe to discount that.

src/microbe_stage/ProcessSystem.cs
```diff
@@ -377,81 +381,115 @@ private void ProcessNode(IProcessable processor, float delta, float inverseDelta
...
             bag.SetUseful(entry.Key);
 
-            // Apply the general modifiers and
-            // apply the environmental modifier
-            var outputAdded = entry.Value * process.Rate * delta * environmentModifier;
-
-            currentProcessStatistics?.AddOutputAmount(entry.Key, outputAdded * inverseDelta);
+            var outputAdded = entry.Value * process.Rate * delta * environmentModifier * spaceConstraintModifier;
 
             // currentProcessStatistics?.AddOutputAmount(entry.Key, 0);
```
adQuid, line 434:
```suggestion
```

also looks like old code that can go now.

simulation_parameters/Constants.cs
```diff
@@ -575,6 +577,9 @@ public static class Constants
             0 :
             -42;
 
+    private const uint MinimumRunnableProcessFractionIsAboveEpsilon =
```
adQuid, line 580: I'm all for hitchhikers guide references, but is this ever used?

adQuid (commented):
I won't hold up a release candidate for it, but I see some easy wins here. The amount of repeated logic is an obvious code smell to me, but a full refactor of process system isn't the goal here.

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: I kept this, for now, because the process panel has multiple issues open against it:
https://github.com/Revolutionary-Games/Thrive/issues/2085
https://github.com/Revolutionary-Games/Thrive/issues/1982
https://github.com/Revolutionary-Games/Thrive/issues/2295

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: It's just a few lines away. And anyway there's 4 closely related formulas in this method already, so I thought that the performance win of avoiding a divide is better here.

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: I wanted to early return if there is no compound to be had.
I wrote the logic the way I did because it was easier for me to reason about and make absolutely sure there isn't funny business going on if there is too little input compound.

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: https://github.com/Revolutionary-Games/Thrive/pull/2488#discussion_r688922023

simulation_parameters/Constants.cs
hhyyrylainen: This will actually prevent the code from compiling if the constant is wrong.
C# doesn't have static assert (like C++ does), so this is the only way to implement static asserts. See the one above this one, there's already one of these kinds of "static asserts" in the code.

src/microbe_stage/ProcessSystem.cs
adQuid: This feels like a weak argument to me, but if you really want to keep it I would suggest at least adding an explanatory comment. Future devs aren't going to guess that.

src/microbe_stage/ProcessSystem.cs
adQuid: An alternative option would be to just multiply everything BUT the delta first, then calculate the input removed afterwards. That saves you even more performance.

src/microbe_stage/ProcessSystem.cs
adQuid: I don't see any return statements here.

I can't argue against it being easier to you, but in my experience nothing confuses a new dev to the codebase like flag variables, which is why I try to avoid them.

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: I'm just keeping these lines, though. So I'm going to pull the "it was this way already" argument as to why these lines are here.

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: I misspoke about the early return. I meant that I skip the entire logic block.

I think flag variables are used elsewhere, at least in my personal coding style I use them a lot.
Anyway the main point here is that this version I'm sure does the right thing. If I were to change this I would want to get at least a couple of people playtesting this extensively to ensure correctness.

84634E1A607A:
- Save: Succeed, works fine
- Load: Succeed, works fine
- #2419 solved: True. Now processes can work at a lower rate.

- Processes don't seem to sum up correctly. In these two pics, photosynthesis speed is lower than respiration speed sum, but I can still gain glucose. 
  image
  image
  image

hhyyrylainen:
The numbers are averaged over some time, that tries to stop them changing really fast, but even that doesn't seem to entirely help. So the respiration process is probably not active all of the time, I think. And besides the numbers are kinda not exact.
This PR doesn't try to make the process numbers exact (I linked earlier in reply to adQuid the actual issues tracking making the process panel numbers match up). I've tried tweaking the process panel a lot, and it seems very hard so I'm not going to focus on fixing those, just going to leave the numbers as is for someone to pick up.

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: That sounds like a good idea. I'll do that once I'm done working on the feature for ThriveDevCenter that I'm currently working on.

src/microbe_stage/ProcessSystem.cs
adQuid: fair enough

src/microbe_stage/ProcessSystem.cs
adQuid: I guess it comes down to subjectivity then. I'd like to see this trimmed down, but I won't block over it.

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: Done

adQuid (approved):

hhyyrylainen:
I'll merge this now. I'm going to check other pending PRs next week, and I'll at least make one PR bumping the version number to 0.5.5.0 and add a save upgrade step for that.

PR #2490, opened 2021-08-14
2D membranes given 3D shading and lighting through normal mapping

### LINKED ISSUE #909: Make lighting affect membranes
Due to some bsf weirdness with the lighting on membranes, light on membranes is currently disabled.
It should be fixed so that light affects the membranes as well.

Nunez2196:

This PR gives 2D membranes the ability to mimic 3D lighting through normal mapping

**Related Issues**

Closes #909

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
I tried to checkout this branch on my computer, and CI is correct, this branch refers to Git LFS files that aren't actually uploaded:
```
Couldn't checkout e7898376ed9444a8259a7c7a5570d3bb6cdb8a57
				Downloading assets/textures/MembraneNormals.png (162 KB)
				Error downloading object: assets/textures/MembraneNormals.png (6407ffb): Smudge error: Error downloading assets/textures/MembraneNormals.png (6407ffb9b7d9304222d9554ae92dd144ca70b4305333aa699563b992d8a85971): batch response: Repository or object not found: https://dev.revolutionarygamesstudio.com/api/v1/lfs/thrive/objects/batch
				Check that it exists and that you have proper access to it
				Errors logged to /home/hhyyrylainen/Projects/thrive/.git/lfs/logs/20210818T220342.778117582.log
				Use `git lfs logs last` to view the log.
				external filter 'git-lfs filter-process' failed
				assets/textures/MembraneNormals.png: smudge filter lfs failed
```

Nunez2196:
> I tried to checkout this branch on my computer, and CI is correct, this branch refers to Git LFS files that aren't actually uploaded:
> 
> ```
> Couldn't checkout e7898376ed9444a8259a7c7a5570d3bb6cdb8a57
> 				Downloading assets/textures/MembraneNormals.png (162 KB)
> 				Error downloading object: assets/textures/MembraneNormals.png (6407ffb): Smudge error: Error downloading assets/textures/MembraneNormals.png (6407ffb9b7d9304222d9554ae92dd144ca70b4305333aa699563b992d8a85971): batch response: Repository or object not found: https://dev.revolutionarygamesstudio.com/api/v1/lfs/thrive/objects/batch
> 				Check that it exists and that you have proper access to it
> 				Errors logged to /home/hhyyrylainen/Projects/thrive/.git/lfs/logs/20210818T220342.778117582.log
> 				Use `git lfs logs last` to view the log.
> 				external filter 'git-lfs filter-process' failed
> 				assets/textures/MembraneNormals.png: smudge filter lfs failed
> ```

strange, all i have to do is push my changes with Git LFS right? Or do I have to upload LFS objects to the DevCenter or something?

hhyyrylainen:
When you push, the LFS objects *should* be automatically uploaded. If the upload fails, it should even cancel the push. You have ended up in a situation where you need to google how to force re-check of LFS objects to upload, as I don't think all objects are re-checked each push, only the LFS objects that are mentioned in the commits you are pushing are checked.

.gitattributes
```diff
@@ -26,3 +26,4 @@
 *.otf filter=lfs diff=lfs merge=lfs -text
 *.shape filter=lfs diff=lfs merge=lfs -text
 *.jpeg filter=lfs diff=lfs merge=lfs -text
+assets/textures/MembraneNormals.png filter=lfs diff=lfs merge=lfs -text
```
hhyyrylainen, line 29: This should not be needed. All *.png files should be already included in here

CIConfiguration.yml
```diff
@@ -5,6 +5,7 @@ jobs:
     image: thrive/godot-ci:v10
     cache:
       loadFrom:
+
```
hhyyrylainen, line 8: unnecessary added blank line

.gitattributes
hhyyrylainen: More precisely I meant that this file already contains line 9 which is: `*.png filter=lfs diff=lfs merge=lfs -text` so this line should be entirely redundant.

shaders/Membrane.shader
```diff
@@ -25,16 +29,26 @@ void vertex(){
 }
 
 void fragment(){
-    vec4 normal = texture(albedoTexture, UV);
+    vec4 albedo = texture(albedoTexture, UV);
+    
+    vec4 normalmap = texture(normalTexture, UV);
+    vec3 normals = vec3 (normalmap.r * 2.0 - 1.0, normalmap.g * 2.0 - 1.0, 
+    normalmap.b * 2.0 - 1.0);
```
hhyyrylainen, line 36:
```suggestion
    vec3 normals = vec3(normalmap.r * 2.0 - 1.0, normalmap.g * 2.0 - 1.0, 
        normalmap.b * 2.0 - 1.0);
```

shaders/Membrane.shader
```diff
@@ -25,16 +29,26 @@ void vertex(){
...
+
+    float y = normals.y;
+    normals.y = -normals.z;
+    normals.z = y;
```
hhyyrylainen, line 40: Why is this flipping done here, instead of like having the normal image file already have correct data?

shaders/Membrane.shader
```diff
@@ -25,16 +29,26 @@ void vertex(){
...
+
     vec4 damaged = texture(damagedTexture, UV);
-    vec4 final = ((normal * healthFraction) + 
-        (damaged * (1.f - healthFraction))) * tint;
+    vec4 final = vec4 ((((albedo.rgb * healthFraction) + 
+        (damaged.rbg * (1.f - healthFraction))) * tint.rgb), albedo.a);
```
hhyyrylainen, line 44: Why is the damaged texture alpha now ignored? (If I understood this change correctly...)

hhyyrylainen (changes requested):

.gitattributes
Nunez2196: > More precisely I meant that this file already contains line 9 which is: `*.png filter=lfs diff=lfs merge=lfs -text` so this line should be entirely redundant.

alright i'll delete that line real quick

shaders/Membrane.shader
Nunez2196: that's how blender saves the bakes. I could probably try to mess with blender to make it bake correctly but, this was faster as i already know how to do this.

hhyyrylainen: I guess we just have to live with this, then. As long as it is like the default option. It would be very confusing in the future if new normal maps are made and suddenly they end up looking wrong because of this swapping.

shaders/Membrane.shader
Nunez2196: If I am remembering right, it was unnecessary and just messed with shading,(or I was messing with stuff and forgot to change it back) i'll have to look again though.

hhyyrylainen: My concern is that if the crack parts in the damaged texture are at locations where the undamaged texture is very transparent, the damage won't show up well. And this limitation of needing the same alpha in the damaged as the undamaged texture, probably will limit art choices etc.

Nunez2196: > My concern is that if the crack parts in the damaged texture are at locations where the undamaged texture is very transparent, the damage won't show up well. And this limitation of needing the same alpha in the damaged as the undamaged texture, probably will limit art choices etc.

now I remember after looking, the person before combined Alpha values very incorrectly, you're supposed to use mix(x, y, 1.0) not just add them together. I added this just now on my local branch.

hhyyrylainen: Don't you need to take the healthFraction into account in the mixing, though?
Wasn't this math right, since it took the healtFraction of alpha from the undamaged texture and (1 - healthFraction) from the damaged one? That results in alpha being 1 if healthFraction is 1 and the undamaged texture is opaque, and otherwise a mix between the two alpha values?

Nunez2196: > Don't you need to take the healthFraction into account in the mixing, though?
> Wasn't this math right, since it took the healtFraction of alpha from the undamaged texture and (1 - healthFraction) from the damaged one? That results in alpha being 1 if healthFraction is 1 and the undamaged texture is opaque, and otherwise a mix between the two alpha values?

it does with the RGB but actually now that I rethink and reread, it seems you are correct, the old math is fine, i guess there was something else i'm just not remembering. Anyways it seems the old stuff was fine so i'll revert.

hhyyrylainen (approved):

PR #2494, opened 2021-08-17
Adjusted storage amounts for binding agent, alongside other misc. cor…

### LINKED ISSUE #2491: Some organelle tooltips are missing osmoregulation cost
Noticed on the community forums that binding agents (and also at least chemosynthesizing proteins and the pilus) is missing osmoregulation from the editor tooltip.

- [ ] These should be fixed
- [ ] It would be nice to have an automatic system fill out that part of the tooltip based on the organelle data from json

buckly90:
…rections

Decreases storage and mass of binding agent, and increases storage of nucleus and vacuoles. Also corrects tooltips.

**Related Issues**
Closes #2491

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/gui_common/tooltip/ToolTipManager.tscn
+++ b/src/gui_common/tooltip/ToolTipManager.tscn
@@ -1135,7 +1135,7 @@ margin_bottom = 20.0
 mouse_filter = 2
 script = ExtResource( 21 )
 DisplayName = "STORAGE"
-ModifierValue = "+15"
+ModifierValue = "+1"
 ModifierValueColor = Color( 0, 1, 0, 1 )
 ModifierIcon = ExtResource( 12 )
 
@@ -1786,7 +1786,7 @@ margin_bottom = 20.0
 mouse_filter = 2
 script = ExtResource( 21 )
 DisplayName = "STORAGE"
-ModifierValue = "+6"
+ModifierValue = "+8"
 ModifierValueColor = Color( 0, 1, 0, 1 )
 ModifierIcon = ExtResource( 12 )
 
@@ -1796,7 +1796,7 @@ margin_right = 24.0
 margin_bottom = 18.0
 custom_fonts/font = ExtResource( 26 )
 custom_colors/font_color = Color( 0, 1, 0, 1 )
-text = "+6"
+text = "+8"
 
 [node name="Name" type="Label" parent="GroupHolder/organelleSelection/vacuole/MarginContainer/VBoxContainer/ModifierList/storage"]
 margin_left = 28.0
--- a/src/gui_common/tooltip/ToolTipManager.tscn
+++ b/src/gui_common/tooltip/ToolTipManager.tscn
@@ -1041,7 +1041,7 @@ margin_bottom = 209.0
 [node name="pilus" parent="GroupHolder/organelleSelection" instance=ExtResource( 5 )]
 visible = false
 DisplayName = "PREDATORY_PILUS"
-Description = "PREDATORY_PILUS_DESCRIPTION"
+ProcessesDescription = "PREDATORY_PILUS_DESCRIPTION"
 MutationPointCost = 30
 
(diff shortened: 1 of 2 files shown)


buckly90:
The pilus tooltip is still missing storage, I haven't quite figured out why it won't show, I'll try investigating again at a later time.

hhyyrylainen:
Note to myself, #2481 has a second part about automatically building the tooltip osmoregulation cost parts

hhyyrylainen:
@Kasterisk might be able to help with the tooltip problem.

buckly90:
When's the cutoff time for the next update? Might just rip my changes to the pilus tooltip and leave it for later so these changes can make it in.

Kasterisk:
I had a look through the tooltip changes, the reason why it didn't show was due to the container node (the one containing osmo cost, storage etc) set to invisible and so its child nodes (modifiers) was never shown. Simple fix is just to turn it back on, I'll do this change along with adding the storage @buckly90.

Also yeah we want to eventually have the tooltip modifier parts to be created automatically rather than manually as of now but that's for another PR.

buckly90:
Who was the monster that did that? >:(

Kasterisk:
> Who was the monster that did that? >:(

It was me :P Well, reason was so the node don't take up space in the tooltip, it creates noticeable padding between tooltip title and description even when empty but since it's now filled it would make sense to turn it back on.

buckly90:
Ah I see, that makes sense. Thanks for the explanation.

hhyyrylainen:
If this is fixed by tomorrow, a little over 24 hours, and someone (or me, as it usually ends up being) can test this, this can still be included in the release.

Kasterisk:
Pushed my fix.

src/gui_common/tooltip/ToolTipManager.tscn
```diff
@@ -530,6 +530,79 @@ margin_bottom = 204.0
...
+
+[node name="Name" type="Label" parent="GroupHolder/organelleSelection/chemoSynthesizingProteins/MarginContainer/VBoxContainer/ModifierList/storage"]
+margin_left = 20.0
+margin_right = 73.0
+margin_bottom = 19.0
+custom_fonts/font = ExtResource( 27 )
+text = "Storage"
+__meta__ = {
+"_editor_description_": "PLACEHOLDER"
+}
```
hhyyrylainen, line 559: Shouldn't this string be translatable? Are the other organelles like this as well, @Kasterisk ?

hhyyrylainen (changes requested):
Looks like the vacuole tooltip was not updated (it still says 6).

src/gui_common/tooltip/ToolTipManager.tscn
Kasterisk: This is just a flavor text for display only in the editor, the actual string will be set by code which is already translatable. The other organelle tooltips are also like this (if I'm not misremembering my own design).

hhyyrylainen: Ah, okay, I guess I don't seem to know enough about this. I was also a bit confused as the numbers where in two different places, but I guess that's also the editor only stuff?
So that just leaves one of the storage change numbers not being also put in a tooltip as a problem.

Kasterisk: Yep, the organelle tooltip system warrants an overhaul eventually.

buckly90:
Can't believe I forgot to update the tooltips to reflect my changes to the vacuole, regardless it is now done and seems to work on my end.

hhyyrylainen (changes requested):
The pilus description and the list of modifiers, is now the wrong way around.
Even though there is no longer description for the pilus, the list of stats should still be after the first line

hhyyrylainen (approved):

Kasterisk:
I made at least the description label and separator hidden so there's no aforementioned padding on the pilus tooltip and dangling elements now that the bottom part is empty.

PR #2506, opened 2021-08-19
Fixed rare crash with invalid shape ids and removed useless lock

FuJa0815:
Follow up to #2391

- Removed the weird check in `OnContactEnd`.
- Removed useless lock in `ReParent`.

I've messed up with debugging in `ReParent`. I've seen a NullPointer exception on the `node.GetParent().RemoveChild(node);` line and have checked the output of `node.GetParent()`, which returned null, therefore I concluded that this had to be a threading problem, but this was wrong. The null exception actually happened in `OnContactEnd` which `RemoveChild` seems to call internally. My IDE couldn't follow the internal godot call and therefore showed my an exception on the `RemoveChild` line. The child has been removed successfully which lead to `node.GetParent()` being null, which confused me.

### DIFF
--- a/src/engine/NodeHelpers.cs
+++ b/src/engine/NodeHelpers.cs
@@ -97,25 +97,13 @@ public static class NodeHelpers
     /// </remarks>
     public static void ReParent(this Node node, Node newParent)
     {
-        // TODO: remove this lock from here. This shouldn't even help at all, unless Godot runs physics callbacks
-        // concurrently. I'm pretty sure that is not the case so this shouldn't even help at all as this won't prevent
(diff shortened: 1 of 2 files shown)


hhyyrylainen:
Are there going to be more changes in this PR, or should I merge this now (to make the last devbuild before the release)?

hhyyrylainen (approved):

FuJa0815:
> 
> 
> Are there going to be more changes in this PR, or should I merge this now (to make the last devbuild before the release)?

I've tried some more things, but nothing truly successfull. Ready to merge.

PR #2545, opened 2021-08-28
Add info text on graphs with no datasets

Kasterisk:

Graphs with empty data now shows "No Data to Show" text if they have no datasets to make it less confusing for players seeing an empty graph which currently looks bugged at a glance.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/gui_common/charts/line/LineChart.cs
```diff
@@ -693,6 +696,19 @@ private void UpdateLineColliders(string datasetName, DataPoint firstPoint, DataP
...
 
+    /// <summary>
+    ///   Draws a text on the chart clarifying that there's no data to show to the user.
+    /// </summary>
+    private void DrawNoDataText()
+    {
+        var font = GetFont("lato_small", "Label");
+
+        var position = new Vector2((drawArea.RectSize.x - font.GetStringSize(
+            TranslationServer.Translate("NO_DATA_TO_SHOW")).x) / 2, drawArea.RectSize.y / 2);
```
Maxonovien, line 707: I find the computations pretty hard to read as such with formatting

Kasterisk: I think this is due to the translation server call as this calculation is already at its simplest form, will put the translate call to a local variable.

Maxonovien: Yeah, I also find it much more readable now that computations for x aren't split on two lines, with one further sharing computations for y.

src/gui_common/charts/line/LineChart.cs
```diff
@@ -577,8 +577,11 @@ private void CreateDropDownLegend(string title)
     /// </summary>
     private void RenderChart()
     {
-        if (dataSets.Count <= 0)
+        if (dataSets == null || dataSets.Count <= 0)
```
84634E1A607A, line 580: Instead of checking `dataSets.Count <= 0`, check if shown count > 0.

I'm able to produce this, where `Count = 3` but Shown ones =0
image

```suggestion
        if (dataSets?.All(d => d.Value.Draw != true) != false)
```

Maxonovien:
More on a minor GUI note, I find the font not matching super well with the others used in the stats tab, does it?

image

src/gui_common/charts/line/LineChart.cs
Kasterisk: That's a good idea, I only had the graph's available datasets in mind but not the _shown_ datas when implementing this.

Kasterisk:
> More on a minor GUI note, I find the font not matching super well with the others used in the stats tab, does it?

It uses Lato font whereas the graph's axes labels uses Jura, I suppose it does looks out of place with the rest of the labels.

src/gui_common/charts/line/LineChart.cs
```diff
@@ -577,7 +577,8 @@ private void CreateDropDownLegend(string title)
     /// </summary>
     private void RenderChart()
     {
-        if (dataSets == null || dataSets.Count <= 0)
+        // Handle empty or entirely hidden datasets
+        if (dataSets?.All(d => !d.Value.Draw) == true)
```
Maxonovien, line 581: It may be my own lack of knowledge of C# and x?. structures speaking here, but why use ==true here?

84634E1A607A: Pls use != false , because when dataset == null it was supposed to draw empty string, but now null == true returns false.

Maxonovien: Yeah, just experienced that on my own PR. Then I'm fine with the code!

FuJa0815: > It may be my own lack of knowledge of C# and x?. structures speaking here, but why use ==true here?

The ? syntax is conditional calling of the method. If the variable in front of the ? is null it won't call the method afterwards and instead automatically return null. Because Any returns a bool, which is a primitive type and therefore non nullable it automatically returns a nullable version of bool instead (`bool?`). A nullable bool can have three states: true, false or null. Thats why explicit == true is required here. Note: because of this == true is different than != false, which can be quite confusing

Maxonovien (approved):

src/gui_common/charts/line/LineChart.cs
84634E1A607A: So why not using !=false ? Should use here

src/gui_common/charts/line/LineChart.cs
```diff
@@ -577,8 +577,12 @@ private void CreateDropDownLegend(string title)
     /// </summary>
     private void RenderChart()
     {
-        if (dataSets.Count <= 0)
+        // Handle empty or entirely hidden datasets
+        if (dataSets?.All(d => !d.Value.Draw) == true)
```
Maxonovien, line 581: Good catch, just tested and it wasn't working anymore for null cases.
```suggestion
        if (dataSets?.All(d => !d.Value.Draw) != false)
```

Maxonovien (changes requested):
Needs fixes to account for empty datasets

src/gui_common/charts/line/LineChart.cs
Maxonovien: Good catch, just tested and it wasn't working anymore for null cases. Requested changes for that...

src/gui_common/charts/line/LineChart.cs
Kasterisk: Fixed it.

Maxonovien (commented):
I'm fine with the code as it is, but it would probably be much nicer if the labels were actually removed at this point when no data is shown (like when there isn't any). Would this be doable?

image

Kasterisk:
It should, I agree that it looks a bit off to have invalid 0 values shown all over the axes. As for the previous entirely hidden datasets check, I just now remembered that I had already made `VisibleDataSets` property which I could use instead of Linq. Well, I guess it's functionally the same and similar in terms of performance but it should make the if statement more easier to comprehend.

Maxonovien (approved):
Very nice with the last changes. The code looks fine and testing-wise it works well, including with previous saves.

src/microbe_stage/editor/MicrobeEditor.tscn
```diff
@@ -2151,7 +2151,7 @@ anchor_right = 0.0
 anchor_bottom = 0.0
 margin_top = 45.0
 margin_right = 303.0
-margin_bottom = 107.0
```
hhyyrylainen, line 2154: This file seems to only have margin changes, so maybe reset this to the master branch version?

hhyyrylainen (changes requested):
I fixed a problem with jura being considered a typo. Besides that I noticed that the editor scene just had the margins changed. I assume that that change would be safe to revert.

One last thing: would it be possible to try to pixel align the text label that is to be rendered? On my monitor it seems that at 720p resolution it ends up at just the wrong position and looks blurry (maximizing the window makes the text look fine).

Kasterisk:
> would it be possible to try to pixel align the text label that is to be rendered? On my monitor it seems that at 720p resolution it ends up at just the wrong position and looks blurry (maximizing the window makes the text look fine).

Not sure how I could do that as there's no option for such settings in `DrawString`. As for the text being a bit wonky in your end, could be an internal Godot rendering problem? This might be a whole another issue.

src/microbe_stage/editor/MicrobeEditor.tscn
Kasterisk: I think these changes is due to the new min size I set for the line charts, and as a result the graphs in the editor gains height a bit thus these margin changes.

hhyyrylainen: Ah, I can accept that explanation.

hhyyrylainen:
I *think* it looks like the font starts at like a half a pixel, hence why it is blurry.
One idea: what if you take the position of the drawable surface into account and use math round to get to a whole pixel value, that way the text would end up at an integer coordinate. That could work.

Kasterisk:
Rounding the text's position values fixed it.

hhyyrylainen:
Can confirm it doesn't look blurry like before. I'm happy enough with how the text looks now (ie. it doesn't look worse than what Godot Label's look at various resolutions).

hhyyrylainen (approved):
I think all that's left is to resolve the merge conflict and this is good to go.

PR #2546, opened 2021-08-28
2544 enemy cell in colony

### LINKED ISSUE #2544: Enemy cell bound in colony
**Important**
Please only submit technical bug reports here. This is not a place to make feature requests
or suggestions. For that please visit our community forums:
https://community.revolutionarygamesstudio.com/

**Describe the bug**

It was reported that a player was able to bind with an enemy cell.

**To Reproduce**
Possibly activate binding and try to bind with a friendly and an enemy cell at the same time. The reporting player said they were effectively doing this.
(Wasn't replicated)

**Expected behavior**
Not being able to bind with the enemy cell

**Screenshots**
image

**System info (please complete the following information):**
Not reported

**Logs**
Not given

**Additional context**
A save file after this has occured is joined to the issue.
[help_i_binded_to_a_enemy_cell.zip](https://github.com/Revolutionary-Games/Thrive/files/7070666/help_i_binded_to_a_enemy_cell.zip)

Maxonovien:

This PR aims at preventing binding a forbidden cell in your colony: as a frame passes during the operation, a new check is applied to avoid changes between checks leading to undesired behaviors.

**Related Issues**

Fixes #2544 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/microbe_stage/Microbe.cs
```diff
@@ -2545,6 +2545,12 @@ private void OnBodyExitedEngulfArea(Node body)
         }
     }
 
+    private bool IsBindingForbidden(Microbe other)
```
hhyyrylainen, line 2548: Usually in programming negative condition checks are harder to reason about, so maybe this should be `CanBind` and is used to check if all the binding checks *pass*

sirdanforth:
Checked 3 things,
1.) Player cell by itself does not bind with enemy cells (PASS)
2.) Player cell with 2-5 other colony members does not bind with enemy cells (PASS)
3.) Player cell with 5-10 other colony members does not bind with enemy cells (PASS)
I couldn't get any of the enemy cells to bind to my colony. If I'm missing something let me know.

src/microbe_stage/Microbe.cs
Maxonovien: The issue I had with naming was that there were already a few functions related in name, like `CheckBinding`, so I though `CanBind` may not be super speaking as a difference.

But I can definitely swap to positive condition instead, and rename it `IsAllowedToBind` (or `canBind` if you deem that the name will not be confused with `checkBinding`)

Maxonovien:
> 
> 
> Checked 3 things,
> 1.) Player cell by itself does not bind with enemy cells (PASS)
> 2.) Player cell with 2-5 other colony members does not bind with enemy cells (PASS)
> 3.) Player cell with 5-10 other colony members does not bind with enemy cells (PASS)
> I couldn't get any of the enemy cells to bind to my colony. If I'm missing something let me know.

Ideally, if you could replicate bug #2544 on master branch and then no longer have it on this branch, this would be great!

84634E1A607A:
image

Sadly, in this PR I can even bind to cells with no Binding agent.

src/microbe_stage/Microbe.cs
```diff
@@ -2566,7 +2566,7 @@ private void CheckBinding()
 
         // We check here if we can bind, not to waste computational power with Invoke,
         // although we'll still have to recheck it later.
-        if (IsBindingForbidden(other))
+        if (CanBindToMicrobe(other))
```
84634E1A607A, line 2569: You are!

src/microbe_stage/Microbe.cs
```diff
@@ -2591,7 +2591,7 @@ private void BeginBind()
 
         // A frame has passed since last check, so we get sure nothing got changed in between.
         // We don't pass other to the *invoked* method for computational purposes.
-        if (IsBindingForbidden(other))
+        if (CanBindToMicrobe(other))
```
84634E1A607A, line 2594: And here too!

84634E1A607A (changes requested):

src/microbe_stage/Microbe.cs
Maxonovien: Yeah, I renamed it but forgot to account for the change in the logic. Don't code at 2am, folks. 
(I'm waiting for the checks to be done and I'll request review then)

src/microbe_stage/Microbe.cs
```diff
@@ -2582,6 +2589,11 @@ private void BeginBind()
             return;
         }
 
+        // A frame has passed since last check, so we get sure nothing got changed in between.
+        // We don't pass other to the *invoked* method for computational purposes.
+        if (!CanBindToMicrobe(other))
```
84634E1A607A, line 2594: Merge this with `var other` on line 2578 too.

84634E1A607A (changes requested):
This solves the problem, but I've got some gameplay experience suggestions

src/microbe_stage/Microbe.cs
Maxonovien: The thing is that we have error printing above, so we may want to keep them for defensiveness. However, it may in fact come in handy to avoid wasting a binding frame... 

I suppose I'll do the change because the latter seems outweight the former in terms of benefices.

Maxonovien:
I went for your changes but changed my mind to keep frame duration constant. This is at the edge of the PR's scope, so I'll stick with what's it currently there. If we decide to go for this change, it may be refactored in a later PR.

84634E1A607A:
> keep frame duration constant

I don't really understand how this do with frame duration.

Maxonovien:
> 
> 
> > keep frame duration constant
> 
> I don't really understand how this do with frame duration.

If I'm not mistaken, FirstOrDefault(T) will go to the end of the set if no element satisfies predicate T. So this is O(n)
On the contrary, FirstOrDefault() will only take the first element (if there's any). So this is O(1).

So it may take more frames, but frames will take equal time. There's a trade-off to have here, and I think it should be right to leave the decision to the final review of the programming lead.

hhyyrylainen:
I think the number of microbes that are touched at once, is at most like 10. Checking only the first element or all 10, is pretty much the same. So I'd prefer to check all because that will avoid strange binding bugs where a player reports that randomly they don't manage to bind (If I understood right that the other touching microbes need to be removed from the list before the other items would be checked).

Maxonovien:
> If I understood right that the other touching microbes need to be removed from the list before the other items would be checked

Yes, indeed this may also happen. If we all agree on the trade-off chosen, I'm moving to bring this changes to the PR.

84634E1A607A (approved):
I think this is simple but effective then.

84634E1A607A:
> So it may take more frames
> frames will take equal time

As far as I know, these codes will be run in one frame. The only side-effect of too complicated and time-consuming operations will only cause long frame time and low frame rate.

Maxonovien:
What I meant is that looping through the set will take more frames if we only check the first element each frame. But if wee loop through the set (at worst) each frame, we have frames taking more time (and potentially unequal frame duration)

Although, as hh pointed it out, as long as we make sure not to have to many cells in the set (i.e. touching cells; and we might have surprises for large colonies though), this is not a major issue.

sirdanforth:
> > Checked 3 things,
> > 1.) Player cell by itself does not bind with enemy cells (PASS)
> > 2.) Player cell with 2-5 other colony members does not bind with enemy cells (PASS)
> > 3.) Player cell with 5-10 other colony members does not bind with enemy cells (PASS)
> > I couldn't get any of the enemy cells to bind to my colony. If I'm missing something let me know.
> 
> Ideally, if you could replicate bug #2544 on master branch and then no longer have it on this branch, this would be great!

Alright, So I tried to replicate this bug on the master branch but no avail. No matter how I tried it the enemy cells would not bind with me or my colony no matter the size. Is this what you wanted and is there anything else I missed?

Maxonovien:
> Alright, So I tried to replicate this bug on the master branch but no avail. No matter how I tried it the enemy cells would not bind with me or my colony no matter the size. Is this what you wanted and is there anything else I missed?

No, normally you should be able to replicate on master. As said on the bug report it seems that the way to replicate it is to have two cells close to one another. @84634E1A607A apparently managed to do it on master, and there are save files related to it on #2544 . What needs to be done for external confirmation is to be able to replicate on master, but not here. I don't know if Ajax did so with the latest version of this PR.

84634E1A607A:
Not yet. I went to sleep.

> with the latest version of this PR.

But master hasn't changed much, and those bugged saves are created on master branch yesterday. I think there's no need to check again.

Maxonovien:
> But master hasn't changed much, and those bugged saves are created on master branch yesterday. I think there's no need to check again.

I meant the latest version of this PR, which is not on master yet. In other words, am I good to call hh's final review then?

sirdanforth:
> > Alright, So I tried to replicate this bug on the master branch but no avail. No matter how I tried it the enemy cells would not bind with me or my colony no matter the size. Is this what you wanted and is there anything else I missed?
> 
> No, normally you should be able to replicate on master. As said on the bug report it seems that the way to replicate it is to have two cells close to one another. @84634E1A607A apparently managed to do it on master, and there are save files related to it on #2544 . What needs to be done for external confirmation is to be able to replicate on master, but not here. I don't know if Ajax did so with the latest version of this PR.

Oh my bad! I'm still getting used to the lingo used so I'm sorry if I'm slow. Do you still need me to test this PR or is it safe for me to move on to another one?

Maxonovien:
Given that a review to hh was requested, I assume Ajax was satisfied by the testing? Although the box isn't checked on the list. 
As testing is a limiting factor, what I propose is to assume it is okay and that we are just awaiting final review. As such, you'll be able to move on to testing other PRs, and I'll come back to you should the need arise.

hhyyrylainen:
I'm pretty happy with the reasoning *why* the bug may happen very rarely. And at least this PR does fix something. Though, we can still be on the lookout if during any testing before 0.5.6 this issue pops up again.

src/microbe_stage/Microbe.cs
```diff
@@ -2545,6 +2545,12 @@ private void OnBodyExitedEngulfArea(Node body)
         }
     }
 
+    private bool CanBindToMicrobe(Microbe other)
+    {
+        // Cannot hijack the player, other species or other colonies (TODO: yet)
+        return other?.IsPlayerMicrobe == false && other.Colony == null && other.Species == Species;
```
hhyyrylainen, line 2551: I think it would be clearer to have `if (other == null)` as a separate check first, so there wouldn't be a need to have this safe navigation operator used here `other?`

hhyyrylainen: Maybe even the entire null check should be removed as we are pretty sure right that touchedMicrobes does not contain any null entries (as I've been told there should already be check in the physics callback adding things to it to ensure it doesn't put a null in the list).

hhyyrylainen (changes requested):
Otherwise looks good, but I'd like one tiny code cleanup done on this (which might be literally just changing one character).

src/microbe_stage/Microbe.cs
Maxonovien: `other` is actually extracted from a filtered list. If no touched microbe matches the filter, then `other == null`.

As for the use of `other?`, it has the merit to cover the cases of "improper" calls to this method (with null other). But, on the other hand, in such case we may want to just print an error in the code... or just let the game crash to remark it even more....

So, I'm happy with either solution.

hhyyrylainen (approved):

PR #2547, opened 2021-08-28
fixed moving to any patch in freebuild

### LINKED ISSUE #2115: Make patch move in freebuild allow moving to any reachable patch
So instead of requiring the player to select each subsequent patch in a path and click the move button, in freebuild any patch (from the current patch where there is a viable path) can be moved to.

Jonathon594:

Added recursive method to find all connected patches, if the target patch is in the list, allows travel.

**Related Issues**

fixes #2115 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


Jonathon594:
Is this a good way to do this? I've used this recursive approach before and it works fairly efficiently in Java. I've tested it, it works fine as far as I can tell. I don't think there's any possible way this could mess with saving/loading.

hhyyrylainen:
This could be rewritten as an enumerator with "yield return" (that approach is used in the microbe spawner) so that a list object doesn't need to be built. Instead the LINQ contains can directly just iterate the sequence once without having to make a concrete list.

Jonathon594:
Yeah I was loosing my mind look at IEnumerable going "but how do I keep track of what I've already put.." I converted it to HashSet, does that look better? Personally I don't know all the differences of the List/Set types yet. Probably should look into that.

src/microbe_stage/Patch.cs
```diff
@@ -79,6 +79,35 @@ public bool AddNeighbour(Patch patch)
         return Adjacent.Add(patch);
     }
 
+    /// <summary>
+    /// Checks all neighbors recursively to find all connected patch nodes
+    /// </summary>
```
Maxonovien, line 84:
```suggestion
    /// <summary>
    ///   Checks all neighbors recursively to find all connected patch nodes
    /// </summary>
```

src/microbe_stage/Patch.cs
```diff
@@ -79,6 +79,35 @@ public bool AddNeighbour(Patch patch)
...
+    {
+        HashSet<Patch> resultSet = new HashSet<Patch>();
+        CollectNeighbors(this, ref resultSet);
+
+        return resultSet;
+    }
+
+    /// <summary>
+    /// Adds all neighbors recursively to the referenced <see cref="HashSet{T}"/>
+    /// </summary>
```
Maxonovien, line 96:
```suggestion
    /// <summary>
    ///   Adds all neighbors recursively to the referenced <see cref="HashSet{T}"/>
    /// </summary>
```

Maxonovien (changes requested):
This seems like a nice implementation of a DFS. Maybe it could be worth it to mention that we use this kind of recursive search somewhere in the functions documentation?

The request for changes is only about XML formatting, otherwise I'm very happy with the changes brought to the code.

Jonathon594:
Done, and done.

Maxonovien (approved):

src/microbe_stage/Patch.cs
```diff
@@ -79,6 +79,35 @@ public bool AddNeighbour(Patch patch)
...
+
+        return resultSet;
+    }
+
+    /// <summary>
+    ///   Adds all neighbors recursively to the referenced <see cref="HashSet{T}"/>
+    /// </summary>
+    /// <param name="patch">The <see cref="Patch"/> to start from</param>
+    /// <param name="set">The <see cref="HashSet{T}"/> to add to</param>
+    public void CollectNeighbors(Patch patch, ref HashSet<Patch> set)
```
hhyyrylainen, line 99:
```suggestion
    public void CollectNeighbours(Patch patch, HashSet<Patch> neighbours)
```

Besides that suggestion: I have a question, why does this take in a patch? This is a non-static method, why not use `this.Adjacent`. Why is it like this?

src/microbe_stage/Patch.cs
```diff
@@ -79,6 +79,35 @@ public bool AddNeighbour(Patch patch)
...
+    {
+        foreach (var neighbour in patch.Adjacent)
+        {
+            if (!set.Contains(neighbour))
```
hhyyrylainen, line 103: HashSet.Add already returns a boolean indicating whether the element was added, we can save one method call like that (so instead of first contains and then add, just directly add and check the return value).
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.add?view=net-5.0

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -1021,6 +1021,10 @@ public bool IsPatchMoveValid(Patch patch)
         if (patch == playerPatchOnEntry)
             return true;
 
+        // We are freebuilding so check if the target patch is connected by any means
```
hhyyrylainen, line 1024:
```suggestion
        // If are freebuilding, check if the target patch is connected by any means, then it is allowed
```

hhyyrylainen (changes requested):

src/microbe_stage/Patch.cs
Jonathon594: Because this method calls itself on line 106 with the neighbour passed in as what to collect from. I probably should have made it static?

hhyyrylainen: Yes, that would make sense.

Jonathon594: I've converted it to static and moved it above all non static members. Along with all the other requested changes. It should look good now.

hhyyrylainen (approved):
lgtm

Issue #1710, opened 2020-10-11
At Cursor feature doesn't detect any toxins


punctdan:
When the cursor hovers over environmental or fired toxins, "At Cursor:" doesn't show you what it is.

hhyyrylainen:
This would be a pretty nice addition.

Kasterisk:
Perhaps the navigator panel could also include any other world entities (like chunks or cell corpses)? Could make identifying things more easier I think

PR #2550, opened 2021-08-29
Update mouse hover panel on pause

### LINKED ISSUE #2531: Mouse hover panel should update while the game is paused
So that the player can pause the game to plan and check things with the mouse. I think the reason why it doesn't is that it is in the main microbe stage update code, which is paused when the game is paused (I mean otherwise many of the game systems wouldn't pause). Probably needs some bit of code architecture design on how to do this sensibly.

Kasterisk:

Separated mouse hover detection logic to a dedicated class called `PlayerHoverInfo` to allow it to keep processing even while paused. Taking some inspiration from the old pre-Godot `player_hover_info.cpp` file.

**Related Issues**

Closes #2531 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+        HoveredCompounds = cloudSystem.GetAllAvailableAt(camera.CursorWorldPos);
+
+        var microbes = GetTree().GetNodesInGroup(Constants.AI_TAG_MICROBE);
+
+        foreach (var microbe in HoveredMicrobes)
+            microbe.IsHoveredOver = false;
+
+        HoveredMicrobes.Clear();
+
+        foreach (Microbe entry in microbes)
```
Maxonovien, line 41: I feel like entry is too generic a name. As the loop is short, this is minor, but it could still benefit a name change, I think.

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+    ///   All compounds the user is hovering over.
+    /// </summary>
+    public Dictionary<Compound, float> HoveredCompounds { get; private set; } = new Dictionary<Compound, float>();
+
+    /// <summary>
+    ///   All microbes the user is hovering over.
+    /// </summary>
+    public List<Microbe> HoveredMicrobes { get; private set; } = new List<Microbe>();
+
+    public bool IsHoveringAnyEntity => HoveredCompounds.Count > 0 || HoveredMicrobes.Count > 0;
```
Maxonovien, line 22: Are compound clouds actually considered entity? If not, this name might be misleading, but if it's clear that compound clouds can fall under this name I'm happy with it.

Maxonovien (changes requested):

src/microbe_stage/PlayerHoverInfo.cs
hhyyrylainen: my suggestion: `foreach (var microbe in microbes)`

I think this might be some of my first C# code that I wrote for Thrive, that's why it is pretty rough...

src/microbe_stage/PlayerHoverInfo.cs
hhyyrylainen: Technically the individual clouds aren't but the game objects holding the fluid simulations technically are (at least I think I've called them entities in a comment or two). `IsHoveringOverAnything` might be a more neutral name (and suitable for https://github.com/Revolutionary-Games/Thrive/issues/1710 as well)

src/microbe_stage/PlayerHoverInfo.cs
Maxonovien: Yeah, entry is a very common var name in foreach loops in the project, I actually supposed it was some automation result. Could use some refactor, perhaps.

I second this choice, but note the loop right above also uses `microbe` for vars looped from another set.

hhyyrylainen: If it's already used, maybe both loop variables need renaming to reflect their relationship with each other...

Maxonovien: The loop above loops through microbes in HoveredMicrobes list (to clear them properly from before update), so I'd suggest `hoveredMicrobe` and `worldMicrobe`. But this is first-thought material, so if anyone comes with a better idea I'm all open.

Kasterisk: Maybe `hoveredMicrobe` for the hovered ones and `microbe` for world microbes, also I'll probably rename `microbes` list to `allMicrobes`.

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System.Collections.Generic;
+using Godot;
+
+/// <summary>
+///   A system that manages reading what the player is hovering over.
```
hhyyrylainen, line 5:
```suggestion
///   A system that manages detecting what the player is hovering over with the cursor.
```

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+/// </summary>
+public class PlayerHoverInfo : Node
+{
+    private MicrobeCamera camera;
+    private CompoundCloudSystem cloudSystem;
+
+    /// <summary>
+    ///   All compounds the user is hovering over.
+    /// </summary>
+    public Dictionary<Compound, float> HoveredCompounds { get; private set; } = new Dictionary<Compound, float>();
```
hhyyrylainen, line 15: I think this could have the set removed as this should never be re-assigned, right?

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    /// <summary>
+    ///   All microbes the user is hovering over.
+    /// </summary>
+    public List<Microbe> HoveredMicrobes { get; private set; } = new List<Microbe>();
```
hhyyrylainen, line 20: same here

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+        HoveredMicrobes.Clear();
+
+        foreach (Microbe entry in microbes)
+        {
+            var distance = (entry.GlobalTransform.origin - camera.CursorWorldPos).Length();
+
+            // Find only cells that have the mouse
+            // position within their membrane
+            if (distance > entry.Radius + Constants.MICROBE_HOVER_DETECTION_EXTRA_RADIUS)
+                continue;
```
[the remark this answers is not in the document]
hhyyrylainen: I think that even multiplying the radius by radius here is probably faster than a single Length call as I assume a float multiply is way, way faster than a square root operation.

hhyyrylainen (changes requested):
Besides what Maxonovien commented, just a few comments. Functionality and save compatibility seemed fine.

src/microbe_stage/PlayerHoverInfo.cs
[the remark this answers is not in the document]
Kasterisk: Makes sense, though I do think we need to overhaul this logic in general eventually. I imagine there would be a noticeable performance impact if we keep the same logic for #1710 as the number of nodes to iterate probably doubles the usual amount. I don't currently have a better idea in mind for optimizations but for CollisionObjects like rigidbodies at least, they have `mouse_entered` and `mouse_exited` signals which we could probably utilize in some way.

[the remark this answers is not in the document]
hhyyrylainen: I have a hunch that the mouse enter callback stuff requires enabling raycasting that Godot does in the background. I guess if our entity amount is huge then, the physics engine might be more performant by being able to filter out physics bodies to check with a quad tree.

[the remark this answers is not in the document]
Kasterisk: > I think that even multiplying the radius by radius here is probably faster than a single Length call as I assume a float multiply is way, way faster than a square root operation.

Yeah, it's what I'm going for (will push the commit in a bit).

src/microbe_stage/PlayerHoverInfo.cs
Kasterisk: Indeed so.

Kasterisk: Wait nevermind, the compouds list specifically needs to be re-assigned every frame due to the `GetAllAvailableAt` call, so we need it be reassignable it seems like.

hhyyrylainen: Well at least the microbe list doesn't need to be reassigned, right?

Kasterisk: Yep, we only need it for the compounds list.

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        HoveredMicrobes.Clear();
+
+        foreach (Microbe microbe in allMicrobes)
+        {
+            var distance = (microbe.GlobalTransform.origin - camera.CursorWorldPos).LengthSquared();
+
+            // Find only cells that have the mouse
+            // position within their membrane
+            if (distance > microbe.RadiusSquared + Constants.MICROBE_HOVER_DETECTION_EXTRA_RADIUS)
```
hhyyrylainen, line 47: One more small thing should MICROBE_HOVER_DETECTION_EXTRA_RADIUS be also doubled now that we use squared distances? (edited in Constants.cs)

hhyyrylainen (changes requested):

src/microbe_stage/PlayerHoverInfo.cs
Kasterisk: Oh right it should, wondered why the detection are doesn't feel as big as before, forgot about the constant.

hhyyrylainen (approved):
All of my comments were resolved. Just waiting for @Maxonovien to verify that their comments were addressed.

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+        var allMicrobes = GetTree().GetNodesInGroup(Constants.AI_TAG_MICROBE);
+
+        foreach (var hoveredMicrobe in HoveredMicrobes)
+            hoveredMicrobe.IsHoveredOver = false;
+
+        HoveredMicrobes.Clear();
+
+        foreach (Microbe microbe in allMicrobes)
+        {
+            var distance = (microbe.GlobalTransform.origin - camera.CursorWorldPos).LengthSquared();
```
Maxonovien, line 43: Well, now this is a `distanceSquared`, isn't it?

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -699,7 +699,7 @@ private void UpdateHoverInfo(float delta)
         }
 
         // Show the species name of hovered cells
-        foreach (var entry in stage.MicrobesAtMouse)
+        foreach (var entry in stage.HoverInfo.HoveredMicrobes)
```
Maxonovien, line 702: As we discussed loop variable name already, perhaps it would be worth to use the opportunity to rename the var away from `entry`, wouldn't it?

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -674,7 +674,7 @@ private void UpdateHoverInfo(float delta)
             hoveredCompoundsContainer.GetParent<VBoxContainer>().Visible = true;
 
             // Create for each compound the information in GUI
-            foreach (var entry in stage.CompoundsAtMouse)
+            foreach (var entry in stage.HoverInfo.HoveredCompounds)
```
Maxonovien, line 677: same here

src/microbe_stage/Microbe.cs
```diff
@@ -323,6 +323,12 @@ public float Radius
         }
     }
 
+    /// <summary>
+    ///   Returns a squared value of <see cref="Radius"/>.
+    /// </summary>
+    [JsonIgnore]
+    public float RadiusSquared => Radius * Radius;
```
Maxonovien, line 330: Given that it will be called each frame, I wonder if storing the value would not be interesting for computation time. But I can't really affirm that float * float multiplication is *that* costly, so if you feel it is better as is I have no opposition.

Maxonovien (changes requested):
My comments were almost fully adressed indeed (I'd still like to discuss radius*radius vs stored value, but I won't force it either), but there are changes related to what we discussed that could be worth adressing as this PR comes in. 

Once that is settled, it should be all good for me

src/microbe_stage/Microbe.cs
hhyyrylainen: The argument here is probably that caching it will be a bit tricky, It'll need to be in Membrane.cs as that's the only thing that knows when Radius becomes dirty and is recomputed. Much simpler to just go with this code for now and come back to this if it turns out to be problematic.

Maxonovien: Alright, didn't think of that. The argument makes sense, I'm fine with keeping it as such.

Maxonovien (approved):
Everything seems to be adressed. Nice work!

PR #2554, opened 2021-08-31
Added more nullchecking to the JSON serializer

hhyyrylainen:
should fix all the warnings that popped up after the previous round of CI tool (and maybe dependency updates)

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/saving/serializers/ThriveJsonConverter.cs
+++ b/src/saving/serializers/ThriveJsonConverter.cs
@@ -347,9 +347,14 @@ public abstract class BaseThriveConverter : JsonConverter
         // Detect ref to already loaded object
         var refId = item[REF_PROPERTY];
 
-        if (refId != null)
+        if (refId is { Type: JTokenType.String })
         {
-            return serializer.ReferenceResolver.ResolveReference(serializer, refId.Value<string>());
+            var refValue = refId.Value<string>();
+
+            if (refValue == null)
+                throw new JsonException($"{REF_PROPERTY} read as string failed");
+
+            return serializer.ReferenceResolver.ResolveReference(serializer, refValue);
         }
 
         var objId = item[ID_PROPERTY];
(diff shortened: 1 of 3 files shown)


src/saving/serializers/ThriveJsonConverter.cs
```diff
@@ -347,23 +347,34 @@ public static bool IsIgnoredGodotMember(string name, Type type)
...
         var refId = item[REF_PROPERTY];
 
-        if (refId != null)
+        if (refId is { Type: JTokenType.String })
         {
-            return serializer.ReferenceResolver.ResolveReference(serializer, refId.Value<string>());
+            var refValue = refId.Value<string>();
+
+            if (refValue == null)
+                throw new JsonException($"{REF_PROPERTY} read as string failed");
```
Maxonovien, line 355: This very code is repeated 3 times with a single variable name, so we might want to factor it?

Maxonovien (commented):
It seems to work fine, both for JetBrains and in-game saving an loading. 

Given that a hotfix is needed, I'm open to this being considered an approval, but I just have a small factoring proposal if you have time to do it. Otherwise, leaving it for refactor with a TODO can be ok to me given that it is a bottleneck for my PR as of now.

src/saving/serializers/ThriveJsonConverter.cs
hhyyrylainen: Well I considered suppressing the warning because Value<string> should *not* be able to ever return null after there is that check that the token type is a string. I thought it would be better like this.

Maxonovien: Hmmm... I guess it would be a bit non-sensical to factor just to add a throwing exception that ismostly here to appease the checks indeed. That's a sufficient argument for me to accept this design.

hhyyrylainen: I already rewrote this as an extension method (though without the per-key specialized error message, but that shouldn't matter as I said, this condition should never be reached).

Maxonovien: > Value should not be able to ever return null after there is that check that the token type is a string

Perhaps it could be worth to mention this in a comment though?

hhyyrylainen: I mean there are `.Value` calls over the codebase, I don't feel like explaining that in a comment here that that conversion succeeds as long as you have checked the json token type beforehand, would help as this is like a general JSON library usage thing.

Maxonovien: Sounds a fair argument to me.

Maxonovien (approved):
I'm fine with the code, and I encountered no issue with testing (except for my usual Russian roulet thread crashes), so I think this can be considered an approval for both programming and testing.
___
Testing process:
1. Start new game
2. Enter/exit editor
3. Save editor exiting auto-save from 2.
4. Saved game
5. Loaded save from 4.
6. Loaded older save
7. Saved from load in 6.
8. Load save from 7.

PR #2564, opened 2021-09-03
Save patch manager current patch

### LINKED ISSUE #2556: Patch move is not detected if a save made in the editor is loaded
Looks like the patch cleanup on patch change doesn't trigger if:
- You enter the editor
- Make a save
- Load that save
- Move to a new patch in the patch map
- Exit the editor

This code doesn't trigger:
https://github.com/Revolutionary-Games/Thrive/blob/71fcf56153c5888a3f923177164b1e715f768f8d/src/microbe_stage/PatchManager.cs#L46

Additional info: https://github.com/Revolutionary-Games/Thrive/pull/2486#issuecomment-909479798

hhyyrylainen:

In order to properly detect when patch changes in a save made in the editor

**Related Issues**

closes #2556

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/Thrive.csproj
+++ b/Thrive.csproj
@@ -238,6 +238,7 @@
     <Compile Include="src\microbe_stage\MicrobeAI.cs" />
     <Compile Include="src\microbe_stage\MicrobeAICommonData.cs" />
     <Compile Include="src\saving\FileHelpers.cs" />
+    <Compile Include="src\saving\IChildPropertiesLoadCallback.cs" />
     <Compile Include="src\saving\IGodotEarlyNodeResolve.cs" />
     <Compile Include="src\saving\ILoadableGameState.cs" />
     <Compile Include="src\saving\InProgressLoad.cs" />
--- a/src/microbe_stage/MicrobeStage.cs
+++ b/src/microbe_stage/MicrobeStage.cs
@@ -28,6 +28,9 @@ public class MicrobeStage : NodeWithInput, ILoadableGameState, IGodotEarlyNodeRe
     private SpawnSystem spawner;
 
     private MicrobeAISystem microbeAISystem;
+
+    [JsonProperty]
+    [AssignOnlyChildItemsOnDeserialize]
     private PatchManager patchManager;
 
     private DirectionalLight worldLight;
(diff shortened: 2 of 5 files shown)


Kasterisk (approved):
`previousPatch` seems to be saved correctly. As in relation to #2486 I believe the double patch settings update call still needs to be fixed in some way so the patch name can be properly shown on loaded editor save after exiting the editor.

hhyyrylainen:
Only one of the calls will result in detection of a patch change. So as your code relies on the patch manager saying if the current patch changed it still should be safe. The only slight issue may be that older saves might be a bit janky with the patch name popup but I don't immediately have an idea how to easily get around that.

Kasterisk:
The tricky part with that is the ordering of which gets called first though. The patch name is only shown if `returningFromEditor` parameter is true and since the only place where it's set true is `OnReturnFromEditor` which is called after the patch settings update in `SetupStage`, it still doesn't get the chance to know what's the actual previous patch.

hhyyrylainen:
Okay, what if the logic is reversed? We give the PatchManager an action it calls when the patch is changed, and that is hooked up to the HUD method to show the patch name. There shouldn't be a problem as the HUD reference is looked up before the PatchManager object is created (and neither of them are JSON saved or loaded so there should be no issues regarding that).

Kasterisk:
Maybe, so that will require the HUD to manually know player is returning from editor right? Will look into ways of fixing this more later.

hhyyrylainen:
What about having the callback trigger here: https://github.com/Revolutionary-Games/Thrive/blob/71fcf56153c5888a3f923177164b1e715f768f8d/src/microbe_stage/PatchManager.cs#L48 ?
I think that might work for all cases except old saves, but as I said I don't think that's too important to fix in this case.

Kasterisk:
Yeah I think that might be a good place.

Kasterisk:
Wait, I just realized also showing the patch name in the first patch settings update call would be the simplest solution... It won't get shown again in the second update call because `previousPatch` will already be the same to the current one, likewise when quick saving/loading in the stage as it's already saved.

PR #2569, opened 2021-09-03
Added a label in options menu showing how complete selected language is

### LINKED ISSUE #2033: Add some kind of warning below language selection when unfinished
I think something like 30-40% not translated strings in a language should warrant a label below the language selection becoming visible that warns that the translation is unfinished and maybe encourages the user to help out.

Additionally there should be a script that detects which locales are quite untranslated. Basically it should be a ruby script that produces a JSON file by reading the .po files and determining which message bodies are empty vs how many have text (and are not marked fuzzy).

hhyyrylainen:

**Related Issues**

closes #2033

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


simulation_parameters/TranslationsInfo.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Collections.Generic;
+using Newtonsoft.Json;
+
+public class TranslationsInfo : IRegistryType
+{
+    [JsonProperty]
+    public Dictionary<string, float> TranslationProgress { get; private set; }
+
+    /// <summary>
+    ///   Unused
```
Kasterisk, line 10: This seems auto-generated?

Kasterisk (changes requested):

simulation_parameters/TranslationsInfo.cs
hhyyrylainen: Nope, copy-pasted from another system, this is needed because `IRegistryType` requires this.

Kasterisk (approved):

PR #2571, opened 2021-09-03
Doubled cheat cloud density and moved to constants

FuJa0815:

In my opinion the cheat clouds are too weak before. I've doubled the amount in this PR. Do you agree?


hhyyrylainen:
I didn't playtest, but I think doubling the amount you get with the cheat clouds should be fine. For glucose I've found it to be pretty onpoint (the existing value) but for ammonia and phosphates it's definitely too slow to test organelle duplication in a reasonable time. So, if the isn't a mess, I'm fine with this change, but I'll ask some other people to review this.

Kasterisk (approved):
I've fiddled with the cheat cloud amount before and gave it a much higher value. This should be fine.

PR #2572, opened 2021-09-03
Fixed diff check on windows

### LINKED ISSUE #2515: Fix detecting formatter made changes on Windows
On windows `git diff --stat` seems to report all kinds of garbage like PIDs and LF CRLF change stuff: https://github.com/Revolutionary-Games/Thrive/pull/2459#issuecomment-903083539

So check formatting should be updated so that the comparison cleans out the created diffs by removing blank lines and any lines containing any of the following: `The file will have its original line endings`, `warning: .* will be replaced by .* in` and `pid 3404 exit 0`
then the diff comparison should also work on windows... though someone on windows will need to program this change to be able to test this.

FuJa0815:

The `core.safecrlf=false` options surpresses the `warning: LF will be replaced by CRLF in` output.
With `-c` you can override a git config for one command

**Related Issues**

closes #2515

### DIFF
--- a/check_formatting.rb
+++ b/check_formatting.rb
@@ -657,7 +657,7 @@ end
 def run_cleanup_code
   return if skip_jetbrains?
 
-  old_diff = runOpen3CaptureOutput 'git', 'diff', '--stat'
+  old_diff = runOpen3CaptureOutput 'git', '-c', 'core.safecrlf=false', 'diff', '--stat'
 
   params = [cleanup_code_executable, 'Thrive.sln', '--profile=full_no_xml']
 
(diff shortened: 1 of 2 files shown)


hhyyrylainen:
Does this fix the problem with `pid 3404 exit 0` getting printed on Windows?

FuJa0815:
> Does this fix the problem with `pid 3404 exit 0` getting printed on Windows?

I haven't seen this message and have not found an instance of this message anywhere on the internet. I have no idea where this may come from.

hhyyrylainen:
Well @84634E1A607A got this message as part of the output:
https://github.com/Revolutionary-Games/Thrive/pull/2459#issuecomment-903083539

which is the reason why in the issue you linked this PR as closing I wrote that that message needs to be ignored as well.

84634E1A607A:
It's a ruby one. It seems that it doesn't cause problems though.

hhyyrylainen (approved):
To me this looks fine and still works. I'll assume this works for windows people so I should merge this.

PR #2573, opened 2021-09-03
Fixed loading text not updating when pressing Esc

### LINKED ISSUE #2501: Pressing Esc while loading pauses loading
**Describe the bug**

When in the loading screen you can press Esc and the text won't update anymore. Pressing Esc resumes the text updating. While in this Esc state the game won't finish loading until you end this buggy state via pressing Esc again.

FuJa0815:

The MicrobeEditor (which updates the loading description text) is now initially in Process-Pausemode. Once loading is complete it changes to Inherit-Pausemode

**Related Issues**

closes #2501

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
I think it would be much better if the pause menu would be prevented from opening while the game is loading.

hhyyrylainen:
I just looked through the code and it looks fine to me and did a bit of playtesting. So once my one concern is addressed this is fine by me to be merged: https://github.com/Revolutionary-Games/Thrive/pull/2573#commitcomment-55986459

84634E1A607A (dismissed):
When you press Editor button then F1 or Esc before animation completes, Editor gets stuck -- Without recovery!

hhyyrylainen:
Sounds like a pretty serious problem and should be fixed. But maybe my suggestion that *closing* the pause menu should always be allowed to avoid very serious getting stuck problems like that.
Thanks for testing this!

src/gui_common/TransitionManager.cs
```diff
@@ -36,6 +37,7 @@ private TransitionManager()
     public List<ITransition> TransitionSequence { get; } = new List<ITransition>();
 
     public bool HasQueuedTransitions => TransitionSequence.Count > 0;
+    public bool HasTransitionInProgress => TransitionSequence.Any(ts => IsInstanceValid((Node)ts));
```
hhyyrylainen, line 40: @Kasterisk you wrote this system, right? So could you comment on how possible it is that there is an *invalid* object in here?
I personally think this should just be a `HasQueuedTransitions and currently running transitions` (however that latter thing could be checked).

FuJa0815: I've slightly based this upon the `CancelQueuedTransitions` method and tested the functionality with a debugger, although I wasn't able to test all cases.

Kasterisk: The transitions queue free themselves after finished which is why there's an invalid object in `TransitionSequence`. In a way usage of `IsInstanceValid` works as a check for any active transitions or currently running transitions in a set sequence.

An alternative in this case that I could think of (haven't tested) without any need for IsInstanceValid is turning the local variable at line 160 to a private member variable maybe as currentTransition, and so you can just make the in progress check as `HasQueuedTransitions && currentTransition != null`. This variable can later be cleared in the if block at line 152 when the sequence finishes.

hhyyrylainen: Why don't the objects remove themselves from the sequence when they are done, then? That seems much better than keeping disposed objects around.

Kasterisk: Yes that's also preferable. I just didn't find a good way to do that back when implementing the class.

hhyyrylainen: I've used this kind of approach:
```
                    if (runSteps.Peek().RunStep(results))
                        runSteps.Dequeue();
```
to remove an object once it was done. A similar thing could probably be adapted here so that the transitions tell the main code when to remove it.

Kasterisk: Now that you mention `Dequeue`, I forgot to mention that `TransitionSequence` is a list and just a helper to capture `queuedTransition`'s initial sequence structure. `queuedTransitions` is the actual queue that's used for advancing through the transition sequence which also removes the current transition from itself. So maybe we can use `queuedTransitions` Count for the in progress check?

hhyyrylainen: Why there's actually multiple of those lists?
Wouldn't it be simpler if we had a method that requires you to give it a list of actions already, which would immediately go to the running things?
That would solve the problem as well, or?

Kasterisk: > Why there's actually multiple of those lists?

It's mainly due to CancelQueuedTransitions. I think I remember the problem when implementing it was that queudTransitions doesn't include the currently running transition (because of Dequeue) so what happen was all subsequent transitions got skipped but not the one that's currently running, the solution I came up with was the `TransitionSequence` to hold references to all transition objects from initial queuedTransitions...

You know, I now realize it's a bit silly. In hindsight it probably makes most sense to go with the `currentTransition` variable I mentioned earlier and then just iterate through queuedTransitions to skip all remaining transitions without needing an extra list, so only one queuedTransitions needed, probably will be more robust as well.

> Wouldn't it be simpler if we had a method that requires you to give it a list of actions already, which would immediately go to the running things?

Yes, the current system works a bit like that. We first give TransitionManager "a list" of actions by adding to the transitions queue by calling either `AddScreenFade` or `AddCutscene` and then we call StartTransitions which will immediately advance through the sequence.

FuJa0815: So, is this line fine? 😅

hhyyrylainen: I say this is not because "IsInstanceValid" is a huge design smell to me. This PR needs to be designed some other way but I think we can leave the overall approach changing to @Kasterisk

Kasterisk: I think for now it's fine to use `HasQueuedTransitions`. It kind of already functions as `HasTransitionInProgress` and is relevant until sequence is finished and the list is cleared. Any further will probably warrant a medium rework as mentioned in earlier discussions.

src/general/PauseMenu.cs
```diff
@@ -99,6 +101,25 @@ private ActiveMenuType ActiveMenu
...
 
+    /// <summary>
+    ///   If true the user may not open or close the pause menu.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Does not automatically close the pause menu when set to true.
+    ///   </para>
+    /// </remarks>
+    public bool IsPausingBlocked()
```
hhyyrylainen, line 112: This could be a get only property

src/gui_common/TransitionManager.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
```
hhyyrylainen, line 2: This added line is probably why CI fails.

hhyyrylainen (changes requested):

src/general/PauseMenu.cs
```diff
@@ -64,6 +64,33 @@ private enum ActiveMenuType
     /// </summary>
     public GameProperties GameProperties { get; set; }
 
+    public bool GameLoading { get; set; }
+
+    /// <summary>
+    ///   If true the user may not open or close the pause menu.
```
hhyyrylainen, line 70: This comment is now wrong.

hhyyrylainen (changes requested):
One tiny code related thing. I now playtested this and (with tutorials enabled) I wasn't able to open the pause menu with ESC at all...

FuJa0815:
> One tiny code related thing. I now playtested this and (with tutorials enabled) I wasn't able to open the pause menu with ESC at all...

Sorry, was only able to make swift changes and haven't had time to test them. This had a logic error.

hhyyrylainen (approved):
I didn't find anything wrong with this. I assume that other issue that someone else found is also fixed so that this is finally good to merge.

PR #2578, opened 2021-09-06
Add Floating Chunk Food Source

adQuid:

Adds food source to account for native floating chunks to the environment for patches that rely on marine snow. Selection pressure doesn't result in very interesting life, but at least the deep sea patches don't just die out once the glucose is gone.

This PR does NOT attempt to calculate chunks from existing species that would be encountered.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


buckly90:
It does seem like more species readily survive in the marine snow patches than before.

Thrive.csproj
```diff
@@ -320,7 +321,6 @@
     <Folder Include="shaders\" />
     <Folder Include="src\microbe_stage\organelle_components\" />
     <Folder Include="src\auto-evo\steps\" />
-    <Folder Include="src\auto-evo\simulation\" />
```
hhyyrylainen, line 323: Once again this folder is attempted to be deleted...

src/auto-evo/simulation/food_source/FloatingChunkFoodSource.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System.Linq;
+
+public class FloatingChunkFoodSource : FoodSource
```
hhyyrylainen, line 3: This name is not accurate because iron are chunks, so this is more accurately just the marine snow thing.

src/auto-evo/simulation/food_source/FloatingChunkFoodSource.cs
```diff
@@ -0,0 +1,59 @@
...
+    private float chunkSize;
+
+    public FloatingChunkFoodSource(Patch patch)
+    {
+        this.patch = patch;
+        biomeConditions = patch.Biome;
+
+        if (patch.Biome.Chunks.ContainsKey("marineSnow"))
+        {
+            ChunkConfiguration chunk = patch.Biome.Chunks["marineSnow"];
```
hhyyrylainen, line 19: use `TryGetValue`

src/auto-evo/simulation/food_source/FloatingChunkFoodSource.cs
```diff
@@ -0,0 +1,59 @@
...
+            chunkSize = chunk.Size;
+            totalEnergy = chunk.Compounds[glucose].Amount * Constants.AUTO_EVO_CHUNK_ENERGY_AMOUNT;
+        }
+    }
+
+    public override float FitnessScore(Species species)
+    {
+        var microbeSpecies = (MicrobeSpecies)species;
+
+        var predatorSize = microbeSpecies.Organelles.Organelles.Sum(organelle => organelle.Definition.HexCount);
```
hhyyrylainen, line 29: It's not really a predator here, right? So `scavenger`?
Also this "engulf size" calculation is quite in many places, would be probably very handy to directly have this in MicrobeSpecies, as I would guess that someone somewhere will forget to take the IsBacteria flag into account.

hhyyrylainen (changes requested):

Thrive.csproj
```diff
@@ -320,7 +321,6 @@
     <Folder Include="shaders\" />
     <Folder Include="src\microbe_stage\organelle_components\" />
     <Folder Include="src\auto-evo\steps\" />
```
adQuid, line 323:
```suggestion
    <Folder Include="src\auto-evo\steps\" />
    <Folder Include="src\auto-evo\simulation\" />
```

adQuid, line 323:
```suggestion
    <Folder Include="src\auto-evo\steps\" />
    <Folder Include="src\auto-evo\simulation\" />
```

src/microbe_stage/MicrobeSpecies.cs
```diff
@@ -36,6 +36,10 @@ public override string StringCode
     [JsonIgnore]
     public float BaseSpeed => MicrobeInternalCalculations.CalculateSpeed(Organelles, MembraneType, MembraneRigidity);
 
+    [JsonIgnore]
+    public float BaseSize => Organelles.Organelles.Sum(organelle => organelle.Definition.HexCount)
```
hhyyrylainen, line 40: I think this is too general. This should be engulf size? Or at least like "EffectiveTotalHexSize" right now this "size" is way too general to me as to what this is related to (because my first thought about size would be how large the cell is visually and that is more connected to the membrane radious).

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -1,21 +1,17 @@
...
     private float preySpeed;
-    private float preySize;
     private float totalEnergy;
 
     public HeterotrophicFoodSource(Patch patch, MicrobeSpecies prey)
     {
         this.prey = prey;
         this.patch = patch;
         preySpeed = prey.BaseSpeed;
-        preySize = prey.Organelles.Organelles.Sum(organelle => organelle.Definition.HexCount);
```
hhyyrylainen, line 18: It would help to compute these values just once instead of each time where the property is referenced it gets recomputed.

adQuid: Does the property here not work that way? https://github.com/Revolutionary-Games/Thrive/pull/2578/files/c3a05d9d6569411931c6eeafb52416763817d92e#diff-ef5f11f4e06e07f19c774dfe97a969568398f94e92a1fcff255cde04ac6e1aa7R40

If not, then I might need some C# tips so that it does.

hhyyrylainen: It does not. Literally each time when that property appears, the entire code for calculating it is re-ran, which as it looks pretty expensive of an operation is not very good. There's no inbuilt caching or anything like that in C# properties.

Properties can have a backing field and a dirty flag so that they can be recomputed only when necessary but here as it depends on the organelle list it's going to be a bit perhaps tricky to setup the callbacks to set the dirty flag properly after those change.

adQuid: I've gone ahead and reused the calculation for the prey size. I think the danger is too high to put it on the property altogether given how often the organelles are modified in the code so far, and while technically this is called O(n^2) times (each species attempting to eat all others), I'd be very concerned if N exceeded 10-20 in any given patch.

FuJa0815: > Properties can have a backing field and a dirty flag so that they can be recomputed only when necessary

Personally I love using the following syntax in case where a dirty flag is required for a caching field:

```
private type? value;
public type Property => value ??= calculation();

public void DirtyProperty() => value = null;
```

This can combine the dirty flag and the value itself into one concise variable. Of course this is only applicable when null is not a valid value.

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -25,22 +23,22 @@ public override float FitnessScore(Species species)
...
-        var predatorSize = microbeSpecies.Organelles.Organelles.Sum(organelle => organelle.Definition.HexCount);
         var predatorSpeed = microbeSpecies.BaseSpeed;
         predatorSpeed += ProcessSystem
             .ComputeEnergyBalance(microbeSpecies.Organelles.Organelles, patch.Biome,
                 microbeSpecies.MembraneType).FinalBalance;
 
         // It's great if you can engulf this prey, but only if you can catch it
         var engulfScore = 0.0f;
-        if (predatorSize / preySize > Constants.ENGULF_SIZE_RATIO_REQ && !microbeSpecies.MembraneType.CellWall)
+        if (microbeSpecies.BaseHexSize / preyHexSize >
```
hhyyrylainen, line 40: microbeSpecies.BaseHexSize is also used twice, first time here

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -70,7 +68,7 @@ public override float FitnessScore(Species species)
         pilusScore *= predatorSpeed;
 
         // Intentionally don't penalize for osmoregulation cost to encourage larger monsters
-        return behaviorScore * (pilusScore + engulfScore + predatorSize + oxytoxyScore);
+        return behaviorScore * (pilusScore + engulfScore + microbeSpecies.BaseHexSize + oxytoxyScore);
```
hhyyrylainen, line 71: and then a second time in the same method here

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
adQuid: Having taken a while to look at this, I think the best next step is to actually pull this variable back off of the MicrobeSpecies class, and add it to a new wrapper used solely by the auto-evo. This information is auto-evo's domain, and a wrapper pattern would allow us to reuse a lot more calculations, including very esoteric ones like "number of times toxin hits you before you die"

hhyyrylainen: Except the same code is used to check if something can engulf something.
It would be highly silly to decouple the auto-evo from the actual game mechanics related calculation...

adQuid: The hex size calculation is separate from base hex size. Base hex size exists only on this branch.

hhyyrylainen: What? Why is it different? The question for "can I engulf this?" should be 100% the same between auto-evo and gameplay code. Why it isn't?

adQuid: Because a Microbe can have a variable amount of organelles (it could be anywhere in the process of division), whereas a MicrobeSpecies cannot. Yes, they both divide the result by 2 if the species is bacteria, but other than that they count different lists.

hhyyrylainen: Ah yes, of course, I forgot about that.
Anyway I think this is really a *property* of the species and belongs there.

adQuid: I'll make my attempt to convince you of open closed principle at a later time. For now, I have no intent on moving this code anywhere.

hhyyrylainen: I think someone, at some point is much more likely to reimplement the hex size computation as they can't find it when it is put in the wrong place, which is just in the depths of the auto-evo system.
Anyway, I won already as you put the code in the MicrobeSpecies class already.

hhyyrylainen (approved):

PR #2588, opened 2021-09-16
updated mutation algorithm to prevent duplicating nuclei

### LINKED ISSUE #2579: Auto-evo mutations can make a duplicate of an organelle that is only allowed once
auto-evo can seemingly duplicate the nucleus organelle (and probably the binding agents as well) to create species that have two nucleai. This should be prevented.

https://github.com/Revolutionary-Games/Thrive/blob/master/src/general/Mutations.cs#L238

should be relatively easy to check if the organelle type is one that there only can be one of before duplicating

adQuid:

Updates mutation code to ensure that organelles with the unique property cannot be duplicated. I have tested this enough to ensure it doesn't crash, but not enough to ensure the original bug cannot happen.

**Related Issues**

fixes #2579 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/general/Mutations.cs
+++ b/src/general/Mutations.cs
@@ -241,7 +241,7 @@ public class Mutations
                         if (organellesThatCanBeDuplicated.Any())
                         {
                             AddNewOrganelle(mutatedOrganelles,
-                                organellesThatCanBeDuplicated.ToList().Random(random).Definition);
+                                organellesThatCanBeDuplicated.Random(random).Definition);
                         }
                         else
                         {
--- a/src/general/Mutations.cs
+++ b/src/general/Mutations.cs
@@ -240,7 +240,7 @@ public class Mutations
                             parentOrganelles.Organelles.Where(organelle => !organelle.Definition.Unique).ToList();
                         if (organellesThatCanBeDuplicated.Any())
                         {
-                            AddNewOrganelle(mutatedOrganelles, 
+                            AddNewOrganelle(mutatedOrganelles,
                                 organellesThatCanBeDuplicated.ToList().Random(random).Definition);
                         }
                         else
--- a/src/general/Mutations.cs
+++ b/src/general/Mutations.cs
@@ -236,7 +236,7 @@ public class Mutations
                     else
                     {
                         // Duplicate an existing organelle, but only if there are any organelles where that is legal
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
Fixed the issue number this is marked as solving.

src/general/Mutations.cs
```diff
@@ -235,7 +235,16 @@ private void MutateBehaviour(MicrobeSpecies parent, MicrobeSpecies mutated)
                     }
                     else
                     {
-                        AddNewOrganelle(mutatedOrganelles, parentOrganelles.Organelles.Random(random).Definition);
+                        // Duplicate an existing organelle, but only if there are any organelles where that is legal
+                        var organellesThatCanBeDuplicated = parentOrganelles.Organelles.Where(organelle => !organelle.Definition.Unique);
+                        if (organellesThatCanBeDuplicated.Any())
```
hhyyrylainen, line 240: I kinda would like to have this be instead so that first ToList() is called, on that then Count > 0 is checked, and then the random is retrieved. I suggest this because I think .Any will first enumerate at least one item, and then the ToList call will re-enumerate that, which seems to be exactly what CI is complaining here about.

hhyyrylainen (changes requested):

src/general/Mutations.cs
```diff
@@ -235,7 +235,18 @@ private void MutateBehaviour(MicrobeSpecies parent, MicrobeSpecies mutated)
...
                     else
                     {
-                        AddNewOrganelle(mutatedOrganelles, parentOrganelles.Organelles.Random(random).Definition);
+                        // Duplicate an existing organelle, but only if there are any organelles where that is legal
+                        var organellesThatCanBeDuplicated =
+                            parentOrganelles.Organelles.Where(organelle => !organelle.Definition.Unique).ToList();
+                        if (organellesThatCanBeDuplicated.Any())
+                        {
+                            AddNewOrganelle(mutatedOrganelles,
+                                organellesThatCanBeDuplicated.ToList().Random(random).Definition);
```
hhyyrylainen, line 244: Aren't you calling `ToList` here on something that is already a list?

hhyyrylainen (changes requested):

hhyyrylainen (approved):
I assume someone playtested this, but the code seems pretty straightforward.

PR #2605, opened 2021-09-23
Retool Toxin mechanics and AI

adQuid:

Does 3 things:

* Makes toxin have a maximum emission of 2 instead of a minimum emission of 1. Microbes can now release toxin below fully capacity, with a proportionally smaller model and proportionally lower damage.
* Full toxin emission now deals 30 damage (albeit for double the compound cost), and generates 50% faster (and more efficiently)
* AI modified to be braver at attacking larger prey with toxin (with high enough opportunism), and to sometimes choose to use toxin at partial emission

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -165,7 +165,7 @@ public static class Constants
     /// <summary>
     ///   The minimum amount of oxytoxy (or any agent) needed to be able to shoot.
     /// </summary>
-    public const float MINIMUM_AGENT_EMISSION_AMOUNT = 1;
+    public const float MINIMUM_AGENT_EMISSION_AMOUNT = 1.0f;
 
     /// <summary>
     ///   The time (in seconds) it takes a cloud being absorbed to halve its compounds.
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -130,7 +130,7 @@ public static class Constants
     /// </summary>
     public const float AGENT_EMISSION_IMPULSE_STRENGTH = 20.0f;
 
-    public const float OXYTOXY_DAMAGE = 10.0f;
+    public const float OXYTOXY_DAMAGE = 30.0f;
 
     /// <summary>
     ///   Delay when a toxin hits or expires until it is destroyed. This is used to give some time for the effect to
@@ -165,7 +165,7 @@ public static class Constants
     /// <summary>
     ///   The minimum amount of oxytoxy (or any agent) needed to be able to shoot.
     /// </summary>
-    public const float MINIMUM_AGENT_EMISSION_AMOUNT = 1.0f;
+    public const float MINIMUM_AGENT_EMISSION_AMOUNT = 2.0f;
 
     /// <summary>
     ///   The time (in seconds) it takes a cloud being absorbed to halve its compounds.
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -163,9 +163,9 @@ public static class Constants
     public const float AGENT_EMISSION_COOLDOWN = 2.0f;
 
     /// <summary>
-    ///   The minimum amount of oxytoxy (or any agent) needed to be able to shoot.
+    ///   The maximum amount of oxytoxy (or any agent) fired in one shot.
     /// </summary>
-    public const float MINIMUM_AGENT_EMISSION_AMOUNT = 2.0f;
+    public const float MAXIMUM_AGENT_EMISSION_AMOUNT = 2.0f;
 
     /// <summary>
     ///   The time (in seconds) it takes a cloud being absorbed to halve its compounds.
--- a/src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
+++ b/src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
@@ -69,6 +69,12 @@ public class HeterotrophicFoodSource : FoodSource
         // Pili are much more useful if the microbe can close to melee
         pilusScore *= predatorSpeed;
 
+        // predators are less likely to use toxin against larger prey, unless they are opportunistic
+        if (preySize > predatorSize)
+        {
+            oxytoxyScore *= microbeSpecies.Opportunism / Constants.MAX_SPECIES_OPPORTUNISM;
+        }
+
         // Intentionally don't penalize for osmoregulation cost to encourage larger monsters
         return behaviorScore * (pilusScore + engulfScore + predatorSize + oxytoxyScore);
     }
--- a/src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
+++ b/src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
@@ -69,7 +69,7 @@
         pilusScore *= predatorSpeed;
 
         // predators are less likely to use toxin against larger prey, unless they are opportunistic
-        if (preySize > predatorSize)
+        if (preyHexSize > microbeSpeciesHexSize)
         {
             oxytoxyScore *= microbeSpecies.Opportunism / Constants.MAX_SPECIES_OPPORTUNISM;
         }
--- a/src/microbe_stage/Microbe.cs
+++ b/src/microbe_stage/Microbe.cs
@@ -868,7 +868,7 @@ public class Microbe : RigidBody, ISpawned, IProcessable, IMicrobeAI, ISaveLoade
                 var direction = new Vector3(random.Next(0.0f, 1.0f) * 2 - 1,
                     0, random.Next(0.0f, 1.0f) * 2 - 1);
 
-                SpawnHelpers.SpawnAgent(props, 10.0f, Constants.EMITTED_AGENT_LIFETIME,
+                SpawnHelpers.SpawnAgent(props, 1.0f, Constants.EMITTED_AGENT_LIFETIME,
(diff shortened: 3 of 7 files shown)


src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -68,6 +68,12 @@ public override float FitnessScore(Species species)
         // Pili are much more useful if the microbe can close to melee
         pilusScore *= predatorSpeed;
 
+        // predators are less likely to use toxin against larger prey, unless they are opportunistic
+        if (preyHexSize > microbeSpeciesHexSize)
+        {
+            oxytoxyScore *= microbeSpecies.Opportunism / Constants.MAX_SPECIES_OPPORTUNISM;
+        }
```
Maxonovien, line 75: I'm not sure that size should play a role in oxytoxy hunting (aside from larger prey meaning larger yield, but that's more general). Is there any reason you wanted to limit the use against larger preys?

adQuid: This area of the code is a reflection of the AI; while we use a pattern (it's actually an anti-pattern, but shh) of parallel code, the Auto-evo should reflect what the AI would do, independent of if the AI is making good choices.

As to the question of the AI choosing to not engage larger prey, I made that decision based on the idea that it's still unwise to get close to things that could engulf you. The AI still doesn't know if this "prey" has enough HP and toxin resistance to ignore the attack, or enough flagellum or a lighter membrane to chase down and kill this microbe. I tried to take the safe middle ground and say that only some of the microbes will take this risk, and having the auto-evo encourage only the bold microbes to invest in more toxin.

hhyyrylainen: > it's actually an anti-pattern, but shh

Which part is the anti-pattern? The use of threads? We either use threading or make the game much slower. Which on my computer is fine but a little bit older CPUs are really going to struggle to run the compound clouds at more than 60 FPS.

adQuid: Sorry, I didn't mean parallel as in multi-threading, I meant parallel in terms of data structures. The auto-evo, in many places (which I wrote) contained code designed to loosely mimic what the AI code does, under an implicit assumption that a developer will be smart enough to not change one without changing the other. At some point, I would like to decompose the AI code into context-independent functions that can be run by both the AI and the auto evo, so that both areas of code can ask "what would this species do in this circumstance?" by reusing the same code.

hhyyrylainen: > that a developer will be **smart enough** to not change one without changing the other

That's too optimistic. I've seen PRs that change the logic and don't bother to update a comment that's a few lines above it.

adQuid: I couldn't agree more so, hence my objection to my own design. But the refactor I would like to do will be a PR in itself, about which I'm sure we will have many conversations.

src/microbe_stage/Microbe.cs
```diff
@@ -639,14 +639,16 @@ public void EmitToxin(Compound agentType = null)
 
         float amountAvailable = Compounds.GetCompoundAmount(agentType);
 
-        if (amountAvailable < Constants.MINIMUM_AGENT_EMISSION_AMOUNT)
+        // Emit as much as you have, but don't start the cooldown if that's zero
+        float amountEmitted = Math.Min(amountAvailable / Constants.MAXIMUM_AGENT_EMISSION_AMOUNT, 1.0f);
```
hhyyrylainen, line 643: Why is the divide by the constant here?
Wouldn't it be simpler to just get the min from the available amount and the maximum amount to emit.
Then in the if just return if the amount is less than the min emission amount (see my last comment about that).

src/microbe_stage/Microbe.cs
```diff
@@ -639,14 +639,16 @@ public void EmitToxin(Compound agentType = null)
...
+        if (amountEmitted < MathUtils.EPSILON)
             return;
 
+        Compounds.TakeCompound(agentType, amountEmitted * Constants.MAXIMUM_AGENT_EMISSION_AMOUNT);
```
hhyyrylainen, line 647: That will also simplify this logic?

src/microbe_stage/Microbe.cs
```diff
@@ -662,11 +664,18 @@ public void EmitToxin(Compound agentType = null)
 
         var position = Translation + (direction * ejectionDistance);
 
-        SpawnHelpers.SpawnAgent(props, 10.0f, Constants.EMITTED_AGENT_LIFETIME,
+        SpawnHelpers.SpawnAgent(props, amountEmitted, Constants.EMITTED_AGENT_LIFETIME,
```
hhyyrylainen, line 667: I think it makes *more* sense also here to pass directly the amount and let the agent code handle conversion from compound amount to how much damage it should do.

src/microbe_stage/Microbe.cs
```diff
@@ -861,16 +870,16 @@ public void Kill()
...
             var agentScene = SpawnHelpers.LoadAgentScene();
 
-            while (amount > Constants.MINIMUM_AGENT_EMISSION_AMOUNT)
+            while (amount > Constants.MAXIMUM_AGENT_EMISSION_AMOUNT)
             {
                 var direction = new Vector3(random.Next(0.0f, 1.0f) * 2 - 1,
                     0, random.Next(0.0f, 1.0f) * 2 - 1);
 
-                SpawnHelpers.SpawnAgent(props, 10.0f, Constants.EMITTED_AGENT_LIFETIME,
+                SpawnHelpers.SpawnAgent(props, 1.0f, Constants.EMITTED_AGENT_LIFETIME,
```
hhyyrylainen, line 878: using the constant here would be safer for / when the constant is changed.

src/microbe_stage/Microbe.cs
```diff
@@ -997,31 +1006,14 @@ public void Kill()
         // Some pre-death actions are going to be run now
     }
 
-    public void PlaySoundEffect(string effect)
+    public void PlaySoundEffectOnce(string effect)
```
hhyyrylainen, line 1009: I think this is unnecessary. Our sound file configuration can take care of things being non-looping. Otherwise this type of code is needed for *every* game entity type (and also later stages). So I don't like this change. Also, we are modifying the returned Godot loaded object, which *might* be the same instance if Godot is caching things properly, so the "loop" parameter is fundamentally broken. Instead the change needs to be in the Godot import settings per sound file.

src/microbe_stage/Microbe.cs
hhyyrylainen: If you really want to keep this code, then "amountEmitted" *must* be renamed as it is not actually the amount, but a fraction of the maximum emit amount.

hhyyrylainen (changes requested):

src/microbe_stage/Microbe.cs
adQuid: Looking at it a second time, I agree that it was very strange. It made sense to me at the time.

simulation_parameters/Constants.cs
```diff
@@ -163,9 +163,9 @@ public static class Constants
     public const float AGENT_EMISSION_COOLDOWN = 2.0f;
 
     /// <summary>
-    ///   The minimum amount of oxytoxy (or any agent) needed to be able to shoot.
+    ///   The maximum amount of oxytoxy (or any agent) fired in one shot.
     /// </summary>
-    public const float MINIMUM_AGENT_EMISSION_AMOUNT = 1;
+    public const float MAXIMUM_AGENT_EMISSION_AMOUNT = 2.0f;
```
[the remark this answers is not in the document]
Kemikal1: Its already checking if its higher then EPSILON.

src/microbe_stage/Microbe.cs
Kemikal1: I dont understand why this change was made? If it was to underline the fact that the sound is played once im good with it.

Kemikal1 (approved):
Playtested this for some 30 mins and i like the changes here. Didnt find any problems with this code.

src/microbe_stage/Microbe.cs
adQuid: The hope here was to not have the sound file be responsible for knowing _how_ the sound is played. Currently our .ogg files are hard-coded to either loop or not loop, and the play sound effect method has no control over it. I feel that that's very unintuitive and causes issues if we want the same sound to sometimes loop and sometimes not, but hhyyrylainen makes a very good point with the caching, so I'm still trying to think of a better solution.

hhyyrylainen: Bring it up with the Godot developers. That's how the engine is designed. I was confused as well, but I got over it. This is just a bad design decision in the engine and we just have to go with it (see my comments above).

hhyyrylainen:
Something very obvious just came to me, we can update the organelle tooltip as well as add a help menu entry to explain the new damage mechanics.

adQuid:
> Something very obvious just came to me, we can update the organelle tooltip as well as add a help menu entry to explain the new damage mechanics.

I don't think that that would help players who weren't already paying close attention anyway. I also don't have the capacity to run the babel script in my environment, and have given up trying again, so I would need someone else to do it for me.

PR #2608, opened 2021-09-24
Prevent saving if the player is extinct

### LINKED ISSUE #2593: Saving if the player is extinct should be prevented
just tested and I was able to make a quicksave when the player is dead.

Two things:
- [ ] prevent quick save when extinct
- [ ] prevent save from the pause menu (maybe extinction could send a signal to the pause menu telling it to disable the save option permanently? starting a new game or loading a save should restore that just fine)

hhyyrylainen:

prevent saving if extinct

**Related Issues**

closes #2593

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/locale/en.po
+++ b/locale/en.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2021-09-17 17:10+0300\n"
-"PO-Revision-Date: 2021-09-10 07:24-0400\n"
+"POT-Creation-Date: 2021-09-24 09:45+0300\n"
+"PO-Revision-Date: 2021-09-24 09:46+0300\n"
 "Last-Translator: Henri Hyyryläinen <hhyyrylainen@revolutionarygamesstudio."
 "com>\n"
 "Language-Team: English <https://translate.revolutionarygamesstudio.com/"
@@ -1372,11 +1372,11 @@ msgstr "Art by {0}"
 msgid "CLOSE"
 msgstr "Close"
 
-#: ../src/general/MainMenu.cs:394
+#: ../src/general/MainMenu.cs:393
 msgid "OK"
 msgstr "OK"
 
-#: ../src/general/MainMenu.cs:395
+#: ../src/general/MainMenu.cs:394
 msgid "Cancel"
 msgstr "Cancel"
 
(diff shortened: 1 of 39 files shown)


src/saving/InProgressSave.cs
```diff
@@ -30,8 +30,14 @@ public class InProgressSave : IDisposable
 
     private bool success;
     private string message;
+
+    /// <summary>
+    ///   Failure exception or description. Depends on exceptionIsException
+    /// </summary>
     private string exception;
 
+    private bool exceptionIsException = true;
```
FuJa0815, line 39: That variable sounds very... strange. Maybe rename the `exception` variable?

src/saving/SaveStatusOverlay.cs
```diff
@@ -80,15 +85,19 @@ public void ShowMessage(string message, float visibleTime = 0.7f)
     /// <param name="message">Message to show</param>
     /// <param name="exception">Extra / exception info to show</param>
     /// <param name="returnToMenu">
-    ///   If true closing the dialog returns to menu. If false the dialog is just closed (and game is unpaused)
+    ///    If true closing the dialog returns to menu. If false the dialog is just closed (and game is unpaused)
```
FuJa0815, line 88: Indentation wrong

locale/en.po
```diff
@@ -3070,11 +3070,19 @@ msgstr "Overwrite existing save:"
 msgid "OVERWRITE_EXISTING_SAVE_TITLE"
 msgstr "Overwrite existing save?"
 
-#: ../src/saving/SaveHelper.cs:379
+#: ../src/saving/SaveHelper.cs:382
+msgid "SAVING_NOT_POSSIBLE"
+msgstr "Saving is not currently possible due to:"
```
FuJa0815, line 3075:
```suggestion
msgstr "Saving is currently not possible due to:"
```

FuJa0815 (dismissed):
In-game testing pending

src/saving/InProgressSave.cs
hhyyrylainen: I did try, couldn't really think of anything better than what I ended up with.

locale/en.po
hhyyrylainen: I found this: https://english.stackexchange.com/questions/5464/currently-not-or-not-currently
which says that both ways are correct. Though, having "not" first emphasises the "not", which I don't think is too much of a difference either way, so... I'm not convinced it's worth the minuscule trouble of changing this.

hhyyrylainen: Would you consider that a fine explanation? If you can get people to reply on discord in the #translation channel or maybe on the community forums, this'll be pretty easy to change on https://translate.revolutionarygamesstudio.com/

FuJa0815: Nah, should be fine. I've talked with somebody who is better in english than I am, and he said that both are acceptable. No need to change it.

hhyyrylainen: Okay that's fine. But as I linked that stackexchange answer it says that there is a slight emphasis difference, which we can discuss in the future if anyone cares enough to want to change it.

PR #2613, opened 2021-09-24
Created a system for safely keeping game object references across frames

### LINKED ISSUE #2029: Design a way for the AI and other game objects to keep references to other game objects
Do we need to have destruction callbacks or maybe just weak references (need to check that weak references are saved correctly)? Otherwise the objects might keep references to disposed objects and that's bad, and caused the saving problems issues.

Probably will also be good to have for binding agents (@FuJa0815 )

For reference see: https://github.com/Revolutionary-Games/Thrive/pull/2028

- [x] designed and implemented (I did some stuff towards this here: https://github.com/Revolutionary-Games/Thrive/pull/2332)
- [x] styleguide is updated to require people to use this approach when references are kept across frames, single frame local usage variables do not need to be updated to follow the new style

hhyyrylainen:

Not fully tested, but here's a potential implementation on how we can keep references across frames without having disposed object problems.

**Related Issues**

closes #2029

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/engine/EntityReference.cs
+++ b/src/engine/EntityReference.cs
@@ -1,0 +1,147 @@
+﻿using System.Collections.Generic;
+using Newtonsoft.Json;
+
+/// <summary>
+///   Allows safely keeping references to game entities across multiple frames. Needs to be used instead of raw
+///   references as those can't clear themselves when the entity is disposed
+/// </summary>
+/// <typeparam name="T">The entity type</typeparam>
+public class EntityReference<T>
+    where T : class, IEntity
+{
+    // TODO: should this be somehow set to null when we detect that the alive marker is no longer alive
+    // Currently set to clear on fetch
+    private T currentInstance;
+    private AliveMarker currentAliveMarker;
+
+    public EntityReference(T value)
+    {
+        Value = value;
+    }
+
+    /// <summary>
+    ///   Creates a reference that doesn't refer to anything
+    /// </summary>
+    public EntityReference()
+    {
+    }
+
+    /// <summary>
+    ///   Gets the referenced entity if it is still alive
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     If you need to access one object a bunch of times it's better to get the value from here once per frame
+    ///     and keep a normal reference around.
+    ///   </para>
+    /// </remarks>
+    public T Value
+    {
+        get
+        {
+            if (currentAliveMarker == null)
+                return null;
+
+            if (currentAliveMarker.Alive)
+                return currentInstance;
+
+            // Clear these references to let objects get garbage collected
+            currentInstance = null;
+            currentAliveMarker = null;
+            return null;
+        }
+        set
+        {
+            if (currentInstance == value)
+                return;
+
+            if (value == null)
+            {
+                currentInstance = null;
+                currentAliveMarker = null;
+            }
+            else
+            {
+                var marker = value.AliveMarker;
+
+                if (marker.Alive)
+                {
+                    currentInstance = value;
+                    currentAliveMarker = marker;
+                }
+                else
+                {
+                    currentInstance = null;
+                    currentAliveMarker = null;
+                }
+            }
+        }
+    }
+
+    [JsonIgnore]
+    public bool IsAlive => currentAliveMarker is { Alive: true };
+
+    public static implicit operator T(EntityReference<T> value)
+    {
+        return value.Value;
+    }
+
+    public static bool operator ==(EntityReference<T> lhs, EntityReference<T> rhs)
+    {
+        return Equals(lhs, rhs);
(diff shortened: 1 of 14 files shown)


src/engine/EntityReference.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            return false;
+        }
+
+        return Equals((EntityReference<T>)obj);
+    }
+
+    public bool Equals(EntityReference<T> obj)
+    {
+        return ReferenceEquals(Value, obj.Value);
```
adQuid, line 137: Would we consider references holding the same object, but with different alive values, equal? I would think we wouldn't.

src/saving/serializers/EntityReferenceConverter.cs
```diff
@@ -0,0 +1,74 @@
...
+            writer.WriteNull();
+            return;
+        }
+
+        var objectType = value.GetType();
+
+        var genericTypes = objectType.GenericTypeArguments;
+
+        if (genericTypes.Length != 1)
+            throw new JsonException("Invalid generic types for EntityReference");
```
adQuid, line 20: Could we print out the objectType variable as well? I imagine that could have handy information.

src/saving/serializers/EntityReferenceConverter.cs
```diff
@@ -0,0 +1,74 @@
...
+        if (property == null)
+            throw new JsonException("Value property not found in EntityReference");
+
+        var internalValue = property.GetValue(value);
+
+        if (internalValue == null)
+        {
+            writer.WriteNull();
+            return;
+        }
```
adQuid, line 34: I could see this getting us in trouble. If the code of a game object expects a EntityReference variable not to be null, but allows for that EntityReference to have a null Value (which this design currently allows, even though I don't know why you would do that), then saving and loading that object might break it. Worse, I imagine that that would be a scenario that wouldn't happen often, making the future save bug harder to track down.

adQuid (commented):
Well this was a thinky PR! I have more in the way of questions than real change requests, but I guess that's fair for a draft.

src/engine/EntityReference.cs
hhyyrylainen: That would be a bug in the entity class. Entities are supposed to return the same alive instance each time the `AliveMarker` property is read. For speed reasons I'd prefer to not add a check about that here.

src/saving/serializers/EntityReferenceConverter.cs
hhyyrylainen: Maybe? But this should *only* happen when someone is working on new code and they mess this up. Meaning that it's probably pretty easy to tell what's wrong. If not, they can turn off save error catching and run their debugger. This is pretty similar to the other JSON errors that aren't about input, but straight up just mistakes in the code.

src/saving/serializers/EntityReferenceConverter.cs
hhyyrylainen: I'm having a little bit of trouble following your line of reasoning here.

I wrote the serializer this way because I wanted it to be *transparent* to the save. So the saves just plain have the *contents* of the EntityReference saved, and no details about the entity reference at all. This way save compatibility is kept and I think it's nicer this way as the internal helper class doesn't need to be output in a save.

I'm thinking along the lines that a EntityReference should *never* be null, but the value can turn null whenever, so client code using the class needs to be designed with these constraints in mind.

adQuid: I was worried about the case of an EntityReference with a value of null. It sounds like that was intended here, so is the idea that the serializer always assumes Entities are wrapped in an EntityReference, so it will read a null (which I presume doesn't write an empty string) then wrap around that? That's not what I took away from reading this, but that strategy should be safe...

hhyyrylainen: > I was worried about the case of an EntityReference with a value of null.

That's very much the default state. And also the state the EntityReference goes in if the thing it refers to is destroyed. So I don't think there's any other way to design it really.

> serializer always assumes Entities are wrapped in an EntityReference

I mean it doesn't really. It just reads in a null (in which case it default constructs an EntityReference) or it reads in an object of T and constructs a EntityReference<T> out of that. So it always returns a valid EntityReference instance, but it may have a null as its Value.

And when writing it works so that it writes a null if the EntityReference is null or the value is null. Otherwise it just serializes the held Value.

As I said I designed the serialization of EntityReference to be entirely transparent for the save format.

hhyyrylainen:
I did a bunch of playtesting myself on this (though it's hard to confirm if this fixes the save errors) and didn't at least find this making the game more broken than before. So I'm going to merge this very soon.

PR #2617, opened 2021-09-30
Behaviour tab

### LINKED ISSUE #1501: Create sliders in the behaviour tab to control the AI variables for the player species
they probably don't need to cost MP initially as their impact on gameplay will be pretty low.

FuJa0815:

Implements the behaviour tab and changes all occurances of behaviour to british english to conform with the [Style guide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md#code-style-rules)

**Related Issues**

closes #1501

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/locale/en.po
+++ b/locale/en.po
@@ -2755,7 +2755,7 @@ msgid "MEMBRANE"
 msgstr "Membrane"
 
 #: ../src/microbe_stage/editor/MicrobeEditor.tscn:1182
-msgid "BEHAVIOR"
+msgid "BEHAVIOUR"
 msgstr "Behaviour"
 
 #: ../src/microbe_stage/editor/MicrobeEditor.tscn:1234
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -17,13 +17,13 @@ https://community.revolutionarygamesstudio.com/
 A clear and concise description of what the bug is.
 
 **To Reproduce**
-Steps to reproduce the behavior:
+Steps to reproduce the behaviour:
 1. Go to '...'
 2. Click on new game / something else '....'
 3. Play and do '....'
 4. See error
 
-**Expected behavior**
+**Expected behaviour**
 A clear and concise description of what you expected to happen, instead of what actually
 happened.
 
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -2,7 +2,7 @@
 
 ## Our Standards
 
-Examples of behavior that contributes to a positive environment for our
+Examples of behaviour that contributes to a positive environment for our
 community include:
 
 * Demonstrating empathy and kindness toward other people
@@ -13,7 +13,7 @@ community include:
 * Focusing on what is best not just for us as individuals, but for the
   overall community
 
-Examples of unacceptable behavior include:
+Examples of unacceptable behaviour include:
 
 * The use of sexualized language or imagery, and sexual attention or
   advances of any kind
@@ -25,7 +25,7 @@ Examples of unacceptable behavior include:
 
 ## Enforcement
 
-Instances of abusive, harassing, or otherwise unacceptable behavior may be
+Instances of abusive, harassing, or otherwise unacceptable behaviour may be
 reported to the community leaders responsible for enforcement at
 revolutionarygamesstudio@gmail.com.
 All complaints will be reviewed and investigated promptly and fairly.
--- a/locale/bg.po
+++ b/locale/bg.po
@@ -2766,7 +2766,7 @@ msgid "MEMBRANE"
 msgstr "Мембрана"
 
 #: ../src/microbe_stage/editor/MicrobeEditor.tscn:1182
-msgid "BEHAVIOR"
+msgid "BEHAVIOUR"
 msgstr "Поведение"
 
 #: ../src/microbe_stage/editor/MicrobeEditor.tscn:1234
--- a/locale/ca.po
+++ b/locale/ca.po
@@ -2793,7 +2793,7 @@ msgid "MEMBRANE"
 msgstr "Membrana"
 
 #: ../src/microbe_stage/editor/MicrobeEditor.tscn:1182
-msgid "BEHAVIOR"
+msgid "BEHAVIOUR"
 msgstr "Comportament"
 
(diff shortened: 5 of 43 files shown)


src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -248,6 +248,21 @@ public float Rigidity
...
+
+    [JsonProperty]
+    public float Fear { get; private set; }
+
+    [JsonProperty]
+    public float Activity { get; private set; }
+
+    [JsonProperty]
+    public float Focus { get; private set; }
+
```
adQuid, line 265: Since you've already added an enum for this, you could cut down on lines by converting these to a dictionary using the enum values as keys. This would also ensure that any code using the personality variables will always include all of them, even if we end up adding or removing one in the future.

locale/en.po
```diff
@@ -2038,32 +2038,67 @@ msgstr ""
...
+#: ../src/gui_common/tooltip/ToolTipManager.tscn:3222
 msgid "HEALTH"
 msgstr "Health"
 
-#: ../src/gui_common/tooltip/ToolTipManager.tscn:228
+#: ../src/gui_common/tooltip/ToolTipManager.tscn:220
+msgid "AGGRESSION_EXPLANATION"
+msgstr ""
+"Aggressive microbes will chase down prey over greater distances\n"
+"and are more likely to fight predators when attacked."
```
adQuid, line 2064:
```suggestion
"and are more likely to fight predators when attacked.\n"
```

locale/en.po
```diff
@@ -2038,32 +2038,67 @@ msgstr ""
...
+"Peaceful microbes won't engage others over greater distances"
```
adQuid, line 2065:
```suggestion
"Peaceful microbes won't engage others over greater distances\n"
```

adQuid (changes requested):
aggression text is formatted poorly

src/microbe_stage/editor/MicrobeEditor.cs
FuJa0815: Is that really a good idea? Other code also has these properties split out and this makes reading the properties way easier.

hhyyrylainen: It might slightly be easier to handle loading an old save with a single object that holds all of these, that way on editor load the object can be created if it is null (loaded from an old save or a freshly entered editor) or if it is present used as is.

src/saving/ISaveUpgradeStep.cs
```diff
@@ -89,6 +90,82 @@ protected override void CheckAndUpdateProperty(JProperty property)
         }
     }
 
+    internal class UpgradeStep055To056 : BaseRecursiveJSONWalkerStep
```
FuJa0815, line 93: I might need some help with that.
This code does not work, it crashes with an `AggregateException`: `Collection was modified; enumeration operation may not execute` on line `ISaveUpgradeStep.cs:181`.

hhyyrylainen: Look at how I did it for the json being contained in the dictionary keys. Basically you need to make a list of the things you want to update and loop that list, that way you aren't modifying the structure you are looping through. Have a look at `DetectAndUpdateKeysThatAreJSON`

hhyyrylainen: I was going to say that it doesn't make super ton of sense to move the behaviour values out of the species object to a separate composite object, that just seems unnecessary to me. But if you guys want that change, I can accept it.
I'd personally just make a class for holding the behaviour values while in the editor and not update the species to use that.

FuJa0815: I can see the point that a dictionary is easier to change in the future. Also, changing the storage in the species goes along well with the editor storage.

hhyyrylainen: Actually, are they stored with string keys?
If they are that yells a huge design problem to me.
I thought you had made based on adQuid's comment something like:
```
public class BehaviourValues {
    public float Aggression ...etc
}
```

I don't want random properties being stored as behaviour values in the species.

adQuid: The dictionary is based on enum keys, not string keys, which is what I was suggesting: https://github.com/Revolutionary-Games/Thrive/pull/2617/commits/5a78125163e86bf8cc0526b1fea4da0ee6cbd83a#diff-3246008dccf5d29ed6d40ffae67cd0c23d2e98886134ce85d4096611b899be6aR31

When serialized, of course, these are presumably stored as strings, which someone could mess with, but the same applies to variable names.

hhyyrylainen: Okay, that's slightly better, but still. I don't see why we should have dynamically created list of properties.
It's so much safer to just have them as normal properties; I don't ever want to see any weird bug or behaviour if some species object ends up missing one value in the dictionary.

adQuid: That's the advantage of collections; you can have one statement iterating over every key, and not have a chance of missing one. With properties you need to remember to set each one in its own statement.

hhyyrylainen: If you want to have proper design, I'd add an interface IBehaviourProperties or something like that, with an extension method to copy the properties. I really don't think adding the potential class of bugs about dictionary missing keys potentially and all code wanting access needing to go through the dictionary than just using a property, is worth it to make it easier to loop through values.

adQuid: I've already backed out my review. I'm not going to argue with you over poor Fuja's PR.

Kemikal1 (approved):
Well i may not be a code style connoisseur but looks good  and works good. I only have one personal suggestion, to add the antonyms of each behavior type to the end of each slides .Could be done in another PR too tho.

FuJa0815:
> Well i may not be a code style connoisseur but looks good and works good. I only have one personal suggestion, to add the antonyms of each behavior type to the end of each slides .Could be done in another PR too tho.

But what is are good antonyms for focused?

Kemikal1:
Procastrinated lol, but for real i think distracted could work.

adQuid:
"responsive" might make it sound like less of an obviously bad choice

Kemikal1:
that too OR stubborn - flexible

Issue #1789, opened 2020-11-01
Population Discrepancy Issue: 'Patch Map' shows a Population that Does Not Match 'Report' Screen - Increments by 10 Every Reproduction


Aquasheep3000:
MASSIVE EDIT: I have explained the issue in detail and consolidated to what is likely a single issue.

1) In the Patch Map Screen, the population is shown to be '650' whereas in the 'Report' Screen, the population is shown to be '50'. This occurred after the first reproduction in a new biome and after subsequent reproductions (still showing '50'). It was consistent across a couple of biomes. This is as of 0.5.2.0 rc-1. Furthermore, this '650' number appears to be 10 * n where 'n' is the number of total reproductions across all biomes.

The inaccurate population count in the 'Patch Map' screen for the Pangonian Epipelagic biome (another newly-inhabited biome had the same exact behavior; this occurred after winning the game).
![10 15 20 - Thrive 0 5 2 - Inaccurate Population on 1st Reproduction New Biome](https://user-images.githubusercontent.com/44108929/97805653-1e206180-1c1d-11eb-907c-fb64e339cc63.png)

The stuck population count in the 'Report' Screen.
![10 15 20 - Thrive 0 5 2 - Inaccurate Population on 1st Reproduction New Biome 2](https://user-images.githubusercontent.com/44108929/97805806-e960da00-1c1d-11eb-9725-fb73e67252f2.png)

hhyyrylainen:
But isn't this exactly how it should work? When you as the player move to a new patch, your population there is 0, then when you reproduce your population increases in your current patch? And that is correctly shown: population is now something when previously it was 0.

Aquasheep3000:
@hhyyrylainen You are right. On second thought, I have begun to edit the issue, posting shortly.

hhyyrylainen:
Does this happen with other species as well?
I'm guessing that this might be a player species specific issue.
If you can provide a save file or exact steps that make the issue happen each time, that would greatly help whoever picks up working on this issue.

Aquasheep3000:
Here is a save file indicating the issue in the Pangonian Coast Biome. The population says 50, but that shouldn't be accurate after evolving; there should be a higher population with the previous 50. It appears it's stuck at 50 with a previous population of 0 and then the 'Report' Screen population is only incrementing by the rule of 10 * the total number of reproductions in all biomes.
[...]

PR #2623, opened 2021-10-02
Patchwide population system

### LINKED ISSUE #1288: Player losing population probably has a bug
I think that when you lose population by dying it only applies to the current patch and that eventually goes negative, so while you can still probably go extinct, this leads to a silly situation where the player needs to have enough negative population in the current patch to override the population from other patches to go extinct. A better way would be for the immediate population reduction effect to take population from other patches as well if the current patch doesn't have enough population.

### LINKED ISSUE #1390: Player should get kicked out of the current patch with no population
Right now the player can just keep playing (or being stuck) in a patch where they constantly die, even after the population in that patch is negative. So I suggest we make it so that the player gets kicked out of a patch if their population in it goes to 0. I think that initially we could just have the player be moved to the closest (adjacency taken into account) patch their species is in, but in the future this would be a good place to have a small GUI for selecting the next patch to play in.

FuJa0815:

This PR does multiple things:
- changed environemntal effects to only affect the current patch
- made PatchPanel its own tscn file
- implemented a system for blocking paches in the patch map
- if the player goes extinct in a patch a new menu shows up informing the player that he died out in this patch and allowes the player to select a new patch to continue in

**Related Issues**

closes #1288
closes #1390
#1789 - needs investigation

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
I'm fine with the other changes, but

> when the player migrates to a new patch 10% of the microbes in the old patch migrate over

seems a bit much to include here, as that's a pretty major change I think. We need to consider the design we want regarding how the player can command their *species* to migrate.

FuJa0815:
> 
> 
> I'm fine with the other changes, but
> 
> > when the player migrates to a new patch 10% of the microbes in the old patch migrate over
> 
> seems a bit much to include here, as that's a pretty major change I think. We need to consider the design we want regarding how the player can command their _species_ to migrate.

I needed to move some amount, otherwise my patch extinction logic would instantly trigger.

hhyyrylainen:
It should trigger only on death? I'm fine with it being so that when the player moves to a new patch, they *must* survive there or they'll get immediately thrown back out of the patch.

buckly90:
Can you not just give the player an initial population of 10-40 or so? That would prevent them from being kicked out of the patch when they first migrate. I agree with hh that it should only trigger on death though...

A population penalty sounds interesting, and would hypothetically encourage players to be more careful about migration. But with how death scales already I'm not sure it would be necessary unless the player's death only effects the patch they are currently in when they die (As in they don't actually lose any significant population because they died in a new patch and get thrown back.)

hhyyrylainen:
> I'm not sure it would be necessary unless the player's death only effects the patch they are currently in when they die (As in they don't actually lose any significant population because they died in a new patch and get thrown back.)

That's exactly my concern. Because if no population is moved when you move patches, then when you do and get kicked out of the patch, you wouldn't actually get the population penalty. So trying to move to a patch could be freely tried without a population penalty, which I guess is also a valid design choice to not punish players trying to move to a patch if they aren't completely sure they can survive there.

FuJa0815:
Realistically it would make sense that some cells migrate along with the player. But maybe we should move the discussion to the devforum.

FuJa0815:
Save compatibility will be a huuuuge PITA, and player migration still has to be discussed, but other than that, I think this should be ready for first reviews.

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -78,7 +78,7 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
                         }
 
                         if (useGlobal)
-                            currentPopulation = currentSpecies.Population;
+                            currentPopulation = parameters.World.GetGlobalSpeciesPopulation(currentSpecies);
```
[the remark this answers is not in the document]
FuJa0815: I've tried reverse engineering the class but I was unable to get a breakpoint on this line to hit. To me it seems like this if never executes?

shaders/Monochrome.shader
```diff
@@ -0,0 +1,7 @@
+shader_type canvas_item;
+
+void fragment() {
+    COLOR = texture(TEXTURE, UV);
+    float avg = (COLOR.r + COLOR.g + COLOR.b) / 3.0;
```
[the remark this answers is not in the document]
Kemikal1: ( (0.3 * R) + (0.59 * G) + (0.11 * B) ).

src/auto-evo/simulation/PopulationSimulation.cs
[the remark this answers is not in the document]
hhyyrylainen: It will definitely hit once https://github.com/Revolutionary-Games/Thrive/pull/2689 is merged.

FuJa0815:
This is ready for review if someone wants to do the first iteration

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -120,7 +141,10 @@ private void UpdateSelectHighlightRing()
         if (highlightPanel == null)
             return;
 
-        highlightPanel.Visible = Highlighted || Selected;
+        if (Enabled)
+            highlightPanel.Visible = Highlighted || Selected;
+        else
+            highlightPanel.Visible = false;
```
Kasterisk, line 147: Include brackets.

Kasterisk (changes requested):
The starting population now seems to be 1, I think this is a quite significant change to the gameplay as the player now only have one chance to stay alive, this might be unfavorable as new players might die more than once on a first playthrough.

Visual-wise I think for consistency the patch extinction screen shouldn't have a black background as we don't have it for the total extinction screen. Regarding the patch selection, I think we don't have to include the patch info panel as the main purpose for the patch selection is to only move to a certain patch and I think the logic here is that a more comprehensive information can be found in the editor instead, by doing this we also simplifies what's shown to the player and reducing visual clutter. Though other opinions would be nice.

Also kind of related but there's also the concern that once a patch map goes big enough it may no longer fit inside the extinction screen and we have to limit what's displayed in some way.

hhyyrylainen:
I hadn't looked, but if the patch extinction indeed has a black entire screen background, I'd also prefer to not have that.

FuJa0815:
> I hadn't looked, but if the patch extinction indeed has a black entire screen background, I'd also prefer to not have that.

This was, in my eyes, more or less neccecerry because you switch to a new patch in that menu. It would be very unnatural to just "pop" into the new patch, this is why I respawn the player in the new patch and then fade in.

Does anybody have a better idea?

hhyyrylainen:
That's where I'd like to see a fade. Like this:
- Fade to black
- Load the new patch (this causes a lag spike)
- Fade back to showing gameplay

src/microbe_stage/editor/MicrobeEditor.tscn
```diff
@@ -2331,14 +2276,12 @@ custom_styles/separator = SubResource( 9 )
 margin_right = 500.0
 margin_bottom = 85.0
 rect_min_size = Vector2( 500, 0 )
-WindowTitle = "NEGATIVE_ATP_BALANCE"
```
Kasterisk, line 2334: Another property godot breaks.

src/microbe_stage/editor/MicrobeEditor.tscn
```diff
@@ -2331,14 +2276,12 @@ custom_styles/separator = SubResource( 9 )
...
 DialogText = "NEGATIVE_ATP_BALANCE_TEXT"
 
 [node name="IslandError" parent="MicrobeEditorGUI/CellEditor" instance=ExtResource( 3 )]
 margin_right = 600.0
 margin_bottom = 85.0
 rect_min_size = Vector2( 600, 0 )
-WindowTitle = "DISCONNECTED_ORGANELLES"
```
Kasterisk, line 2341: And here.

src/microbe_stage/editor/MicrobeEditor.tscn
```diff
@@ -2932,870 +2875,13 @@ anchor_bottom = 0.0
...
-margin_right = 140.0
-margin_bottom = 40.0
-mouse_filter = 1
-custom_styles/hover = SubResource( 26 )
-custom_styles/normal = SubResource( 27 )
-custom_colors/font_color_pressed = Color( 1, 1, 1, 1 )
-text = "NEXT_CAPITAL"
-__meta__ = {
-"_edit_use_anchors_": false
-}
```
Kasterisk, line 3798: The next button in the patch map tab got removed.

src/microbe_stage/gui/PatchExtinctionBox.cs
```diff
@@ -0,0 +1,65 @@
...
+        patchDetailsPanel.CurrentPatch = Map.CurrentPatch;
+        patchDetailsPanel.Patch = null;
+        patchDetailsPanel.OnMoveToPatchClicked = NewPatchSelected;
+
+        patchMapDrawer.OnSelectedPatchChanged = SelectedPatchChanged;
+
+        foreach (var patch in Map.Patches.Values)
+        {
+            patchMapDrawer.SetPatchEnabledStatus(patch, patch.GetSpeciesPopulation(PlayerSpecies) > 0);
+        }
```
Kasterisk, line 39: Maybe good to move this to its own public method.

Kasterisk (changes requested):

src/microbe_stage/editor/MicrobeEditor.tscn
```diff
@@ -2932,870 +2877,13 @@ anchor_bottom = 0.0
...
```
Kasterisk, line 3798: This is still missing.

Kasterisk (changes requested):

src/microbe_stage/editor/MicrobeEditor.tscn
FuJa0815: sorry, missed that

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -2124,6 +2119,21 @@ toggle_mode = true
 ActionIcon = ExtResource( 82 )
 ActionName = "g_toggle_binding"
 
+[node name="ProcessPanel" parent="MicrobeHUD" instance=ExtResource( 58 )]
+margin_left = 8.0
+margin_top = 72.0
+margin_right = -872.0
+margin_bottom = -48.0
```
Kasterisk, line 2126: Actually this seems to be a duplicate as I don't see any removals in the diff.

Kasterisk (changes requested):
One more scene related stuff. Overall I guess I'm fine with the UI layout for now.

Kasterisk (approved):

84634E1A607A:
I'm pretty sure that Merge has caused severe problems to src/microbe_stage/editor/MicrobeEditor.tscn

src/microbe_stage/editor/MicrobeEditor.tscn
```diff
@@ -1,4 +1,4 @@
-[gd_scene load_steps=129 format=2]
+[gd_scene load_steps=108 format=2]
```
84634E1A607A: Severe damages.

src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -170,87 +170,9 @@ public class MicrobeEditorGUI : Control, ISaveLoadedTracked
     [Export]
     public NodePath MapDrawerPath;
 
-    [Export]
```
84634E1A607A: Here too

84634E1A607A (commented):
Due to these damages, I'm not able to test this PR.

src/microbe_stage/editor/MicrobeEditor.tscn
FuJa0815: umm, I don't think so? That should be correct

src/microbe_stage/gui/ExtinctionBox.tscn
```diff
@@ -26,6 +26,7 @@ tracks/0/keys = {
 }
 
 [node name="ExtinctionBox" type="Popup"]
+modulate = Color( 1, 1, 1, 0 )
```
hhyyrylainen: Is this an unintended change? Seems a bit suspicious to be modifying the extinction screen in this PR.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -22,10 +23,12 @@ public class PatchMapDrawer : Control
     [Export]
     public Color ConnectionColour = new(1.0f, 1.0f, 1.0f, 1.0f);
 
+    [Export]
+    public ShaderMaterial MonochromeMaterial = null!;
```
hhyyrylainen: Can't Control modulation be used to make things monochrome? That's how it's done in the editor, I think.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -89,16 +94,6 @@ public override void _Ready()
         }
     }
 
-    public override void _Process(float delta)
```
hhyyrylainen: Please keep this approach where the data is only rebuilt *once* per any sequence of changes to the data on the next frame, instead of setting one property causing the expensive rebuild to happen. Or do you have a very good reason why this must be changed?

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -242,7 +242,9 @@ public class MicrobeHUD : Control
     private AnimationPlayer patchOverlayAnimator = null!;
     private TextureButton editorButton = null!;
     private CustomDialog? extinctionBox;
+    private PatchExtinctionBox? patchExtinctionBox;
     private CustomDialog? winBox;
+    private PackedScene patchExtinctionBoxScene = null!;
```
hhyyrylainen: Why aren't these two newly added very closely related variables next to each other?

src/auto-evo/ExternalEffect.cs
```diff
@@ -24,4 +25,10 @@ public ExternalEffect(Species species, int constant, float coefficient, string e
 
     [JsonProperty]
     public string EventType { get; set; }
+
+    /// <summary>
+    ///   Patch where this occured. Null if global.
```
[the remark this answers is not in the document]
FuJa0815: There is one global event: When the player dies out in a patch.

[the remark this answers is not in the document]
hhyyrylainen: Doesn't that also make sense to be marked to have happened in the patch the player went extinct?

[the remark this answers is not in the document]
FuJa0815: Oh, just noticed that this is a bad description. This should be the patch that is affected by this event (which is the patch where something occured 99% of the time, but not always)

src/microbe_stage/editor/PatchMapDrawer.cs
FuJa0815: I don't really get what you mean by that. In the editor everything is made grayscale by disabling buttons.

hhyyrylainen: Yes. So I'm asking *why* you use a custom ShaderMaterial here instead of adding the module to gray option. This is for example done on the organelle selection buttons, we don't have separate icons for their selected or unselected state.

FuJa0815: Because modulation only darkens the tiles. I wanted to go for a monochrome effect. I guess this is personal taste, but I like the monochrome one more.

image
image

hhyyrylainen: Hang on but then how do the selected organelle icons swap colours?

FuJa0815: The background gets light by "pressing" the button and gets gray by disabling it.
The Icon and text get gray by modulating down the white text to a grayer color.

hhyyrylainen: Okay, fine, I guess there's no other way to do this than to use a custom shader then.

shaders/Monochrome.shader
```diff
@@ -0,0 +1,9 @@
+shader_type canvas_item;
+
+uniform vec3 rgbMultipliers = vec3(1, 1, 1);
```
hhyyrylainen, line 3: Aren't these values wrong per the earlier discussion on this file?

FuJa0815: What exactly do you mean? Monochrome works by calculating a weighted average of the rgb values and setting all color vectors to that value. This variable defines the weights for the three color vectors. By default all channels should be treated equaliy.

hhyyrylainen: Umm, no? Didn't you see those numbers posted here: https://github.com/Revolutionary-Games/Thrive/pull/2623/files/089fc3743df646ce7a68489651d837d4b8c66749#r739664387 ? Each RGB value should have a different default contribution factor to the final luminosity.

shaders/Monochrome.shader
```diff
@@ -0,0 +1,9 @@
+shader_type canvas_item;
+
+const vec4 rgbMultipliers = vec4(0.30, 0.59, 0.11, 1);
+
+void fragment() {
+    COLOR = texture(TEXTURE, UV) * rgbMultipliers;
+    float avg = (COLOR.r + COLOR.g + COLOR.b) / 3.0;
+    COLOR.rgb = vec3(avg);
```
hhyyrylainen, line 8: Isn't the math here still wrong? Shouldn't the RGB multipliers only add up to 1 at most, so taking the average on top of that seems not right to me?

Should it more be like: `luminosity = sum( texture(TEXTURE, UV) * rgbMultipliers)` (I'm not sure such "sum" function exists) and then `COLOR.rgb = vec3(luminosity)`

src/general/Species.cs
```diff
@@ -126,6 +126,21 @@ public void ApplyImmediatePopulationChange(long constant, float coefficient)
             Population = 0;
     }
 
+    public void ApplyImmediatePopulationChangeInPatch(long constant, float coefficient, Patch patch)
+    {
+        var oldPopulation = patch.GetSpeciesPopulation(this);
```
hhyyrylainen, line 131: This method should throw an exception if used for something else than the player. As that should not be allowed as I think that'll make it more likely for auto-evo to fail with a collection modified error.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -712,6 +716,21 @@ public void ShowExtinctionBox()
...
+            return;
+
+        winExtinctBoxHolder.Show();
+
+        patchExtinctionBox = (PatchExtinctionBox)patchExtinctionBoxScene.Instance();
+        patchExtinctionBox.Map = map;
+        patchExtinctionBox.PlayerSpecies = playerSpecies;
+        patchExtinctionBox.GoToNewPatch = GoToNewPatch;
+
+        winExtinctBoxHolder.AddChild(patchExtinctionBox);
```
hhyyrylainen, line 731: Why isn't the patch extinction box always in the win box? I think it should be there, just invisible when not required like the other screens in that holder.

src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -0,0 +1,2363 @@
+﻿using System;
```
hhyyrylainen, line 1: It seems you've unintentionally added this file back?

src/microbe_stage/gui/PatchExtinctionBox.cs
```diff
@@ -0,0 +1,65 @@
...
+        TransitionManager.Instance.AddScreenFade(ScreenFade.FadeType.FadeOut, animationPlayer.CurrentAnimationLength,
+            false);
+        TransitionManager.Instance.StartTransitions(this, nameof(OnFadedToBlack));
+        patchDetailsPanel.MouseFilter = MouseFilterEnum.Ignore;
+    }
+
+    private void OnFadedToBlack()
+    {
+        if (patchDetailsPanel.Patch == null)
+            throw new NullReferenceException("The patch must not be null at this point");
```
hhyyrylainen, line 51:
```suggestion
            throw new InvalidOperationException("The patch must not be null at this point");
```

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -140,5 +189,14 @@ private void UpdateIcon()
...
         iconRect.Texture = PatchIcon;
+        iconDirty = false;
+    }
+
+    private void UpdateGrayscale()
+    {
+        if (iconRect != null)
+            iconRect.Material = Enabled ? null : MonochromeShader;
+
+        grayscaleDirty = false;
```
hhyyrylainen, line 200: Should the dirty flag be cleared when the material could not be actually applied yet?

src/microbe_stage/MicrobeStage.cs
```diff
@@ -423,6 +429,15 @@ public override void _Process(float delta)
             return;
         }
 
+        if (playerExtinctInCurrentPatch)
+        {
+            guidanceLine.Visible = false;
```
hhyyrylainen, line 434: Should the chemoreceptor lines be hidden here as well?

src/microbe_stage/MicrobeStage.cs
```diff
@@ -539,6 +554,23 @@ public void ToggleGUI()
...
+        if (CurrentGame == null)
+            throw new InvalidOperationException("Going to a new patch without a game setup");
+
+        CurrentGame.GameWorld.Map.CurrentPatch = patch;
+        UpdatePatchSettings();
+        playerExtinctInCurrentPatch = false;
+        SpawnPlayer();
+
+        // Auto save is wanted once possible
+        wantsToSave = true;
```
hhyyrylainen, line 571: I don't think saving is wanted in this case at all... seems like a bad idea to delete recent autosaves where the player might better be able to recover...

src/auto-evo/AutoEvoRun.cs
```diff
@@ -268,25 +270,34 @@ public void AddExternalPopulationEffect(Species species, int constant, float coe
...
         foreach (var entry in combinedExternalEffects)
         {
-            // entry.Value is the amount, Item2 is the reason string
-            builder.Append(new LocalizedString("AUTO-EVO_POPULATION_CHANGED",
-                entry.Key.Item1.FormattedName, entry.Value, entry.Key.Item2));
+            if (entry.Key.Patch == null)
+            {
+                builder.Append(new LocalizedString("AUTO-EVO_POPULATION_CHANGED_EVERYWHERE",
+                    entry.Key.Species.FormattedName, entry.Value, entry.Key.Event));
+            }
```
hhyyrylainen, line 294: I may have totally forgotten, but could you remind me what's the case when a global population change is created?

FuJa0815: The player gets a global malus whenever they die out in a patch and are forced in another one.

hhyyrylainen: Oh yeah that was it. But isn't that like totally unrealistic? I think it should be more like the player should get a population penalty to one patch in this case? For example the patch they chose to move to.

hhyyrylainen: If you don't mind, I think it would be much simpler to always require a population effect to be in *some* patch. It'll simplify the code design and allow making the patch field non-nullable reducing needed checks.

hhyyrylainen: What's "malus"? I think I'd prefer an approach where it is just calculated normally and if the reduction causes negative population in the target patch, just set the population there to 0. Would that work fine or is there a problem with that?

FuJa0815: Oh, didn't notice that malus is only known in German, thought due to its latin origin it is also an English term. Malus is a penalty.
The problem is that currently the population penalty is applied to every patch and then let the player select a patch where the population is above 0. If we only apply the penalty to the target patch this does not work and we need to calculate whether or not the player would go extinct there with the penalty applied. If he would then the patch should not be selectable.

hhyyrylainen: My opinion is that the player should be able to move to any patch they want. And if the penalty causes their population to become 0 there, then let it be so. If the player survives to the editor they'll get a bit of population, if not and they keep on dying they need to next select a new patch where they have population. Once the player dies with no population in a patch and there is no other patches where the player population is > 0, then it's game over.

FuJa0815: Oh, I forgot that population=0 is a valid state! Problem resolved.

src/auto-evo/RunResults.cs
```diff
@@ -704,7 +704,7 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
                     {
                         foreach (var effect in effects)
                         {
-                            if (effect.Species == entry.Species)
+                            if (effect.Species == entry.Species && effect.Patch?.ID == patchPopulation.Key.ID)
```
hhyyrylainen, line 707: I think the ? null suppression is unnecessary here.

src/general/GameWorld.cs
```diff
@@ -300,12 +301,22 @@ public void ResetAutoEvoRun()
...
 
-        autoEvo!.AddExternalPopulationEffect(species, constant, coefficient, description);
+        autoEvo!.AddExternalPopulationEffect(species, constant, coefficient, description, patch);
+    }
+
+    public void AlterSpeciesPopulationInCurrentPatch(Species species, int constant, string description,
+        bool immediate = false, float coefficient = 1)
+    {
+        if (Map.CurrentPatch == null)
+            throw new InvalidOperationException("The player is not in a chunk");
```
hhyyrylainen, line 317:
```suggestion
            throw new InvalidOperationException("The player is not in a patch");
```
wrong word?

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -2248,6 +2249,7 @@ custom_styles/separator = SubResource( 36 )
 anims/FadeInOut = SubResource( 37 )
 
 [node name="PauseMenu" parent="." instance=ExtResource( 11 )]
+visible = true
```
hhyyrylainen, line 2252: Probably an intended change...

src/auto-evo/AutoEvoRun.cs
```diff
@@ -268,25 +270,29 @@ public void AddExternalPopulationEffect(Species species, int constant, float coe
...
-                entry.Constant + (long)(entry.Species.Population * entry.Coefficient) - entry.Species.Population;
+                entry.Constant + (long)(speciesPopulation * entry.Coefficient) - speciesPopulation;
         }
 
         var builder = new LocalizedStringBuilder(300);
 
         foreach (var entry in combinedExternalEffects)
         {
-            // entry.Value is the amount, Item2 is the reason string
+            if (string.IsNullOrEmpty(entry.Key.Event))
```
hhyyrylainen, line 290: What's the case when the event is null or empty? Does this PR affect that or is this an unnecessary check?

src/general/Species.cs
```diff
@@ -102,28 +102,21 @@ public void SetPopulationFromPatches(long population)
         }
     }
 
-    /// <summary>
```
hhyyrylainen, line 105: Why was this documentation removed?

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -22,11 +23,17 @@ public class PatchMapDrawer : Control
...
+    [Export]
+    public ShaderMaterial MonochromeMaterial = null!;
+
     private readonly List<PatchMapNode> nodes = new();
 
     private PatchMap? map;
+
     private bool dirty = true;
 
+    private Dictionary<Patch, bool>? statusesToBeApplied;
```
hhyyrylainen, line 35: Could you rename this to be more descriptive. To me "status" of a patch that is a bool doesn't let me really know anything. Is this like the enabled patches thing?

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -0,0 +1,400 @@
+﻿using System;
```
hhyyrylainen, line 1: I assume this is a split off part from the microbe editor? If so you should edit the new separate patch map tab scene so that it uses this code and make sure there weren't any new additions in there that your version hasn't picked up.

stoptakingusernames:
There should be some kind of indicator for the total population of your species in your current patch while in game. Currently you don't know how many times you can die before having to relocate to a different patch

hhyyrylainen:
I think we can leave that later for GUI designers to decide. It's very critical that this is not slowed down as this needs to make it into 0.5.9 so that we don't need to make multiple versions in a row that break save compatibility.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -712,6 +716,30 @@ public void ShowExtinctionBox()
         extinctionBox.Show();
     }
 
+    public void ShowPatchExtinctionBox()
+    {
+        winExtinctBoxHolder.Show();
+
+        if (patchExtinctionBox == null)
+        {
+            patchExtinctionBox = PatchExtinctionBoxScene.Instance<PatchExtinctionBox>();
```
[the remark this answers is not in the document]
FuJa0815: I've adopted the way the other two boxes do this

src/auto-evo/AutoEvoRun.cs
FuJa0815: This is a bit of a hacky solution to avoid an event from showing up in the AutoEvo result. I didn't want to majorly rewrite this code just for the tiny edge case where after applying the population penalty for dying out in a patch the player has 0 global population which is invalid and might lead to bugs, therefore I summon some members to avoid this state: https://github.com/Revolutionary-Games/Thrive/blob/07932a5ac84e194cf45eaa94cacf98614e540e23/src/microbe_stage/MicrobeStage.cs#L391-L396

hhyyrylainen: Why? The case when the player is extinct is when they have 0 global population. So why is that required?

FuJa0815: This code runs when switching to a patch after the patch extinction screen. It's weird showing the player the game over screen (which would also require a specialized check) after telling him: hey, it's not over yet! You can still swich to another patch!

hhyyrylainen: Wouldn't it be better in that case to just  not show the extinction screen? I think that would be a cleaner design than adding a random population increase. Also this seems to give a pretty large population increase if the player was about to lose the game, in effect giving the player bunch of free lives if they managed to almost lose just right. So I'd say get rid of that, and just add some special flag in the stage that the player gets to respawn once more before extinction.

FuJa0815: I avoided this because I would consider it a bug being alive with 0 population. Also there are quite many places in the code where `Population > 0` is a check.

hhyyrylainen: In a new patch the player also has 0 population. And those checks for no population won't apply until the player goes to the editor at which point they gain the population bonus from that. So I do not see this as a problem at all. I find the randomly applied population bonus much more problematic design. If ever any player would noticed that and reported that as a bug I would say "told you so" at that point that it is no good design at all...

Kasterisk:
Anything I can do to finish this PR?

hhyyrylainen:
As far as I remember `src/microbe_stage/gui/PatchDetailsPanel.tscn` needs to be checked that we didn't make any new changes to that in master, and then all places that (which should be just the patch map tab scene) use that component need to be refactored to now use the extracted scene. I *think* that was the only thing left to do for this (and fixing the translation conflicts but that should be easy like it normally is).

hhyyrylainen:
Sorry, accidentally closed this.

hhyyrylainen:
Looks like Jetbrains caught some nullability related problems. I'll try to review the new changes ASAP to be able to confirm that this would be good to go.

src/auto-evo/AutoEvoRun.cs
hhyyrylainen: Seems like this is still the case...
If no one else can fix this, I think I might give it a shot to get this done in the next few weeks.

src/microbe_stage/gui/PatchExtinctionBox.cs
```diff
@@ -0,0 +1,79 @@
...
+    private PatchDetailsPanel detailsPanel = null!;
+    private AnimationPlayer animationPlayer = null!;
+
+    public PatchMap? Map
+    {
+        get => mapDrawer.Map;
+        set
+        {
+            mapDrawer.Map = value;
+            mapDrawer.SetPatchEnabledStatuses(value!.Patches.Values, p => p.GetSpeciesPopulation(PlayerSpecies) > 0);
```
hhyyrylainen, line 25: This should throw an explicit exception if value is null, if this doesn't work with null values

src/microbe_stage/MicrobeStage.tscn
hhyyrylainen: Still here

Oliveriver:
Most of this seems to work well, but I did find one thing that might be an issue. I went extinct in a patch and moved to an adjacent patch, but when I next entered the editor, there were still members of my species in the patch I'd gone extinct in and I could still move there. Is this intended?

hhyyrylainen:
I assume that patch performed so well in auto-evo that it generated more population than the population penalty from external effects was. That's pretty much working as intended. I'm sure no matter which way around we mangle auto-evo and external effects to to interleave, someone will run into a situation they would consider at least peculiar.

Oliveriver:
Fair enough. Just wanted to check that wasn't broken.

I have realised this and the procedural patch map are going to have pretty severe merge and functionality conflicts. In particular, the screen to select a new patch on patch extinction in this PR will need to have all the navigational changes that end up being added to the precedural patch map PR (so for the moment, just the scrollbars and scroll centering).

src/microbe_stage/MicrobeHUD.cs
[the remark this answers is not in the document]
hhyyrylainen: This is still a problem...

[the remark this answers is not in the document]
hhyyrylainen: Never mind, I see the other win/extinct boxes also do the same thing.

hhyyrylainen (approved):
I did code cleanup and refactoring as best I could. There's still 2 bugs I know of with this, but I'll merge this anyway to get this done.

hhyyrylainen:
Here's the follow up issues:

- https://github.com/Revolutionary-Games/Thrive/issues/3459
- https://github.com/Revolutionary-Games/Thrive/issues/3460 (if no one else does this, I will make time for myself to fix this bug before the RC1 is out)

Issue #2603, opened 2021-09-22
Softlock when messing with pause on a new game


FuJa0815:
**Describe the bug**

When pressing pause after the new game intro the game can softlocked. You cannot move and the Pangonian Vents-text is semi-visible and does not fade. Opening and closing the pause menu fixes this.

**To Reproduce**
1. Create a new game
2. After the cutscene press esc (before the tutorial popup shows up!)
3. Wait for the tutorial to pop up
4. Close the pause menu
5. Press the x on the tutorial popup
6. Get stuck

**Screenshots**
![image](https://user-images.githubusercontent.com/30809803/134393385-b9b77ff6-6cc6-4a62-955e-0ff8cc27efca.png)

hhyyrylainen:
This is probably as the tutorial remembers the game is paused, it returns to the paused state when it is closed (and not to the unpaused state). I think it would make sense to block pausing the game until after the piece of code runs that pops up the tutorial (if enabled).

PR #2637, opened 2021-10-09
Softlock with pause on a new game

takko:

Blocked pausing until tutorial opens, to avoid softlock.

**Related Issues**

closes Revolutionary-Games/Thrive#2603

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/microbe_stage/MicrobeStage.cs
+++ b/src/microbe_stage/MicrobeStage.cs
@@ -39,6 +39,7 @@ public class MicrobeStage : NodeWithInput, ILoadableGameState, IGodotEarlyNodeRe
     private GuidanceLine guidanceLine;
     private Vector3? guidancePosition;
     private PauseMenu pauseMenu;
+    private bool transitionFinished;
 
     /// <summary>
     ///   Used to control how often compound position info is sent to the tutorial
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
Is this related to https://github.com/Revolutionary-Games/Thrive/pull/2573 ?

takko:
> Is this related to #2573 ?
Yes, this is related, as this is the same problem between MicrobeStage and PauseMenu instead of MicrobeEditor and PauseMenu.
And I can use the GameLoading property from #2573.

takko:
#2573 added a check for `TransitionManager.Instance.HasQueuedTransitions` in the PauseMenu.cs. If there would also be a check if the game is currently transitioning, then this (#2603) would not happen.
But I don't know, if there are transitions where pause is allowed.

In general, when pausing the game has so much impact, then the PauseMenu should be more restrictive. (E.g. default block pausing until allowed.)

hhyyrylainen:
@Kasterisk @FuJa0815 thoughts on that point? (https://github.com/Revolutionary-Games/Thrive/pull/2637#issuecomment-939315232)

FuJa0815:
Making the pause menu a whitelist instead of a blacklist might actually not be a bad idea. I can think of many instances where pausing should be blocked but only two where it should be allowed: during the loaded microbe editor and during the loaded microbe stage.

The changes in this PR are very similar to `MicrobeEditor.Ready` and implements this logic in the `MicrobeStage`. To be honest, this was an oversight of my PR and this is a fix to that, therefore this PR can be merged imo. Still, changing the pause blocking system to a more restricted implementation sounds good and should be tackled, but probably in a seperate PR.

PR #2659, opened 2021-10-15
add new bbcode for constant

### LINKED ISSUE #2650: Add new bbcode to show game constants
Regarding https://github.com/Revolutionary-Games/Thrive/issues/2649 I realized that if we want the tooltip to say the max amount of toxin to shoot at once, we need a new bbcode tag that can show the values from Constants (and probably in the future some JSON loaded stuff). Basically should have a big switch statement of allowed values to show and then just find that number and format it into the tooltip. I don't think we should use reflection here as I think we'll want to control which values are allowed to be retrieved from Constants.

takko:

Added new BBCode `[thrive:constant][/thrive]` for custom constant values to be added to that switch/case.

**Related Issues**

closes #2650

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
That hack looks so ugly that I'd actually prefer the normal switch statement...

takko:
me too, so switch back?

takko:
Listed the new tag on the wiki 👍

PR #2693, opened 2021-10-23
Colony mass and speed

### LINKED ISSUE #2355: Colony members should give some kind of movement bonus
I think colony member cells (and maybe their movement organelles as well), should add some movement force to the colony leader, otherwise big colonies might be entirely impossible to move.

https://forum.revolutionarygamesstudio.com/t/beneficial-effects-of-colonies-and-binding/815/19

Kemikal1:

Added colony members mass to the master, also made the flagella of children affect colony speed.
Lowest speed cap is 75% of original microbe, maybe will be  lower when muulticell editor is implemented.

**Related Issues**
Fixes #2355 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/microbe_stage/Microbe.cs
```diff
@@ -420,14 +425,17 @@ public override void _Process(float delta)
                 }
 
                 totalMovement += queuedMovementForce;
-
```
Kasterisk, line 423: Unnecessary blank line removal.

src/microbe_stage/MicrobeColony.cs
```diff
@@ -100,7 +101,7 @@ public void AddToColony(Microbe microbe, Microbe master)
             throw new ArgumentException("Microbe or master null or microbe already is in a colony");
 
         ColonyMembers.Add(microbe);
-
```
Kasterisk, line 103: Unnecessary blank line removal.

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -86,7 +86,28 @@ private Vector3 CalculateMovementForce(Microbe microbe, float elapsed)
         // The movementDirection is the player or AI input
         Vector3 direction = microbe.MovementDirection;
 
-        var forceMagnitude = force.Dot(direction);
+        // TO DO: Once this is merged, with the colony rotation PR
```
Kasterisk, line 89:
```suggestion
        // TODO: Once this is merged, with the colony rotation PR
```

src/microbe_stage/Microbe.cs
```diff
@@ -588,6 +596,17 @@ private Vector3 DoBaseMovementForce(float delta)
 
         float force = Constants.CELL_BASE_THRUST;
 
+        if (Colony != null && Colony.Master == this)
+        {
+            // Multiplies the movement factor as if the colony has the normal microbe speed
+            // Then it substracts movement speed from 100% up to 75%(soft cap),
+            // using a series that converges to 1 , value = (1/2 + 1/4 + 1/8 +.....) = 1 - 1/2^n
+            // when specialized cells become a reality the cap could b lowered to encourage cell specialization
```
Kasterisk, line 604:
```suggestion
            // when specialized cells become a reality the cap could be lowered to encourage cell specialization
```

Kasterisk (changes requested):
Some small comments on the code.

src/microbe_stage/Microbe.cs
Kemikal1: sorry whats the problem there? isnt it the same logic?

hhyyrylainen: Blank lines are friends, not enemies. They should not be erased without a reason.

Kemikal1: i dont understand the logic for this blank line.

hhyyrylainen: There's no reason really to remove it, so it'd be better to just let it be.
It still makes logical sense to me to keep the parts building totalMovement separate from when then the movement force is applied. For example if someone wants to add a new force, they can see the blank line after line 427 and add it there.

FuJa0815:
I really don't like how you mess around with the master's mass in this implementation. Wouldn't a `Mass` property in the Colony class work better?

Kemikal1:
I dont know what to tell you except that, that the player controls the microbe, so its mass should be the one affected here.

Maxonovien:
Without much thought put into, why not make the movement depend on the colony's mass if a microbe's indeed the colony master?

hhyyrylainen:
Good luck trying to convince the physics engine of that...
The only thing that can be done is having a separate variable for storing what the mass for the current microbe is if it isn't in a colony, and then the RigidBody.Mass has to be the colony's mass for the physics engine to work with it.

FuJa0815:
Oh, sorry, forgot Mass is a godot variable and not made by us :D
In that case this should hopefully be fine

FuJa0815 (approved):

Kasterisk (approved):
Works pretty nicely. Moving big iron chunks are easier when in a big colony than individually.

src/microbe_stage/PlacedOrganelle.cs
```diff
@@ -478,9 +478,12 @@ private void FinishAttachToMicrobe()
         }
 
         // Physics
-        // TODO: shouldn't we also add the mass to the colony master?
         ParentMicrobe.Mass += Definition.Mass;
 
+        // Add the mass of the organelles to the master
```
hhyyrylainen, line 483:
```suggestion
        // Add the mass of the organelles to the colony master
```

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -101,7 +101,10 @@ private Vector3 CalculateMovementForce(Microbe microbe, float elapsed)
         // The movementDirection is the player or AI input
         Vector3 direction = microbe.MovementDirection;
 
-        var forceMagnitude = force.Dot(direction);
+        // Real force the flaggela applied to the colony (considering rotation)
```
hhyyrylainen, line 104:
```suggestion
        // Real force the flagella applied to the colony (considering rotation)
```

RacerBG (approved):

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -101,7 +101,10 @@ private Vector3 CalculateMovementForce(Microbe microbe, float elapsed)
         // The movementDirection is the player or AI input
         Vector3 direction = microbe.MovementDirection;
 
-        var forceMagnitude = force.Dot(direction);
+        // Real force the flagella applied to the colony (considering rotation)
+        var realForce = organelle.RotatedPositionInsideColony(force);
```
hhyyrylainen, line 105: This variable doesn't look like it is used anywhere?

hhyyrylainen: Never mind, I just noticed the change done to the next line.

hhyyrylainen, line 105: I think to save on performance it would be nice to have an if guarding this something like:
```
if(is inside colony)
    force = organelle.RotatedPositionInsideColony(force);
```

src/microbe_stage/Microbe.cs
```diff
@@ -426,6 +426,11 @@ public override void _Process(float delta)
 
         HandleOsmoregulation(delta);
 
+        // Colony members have their movement update before organelle update,
+        // so that the movement organelles see the direction
```
hhyyrylainen, line 430: Doesn't this rely on the fact that _Process would be called first on the master microbe? Or would this work correctly in the case where the stale movement direction value is used here? I think this should at least be clarified in a comment.

src/microbe_stage/Microbe.cs
```diff
@@ -455,6 +460,10 @@ public override void _Process(float delta)
                     movementAudio.Play();
             }
         }
+        else
+        {
+            Colony?.Master?.AddMovementForce(queuedMovementForce);
```
hhyyrylainen, line 465: Should it maybe be assumed that if ColonyParent is not null, then Colony can't be null either? So the first ? could be removed here very safely.
Also doesn't this also have a multi frame dependency? As in if the master microbe runs its _Process first this movement force will only apply on the next frame. This dependency over multiple _Process calls should be documented here.

hhyyrylainen (changes requested):
A few code comments more in terms of documenting why the implementation is the way it is than requests to really change things.

Edit: and also need to resolve the conflicts this PR has.

src/microbe_stage/organelle_components/MovementComponent.cs
hhyyrylainen: Actually reading the other parts of the code, I think `RotatedPositionInsideColony` should have an early exit if the microbe is not in a colony to save on some performance when spawning cells with a lot of organelles as well as when a cell with a lot of movement organelles is trying to move (while not being in a colony).

src/microbe_stage/Microbe.cs
Kemikal1: It does indeed rely on that fact. This is what this comment says, tell me what clarifications you need, for me that comment is enough.

Kemikal1: Normally this wouldnt be a problem i think, the player shouldnt notice the fact that the organelles movements for the miliseconds that it takes to adjust.

Kemikal1: Well if the parent gets the process first then its really no problem or delay. I will clarify that then

hhyyrylainen: Note that I *think* that is what happens (as Godot processes signals in scene tree order), you should verify with a debugger to confirm before writing a comment how things are.

Kemikal1: It really is that way no need for debugger.
https://docs.godotengine.org/en/stable/getting_started/step_by_step/scene_tree.html#tree-order

src/microbe_stage/Microbe.cs
Kemikal1: I think here the formatter didnt let me do that but ill check.

hhyyrylainen: If that's the case I think `else if(Colony != null)` with a comment explaining why it is like that would be better than this code.

src/microbe_stage/Microbe.cs
hhyyrylainen: Even if the behaviour is confirmed, remember that there could be a bug in the engine, then explaining it here is necessary to avoid future confusion. See my styleguide entry that if there's enough confusion to warrant debating about something, that's enough to require it to be commented properly in the code.

src/microbe_stage/Microbe.cs
Kemikal1: huh? why? why notleave it compressed like that? standard null propagation

hhyyrylainen: In the If it's already checked that the cell is a part of a colony, so Colony variable should not be null, also (I didn't design the colony system, but I think this is true) the Master in a colony should never be null. A colony can't exist without a master. So it doesn't make sense to do null checking on them here, if the linter truly complains here, I'll accept null checks, but logically they don't make any sense here.

src/microbe_stage/Microbe.cs
Kemikal1: Well anything could be a bug and be commented when thinking like that but ok.

hhyyrylainen: In unclear situations I prefer much more to just test something rather than believing documentation is right. At least after relying on the documentation I do at least a quick test of the overall operation to make sure I understood the documentation correctly / there wasn't a mistake in the documentation.

src/microbe_stage/Microbe.cs
Kemikal1: The linter complains
image

Kemikal1: the inspect_results says nothing more usually then what is printed.
<Issue TypeId="PossibleNullReferenceException" File="src\microbe_stage\Microbe.cs" Offset="14746-14752" Line="466" Message="Possible 'System.NullReferenceException'" />

Kemikal1: image

hhyyrylainen: The offset points to the exact variable that can be null. At least in emacs you can go to the n:th character in a file so you can use those.

Kemikal1:
Ill do the conflicts tomorrow

Kemikal1:
Or not

hhyyrylainen (changes requested):
I did code style and typo fixes I still wanted. I think the basic code and implementation is fine now.
However this breaks save compatibility. If you load a save made with a previous version where the player is in a colony, then unbinding causes negative mass for the player.
```
ERROR: Condition "p_mass <= 0" is true.
   at: set_mass (scene/3d/physics_body.cpp:499)
```

This needs a save upgrade step that adds the colony member mass values to the colony master mass. You should probably coordinate with @Kasterisk as the timeline PR bumps the version to 0.5.7-alpha and adds a save upgrade step. I think modifying that same step or adding a alpha2 version bump and upgrade step need to be done.

Here's a save where the problem occurs if you press the unbind button:
[quick_save_12.zip](https://github.com/Revolutionary-Games/Thrive/files/7869004/quick_save_12.zip)

Kemikal1:
i was thinking of just making an if situation for the case of an older version save.

hhyyrylainen:
Okay, using the save loadable interface and fixing things up in the callback is also valid, though I can't immediately tell how you would recognize an older save being loaded there, as there's no data field difference.

Kemikal1:
basically add all the mass of organelles of the colony and see if its bigger then the current microbe mass

hhyyrylainen:
I guess that could work, though I'm not super happy that we would need to basically forever run that code always when loading a save, instead of just doing it once when upgrading the save version.

export_presets.cfg
```diff
@@ -56,8 +56,8 @@ codesign/digest_algorithm=1
 codesign/description=""
 codesign/custom_options=PoolStringArray(  )
 application/icon="res://assets/misc/icon.ico"
-application/file_version="0.5.7.0"
-application/product_version="0.5.7.0"
+application/file_version="0.5.8.0-alpha"
```
hhyyrylainen, line 59: This is incorrect, you should not put the version suffix in this file at all.

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -140,7 +143,10 @@ private Vector3 CalculateMovementForce(Microbe microbe, float elapsed)
             forceMagnitude / 100.0f;
 
         // Rotate the 'thrust' based on our orientation
-        direction = microbe.Transform.basis.Xform(direction);
+        if (microbe.Colony?.Master == null)
+            direction = microbe.Transform.basis.Xform(direction);
+        else
+            direction = microbe.Colony.Master.Transform.basis.Xform(direction);
```
hhyyrylainen, line 149: Braces need to be used in multiline if statements.

src/microbe_stage/Microbe.cs
```diff
@@ -282,20 +280,24 @@ public override void _Ready()
                 }
             }
 
+            Mass = Constants.MICROBE_BASE_MASS;
+
```
hhyyrylainen, line 284: Wouldn't this be fine to be on line 269 still? I don't understand why we shouldn't initialize mass for non save loaded microbes.

Kemikal1: its strange taht i didnt see any kind of difference with other microbes

hhyyrylainen: I think the base mass is so small related to organelle mass that it is very insignificant... that might explain it.

src/microbe_stage/PlacedOrganelle.cs
```diff
@@ -494,10 +498,13 @@ private void FinishAttachToMicrobe()
         }
 
         // Physics
-        // TODO: shouldn't we also add the mass to the colony master?
         ParentMicrobe!.Mass += Definition.Mass;
 
-        MakeCollisionShapes(ParentMicrobe.Colony?.Master ?? ParentMicrobe);
+        // Add the mass of the organelles to the colony master
+        if (ParentMicrobe.Colony != null && ParentMicrobe != ParentMicrobe.Colony.Master)
+            ParentMicrobe.Colony.Master.Mass += Definition.Mass;
```
hhyyrylainen, line 505: I'm a bit suspicious that this line might cause duplicate mass to be added to the colony leader. As If we load a save there's we add the masses of the individual cells to the leader, and then call FinishAttachToMicrobe for all organelles, which adds the mass again. So I think there is a bug maybe here.
And I think maybe the bug would go entirely away if this wouldn't add any mass to the colony master.

Kemikal1: cant colony members grow themselves organelles in colony?

Kemikal1: at least that was what i was thinking

hhyyrylainen: They cannot. That is entirely prevented. Even in my multicellular prototype there's no organelle division. Only fully formed cells are added once there are enough resources.

Then could you comment out that bit of code and put like a TODO:
`TODO: if organelles can grow while cells are in a colony this will be needed (when enabled something needs to be modified to ensure no duplicate mass is added with loading a save)`

Kemikal1: ill modify it to only wotk when not loading a save and then comment

hhyyrylainen: It seems you commented it out already. Are you going to not do this:

> ill modify it to only wotk when not loading a save and then comment

Also I'd like the warning bit that it will cause a bug if it isn't adjusted in the future, as I'm quite sure that whoever uncomments the code in teh future won't know that we had this conversation.

Kemikal1: i did? i put a !IsLoadedFromSave in the if

hhyyrylainen: Ah, sorry, I didn't see that. I think it's fine then once the comment format is fixed so that the CI is happy.

PR #2696, opened 2021-10-24
Implement a popup for exit confirmation

### LINKED ISSUE #2156: Make sure you want to quit
"Make sure you want to quit" message box.

Kasterisk:
closes #2208 (previous PR)
closes #2156

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


FuJa0815 (dismissed):
I'm sorry for my blunder, I assumed that every commit in #2208 was made by ender-titan. There is only this tiny comment, everything else is different enough or is the only sensible implementation.

src/saving/SaveHelper.cs
```diff
@@ -38,6 +40,18 @@ public enum SaveOrder
...
+    ///   Checks whether the last save is made within a timespan of set duration.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Used for knowing whether to show confirmation dialog on the pause menu when exiting the game.
+    ///     TODO: Implement a method overriding this flag as old in some way.
+    ///   </para>
+    /// </remarks>
+    /// <returns>True if the last save is still recent, false if otherwise.</returns>
+    public static bool SavedRecently => DateTime.Now - lastSave < Constants.RecentSaveTime;
```
hhyyrylainen, line 53: Doesn't this need a null check?

simulation_parameters/Constants.cs
```diff
@@ -600,6 +600,16 @@ public static class Constants
...
 
+    /// <summary>
+    ///   The duration for which a save is considered recently performed.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Not a const because TimeSpan is not a primitive.
+    ///   </para>
+    /// </remarks>
+    public static readonly TimeSpan RecentSaveTime = TimeSpan.FromSeconds(10);
```
hhyyrylainen, line 611: I'd increase this to 15 for now, and we can tweak this later. Because I think some people might deliberate a bit after saving the game before finding the exit button.

src/general/PauseMenu.cs
```diff
@@ -27,11 +27,21 @@ public class PauseMenu : ControlWithInput
     [Export]
     public NodePath LoadSaveListPath;
 
+    [Export]
+    public NodePath ExitConfirmationDialogPath;
```
hhyyrylainen, line 31: I'd rename anything with "ExitConfirmation" to "UnsavedProgressWarning" that way the last few bits I find similar to the old PR would be gone.

src/general/PauseMenu.cs
```diff
@@ -49,6 +59,15 @@ public class PauseMenu : ControlWithInput
     [Signal]
     public delegate void MakeSave(string name);
 
+    /// <summary>
+    ///   Types of exit the player requests. Used for game exit confirmation.
```
hhyyrylainen, line 63:
```suggestion
    ///   Types of exit the player can request. Used to store the action for when the warning popup about this is closed.
```

hhyyrylainen (changes requested):
Functionality seems fine, code is mostly fine and I can't see surviving code from the previous PR, though I added one comment regarding how to make it clearer that this is not copied from there.
Besides those I actually have a small improvement request: it would be nice to have a checkbox in the options menu to not warn about this, it would be useful when testing the game as I close the game a lot. Not absolutely required, I can open a new issue, but it shouldn't take too long to include that in this PR.

src/saving/SaveHelper.cs
Kasterisk: Probably, it works on first save attempt though...
Should `GetValueOrDefault` be used or should ternary be used here?
```csharp
DateTime.Now - lastSave.GetValueOrDefault() < Constants.RecentSaveTime;
```

hhyyrylainen: I would not use GetValueOrDefault here

Kasterisk: Alright.

src/saving/SaveHelper.cs
```diff
@@ -128,6 +142,7 @@ public static void LoadSave(string name)
 
         GD.Print("Starting load of save: ", name);
         new InProgressLoad(name).Start();
+        lastSave = DateTime.Now;
```
hhyyrylainen, line 145: Actually now that I'm looking at this gain, these time resets should be in the InProgressLoad and InProgressSave classes on the final step, because it can take up to like 5 seconds to load a save now (and likely more in the future), so this time is way too early.

src/saving/SaveHelper.cs
```diff
@@ -38,6 +40,18 @@ public enum SaveOrder
         FileSystem,
     }
 
+    /// <summary>
+    ///   Checks whether the last save is made within a timespan of set duration.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Used for knowing whether to show confirmation dialog on the pause menu when exiting the game.
+    ///     TODO: Implement a method overriding this flag as old in some way.
```
hhyyrylainen, line 49: Isn't this TODO not necessary, or am I not understanding it correctly?

FuJa0815:
There are fuzzy translations in en.po

hhyyrylainen (changes requested):
Functionality is good, but I have slight concern regarding the timing.

src/saving/SaveHelper.cs
Kasterisk: In a scenario where you 1. save the game, 2. return to menu, and 3. start a new game, it may be possible to return to the main menu without the warning to popup (although I haven't got this to yet but still a possibility). So we may want to override the saved recently flag to manually make the popup to appear. This isn't too significant at the moment so it's why I added the TODO here.

hhyyrylainen: Ah okay, could you clarify this here?
And this would be pretty easy to make by adding code in MainMenu._Ready to set the last saved time back to null (through a method to not allow any code to set the property).

Kasterisk: Hang on, that should actually work, I somehow didn't think of setting lastSave to null into account...

Kasterisk:
Just crossed my mind, should we show the warning in freebuild by default anyways?

hhyyrylainen:
I looked at PauseMenu and it would be pretty fiddly to have the warning suppressed there. Anyway people might spend 30 minutes in it making their creature so I think having the save warning there also makes sense.

Kasterisk:
Alright, this should be good to go then.

hhyyrylainen (approved):

PR #2702, opened 2021-10-28
Make the mouse hover panel categorize compound densities

### LINKED ISSUE #1525: Compound panel indicator goes crazy
If a cell(the player or any other organism) is over some compound clouds, and you hover your cursor over it, the panel showing what compounds are at the cursor starts adding and removing compounds very quickly.

### LINKED ISSUE #2202: Make the mouse hover panel categorize compound densities
I've been thinking that we could maybe change it to be a scale like "very little, little, some, a fair amount, quite a bit, an abundance".
I didn't make this an issue back when this was last discussed, but now people have asked for this again.

takko:

Added categories for the hovered compound amount:
```csharp
COMPOUND_DENSITY_CATEGORY_VERY_LITTLE = 0.5f;
COMPOUND_DENSITY_CATEGORY_LITTLE = 5f;
COMPOUND_DENSITY_CATEGORY_SOME = 20f;
COMPOUND_DENSITY_CATEGORY_FAIR_AMOUNT = 50f;
COMPOUND_DENSITY_CATEGORY_QUITE_A_BIT = 200f;
COMPOUND_DENSITY_CATEGORY_AN_ABUNDANCE = 500f;
```

image

Reduced text flickering with a threshold and delayed removing of label for 1 second.

Changed the code for adding the lines to the HoverInfoPanel, so that the Controls are only Hidden/Visible, and not always recreated.

**Related Issues**

closes #2202
closes #1525

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


Kasterisk:
Really nice work on this. Though one thing I thought of that could be a nice QoL feature is maybe adding distinct colours to the categories? This would be a very useful visual cue for players to easily distinguish densities at a glance, so lowest density could have have red color and gradually move up to green at the highest.

takko:
added colors for the categories:
```csharp
            >= Constants.COMPOUND_DENSITY_CATEGORY_AN_ABUNDANCE => Color.Color8(75, 187, 16),
            >= Constants.COMPOUND_DENSITY_CATEGORY_QUITE_A_BIT => Color.Color8(117, 186, 15),
            >= Constants.COMPOUND_DENSITY_CATEGORY_FAIR_AMOUNT => Color.Color8(174, 186, 15),
            >= Constants.COMPOUND_DENSITY_CATEGORY_SOME => Color.Color8(186, 173, 15),
            >= Constants.COMPOUND_DENSITY_CATEGORY_LITTLE => Color.Color8(186, 137, 15),
            >= Constants.COMPOUND_DENSITY_CATEGORY_VERY_LITTLE => Color.Color8(186, 78, 15),
```
image
image

hhyyrylainen:
One style thing: elsewhere in the code we've used floats to define colours, not integer values.

Kasterisk (changes requested):
There seems to be an extra height/margin at the bottom of the panel now.
| Before | After |
| ------------- | ------------- |
|  |  |

icedjuro (approved):
Flickering is gone

takko:
> That looks pretty awesome! However I noticed there seems to be an extra height/margin at the bottom of the panel now.
> 
> Before	After
> 	

fixed the height

Kasterisk:
This might just be me but on brighter patches, I noticed that the colors are a bit too undersaturated and may be hard to see in this case:



I guess we could add a little bit brightness to the colors but not too much that it becomes distracting.

hhyyrylainen:
Maybe the colours could be selected to be like blue to green (in addition to tweaking their visibility)? I think this would be a better pair of colours for colourblind accessibility. So blue would be a mellow colour indicating lackluster compounds, and green would be good amount of compounds.

Kasterisk:
I feel like that would be unintuitive, to me at least green and blue are on equal level and indicates the same thing. As for colorblind accessibility we do have the colorblind filter and I think it could help with the issue.

hhyyrylainen:
While they do probably help, it probably helps even more if we pick colours that are distinct to begin with.

Kasterisk:
Red and green **are** distinct already but I'm not absolutely against trying out the mellow blue and green color, I'd be happy to choose, just need to see if the end result communicates better or the same while keeping a balanced color saturation and visibility.

hhyyrylainen:
Sorry, I wasn't clear that I meant distinct even when considering common colour-blindness variants (red-green colour-blindness specifically).

takko:
image
image

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -1076,4 +1106,42 @@ private void OnAbilitiesHotBarDisplayChanged(bool displayed)
     {
         hotBar.Visible = displayed;
     }
+
+    private class HoveredCompoundControl : HBoxContainer
+    {
+        private readonly Label compoundName = new Label();
+        private readonly Label compoundValue = new Label();
```
hhyyrylainen, line 1113: These should be created in _Ready

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -1076,4 +1106,42 @@ private void OnAbilitiesHotBarDisplayChanged(bool displayed)
...
+        public HoveredCompoundControl(Compound compound)
+        {
+            MouseFilter = MouseFilterEnum.Ignore;
+            Compound = compound;
+            TextureRect compoundIcon = GUICommon.Instance.CreateCompoundIcon(compound.InternalName, 20, 20);
+            compoundName.SizeFlagsHorizontal = (int)SizeFlags.ExpandFill;
+            compoundName.Text = Compound.Name;
+            AddChild(compoundIcon);
+            AddChild(compoundName);
+            AddChild(compoundValue);
```
hhyyrylainen, line 1124: For consistency with other code, these also should be put in _Ready (as well as the MouseFilter line)

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -6,13 +6,16 @@
 /// </summary>
 public class PlayerHoverInfo : Node
 {
+    private readonly Dictionary<Compound, float> currentHoveredCompounds = new Dictionary<Compound, float>();
```
hhyyrylainen, line 9: Could you add a comment on this why this exists and how this differs from HoveredCompounds?

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -6,13 +6,16 @@
...
+    private readonly Dictionary<Compound, float> compoundDelayTimer = new Dictionary<Compound, float>();
     private MicrobeCamera camera;
     private CompoundCloudSystem cloudSystem;
+    private Vector3 lastCursorWorldPos = Vector3.Inf;
```
hhyyrylainen, line 13:
```suggestion

    private Vector3? lastCursorWorldPos;
```

can use a nullable type here to indicate the invalid value

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -29,7 +32,42 @@ public void Init(MicrobeCamera camera, CompoundCloudSystem cloudSystem)
 
     public override void _Process(float delta)
     {
-        HoveredCompounds = cloudSystem.GetAllAvailableAt(camera.CursorWorldPos);
```
hhyyrylainen, line 32: Might be good to check against negative delta here, but I don't see anything in here that would really break if the delta is occasionally negative in here, so up to you if you want to add that or not.

```
        // https://github.com/Revolutionary-Games/Thrive/issues/1976
        if (delta <= 0)
            return;
```

hhyyrylainen (changes requested):
Functionality seems pretty good, though I think there needs to be a tiny bit more of reserved space as hydrogen sulfide can now stretch the box a little bit:



besides that I added just a few tiny code comments.

src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -412,8 +412,7 @@ public void GetCompoundsAt(int x, int y, Dictionary<Compound, float> result)
                 break;
 
             float amount = HackyAddress(Density[x, y], i);
-            if (amount > 0)
-                result[Compounds[i]] = amount;
+            result[Compounds[i]] = amount;
```
hhyyrylainen, line 415: Why was this change made?

simulation_parameters/Constants.cs
```diff
@@ -600,6 +600,46 @@ public static class Constants
     public const int KIBIBYTE = 1024;
     public const int MEBIBYTE = 1024 * KIBIBYTE;
 
+    /// <summary>
+    ///  Delay for the compound row to hide when standing still and compound amount is 0.
+    /// </summary>
+    public const float COMPOUND_HOVER_INFO_REMOVE_DELAY = 0.5f;
+
+    /// <summary>
+    ///  Compound changes below that value are ignored while mouse world position doesn't change.
```
hhyyrylainen, line 609:
```suggestion
    ///  Compound changes below this value are ignored while mouse world position doesn't change.
```

simulation_parameters/Constants.cs
```diff
@@ -600,6 +600,46 @@ public static class Constants
     public const int KIBIBYTE = 1024;
     public const int MEBIBYTE = 1024 * KIBIBYTE;
 
+    /// <summary>
+    ///  Delay for the compound row to hide when standing still and compound amount is 0.
```
hhyyrylainen, line 604: All of these XML comments are missing 2 spaces of needed indentation.

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -6,13 +6,21 @@
...
+    /// <summary>
+    ///   Used to query the real hovered compound values.
+    ///   This is a member variable to reduce GC pressure.
+    /// </summary>
+    private readonly Dictionary<Compound, float> currentHoveredCompounds = new Dictionary<Compound, float>();
+
+    private readonly Dictionary<Compound, float> compoundDelayTimer = new Dictionary<Compound, float>();
     private MicrobeCamera camera;
     private CompoundCloudSystem cloudSystem;
+    private Vector3? lastCursorWorldPos;
```
hhyyrylainen, line 18:
```suggestion

    private Vector3? lastCursorWorldPos;
```

src/microbe_stage/PlayerHoverInfo.cs
```diff
@@ -29,7 +37,46 @@ public void Init(MicrobeCamera camera, CompoundCloudSystem cloudSystem)
...
+        }
+
+        foreach (var compound in currentHoveredCompounds)
+        {
+            HoveredCompounds.TryGetValue(compound.Key, out float oldAmount);
+
+            // Delay removing of label to reduce flickering.
+            if (compound.Value == 0f && oldAmount > 0f)
+            {
+                compoundDelayTimer.TryGetValue(compound.Key, out float delayDelta);
```
hhyyrylainen, line 59: If the TryGetValue fails, then this assumes that delayDelta was initialized to 0, did I get that right?

hhyyrylainen (changes requested):
Playtesting didn't turn up any further issues, just some code related comments.

src/microbe_stage/CompoundCloudPlane.cs
takko: I changed the code, so that you can reuse the same instance of the Dictionary to not always create a new Dictionary on every frame.
So this is needed to zero the value which maybe already in the Dictionary from the previous query.

hhyyrylainen: Why not clear the dictionary then (in the hover info)? This changes the functionality here from "what clouds there are here" to "what cloud types are supported", and I think that kind of assumption change needs more careful investigation regarding the overall code structure that uses the clouds.

src/microbe_stage/PlayerHoverInfo.cs
takko: Yes, TryGetValue assigns default value on fail which is 0.0f.

src/microbe_stage/CompoundCloudPlane.cs
takko: reverted and cleared dictionary.

hhyyrylainen (approved):
I think all of my concerns were addressed. I didn't re-playtest this, maybe some of the other pending reviewers will be able to do a final review?

FuJa0815 (approved):
Code looks good

Kasterisk (approved):

PR #2707, opened 2021-10-28
Background doesn't work correctly in really wide aspect ratio

### LINKED ISSUE #1771: Background doesn't work correctly in really wide aspect ratio
On ultrawide monitors you can see bars on the left and the right when zoomed out far enough
Tested on a 2560 x 1080 monitor

grafik

takko:

Doubled the width for the backgroundplane and adjusted the shader as recommended in the issue.

**Related Issues**

closes #1771

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


takko:
Using the vertex shader solved it.

Maybe found an issue:
I noticed those speed values in the shader, which are doubled for y axis. I draw a round circle on the background and tried to reach 45°. Microbe always went for 22°, so the background moves double so fast on the x axis.
Not sure if this is the wanted behavior or a bug.

hhyyrylainen:
Did you compare the background moving on the master branch and this branch? The speeds should match in all directions (or at least be so close that you can't tell the difference)

takko:
Yes, checked if I can reach the same spot on master in the same time.

hhyyrylainen:
But did you see on master the up/down speed being different in the background? In my experience that isn't the case, so I'm a bit confused why you mentioned that after you said you tweaked things to be correct.

takko:
I tweaked it so it behaves like on master.
I found on master the background speed on x is double the y.
The microbe speed is correct on both branches.
Im not sure if this is a bug or wanted for a depth like feeling.

hhyyrylainen:
My intention was that the speed was the same on both axes. If you could make the background in this PR behave like that, then I'd say that this is even better than master branch version.

takko:
I can upload the test graphic I used later, so you can test what I mean.

Kasterisk (changes requested):

takko:
I took the faster speed (the speed of the x axis).

The circle in the image is directly on the Thrive_vent0.png

Master:
image

Branch:
image

Kasterisk (approved):
Can't really comment on the size fix since I don't have ultrawide monitors, but the background speed fix and other changes seems good so this is good to me.

hhyyrylainen (approved):
Seems to work fine, I think this does fix a problem that was on master so the background speed looks a bit different know, but if someone dislikes it we can open a separate issue to tweak things.

PR #2717, opened 2021-10-30
Pressing space during microbe intro causes it to restart

### LINKED ISSUE #2356: Pressing space or enter during the microbe intro causes it to restart
Looks like the GUI button or whatever stays selected during the cutscene, so when you press space the button is pressed again and the cutscene restarts (for me it restarts, another person reported that the game gets stuck instead).

takko:

This PR does some stuff...

**Related Issues**

closes #2356

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/general/MainMenu.cs
+++ b/src/general/MainMenu.cs
@@ -263,6 +263,9 @@ public class MainMenu : NodeWithInput
         // Ignore mouse event on the button to prevent it being clicked twice
         newGameButton.MouseFilter = Control.MouseFilterEnum.Ignore;
 
+        // Release keyboard focus on the button to prevent it being executed again.
+        newGameButton.ReleaseFocus();
+
         // Stop music for the video (stop is used instead of pause to stop the menu music playing a bit after the video
         // before the stage music starts)
         Jukebox.Instance.Stop();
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
Instead of releasing the focus, I think a boolean flag that the button has been pressed should be used instead (the same flag can be used for all buttons that can cause the problem). That way even if the user manages to move the focus back to the button, for example with keyboard navigation or maybe controller navigation of the menus works, they still can't mess it up.

Kasterisk:
Yeah I think it would be a more robust fix. To be more specific I'm assuming you mean the [focus mode](https://docs.godotengine.org/en/stable/classes/class_control.html#enum-control-focusmode) enum.

hhyyrylainen:
I don't actually mean that. A fix like that, I'd actually prefer if the button become disabled so that there is no way to trigger the press action for it again.
I was actually talking about just adding a bool in the MainMenu class to keep track if the transition to game has been started, and if so ignore any further presses to the button(s).

Kasterisk:
Ah yeah, that's actually better. Though it reminds me, I once entertained a similar idea which is bit more general in terms of implementation and that is to have TransitionManager to block any inputs (except ESC) once a transition has started but that may not actually be feasible.

hhyyrylainen (approved):
Seems to work perfectly.

PR #2718, opened 2021-10-31
Optimize data points interpolation in line charts

### LINKED ISSUE #2715: Resolve performance degradation from LineChart
When messing in freebuild I noticed that I was only getting 150 FPS in the editor, when I usually get 400 (the limit). After stopping in debugger I noticed the main thread was busy in LineChart.cs _Process, and returning early from there (using save made, I also went back to the editor to confirm that nothing I did in the previous session caused it) greatly improved performance.

Before:

This is the code change I tested:
Kuvakaappaus - 2021-10-29 17-05-44

And the FPS recovered after rebuild and restart (I also confirmed that undoing that change and loading the same save, gives 150 FPS again):

So I think that LineChart needs to smartly only update points while animating, and rest of the time skip updating them to not sap performance.

Kasterisk:

Made use of Tween nodes for animating the points and got rid of the horribly optimized code in `LineChart._Process`.

**Related Issues**

Fixes #2715 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/gui_common/charts/line/LineChart.cs
+++ b/src/gui_common/charts/line/LineChart.cs
@@ -231,39 +231,6 @@ public class LineChart : VBoxContainer
         UpdateAxesName();
     }
 
-    public override void _Process(float delta)
-    {
-        // https://github.com/Revolutionary-Games/Thrive/issues/1976
-        if (delta <= 0)
-            return;
-
-        // Apply coordinates (with interpolation for smooth animations)
-        foreach (var data in dataSets)
-        {
-            foreach (var point in data.Value.DataPoints)
-            {
-                if (!IsInstanceValid(point))
-                    continue;
-
-                var coordinate = Vector2.Zero;
-
-                if (IsMinMaxValid())
-                {
-                    coordinate = ConvertToCoordinate(point.Value);
-                }
-                else
-                {
-                    // Hide the marker as its positioning won't be pretty at zero coordinate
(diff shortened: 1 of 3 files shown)


Kasterisk:
Tested with 10 data series and each 10 data points in 3 charts. Got an increase of 700% in FPS and the UI is now very responsive whereas previously I could barely interact with it. I'd say this is a definite improvement :)

src/gui_common/charts/line/LineChart.cs
```diff
@@ -1019,9 +1013,19 @@ public DataLine(LineChartData data, bool isDefault)
...
 
+            tween = new Tween();
+            AddChild(tween);
+
             // Antialiasing is turned off as it's a bit unreliable currently
         }
 
+        public void InterpolatePointPosition(int i, Vector2 initialPos, Vector2 targetPos)
+        {
+            tween.InterpolateMethod(this, "ChangePointPos", new Vector3(i, initialPos.x, initialPos.y),
```
hhyyrylainen, line 1024: This should use `nameof(ChangePointPos)`

hhyyrylainen (changes requested):
I didn't test excessively but I didn't see any performance drop anymore. Noticed just one tiny code related thing and then this is good to go.

hhyyrylainen (approved):

PR #2720, opened 2021-10-31
Support for store versions of thrive

hhyyrylainen:
script and other changes needed to make releases of thrive for game stores

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/steam/SteamHandler.cs
```diff
@@ -0,0 +1,127 @@
...
+
+    public override void _Ready()
+    {
+        OnSteamInit();
+    }
+
+    public override void _Process(float delta)
+    {
+        steamClient?.Process(delta);
+        PauseMode = PauseModeEnum.Process;
```
FuJa0815, line 45: Why do you set the `PauseMode` for each frame?

hhyyrylainen: That's a mistake, I meant to put this in _Ready

hhyyrylainen: Fixed: https://github.com/Revolutionary-Games/Thrive/pull/2720/commits/9eae0ccf12d067f3ffdffe7f42eb47ed54aa9f8a

doc/steam_license_readme.txt
```diff
@@ -0,0 +1,20 @@
+﻿Thrive Game
+Copyright (C) 2013-2021  Revolutionary Games
+
+This version of Thrive is distributed through Steam by Revolutionary Games Studio
```
FuJa0815, line 4: Is this a premade text? If not I would rewrite this a bit because this sounds a bit like Steam was made by us to me

make_release.rb
```diff
@@ -54,19 +55,39 @@
...
-INFORMATIONAL_VERSION = /AssemblyInformationalVersion\("([^"]*)"\)/.freeze
-
 SET_EXECUTE_FOR_MAC = false
 
+SPECIAL_BUILDS = { steam:
+                     {
+                       prepare_compile: lambda {
+                         warning 'This is the Steam build. This can only be distributed '\
+                                 'by Revolutionary Games Studio (under a special license) '\
+                                 'due to Steam being incompatible with GPL license!'
```
FuJa0815, line 65:
```suggestion
                                 'due to Steam being incompatible with the GPL license!'
```

make_release.rb
```diff
@@ -372,16 +404,20 @@ def zip_package(target, target_name, target_folder, target_file)
 
 def package(target, target_name, target_folder, target_file)
   if target_mac? target
-    puts 'Including licenses in mac .zip'
+    if @options[:special]
+      SPECIAL_BUILDS[@options[:special]][:mac_zip_postprocess].call target_folder, target_file
+    else
+      puts 'Including licenses in mac .zip'
```
FuJa0815, line 410: I'd write `macOS` just to clarify

src/general/MainMenu.tscn
```diff
@@ -79,6 +81,29 @@ __meta__ = {
...
+margin_right = -5.0
+margin_bottom = 26.0
+grow_horizontal = 2
+grow_vertical = 2
+mouse_filter = 1
+size_flags_horizontal = 3
+size_flags_vertical = 6
+theme = ExtResource( 1 )
+custom_fonts/font = ExtResource( 13 )
+text = "Store user: name"
```
FuJa0815, line 97:
```suggestion
text = "Logged in as: name"
```
To match the en.po file

src/steam/ISteamClient.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Godot;
+
+/// <summary>
+///   Interface specifying the Steam operations we use. Implemented in the SteamClient.cs file
+/// </summary>
+public interface ISteamClient : ISteamSignalReceiver
+{
+    /// <summary>
+    ///   True if steam client has been initialized
```
FuJa0815, line 9:
```suggestion
    ///   True if the Steam client has been initialized
```

src/steam/ISteamClient.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// <summary>
+    ///   Sets up receiver to receive callbacks
+    /// </summary>
+    /// <param name="receiver">The receiver. Must implement <see cref="ISteamSignalReceiver"/></param>
+    /// <remarks>
+    ///   <para>
+    ///     Note that for many of the received signals they need to be forwarded to this object for further processing.
+    ///   </para>
+    /// </remarks>
+    void ConnectSignals(Object receiver);
```
FuJa0815, line 26: Wouldn't something like this work here?
```cs
    void ConnectSignals<T>(T receiver)
        where T : Object, ISteamSignalReceiver;
```

src/steam/SteamClient.cs
```diff
@@ -0,0 +1,129 @@
...
+    public bool? IsOnline { get; private set; }
+    public ulong? SteamId { get; private set; }
+    public bool? IsOwned { get; private set; }
+
+    public string DisplayName => Steam.GetPersonaName();
+
+    public void Init()
+    {
+        if (initStarted)
+            throw new InvalidOperationException("Can't initialize steam more than once");
```
FuJa0815, line 24:
```suggestion
            throw new InvalidOperationException("Can't initialize Steam more than once");
```

src/steam/SteamClient.cs
```diff
@@ -0,0 +1,129 @@
...
+            SetError(TranslationServer.Translate("STEAM_CLIENT_INIT_FAILED"), stats["verbal"] as string);
+            return;
+        }
+
+        GD.Print("Steam load finished");
+        IsLoaded = true;
+
+        RefreshCurrentUserInfo();
+
+        if (IsOwned == true)
```
FuJa0815, line 44: Can it happen that somebody runs the executable via Steam without owning the game? Why is this check here?

src/steam/SteamHandler.cs
```diff
@@ -0,0 +1,192 @@
...
+        steamClient?.Process(delta);
+    }
+
+    /// <summary>
+    ///   Should only be called by the Steam handling library when its loaded
+    /// </summary>
+    /// <param name="client">The Steam handler to register</param>
+    /// <exception cref="ArgumentException">If client is invalid</exception>
+    /// <exception cref="InvalidOperationException">If a client has already registered</exception>
+    public void RegisterSteamClient(ISteamClient client)
```
FuJa0815, line 79: I'd make this internal just to show even clearer that this is not meant for public use.

FuJa0815 (changes requested):
I guess adding further Steam operations (achivement tracking) in the future isn't difficult, right?

doc/steam_license_readme.txt
hhyyrylainen: It doesn't sound like that to me but what would you suggest instead?.

make_release.rb
hhyyrylainen: I'm pretty sure it was like this before. Also I refer to mac as mac in plenty of other places, also the ruby gem refers to it like `OS.mac?`

src/steam/SteamClient.cs
hhyyrylainen: I mean, if you get the Steam game files, run them on your computer, and have Steam open but you don't own the game, I think that's the case where this triggers.

src/steam/ISteamClient.cs
hhyyrylainen: That does indeed work.

src/steam/SteamHandler.cs
hhyyrylainen: In my original design this actually *wouldn't* be possible to be internal, as this would be called from a separately loaded DLL...
So this kinda isn't Thrive library internal method.

src/steam/ISteamClient.cs
hhyyrylainen: https://github.com/Revolutionary-Games/Thrive/pull/2720/commits/4bfc47350eed44a4673e7fd42c17e0d38c122bed

hhyyrylainen:
Once achievements are implemented in general, I don't think it is too difficult to feed (and read current status) from the steam achievements system. We also need achievement icons.

doc/steam_license_readme.txt
FuJa0815: Maybe `This version of Thrive is distributed through Valve's digital distribution service Steam by Revolutionary Games Studio`

hhyyrylainen: Well that sounds pretty clunky, to be fair my version does as well, but I'll put this suggestion in.

Kasterisk:
Does the Steam name affects the username in saves? If not, I think now would be a good time to support that. And perhaps the custom username option in the options menu needs special handling in regards to that.

hhyyrylainen:
It doesn't. I can open an issue, but I don't have enough hours in today, tomorrow, and half of Friday to do this extra thing.

hhyyrylainen:
I'll merge this now and prepare to shift focus to the mod loader.

PR #2721, opened 2021-10-31
Quick save slot rotation doesn't work

### LINKED ISSUE #2719: Quick save slot rotation doesn't work
I just noticed that when I quick save (with quick saves to keep configured to be 10) I always get the save to go to the same slot:
```
quick saving microbe stage
Creating a save with name: quick_save_10.thrivesave
save finished, success: True message: Saving succeeded elapsed: 00:00:00.3992559
quick saving microbe stage
Creating a save with name: quick_save_10.thrivesave
save finished, success: True message: Saving succeeded elapsed: 00:00:00.3407616
Jukebox: starting track: res://assets/sounds/soundeffects/microbe-ambience.ogg position: 0
```
so the save rotation system seems broken (at least auto save rotation seems to work normally, but there I have the default limit of 5 configured).

takko:

The + behind the regex capture group separates the digits in different groups.
The + within the capture group reads a 10 as 10 and not as 1 and 0.

So quick save rotation was broken for save names > 9.

**Related Issues**

closes #2719

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/saving/InProgressSave.cs
+++ b/src/saving/InProgressSave.cs
@@ -239,7 +239,7 @@ public class InProgressSave : IDisposable
 
                 foreach (var name in SaveHelper.CreateListOfSaves(SaveHelper.SaveOrder.FileSystem))
                 {
-                    var match = Regex.Match(name, "^save_(\\d)+\\." + Constants.SAVE_EXTENSION);
+                    var match = Regex.Match(name, "^save_(\\d+)\\." + Constants.SAVE_EXTENSION);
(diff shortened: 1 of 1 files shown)


hhyyrylainen (approved):
Quick saving now rotates correctly, and now that it's pointed out I can pretty clearly see the mistake regarding the capture groups. Checked with text search that this probably got all of the places there are save name regexes.

PR #2736, opened 2021-11-05
1880 exception collection was modified in auto evo

### LINKED ISSUE #1880: InvalidOperationException: Collection was modified in auto-evo
Just saw this error being printed and auto-evo failed to run (this was when directly going to the microbe freebuild editor):
```
 Auto-evo failed with an exception: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
  at System.Collections.Generic.Dictionary`2+Enumerator[TKey,TValue].MoveNext () [0x00013] in <46e2faba55964e57bc4a72159b9574b8>:0 
  at AutoEvoRun.GatherInfo () [0x000e9] in /home/hhyyrylainen/Projects/thrive/src/auto-evo/AutoEvoRun.cs:368 
  at AutoEvoRun.Step () [0x00023] in /home/hhyyrylainen/Projects/thrive/src/auto-evo/AutoEvoRun.cs:324 
  at AutoEvoRun.Run () [0x0000f] in /home/hhyyrylainen/Projects/thrive/src/auto-evo/AutoEvoRun.cs:301 

```

This caused the auto-evo to not run at all so this is quite serious.

takko:

The exception from the issue is pretty old. I could not reproduce it, but I think it is dangerous to iterate over a global editable dictionary in a thread.
As GatherInfo is only called once per AutoEvo run, the ToList() is not a big addition.

**Related Issues**

closes #1880

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/auto-evo/AutoEvoRun.cs
+++ b/src/auto-evo/AutoEvoRun.cs
@@ -306,8 +306,12 @@ public class AutoEvoRun
 
         foreach (var entry in map.Patches)
         {
+            // TODO: No one should be allowed to update the SpeciesInPatch.
+            // If this happens, the root cause must be addressed.
+
             // Iterate over a copy to be secure from changes to the dictionary.
-            foreach (var speciesEntry in entry.Value.SpeciesInPatch.ToList())
+            var speciesInPatchCopy = entry.Value.SpeciesInPatch.ToList();
+            foreach (var speciesEntry in speciesInPatchCopy)
             {
                 if (alreadyHandledSpecies.Contains(speciesEntry.Key))
                     continue;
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
I guess this resolves the immediate issue, but we won't know if this now masks an issue there should *not* be any code updating that dictionary once auto-evo starts. With this PR, if such a change happens we won't ever find out...
Though I'm willing to accept this like this, because the issue is extremely rare, I've not seen this for months, if you add a TODO comment saying that no one should be allowed to update it and we should still be on the lookout for the root cause.

takko:
If you like I could add code checking if the Dictionary changed (checking length) after the loop and print something.

hhyyrylainen:
Maybe the length doesn't actually change? I don't actually know, but in C# would adding the same item again to a dictionary invalidate iteration even if the item number doesn't change?

FuJa0815:
I don't think this changes anything. You cannot modify a collection while iterating it via foreach no matter if it's a enumerable or a list.

takko:
Yes, adding the same value would raise that exception. I could also check every Item in the Dictionary, but that costs more time, just for a check, but I think it's still nowhere to be realized.
@FuJa0815 this is the common use case if you enumerate something and want to change it while enumerating. ToList() creates a copy. So you enumerate another object. This is widely used, but with a game loop you have to be careful as you don't want to create a copy every frame.

takko:
Added the TODO and a check.
The check will not find if the same value was written.
Also this fix is not 100% as the ToList() can also fail with the same Exception.
The Problem is, that it is not a ConcurrentDictionary, but used on multiple threads.
I could also add a try/catch to further save it.

hhyyrylainen:
> The Problem is, that it is not a ConcurrentDictionary, but used on multiple threads.

Kind of true. The overall design here is that it is *forbidden* from being modified during gameplay. Only during game initialization and auto-evo results being applied after the run finishes should it be modified. That's the problem that there is some buggy piece of code that for some reason very rarely causes that exception.

hhyyrylainen (approved):

PR #2738, opened 2021-11-06
Default font doesn't display "ž" correctly (in czech translation)

### LINKED ISSUE #2005: Default font doesn't display "ž" correctly (in czech translation)
The default font causes `ž` to disappear and a mark above the previous letter to appear.
Here's correct (when changed explicitly to Lato font):

And here's how it looks with the default font:

I suspect that the default font has a broken definition for `̌` as ž is a character composed of two parts:
Kuvakaappaus - 2021-01-03 20-23-20

and we probably either need to find a newer version of the font, or edit the ttf file ourselves to correct it. I think Jura is the default font, so the jura .ttf files are the ones that need modification.

takko:

Updated jura font version from 2.5.1 to 5.106. The new version was corrected.

**Related Issues**

closes #2005

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/assets/fonts/Jura-DemiBold.ttf
+++ b/assets/fonts/Jura-DemiBold.ttf
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:aa720099a548cb027ff3d65d3a1f4e73af449135d6ee7d6cb4cecdace7dfbc2c
-size 326140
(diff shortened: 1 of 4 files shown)


Kasterisk (approved):
Font looks better too apparently so this is great improvement all around.

hhyyrylainen (approved):
The problem is now gone, and the font definitely looks different. I guess we'll see how people like the new look in due time.

PR #2743, opened 2021-11-07
created the hex grid within the shader

### LINKED ISSUE #2274: Fix warnings about non-power-of-2 textures
When exporting this warning is printed (for a few files):
```
WARNING: import: res://assets/textures/background_particle_fuzzy.png: Imported a repeating texture with a size of 182x182, but the project is configured to allow falling back to GLES2.
Non-power-of-2 repeating textures may not display correctly on some platforms such as HTML5. This is because GLES2 does not mandate support for non-power-of-2 repeating textures.
```

I guess a fix is to change the textures that don't need to repeat to have the non-repeat flag on them (build cache versions need to be bumped with this), and remaining files are fixed to have power of 2 dimensions.

takko:

While trying to fix #2274 I realized that the editor_grid.png is a non power of 2 texture but uses repeating.
Checking the shader I found the todo:
 > TODO: implement our own version of procedural hex grid tiling for a better looking and more uniform hexes.

So here it is, a hex grid created within the shader.

I'm absolutely new to that shader stuff, so feel free to optimize or reject.

**Related Issues**

#2274

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


shaders/EditorHexGrid.shader
```diff
@@ -1,27 +1,50 @@
...
 void vertex(){
-    worldPos = (WORLD_MATRIX * vec4(VERTEX, 1.0)).xyz;
+    worldPos = (CAMERA_MATRIX * vec4(0., 0., 0., 1.)).xz;
 }
 
 void fragment(){
-    vec4 gridTex = texture(gridTexture, (worldPos.xz / gridSize) *
-        vec2(2.778, 4.811));
+    // Antialiasing for small grids.
+    float aa = fwidth(VERTEX.xy).x;
```
Kasterisk, line 33: `fwidth` doesn't work on GLES2. To preserve as much compatibility with older devices, maybe an alternative calculation could be used here or another way would be for the shader to support the two approach: procedural and texture based, these two would be toggleable. The procedural grid would be the default and if GLES2 is detected then the shader switches to the texture grid, I think this can be done by setting a uniform bool variable through the microbe editor script.

Kasterisk (changes requested):

shaders/EditorHexGrid.shader
takko: fwidth is only used to make the lines thicker when the camera is far away. So I can take the camera height for that.
The problem with GLES2 and the bitmap is, that it is non 2 square and repeating.

hhyyrylainen: Well the warning is about some implementations not working in GLES2 with those kind of repeating textures. It seems to not be a problem on desktop platforms, but some other platforms probably have issues and warranted Godot adding the warning.

And yeah, 100% the shader needs to work with GLES2 mode, or a workaround needs to be added.

takko: updated shader to work with GLES2

shaders/EditorHexGrid.shader
```diff
@@ -1,27 +1,48 @@
...
-uniform vec2 gridSize;
+const vec2 hexSize = vec2(1.7320508, 1); // 1.7320508 = sqrt(3)
 
 varying vec3 worldPos;
+varying float minLineWidth;
+
+// Returns normalized distance to the nearest hex center.
+float calcHexCenterDistance(vec2 coord)
+{
+    vec4 hexCenter = round(vec4(coord, coord - vec2(1., .5)) / hexSize.xyxy);
```
hhyyrylainen, line 21: I really dislike writing numbers out as `1.` or `.5` I would greatly appreciate if you could change all these to include the leading / trailing zero `1.0, 0.5`

hhyyrylainen:
Does this now work on GLES 2?

shaders/EditorHexGrid.shader
```diff
@@ -1,27 +1,53 @@
...
+varying float minLineWidth;
+
+// Returns normalized distance to the nearest hex center.
+float calcHexCenterDistance(vec2 coord)
+{
+    vec4 hexCenter = round(vec4(coord, coord - vec2(1.0, 0.5)) / hexSize.xyxy);
+    vec4 offset = vec4(coord - hexCenter.xy * hexSize, coord - (hexCenter.zw + 0.5) * hexSize);
+    vec2 position = dot(offset.xy, offset.xy) < dot(offset.zw, offset.zw) ? abs(offset.xy) : abs(offset.zw);
+    return max(dot(position, hexSize * 0.5), position.y) + 0.5;
+}
```
FuJa0815, line 25: Please make it more clear that this implementation is an adapted variant of Andrew Hung's code

FuJa0815 (changes requested):
Looks very good ingame!

shaders/EditorHexGrid.shader
hhyyrylainen: Is there any licensing info available regarding that tutorial? I think it's a bit iffy if we have copypasted an entire function from there without any kind of info regarding what's the license for that code...

FuJa0815: No mentioning of any licensing on the blog anywhere, but the closer
```
Armed with these functions, you now have everything you need to create some cool and unique hex-based effects.
[...]

Happy shader-writing! The only limit is your imagination. And your coding ability.
```
stongly suggests that he intends that the code he showed is meant to be used in projects

takko: Further adopted the code. But still, this is a tutorial. I learned this technique from there. There will always be some recognizable matches. The function is not the same. It is heavily adopted to our use case.

I have absolutely no experience with licenses. So I'm not sure about this.

shaders/EditorHexGrid.shader
```diff
@@ -1,27 +1,51 @@
...
 }
 
 void fragment(){
-    vec4 gridTex = texture(gridTexture, (worldPos.xz / gridSize) *
-        vec2(2.778, 4.811));
     vec4 mask = texture(maskTexture, UV);
 
-    ALBEDO = gridTex.rgb * color.rgb;
-    ALPHA = gridTex.a * mask.a;
+    if (mask.a == 0.0){
```
hhyyrylainen, line 37: Is this actually necessary? I see a multiply by mask.a in ALPHA later. I'm not a super shader expert so this might be incorrect, but I think in shaders if statements are way more expensive than doing unnecessary processing (as GPUs are not very efficient regarding branching).

hhyyrylainen (changes requested):
Looked pretty good in game, though at 720p I see that the lines are slightly jaggy, at higher resolutions it looks pretty good except when the camera is zoomed all the way out, that only looks good at 4K without jagged edges.

shaders/EditorHexGrid.shader
takko: I thought so two, but found this [link](https://stackoverflow.com/a/30256689/2236081)
So I think this improves a bit.

hhyyrylainen: I guess the vast majority of pixel shader executions do early exit here, and the ones that don't are all grouped.

Kasterisk:
There now some sort of black borders around the edges of the grid plane. This is on GLES2.

Also, the grid goes off sync the further you go from the center, this is the problem I also had with the previous method and I think I fixed it by meticulously adjusting the UV offset. Similar solution is probably applicable here.

hhyyrylainen:
Strange that I didn't see that black area (even in GLES2 mode). Maybe that's related to the early alpha return?

takko:
Grid off sync seemed to be an rounding problem. With 1.299 instead of 1.3 i could not get it off sync.
The black corners really seem the come with this early exit. I could not directly reproduce it, but when turning on alpha_scissors these edges appeared with early exit and disappeared without.
Anyway I realized, that the shader only renders a small square around the mask texture. So the early exit may have been worthless anyway.

I could not get rid off those jaggy lines. I guess this would need some kind of super sampling(rendering to higher resolution and than scale down). Not sure if I should give this a try or if this would be overkill.

hhyyrylainen:
I wonder if the smoothness when using the texture is a result of downscaling the texture file. 
I guess this is fine enough even with the slight jagginess of the lines to merge as long as the other problems are solved. I can open an issue to improve the line jagginess later.

Kasterisk (approved):
No more major visual oddities found, overall looks great!

hhyyrylainen (approved):
Code looks fine now.

PR #2744, opened 2021-11-07
update to Godot v3.4

### LINKED ISSUE #2742: Update to Godot 3.4
Godot 3.4 officially released today. a blog post covering the main changes can be found [here](https://godotengine.org/article/godot-3-4-is-released).

Unfortunately thrive does not currently build on 3.4. There are five errors related to `ulong` and `long` types. Typically 3.x releases shouldn't break compatibility, and I'm not great at C# so I'm not sure what's going on here. Either something in the engine broke, changed, or it's finding errors it wasn't finding before. If something in Godot is broken an issue should be opened on its repository.

takko:

changed godot version.
uploaded new godot ci (v14)

changed some errors with the new Godot version where longs were changed to ulongs.

Game is running, just did a quick test.

**Related Issues**

closes #2742

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
As I mentioned in the issue about upgrading Godot, I'm currently working on getting the steam builds working (which require me to make custom Godot builds with mono and GodotSteam) so this update requires extra work from me regarding that. And I haven't yet managed to get the Windows builds working, so I'm going to say that this might not make it in for 0.5.6. I'll look at this after I've prepared everything else for the release (so I'll look at this on the 19th before the RC on the 20th, if I'm not trying to finish everything else at breakneck speed).

src/engine/GodotFileStream.cs
```diff
@@ -17,11 +17,11 @@ public GodotFileStream(File file)
     public override bool CanRead => true;
     public override bool CanSeek => true;
     public override bool CanWrite => true;
-    public override long Length => file.GetLen();
+    public override long Length => checked((long)file.GetLen());
```
hhyyrylainen, line 20: Would it be cleaner to change our data types regarding file lengths to use ulongs? I don't think we do much match with the file lengths, so I think it should be a safe change.

takko: Yes, problem is, we subclass IO.Stream, which needs long.

hhyyrylainen: Ah, I see. I guess then this'll have to do. I guess some of the stuff that doesn't follow a fixed interface could be changed, but it would be pretty weird to use half and half of long and ulong for file lengths.

hhyyrylainen:
Here's some notes I think we should check that I picked up on in the Godot 3.4 release notes:
- godot 3.4 frame delta smoothing is on by default
- godot 3.4 negative frame delta is finally fixed
- godot 3.4 has physical keys, and now also better axis handling built in (or we can use our own with get_action_strength)

hhyyrylainen:
Did anyone check those? Now would be the time to get this merged.

RacerBG:
By the way the issue of this PR is still assigned to release 0.5.7.

hhyyrylainen:
Thanks for the heads up.

hhyyrylainen:
Looks like ultimately that the 3.4 caused at least one bug (on top of the video player "issue"): https://github.com/Revolutionary-Games/Thrive/issues/2959

PR #2750, opened 2021-11-09
Add fading between list tracks in jukebox

### LINKED ISSUE #1130: Add fading between list tracks in jukebox
microbe-ambience and microbe-ambience-2 have a noticeable change between them and it would be nicer with a fade.

takko:

Added fading between tracks.
    Added separate operations for each AudioPlayer to fade only that track.

**Related Issues**

closes #1130

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/general/MusicCategory.cs
```diff
@@ -23,24 +23,21 @@ public enum ReturnType
...
+        None,
 
-    public enum TrackTransitionType
-    {
         /// <summary>
-        ///   No transition between tracks
+        ///   There is a fade between the tracks/categories
         /// </summary>
-        None,
+        Fade,
```
FuJa0815, line 33: I'd name this `Crossfade` so that it's clear what type of fading is meant.

FuJa0815 (changes requested):

src/general/Jukebox.cs
```diff
@@ -262,21 +301,48 @@ private void OnCategoryChanged()
...
     }
 
-    private void AddVolumeChange(float duration, float startVolume, float endVolume)
+    private void AddWait(float duration, AudioPlayer audioPlayer = null)
     {
-        var data = new TimedOperationData(duration) { StartVolume = startVolume, EndVolume = endVolume };
+        var data = new TimedOperationData(duration, audioPlayer);
 
-        operations.Enqueue(new Operation(delta =>
+        var ops = operations;
```
hhyyrylainen, line 318:
```suggestion
        var targetOperations = operations;
```

src/general/Jukebox.cs
```diff
@@ -262,21 +301,48 @@ private void OnCategoryChanged()
...
+
+            return false;
+        }));
+    }
+
+    private void AddVolumeChange(float duration, float startVolume, float endVolume, AudioPlayer audioPlayer = null)
+    {
+        var data = new TimedOperationData(duration, audioPlayer) { StartVolume = startVolume, EndVolume = endVolume };
+
+        var ops = operations;
```
hhyyrylainen, line 339:
```suggestion
        var targetOperations = operations;
```

src/general/Jukebox.cs
```diff
@@ -287,32 +353,44 @@ private void AddVolumeChange(float duration, float startVolume, float endVolume)
...
 
             return false;
         }));
     }
 
-    private void AddVolumeRestore()
+    private void AddVolumeRestore(AudioPlayer audioPlayer = null)
     {
-        operations.Enqueue(new Operation(_ =>
+        var ops = operations;
```
hhyyrylainen, line 370:
```suggestion
        var targetOperations = operations;
```

src/general/Jukebox.cs
```diff
@@ -287,32 +353,44 @@ private void AddVolumeChange(float duration, float startVolume, float endVolume)
...
+            SetVolume(NORMAL_VOLUME, audioPlayer);
             return true;
         }));
     }
 
-    private void AddVolumeRemove()
+    private void AddVolumeRemove(AudioPlayer audioPlayer = null)
     {
-        operations.Enqueue(new Operation(_ =>
+        var ops = operations;
```
hhyyrylainen, line 385:
```suggestion
        var targetOperations = operations;
```

src/general/MusicCategory.cs
```diff
@@ -23,24 +23,21 @@ public enum ReturnType
...
     {
         /// <summary>
-        ///   There is a fade between the categories
+        ///   No transition between tracks/categories
         /// </summary>
-        Fade,
-    }
+        None,
 
-    public enum TrackTransitionType
```
hhyyrylainen, line 31: I guess it kinda makes sense to combine these two enums into one. At least I can't right now think of a case where a value would be valid for one and not make sense in the other one.

src/general/Jukebox.cs
```diff
@@ -223,6 +248,14 @@ private void PlayTrack(AudioPlayer player, TrackList.Track track, string trackBu
 
         if (changedTrack || !player.Playing)
         {
+            var target = categories[PlayingCategory];
+            if (target.TrackTransition == MusicCategory.Transition.Crossfade)
```
hhyyrylainen, line 252: Could you add a TODO comment here that it might be nice to skip the fade if the *same* track is going to be played again? (so a special TrackTransition would be added that only triggers the fade if changedTrack is true) With microbe ambiance2 I think that would sound better when it loops that it doesn't have the fade in the middle of it.

src/general/MusicCategory.cs
hhyyrylainen: Though maybe with my latest review comment it does make sense to keep these separate?

hhyyrylainen (changes requested):
Seems to work fine in-game, just a few code cleanup comments.

src/general/MusicCategory.cs
takko: I think this option could also be applied to category change. If changing to same category skip fade.
I currently can't think of a transition that is only valid for tracks or categories.
Also as long, as the Transition enum is not used within a safe, this can be splitted easily.

But it's up to you.

hhyyrylainen: Okay fine, let's go with the combined enum for now, and change it in the future if needed.

takko:
But the FadeOut is added for the end of the track, when the track is started. You can skip FadeIn easily with the changed track, but the FadeOut needs to know which track is playing next.

hhyyrylainen:
Ah I see. I didn't realize that point. Okay, it's fine then. In the future though doing something like always keeping a queue of 2 things to play would be a doable design, I think.

hhyyrylainen (approved):
Code looks fine now, I assume the one small renaming things commit didn't break this, so I won't re-playtest this.

PR #2753, opened 2021-11-14
Compound production

### LINKED ISSUE #570: Photosynthesisers should affect environmental oxygen
Split from this issue: https://github.com/Revolutionary-Games/Thrive/issues/553

- [ ] It should be possible for the cell species to affect the amount of oxygen and CO2 (photosynthesisers)

Maxonovien:

This PR adds production/consumption on gas compounds in patches depending on population in them.
It also paves the way for further dynamism in patches composition.

Planned changes:
- [x] Add compound amount manipulation infrastructure (patch volume...)
- [x] Have input/output of processes from microbes affect the world
- [x] Add a save upgrader
- [x] Add gas output to microbe processes (to avoid depletion)
- ~~[ ] Add patch volume in GUI (larger patches are more stable) & translation support~~
- [x] Add constant input to patches for balance
- [x] Balance production factor to get an interesting/not game-breaking evolution of patches

**Related Issues**

Closes #570.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -77,7 +77,7 @@ public static PatchMap Generate(WorldGenerationSettings settings, Species defaul
         map.AddPatch(bathypelagic);
 
         var abyssopelagic = new Patch("PATHCH_PANGONIAN_ABYSSOPELAGIC", 5,
-            GetBiomeTemplate("abyssopelagic"))
+            GetBiomeTemplate("abyssopelagic"), (long)(2000 * 2000) * 2000)
```
hhyyrylainen, line 80: `2000L`?

src/general/IWorldEffect.cs
```diff
@@ -53,3 +54,71 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+        this.targetWorld = targetWorld;
+    }
+
+    public void OnRegisterToWorld()
+    {
+    }
+
+    public void OnTimePassed(double elapsed, double totalTimePassed)
+    {
+        List<Compound> gasCompounds = SimulationParameters.Instance.GetGasCompounds();
```
adQuid, line 74: Will this list be changing? I'd look into doing this just once in OnRegisterToWorld

src/microbe_stage/Patch.cs
```diff
@@ -23,6 +23,12 @@ public class Patch
     [JsonProperty]
     public readonly int[] Depth = new int[2] { -1, -1 };
 
+    /// <summary>
+    ///   The volume of the patch, in meters cube.
```
adQuid, line 27: Do we really want to restrict ourselves to exact units here?

src/microbe_stage/ProcessSpeedInformation.cs
```diff
@@ -37,7 +37,11 @@ public ProcessSpeedInformation(BioProcess process)
     public IReadOnlyDictionary<Compound, float> FullSpeedRequiredEnvironmentalInputs =>
         WritableFullSpeedRequiredEnvironmentalInputs;
 
-    public IEnumerable<KeyValuePair<Compound, float>> Outputs => WritableOutputs;
+    public IEnumerable<KeyValuePair<Compound, float>> Outputs =>
+        WritableOutputs.Where(p => !p.Key.IsEnvironmental);
+
+    public IEnumerable<KeyValuePair<Compound, float>> EnvironmentalOutputs =>
+        WritableOutputs.Where(p => p.Key.IsEnvironmental);
```
adQuid, line 44: Why were these needed?

adQuid (commented):
Overall surprisingly simple. No complaints on the grand scale.

src/general/IWorldEffect.cs
hhyyrylainen: I'd slightly prefer this being here, because after all this runs just once for each editor entry. And re-checking the gas compounds list each time would make this more compatible with loading mods after creating a save that add more compound types.

hhyyrylainen:
Is this related to https://github.com/Revolutionary-Games/Thrive/issues/570 ?

simulation_parameters/SimulationParameters.cs
```diff
@@ -184,6 +184,26 @@ public List<Compound> GetCloudCompounds()
         return result;
     }
 
+    /// <summary>
+    ///   Returns all *molecular* compounds that are dissolved in the environment, i.e. gas.
+    /// </summary>
+    /// <remarks>This excludes sunlight</remarks>
```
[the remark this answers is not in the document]
Maxonovien: No, it is actually limited to environmental compounds like CO2, O2 and N. Perhaps some rephrasing could phrase this better indeed.

src/microbe_stage/Patch.cs
Maxonovien: Tough question that has been haunting my nights for a while (do I exaggerate here? Yes, definitely.).

Depth itself is in meters, and so the volume, defined as deltaDepth^3 is in meters cube. 
I feel like having real units helps getting a clear view of what values represent physically, but I agree it is a restriction.

I don't have much of an opinion on this.

src/microbe_stage/ProcessSpeedInformation.cs
Maxonovien: These are not useful as of now because there is no environmental output, but it will start to be useful when some processes like photosynthesis will start rejecting environmental compounds. Yet, the previous infrastructure assumes there is no environmental outputs, so I filtered them out to keep things working.

Having access to Environmental ones rather than the full list is just sugar to mimick inputs and help readability.

Maxonovien:
It is. I mostly put the draft up to get it accessible as I wanted Thim's opinion on it before it gets any more complicated, so I did not bother to link issues. I'll have to check for other relations, but I can add this issue to the list already.

buckly90:
You're making great progress, I'm happy with how it's turning out. Noticed oxygen became very abundant in the surface patches, while the depths were very abundant with CO2.

The most glaring problems I see so far are that patches end up with little to no CO2 or O2, depending on where you are; as well as the effects of gasses on processes making them become very powerful.
Oh, and the tidepool was stripped of it's gasses rather spontaneously, which was odd.

I suggest that the rate of gas exchange between patches be increased so that the oxygen and CO2 produced in both environments mixes a little better. In addition, and this doesn't need to be done now, I believe we should adjust how processes scale with gasses so that they grant diminishing returns.

Keep up the great work!

Kasterisk:
Any numerical compound amount percentage display in the UI needs rounding to only 3 decimal places, similar to how the compound amounts is handled.

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -16,7 +16,7 @@ public static PatchMap Generate(WorldGenerationSettings settings, Species defaul
 
         // Predefined patches
         var vents = new Patch("PATCH_PANGONIAN_VENTS", 0,
-            GetBiomeTemplate("aavolcanic_vent"))
+            GetBiomeTemplate("aavolcanic_vent"), 500 * 500 * 500)
```
Kasterisk, line 19: Can a `Math.Pow` be used for these?

src/saving/ISaveUpgradeStep.cs
```diff
@@ -104,7 +105,7 @@ internal class UpgradeStep055To056 : BaseRecursiveJSONWalkerStep
         protected override void CheckAndUpdateProperty(JProperty property)
         {
             var children = property.Value.Children<JProperty>();
-            var childrenNames = children.Select(c => c.Name);
+            var childrenNames = children.Select(c => c.Name).ToList();
```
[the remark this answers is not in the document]
Maxonovien: Oh, it's probably an old change (I think I originally had the save upgrade step in this call), and I did not remove all the changes I did when moving all to a new step. I'll correct this tomorrow. Good catch!

simulation_parameters/SimulationParameters.cs
```diff
@@ -190,6 +190,26 @@ public List<Compound> GetCloudCompounds()
         return result;
     }
 
+    /// <summary>
+    ///   Returns all environmental *molecular* compounds that are dissolved in the environment, i.e. gas.
+    /// </summary>
+    /// <remarks>This excludes sunlight, and includes O2, CO2, N...</remarks>
+    public List<Compound> GetGasCompounds()
```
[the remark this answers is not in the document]
Maxonovien: Get Cloud compounds was similar to my implementation, but I agreed that it would be better as LINQ, so I changed both.

adQuid (changes requested):
Just one thing I would change. I've also noticed that nothing releases nitrogen into the environment so we'll be dependent on the hand of God until we add something, but I think that that is outside the scope of this change.

src/general/IWorldEffect.cs
```diff
@@ -93,3 +95,167 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+                        if (individualCompoundProduction.TryGetValue(compound, out CompoundBalance compoundProduction))
+                        {
+                            if (!compoundsProduced.ContainsKey(compound))
+                                compoundsProduced[compound] = 0;
+
+                            compoundsProduced[compound] += compoundProduction.Balance * patch.SpeciesInPatch[species] *
+                                Constants.DISSOLVED_PRODUCTION_FACTOR;
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
Maxonovien: Shouldn't that start with a non-capital letter? But good idea otherwise?

Maxonovien:
Oh, you mean nothing releases it IRL? Actually, there might be some azote trapped within cavities getting slowly rejected to the ocean, or whatever.

src/general/IWorldEffect.cs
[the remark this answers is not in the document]
adQuid: You're right; that's what I get for coding in github.

adQuid:
> Oh, you mean nothing releases it IRL? Actually, there might be some nitrogen trapped within cavities getting slowly rejected to the ocean, or whatever.

I have no idea how real science works. I just meant in the simulation no biological process releases it, so we have to remember to never turn off the magic nitrogen hose in a future update.

src/microbe_stage/Patch.cs
```diff
@@ -24,6 +24,12 @@ public class Patch
...
```
hhyyrylainen, line 28:
```suggestion
    ///   The volume of the patch, in meters cubed.
```
Isn't "meters cube" a typo? At least I've seen more "cubed" so it sounds more correct to me.

src/microbe_stage/ProcessSpeedInformation.cs
hhyyrylainen: I'd like a comment here detailing this aspect so that when someone is fixing this up properly they'll have the necessary context to know to change these back.

src/saving/ISaveUpgradeStep.cs
```diff
@@ -162,6 +163,174 @@ private void UpgradeBehaviouralValues(JProperty property, JEnumerable<JProperty>
...
+            {
+                if (depth[1].Value<int?>() == null || depth[0].Value<int?>() == null)
+                {
+                    throw new ArgumentException("Depth array does not start with two integers!");
+                }
+
+                var depthDifference = depth[1].Value<int>() - depth[0].Value<int>();
+
+                // Assume cubic patches for upgrade
+                container.Add("Volume", (long)Math.Pow(depthDifference, 3));
```
hhyyrylainen, line 268: Why is this done like this? Surely it would be better to either: specify the volume in the biome definitions *or* also update the patch generator to use the same depth difference logic, otherwise I fear there's too much highly specific math duplication between this and the patch generation.

src/saving/ISaveUpgradeStep.cs
```diff
@@ -162,6 +163,174 @@ private void UpgradeBehaviouralValues(JProperty property, JEnumerable<JProperty>
...
+        ///       ...
+        ///   </para>
+        /// </remarks>
+        private void UpgradePatchesVolume(JObject container, JEnumerable<JProperty> children)
+        {
+            if (children.First(p => p.Name == "Depth").Value is JArray depth)
+            {
+                if (depth[1].Value<int?>() == null || depth[0].Value<int?>() == null)
+                {
+                    throw new ArgumentException("Depth array does not start with two integers!");
```
hhyyrylainen, line 262: Should throw json exceptions like the other upgrade steps as these are about invalid json data.

src/saving/ISaveUpgradeStep.cs
```diff
@@ -162,6 +163,174 @@ private void UpgradeBehaviouralValues(JProperty property, JEnumerable<JProperty>
...
+            var childrenNames = children.Select(c => c.Name).ToList();
+
+            // Add volume for named patches (properties)
+            if (childrenNames.Contains("Depth") && !childrenNames.Contains("Volume"))
+            {
+                UpgradePatchesVolume((JObject)property.Value, children);
+            }
+
+            // Most patches are defined through references to non-named tokens of array "Adjacent"
+            if (property.Name == "Adjacent")
```
hhyyrylainen, line 182: I'm a bit suspicious about this as I thought that the base step should already recurse through everything... so in my head, this design should not be required. It seems the call to the base CheckAndUpdateProperty is missing in this method.

simulation_parameters/SimulationParameters.cs
```diff
@@ -177,17 +178,16 @@ public bool DoesCompoundExist(string name)
...
-            {
-                result.Add(entry.Value);
-            }
-        }
+        return compounds.Values.Where(c => c.IsCloud).ToList();
+    }
 
-        return result;
+    /// <summary>
+    ///   Returns all environmental *molecular* compounds that are dissolved in the environment, i.e. gas.
```
hhyyrylainen, line 185: I feel like this comment should be a bit reworded to be just "Returns compounds that are dissolved in the environment (aren't clouds)" just so that the comment won't become incorrect in the future.

simulation_parameters/SimulationParameters.cs
```diff
@@ -177,17 +178,16 @@ public bool DoesCompoundExist(string name)
...
+    /// </summary>
+    /// <remarks>This excludes sunlight, and includes O2, CO2, N...</remarks>
```
hhyyrylainen, line 187: remarks needs to have a para tag in it and it needs to be formatted like shown in the styleguide.

simulation_parameters/SimulationParameters.cs
```diff
@@ -177,17 +178,16 @@ public bool DoesCompoundExist(string name)
...
+    public List<Compound> GetGasCompounds()
+    {
+        return compounds.Values.Where(c => c.IsCloud && c.IsEnvironmental && c.CanBeDistributed).ToList();
```
hhyyrylainen, line 190: I feel a bit confused by these flag checks here. It *should* be the case that IsCloud and IsEnvironmental can't be true for the same compound. Also CanBeDistributed is related to colony compound sharing, so this confuses me a lot.

src/general/IWorldEffect.cs
```diff
@@ -93,3 +95,167 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+    {
+        List<Compound> gasCompounds = SimulationParameters.Instance.GetGasCompounds();
+
+        foreach (var patch in targetWorld.Map.Patches.Values)
+        {
+            // First, add the constant intake to the patch, since it will influence the available amount for species.
+            // TODO move to own effect
+            AddConstantIntakeToPatch(patch);
+
+            Dictionary<Compound, float> compoundsProduced = new Dictionary<Compound, float>(gasCompounds.Count);
```
hhyyrylainen, line 124: You seem to be a bit underusing `var` in general, but especially this line is unnecessarily long due to not using `var`.

src/general/IWorldEffect.cs
```diff
@@ -93,3 +95,167 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+                }
+            }
+
+            compoundsProduced = ScaleByOverconsumption(compoundsProduced, patch);
+
+            foreach (var compound in compoundsProduced.Keys)
+            {
+                if (patch.Biome.Compounds.TryGetValue(compound, out EnvironmentalCompoundProperties compoundValue))
+                {
+                    // TODO if capped here, use something to scale production
```
hhyyrylainen, line 155:
```suggestion
                    // TODO: if capped here, use something to scale production
```

src/general/IWorldEffect.cs
```diff
@@ -93,3 +95,167 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+    public void AddConstantIntakeToPatch(Patch patch)
+    {
+        var oxygen = SimulationParameters.Instance.GetCompound("oxygen");
+        var carbonDioxide = SimulationParameters.Instance.GetCompound("carbondioxide");
+        var nitrogen = SimulationParameters.Instance.GetCompound("nitrogen");
+        var constantCompoundsDissolvedIntake = new Dictionary<Compound, float>(3);
+
+        constantCompoundsDissolvedIntake[oxygen] = Constants.PATCH_CONSTANT_OXYGEN_INPUT;
+        constantCompoundsDissolvedIntake[carbonDioxide] = Constants.PATCH_CONSTANT_CARBON_DIOXYDE_INPUT;
+        constantCompoundsDissolvedIntake[nitrogen] = Constants.PATCH_CONSTANT_NITROGEN_INPUT;
```
hhyyrylainen, line 181: You can rewrite this as a dictionary with initializer.

src/general/IWorldEffect.cs
```diff
@@ -93,3 +95,167 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+
+        foreach (var intake in constantCompoundsDissolvedIntake)
+        {
+            if (patch.Biome.Compounds.TryGetValue(intake.Key, out EnvironmentalCompoundProperties compoundValue))
+            {
+                // TODO if capped here, use something to scale production
```
hhyyrylainen, line 187:
```suggestion
                // TODO: if capped here, use something to scale production
```

src/general/IWorldEffect.cs
```diff
@@ -93,3 +95,167 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+    /// <summary>
+    ///   Reduce ALL compound production by a scaling factor if one compounds happens to be overproduced.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     This is a heuristic to prevent production of compounds from non-existing ones.
+    ///     A better yet more complicated approach would be to only limit the processes consuming the target compounds.
+    ///   </para>
+    /// </remarks>
+    public Dictionary<Compound, float> ScaleByOverconsumption(Dictionary<Compound, float> compoundsProduced,
```
hhyyrylainen, line 246: Shouldn't this return void as this modifies the parameter in-place?

hhyyrylainen (changes requested):
I've now reviewed all the code here and left quite many code style comments. I didn't see any big issues regarding the code architecture.

src/saving/ISaveUpgradeStep.cs
```diff
@@ -162,6 +163,174 @@ private void UpgradeBehaviouralValues(JProperty property, JEnumerable<JProperty>
...
+                }
+
+                string targetWorldRef = effectsArray[0]["targetWorld"]?["$ref"]?.ToString() ??
+                    throw new JsonException("Missing field from the first element of effect array: " +
+                        "targetWorld or its $ref!");
+
+                var typeTokenText = "'$type': '" + gasProductionEffectName + "'";
+                var targetWorldTokenText = "'targetWorld': { '$ref': '" + targetWorldRef + "' }";
+                var effectTokenText = "{" + typeTokenText + "," + targetWorldTokenText + "}";
+                var effectToken = JToken.Parse(effectTokenText);
```
hhyyrylainen, line 322: Wouldn't it be better to actually build the JObject properly instead of doing a bunch of string concatenation?
(I've seen some r/programminghorrorposts about manually created JSON strings...)

src/general/IWorldEffect.cs
Maxonovien: I think that for some reason it was not actually doing in place changes, so I used a return instead

src/general/IWorldEffect.cs
```diff
@@ -93,3 +95,167 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+                    patch.Biome.Compounds[compound] = new EnvironmentalCompoundProperties
+                    {
+                        Amount = 0, Density = 0, Dissolved = compoundsProduced[compound] / patch.Volume,
+                    };
+                }
+            }
+        }
+    }
+
+    public void AddConstantIntakeToPatch(Patch patch)
```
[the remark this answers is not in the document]
Maxonovien: Should, I also swapped a bunch of similar methods as such.

src/general/IWorldEffect.cs
hhyyrylainen: `struct` would be a value type, but Dictionary is a `class`

src/general/IWorldEffect.cs
```diff
@@ -93,3 +95,167 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+                    compoundValue.Dissolved = Math.Max(
+                        compoundValue.Dissolved + compoundsProduced[compound] / patch.Volume,
+                        Constants.DISSOLVED_MIN);
+                    patch.Biome.Compounds[compound] = compoundValue;
+                }
+                else
+                {
+                    patch.Biome.Compounds[compound] = new EnvironmentalCompoundProperties
+                    {
+                        Amount = 0, Density = 0, Dissolved = compoundsProduced[compound] / patch.Volume,
```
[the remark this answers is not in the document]
Maxonovien: I did a shared method that should use it properly.

src/general/IWorldEffect.cs
Maxonovien: I guess old habits die hard. Fixed.

simulation_parameters/SimulationParameters.cs
Maxonovien: I agree that by the JSON compounds, this should return an empty list, yet since changes are applied, it probably doesn't... I'm as confused as you on this, I don't recall why I used `IsCloud`... I'll have to test this when I'll have a moment.

`CanBeDistributed` is used to exclude sunlight. That was both the best heuristic I had at hand, and something logical, since globally "gas compounds" should be spread in colonies, unlike whatever compound type is sunlight.

hhyyrylainen: Perhaps if there isn't a good existing flag for it you should add a new property to the Compound class that specifies that it is a gas?

src/saving/ISaveUpgradeStep.cs
Maxonovien: Patch generator is hardcoded. I used cubes as a simple model for patches, but nothing should force patches to be cubes. Now this upgrade step just intends to give a volume to patches of older versions, and since I can't access the exact value (unless I just move every patch volume to `Constants.cs` but then any modification would require to change multiple files at once, which seems terrible an idea, if I may), I just go for the simple model of a cube, which is enough to get older versions to be *playable*.

The only way I see to change this would be to rework the patch generator to work from either Constants or JSONs, and that would be a PR of its own.

src/microbe_stage/Patch.cs
Maxonovien: Actually, the official naming is "cubic metre", as it seems : https://en.wikipedia.org/wiki/Cubic_metre

I'll change that.

src/saving/ISaveUpgradeStep.cs
hhyyrylainen: I more suggested that you add the "Volume" as a parameter for a biome, then use that as the value here and in the patch map generator. I guess if we don't care about the math being off (and thus perhaps balancing as well) when comparing upgraded and other saves, this is fine if that is explained with a comment why this approach is taken.

Maxonovien:
I guess, I just keep forgetting that `out` creates variables outside of the `if` scope...

src/microbe_stage/Patch.cs
hhyyrylainen: At least don't change the spelling to "metre" that spelling is disgusting to me as it's farther away from how the word is actually said (as a Finnish person I'm always offended when a word is not spelled the same way it is pronounced).

hhyyrylainen:
> I guess, I just keep forgetting that out creates variables outside of the if scope...

Yeah, that's something that I didn't know initially either, but I've now found that it is helpful in reducing the amount of code that is needed.

buckly90:
It has been some time since I last looked at this PR and now it seems that the compounds barely change at all now.
I tested this PR with two separate playthroughs, one as a strict carnivorous hunter, and another as a docile photosynthesizer; Both concluded with the same results.

I'm unsure if this is because cells no longer have a significant impact, or if I just happened to live in a world that struck a perfect balance both times.

Either way, when I have some time on my hands I'll see about playing with the numbers and see if I can make things feel right for testing purposes.

src/microbe_stage/Patch.cs
Maxonovien: Well, the thing is, the guideline is to keep by British English, whose spelling is actually "metre". I'd go against adding exception for one's sole personal taste (especially given that we'd then have to change about half the vocabulary by this metric --English be English :) )

src/saving/ISaveUpgradeStep.cs
Maxonovien: You mean having a predetermined volume for a biome? I see no reason why we should have a predefined volume for caves : I think this goes along the same logic as the one used for depth measures.  I'll go along the comment route, I think.

Ultimately, I guess this shows how much the current patch generator system should be updated...

simulation_parameters/SimulationParameters.cs
Maxonovien: I guess it would make sense to have a good flag for sunlight (and potential other similar compounds), but I'm short for ideas at the moment. Unless someone comes by with a great idea, I'll simply add a "isSunlight" flag whenever I have time to wrap everything up.

Maxonovien:
In which patch are you playing ? Remember that larger patch are less affected by the same amount of cells (and since we have very tiny patches like the pond, it must be a bit controlled). 

Also, obviously the idea is not to get it to change drastically every generation, otherwise it would be fairly unplayable. But sure, some help in balancing this would be great!

buckly90:
> In which patch are you playing ? Remember that larger patch are less affected by the same amount of cells (and since we have very tiny patches like the pond, it must be a bit controlled).

I tried all of the surface patches, and the vents.

The vents had no visible change at all, but the epipelagic and tidepools atleast had an occasional 0.0001% or so change in concentration

Maxonovien:
Maybe I went to harsh on the species then ? Well, I'll give it a look when I have time. But as far as I remember I had some decent changes in some places... Perhaps is it also due to a small pop? I'll see.

src/saving/ISaveUpgradeStep.cs
hhyyrylainen: I was suggesting that as it would fit the current patch generation model and also make it consistent when upgrading saves.
But feel free to do this:
> I guess if we don't care about the math being off (and thus perhaps balancing as well) when comparing upgraded and other saves, this is fine if that is explained with a comment why this approach is taken.

simulation_parameters/SimulationParameters.cs
hhyyrylainen: Why not "IsGas"?

src/microbe_stage/Patch.cs
hhyyrylainen: https://github.com/Revolutionary-Games/Thrive/pull/3107

Kasterisk (changes requested):

src/general/GameWorld.cs
```diff
@@ -72,6 +72,9 @@ public GameWorld()
 
         // Register glucose reduction
         TimedEffects.RegisterEffect("reduce_glucose", new GlucoseReductionEffect(this));
+
+        // Register gas compounds production
+        TimedEffects.RegisterEffect("produce_gases", new GasProductionEffect(this));
```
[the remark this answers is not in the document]
hhyyrylainen: Actually seems like this is correct form as this is the plural of gas and not a verb.
https://grammarist.com/usage/gases-gasses/

adQuid (approved):

Maxonovien:
Currently impeded due to the fact that thylakoid species still somewhat thrive without photosynthesis once they appeared. We thus need photosynthetic elements to be a burden under low CO2/high O2. See photosynthesis.

src/general/IWorldEffect.cs
```diff
@@ -93,3 +95,160 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+        this.targetWorld = targetWorld;
+    }
+
+    public void OnRegisterToWorld()
+    {
+    }
+
+    public void OnTimePassed(double elapsed, double totalTimePassed)
+    {
+        List<Compound> gasCompounds = SimulationParameters.Instance.GetGasCompounds();
```
Kemikal1, line 116: i debugged this and for some reason this always gets me gasCompunds.Count = 0

simulation_parameters/SimulationParameters.cs
Kemikal1: For some  reason those flags are never simultaneously true?

hhyyrylainen:
I won't close this PR as there's still probably a chance that this will be finished.

hhyyrylainen:

PR #2754, opened 2021-11-14
Fixes display of speed and ATP consumption of flagella

### LINKED ISSUE #2406: Fix flagella draining atp when visually inactive
When having multiple flagella, with only one visually (and normally) active, for a build able to sustain it (from the editor charts), the atp still decreases.

Further tests highlighted that this could occur for flagella colinearly and (especially) diagonally opposed to the movement.
This occurs even when flagella are placed with valid editor rotation, although it seems more to be in see-saw (major drop then atp rises again).

Perhaps comes from proportional reduction of flagella? 

Perhaps related to #1500 and #1297 . 
_____
Screens from initial, replicated report

One flagella sustainability
image

Final design 
image

Replication with one opposed flagella

image

Kemikal1:

Fixes the speed and ATP cost display.

**Related Issues**
Fixes #2406 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


Kemikal1:
Turns out it was harder then i bargained for.... oufff

buckly90:
Seems to work without any visible error.

I noticed that this makes it so that backwards facing flagella display positive speed instead of negative. When I place two flagella facing opposite directions, it results in an overall decrease in speed. Am I mistaken in assuming that it calculates top speed in one direction? Whatever is the case, it is pulled off quite well so excellent job!

Kemikal1:
Thats indeed the intended function, because a microbe cant go in opossite direction at thr same time. Well for a given microbe theres a single direction that represent the microbes top speed. Maybe we could also display that direction in a new PR.

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -1,39 +1,147 @@
...
+        var organelles = inputOrganelles.Where(o => o.Definition.HasComponentFactory<MovementComponentFactory>())
+            .ToList();
+
+        foreach (var organelle in organelles)
+        {
+            maximumMovementDirection += GetOrganelleDirection(organelle);
+        }
+
+        // After calculating the sum of all organelle directions we substract the movement components which
+        // are symetric and we chose the one who would benefit the max-speed the most.
```
hhyyrylainen, line 20:
```suggestion
        // are symetric and we choose the one who would benefit the max-speed the most.
```

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -1,39 +1,147 @@
...
+
+    private static float MovementForce(float movementForce, float directionFactor)
+    {
+        if (directionFactor < 0)
+            return 0;
+
+        return movementForce * directionFactor;
+    }
+
+    // Symetric flagella are a corner case for speed calculations because the sum of all
```
hhyyrylainen, line 121: This should be an xml remark

src/microbe_stage/ProcessSystem.cs
```diff
@@ -49,14 +49,17 @@ public ProcessSystem(Node worldRoot)
...
+        return ComputeEnergyBalance(organelles, biome, membrane, maximumMovementDirection);
     }
 
     /// <summary>
     ///   Computes the energy balance for the given organelles in biome
     /// </summary>
-    public static EnergyBalanceInfo ComputeEnergyBalance(IEnumerable<OrganelleDefinition> organelles,
-        BiomeConditions biome, MembraneType membrane)
+    public static EnergyBalanceInfo ComputeEnergyBalance(IEnumerable<OrganelleTemplate> organelles,
+        BiomeConditions biome, MembraneType membrane, Vector3 maximumMovementDirection)
```
hhyyrylainen, line 62: `maximumMovementDirection` should be renamed here to something that explains it is used to filter out movementOrganelles that aren't active if they don't allow moving in that direction. Maybe something like `excludeMovementIfCannotPushInThisDirection` though that's a bit wordy

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -1,39 +1,147 @@
...
+    // directions is kinda of broken in their case, so we have to choose which one of the symmetric flagella
+    // we must discard from the direction calculation
+    // Here we only discared if the flagella we input is the "bad" one
+    private static Vector3 ChooseFromSymetricFlagella(IEnumerable<OrganelleTemplate> inputOrganelles,
+        OrganelleTemplate testedOrganelle, Vector3 maximumMovementDirection)
+    {
+        var organelles = inputOrganelles.ToList();
```
hhyyrylainen, line 128: This ToList call is completely unnecessary

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -1,39 +1,147 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using Godot;
 
 public static class MicrobeInternalCalculations
 {
+    public static Vector3 MaximumSpeedDirection(IEnumerable<OrganelleTemplate> inputOrganelles)
```
hhyyrylainen, line 7: Why is this called `inputOrganelles`I think the parameter should just be called `organelles` so that it doesn't confuse the future code writer as to why this takes in "inputOrganelles" and not just "organelles"?
You can rename the local variable to movementOrganelles.

hhyyrylainen (changes requested):
I guess all the parts are necessary, as it's realistic to only calculate the maximum movement force providing flagella energy consumption.
I have some code clarity related comments, I was not able to understand all of the code, but I can let that slide, I didn't really understand the previous speed calculation math either...

RacerBG (approved):

src/microbe_stage/MicrobeInternalCalculations.cs
Kemikal1: jetbrains gives me an error for multiple enumerations if i iterate multiple times through it.

src/microbe_stage/ProcessSystem.cs
Kemikal1: why should i rename  a variable like that? thats not what the variable represents. Armed with some basic vector math and seeing its use in calcualtions you should know in what way and why the variable is used.

src/microbe_stage/MicrobeInternalCalculations.cs
hhyyrylainen: But you do not? I don't see the organelles variable being used anywhere else in this method than on the next line.

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: It doesn't fit the design of ComputeEnergyBalance. It should not have to know that maximum movement direction is the only direction flagella are ever used in. This is a lower level building block that should be made with the assumption that whoever is looking at it doesn't know anything about maximum movement direction, just that if you want this lower level method can exclude any movement organelles that can't move the cell in a certain direction from the computation.

Kemikal1: Totally disagree, when computing the maximum energy consumption of a microbe the direction in which the speed and, thus, the amount of flagella activation is the greatest is pretty relevant since both are tied together.

src/microbe_stage/MicrobeInternalCalculations.cs
Kemikal1: yea, i didint see that there

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -118,15 +119,17 @@ private static float MovementForce(float movementForce, float directionFactor)
...
+    ///  directions is kinda of broken in their case, so we have to choose which one of the symmetric flagella
+    ///  we must discard from the direction calculation
+    ///  Here we only discard if the flagella we input is the "bad" one
+    /// </summary>
+    private static Vector3 ChooseFromSymetricFlagella(IEnumerable<OrganelleTemplate> organelles,
         OrganelleTemplate testedOrganelle, Vector3 maximumMovementDirection)
     {
-        var organelles = inputOrganelles.ToList();
-        foreach (var organelle in organelles)
+        var movementOrganelles = organelles;
```
hhyyrylainen, line 131: Why have this separate variable assignment here?

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: I would be more willing to accept that, if this method wasn't public. This is a public method and now it has this new confusing parameter that doesn't seem to be related to this method's purpose at all. I might accept this naming if you add xml parameter comments.

Kemikal1: For all the parameters?

hhyyrylainen:
I had this TODO note:
After merging this need to open an issue for auto-evo to consider the maximum speed in a direction so it can't cheat by placing sideways flagella that don't cost energy but in the simple calculations they give movement speed. Probably should cache the maximum speed direction and speed value for efficiency

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: I'll do this myself.

hhyyrylainen (approved):
Did the final fixes I wanted. I still don't fully understand why I got the playtesting results I got, but I'll merge this now anyway.

PR #2756, opened 2021-11-15
Organelle models flicker sometimes render order

### LINKED ISSUE #1077: Organelle models flicker sometimes (render order)
editor placed models seem to flicker sometimes when the hover hex changes from valid to invalid, not visible until a lot of organelles are added.

takko:

Calculate different RenderPriorities for the HexPositions of the Organelle. Min hex distance for same priority is 4.

**Related Issues**

closes #1077

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


buckly90:
Upon testing this with a few freebuild creations I noticed no flickering with overlapping organelles, both normal and growing.

Nice job!

src/general/Hex.cs
```diff
@@ -231,19 +231,50 @@ public static Hex FlipHorizontally(Hex hex)
...
+            {
+                if (currentHex == hex)
+                {
+                    return (radius % 6 + 1) * 6 + (indexInRing % 6 + 1);
+                }
+
+                indexInRing++;
+                currentHex += HexNeighbourOffset[(HexSide)(i + 1)];
+            }
+        }
```
hhyyrylainen, line 260: Can't this looping be fixed somehow? This seems to just look for a hex that matches the current given position?
Why can't a direct computation from `hex` to render index without looping be done?

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -2302,9 +2303,26 @@ private void UpdateAlreadyPlacedVisuals()
...
+            {
+                geometry = (GeometryInstance)organelleModel.CurrentlyShown;
+            }
+            else
+            {
+                geometry = organelleModel.CurrentlyShown.GetNode<GeometryInstance>(definition.DisplaySceneModelPath);
+            }
+
+            geometry.MaterialOverride.RenderPriority = renderPriority;
+        }
```
Kasterisk, line 2325: Unless it's inconvenient, I think this can be made into a helper method in SceneDisplayer and CurrentlyShown can be kept as private.

Kasterisk (changes requested):

src/microbe_stage/editor/MicrobeEditor.cs
hhyyrylainen: I think this DisplaySceneModelPath handling is in pretty many places, so unifying those to a single helper method would also be nice.

Kasterisk (approved):

src/engine/NodeHelpers.cs
```diff
@@ -115,4 +115,27 @@ public static void ReParent(this Node node, Node newParent)
...
+    /// <param name="modelPath">Path to model within the scene. If null takes scene root as model.</param>
+    /// <returns>ShaderMaterial or null if not found.</returns>
+    public static ShaderMaterial GetMaterial(this Node node, string modelPath = null)
+    {
+        GeometryInstance geometry;
+
+        // Fetch the actual model from the scene
+        if (string.IsNullOrEmpty(modelPath))
+        {
+            geometry = node as GeometryInstance;
```
hhyyrylainen, line 132: I think this should use a throwing cast to make it more obvious if you call this method on the wrong type of object.

src/general/SceneDisplayer.cs
```diff
@@ -21,6 +21,16 @@ public string Scene
         }
     }
 
+    /// <summary>
+    ///   Get the material of this scenes model.
```
hhyyrylainen, line 25:
```suggestion
    ///   Get the material of this scene's model.
```

src/engine/NodeHelpers.cs
```diff
@@ -115,4 +115,27 @@ public static void ReParent(this Node node, Node newParent)
...
+        }
+        else
+        {
+            geometry = node.GetNode<GeometryInstance>(modelPath);
+        }
+
+        return geometry?.MaterialOverride as ShaderMaterial;
```
hhyyrylainen, line 139: Also this cast should use the throwing variant to avoid silent errors where a wrong type of object is here.

src/microbe_stage/PlacedOrganelle.cs
```diff
@@ -607,6 +595,11 @@ private void SetupOrganelleGraphics()
         // Position the intermediate node relative to origin of cell
         var transform = new Transform(Quat.Identity,
             Hex.AxialToCartesian(Position) + Definition.CalculateModelOffset());
+        if (organelleMaterial != null)
+        {
+            organelleMaterial.RenderPriority = Hex.GetRenderPriority(Position);
+        }
+
```
hhyyrylainen, line 602: I'd put this after line 585, instead of here in *between* the transform setting.

hhyyrylainen (changes requested):
Code looks almost perfect now, good job on simplifying the render priority calculation logic. I can give this one quick final playtest once the code changes are done, seeing as multiple people already confirmed the graphics worked.

hhyyrylainen (changes requested):
Just that my one comment about UpdateOrganellePlaceHolderScene needs to be addressed, I'm fine enough with the other ones as long as there isn't any organelle that causes a crash now in the game.

takko:
tested with all structure parts

src/engine/NodeHelpers.cs
```diff
@@ -115,4 +115,27 @@ public static void ReParent(this Node node, Node newParent)
...
         newParent.AddChild(node);
     }
+
+    /// <summary>
+    ///   Get the material of this scene's model.
+    /// </summary>
+    /// <param name="node">Node to get material from.</param>
+    /// <param name="modelPath">Path to model within the scene. If null takes scene root as model.</param>
+    /// <returns>ShaderMaterial of the GeometryInstance.</returns>
+    public static ShaderMaterial GetMaterial(this Node node, string modelPath = null)
```
FuJa0815, line 125: I'd use godot's `NodePath` datatype here, as well as in `SceneDisplayer.GetMaterial`

src/general/MathUtils.cs
```diff
@@ -74,4 +74,14 @@ public static float Lerp(float from, float to, float weight, float tolerance = M
...
     }
+
+    /// <summary>
+    ///   Standard modulo for negative values in C# produces negative results.
+    ///   This function returns modulo values between 0 and mod-1.
+    /// </summary>
+    /// <returns>The positive modulo</returns>
+    public static int PositiveModulo(this int val, int mod)
+    {
+        return (val % mod + mod) % mod;
```
FuJa0815, line 85: What is this supposed to do? From my experiments `5.PositiveModulo(3)` returns 2, makes sense, but `(-5).PositiveModulo(3)` returns 1?

src/general/SceneDisplayer.cs
```diff
@@ -21,6 +21,16 @@ public string Scene
         }
     }
 
+    /// <summary>
+    ///   Get the material of this scene's model.
+    /// </summary>
+    /// <param name="modelPath">Path to model within the scene. If null takes scene root as model.</param>
+    /// <returns>ShaderMaterial of the GeometryInstance. Null if no scene.</returns>
+    public ShaderMaterial GetMaterial(string modelPath = null)
```
FuJa0815, line 29: See comment above

FuJa0815 (changes requested):
Please refrain from the use of brackets in the branch name :D

src/general/MathUtils.cs
takko: This is the normal mathematical modulo(at least I think I learned that in school). For negative numbers add the mod until it is in range [0;mod). For positive numbers subtract the mod until it is in range [0;mod)
[link](https://torstencurdt.com/tech/posts/modulo-of-negative-numbers/) e.g. Rust would return positive modulo for %.

takko:
> Please refrain from the use of brackets in the branch name :D
Sure, does the checkout break for some systems?

FuJa0815:
> 
> 
> > Please refrain from the use of brackets in the branch name :D
> 
> Sure, does the checkout break for some systems?

You have to add brackets on bash and it's just not good practice

src/general/MathUtils.cs
FuJa0815: This article also shows this code:
```
int mod(a, b) {
  c = a % b
  return (c < 0) ? c + b : c
}
```

I think this is way readable and should be more efficient because modulo gets used only once

takko:
> > > Please refrain from the use of brackets in the branch name :D
> > 
> > 
> > Sure, does the checkout break for some systems?
> 
> You have to add quotes on bash and it's just not good practice

In my office all use the same git client and it does handle the brackets without complaining. Normally I just paste the issue text as branch name and it replaces all not allowed chars with underscore.
So is there a list which chars are allowed? Or should I just replace all special characters?

hhyyrylainen:
I think all special characters should be not used, so just branch names with alphanumeric characters, underscores and dashes, would be the best.

hhyyrylainen (approved):
Code looks good now and functionality seems to be still working.

PR #2758, opened 2021-11-16
Assortment of tweaks, polishes and fixes to the gui

Kasterisk:

- Darken the autoevo results panel and added fades at the top and bottom to spice up the look as done in the gui mockup.
- Decreased the x axis ticks of the patch details graphs to lessen crowding in tight spaces.

- Made markers in graphs also be highlighted when hovering over the lines.
- Changed the font size in some buttons in the microbe editor.
- Moved cell editor's left panel down a bit to match with the vertical height of other panels in the other tabs.
- Fixed MP bar shown as empty the first time switching to the cell editor tab.
- Changed how the organelle tooltips are positioned. They are now appears at the bottom right edge of the part selection controls instead of following the mouse position. This makes the tooltips more steadier and I honestly think this feels much better.

https://user-images.githubusercontent.com/54026083/142030392-89b48f0e-f086-4b56-848e-1bb806f9a474.mp4

- The species name textbox now flash if the player attempts to enter invalid name.
- Removed pressed animation of the graph dataset button.
- Increased the population label width in the microbe HUD, should fix a bug where the box doubles in height.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
I think the bigger font size actually looks better, at least I was immediately hit with the impression in the reduced font size screenshot that the text is small.

Kasterisk:
I think it definitely looks better in some areas, it's not really big of a difference actually though may just need some time to get used to. It's still readable but more compact, for example in the options menu:

In master:

I think it helps the UI to not feel "bloated"(?) for a lack of better word. Would also like other people's opinions on this.

hhyyrylainen:
To me it looks like that our already large buttons are wasting space as the top of the button looks to be just pointless blank space with the smaller font size.

hhyyrylainen:
In case other people don't chime in, would you be willing to remove the font size changes from this PR? I can then approve this (relatively) quickly. The font size change can be put in another PR waiting for example us to hold a poll on the community forums / discord after the next release so that people see how the current font size looks in game with the new font version.

Kasterisk:
Moved it to another PR.

hhyyrylainen (changes requested):
I noticed too many small isues to merge this right away:
- the tutorial text size is now off. I specifically crafted each English tutorial to nicely fill the box, but now with the smaller font they have bad looking empty blank space
- the jura deleted font file should be kept, because there's the same size font for Lato, so for consistency it should be kept
- The cytoplasm tooltip seems to be the only one that isn't bottom aligned in the microbe editor

Other than those I didn't notice any issues, but I also didn't test saving and loading.

Kasterisk:
Addressed those two.

> The cytoplasm tooltip seems to be the only one that isn't bottom aligned in the microbe editor

Looks fine on my screen? If anything it could be that for some reason in your end the tooltip is offset upwards due to its height goes slightly beyond the bottom of the screen (so it doesn't get cut off) like some of the other tooltips.

hhyyrylainen:
This is what I see:


Unless I misunderstood. I took what you said to mean that the tooltips would be aligned to the bottom of the *screen*.

Kasterisk:
No, I meant the tooltips would be aligned to the bottom right edge of a Control which in this case is the organelle buttons or rigidity slider.

hhyyrylainen:
Okay, I get it now. I misunderstood what you meant by bottom aligning the tooltips.

hhyyrylainen (approved):

PR #2762, opened 2021-11-17
Set auto-evo to again make copy of player on editor

adQuid:

After some playtesting, it was clear that it's just too easy to completely outrun the other species, which still take a lot of rounds to propagate around the world. In the future we should solve this by making the player migrate slower and the AI migrate faster, but in the short term this is the only practical way to make a fun 0.5.6

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/microbe_stage/editor/MicrobeEditor.cs
+++ b/src/microbe_stage/editor/MicrobeEditor.cs
@@ -1491,7 +1491,7 @@ public class MicrobeEditor : NodeWithInput, ILoadableGameState, IGodotEarlyNodeR
 
         // Disabled warning as this is a tweak constant
         // ReSharper disable ConditionIsAlwaysTrueOrFalse HeuristicUnreachableCode
-        if (Constants.CREATE_COPY_OF_EDITED_SPECIES && new Random().NextDouble() > 0.5)
+        if (new Random().NextDouble() < Constants.CREATE_COPY_OF_EDITED_SPECIES_CHANCE)
         {
             // Create a mutated version of the current species code to compete against the player
             CreateMutatedSpeciesCopy(species);
--- a/src/microbe_stage/editor/MicrobeEditor.cs
+++ b/src/microbe_stage/editor/MicrobeEditor.cs
@@ -1490,7 +1490,6 @@ public class MicrobeEditor : NodeWithInput, ILoadableGameState, IGodotEarlyNodeR
 #pragma warning disable 162
 
         // Disabled warning as this is a tweak constant
-        // ReSharper disable ConditionIsAlwaysTrueOrFalse HeuristicUnreachableCode
         if (new Random().NextDouble() < Constants.CREATE_COPY_OF_EDITED_SPECIES_CHANCE)
         {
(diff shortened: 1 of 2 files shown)


hhyyrylainen:
Code-wise there isn't much to say.
Needs playtesting / design team opinions.

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -1491,7 +1491,7 @@ private void SetupEditedSpecies(MicrobeSpecies species)
 
         // Disabled warning as this is a tweak constant
         // ReSharper disable ConditionIsAlwaysTrueOrFalse HeuristicUnreachableCode
-        if (Constants.CREATE_COPY_OF_EDITED_SPECIES)
+        if (Constants.CREATE_COPY_OF_EDITED_SPECIES && new Random().NextDouble() > 0.5)
```
Maxonovien, line 1494: Probability probably should be a Constant as we may want to tweak in the future, shouldn't it?

adQuid: Yeah, that makes a lot of sense.

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -1491,7 +1491,7 @@ private void SetupEditedSpecies(MicrobeSpecies species)
 
         // Disabled warning as this is a tweak constant
         // ReSharper disable ConditionIsAlwaysTrueOrFalse HeuristicUnreachableCode
```
FuJa0815, line 1493: Outdated

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -1491,7 +1491,7 @@ private void SetupEditedSpecies(MicrobeSpecies species)
...
-        if (Constants.CREATE_COPY_OF_EDITED_SPECIES)
+        if (new Random().NextDouble() < Constants.CREATE_COPY_OF_EDITED_SPECIES_CHANCE)
```
FuJa0815, line 1494: Creating a new random here sounds wrong to me. This gets called multiple times at roughly the same time, right? The implementation of random bases it's seed upon the current time. If this gets called twice at the same time the random number will be the same.
Creating the random as an object variable fixes this.

FuJa0815 (changes requested):

src/microbe_stage/editor/MicrobeEditor.cs
hhyyrylainen: I'm pretty sure this gets called exactly once like 5 minutes (only once per editor object creation), so the current time as randomness seed is probably sufficient here.

adQuid: It's currently being called exactly twice each time the editor opens, since the helper method below also creates a random. I could combine the two, but I tend to shy away from passing down helper objects like that. I'm happy to code it either way.

adQuid:
Found a better way to solve this problem

Issue #2751, opened 2021-11-12
Fix the custom dialog properties almost always breaking automatically


hhyyrylainen:
Lately with the new custom dialog classes I've noticed that Godot almost always automatically breaks the main scene like this:
![kuva](https://user-images.githubusercontent.com/3796411/141430329-1fb749d0-aa86-454b-aa17-c58c2d8d6340.png)

Might be an engine bug related to the compiling of C# code on load or something causing the properties to not be ready by the time the main scene is loaded.

hhyyrylainen:
This is really starting to annoy me. Perhaps if we moved that dialog box to a child scene / class, then it would work as the first loaded scene wouldn't have that kind of custom property in it?.

hhyyrylainen:
Found another victim of this bug:
![2021-11-21_11 37 19 1581](https://user-images.githubusercontent.com/3796411/142756987-c54b0a56-bb76-423a-9b5d-4c6e891be02a.png)


I think we need an automatic check against this as this seems to happen super often.

I'll go through all the tutorial scenes on my modding branch and fix the titles.

hhyyrylainen:
What if the "tool" attributes are commented out? Maybe that would help. It would 100% be better to not have them display properly in the editor than to constantly, I mean yesterday I probably re-set a window's title like 10 times, fight against this.

Kasterisk:
That's probably for the best, and if live visual feedback is needed the tool attribute can just be uncommented locally.

hhyyrylainen:
With the update to Godot 4 this is hopefully possible to do now.

PR #2763, opened 2021-11-17
Turn licenses display into a popup dialog

Kasterisk:

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
This needs to be updated for the latest master version. Also note for me: I need to test that in the Steam version the use of [Tool] won't crash the Godot editor due to the Steam load check there.

What if we add a menu that is so large that it is visible from underneath the dialog?

Also just a general note but https://github.com/Revolutionary-Games/Thrive/issues/2751 might get worse with this change...

Kasterisk:
> What if we add a menu that is so large that it is visible from underneath the dialog?

What do you mean by this?

hhyyrylainen:
It looks like you removed the menu content hiding when showing the popup. So if the popup is small or there are a lot of menu buttons the menu buttons will be visible from behind the dialog box.

Kasterisk:
Ah in that case I have already set the popup as exclusive so even if there are menu buttons that are visible the player shouldn't be able to click them.

hhyyrylainen:
They would still be a visual distraction and different from all the other separate sub menus available from the main menu.

Kasterisk:
I've now set the popup to be the same size as the viewport, should address the issue.

hhyyrylainen:
Ah okay.

hhyyrylainen (changes requested):
I don't like how the popup is the entire screen size now. Much better when it wasn't sized so huge. What I wanted was that the main menu would be hidden when the popup is open (like in my previous code managing that, that this PR removes).

hhyyrylainen (approved):
This looks and works well now I think.

PR #2782, opened 2021-11-19
Add support for playing a track list only once

### LINKED ISSUE #2698: Extinction theme is played on loop
Found that on game lose the extinction theme is still actually looping despite its loop property is disabled in the import setting. Underlying cause most likely is in the Jukebox class itself. Need a way to play a list of track without repeating.

Kasterisk:

Track list can now be non-repeating. Not really confident with the implementation so feel free to suggest a better approach.

**Related Issues**

Closes #2698 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/general/Jukebox.cs
```diff
@@ -538,6 +543,8 @@ private void OnCategoryEnded()
             {
                 foreach (var track in list.Tracks)
                 {
+                    track.PlayedOnce = false;
```
hhyyrylainen, line 546: I found a bug regarding this (this reset needs to *always* happen when `category` is not null.
If you go into a new game and go extinct, the track plays fine once, but if you exit to menu and then start a new game where you go extinct again, you don't hear the extinct track again even though you should.

hhyyrylainen (changes requested):

hhyyrylainen (changes requested):
Still does not work:
```
Jukebox now playing from: Extinction
Jukebox: starting track: res://assets/sounds/extinct2.ogg position: 63.46594
```
Did it work for you?

Kasterisk:
Oh do you mean the track position should reset instead of continuing? From your jukebox log it seems the track is resumed at nearing the end which has quite a long silence in the audio stream but not yet ended, so that may give the impression the track never plays once it does.

FuJa0815:
Couldn't you add a property like `NeedsToBeginFromStart`?

Kasterisk:
The actual solution would be adding a new member in MusicCategory.ReturnType enum to not continue.

hhyyrylainen:
I guess I didn't implement that? I guess there wasn't a music category before that didn't continue? Isn't the main menu category like that, or does that continue as well?

Kasterisk:
Nope, there is only one return type that is to continue. All categories currently continues including the main menu.

hhyyrylainen (approved):

PR #2790, opened 2021-11-21
Make auto-evo results translatable

### LINKED ISSUE #2785: Auto-evo results text doesn't react to language change
image

takko:

Added the possibility to save a string with Translations.
For this I added TranslatingStringBuilder and TranslatingString classes.
TranslatingString takes the translationKey and optional stringFormat args.
ToString returns the translated text for the current locale.

The TranslatingStringBuilder generates a FormatString with {0} for IFormattable types and saves the IFormattable objects in a list.
ToString calls string.Format which applies all translations.

Everything is Json serializable.
To load old saves, the TranslatingStringBuilder has an explicit type cast from string. The texts for old saves do not translate.

I'm not sure if TranslatingString is a good name, but didn't find a better one.

Also added TranslatingString to the pybabel keywords.
I could not add a function AppendTranslation(translationKey, args) to TranslatingStringBuilder, as pybabel does not accept wildcards like *.AppendTranslation

**Related Issues**

closes #2785

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


84634E1A607A (changes requested):
**Testing result:**
1. Code functionality confirmed. Code works as intended.
2. Save / Load OK
3. Save upgrade OK
4. Due to this PR doesn't increase version number (Such as from 0.5.6.0-alpha to 0.5.6.1-alpha), I can save using this PR then open using master, generating load error; and loading a master save doesn't shows upgrade confirmation.

takko:
Updated the AssemblyVersion. When loading a master save there is that dialog now. When loading a branch save from master there is no dialog, but it fails, as it is not compatible in this way.

84634E1A607A:
> When loading a branch save from master there is no dialog

There is.

image

84634E1A607A (approved):

hhyyrylainen:
Seeing as the version number was incremented to 0.5.6.1 I guess you don't mind this being merged after 0.5.6 release. And I think we'll probably anyway have a 0.5.6.1 patch release this time around as I think there'll be all kinds of small mod etc. issues that are good to take care of in a week or two after the release. (another solution would have been to increment the version to 0.5.6.0 with no suffix as that is treated as a different version from 0.5.6.0-alpha and even 0.5.6.0-rc1, which reminds me that I need to add save upgraders for those in my branch).

RacerBG:
Will this PR fix every instance of non-translatable strings or just the auto-evo ones?

This picture made by Kasterix, shows that "Pangonian Vents" and "N/A" are not updated when the language is changed as well:
142730979-9e7ec734-0228-462d-ba9c-4a427db8635c

takko:
Just added the translations for the auto evo prediction to update on language change.

hhyyrylainen:
> Just added the translations for the auto evo prediction to update on language change.

I think that's better as a separate PR, because this can't be merged before 0.5.6 release as this bumps the version number.

Thrive.csproj
```diff
@@ -34,6 +34,8 @@
     <Compile Include="src\engine\CheatManager.cs" />
     <Compile Include="src\engine\ChildObjectCache.cs" />
     <Compile Include="src\engine\ChromaticFilter.cs" />
+    <Compile Include="src\engine\TranslatingString.cs" />
+    <Compile Include="src\engine\TranslatingStringBuilder.cs" />
```
Kasterisk, line 38: How about naming these `LocalizedString` or `LocalizationString`?

src/engine/TranslatingString.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Globalization;
+using Godot;
+using Newtonsoft.Json;
+
+[JSONDynamicTypeAllowed]
+public class TranslatingString : IFormattable
```
Kasterisk, line 7: Add a doc here explaining how the class can be used and its purpose.

src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -320,6 +320,21 @@ public class MicrobeEditorGUI : Control, ISaveLoadedTracked
     [JsonProperty]
     private float initialCellHp;
 
+    [JsonProperty]
+    private bool? autoEvoRunSuccessfull;
```
Kasterisk, line 324:
```suggestion
    private bool? autoEvoRunSuccessful;
```

Kasterisk (changes requested):

RacerBG:
Alright, everything seems to work just fine, good work. The log gives errors which are not related to this PR. I tried the following things:
1. Changing the languages back and forward between English and Bulgarian. Everything was working as intended.
2. Did the same thing with the Patch Map and the Editor - no problems noticed.
3. I had played the game for a few generations - no crashes or at least no new ones.

Thank you for creating this PR.

Kasterisk (approved):
LGTM.

src/auto-evo/RunResults.cs
```diff
@@ -389,27 +387,26 @@ public void PrintSummary(PatchMap previousPopulations = null)
...
             bool playerReadable = false, List<ExternalEffect> effects = null)
         {
             const bool resolveMigrations = true;
             const bool resolveSplits = true;
 
-            var builder = new StringBuilder(500);
+            var builder = new LocalizedStringBuilder(500);
 
-            string PatchString(Patch patch)
+            LocalizedStringBuilder PatchString(Patch patch)
```
hhyyrylainen, line 398: I think this should be refactored so that this is given in the LocalizedStringBuilder where the patch data is added to. This would reduce this nested LocalizedStringBuild stuff, which I kind of dislike as to me it seems we might end up with very deep object hierarchies unnecessarily.

src/engine/LocalizedString.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        return ToString(null, null);
+    }
+
+    public string ToString(string format, IFormatProvider formatProvider)
+    {
+        return formatStringArgs == null || formatStringArgs.Length == 0 ?
+            format ?? TranslationServer.Translate(translationKey) :
+            string.Format(formatProvider ?? CultureInfo.CurrentCulture,
+                format ?? TranslationServer.Translate(translationKey), formatStringArgs);
```
hhyyrylainen, line 42: I think this is way too long of a single statement to be able to understand what is happening here.

hhyyrylainen (changes requested):
Functionality and save compatibility seem to work fine. I have a few code related comments through. I realize that the dynamic type is probably related to the save compatibility? If so I think we should open an issue to eventually maybe in like 0.5.8 or so to change it?
Alternatively I'd like to see a save upgrader to 0.5.6.1-alpha as the save load menu will give an outdated save warning anyway. And eventually a save upgrade from 0.5.6 to like 0.5.8 would fail even if we have version upgraders for all intermediate versions.

src/engine/LocalizedString.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Globalization;
+using Godot;
+using Newtonsoft.Json;
+
+/// <summary>
+///   This class caches the translationKey.
+///   ToString returns the translated text for the current locale.
+///   This class can be used on its own, but was designed for the use within LocalizedStringBuilder.
+/// </summary>
+[JSONDynamicTypeAllowed]
```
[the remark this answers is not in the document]
takko: Yes for the saves.
I got the ```        throw new JsonException($"Dynamically typed JSON object is not allowed to be {typeName}");``` exception because the LocalizedString has an object[] as JsonProperty. So I added this Attribute.

src/engine/LocalizedStringBuilder.cs
```diff
@@ -0,0 +1,110 @@
...
+using Newtonsoft.Json;
+
+/// <summary>
+///   Like a StringBuilder but with localization.
+///   When adding a IFormattable(e.g. Number, or LocalizedString),
+///   the object is cached in a list and a string format index ({0}) is added to the StringBuilder.
+///   Calling ToString formats all the IFormattable's to the current culture and returns the formatted string.
+///   This class is json serializable as long as all IFormattable's are serializable.
+/// </summary>
+[JSONDynamicTypeAllowed]
```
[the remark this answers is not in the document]
takko: I did not fully understand the serialization with the thrive serializer.
I just wanted the LocalizedStringBuilder and the LolizedString to be serializable. Not sure if this is the correct way.
What does dynamically mean. There is the List<IFormattable> which is dynamic as it is just an interface.
But I just want this class to get serialized. If there is another way, I'll take it.

src/engine/LocalizedString.cs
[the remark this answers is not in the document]
takko: So the dynamic type is added to make serialization possible. The save compatibility is done via the implicit conversion from string.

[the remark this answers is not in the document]
hhyyrylainen: Oh yeah, I see that now. I guess there isn't a way to avoid that? I ask because the dynamic types should be avoided if possible as they leave more exploit surface for malicious save files to try to exploit.

src/engine/LocalizedStringBuilder.cs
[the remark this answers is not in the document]
hhyyrylainen: I would imagine that if you remove the nexted LocalizedStringBuilder usage, you would be able to just remove this line or if that doesn't work replace with `UseThriveSerializerAttribute`

src/auto-evo/RunResults.cs
takko: I dislike this too, but this stringbuilder is used in another formatString
```csharp
builder.Append(new LocalizedString("WENT_EXTINCT_IN", patchName));
```
So this nesting is needed

hhyyrylainen: I guess it can't be avoided, then. At least a good design that takes that into account and avoid the nesting isn't immediately popping into my mind.

hhyyrylainen (approved):
I think my comments were sufficiently addressed and I don't think the changes could have broken this, so I'll merge this now.

PR #1930, opened 2020-12-15
 Cell Movement pitch scales with microbe speed

icedjuro:
closes #1610

PR #2793, opened 2021-11-21
Cell movement sound handling overhaul

Kasterisk:

An alternative to (or complements) #1930.

Overhauled the way how the microbe movement sound is played. Currently the audio is played continuously every time a cell moves and with the current sound effect we have this can probably get to some people's nerves.

Also I doubt the sound effects are intended to be played constantly in the first place as the sound structure doesn't seem to indicate that, rather it kind of implies to be played on a cell turning?

So in this PR I introduced a few changes which includes: movement audio cooldown and "acceleration-timed" audio play. The cooldown adds a certain delay to when a move sound is played between each other, and the other makes the sound only be emitted when there's a considerable change in the cell's acceleration. In effect, this causes the move sound to be only emitted once when starting to move ~~and when the cell turns or any acceleration change in general~~ (see https://github.com/Revolutionary-Games/Thrive/pull/2793#issuecomment-993267103). This may not always happen given the previous mentioned cooldown modifier.

Besides those, I added one more move sound to add variety and to make good use of some existing sound assets that are just gathering dust. Both sounds will be played randomly.

All in all, I think this direct the movement sound to the way how I assume it is intended to be played. Though other opinions/suggestions are welcome.

In the code side, the movement handling logic is moved entirely to `_PhysicsProcess`. This is needed to get the acceleration calculation right as doing it in `_Process` gives some weird results. On playtest this change doesn't seem to affect how the cell movement works.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/microbe_stage/Microbe.Interior.cs
```diff
@@ -736,6 +736,40 @@ private void HandleOsmoregulation(float delta)
...
 
+    private void HandleMovement(float delta)
+    {
+        if (MovementDirection != new Vector3(0, 0, 0) ||
+            queuedMovementForce != new Vector3(0, 0, 0))
+        {
+            // Movement direction should not be normalized to allow different speeds
+            Vector3 totalMovement = new Vector3(0, 0, 0);
+
+            if (MovementDirection != new Vector3(0, 0, 0))
```
FuJa0815, line 747: Use Vector3.Zero

FuJa0815 (changes requested):

src/microbe_stage/Microbe.Interior.cs
Kasterisk: Changed.

RacerBG:
Could someone merge the latest changes from master first?

Kasterisk:
Merged it.

Kasterisk:
Nice to hear that, will try implementing your suggestion. Btw I noticed a minor issue with this PR, colliding with chunks or objects triggers the movement sound where it should've not. _Maybe_ I could cheat this by checking if there's any collision with the `collisionForce` variable in #2847.

buckly90:
Something seems to very wrong with this.

The movement sound often doesn't play when it's supposed to. And for some reason it will occasionally break and start looping at an incredibly high pitch.

Hate to be the bearer of bad news...

RacerBG:
> Something seems to be very wrong with this.
> 
> The movement sound often doesn't play when it's supposed to. And for some reason it will occasionally break and start looping at an incredibly high pitch even after I stop moving.
> 
> Hate to be the bearer of bad news...

Isn't this intended or I misunderstood the description:

> The cooldown adds a certain delay to when a move sound is played between each other, and the other makes the sound only be emitted when there's a considerable change in the cell's acceleration.

Kasterisk:
@buckly90 Assuming you're running Thrive through Godot, there might be incorrect settings in your movement sound assets, maybe the Loop property is enabled. No need to worry though, this can happen automatically in a freshly cloned repo.

buckly90:
> @buckly90 Assuming you're running Thrive through Godot, there might be incorrect settings in your movement sound assets, maybe the Loop property is enabled. No need to worry though, this can happen automatically in a freshly cloned repo.
> 
> > Isn't this intended or I misunderstood the description:
> 
> No the problem seems to be the sound starts to loop continuously which is not intended.

I ran it through the Thrive launcher.

I went ahead and gave this a try again, and was unable to trigger the looping audio glitch so that's good! Nice job.

However, the movement audio only plays once when I first begin to move, and then doesn't seem to audibly play as I am moving. Is this intended? I tried playing with different speeds but the end result remained the same.

Kasterisk:
Yes that's the main goal of this PR... the movement sound should only play when starting to move.

RacerBG (approved):

hhyyrylainen (approved):
Overall I'm a bit hesitant about the change to applying the movement force processing in the _PhysicsProcess callback, though I suppose it should still match the ATP drain etc., but it might not fully match the previous behaviour with trying to move without ATP...
Overall I'm fine enough with this change as long as other people test that ATP damage from movement and are happy with how this sounds.

Kasterisk:
Playing in 60 fps, the behavior of atp damage from movement is still largely the same as far I can tell. Re the sound, I think it would be great if we have two types of sound for movement where one is for when starting to move and the other is a movement looping sound, this needs to be subtle enough to not annoy, possibly a low profile seamless sloshing noise. But this is for another time.

Kasterisk:
Merging due to approvals, can probably reverted if this turns out to be problematic.

PR #2795, opened 2021-11-21
More volume tweaks

Kasterisk:

Tweaked some more quite overly loud sounds and also added a method to play non positional audio players in Microbe.cs.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/microbe_stage/Microbe.Contact.cs
+++ b/src/microbe_stage/Microbe.Contact.cs
@@ -495,7 +495,7 @@ public partial class Microbe
         if (IsPlayerMicrobe)
         {
             // Playing from a positional audio player won't have any effect since the listener is
-            // directly in it.
+            // directly on it.
             PlayNonPositionalSoundEffect("res://assets/sounds/soundeffects/microbe-death-2.ogg", 0.3f);
         }
         else
--- a/src/microbe_stage/Microbe.Contact.cs
+++ b/src/microbe_stage/Microbe.Contact.cs
@@ -496,7 +496,7 @@ public partial class Microbe
         {
             // Playing from a positional audio player won't have any effect since the listener is
             // directly on it.
-            PlayNonPositionalSoundEffect("res://assets/sounds/soundeffects/microbe-death-2.ogg", 0.3f);
+            PlayNonPositionalSoundEffect("res://assets/sounds/soundeffects/microbe-death-2.ogg", 0.5f);
         }
         else
         {
--- a/src/microbe_stage/Microbe.Contact.cs
+++ b/src/microbe_stage/Microbe.Contact.cs
@@ -492,7 +492,16 @@ public partial class Microbe
             OnReproductionStatus?.Invoke(this, false);
         }
 
-        PlaySoundEffect("res://assets/sounds/soundeffects/microbe-death-2.ogg");
+        if (IsPlayerMicrobe)
+        {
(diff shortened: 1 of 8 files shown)


hhyyrylainen:
Could you maybe boost the main menu songs up to like 0.3 or 0.4 as they are way quieter than the ingame music, I think it should be at least as loud as them...

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -492,7 +492,16 @@ public void Kill()
             OnReproductionStatus?.Invoke(this, false);
         }
 
-        PlaySoundEffect("res://assets/sounds/soundeffects/microbe-death-2.ogg");
+        if (IsPlayerMicrobe)
+        {
+            // Playing from a positional audio player won't have any effect since the listener is
+            // directly in it.
```
FuJa0815, line 498:
```suggestion
            // directly on it.
```

buckly90:
Great work on these tweaks, the ambient volume seems perfect, along with most other things.

My only complaint is that the sfx for clicking buttons and dying feels just barely too quiet
If those are turned up just a little, I think it would be perfect. (I tested this with all sounds at max volume)

src/microbe_stage/Microbe.cs
```diff
@@ -329,6 +330,32 @@ public void PlaySoundEffect(string effect)
...
+        // TODO: make these sound objects only be loaded once
+        var sound = GD.Load<AudioStream>(effect);
+
+        // Find a player not in use or create a new one if none are available.
+        var player = nonPositionalAudioPlayers.Find(nextPlayer => !nextPlayer.Playing);
+
+        if (player == null)
+        {
+            // If we hit the player limit just return and ignore the sound.
+            if (otherAudioPlayers.Count >= Constants.MAX_CONCURRENT_SOUNDS_PER_ENTITY)
```
hhyyrylainen, line 344: Wrong variable used here

hhyyrylainen (changes requested):
If the sound levels are fine by buckly, I can accept that as well for this release. Noticed one wrong variable used in the code.

hhyyrylainen:
Oh and one more thing. What about the intro video audio levels? I just played on my laptop where I have those enabled and I think at least the microbe intro volume level was much higher than the menu or in game music now.
I sort of feel like that reducing the overall master volume may have been the right call with a few music tracks getting a boost in volume instead of needing to reduce everything one by one, but at least for 0.5.6 I don't think the approach should be switched at this point.

Kasterisk:
I didn't change the in-game music much so I guess the intro cutscene was just a bit loud to begin with, it can be easily tuned as VideoPlayer provides VolumeDb property. The problem with reducing the overall master volume approach is that it doesn't preserve consistency, by that I mean some sounds will be loud and the user have to lower it and some will end up too quiet and they have to increase it.

A good example is probably the bubble ambience we have right now, it's pretty jarring in contrast to its other soft sounding counterpart and the player will then have to alternate between different volume settings to be able listen to varieties of ambience sounds comfortably.

hhyyrylainen:
Well I tested again on my laptop with the latest master branch code (the Steam build) and it's definitely the case that the game start intro video is noticeable louder than the main menu music.

hhyyrylainen (approved):
Code looks fine now, I'll merge this now assuming the gameplay has *not* broken due to the latest code changes.

PR #2847, opened 2021-11-25
Implemented microbe collision sound effect

Kasterisk:

Adds a collision sound effect that plays when a certain collision force is reached between microbes.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


FuJa0815:
There isn't any sound when bumping into iron chunks

src/microbe_stage/Microbe.cs
```diff
@@ -531,6 +533,13 @@ public override void _IntegrateForces(PhysicsDirectBodyState physicsState)
         // TODO: should movement also be applied here?
 
         physicsState.Transform = GetNewPhysicsRotation(physicsState.Transform);
+
+        collisionForce = Vector3.Zero;
+
+        for (var i = 0; i < physicsState.GetContactCount(); ++i)
+        {
+            collisionForce += physicsState.GetContactImpulse(i) * physicsState.GetContactLocalNormal(i);
```
adQuid, line 541: Does this mean that, if a microbe impacts two objects on opposite sides at the same time (like squeezing through a gap), that it will have near zero collision force?

Kasterisk: Hmm, I actually haven't thought about this. I will look into this later when I have more time.

Kasterisk:
> There isn't any sound when bumping into iron chunks

I originally intended this, looked through some really old implementation and the sound only plays on collision between microbes though now that I think about it, it might just because floating chunks haven't yet added back then...

Nonetheless would be really nice to have a variety of collision sounds going this route.

src/microbe_stage/Microbe.cs
hhyyrylainen: I would imagine that results in two different contacts as they are different objects that hit the one in the center.

hhyyrylainen: And also at least in bullet two objects colliding can result in multiple contact points depending on their shapes.

hhyyrylainen: Ah, I see these are summed here. Maybe the squared lengths should already be retrieved here so that this problem won't be able to happen.

RacerBG (commented):
Probably I'm doing something wrong here but I can't hear anything new.

First, I had disabled the music, then the environment sounds, then I had tried to disable the effects and leave the environment sounds on but still nothing. I tried with with pilus, without a pilus, with different membranes - still nothing.*

* I tried to collide only with my own cells from which I had divided.

Kasterisk:
Maybe you didn't collide with enough force?

RacerBG:
> Maybe you didn't collide with enough force?

How much is that?

Kasterisk:
I think about quite a distance away from other cell and then you just start ram into it. For easier testing you need to disable the AI first through the cheat menu so the other cell don't move around.

RacerBG:
> I think about quite a distance away from other cell and then you just start ramming into it. For easier testing you need to disable the AI first through the cheat menu so the other cell don't move around.

Its works now but I had to rely heavily on the cheats for that. I would like to see more opinions on the force needed to create this sounds but overall it's a nice addition.

RacerBG (approved):

Kasterisk:
I intended the sound mainly as an auditory feedback for accidental cell collisions which is probably like once in a few minutes or more occurrence (given the player didn't purposefully ram into other cells) so the impact sound shouldn't happen at every cell bump, it probably can get pretty annoying very quick if it were to be so.

RacerBG:
> I intended the sound mainly as an auditory feedback for accidental cell collisions which is probably like once in a few minutes or more occurrence (given the player didn't purposefully ram into other cells) so the impact sound shouldn't happen at every cell bump, it probably can get pretty annoying very quick if it were to be so.

I see, if this is the purpose, then it's fine with me.

hhyyrylainen (changes requested):
Hang on, why is the logic so that on a contact begin, with a microbe, the sound is played if the total forces of *all* collisions on the microbe as detected in the physics always running callback is higher than a thresshold?
Shouldn't it be like the other way around? Or purely so that the microbe contact callback would detect everything?

Kasterisk:
Yeah, the threshold is there to filter out weak collisions so it doesn't play the sound on every tiniest collisions.

hhyyrylainen:
Would it actually be possible to program this correctly, ie. so that the actual physics force of a collision is used to play the sound, rather than relying on there not being more than one collision going on at once for a cell (otherwise the total force does not match the force of a single collision)?

Kasterisk:
Unfortunately, Godot does not provide some convenient way to access a total collision impulse, this is one of the solution I found to calculate it from here https://godotengine.org/qa/13155/get-collision-force?show=94293#a94293. Maybe to ignore the resulting force possibly cancelling each other due to having multiple side of contact would be to remove the second part of the sum calculation?
```diff
-             collisionForceSqr +=
-                (physicsState.GetContactImpulse(i) * physicsState.GetContactLocalNormal(i)).LengthSquared();
+             collisionForce += physicsState.GetContactImpulse(i);
```

hhyyrylainen:
I guess we'll need to go with a workaround then, but a TODO comment and a github issue should be opened about sometime in the future managing to do this properly.

hhyyrylainen (approved):
I guess this is fine to merge now.

PR #2868, opened 2021-11-28
Auto-evo data caching and other speedups

### LINKED ISSUE #2451: Make auto-evo population simulation for mutations only simulate necessary patches
Instead of simulating each patch for each mutation for a species, it would be more efficient to only simulate patches that a species is currently in. So this way each mutation effectiveness calculation could probably simulate only 20% of the patches. This should make auto-evo noticeably faster in the long run once we have more complex calculations, more patches, or more species.

hhyyrylainen:

added caching of energy balances and made migrations and mutations only simulate the patches they need. In my testing the first change cut down the total run time by 10 times and then the second change about halved it, so I think these changes in total made a 20-fold reduction in auto-evo processing time

**Related Issues**

closes #2451

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/auto-evo/simulation/PopulationSimulation.cs
+++ b/src/auto-evo/simulation/PopulationSimulation.cs
@@ -14,10 +14,12 @@
         private static readonly Compound Glucose = SimulationParameters.Instance.GetCompound("glucose");
         private static readonly Compound HydrogenSulfide = SimulationParameters.Instance.GetCompound("hydrogensulfide");
         private static readonly Compound Iron = SimulationParameters.Instance.GetCompound("iron");
+        private static readonly Compound Sunlight = SimulationParameters.Instance.GetCompound("sunlight");
 
         public static void Simulate(SimulationConfiguration parameters)
         {
             var random = new Random();
+            var cache = new SimulationCache();
 
             var speciesToSimulate = CopyInitialPopulationsToResults(parameters);
 
@@ -31,7 +33,7 @@
 
             while (parameters.StepsLeft > 0)
             {
-                RunSimulationStep(parameters, speciesToSimulate, patchesToSimulate, random);
+                RunSimulationStep(parameters, speciesToSimulate, patchesToSimulate, random, cache);
                 --parameters.StepsLeft;
             }
         }
@@ -115,31 +117,31 @@
         }
 
         private static void RunSimulationStep(SimulationConfiguration parameters, List<Species> species,
-            IEnumerable<KeyValuePair<int, Patch>> patchesToSimulate, Random random)
+            IEnumerable<KeyValuePair<int, Patch>> patchesToSimulate, Random random, SimulationCache cache)
         {
             foreach (var entry in patchesToSimulate)
             {
                 // Simulate the species in each patch taking into account the already computed populations
                 SimulatePatchStep(parameters.Results, entry.Value,
-                    species.Where(item => parameters.Results.GetPopulationInPatch(item, entry.Value) > 0).ToList(),
-                    random);
+                    species.Where(item => parameters.Results.GetPopulationInPatch(item, entry.Value) > 0),
+                    random, cache);
             }
         }
 
         /// <summary>
         ///   The heart of the simulation that handles the processed parameters and calculates future populations.
         /// </summary>
-        private static void SimulatePatchStep(RunResults populations, Patch patch, List<Species> genericSpecies,
-            Random random)
+        private static void SimulatePatchStep(RunResults populations, Patch patch, IEnumerable<Species> genericSpecies,
+            Random random, SimulationCache cache)
         {
             _ = random;
 
-            // Skip if there aren't any species in this patch
-            if (genericSpecies.Count < 1)
-                return;
-
             // This algorithm version is for microbe species
             var species = genericSpecies.Select(s => (MicrobeSpecies)s).ToList();
+
+            // Skip if there aren't any species in this patch
+            if (species.Count < 1)
+                return;
 
             var energyBySpecies = new Dictionary<MicrobeSpecies, float>();
             foreach (var currentSpecies in species)
@@ -149,7 +151,7 @@
 
             var niches = new List<FoodSource>
             {
-                new EnvironmentalFoodSource(patch, "sunlight", Constants.AUTO_EVO_SUNLIGHT_ENERGY_AMOUNT),
+                new EnvironmentalFoodSource(patch, Sunlight, Constants.AUTO_EVO_SUNLIGHT_ENERGY_AMOUNT),
                 new CompoundFoodSource(patch, Glucose),
                 new CompoundFoodSource(patch, HydrogenSulfide),
                 new CompoundFoodSource(patch, Iron),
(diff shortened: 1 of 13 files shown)


RacerBG:
While the FPS in-game is unchanged, now literally I can't see the loading screen (in a good way). I used to wait for around 5-6 seconds with ~300 steps (Freebuild mode), now I'm waiting around 1 second - I see a black screen and the Auto-Evo results are showing immediately.

I tested the game for a few minutes - no crashes, so I believe it's fine. I can't comment about the code though.

Very good work!

Maxonovien (commented):
There might be an unnecessary operation somewhere, but else everything's clear, with a few welcomed small code cleaning operations.

Testing-wise the auto-evo time decreased indeed by a decent factor, and loading a save didn't seem to break things.

src/auto-evo/steps/FindBestMigration.cs
```diff
@@ -52,6 +54,11 @@ protected override IAttemptResult TryVariant()
                 return new AttemptResult(null, -1);
 
             var config = new SimulationConfiguration(map, Constants.AUTO_EVO_VARIANT_SIMULATION_STEPS);
+
+            config.SetPatchesToRunBySpeciesPresence(species);
+            config.PatchesToRun.Add(migration.From);
```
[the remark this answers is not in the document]
hhyyrylainen: I *think* so, but I just put this here for safety, adding a duplicate entry to a HashSet won't do anything.

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -115,32 +117,32 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
         }
 
         private static void RunSimulationStep(SimulationConfiguration parameters, List<Species> species,
-            IEnumerable<KeyValuePair<int, Patch>> patchesToSimulate, Random random)
+            IEnumerable<KeyValuePair<int, Patch>> patchesToSimulate, Random random, SimulationCache cache)
```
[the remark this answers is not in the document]
hhyyrylainen: This class creates the object and passes it only to its private member methods, so I thought that a more general name is fine here, and a longer name is better in the interface definition.

src/auto-evo/steps/FindBestMigration.cs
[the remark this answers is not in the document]
Maxonovien: The gain in performance would probably be minimal, so I'm happy with this safety reason.

Maxonovien (approved):
With my only real questioning answered, I'm happy with code changes.

I simply tested it and it seems to work, and I'll also trust RacerBG's more in-depth test.
Nice work!

src/auto-evo/simulation/PopulationSimulation.cs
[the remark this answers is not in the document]
Maxonovien: I assumed it was the reason, just wanted to make sure this was voluntary. I'm fine with it as well.

hhyyrylainen:
I'll merge this in a bit, once we (hopefully) get a conclusion on how to edit the launcher as well to resolve some problems. And then put out a notice of a hotfix on Steam.

Maxonovien:
Indeed. In the mean time I'll try to make a hotfix to limit steps numbers, but that may have more gameplay impact.

RacerBG:
> Indeed. In the mean time I'll try to make a hotfix to limit steps numbers, but that may have more gameplay impact.

I don't think this is huge problem right now. As I wrote earlier, this PR improves the loading times by a lot. Mind you, I was testing this in a VM with 4GB RAM and 4 core CPU. Improving the clouds rendering code, implementing species  and objects limit will be better in my opinion.

Maxonovien:
Whatever we do with performance, there isn't currently any cap on possible requirements for auto evo. Even at 10 steps per second we'll start having issues if we need 5k steps.

RacerBG:
> Whatever we do with performance, there isn't currently any cap on possible requirements for auto evo. Even at 10 steps per second we'll start having issues if we need 5k steps.

I see, thanks for clarifying that. :smiley:

src/auto-evo/simulation/food_source/FoodSource.cs
```diff
@@ -9,11 +11,14 @@
     ///   Provides a fitness metric to determine population adjustments for species in a patch.
     /// </summary>
     /// <param name="microbe">The species to be evaluated.</param>
+    /// <param name="simulationCache">
+    ///   Cache that should be used to reduce amount of times expensive computations are ran
```
adQuid, line 15:
```suggestion
    ///   Cache that should be used to reduce amount of times expensive computations are run
```

FuJa0815 (approved):
no changes of the numbers with much better performance

src/auto-evo/simulation/food_source/MarineSnowFoodSource.cs
```diff
@@ -19,14 +19,15 @@ public MarineSnowFoodSource(Patch patch)
...
         var microbeSpecies = (MicrobeSpecies)species;
 
         var predatorSpeed = microbeSpecies.BaseSpeed;
-        predatorSpeed += ProcessSystem
-            .ComputeEnergyBalance(microbeSpecies.Organelles.Organelles, patch.Biome,
-                microbeSpecies.MembraneType).FinalBalance;
+
+        var energyBalance = simulationCache.GetEnergyBalanceForSpecies(microbeSpecies, patch);
+
+        predatorSpeed += energyBalance.FinalBalance;
```
adQuid, line 30:
```suggestion
        predatorSpeed += energyBalance.FinalBalance;
```

personally I'd group these related statements

src/auto-evo/steps/FindBestMigration.cs
```diff
@@ -36,6 +36,8 @@ protected override IAttemptResult TryCurrentVariant()
         {
             var config = new SimulationConfiguration(map, Constants.AUTO_EVO_VARIANT_SIMULATION_STEPS);
 
+            config.SetPatchesToRunBySpeciesPresence(species);
+
             PopulationSimulation.Simulate(config);
```
adQuid, line 41: Is it safe to say that you should ALWAYS call SetPatchesToRunBySpeciesPresence() prior to running PopulationSimulation.Simulate(config)?

If so, the latter should call the former internally, to prevent misuse.

adQuid (changes requested):
I am surprised, but pleasantly surprised, that this results in such a speed boost. I like the strategy, but I think the cache is dangerously independent. I think more dependency inversion should be introduced to the design to avoid the risk of running computation on a dirty cache.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace AutoEvo
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    ///   Caches some information in auto-evo runs to speed them up
+    /// </summary>
+    public class SimulationCache
+    {
+        private readonly Dictionary<(MicrobeSpecies, Patch), EnergyBalanceInfo> cachedEnergyBalances = new();
```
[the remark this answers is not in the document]
hhyyrylainen: I did not add a note or anything like that *because* I didn't overengineer this, I just added it in the smallest scope it makes sense to use for now.

src/auto-evo/simulation/food_source/MarineSnowFoodSource.cs
hhyyrylainen: I wouldn't because the energy balance is also used a few lines later. If anything this line should be swapped with 26.

src/auto-evo/steps/FindBestMigration.cs
hhyyrylainen: You don't have to. For example when simulating the final populations all species are simulated at once for all patches.

adQuid: Maybe a parameter on the Simulate method then? This is something that would be very easy to miss if a future developer copied off of the time you don't need to use the config in this way as a template for a case where you do have to do this.

hhyyrylainen: They should read the simulation config, this is explained in there.

hhyyrylainen:
I'm going to merge this now so that I don't have to work the entire evening.

PR #2872, opened 2021-11-28
Auto evo limitation

Maxonovien:

This PR aims at limiting the number of species in each patch after auto-evo through extinction events, for performance limitation concerns.

Extinction events do not affect the player and newly spawned species. It can also be set to not affect migrations, but current configuration does kill the migrated cells. The migration is then not reported by the auto-evo, but there is an additional pop loss to the "internal" effect.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


Thrive.csproj
```diff
@@ -19,6 +19,7 @@
     <Reference Include="System" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="src\auto-evo\steps\ForceExtinction.cs" />
```
hhyyrylainen, line 22: Everything else in this file seems to be alphabetical except this line...

src/auto-evo/AutoEvoConfiguration.cs
```diff
@@ -41,6 +41,29 @@ public class AutoEvoConfiguration : IRegistryType
     [JsonProperty]
     public bool BiodiversitySplitIsMutated { get; private set; }
 
+    /// <summary>
+    ///   Maximum number of species kept in a patch at the end of auto-evo.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Always ignore the player species, and may ignore new cells and migrations.
```
hhyyrylainen, line 49:
```suggestion
    ///     Always ignores the player species, and may ignore new cells and migrations.
```

src/auto-evo/RunResults.cs
```diff
@@ -118,6 +118,27 @@ public void AddSplitResultForSpecies(Species species, Species splitSpecies, List
             results[splitSpecies].SplitFrom = species;
         }
 
+        public void KillSpeciesInPatch(Species species, Patch patch, bool refundMigrations = false)
+        {
+            AddPopulationResultForSpecies(species, patch, 0);
+
+            // We copy migration list to be able to modify it
+            foreach (var migration in results[species].SpreadToPatches.ToArray())
```
hhyyrylainen, line 126: For consistency with other code I'd use ToList here. If it turns out that ToArray is better in the usual case, we have a *ton* of code to convert.

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +389,129 @@ public long GetPopulationInPatch(Species species, Patch patch)
             }
         }
 
+        /// <summary>
+        ///   Computes the final population of species, by patch.
+        /// </summary>
+        /// <remarks>
+        ///   <para>
+        ///     TODO: Currently no enforced correlation with actual changes applied.
```
hhyyrylainen, line 397: What's this TODO about?

src/general/DictionaryUtils.cs
```diff
@@ -75,4 +75,49 @@ public static void DivideBy<T>(this Dictionary<T, float> dictionary, float divis
...
     }
+
+    public static T[] GetKeyArray<T>(this Dictionary<T, long> dictionary)
+    {
+        var keyArray = new T[dictionary.Count];
+        dictionary.Keys.CopyTo(keyArray, 0);
+        return keyArray;
+    }
+
+    public static T[] GetSortedKeyArray<T>(this Dictionary<T, long> dictionary)
```
hhyyrylainen, line 86: Why do we have manually written sort? What's wrong with LINQ?

Thrive.csproj
Maxonovien: Oh, indeed, it seems to be alphabetical, but only until a certain point (l.171 has a break in logic and it starts being messy afterwards). Still not a reason to fix it.

src/auto-evo/RunResults.cs
hhyyrylainen: Wouldn't it then be better to put this data for example in a separate object? Perhaps internal to the extinction step if this data has no actual relation to the results.

src/auto-evo/RunResults.cs
Maxonovien: Actually, this is already a list, and so it seems to not have ToList() nor Copy() methods. Only ToArray() and CopyTo(var array). Figured out it'd be simpler to use an already existing method that might be faster.

src/general/DictionaryUtils.cs
Maxonovien: Nothing but the fact that I just didn't think of it. I just swapped to it

src/auto-evo/RunResults.cs
```diff
@@ -703,6 +847,13 @@ public class SpeciesResult
         {
             public Species Species;
 
+            /// <summary>
+            ///   Dictionary of the new species population for relevant patches,
+            ///   limited to natural increase/decrease and emergence.
+            /// </summary>
+            /// <remarks>
+            ///   Does not consider migrations.
```
hhyyrylainen, line 855: Needs to be in a para, also this doesn't consider split populations, which are similar to migrations that you need to tell the code to dynamically calculate those when needed

src/auto-evo/steps/ForceExtinction.cs
```diff
@@ -0,0 +1,80 @@
...
+        public bool CanRunConcurrently => false;
+
+        public int TotalSteps => 1;
+
+        public bool RunStep(RunResults results)
+        {
+            var populationsByPatch = results.GetPopulationsByPatch(!configuration.ProtectMigrationsFromSpeciesCap,
+                !configuration.AllowNoMutation, !configuration.ProtectNewCellsFromSpeciesCap);
+
+            foreach (Patch patch in patches)
```
hhyyrylainen, line 37:
```suggestion
            foreach (var patch in patches)
```

src/auto-evo/steps/ForceExtinction.cs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        public bool CanRunConcurrently => false;
+
+        public int TotalSteps => 1;
+
+        public bool RunStep(RunResults results)
+        {
+            var populationsByPatch = results.GetPopulationsByPatch(!configuration.ProtectMigrationsFromSpeciesCap,
+                !configuration.AllowNoMutation, !configuration.ProtectNewCellsFromSpeciesCap);
```
hhyyrylainen, line 35: What's the nullability override usage here `!`? I don't think there's any other place where it is used in the codebase.

src/auto-evo/steps/ForceExtinction.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            foreach (Patch patch in patches)
+            {
+                var speciesInPatch = populationsByPatch[patch];
+
+                // This does not take player into account as the species can never go extinct this way.
```
hhyyrylainen, line 41:
```suggestion
                // This does not take player into account as the player species can never go extinct this way.
```

src/auto-evo/steps/ForceExtinction.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                // Only bother if we're above the limit
+                if (speciesInPatch.Count + protectedSpeciesCount <= configuration.MaximumSpeciesInPatch)
+                    continue;
+
+                GD.Print("Running extinction step in patch ", patch.Name, ". ",
+                    "Total count:", speciesInPatch.Count + protectedSpeciesCount);
+
+                // Sort the species in the patch, unless protected species fill up all the place already...
+                var orderedSpeciesInPatch = (configuration.MaximumSpeciesInPatch > protectedSpeciesCount) ?
```
hhyyrylainen, line 57: Is this ternary here a "useless" optimization. I'm very certain that a OrderBy with less than 100 list items cannot be slow.

src/auto-evo/steps/ForceExtinction.cs
```diff
@@ -0,0 +1,80 @@
...
+                    speciesInPatch.Keys;
+
+                var speciesToRemoveCount = speciesInPatch.Count - Math.Max(
+                    configuration.MaximumSpeciesInPatch - protectedSpeciesCount, 0);
+
+                // Remove worst-faring species, except for the player's and protected species
+                foreach (var speciesToRemove in orderedSpeciesInPatch.Take(speciesToRemoveCount))
+                {
+                    if (speciesToRemove.PlayerSpecies)
+                        continue;
```
hhyyrylainen, line 68: Because you limit things with Take, I think continue here, makes the counts inaccure. Instead I'd just loop without the .Take and have a separate variable tracking how many things are left to remove, and the loop is broken early once that reaches 0.

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +389,129 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
 
+        /// <summary>
+        ///   Computes the final population of species, by patch.
+        /// </summary>
+        /// <remarks>
+        ///   <para>
+        ///     TODO: Currently no enforced correlation with actual changes applied.
+        ///   </para>
+        ///   <para>
+        ///     Species are only registered if they population go above 0.
```
hhyyrylainen, line 400:
```suggestion
        ///     Species are only returned if their population is above 0.
```
? is that what you meant?

src/auto-evo/RunResults.cs
hhyyrylainen: This is still not solved.

src/auto-evo/steps/ForceExtinction.cs
```diff
@@ -0,0 +1,80 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using Godot;
+
+    /// <summary>
+    ///   Forces extinction of worse-faring species to limit the number of steps in next auto-evo run
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     This was a hotfix to the 0.5.6 release.
```
hhyyrylainen, line 13:
```suggestion
    ///     This was a hotfix to the 0.5.6.1 release.
```

src/auto-evo/steps/ForceExtinction.cs
Maxonovien: I don't get what you mean. `!` is just a logical negation, isn't it?

hhyyrylainen: Whoops, I misread this. Anyway is it so necessary to use the negation *so* much? Can't you just assume that (at least the non-new options like AllowNoMutation), are in the worst possible option and just let tiny bit of extra work happen?

src/auto-evo/steps/ForceExtinction.cs
Maxonovien: No, this is intended: you don't penalize other species if the player is not doing good. So if you keep only first 10 species and the player is 11th, you still keep the first 10. You just also keep the player.

src/auto-evo/steps/ForceExtinction.cs
hhyyrylainen: I was thinking of https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving

Maxonovien:
I know some things are still to fix, I just fixed whatever I had the time to tonight.

src/auto-evo/steps/ForceExtinction.cs
hhyyrylainen: Ah so you already computed speciesToRemoveCount to take the player into account?
Reading things again, I'm not seeing that. You would need to somehow skip the protected species before you .Take, otherwise this logic just does not sit right with me.

Maxonovien: No. There are two cases (assuming the threshold is 10)

a) the player is in the top 10 species with most pop, in which case everything's normal, since they are not in the list to remove.
b) the player isn't and as such they are into the list to remove. Then you end up with 11 species remaining : the top 10 + the player. 
This is an exception to the 10 species limit, since it only concerns 1 single species that is consistently spared (unlike a new species that is at best only spared once).

RacerBG (approved):
This PR introduced no new errors or unexpected behaviour. I played for 8 generations and the population numbers weren't ridiculously high. I was accompanied by 7 species, visually I can't see any difference which is good, the game perfomance as always was getting worse at generation 5+. This specific test was done in Freebuild Mode, during normal gameplay the overall population is lower compared to FB and there is almost no slow downs for the first 15-20 generations.

EDIT: Now I know that the goal of this PR wasn't to improve the perfomance after all. :smile:

src/auto-evo/steps/ForceExtinction.cs
hhyyrylainen: But what about when there are more than 1 protected species due to the moves etc.? Wouldn't then end up not doing the right thing?

hhyyrylainen:
Could you test freebuild up to like 30-40 generations? I think that's when the too many species problem really starts to appear which this PR tries to solve.

RacerBG:
> Could you test freebuild up to like 30-40 generations? I think that's when the too many species problem really starts to appear which this PR tries to solve.

I would if I have more powerful PC. At 30-40 generations my FPS would be around 5-15.

buckly90:
Loading time for autoevo seems to have been completely erased, atleast for my computer!

Species diversity is ok, had about 6 or so species in each patch on average. Most of the ones I saw seemed to be on the small side prokaryotes.

src/auto-evo/steps/ForceExtinction.cs
Maxonovien: Other protected species, which are protected only for one generation, are excluded of `orderedSpeciesInPatch`, and are handled by another logic (I should change the var name, actually)

In the end, unless you only have protected species, you have, at most `cap` + 1 (the player's) species.

src/auto-evo/steps/ForceExtinction.cs
Maxonovien: No, because then the result I get would not accurately display what the auto-evo is doing. It's not about avoiding extra work.

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +389,129 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
+        ///   </para>
+        /// </remarks>
+        public Dictionary<Patch, Dictionary<Species, long>> GetPopulationsByPatch(bool resolveMigrations = false,
+            bool resolveSplits = false, bool includeNewSpecies = true)
+        {
+            var speciesInPatches = new Dictionary<Patch, Dictionary<Species, long>>();
+            foreach (var speciesResultEntry in results)
```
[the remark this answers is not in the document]
Maxonovien: Agreed some refactor would be legitimate. However, there might be issues with logs & reports.

src/auto-evo/steps/ForceExtinction.cs
Maxonovien: Does it hurt that much to keep it though? Not to mention that there is no theoretical limit on the number of species in the patch: if we find a way to have a fast auto-evo even with 100 species, then this part would be of interest.

src/auto-evo/steps/ForceExtinction.cs
hhyyrylainen: Okay, maybe I can now understand this. Though, that variable renaming and perhaps one or two more comments would be welcome to make this clear to future me and other future developers.

src/auto-evo/steps/ForceExtinction.cs
hhyyrylainen: Sorting a list with less than 10000 items should be very fast anyway. I think it'll be a big understandability and readability win to not use the ternary operator here. I had to look at this for like 30 seconds to figure this out, whereas without the ternary understanding would be as easy as reading the line.

src/auto-evo/steps/ForceExtinction.cs
Maxonovien: I did change a few variable names and added comments. Do you think it is clear enough now?

src/auto-evo/steps/ForceExtinction.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        public bool CanRunConcurrently => false;
+
+        public int TotalSteps => 1;
+
+        public bool RunStep(RunResults results)
+        {
+            // We gather populations that may be targeted in the patch,
+            // Depending on the parameters, this may exclude migrated or newly created species.
+            // Excluded species are only protected for one generation, so the player is a target as well,
```
hhyyrylainen, line 36: Isn't it a lie that the player is a target as the player is skipped anyway on line 70?

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +328,155 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
+            var speciesInPatches = new Dictionary<Patch, Dictionary<Species, long>>();
+            foreach (var species in results.Keys)
+            {
+                var populations = GetPopulationsByPatch(species, resolveMigrations, resolveSplits, includeNewSpecies);
+                foreach (var patchEntry in populations)
+                {
+                    if (!speciesInPatches.ContainsKey(patchEntry.Key))
+                        speciesInPatches[patchEntry.Key] = new Dictionary<Species, long>();
+
+                    patchEntry.Value.ToList().ForEach(s =>
```
hhyyrylainen, line 346: Why the ToList here? I'd prefer explicit or LINQ loop here to avoid allocating an unnecessary temporary list.

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +328,155 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
+            bool resolveSplits = false, bool includeNewSpecies = true)
+        {
+            var speciesInPatches = new Dictionary<Patch, Dictionary<Species, long>>();
+            foreach (var species in results.Keys)
+            {
+                var populations = GetPopulationsByPatch(species, resolveMigrations, resolveSplits, includeNewSpecies);
+                foreach (var patchEntry in populations)
+                {
+                    if (!speciesInPatches.ContainsKey(patchEntry.Key))
+                        speciesInPatches[patchEntry.Key] = new Dictionary<Species, long>();
```
hhyyrylainen, line 344: Use try get value here to determine if the value exists.

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +328,155 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
+
+                    patchEntry.Value.ToList().ForEach(s =>
+                        speciesInPatches[patchEntry.Key].Add(s.Key, s.Value));
```
hhyyrylainen, line 347: That way this dictionary lookup also doesn't have to happen for each iteration of the ForEach.

src/auto-evo/RunResults.cs
hhyyrylainen: Actually a manual loop might be preferrable here just to make sure we get the most performance possible out of this code which is relatively heavily used in the auto-evo.

src/auto-evo/RunResults.cs
```diff
@@ -118,6 +119,27 @@ public void AddSplitResultForSpecies(Species species, Species splitSpecies, List
...
+            foreach (var migration in results[species].SpreadToPatches.ToList())
+            {
+                if (results[species].SplitOff != null && results[species].SplitOffPatches?.Contains(patch) == true)
+                    continue;
+
+                if (migration.To == patch)
+                {
+                    results[species].SpreadToPatches.Remove(migration);
+
+                    // We may still penalize the origin patch, the migration would just have died off on its way.
```
hhyyrylainen, line 136: Please add a TODO comment here saying that it would be nice to leave some trace of this happening so that it can be tracked why the population in this patch was reduced.

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +328,155 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
+                    patchEntry.Value.ToList().ForEach(s =>
+                        speciesInPatches[patchEntry.Key].Add(s.Key, s.Value));
+                }
+            }
+
+            return speciesInPatches;
+        }
+
+        /// <summary>
+        ///   Computes the final population of a species and its splitted forms, by patch.
```
hhyyrylainen, line 355:
```suggestion
        ///   Computes the final population of a species and its split forms, by patch.
```

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +328,155 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
+        ///   </para>
+        /// </remarks>
+        public Dictionary<Patch, Dictionary<Species, long>> GetPopulationsByPatch(Species species,
+            bool resolveMigrations = false, bool resolveSplits = false, bool includeNewSpecies = true)
+        {
+            var speciesInPatches = new Dictionary<Patch, Dictionary<Species, long>>();
+
+            if (!results.ContainsKey(species))
+            {
+                GD.Print("Species ", species.FormattedName, "not found in results, returning empty dictionary.");
```
hhyyrylainen, line 369:
```suggestion
                GD.PrintErr("Species ", species.FormattedName, "not found in results, returning empty dictionary.");
```
I'd go even as far as to say that I think this should throw an exception. There's a reason I think in the original code I had exception throwing, it's a bad logic error to try to read the details of a species that wasn't included in the simulation.

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +328,155 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
+                return speciesInPatches;
+            }
+
+            var speciesResult = results[species];
```
hhyyrylainen, line 373: This should also use TryGetValue to avoid 2 dictionary lookups (once for Contains and once for the actual value).

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +328,155 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
+                }
+            }
+
+            if (resolveMigrations)
+            {
+                // Apply migrations
+                foreach (var migration in speciesResult.SpreadToPatches)
+                {
+                    // Only consider valid migrations
+                    if (migration.From == null || migration.To == null)
```
hhyyrylainen, line 411: I really dislike this silencing of errors.

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +328,155 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
+                        continue;
+
+                    // Add entries if necessary
+                    if (!speciesInPatches.ContainsKey(migration.From))
```
hhyyrylainen, line 415: use TryGetValue

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +328,155 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
+                        speciesInPatches[migration.From] = new Dictionary<Species, long>();
+
+                    if (!speciesInPatches.ContainsKey(migration.To))
+                        speciesInPatches[migration.To] = new Dictionary<Species, long>();
+
+                    // Only consider possible migrations
+                    // TODO check second condition to mimick actual effect. (> 0 ?)
+                    if (speciesInPatches[migration.From].TryGetValue(species, out var population) &&
+                        population >= migration.Population)
+                    {
```
hhyyrylainen, line 425: I think this is new? This isn't in my original code because there is a separate step to filter out invalid migrations.
I don't think this checking should be added here...

src/auto-evo/RunResults.cs
```diff
@@ -368,6 +328,155 @@ public long GetPopulationInPatch(Species species, Patch patch)
...
+        /// <summary>
+        ///   Computes the final population of a species and its splitted forms, by patch.
+        /// </summary>
+        /// <remarks>
+        ///   <para>
+        ///     Species are only returned if their population is above 0.
+        ///   </para>
+        /// </remarks>
+        public Dictionary<Patch, Dictionary<Species, long>> GetPopulationsByPatch(Species species,
+            bool resolveMigrations = false, bool resolveSplits = false, bool includeNewSpecies = true)
```
hhyyrylainen, line 363: I find this naming pretty confusing so this is given the species but it also returns split off (ie. different species objects) from this? I'd rename this method. And add xml docs for all the parameters to explain this better.

hhyyrylainen (changes requested):
I thought this was going to be fine now, but reading through RunResults I saw you changed the logic in there a *lot*, so I had to make a lot of comments.

src/auto-evo/steps/ForceExtinction.cs
Maxonovien: The line right below precise the player would be rescued if needed. But it is still treated as any other species, otherwise

hhyyrylainen: Could you maybe then rename this? As these aren't even the extinction targets, then...

Maxonovien: I could rename it all as `targetAndPlayer...` but I'm not sure that'd be an improvement in clarity. Really the player is treated as a target for the most part of the computation...

This especially allows to 1) tell the player they are lagging very hard in the competition (as you save them); 2) potentially narrow this rescue to the patch the player currently plays in; 3) enable a potential hardcore mode where the player can go extinct

hhyyrylainen: Why is the player even included here, btw?
Even if the player was forced to 0 population by auto-evo, they actually won't even go extinct.
I think I would feel better about this whole variable if the player was excluded.

src/auto-evo/RunResults.cs
Maxonovien: Would `GetOffspringPopulationByPatch` look nice to you? During a splitoff, even the main branch *technically* becomes an offspring, I'd say.

src/auto-evo/RunResults.cs
Maxonovien: The main reason I didn't bother with errors was that it was a duplicate, for starters. Now that it became central I agree it really should include errors.

src/auto-evo/RunResults.cs
hhyyrylainen: Maybe "clade" or some other species family tree related term would be better than offspring? I've never seen species being referred to as offspring.

src/auto-evo/RunResults.cs
hhyyrylainen: Also probably good to throw exceptions as we have a serious algorithm bug if a migration has invalid data. See for example RemoveInvalidMigrations, which I put effort into in order to ensure migrations are always valid in the results (well after that step has ran).

src/auto-evo/steps/ForceExtinction.cs
Maxonovien: Yeah, but they would be a burden to biodiversity. 

Unlike protected cells who are protected to allow variation in the cellular fauna, the player is protected because it is the player, but they actually don't bring much. So this approach allows the player not to penalize other species whose only wrong would have been to not be protected by the god of entertainment.

src/auto-evo/RunResults.cs
Maxonovien: Yeah, I guess "clade" is probably more accurate a term. I'll use that for the time being, and I'll try to catch a biologist tomorrow to confirm.

src/auto-evo/RunResults.cs
Maxonovien: Added a bunch of errors throwing, especially one to check that you don't move more pop than you can.

PR #2877, opened 2021-11-29
Cell Process panel (and tutorial popups) is on top of menu

### LINKED ISSUE #2326: Cell Process panel (and tutorial popups) is on top of menu

takko:

Created a draggable and resizable Control and changed Cell Process, CheatMenu, EditorCheatMenu and Tutorial Dialogs to this Control.

**Related Issues**

closes #2326

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/Thrive.csproj
+++ b/Thrive.csproj
@@ -107,6 +107,7 @@
     <Compile Include="src\gui_common\dialogs\CustomConfirmationDialog.cs" />
     <Compile Include="src\gui_common\dialogs\ErrorDialog.cs" />
     <Compile Include="src\gui_common\dialogs\LicensesDisplay.cs" />
+    <Compile Include="src\gui_common\CustomControlDialog.cs" />
     <Compile Include="src\gui_common\tooltip\DefaultToolTip.cs" />
     <Compile Include="src\gui_common\tooltip\ICustomToolTip.cs" />
     <Compile Include="src\gui_common\tooltip\microbe_editor\ModifierInfoLabel.cs" />
--- a/src/engine/ControlHelpers.cs
+++ b/src/engine/ControlHelpers.cs
@@ -28,4 +28,50 @@ public static class ControlHelpers
             });
         }
     }
+
+    /// <summary>
+    ///   Shows the control in the center of the screen and shrinks it to the minimum size.
+    /// </summary>
+    public static void ControlCenteredShrink(this Control control, bool runSizeUnstuck = true)
+    {
+        var parentRect = control.GetViewport().GetVisibleRect();
+
+        var rectSize = control.GetMinimumSize();
+        if (rectSize == default(Vector2))
+        {
+            foreach (Control child in control.GetChildren())
+            {
+                if (child != null)
+                {
+                    var pos = child.RectPosition;
+                    var min = child.GetCombinedMinimumSize();
+
+                    rectSize.x = Mathf.Max(pos.x + min.x, rectSize.x);
+                    rectSize.y = Mathf.Max(pos.y + min.y, rectSize.y);
+                }
+            }
+        }
+
+        var rectPos = parentRect.Position + (parentRect.Size - rectSize) / 2;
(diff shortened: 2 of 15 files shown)


RacerBG:
I had a fair bit of testing with this PR and here are the results:

1. No new crashes or errors - the log is clean.
2. The pop ups which you mentioned are all resizeable and the text formatting seems to adapt pretty well.
3. The cell processes pop up window does not intervene with the menu, the other pop ups are working as intended too.
4. (Optional) There could probably be a maximum resize size for every type of pop up but this might be included in another PR.

In summary: this PR fixes one of my biggest annoyances with the cell processes window and I'm grateful to you for doing this!

Kasterisk:
Regarding the implementation, I'm a bit on the fence about the duplicated logic between the two classes. Just crossed my mind, would it work if we instead make the PauseMenu's root node be a popup? I think that will block all other lower popups.

hhyyrylainen:
I think it would make sense to instead fix the pause menu to be on top of everything else than trying to fix everything else separately.

takko:
But wouldn't that need the Load and Save Dialog to also be on top of the PauseMenu. I think having those TopMost Dialog's is always a problem as you cannot simply control the order.
So either have all (Dialogs, Menus, ...) as Popups or as Controls. Mix is a problem.

I could get rid of some duplicated code via composition. But I did not want to change that code, as it is copied somewhere from godot, and if that code needs change it would be easier to copy the changed godot code again if I do not edit it.

So it's your decision. If you want the PauseMenu as Popup I do that. But I would still keep that new class as it may come handy sometime.

hhyyrylainen:
All of the pause menu content like the load, options, save, would be part of the pause menu popup, so they would stay along with it on top?

takko:
Yes, I guess this may could be. But still, there is also the Extinction Screen Menu. And I'm not sure about Win Menu. Maybe there are more.
I just think that the behavior of the Popup as Dialog is wrong. Not drawn on top but receiving mouse input first is a bug, which we could avoid using Control Dialogs.
I think Popups are only usable as Modal Dialogs.

hhyyrylainen:
Extinction screen and the win screen are contained in the same container, I think.

takko:
They have the same control as holder but are dynamically added via code.
So I guess changing the holder from Control to WindowDialog would be enough.

hhyyrylainen:
Doesn't window dialog have decorations by default? Wasn't the suggestion to use the base popup class to just act as a popup but with no visuals?

84634E1A607A:
> make the PauseMenu's root node be a popup?

This do solve all the problems.

By the way, resize at corners are not handled in CustomWindowDialog, and I suggest make Dialog resizable zone thicker. The current setting is not that easy to use.

84634E1A607A:
> So I guess changing the holder from Control to WindowDialog would be enough.

To Popup. Window dialog has backgrounds.

hhyyrylainen (changes requested):
https://github.com/Revolutionary-Games/Thrive/pull/2877#issuecomment-986000940

84634E1A607A:
Maybe this can be achieved in another PR.

hhyyrylainen:
If someone wants to make that change as a separate PR, I'll welcome it.

84634E1A607A:
would it work if we instead make the PauseMenu's root node be a popup? I 

Wouldn't it?

hhyyrylainen:
When I said "If someone wants to make that change as a separate PR, I'll welcome it." I was referring to 

> make the PauseMenu's root node be a popup?

84634E1A607A:
No. What I wanted to say is that I agree with @Kasterisk , that this pr is actually not worth the trouble.

PR #2881, opened 2021-11-30
Rewrite mutation code to not need to retry in case of islands

### LINKED ISSUE #1796: Rewrite mutation code to not need to retry in case of islands
As that approach can eventually fail to run, it'd be nice to write a new species mutate method that is built with the island check in mind and it doesn't need to retry the entire mutation attempt if there is a problem.
As the current 10 attempts limit can be reached quite easily, if the player makes very silly cells with a lot of thin cytoplasm strands, a new algorithm that doesn't have this problem should be made.

takko:

added code which moves the islands to the base.

**Related Issues**

closes #1796

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/general/Mutations.cs
```diff
@@ -170,96 +170,132 @@ private void MutateBehaviour(Species parent, Species mutated)
     {
         var nucleus = SimulationParameters.Instance.GetOrganelleType("nucleus");
 
-        for (var iteration = 0; iteration < 10; iteration++)
+        mutatedOrganelles.Clear();
```
adQuid, line 173: Is this still needed without iterations to wipe away?

src/general/Mutations.cs
```diff
@@ -170,96 +170,132 @@ private void MutateBehaviour(Species parent, Species mutated)
...
+                        // early exit if distance == 2
+                        if (minDistance == 2)
+                            break;
+                    }
+                }
 
-                // If still empty, copy the first organelle of the parent
-                if (mutatedOrganelles.Count < 1)
-                    mutatedOrganelles.Add((OrganelleTemplate)parentOrganelles[0].Clone());
+                // early exit if distance == 2
```
adQuid, line 282: why does distance == 2 mean early exit? And is that the same as minDistance == 2?

adQuid (commented):
Not able to run proper tests right now, and not an expert in hex coordinate math, but I've just got two comments in the code. This isn't code I would expect to fail silently, so I'm happy to approve if people test this enough times and nothing crashes.

src/general/Mutations.cs
takko: This list is provided as parameter and needs to be cleared here, as it contains the copied organelles from the parent.

src/general/Mutations.cs
takko: Updated the comment to make it more clear.
distance 1 == direct neighbor, distance 2 == min distance for island
distance is always >= minDistance, but changed the comment

src/general/Mutations.cs
adQuid: We provide a list as a param, specifically to clear it out? Sounds like we can just drop the param and declare the variable in here.

hhyyrylainen: It's easier to pass in the mutated list of organelles. As the mutated species is created by cloning, so it would actually be more work to specifically remove the organelles from it to process here.

hhyyrylainen:
Unless someone else can take a look at this, I'll merge this immediately after 0.5.6.1 to give this time to have more people experience the game with this in it.

src/general/Mutations.cs
```diff
@@ -170,96 +170,132 @@ private void MutateBehaviour(Species parent, Species mutated)
...
             if (mutatedOrganelles.Count < 1)
+                mutatedOrganelles.Add((OrganelleTemplate)parentOrganelles[0].Clone());
+        }
+
+        var islandHexes = mutatedOrganelles.GetIslandHexes();
+
+        // Attach islands
+        while (islandHexes.Count > 0)
+        {
+            var mainHexes = mutatedOrganelles.ComputeHexCache().Except(islandHexes);
```
hhyyrylainen, line 260: Does this work correctly if there are multiple islands on different sides of the microbe? What if those are exactly symmetrical and the average distance is zero?

That's basically my only code concern. This PR basically just needs a lot of testing to make sure it never fails.

hhyyrylainen:
This needs conflict resolving and then testing to make sure this works fine. I'm sure @RacerBG can get some testing done once master is merged into this branch.

buckly90:
I gave it a spin and did not notice any errors in the autoevo notes throughout 25 generations

I was originally aiming for about 30 generations but auto-evo became unbearably slow near the end. Not sure if that is related to this PR or not.

hhyyrylainen:
I guess this PR can slightly increase the auto-evo running time. Or if things get unlucky maybe a lot because 5 mutations are generated for each species so if this doubles the time it takes to generate mutations (which I doubt), it could be impactful. Though I suspect the overall auto-evo running time is still within 5-10% of what it is one the master branch at this point.

Actually as the conflict is not yet fixed, master has not been merged into this, so it might be that my auto-evo running performance improvements aren't included here.

RacerBG:
> I guess this PR can slightly increase the auto-evo running time. Or if things get unlucky maybe a lot because 5 mutations are generated for each species so if this doubles the time it takes to generate mutations (which I doubt), it could be impactful. Though I suspect the overall auto-evo running time is still within 5-10% of what it is one the master branch at this point.
> 
> Actually as the conflict is not yet fixed, master has not been merged into this, so it might be that my auto-evo running performance improvements aren't included here.

I will test this PR after the merge. In terms of lag and perfomance issues I'm probably the most suited one to tell the difference thanks to my hardware.

hhyyrylainen (changes requested):
Needs to have the merge conflict fixed and playtesting before this can proceed.

PR #2888, opened 2021-12-01
Clip color picker HSV & Raw checkbuttons text to fix too long texts

### LINKED ISSUE #2867: Membrane tab width issue in specific languages
It seems that there is possibility of a really long translation messing with the membrane tab width in the editor:
https://youtu.be/32zs9Vnp7Fg?t=639

probably needs to add line wrapping or something to a label that probably causes that...

Kasterisk:

Too long texts are now clipped to avoid the editor's panel widening in specific languages. Also probably needs to add notice in the translation site that HSV and Raw translation should be as short as possible or just leave it unchanged.

**Related Issues**

Fixes #2867 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
Added the explanations on weblate: https://translate.revolutionarygamesstudio.com/translate/thrive/thrive-game/en/?checksum=aace9ecdaba9ab08
https://translate.revolutionarygamesstudio.com/translate/thrive/thrive-game/en/?checksum=ba347f3999d8fcf

hhyyrylainen (changes requested):
Could the checkbox widths be a bit more generous? In Finnish I've translated "raw" as "raaka" which would fit in the space in the UI (at least if a little bit of space between HSV and the Raw checkboxes could be taken up by an extra letter).

Kasterisk:
Widen the minimum widths, how does this look now?

RacerBG:
I made longer text on purpose but it gets cut off after 4 characters. If this is the intended behaviour then someone should made an explanation in Weblate that the maximum length of the string should be 4 characters and then this PR will be ready for a merge.

Kasterisk:
Yes it's the intended behavior, though I may want to increase this to allow 5 characters minimum. Specifically mentioning 5 characters in weblate might be a good idea. As for general solution, I think I'll just enable horizontal fill and expand flag for the checkboxes at this point.

hhyyrylainen:
> Maybe it's just me but the field about the color code seems useless, considering that the color picker does the same job. And if this is removed then longer strings could be feasible.

Every colour picker on the planet has support for pasting in a colour code you got from a website for example (that helps you design a coherent colour scheme).

Kasterisk:
I think fundamentally the color picker is already loaded with excessive functionalities in the first place for a simple membrane color picker. The checkboxes can just be removed for example as does the preset button and sliders. But this is kind of all or nothing situation and we may have to just settle with leaving these on.

hhyyrylainen (changes requested):
It still doesn't look like 5 characters fit. Could you maybe edit the label's max visible characters property to fix it at 5? Otherwise the min width needs a bit more changing still.

Kasterisk:
Should now fit.

hhyyrylainen:
Still looks like this:



Am I doing something wrong regarding checking out the latest changes? Did you try with Finnish to see if "Raaka" fits?

Kasterisk:
Apparently I had my Thrive window in different size which changes the font size... though this gives me the idea to just reduce the font size now so that it fits.

hhyyrylainen (approved):
Yeah, the smaller font looks overall better and now it fits.

RacerBG (approved):

PR #2911, opened 2021-12-06
Emergency auto-evo changes

hhyyrylainen:

Just some things I hope makes auto-evo more playable with the player being affected by it now for 0.5.6.1 so we can then later improve things
added explanation of auto-evo prediction with numbers included.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -207,7 +221,7 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
                     / energyBalanceInfo.FinalBalanceStationary);
 
                 // Severely penalize a species that can't osmoregulate
-                if (energyBalanceInfo.FinalBalance < 0)
+                if (energyBalanceInfo.FinalBalanceStationary < 0)
```
adQuid, line 224: It's badly labeled, but my reasoning for this was that most AI is always moving, so the moving ATP balance is the one that counts. IIRC I noticed a lot of AI microbes that just immediately killed themselves in motion in 0.5.5, so if you want to keep this change it would be wise to test for a while to make sure you don't get self-starving microbes.

Personally, I would suggest just not making this change at all. This doesn't address the "emergency" since players weren't complaining about ATP-negative species not doing well in auto-evo, and an AI that kills itself is one of those "stupid" things that really stand out to players.

A middle ground would be to do a sessile check, looking at the Activity attribute and picking a balance based on being in the bottom tenth range. That would have the drawback of being brittle in the future though.

src/auto-evo/simulation/food_source/MarineSnowFoodSource.cs
```diff
@@ -1,4 +1,5 @@
-﻿using AutoEvo;
+﻿using System;
```
adQuid, line 1: import still needed?

src/auto-evo/simulation/food_source/MarineSnowFoodSource.cs
```diff
@@ -36,7 +39,9 @@ public override float FitnessScore(Species species, SimulationCache simulationCa
             score *= Constants.AUTO_EVO_CHUNK_LEAK_MULTIPLIER;
         }
 
-        score /= energyBalance.FinalBalanceStationary;
+        // I don't think it makes sense here to penalize species with positive ATP balance (with higher balance
+        // reducing the available energy), so that divide by final balance stationary here is gone - hhyyrylainen
+        score *= Constants.AUTO_EVO_MARINE_SNOW_BASE_MULTIPLIER;
```
adQuid, line 44: two things here:

- food source fitness scores should only be used to compare to other scores from the same source, so I hope that a constant multiplier is useless (or I designed this wrong)
- This was intended to be a penalty for higher osmoregulation costs, so scavengers don't get bigger than they need to be. Combined with the bonus for energy balance on line 30, I think you will start to see a bias towards progressively larger species with smaller and smaller populations. Ironically this might lead to patches periodically dieing off, but I wouldn't know without a lot of testing.

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -177,10 +181,20 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
...
+                    if (strictCompetition)
+                    {
+                        // Softly enforces https://en.wikipedia.org/wiki/Competitive_exclusion_principle
+                        // by exaggerating fitness differences
+                        thisSpeciesFitness =
+                            Mathf.Max(Mathf.Pow(niche.FitnessScore(currentSpecies, cache), 2.5f), 0.0f);
+                    }
+                    else
+                    {
+                        thisSpeciesFitness = Mathf.Max(niche.FitnessScore(currentSpecies, cache), 0.0f);
```
adQuid, line 195: I am pretty confident that this would backfire VERY badly. My guess is that the single most important part of the auto-evo refactor was adding the exclusion principle, which allows for a diversity of species instead of one all-rounder dominating the field. If a lot of testing proves otherwise I'm happy to re-evaluate, but otherwise if you take no other suggestions from this review multiply this by an exponent meaningfully greater than one. Doing otherwise is very likely to result in a very undynamic environment of NPC species.

src/microbe_stage/EnergyBalanceInfo.cs
```diff
@@ -18,7 +18,15 @@ public class EnergyBalanceInfo
     public float TotalProduction { get; set; }
     public float TotalConsumption { get; set; }
     public float TotalConsumptionStationary { get; set; }
+
+    /// <summary>
+    ///   The absolutely final balance of ATP when a cell is going all out and running everything and moving
```
adQuid, line 23: does this include stuff like engulfing, or is it JUST movement?

adQuid (changes requested):
In general I think this goes a touch over into making the auto-evo not strict enough to keep the AI species challenging and diverse. In particular my guess is that dropping the niche exclusion entirely will be an "out of the frying pan, into the fire" situation that causes a bigger problem than this solves.

src/auto-evo/simulation/food_source/MarineSnowFoodSource.cs
adQuid, line 44: This was supposed to be total consumption, not final balance. There are also other places this mistake is made. This is supposed to be a penalty for wasting energy that doesn't help you get this food source.

src/auto-evo/simulation/food_source/MarineSnowFoodSource.cs
hhyyrylainen: System import was added as the first one. And both are needed according to my IDE.

src/auto-evo/simulation/PopulationSimulation.cs
hhyyrylainen: The reason I made this change is due to players experiencing that adding basically *any* organelle just saps their population. So I thought that if the competition was less harsh, then adding an organelle that is not very heavily competed, *should* be better.
So as the branch says, this is an emergency change. Though, if this actually makes the gameplay experience worse, I'll disable it before merging.

src/microbe_stage/EnergyBalanceInfo.cs
hhyyrylainen: Engulfing is not included *at all* in the energy balance info. Should I add a remark about that? Also the energy cost of binding mode is not included anywhere.

src/auto-evo/simulation/food_source/MarineSnowFoodSource.cs
hhyyrylainen: Shouldn't this be, then `TotalConsumptionStationary`? As you still need to move a bit to find chunks.

src/auto-evo/simulation/PopulationSimulation.cs
hhyyrylainen: > Personally, I would suggest just not making this change at all. This doesn't address the "emergency" since players weren't complaining about ATP-negative species not doing well in auto-evo, and an AI that kills itself is one of those "stupid" things that really stand out to players.

I disagree in that it often seems that the auto-evo *really* wants the player to go extinct. Which would make sense if they triggered this divide by 10 line.
Shouldn't the AI by this point know how to stop to recharge ATP?

adQuid: Sadly it does not, since the AI refactors never got around to "out of ATP vs starving and need to find food" and "want to run away vs need to run away" calculations.

A compromise option might be to hold the player and AI to different rules, letting the more responsible humans play with ATP-flickering species but not the AI. Even then though the player's sibling will likely run off and die.

src/microbe_stage/EnergyBalanceInfo.cs
adQuid: "running everything" sure implies that it's with all possible ATP-consuming things. Other than that I don't think it's too big a deal.

src/auto-evo/simulation/food_source/MarineSnowFoodSource.cs
adQuid: That makes sense to me.

src/microbe_stage/EnergyBalanceInfo.cs
hhyyrylainen: You run processes. Though I guess you don't really *run* osmoregulation
This'll get more complicated if / once we add the sprint mechanic.
I'll add a comment to try to clarify this (as I myself have found to struggle a bit to remember what all is included in all of these calculations).

src/auto-evo/simulation/PopulationSimulation.cs
hhyyrylainen: I guess technically we could check the PlayerSpecies flag here... well it's not too dirty for me to consider putting it in this PR.

adQuid: We're in a tough position: the same mechanic needs to do different things for the player and AI. I'm sorry I didn't think about this when you first made a PR bringing auto-evo into the player experience.

hhyyrylainen: I added in the hack I mentioned.

src/microbe_stage/EnergyBalanceInfo.cs
hhyyrylainen: Added some clarifying comments there.

src/auto-evo/simulation/food_source/MarineSnowFoodSource.cs
hhyyrylainen: Changed to that.

src/auto-evo/simulation/PopulationSimulation.cs
adQuid: The problem is that the AI is trying on random organelles far more often than the player does, so if that's NOT being punished I suspect you will find the "one of everything" species we saw in 0.5.4. Is there any chance I could convince you to pull just this change out to a separate PR, to be tested independently? I'm certainly interested in double-checking my assumptions, but I don't want to conflate variables with the other changes being done here, which I suspect are much safer.

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -206,10 +234,29 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
...
+                        newPopulation /= 10;
+                    }
+                }
+                else
+                {
+                    // Severely penalize a species that can't move indefinitely
+                    if (energyBalanceInfo.FinalBalance < 0)
+                    {
+                        newPopulation /= 10;
+                    }
```
Maxonovien, line 259: Shouldn't we also add in some sessility factor? Otherwise that could be very detrimental to AI plant species.

hhyyrylainen: Maybe but that'll have to wait for another PR.

RacerBG:
I will write a longer review later today. I will focus on aesthetics for the moment:
1. The Auto-Evo prediction on the bottom right should be made taller in my opinion because with some languages it could lead to necessary scrolling.
2. The values should be rounded, I already suggested this in different PR's. Generally speaking every value in the game which is a decimal value should use the same rules for example 2.07 or 3.4578.
3. It is quite unclear what is the meaning of "energy" in the Prediction window (for a new player that is)  also the explanation text needs improving.

hhyyrylainen:
With the player species 2 patches, the text can't fit on screen even if the popup is full screen height.
I'll try to clarify energy. And also throw in rounding to three (?) decimal places.

RacerBG:
> With the player species 2 patches, the text can't fit on screen even if the popup is full screen height. I'll try to clarify energy. And also throw in rounding to three (?) decimal places.

It's up to you. The rounding will improve readability but as I said the same rule should apply to every part of the game. Either 2 or 3 numbers after the ".". Maybe we should ask this in Discord?

hhyyrylainen:
Different places in the game require different shown precisions. You may notice in the editor screenshot already 1 and 3 decimal places being used. I'll use 3 decimals here as well as some pretty small numbers will need to be shown.

PR #2913, opened 2021-12-07
Sector spawn system

FuJa0815:

Overhauls the spawn system to utilize sector spawning.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


FuJa0815:
I think the first iteration of the technical implementation should be more or less done. Balancing will be horrible in this state.

src/general/Sector.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return $"{X};{Y}";
+    }
+
+    private static (int X, int Y) GetSectorCoords(Vector2 position)
+    {
+        var x = (int)(position.x / Constants.SECTOR_SIZE);
+        var y = (int)(position.y / Constants.SECTOR_SIZE);
+        return (x, y);
+    }
```
Maxonovien: Sector size should probably be factored into a parameter for the sector, seeing that this class could be used for other purposes, especially in cloud systems (which uses a sector logic without mentioning it)

src/microbe_stage/Membrane.cs
```diff
@@ -269,10 +271,11 @@ public Vector2 FindClosestOrganelles(Vector2 target)
     {
         // The distance we want the membrane to be from the organelles squared.
         float closestSoFar = 4;
-        Vector2 closest = new Vector2(INVALID_FOUND_ORGANELLE, INVALID_FOUND_ORGANELLE);
+        Vector2 closest = invalidVector;
 
-        foreach (var pos in OrganellePositions)
+        for (var i = 0; i < OrganellePositions.Length; i++)
         {
+            var pos = OrganellePositions[i];
```
Maxonovien: Why did you change this? The previous handling looked better in my eyes.

More generally I feel you changed quite a lot of files (sometimes with good reasons) with no relation to what you wanted to bring, and I feel this makes this PR very dense...

FuJa0815: This change was one of many tiny performance improvements I've tried to spawning microbes

hhyyrylainen: A separate PR is probably warranted as this doesn't fundamentally have anything to do with spawn sectors...

src/microbe_stage/PatchManager.cs
```diff
@@ -156,60 +152,40 @@ private void HandleCellSpawns(Patch patch)
         foreach (var entry in patch.SpeciesInPatch)
         {
             var species = entry.Key;
+            var populationHere = entry.Value;
```
Maxonovien, line 155: Two concerns on this:
a) wouldn't it be more efficient not to instantiate variables here
b) if we stick by this, we probably should settle for a solution when dealing with dictionary entries, cause other parts in the code do not create such variables.

And, if clarity is the issue, perhaps a more telling name than entry could be considered.

src/microbe_stage/SpawnedRigidBody.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Godot;
+
+/// <summary>
+///   All nodes that can be spawned with the spawn system must derivative from this class
```
Maxonovien, line 4:
```suggestion
///   All nodes that can be spawned with the spawn system must be a derivative from this class
```

And perhaps we'd even prefer another term like "must inherit this class"?

src/microbe_stage/Spawner.cs
```diff
@@ -1,50 +1,113 @@
...
 
 /// <summary>
 ///   Spawner that can be added to a SpawnSystem to be used for spawning things
 /// </summary>
 public abstract class Spawner
 {
+    private float[]? binomialValuesCache;
+
+    public virtual int BinomialN => 0;
+    public virtual float BinomialP => 0;
```
Maxonovien, line 14: Ok, there is the explanation for the factorial args. However, would we not want to swap that to an actual utils class such as the math one?

src/microbe_stage/Spawner.cs
```diff
@@ -1,50 +1,113 @@
...
     /// </summary>
-    /// <param name="worldNode">The parent node of spawned entities</param>
-    /// <param name="location">Location the spawn system wants to spawn a thing at</param>
-    /// <returns>An enumerator that on each next call spawns one thing</returns>
-    public abstract IEnumerable<ISpawned>? Spawn(Node worldNode, Vector3 location);
+    /// <returns>Returns the amount of spawns in a sector.</returns>
+    protected virtual int GetSpawnsInASector(float sectorDensity, Random random)
+    {
+        if (BinomialP == 0)
+            return 0;
```
Maxonovien, line 91: What is binomial P supposed to stand for?

src/microbe_stage/Spawner.cs
```diff
@@ -1,50 +1,113 @@
...
+
+        var nextRandom = random.NextFloat() * sectorDensity;
+        var binomialSum = 0f;
+        var i = 0;
+        do
+        {
+            binomialSum += BinomialValues[i++];
+        }
+        while (nextRandom >= binomialSum);
```
Maxonovien, line 100: Why not using a simple while loop here?

src/microbe_stage/Spawner.cs
```diff
@@ -1,50 +1,113 @@
...
+
+        return i - 1;
+    }
```
Maxonovien, line 103: Why this probability law in particular? I assume it at least privileges central values, since nCp is highest for p = n/2 (if I'm not mistaken).

If it is a known method, could you please add a reference in the doc? That would be super helpful in reviewing the mechanics, especially if we consider tweaking it.

src/microbe_stage/Spawners.cs
```diff
@@ -366,7 +364,11 @@ public CompoundCloudSpawner(Compound compound, CompoundCloudSystem clouds, float
         this.amount = amount;
     }
 
-    public override IEnumerable<ISpawned>? Spawn(Node worldNode, Vector3 location)
+    public override int BinomialN => (int)(amount / 50000);
+    public override float BinomialP => 0.1f;
```
Maxonovien, line 368: I'm definitely confused as to what these values represent in your algorithm, and why you give the these values.

Maxonovien (commented):
Some general comments about the maths and a few minor remarks. Overall it seems well thought, but I think some parts remain to obscure in their details.  Keep up the great work!

src/microbe_stage/PatchManager.cs
hhyyrylainen: I've recently discovered that you can even destructure variables (usually) with something like: `foreach(var (thing, secondThing) in container)` which is very awesome. I would expect the compiler to optimize away any temporary variables or at least that isn't much of a problem if this is code that isn't ran every frame.

RacerBG:
This PR should be superseded already by PR #3235 .

hhyyrylainen:
This has bunch of stuff like storing the spawned in sector data etc. that this is actually even more advanced than that spawn system change that did get done.

PR #2915, opened 2021-12-07
Show label in empty load game screen

### LINKED ISSUE #2748: Load game screen is totally empty without saves.
Without a save to load, the load game screen is totally empty. This looks very weird.
There should at least be a message saying no saves to load.
Or maybe the load game button should be disabled.

lordiii:

Implements a label to be displayed if no saves are found.
Also adds a translatable string related to that.

**Related Issues**

Closes #2748 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/saving/SaveList.tscn
```diff
@@ -104,7 +112,6 @@ WindowTitle = "SELECTED_SAVE_IS_UPGRADEABLE"
 DialogText = "SAVE_IS_UPGRADEABLE_DESCRIPTION"
 
 [node name="SaveUpgradeFailedDialog" parent="MarginContainer/VBoxContainer" instance=ExtResource( 4 )]
-visible = false
```
Kasterisk, line 107: This may be unintended change.

src/saving/SaveList.tscn
```diff
@@ -43,10 +44,17 @@ margin_bottom = 715.0
 size_flags_horizontal = 3
 size_flags_vertical = 3
 
+[node name="NoSavesLabel" type="Label" parent="MarginContainer/VBoxContainer"]
+visible = false
+margin_right = 1270.0
+margin_bottom = 25.0
+text = "NO_SAVEGAMES_FOUND"
+align = 1
```
Kasterisk, line 52: Having the text vertically aligned to the center of the save list might be better.

Kasterisk (changes requested):

RacerBG (commented):
After applying the changes which Kasterisk suggested this PR will be good to go. It's quite clever implementation.

src/saving/SaveList.tscn
lordiii: Godot seems to remove any lines containing `visible = false`.
Cause might be Godot interpreting missing `visible`-values as false.
Re-Adding that in line in the SaveList.tscn file, results in Godot removing it.

hhyyrylainen: I've seen this type of change many times before. This is caused by setting a node that we have purposefully hidden, to be visible. The default value for a node is to be visible, meaning that when Godot removes this line, that's the same as making the node visible. So you need to in the Godot editor find this node and click it back to hidden state, or use git to revert this line.

lordiii: Node was already hidden, got removed anyway.
Used git to revert the line.

Kasterisk (approved):
LGTM. Thanks for contributing!

hhyyrylainen:
This is good to merge once the translation merge conflicts are resolved. It may be easier to reset the changes to the translation files, merge master, and then run the localize script again, and put back in the English text.

hhyyrylainen:
I took care of the merge conflict even though it turned out to be very simple (just the POT update header lines were conflicting).

hhyyrylainen (approved):
Looks good enough now with the small change I did. If I were to nitpick I'd extract the now a bit too indented for my taste block to a new method.

PR #2916, opened 2021-12-07
Compound venting on engulf

Maxonovien:

This PR makes cells vent excess compounds when engulfing a chunk.

It also fixes a bug in `AddCompound` method from `CompoundBag`, where the return value was the previous amount in the bag rather than what was added. This return was mistakenly used for colonies compounds sharing; no other call did make actual use of the return value.

*Nota: Observing the effect is much easier when disabling chunk venting.*

Also, since I changed returns of functions used by colony, it would be good to confirm everything still works.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/microbe_stage/ColonyCompoundBag.cs
+++ b/src/microbe_stage/ColonyCompoundBag.cs
@@ -88,18 +88,20 @@ public class ColonyCompoundBag : ICompoundStorage
 
     public float AddCompound(Compound compound, float amount)
     {
-        var originalAmount = amount;
+        var totalAmountAdded = 0.0f;
 
         foreach (var bagToAddTo in GetCompoundBags())
         {
             var amountAdded = bagToAddTo.AddCompound(compound, amount);
+
+            totalAmountAdded += amountAdded;
             amount -= amountAdded;
 
             if (amount <= MathUtils.EPSILON)
                 break;
         }
 
-        return originalAmount - amount;
+        return totalAmountAdded;
     }
 
     public void ClearCompounds()
--- a/src/microbe_stage/CompoundBag.cs
+++ b/src/microbe_stage/CompoundBag.cs
@@ -78,7 +78,7 @@ public class CompoundBag : ICompoundStorage
 
         Compounds[compound] = newAmount;
 
-        return newAmount - amount;
+        return newAmount - existingAmount;
     }
 
     public IEnumerator<KeyValuePair<Compound, float>> GetEnumerator()
--- a/src/microbe_stage/FloatingChunk.cs
+++ b/src/microbe_stage/FloatingChunk.cs
@@ -271,8 +271,6 @@ public class FloatingChunk : RigidBody, ISpawned, ISaveLoadedTracked
                     // Can engulf
                     if (ContainedCompounds != null)
                     {
-                        // TODO: could spill the amount that was wasted here as a cloud
-
                         foreach (var entry in ContainedCompounds)
                         {
                             var added = microbe.Compounds.AddCompound(entry.Key, entry.Value /
--- a/src/microbe_stage/FloatingChunk.cs
+++ b/src/microbe_stage/FloatingChunk.cs
@@ -275,8 +275,10 @@ public class FloatingChunk : RigidBody, ISpawned, ISaveLoadedTracked
 
                         foreach (var entry in ContainedCompounds)
                         {
-                            microbe.Compounds.AddCompound(entry.Key, entry.Value /
-                                Constants.CHUNK_ENGULF_COMPOUND_DIVISOR);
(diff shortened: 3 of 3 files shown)


src/microbe_stage/CompoundBag.cs
```diff
@@ -78,7 +78,7 @@ public float AddCompound(Compound compound, float amount)
 
         Compounds[compound] = newAmount;
 
-        return newAmount - amount;
+        return newAmount - existingAmount;
```
hhyyrylainen, line 81: I very strongly think that this will have a bunch of effects elsewhere. I think the return value was changed for colonies to work correctly. So changing the return value here is very likely to cause issues.

RacerBG:
> cells vent excess compounds when engulfing a chunk

Could you elaborate on that?

src/microbe_stage/CompoundBag.cs
Maxonovien: As I wrote, the only possible effects are for colonies: it is called 9 times, 6 of which are calls without storing the variable. Out of the 3 remaining calls, 2 are from `MicrobeInterior.cs`, with no effect, as you can see right bellow

```
                var amountAdded = copyEntity.Compounds.AddCompound(compound, amount * 0.5f);

                if (amountAdded < some_var)
                {
                    // TODO: handle the excess compound that didn't fit in the other cell
                }
         } // exit function, the var was local
```

The last call is used in `ColonyCompoundBag.cs`, as follows:
```
public float AddCompound(Compound compound, float amount)
    {
        var originalAmount = amount;

        foreach (var bagToAddTo in GetCompoundBags())
        {
            var amountAdded = bagToAddTo.AddCompound(compound, amount);
            amount -= amountAdded;

            if (amount <= MathUtils.EPSILON)
                break;
        }

        return originalAmount - amount;
    }
```
As I said, the logic seems to use the amount added (which is what is documented and what I changed the return to), and not the previous amount (which was the previous logic). But you are right the the return is wrong again...

Maxonovien:
Currently, if you engulf a chunk that has too much compounds, they just disappear. This PR would change this.

src/microbe_stage/CompoundBag.cs
hhyyrylainen: For testing simplicity, I would *really* prefer if you could do the required math in the chunk code and not change the compound bag code as that is a truly core component. It took a lot of convincing of me for the colony code to be allowed to change the return value of that method.

Maxonovien: Actually, the colony compound bag return doesn't seem to be used elsewhere either. I still fixed the logic there.

Maxonovien: The issue is that the return value of the compound bag code is said to be precisely what I changed it to! And all occurences of named variable take into account that this *should* return the added value, yet it isn't.

To me this is clearly a bug in the code, that may mess up things later on (if it didn't already).

RacerBG (approved):
The code doesn't seem to break anything.

1. The new functionality works as intended.
2. The cell colonies are not broken by this PR.
3. No new errors or strange behaviours were observed.

src/microbe_stage/ColonyCompoundBag.cs
```diff
@@ -88,18 +88,20 @@ public float TakeCompound(Compound compound, float amount)
 
     public float AddCompound(Compound compound, float amount)
     {
-        var originalAmount = amount;
+        var totalAmountAdded = 0.0f;
```
adQuid, line 91: Why this change from a count down to a count up system? Do you still need the declining `amount` variable?

Maxonovien: Actually, rereading back, I realized the refactor in this file wasn't necessary *per se*, but the fact I was mislead in the first place (arguably because the `CompoundBag.cs` equivalent was wrong), was enough of a reason for me to propose a refactor.

Diminishing `amount` is still necessary to take more than you can. 

To be note that the refactor may do a few more operations, however. I'm not that attached to it, so I'm open to dropping the last commit changing this.

adQuid: If the code became clearer to you, that's an improvement.

adQuid (approved):

src/microbe_stage/CompoundBag.cs
hhyyrylainen: Okay, I looked through the code base and this seems like a safe change.

hhyyrylainen (approved):
Code looks fine to me. I think testing was also done already so this is probably fine to merge.

PR #2917, opened 2021-12-07
fix auto-evo variable mistake

adQuid:

Reviewing another PR made me realize that auto evo divides fitness scores by the net ATP, when it is supposed to divide by ATP cost. This might lower populations a little, which I don't think would be a problem, but might also be the reason that better ATP generators are punished. It might also help with the "spikes make everything better" issue.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.

### DIFF
--- a/src/auto-evo/simulation/food_source/CompoundFoodSource.cs
+++ b/src/auto-evo/simulation/food_source/CompoundFoodSource.cs
@@ -26,7 +26,7 @@ public class CompoundFoodSource : FoodSource
 
         var compoundUseScore = EnergyGenerationScore(microbeSpecies, compound);
 
-        var energyCost = simulationCache.GetEnergyBalanceForSpecies(microbeSpecies, patch).FinalBalanceStationary;
+        var energyCost = simulationCache.GetEnergyBalanceForSpecies(microbeSpecies, patch).TotalConsumptionStationary;
 
         return compoundUseScore / energyCost;
     }
--- a/src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
+++ b/src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
@@ -23,7 +23,7 @@ public class EnvironmentalFoodSource : FoodSource
 
         var energyCreationScore = EnergyGenerationScore(microbeSpecies, compound);
 
-        var energyCost = simulationCache.GetEnergyBalanceForSpecies(microbeSpecies, patch).FinalBalanceStationary;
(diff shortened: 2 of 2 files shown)


src/auto-evo/simulation/food_source/CompoundFoodSource.cs
```diff
@@ -26,7 +26,7 @@ public override float FitnessScore(Species species, SimulationCache simulationCa
 
         var compoundUseScore = EnergyGenerationScore(microbeSpecies, compound);
 
-        var energyCost = simulationCache.GetEnergyBalanceForSpecies(microbeSpecies, patch).FinalBalanceStationary;
+        var energyCost = simulationCache.GetEnergyBalanceForSpecies(microbeSpecies, patch).TotalConsumptionStationary;
 
         return compoundUseScore / energyCost;
     }
```
Maxonovien, line 32: That hasn't changed much, but it makes me realize that the fitness score seems not to be actually dependent on sessility nor speed. However, this is arguably very important when scavenging for compound food.

I could get that you'd deem this outside of the scope of the PR, but perhaps it could be worth adding a small comment on this by the way?

Maxonovien (approved):
Definitely fixing the bug it intended to, and little to say code-wise.

Still, it may be worth taking the opportunity to highlight further possible issues?

src/auto-evo/simulation/food_source/CompoundFoodSource.cs
adQuid: What if I make an issue instead? TODO comments are infamous for never being addressed, and if we make a comment for every scientific inaccuracy in auto evo you're not going to be able to see the code :)

adQuid: I made a new issue to cover this https://github.com/Revolutionary-Games/Thrive/issues/2918

Maxonovien: You are correct, although that's not only about scientific accuracy, but also matching experience from gameplay and auto-evo. I'm happy with this though.

Maxonovien:
Actually, I just realized you did not apply this change to Marine Snow. Is it intended?

adQuid:
> Actually, I just realized you did not apply this change to Marine Snow. Is it intended?

Yes, because https://github.com/Revolutionary-Games/Thrive/pull/2911 already includes that change, and I wanted to avoid conflicts

hhyyrylainen (approved):
I guess if this is better, this can be put in the game now to improve the auto-evo problems

PR #2919, opened 2021-12-08
Make compound clouds smaller but more common

adQuid:

Modifies the biome file for the vents to make clouds 10 times smaller, but 10 times more common. This leaves a single cloud more than big enough to fill up an early game microbe, but makes it very hard to get stuck in a compound desert at the start of the game.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


Maxonovien:
I think you only changed that in the Epipelagic, while this should then be applied to every biome, especially the Vents since this is where players actually start.

RacerBG:
I haven't looked at this PR yet but I think that this is a little too much. My opinion is biased because I don't like when the game is holding your hand while you are playing it but making the clouds 10 times smaller and 10 times more common is like finding a cloud on every 5 to 10 seconds. Also I think that the point which Maxonovien made is quite valid.

adQuid:
My bad with changing the wrong patch; I had to commit a different version of the branch than I tested due to the VM messing up. As for the spirit of the change, I brought this up before and we all seemed in agreement that the only point of the clouds in the early game was to get the player familiar with controls: https://forum.revolutionarygamesstudio.com/t/compound-clouds-and-the-early-game/835

I've found you tend to get a cloud after 15-30 seconds with this change on a bad day, which seems roughly right to me. This PR is lower testing priority than the others here right now though.

RacerBG:
> I've found you tend to get a cloud after 15-30 seconds with this change on a bad day, which seems roughly right to me. This PR is lower testing priority than the others here right now though.

If it is related only to the early game I'm fine with it and 15-30 seconds sounds reasonable.

RacerBG (commented):
It wasn't hard to find the clouds at all. Now they are definitely smaller and the older big blobs of clouds now looks like a combination of circle shapes.  I would like others to test it as well.

hhyyrylainen:
I guess there wasn't a consensus to rebalance the game like this in the end. I'll delete this branch now.

PR #2920, opened 2021-12-08
Implemented timeline tab

Kasterisk:

As mentioned in my 0.5.7 release planning forum post. Here's the first version of the timeline tab implementation based on @DonGororo's report tab concept art. Timeline is subdivided into separate time periods (time skips in the editor) where associated logged world events are listed under it. Events that are highlighted means it's related to the player or player's species.

In the future the underlined species names can be clicked and it will redirect the player to some kind of detailed page/section regarding that species. Possibly also need to make this work nicely with hhyyrylainen's auto-evo report overhaul plan.

Todo:
- [x] Save upgrade to fix timeline dictionary being null in older saves

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


RacerBG (changes requested):
I have a few things to say about the current implementation:
1. If the names of the species are meant to be underlined why they are not in some cases (like Virotpion Obagko and Ohetcanus Emanox)?
2. The order of the list should be reversed in my opinion (it's more important what happened currently because if you are in a situation where 1 000 000 000 years had passed you will scroll indefinitely to see the current events) or have a filter to be shown in either ascending or descending order.
3. I would like to see an option to expand and collapse the sections between every 100 Myr.

And a have a few questions:
1. What will happen in the future when the player can click on the names of the species for example? If they can open Encyclopedia pages or something similar shouldn't we include the same logic for the compounds and patches too?
2. Could we somehow hook up this features with the Statistics tab (the auto-evo report as it stands is starting to be very granular)?

Other than that I love the new icons, the highlighting color is nice and overall this feature adss to the feeling of progress to the player.

PS: Sorry for the long review.

hhyyrylainen:
> Could we somehow hook up this features with the Statistics tab (the auto-evo report as it stands is starting to be very granular)?

I think that should be talked about later and not solved in this PR. https://forum.revolutionarygamesstudio.com/t/explaining-the-auto-evo-prediction-to-the-player/836/2

RacerBG:
> > Could we somehow hook up this features with the Statistics tab (the auto-evo report as it stands is starting to be very granular)?
> 
> I think that should be talked about later and not solved in this PR. https://forum.revolutionarygamesstudio.com/t/explaining-the-auto-evo-prediction-to-the-player/836/2

No problems, I'm just writing what I observed and think about this PR. This doesn't mean that everything which I said must be implemented or included in a single PR. :smiley:

Kasterisk:
> If the names of the species are meant to be underlined why they are not in some cases (like Virotpion Obagko and Ohetcanus Emanox)?

Oh yeah, looks like I forgot to underline those.

> The order of the list should be reversed in my opinion (it's more important what happened currently because if you are in a situation where 1 000 000 000 years had passed you will scroll indefinitely to see the current events) or have a filter to be shown in either ascending or descending order.

Currently an autoscroll move down the scrollbar into the beginning of the last timeskip. And from there, the player can manually scroll down in chronological order, I think this is the most intuitive. A sort button would be nice though, you mean sorting the time periods (the 300 Myr, etc labels) and not the events under it right?

> What will happen in the future when the player can click on the names of the species for example? If they can open Encyclopedia pages or something similar shouldn't we include the same logic for the compounds and patches too?

That makes sense, I think this needs to be taken into consideration once we implement this functionality. But yeah the plan is for them be able to open up the associated page in Thrivepedia or maybe statistics menu but this probably needs more thought into how it will be done.

RacerBG:
> > The order of the list should be reversed in my opinion (it's more important what happened currently because if you are in a situation where 1 000 000 000 years had passed you will scroll indefinitely to see the current events) or have a filter to be shown in either ascending or descending order.
> 
> A sort button would be nice though, you mean sorting the time periods (the 300 Myr, etc labels) and not the events under it right?

Yes, exactly.

hhyyrylainen:
> Currently an autoscroll move down the scrollbar into the beginning of the last timeskip. And from there, the player can manually scroll down in chronological order, I think this is the most intuitive. A sort button would be nice though, you mean sorting the time periods (the 300 Myr, etc labels) and not the events under it right?

That sounds good.

> That makes sense, I think this needs to be taken into consideration once we implement this functionality. But yeah the plan is for them be able to open up the associated page in Thrivepedia or maybe statistics menu but this probably needs more thought put into how it will be done.

The link clicked handlers should be able to work in our custom bbcode to allow a link like "thrivepedia://species/123" to open a species info or for example a compound.

Properties/AssemblyInfo.cs
```diff
@@ -19,9 +19,9 @@
 // NOTE: the version info is read by a regex in a script so don't add complicated things to this file
 // NOTE: When changing the version number you need to also change "export_presets.cfg"
 
-[assembly: AssemblyVersion("0.5.6.1")]
+[assembly: AssemblyVersion("0.5.7.0")]
 
-[assembly: AssemblyInformationalVersion("")]
+[assembly: AssemblyInformationalVersion("alpha")]
```
hhyyrylainen, line 24: doesn't this need to be "-alpha"? Or am I misremembering?

locale/en.po
```diff
@@ -911,59 +911,88 @@ msgstr "Starting"
...
 
-#: ../src/auto-evo/RunResults.cs:502
+#: ../src/auto-evo/RunResults.cs:603
 msgid "RUN_RESULT_NICHE_FILL"
 msgstr "emerged to fill a niche"
 
-#: ../src/auto-evo/RunResults.cs:505
+#: ../src/auto-evo/RunResults.cs:604
+msgid "TIMELINE_NICHE_FILL"
+msgstr "[u]{0}[/u] has split into [u]{1}[/u] to fill a niche"
```
hhyyrylainen, line 936: I think this wording is a bit bad. Originally for my auto evo results I used the wording that the new species split off from another species to make the wording better. For here something that says along the lines of "{1} split off as a new species to fill a niche" or something.

locale/en.po
```diff
@@ -911,59 +911,88 @@ msgstr "Starting"
...
 msgstr "went extinct from the planet"
 
-#: ../src/auto-evo/simulation/food_source/ChunkFoodSource.cs:64
+#: ../src/auto-evo/RunResults.cs:819
+msgid "TIMELINE_SPECIES_EXTINCT"
+msgstr "[u]{0}[/u] has gone extinct!"
+
+#: ../src/auto-evo/RunResults.cs:830
+msgid "TIMELINE_SPECIES_POPULATION_INCREASE"
+msgstr "[u]{0}[/u] populations have increased to {1}"
```
hhyyrylainen, line 989: I think using "population" without the plural sounds better here. As populations has the implication that they are like separate, multiple different populations.

locale/en.po
```diff
@@ -1408,12 +1437,29 @@ msgstr "Art by {0}"
...
 #: ../src/microbe_stage/ProcessPanel.tscn:72
-#: ../src/microbe_stage/editor/MicrobeEditor.tscn:4001
+#: ../src/microbe_stage/editor/MicrobeEditor.tscn:4072
 #: ../src/modding/ModManager.tscn:972
 #: ../src/tutorial/microbe_editor/MicrobeEditorTutorialGUI.tscn:389
 msgid "CLOSE"
 msgstr "Close"
 
+#: ../src/general/IWorldEffect.cs:77
+msgid "COMPOUND_CONCENTRATIONS_DRASTICALLY_DROPPED"
```
hhyyrylainen, line 1447: When does this trigger? I don't think this is clear at all, unless literally all compound concentrations dropped by like 20%, which I don't think is ever able to happen...

Properties/AssemblyInfo.cs
Kasterisk: Apparently it is.

locale/en.po
Kasterisk: Usually happens in freebuild, initial huge population and plentiful resources, thus a big drop in compounds concentrations. I leave this here in case such scenario happens in normal game.

RacerBG: > Usually happens in freebuild, initial huge population and plentiful resources, thus a big drop in compounds concentrations. I leave this here in case such scenario happens in normal game.

That's correct. It happens in Freebuild only.

RacerBG (commented):
The autoscroll is working as expected, the underlined names are fixed. The thing which still bothers me is the amount of text shown on the Timeline *. It's there for a reason but as I said earlier some sort of Filter function or a Sort function would be nice. Feel free to skip this for the current PR though. I'm ready to approve it if no other concerns are raised.

* Basically the content of the Timeline is overlapping with 90% of the Auto-Evo text and this **currently** creates a duplicate functionality. However, this could be improved in the future so it's not a big showstopper for me.

src/auto-evo/RunResults.cs
```diff
@@ -508,23 +508,24 @@ public List<Species> GetNewSpecies()
...
-            GD.Print(MakeSummary(previousPopulations));
+            GD.Print(MakeSummary(world, previousPopulations));
 
             GD.Print("End of results summary");
         }
 
         /// <summary>
         ///   Makes summary text
         /// </summary>
+        /// <param name="world">The current global game data</param>
```
hhyyrylainen: These need to be updated, because this is not purely a summary making code anymore, this is actually now impacting the state of the GameWorld object.

src/auto-evo/RunResults.cs
```diff
@@ -675,6 +686,20 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+                    if (numberOfPatches < 2)
+                    {
+                        world.LogWorldEvent(new LocalizedString("TIMELINE_SPECIES_MIGRATED_TO",
+                                entry.Species.FormattedName, TranslationServer.Translate(
+                                    entry.SpreadToPatches[0].To.Name)), entry.Species.PlayerSpecies,
+                            "res://assets/textures/gui/bevel/newSpecies.png");
+                    }
+                    else
+                    {
+                        world.LogWorldEvent(new LocalizedString("TIMELINE_SPECIES_MIGRATED_TO_MULTIPLE",
```
hhyyrylainen: Wouldn't it be better to numerically show 2 of the migrations at least? That would have a higher chance of always showing the info, but won't take too much space.

Also the term "migration" perhaps needs tweaking as species can spread to a patch they are already in. Then again I suppose "spread to" would be even more misleading...

src/auto-evo/RunResults.cs
```diff
@@ -782,11 +807,37 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+                var globalPopulation = GetGlobalPopulation(entry.Species, resolveMigrations, resolveSplits);
+                var previousTotalPopulations = previousPopulations.GetSpeciesPopulation(entry.Species);
+
+                if (globalPopulation <= 0)
                 {
                     builder.Append(' ');
                     builder.Append(new LocalizedString("WENT_EXTINCT_FROM_PLANET"));
                     builder.Append('\n');
+
+                    world.LogWorldEvent(new LocalizedString("TIMELINE_SPECIES_EXTINCT",
```
hhyyrylainen: This is now also vulnerable to this issue: https://github.com/Revolutionary-Games/Thrive/issues/2958

src/general/GameWorld.cs
```diff
@@ -26,6 +26,15 @@ public class GameWorld
     [JsonProperty]
     private Dictionary<uint, Species> worldSpecies = new Dictionary<uint, Species>();
 
+    [JsonProperty]
+    private Dictionary<double, List<WorldEventDescription>> worldTimeline = new();
+
+    /// <summary>
+    ///   Another placement for worldTimeline keys to allow removing old dictionary keys
```
hhyyrylainen:
```suggestion
    ///   Another place for worldTimeline keys to allow removing old dictionary keys
```

src/general/GameWorld.cs
```diff
@@ -310,11 +319,53 @@ public Species GetSpecies(uint id)
...
+        {
+            EventDescription = description,
+            IconPath = iconPath,
+            Highlighted = highlight,
+        });
+
+        worldTimelineKeys.Enqueue(TotalPassedTime);
+    }
+
+    public IReadOnlyDictionary<double, List<WorldEventDescription>> GetTimeline()
```
hhyyrylainen: Instead of a getter, make a read only property to read the timeline through it as this type.

src/general/IWorldEffect.cs
```diff
@@ -36,19 +38,51 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+
+        if (finalDensity > 0)
+        {
+            var initialGlucose = Math.Round(initialDensity * totalAmount + totalChunkAmount, 3);
+            var finalGlucose = Math.Round(finalDensity * totalAmount + totalChunkAmount, 3);
+            var percentage = Math.Round((initialGlucose - finalGlucose) / initialGlucose * 100, 1);
+
+            if (percentage >= 20)
+            {
+                targetWorld.LogWorldEvent(new LocalizedString("COMPOUND_CONCENTRATIONS_DRASTICALLY_DROPPED",
```
hhyyrylainen: Ah I see you used it here. I very strongly think that this message must be changed to be consistent with the glucose reduction message, as that's the compound that reduces.

src/microbe_stage/PatchMap.cs
```diff
@@ -201,6 +201,18 @@ public void UpdateGlobalPopulations()
         }
     }
 
+    public long GetSpeciesPopulation(Species species)
```
hhyyrylainen:
```suggestion
    public long GetSpeciesGlobalPopulation(Species species)
```
just to be extra clear what this returns and to not have a danger of mixing up the patch map and individual patch objects.

src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -1009,6 +1045,95 @@ public void UpdateReportTabStatistics(Patch patch)
         OnPhysicalConditionsChartLegendPressed("temperature");
     }
 
+    public void UpdateTimeline()
+    {
+        var mainContainer = timelineSubtab.GetNode<Control>(
+            "MarginContainer/ScrollContainer/MarginContainer/VBoxContainer");
```
hhyyrylainen: Why not use an Export path here? This seems a very long path that is bound to get changed due to editor GUI changes at some point.

locale/en.po
hhyyrylainen: Now after reading everything, turns out that what I really should have said that this should be "GLUCOSE_CONCENTRATIONS_DRASTICALLY_DROPPED" because it's only related to glucose not all compounds.

Kasterisk: It's not all compounds which is why it's not plural in the translation key. In the future other compounds might also possible to get reduced significantly hence why I made the naming more general.

src/auto-evo/RunResults.cs
Kasterisk: Migrate in its broader sense can also mean relocate/move, in this case it's the species' sub-population to another place its kind already inhabits and not necessarily mean the entire species. Perhaps for this the wording could be "All or part of (species)'s population has migrated to (patch)"?

src/auto-evo/RunResults.cs
Kasterisk: Added TODO.

locale/en.po
hhyyrylainen: So this relies on the glucose icon being next to this line? And that clarifies that this is about glucose.
I don't imagine that many other compounds follow such a nice pattern, I think most other compounds would go up and down all over the patch map without a clear direction.

src/auto-evo/RunResults.cs
hhyyrylainen: All the population can't be part of a spread action in auto-evo, so that's irrelevant to include in the wording of the text.

Kasterisk:
> The thing which still bothers me is the amount of text shown on the Timeline *. It's there for a reason but as I said earlier some sort of Filter function or a Sort function would be nice. Feel free to skip this for the current PR though. I'm ready to approve it if no other concerns are raised.

Keep in mind that this is still a basic implementation of the tab. I'll try to iteratively improve upon it as time goes by. I'll note this filter and sort function in a new issue.

src/auto-evo/RunResults.cs
Kasterisk: > Wouldn't it be better to numerically show 2 of the migrations at least? That would have a higher chance of always showing the info, but won't take too much space.

Btw, I never see species that migrated into multiple patches in a single auto-evo run so this may not even currently happen in the game at all.

hhyyrylainen: It doesn't currently (out of a few candidate migrations only the most beneficial is selected), but it wouldn't be very fair to keep AI species from doing that once the player can command their own population around.

locale/en.po
Kasterisk: Changed this to only referring to glucose.

RacerBG (approved):

src/general/GameWorld.cs
```diff
@@ -26,6 +26,15 @@ public class GameWorld
...
     private Dictionary<uint, Species> worldSpecies = new Dictionary<uint, Species>();
 
+    [JsonProperty]
+    private Dictionary<double, List<WorldEventDescription>> worldTimeline = new();
+
+    /// <summary>
+    ///   Another place for worldTimeline keys to allow removing old dictionary keys
+    ///   through Queue's functionality
+    /// </summary>
+    private Queue<double> worldTimelineKeys = new();
```
hhyyrylainen: Does this need JSON saving? Also if the timeline uses just numbers to order events, can't we just sort the keys and take from that sorted list to find old items to be removed?

Kasterisk: After some further thought I think it makes sense to store this timeline events in individual patches. That way it will make it easier to support a patchwide-only filter.

Kasterisk:
I think I'll do a major rework of the current implementation, in the meantime will make this a draft.

Kasterisk:
Alright did a first commit of that, timeline now defaults to patch-specific events. Will have to add support for global events instead.

src/general/GameWorld.cs
Kasterisk: In the end having GameWorld store its own logs seems the best way and your sort keys and remove solution works great.

Kasterisk:
I have now added support to filter timeline events by globally or locally.

RacerBG (approved):
It works really well. I will approve it but a task for a future PR could be to implement stylistic icons for the "G" and "L" buttons. I don't have an opinion about their position though, I would love to read the opinions of more people on that topic.

Note: When switching from a Timeline without a scrollbar and one with a scrollbar the horizontal alignment of the buttons changes. This could be fixed but it's not a game breaker.

Kasterisk:
> Note: When switching from a Timeline without a scrollbar and one with a scrollbar the horizontal alignment of the buttons changes. This could be fixed but it's not a game breaker.

Yes this is intended, the buttons are displaced slightly to the left to make room for the scrollbar and returns when the scrollbar is not visible, otherwise they overlap. I guess alternatively their position could be preserved while part of the scrollbar length is cut off at the top but I don't think that's currently possible.

hhyyrylainen:
What if the buttons are just always positioned so that there's room for the scrollbar? I don't think that's too much of a problem but if that is easy to do, it might be nicer than the buttons jumping around.

Kasterisk:
I guess the right and top margin can be increased, but that will start taking up unnecessary space.

hhyyrylainen:
Isn't it pretty common to have buttons that aren't exactly in the corner anyway? So stylewise it wouldn't be that bad of a decision, right?

Kasterisk:
Doesn't look too bad I guess...

src/auto-evo/AutoEvoRun.cs
```diff
@@ -322,13 +322,14 @@ protected virtual void GatherInfo(Queue<IRunStep> steps)
                 }
                 else
                 {
-                    steps.Enqueue(new FindBestMutation(autoEvoConfiguration, map, speciesEntry.Key,
+                    steps.Enqueue(new FindBestMutation(autoEvoConfiguration, map, speciesEntry.Key, entry.Value,
```
hhyyrylainen: Why is the patch passed through here? This doesn't make sense as the patch the species first appears in is arbitrary (I mean the first patch that appears in the list of patches that a species is in, is not in any order that makes sense to only work on the one patch), it should not be relied on for anything.

src/general/GameEventDescription.cs
```diff
@@ -0,0 +1,9 @@
+﻿/// <summary>
+///   A text-based description of what has happened in a game environment. Decorated with an icon if there's any.
+/// </summary>
+public class GameEventDescription
+{
+    public LocalizedString Description;
+    public string IconPath;
+    public bool Highlighted;
```
hhyyrylainen: I think these fields should all be documented, also changing these all to properties would be nice (maybe with private sets, so that no one edits a game event after it has been created).

src/gui_common/GUICommon.cs
```diff
@@ -89,6 +89,23 @@ public static void SmoothlyUpdateBar(TextureProgress bar, float target, float de
         bar.Value = MathUtils.Lerp((float)bar.Value, target, weight, 1.0f / (float)bar.MaxValue);
     }
 
+    /// <summary>
+    ///   Loads a Texture from predefined GUI asset texture folder path.
+    /// </summary>
+    public static Texture LoadGuiAsset(string file)
```
hhyyrylainen: This loads a texture, so this should be named as loading a texture instead of a generic "asset"

src/auto-evo/AutoEvoRun.cs
Kasterisk: To log into that particular patch that some species has mutated, and only for that reason. Is there an alternative way to do this by any chance? I don't really understand well this area of the codebase.

hhyyrylainen: To do it properly you would have to add a step that runs at the end of auto-evo, and it needs to do a loop of each species checking if it has mutations and each patch they have population in and add the notice that they had a mutation there. That's the only way to do this properly as mutations don't happen in one patch. Also one potentially tricky thing is that you'd also need to handle the species split checking there as the species may have split and some patches no longer have their population.

So in summary I think it's fundamentally flawed approach to have this here, so I suggest you remove the patch parameters you added to the auto-evo steps and instead add an overall step at the end that goes through the results to create mutation events for all species in patches they are in.

src/general/GameEventDescription.cs
Kasterisk: `readonly` modifier is maybe good for this.

hhyyrylainen: I kinda think that properties (get only properties would work here) make more sense from an OOP perspective. I'm still not really clear on best C# practices but it seems that public properties for a general class are really frowned upon. That's why I think this data class should use properties.

Kasterisk: Sure, but imo they're just kinda the same.

hhyyrylainen: I think this is a similar thing as the `struct` vs `class` distinction in C++, really the only thing different is the default access modifier of members, but selecting class versus struct actually conveys information to the reader as styleguides say to use them for different purposes. Struct for plain data and class for actual objects that hide their state in private variables and provide methods to access their functionality.

src/auto-evo/AutoEvoRun.cs
Kasterisk: It may actually be too early if this overall step is ran in `RunStage.Ended` as it happens before GameWorld's time skip and the events logging need to happen after that.

hhyyrylainen: Yeah that approach of having a separate method to record the events from RunResults sounds fine.

Also skipping including mutation events in the timeline is probably good as probably most species have mutations pretty often.

Kasterisk: Alright, sounds good.

locale/en.po
```diff
@@ -3084,274 +3132,290 @@ msgid "COMPOUND_BALANCE_TITLE"
...
-#: ../src/microbe_stage/editor/MicrobeEditor.tscn:2590
+#: ../src/microbe_stage/editor/MicrobeEditor.tscn:2598
 msgid "EXTERNAL_EFFECTS"
 msgstr "External effects:"
 
-#: ../src/microbe_stage/editor/MicrobeEditor.tscn:2673
-#: ../src/microbe_stage/editor/MicrobeEditor.tscn:2701
+#: ../src/microbe_stage/editor/MicrobeEditor.tscn:2696
+msgid "TIMELINE_GLOBAL_FILTER_TOOLTIP"
+msgstr "Filter events by globally"
```
hhyyrylainen, line 3354: I think this wording is a bit strange, something like "Show/hide global events" should be used instead. The same goes for the local events as well.

locale/en.po
```diff
@@ -3084,274 +3132,290 @@ msgid "COMPOUND_BALANCE_TITLE"
...
+
+#: ../src/microbe_stage/editor/MicrobeEditor.tscn:2702
+msgid "GLOBAL_INITIAL"
```
hhyyrylainen, line 3357: `GLOBAL_INITIAL_LETTER`

src/general/GameEventDescription.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    public LocalizedString Description { get; private set; }
+
+    /// <summary>
+    ///   The resource path to the associated icon
+    /// </summary>
+    public string IconPath { get; private set; }
+
+    /// <summary>
+    ///   If true, this event will be highlighted in a timeline UI
```
hhyyrylainen:
```suggestion
    ///   If true, this event will be highlighted in the timeline UI
```

src/general/GameWorld.cs
```diff
@@ -310,6 +318,36 @@ public Species GetSpecies(uint id)
         return worldSpecies[id];
     }
 
+    /// <summary>
+    ///   Stores a description of a global event into the game world record.
```
hhyyrylainen:
```suggestion
    ///   Stores a description of a global event into the game world records.
```

src/general/GameWorld.cs
```diff
@@ -310,6 +318,36 @@ public Species GetSpecies(uint id)
...
+    /// </summary>
+    /// <param name="description">The event's description</param>
+    /// <param name="highlight">If true, the event will be highlighted in a timeline UI</param>
```
hhyyrylainen:
```suggestion
    /// <param name="highlight">If true, the event will be highlighted in the timeline UI</param>
```

src/general/GameWorld.cs
```diff
@@ -310,6 +318,36 @@ public Species GetSpecies(uint id)
...
+                oldestKey = Math.Min(entry.Key, oldestKey);
+
+            eventsLog.Remove(oldestKey);
+        }
+
+        if (!eventsLog.ContainsKey(TotalPassedTime))
+            eventsLog.Add(TotalPassedTime, new List<GameEventDescription>());
+
+        // Event already logged in timeline
+        if (eventsLog[TotalPassedTime].Any(entry => entry.Description.ToString() == description.ToString()))
```
hhyyrylainen: I think LocalizedString might need to have a more complex comparison operator, because this won't match if the user changes language...

src/general/GameWorld.cs
```diff
@@ -310,6 +318,36 @@ public Species GetSpecies(uint id)
...
+    /// <param name="highlight">If true, the event will be highlighted in a timeline UI</param>
+    /// <param name="iconPath">Resource path to the icon of the event</param>
+    public void LogEvent(LocalizedString description, bool highlight = false, string iconPath = null)
+    {
+        if (eventsLog.Count > Constants.GLOBAL_EVENT_LOG_CAP)
+        {
+            var oldestKey = double.MaxValue;
+
+            foreach (var entry in eventsLog)
+                oldestKey = Math.Min(entry.Key, oldestKey);
```
hhyyrylainen: This looks a bit suspicious to me. I'd prefer either to do an entirely LINQ operation, I think eventsLog.Keys.Min() should work?, or code this "normally" with an if to check if the oldestKey variable needs to be updated or not.

src/general/IWorldEffect.cs
```diff
@@ -36,20 +38,63 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+
+                var initialGlucose = Math.Round(initialDensity * glucoseValue.Amount +
+                    patch.GetTotalChunkCompoundAmount(glucose), 3);
+                var finalGlucose = Math.Round(glucoseValue.Density * glucoseValue.Amount +
+                    patch.GetTotalChunkCompoundAmount(glucose), 3);
+
+                var localReduction = Math.Round((initialGlucose - finalGlucose) / initialGlucose * 100, 1);
+
+                patch.LogEvent(new LocalizedString("COMPOUND_CONCENTRATIONS_DECREASED",
+                        glucose.Name, string.Format(CultureInfo.CurrentCulture, TranslationServer.Translate(
```
hhyyrylainen: Format should not be used here, instead a nested LocalizedString needs to be used, otherwise this substring wouldn't react to translations. I've used this approach before in Thrive, I'm not exactly sure but it might be in the auto-evo results regarding species splitting.

src/general/IWorldEffect.cs
```diff
@@ -36,20 +38,63 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
-                    patch.Biome.Compounds[glucose] = glucoseValue;
-                }
+                var initialDensity = glucoseValue.Density;
+
+                glucoseValue.Density = Math.Max(glucoseValue.Density * Constants.GLUCOSE_REDUCTION_RATE,
+                    Constants.GLUCOSE_MIN);
+                patch.Biome.Compounds[glucose] = glucoseValue;
+
+                var initialGlucose = Math.Round(initialDensity * glucoseValue.Amount +
+                    patch.GetTotalChunkCompoundAmount(glucose), 3);
```
hhyyrylainen: Couldn't this be calculated on line 61 so that the initialDensity doesn't need to be in a variable?

src/general/IWorldEffect.cs
```diff
@@ -36,20 +38,63 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
+
+        if (globalReduction >= 50)
+        {
+            targetWorld.LogEvent(new LocalizedString("GLUCOSE_CONCENTRATIONS_DRASTICALLY_DROPPED"),
+                false, "glucoseDown.png");
+        }
+        else
+        {
+            targetWorld.LogEvent(new LocalizedString("COMPOUND_CONCENTRATIONS_DECREASED",
+                    glucose.Name, string.Format(CultureInfo.CurrentCulture, TranslationServer.Translate(
```
hhyyrylainen: Same thing here with the LocalizedString as a sub value.

src/gui_common/GUICommon.cs
```diff
@@ -89,6 +89,23 @@ public static void SmoothlyUpdateBar(TextureProgress bar, float target, float de
...
+        var assumedPath = PathUtils.Join(Constants.ASSETS_GUI_BEVEL_FOLDER, file);
+
+        if (ResourceLoader.Exists(assumedPath, "Texture"))
+            return GD.Load<Texture>(assumedPath);
+
+        // Fail-safe if file itself is the absolute path
+        if (ResourceLoader.Exists(file, "Texture"))
+            return GD.Load<Texture>(file);
+
+        return null;
```
hhyyrylainen: I think attempting to load anyway would be better than returning null, that way Godot can log the missing texture file.

src/microbe_stage/Patch.cs
```diff
@@ -70,6 +71,11 @@ public double TimePeriod
     [JsonIgnore]
     public BiomeConditions Biome => currentSnapshot.Biome;
 
+    /// <summary>
+    ///   All logged events occured specific to this patch.
```
hhyyrylainen:
```suggestion
    ///   Logged events that specifically occurred in this patch.
```

src/microbe_stage/Patch.cs
```diff
@@ -232,8 +238,28 @@ public void RecordConditions()
...
-        history.AddToFront(conditions);
+        var snapshot = (PatchSnapshot)currentSnapshot.Clone();
+        history.AddToFront(snapshot);
+    }
+
+    /// <summary>
+    ///   Logs description of an event into the patch's history.
+    /// </summary>
+    /// <param name="description">The event's description</param>
+    /// <param name="highlight">If true, the event will be highlighted in a timeline UI</param>
```
hhyyrylainen:
```suggestion
    /// <param name="highlight">If true, the event will be highlighted in the timeline UI</param>
```

src/microbe_stage/Patch.cs
```diff
@@ -232,8 +238,28 @@ public void RecordConditions()
...
+    /// <param name="iconPath">Resource path to the icon of the event</param>
+    public void LogEvent(LocalizedString description, bool highlight = false, string iconPath = null)
+    {
+        // Event already logged in timeline
+        if (currentSnapshot.EventsLog.Any(entry => entry.Description.ToString() == description.ToString()))
```
hhyyrylainen: Same thing with comparing the localized strings here as in the GameWorld.

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -2521,7 +2525,8 @@ private void ApplyAutoEvoResults()
         // for reference in patch history (e.g. displaying it as zero on the species population chart)
         foreach (var entry in CurrentGame.GameWorld.Map.Patches)
         {
-            entry.Value.RecordConditions();
+            entry.Value.RecordSnapshot();
+            entry.Value.ClearLoggedEvents();
```
hhyyrylainen: Would it make sense to make ClearLoggedEvents private and make RecordSnapshot call it? I think it would be less dangerous to combine the recording of the events with clearing the events list, than leave this method public.

src/saving/ISaveUpgradeStep.cs
```diff
@@ -161,6 +161,19 @@ private void UpgradeBehaviouralValues(JProperty property, JEnumerable<JProperty>
         }
     }
 
+    internal class UpgradeStep0561To057 : BaseRecursiveJSONWalkerStep
+    {
+        protected override string VersionAfter => "0.5.7.0-rc1";
+
+        protected override void CheckAndUpdateProperty(JProperty property)
+        {
+            if (property.Name.Contains("GameWorld"))
```
hhyyrylainen: Do we need to also upgrade patches or just the GameWorld?

src/microbe_stage/editor/TimelineTab.cs
```diff
@@ -0,0 +1,249 @@
...
+    {
+        get => eventFilter;
+        set
+        {
+            eventFilter = value;
+            ApplyEventsFilter();
+        }
+    }
+
+    public void Init(MicrobeEditor editor, PatchMapDrawer drawer)
```
hhyyrylainen: Passing this these objects is a bit more coupling than I'd like. Couldn't these be passed in the UpdateTimeline method?

src/microbe_stage/editor/TimelineTab.cs
```diff
@@ -0,0 +1,249 @@
...
+        GUICommon.Instance.PlayButtonPressSound();
+
+        EventFilter = (Filters)index;
+    }
+
+    private class TimelineSection : VBoxContainer
+    {
+        private Control headerContainer;
+
+        private PackedScene customRtlScene;
```
hhyyrylainen: `Rtl` should not be abbreviated.

src/microbe_stage/editor/TimelineTab.cs
```diff
@@ -0,0 +1,249 @@
...
+        private StyleBoxTexture eventHighlightStylebox;
+
+        private (double TimePeriod, List<GameEventDescription> Events) data;
```
hhyyrylainen: Could this be `readonly`?

src/microbe_stage/editor/TimelineTab.cs
```diff
@@ -0,0 +1,249 @@
...
+            this.data = data;
+        }
+
+        public Rect2 HeaderGlobalRect => headerContainer.GetGlobalRect();
+
+        public override void _Ready()
+        {
+            customRtlScene = GD.Load<PackedScene>("res://src/gui_common/CustomRichTextLabel.tscn");
+            eventHighlightStylebox = GD.Load<StyleBoxTexture>(
+                "res://src/microbe_stage/editor/TimelineEventHighlight.tres");
```
hhyyrylainen: I think it'd be slightly nicer if these were loaded by the Timeline object so that we don't need to load these 100 times (even though Godot probably caches these, I'd feel better not loading these a ton).

src/auto-evo/RunResults.cs
```diff
@@ -496,6 +496,11 @@ public List<Species> GetNewSpecies()
             return newSpecies;
         }
 
+        public NewSpeciesType? GetNewSpeciesResult(Species species)
```
hhyyrylainen: These new methods you added don't seem to be used anywhere...

src/auto-evo/RunResults.cs
```diff
@@ -518,6 +523,16 @@ public List<Species> GetNewSpecies()
             return results[species].EnergyResults;
         }
 
+        public Species GetMutationResults(Species species)
+        {
+            return results[species].MutatedProperties;
+        }
+
+        public (Species SplitOff, Species SplitFrom, List<Patch> SplitOffPatches) GetSplitResults(Species species)
```
hhyyrylainen: Here's the other two

src/auto-evo/RunResults.cs
```diff
@@ -811,6 +826,144 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+                            }
+                        }
+                    }
+
+                    if (globalPopulation <= 0)
+                    {
+                        // TODO: see https://github.com/Revolutionary-Games/Thrive/issues/2958
+                        LogEventGloballyAndLocally(world, patch, new LocalizedString(
+                                "TIMELINE_SPECIES_EXTINCT", species.FormattedName),
+                            species.PlayerSpecies, "extinction.png");
```
hhyyrylainen: using a continue here would reduce nesting for quite a bit of code that's in the else.

src/auto-evo/RunResults.cs
```diff
@@ -811,6 +826,144 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+                    }
+                    else
+                    {
+                        if (hasPopulationBeforeMigration)
+                        {
+                            if (finalPatchPopulation > previousPatchPopulation)
```
hhyyrylainen: What if the population is exactly the same? I think we wouldn't want to accidentally add the population decrease event in that case.

src/auto-evo/RunResults.cs
```diff
@@ -811,6 +826,144 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+                            {
+                                patch.LogEvent(new LocalizedString("TIMELINE_SPECIES_POPULATION_INCREASE",
+                                        species.FormattedName, ((double)finalPatchPopulation).FormatNumber()),
```
hhyyrylainen: Why are the population numbers as doubles here? They can't be fractional so why don't we just include the raw number?

src/auto-evo/RunResults.cs
```diff
@@ -811,6 +826,144 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+                                species.PlayerSpecies, "extinctionLocal.png");
+                        }
+
+                        if (globalPopulation > previousGlobalPopulation)
+                        {
+                            world.LogEvent(new LocalizedString("TIMELINE_SPECIES_POPULATION_INCREASE",
+                                    species.FormattedName, ((double)globalPopulation).FormatNumber()),
+                                species.PlayerSpecies, "popUp.png");
+                        }
+                        else
```
hhyyrylainen: Populations being equal should probably also be checked here.

src/auto-evo/RunResults.cs
```diff
@@ -811,6 +826,144 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+                                species.PlayerSpecies, "popDown.png");
+                        }
+                    }
+                }
+
+                foreach (var migration in GetMigrationsTo(patch))
+                {
+                    // Log to destination patch
+                    patch.LogEvent(new LocalizedString("TIMELINE_SPECIES_MIGRATED_FROM", migration.Key.FormattedName,
+                            TranslationServer.Translate(migration.Value.From.Name)),
```
hhyyrylainen: This should also use a nested LocalizedString so that this reacts to language change properly.

src/auto-evo/RunResults.cs
```diff
@@ -811,6 +826,144 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+                    // Log to game world
+                    world.LogEvent(new LocalizedString("GLOBAL_TIMELINE_SPECIES_MIGRATED_TO",
+                            migration.Key.FormattedName,
+                            TranslationServer.Translate(migration.Value.To.Name),
+                            TranslationServer.Translate(migration.Value.From.Name)),
+                        migration.Key.PlayerSpecies, "newSpecies.png");
+
+                    // Log to origin patch
+                    migration.Value.From.LogEvent(new LocalizedString("TIMELINE_SPECIES_MIGRATED_TO",
+                            migration.Key.FormattedName, TranslationServer.Translate(migration.Value.To.Name)),
```
hhyyrylainen: A few other places in these lines that have the same issue as my comment above.

src/auto-evo/RunResults.cs
```diff
@@ -811,6 +826,144 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+                        migration.Key.PlayerSpecies, "newSpecies.png");
+                }
+
+                foreach (var newSpeciesEntry in newSpecies)
+                {
+                    GetSpeciesPopulationsByPatch(newSpeciesEntry, true, true).TryGetValue(patch, out var population);
+
+                    var newSpeciesResult = GetNewSpeciesResult(newSpeciesEntry);
```
hhyyrylainen: Isn't the new species result data available through the newSpeciesEntry?

src/auto-evo/RunResults.cs
```diff
@@ -811,6 +826,144 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+                                break;
+                        }
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        ///   Logs an event description into game world and a patch. Use this if the event description in question
+        ///   is an exact same.
```
hhyyrylainen:
```suggestion
        ///   is exactly the same.
```

hhyyrylainen (changes requested):
Okay, I've now finished taking a look at this entire PR. I didn't notice any glaring design problems anymore, but I did notice a lot of small translation issues and code that I'd like to see cleaned up a bit. I *think* it is still possible to get this finished by tomorrow morning (my time) so that this can make it into the RC.

src/auto-evo/RunResults.cs
Kasterisk: It is used, line 928. https://github.com/Revolutionary-Games/Thrive/blob/4eb6e8fd5704d1098d8ba8317ea6fc8a9b510763/src/auto-evo/RunResults.cs#L928 Originally for accessing these results from a IRunStep object, which no longer the case but thought these public methods will also be good for future uses.

src/auto-evo/RunResults.cs
Kasterisk: Related to https://github.com/Revolutionary-Games/Thrive/pull/2920#discussion_r794342880. The latter is in a few lines down: https://github.com/Revolutionary-Games/Thrive/blob/4eb6e8fd5704d1098d8ba8317ea6fc8a9b510763/src/auto-evo/RunResults.cs#L932

src/auto-evo/RunResults.cs
Kasterisk: Have no clue how to make `FormatNumber` takes in any numeric data types. Any suggestions?

src/auto-evo/RunResults.cs
hhyyrylainen: I did see that later on. Is there a reason why the species entry doesn't already have the required information?
I think these expose a bit too much of the internal workings to be public.

src/auto-evo/RunResults.cs
hhyyrylainen: Why don't these actually use the full species result objects like done in other methods in this class? That's why these "get this one specific bit of the results" aren't needed for everything.

src/auto-evo/RunResults.cs
hhyyrylainen: Does it need to be formatted? In the auto-evo results the longs are just shown as-is.

src/auto-evo/RunResults.cs
Kasterisk: It only provides `Species` data and not the species type result that we want.

hhyyrylainen: I think it would be better to have a method to go from Species to the SpeciesResult object, that way it's general and helpful to (future) code in this class.

src/saving/ISaveUpgradeStep.cs
Kasterisk: Just the GameWorld apparently, the patches seem to have `eventsLog` initialized already on save upgrade, as evidenced by upgrading and then loading an older save which will show a "No events recorded" text under passed time periods on the local timeline.

Relevant code line: https://github.com/Revolutionary-Games/Thrive/blob/4eb6e8fd5704d1098d8ba8317ea6fc8a9b510763/src/microbe_stage/editor/TimelineTab.cs#L233

hhyyrylainen: I guess it's fine then if it is intended that the data might be missing.

src/auto-evo/RunResults.cs
Kasterisk: Yes, for easier digestible numbers, otherwise it can be hard to understand the long string of numbers added on top of columns of events.

hhyyrylainen: I think there's no other way than to add an overload of the format method for longs. I had to do that for one place where I couldn't have a general "any number" generic. I think that's a C# limitation we just need to work around.

src/microbe_stage/editor/MicrobeEditor.cs
Kasterisk: I guess it would, I added a public clearing method in case the programmer wants an extra option to whether to clear the event records on snapshot recording or not though that should be possible by passing in a bool parameter nonetheless...

src/general/GameWorld.cs
Kasterisk: A multi-level condition check would work for the comparison, mayhaps, so like 1. check if actual string match 2. check if the internal string match 3. check if reference match.

hhyyrylainen: LocalizedString should get specialized Equals method that first compares the base text and then uses Equals on all the values it has. That way it should compare equally if things are the same, not considering when the text was saved.
That would be the proper way. But actually now that I think about it converting to string *now* (as long as all substrings are LocalizedStrings as well) should result in the same thing. Though that's much less efficient as string builders and recursive string formatting will be performed.

src/auto-evo/RunResults.cs
Kasterisk: Yeah, I guess it's too specific of a use case now. I think I'll clean these up.

src/auto-evo/RunResults.cs
Kasterisk: In that case, methods might be redundant, could probably just use the `results` dicitionary.

hhyyrylainen: Yeah, that's what the apply results does. I assumed there was *some* reason why you didn't do that here, so I didn't suggest it...

Kasterisk: The methods were just left-overs from previous implementation done outside RunResults which is why there are public methods. I'll remove them in favor in direct accessing.

src/microbe_stage/editor/MicrobeEditor.cs
Kasterisk: Changed.

src/microbe_stage/editor/TimelineTab.cs
Kasterisk: Done.

src/microbe_stage/editor/TimelineTab.cs
Kasterisk: Changed.

src/auto-evo/RunResults.cs
```diff
@@ -811,6 +811,151 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+                    finalPatchPopulation += CountSpeciesSpreadPopulation(species, patch);
+
+                    if (results[species].SplitOffPatches?.Contains(patch) == true)
+                    {
+                        // All population splits off
+                        finalPatchPopulation = 0;
+                    }
+
+                    // Apply external effects
+                    if (effects != null && world.Map.CurrentPatch.ID == patch.ID)
```
hhyyrylainen: Just for future note, @FuJa0815's PR needs to adjust this part. I'm really hoping I can commit to getting the dynamic MP and patchwide population system merged early for 0.5.8 this time.

hhyyrylainen (approved):
Code looks fine now. Playtested, including save upgrading, and I didn't find any issues. I fixed a few remaining spelling issues and resolved the conflicts with translation files. I think I'll merge this now. Feel free to open a correction PR before tomorrow morning (my time).

Issue #2912, opened 2021-12-06
Investigate and fix iron compound food source


adQuid:
**Describe the bug**
I'm pretty sure the compound food source for iron is broken.

It is defined here
(https://github.com/Revolutionary-Games/Thrive/blob/master/src/auto-evo/simulation/PopulationSimulation.cs#L159)

Like all compound food sources, it takes dissolved compounds.
https://github.com/Revolutionary-Games/Thrive/blob/master/src/auto-evo/simulation/food_source/CompoundFoodSource.cs#L15

But iron is never found dissolved, it's only found in chunks.
https://github.com/Revolutionary-Games/Thrive/blob/master/simulation_parameters/microbe_stage/biomes.json#L101

Either the compound food source needs to be modified to handle this, a new food source class should be made, or I'm totally wrong and there's some other reason the auto-evo hates iron eaters. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to a patch with iron but no glucose (modifying the biomes file will speed this up)
2. Be a pure iron-eater
3. Observe your auto-evo prediction

**Expected behaviour**
A non-zero population prediction

**Screenshots**

**System info (please complete the following information):**
- OS: [e.g. Windows 10]
- Graphics card (you can find this in the game logs) [e.g. OpenGL ES 3.0 Renderer:
    GeForce GTX 1080 Ti/PCIe/SSE2]
- Thrive version [e.g. 0.4.0.2]

**Logs**

Please use a text hosting service (for example https://pastebin.com/) to submit your logs.
You can copy the launcher output or the contents of the following files: `log.txt` and mono
log file. You can find these log files by reading the game log to see their location.
For example:
```
Mono: Logfile is: /home/hhyyrylainen/.local/share/godot/mono/mono_logs/2020_07_21 23.10.26 (265121).txt
Game logs are written to: /home/hhyyrylainen/.local/share/Thrive/logs latest log is 'log.txt'
```

**Additional context**

hhyyrylainen:
Isn't it wrong for all non-environmental compounds to use that source?
For iron it is found in clouds as well as iron chunks. So the amounts available from those need to be summed up for the iron eating species to get energy from.

Maxonovien:
Iron isn't found (currently) in native clouds, but in clouds vented from chunks. Of course, this could happen, but this is the same logic as marine snow in a sense. I'll try to propose something based on that.

hhyyrylainen:
[...]

PR #2921, opened 2021-12-08
2912 fix iron food source

### LINKED ISSUE #2923: Auto-evo do not take into account marine snow density
As per the title, I can scavenge marine snow in the vents while its density is 0. 

This issue should be fixed by #2921 , but I open it here to keep track of it should it change.

Maxonovien:

This PR adds iron chunks as a possible food source by generalizing marine snow food source. It also considers density in the computations, and so I changed a few constants for balance.

Note that I specifically target iron & glucose as compounds of interest in chunks, although this could be expanded to other energy source compounds. The reason to keep iron and glucose selection is that considering all compounds should not break chunk sources, as long as we generalize this to the whole system. Discussions on this part welcome.

**Related Issues**

Closes (at least partially) #2912 
Closes #2923 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


hhyyrylainen:
I'll try to finish my emergency changes tonight so that it could be merged first as I also touched the marine snow food source file.

Maxonovien:
I tried to avoid making potential conflicts. We'll see if that works.

hhyyrylainen:
I guess it might work. At least theoretically git can handle the rename and changes in other branch.

src/auto-evo/simulation/PopulationSimulation.cs
```diff
@@ -157,7 +157,9 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
                 new CompoundFoodSource(patch, Glucose),
                 new CompoundFoodSource(patch, HydrogenSulfide),
                 new CompoundFoodSource(patch, Iron),
```
adQuid, line 159: This shouldn't do anything in the current state of the game, but I like that you kept it in. +1 for versitility!

src/auto-evo/simulation/food_source/ChunkFoodSource.cs
```diff
@@ -38,11 +46,24 @@ public override float FitnessScore(Species species, SimulationCache simulationCa
 
         score /= energyBalance.FinalBalanceStationary;
 
+        var ponderedEnergyGenerationScore = energyCompounds.Sum(
+            c => EnergyGenerationScore(microbeSpecies, c.Key) * ProportionOfTotalEnergy(c.Key));
```
adQuid, line 50: Is there a reason this needs to be proportional? I think you can save some division and seven code lines by just multiplying by amounts.

Remember that fitness scores are only used in ratio of other fitness scores of the same type, so scalar multipliers shouldn't matter.

src/auto-evo/simulation/food_source/ChunkFoodSource.cs
```diff
@@ -38,11 +46,24 @@ public override float FitnessScore(Species species, SimulationCache simulationCa
...
     }
+
+    public float ProportionOfTotalEnergy(Compound compound)
+    {
+        if (energyCompounds.TryGetValue(compound, out var quantity))
+            return quantity / energyCompounds.Sum(c => c.Value);
+
+        return 0;
+    }
 }
```
adQuid, line 69: nice

adQuid (commented):
I think there's one unnecessary calculation in there, but the approach is right on point.

src/auto-evo/simulation/food_source/ChunkFoodSource.cs
Maxonovien: That was a point on which I wasn't sure, so I went the safe way. So we agree that I can add/multiply any constant to the score, that won't matter, right?

adQuid (approved):

hhyyrylainen:
And there we go, this now has a conflict...

src/auto-evo/simulation/food_source/ChunkFoodSource.cs
```diff
@@ -38,6 +46,13 @@ public override float FitnessScore(Species species, SimulationCache simulationCa
 
         score /= energyBalance.FinalBalanceStationary;
 
+        // We ponder the score for each compound by its amount, leading to pondering in proportion of total quantity,
+        // with a constant factor that will be eliminated when making ratios of scores for this niche.
+        var ponderedEnergyGenerationScore = energyCompounds.Sum(
```
hhyyrylainen, line 51: I think this doesn't need a separate variable. Also this is part of the *base* score, so to me this multiplying is pretty weird here. Instead I'd move this to line 39 and make this an addition.

Maxonovien: It is not part of the base score, it *is* the base score. All the rest are multiplicative factors to this base.

hhyyrylainen: Yes, but you *multiply* it into the score *after* all the other stuff has been already calculated, at least github diff shows it to me like that.

hhyyrylainen: Kuvakaappaus - 2021-12-09 14-08-29

Maxonovien: I've just been reordering this, because I agree this may be misleading. But I *think* I saw it in another food source. Not sure though.

hhyyrylainen: Still you are *multiplying* the *final* score after all the adjustments, with what should be the *base* number and the adjustments should apply on top of that number.

Maxonovien: As I wrote just above, I already reordered that part of the logic.

hhyyrylainen:
I'll try to finish this PR now.

hhyyrylainen:
Marine snow and small iron chunks are now pretty ridiculously nerfed:

hhyyrylainen:
Looks like there's translation conflicts, Maybe I'll merge this anyway, and someone can in the meantime try to come up with better numbers to use. Remember that there's a release planned for *tomorrow*.

hhyyrylainen (approved):
The code is now fine to me, but big iron being OP and marine snow being nerfed (the total energy should at least be closer to 5000 for the marine snow) are problems, but I think I'll merge this soon anyway and wait for a follow up PR after that.

PR #2934, opened 2021-12-10
Tweaked engulfing sound volume and handling

Kasterisk:

Reduced engulfing sound volume to 0.6 and the sound now fades out smoothly when stopping engulf mode.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

the PR author.


src/microbe_stage/Microbe.cs
```diff
@@ -712,4 +720,113 @@ private void CountHexes()
...
+    /// <summary>
+    ///   Provides extra level of abstraction to allow simultaneous switching between 3D positional and non positional
+    ///   audio players in a single Spatial-derived class.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Useful in cases where the playing of an audio stream must be able to fulfill both of these conditions.
+    ///   </para>
+    /// </remarks>
+    private class AudioPlayer : Spatial
```
hhyyrylainen, line 733: This should probably be in a new file in the engine folder, as I presume for 3D character stages we might have the same issue as currently.

src/general/HybridAudioPlayer.cs
```diff
@@ -0,0 +1,110 @@
...
+    private AudioStreamPlayer3D player3D;
+    private AudioStreamPlayer playerNonPositional;
+
+    private float volume;
+
+    public HybridAudioPlayer()
+    {
+        player3D = new AudioStreamPlayer3D();
+        playerNonPositional = new AudioStreamPlayer();
+
```
hhyyrylainen, line 23: Would it be completely crazy to suggest that these would be created on demand? With this approach we create potentially hundreds of unnecessary audio player objects.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -782,13 +782,29 @@ private void HandleEngulfing(float delta)
             if (!engulfAudio.Playing)
                 engulfAudio.Play();
 
+            // To balance loudness, here the engulfment audio's max volume is reduced to 0.6 in linear energy
```
hhyyrylainen, line 785: `linear volume`?

src/microbe_stage/Microbe.cs
```diff
@@ -179,12 +179,26 @@ public override void _Ready()
...
         cellBurstEffectScene = GD.Load<PackedScene>("res://src/microbe_stage/particles/CellBurstEffect.tscn");
 
+        engulfAudio = new HybridAudioPlayer
+        {
+            Stream = GD.Load<AudioStream>("res://assets/sounds/soundeffects/engulfment.ogg"),
+            Positional = !IsPlayerMicrobe,
+            Bus = "SFX",
+        };
+
+        AddChild(engulfAudio);
```
hhyyrylainen, line 194: engulfAudio could use _Ready to determine which player type to create as that should trigger here after the properties have been set.

src/general/HybridAudioPlayer.cs
Kasterisk: No that's perfectly sensible, I think Positional property should be readonly.

Kasterisk: Added.

RacerBG (approved):
Sounds good to me while testing it.

src/general/HybridAudioPlayer.cs
```diff
@@ -0,0 +1,146 @@
...
+///   audio players in a single Node-derived class. Though setting this can only be done in instantiation.
+/// </summary>
+/// <remarks>
+///   <para>
+///     Useful in cases where the playing of an audio stream must be able to fulfill both of these conditions.
+///   </para>
+/// </remarks>
+public class HybridAudioPlayer : Node
+{
+    public readonly bool Positional;
```
hhyyrylainen, line 14: I'd make this a get only property

hhyyrylainen: Code looks good except this one thing.

hhyyrylainen (changes requested):
I'd like to see one small change that shouldn't impact functionality at all, so if that is fixed by tomorrow morning (my time), this can be included in 0.5.6.1

src/general/HybridAudioPlayer.cs
Kasterisk: Changed.

hhyyrylainen (approved):
Code looks good now.

PR #2949, opened 2021-12-11
Print player species first in auto evo results

### LINKED ISSUE #2772: Make player species always print first/last in the auto-evo results report
so that's it's easier to find what happened to your species

ArrayOfCells:

First and foremost the player species is printed first in the auto evo-results. The change is implemented more generalised and would work for a set of multiple species (e.g. if the player could define species "to follow"). In addition to that the species in the results are now ordered alphabetically.

**Related Issues**

closes #2772 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/auto-evo/RunResults.cs
```diff
@@ -793,6 +797,23 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
                     builder.Append('\n');
             }
 
+            // Prepare highlight species to remove the need to check those things in every loop iteration
+            var relevantHighlightSpecies = new HashSet<Species>(highlightSpecies) ?? new HashSet<Species>();
+            relevantHighlightSpecies.RemoveWhere(s => s == null || !results.ContainsKey(s));
```
Maxonovien, line 802: Creating a new var (especially a copy) here is probably costlier than looping through a few absent species. I assume it isn't intended to include the whole biodiversity anyways.

Also, I'm not sure that there's any reason for species to be null in this set, so I think that should rather raise an error.

ArrayOfCells: I agree and changed the code accordingly

src/auto-evo/RunResults.cs
```diff
@@ -793,6 +797,24 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
                     builder.Append('\n');
             }
 
+            highlightSpecies ??= new HashSet<Species>();
+            if (highlightSpecies.Contains(null))
+                throw new ArgumentNullException("Highlight species entry must not be null");
```
Maxonovien, line 802: `ArgumentNullException` should be reserved to cases where an argument it itself is null, not when an enumerable argument contains null argument. Besides that, it is the reason why Jetbrains (auto-checks) reject the PR, since it expects `ArgumentNullException` to refer to the null argument.

src/auto-evo/RunResults.cs
```diff
@@ -793,6 +797,24 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
...
+
+            // Print highlight species results
+            foreach (Species species in highlightSpecies.OrderBy(s => s.FormattedName))
+            {
+                if (results.ContainsKey(species))
+                    AddSpeciesResultString(results[species]);
```
Maxonovien, line 808: The dictionary method `TryGetValue(key, out var value)` is precisely designed for these cases, allowing you to define `dict[key]` in `var value`, which you can then use (even outside the `if` statement, don't ask me about this magic).

This is both easier to read and more efficient computationally.

Maxonovien (dismissed):
Few remarks on the code itself, but @hhyyrylainen suggested another, more efficient and cleaner approach in our eyes: using a custom sorter to get everything ordered, which would allow much more flexibility. The only potential harder part is to write the sorter cleanly.

Otherwise I want to point out that this approach does not allow for more than 2 priority levels anyways (base + highlight), while we would most likely have several in general (since the player should always be at the top & alone).

ArrayOfCells:
> Few remarks on the code itself, but @hhyyrylainen suggested another, more efficient and cleaner approach in our eyes: using a custom sorter to get everything ordered, which would allow much more flexibility. The only potential harder part is to write the sorter cleanly.
> 
> Otherwise I want to point out that this approach does not allow for more than 2 priority levels anyways (base + highlight), while we would most likely have several in general (since the player should always be at the top & alone).

While thinking about a solution I wondered about my own initial design. At a bare minimum the following OrderBy would solve the problem as well (without the additional parameter for MakeSummary). I guess I will not be able to build a faster sort than this myself. A disadvantage is that sort conditions can not be injected easily without changing the RunResult class itself. What do you think about that?

`foreach (var entry in results.OrderByDescending(s => s.Key.PlayerSpecies).ThenBy(s => s.Key.FormattedName))
 {
     AddSpeciesResultString(entry.Value);
 }`

I could build a custom sorter as well but that would inevitably lead to some overhead and a slower sort. It could still be a good idea if the sort result should provide some metadata as well, like a priority identifier for the current element or some kind of control break indicator for easier handling of the sort result.

src/auto-evo/RunResults.cs
ArrayOfCells: Thanks for the info, I already wondered what was wrong with my code

src/auto-evo/RunResults.cs
ArrayOfCells: Thanks, I will keep that in mind. I guess I have a lot to learn, especially when it comes to performance sensitive operations like in a game.

hhyyrylainen:
> foreach (var entry in results.OrderByDescending(s => s.Key.PlayerSpecies).ThenBy(s => s.Key.FormattedName)) { AddSpeciesResultString(entry.Value); }

I think that would be a really nice and pretty clean solution to get this done. I kind of want to replace the whole textual description eventually with something else so future expandability is not really a concern here (https://forum.revolutionarygamesstudio.com/t/explaining-the-auto-evo-prediction-to-the-player/836/2).

src/auto-evo/RunResults.cs
Maxonovien: That's normal, we all learn a lot thanks to Thrive. Actually, I've especially just got to learn a lot about this method with my latest PRs. 😉

hhyyrylainen:
@ArrayOfCells is there a reason why you haven't signed the CLA? I've purposefully not looked at your code yet, because we won't be able to use your code if you don't sign the CLA.

src/auto-evo/RunResults.cs
```diff
@@ -793,6 +798,11 @@ void OutputPopulationForPatch(Species species, Patch patch, long population)
                     builder.Append('\n');
             }
 
+            foreach (var entry in results.OrderByDescending(s => s.Key.PlayerSpecies).ThenBy(s => s.Key.FormattedName))
+            {
+                AddSpeciesResultString(entry.Value);
+            }
```
Maxonovien, line 804: Since you only use the value of the elements you iterate over in the end, I think you can directly call `dict.Values`.

This would also allow for a proper renaming of `entry` to a more explicit name.

hhyyrylainen: I agree. As far as I can tell, this entire PR could be refactored to just modify the one foreach loop that already exists with that new `OrderByDescending...` stuff and it would allow all of the other code to stay the same.

ArrayOfCells: I made the changes but I still renamed "entry" to "speciesResult". I could change it back to entry too or give it an even clearer name but I stayed at the middle ground for now.

hhyyrylainen (changes requested):
You still have renamed a bunch of variables causing a lot of changes in the shown diff.

ArrayOfCells:
Yes, I commented above about this. I followed the suggestion from Maxonovien and still left the renamed variables there. Just to be sure now, you want me to change the loop variable name back to "entry"?

hhyyrylainen (approved):

hhyyrylainen:
Looks like when I merged master into this, the newer version of Jetbrains tools doesn't like the amount of indentation in the code anymore.

ArrayOfCells:
Is there something I have to correct on my end as well? I am a bit confused by the error message because 9 spaces seem odd. I just pulled master and my local format checks also do not run anymore (it complains about my installed .NET Framework 6.0 version and that it's missing .NET Core 3.1). I guess that's just a configuration issue on my end, but I just wanted you to know because it could also affect others as well.

hhyyrylainen:
My recent PR, which your branch probably didn't have, switched us over to using dotnet tool system to download and run the jetbrains tools. I've updated the setup instructions regarding this.

PR #2952, opened 2021-12-12
Changed report tab patch name label into a dropdown list of patches

### LINKED ISSUE #2885: Make it possible to switch patch from the patch name shown on the report tab
I think this is a very useful feature to add. Instead having to switch to the patch map to select another patch we can instead select it directly from the report tab. I think a simplest way this can be done is by turning the patch name text into a dropdown list.

Kasterisk:

This PR adds the ability to select patches in the report tab from the patch list dropdown.

**Related Issues**

Closes #2885 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


RacerBG (changes requested):
This PR needs more work. I have 2 things to say but a screenshot is worth 1000 words as they say:


And the second thing is related to the way how the drop-down list works. In a Freebuild everything is fine but during a normal gameplay this can be very confusing for a new player because you can still change the patch through it but you will remain in your old patch if you don't know which patch is next to you. I don't know if the following situation should be allowed in a normal gameplay (on the patch map it is allowed as a view feature) but if you select the patch from here and then go to the patch map it can get even more confusing because the "Move to this patch button" will be greyed out (your selected patch probably will be inaccessible):

hhyyrylainen:
I think the dropdown shouldn't affect the patch the player is in, and we could have a piece of help text showing at the start of the report saying that you have not moved to the patch.
Also it needs to be verified that selecting a different patch there *doesn't* interfere with the patch map tutorial.
One final concern I have is that if we have huge patch maps, then we might need to limit the selection to like the closest 20 patches to the one the player is in.

RacerBG:
> I think the dropdown shouldn't affect the patch the player is in, and we could have a piece of help text showing at the start of the report saying that you have not moved to the patch. Also it needs to be verified that selecting a different patch there _doesn't_ interfere with the patch map tutorial. One final concern I have is that if we have huge patch maps, then we might need to limit the selection to like the closest 20 patches to the one the player is in.

This feature (PR) alone could apply only for the Statistics report. The Auto-Evo report is already patch independent. I'm also on the opinion that you shouldn't be able to change the patch directly from here.

hhyyrylainen:
I *think* it would be fine to make it so that selecting a different patch from the dropdown also selects the one of the patch map. That way the player can switch between the two tabs and have the data they show be in sync, which I think would be good to reduce confusion.

RacerBG:
> I _think_ it would be fine to make it so that selecting a different patch from the dropdown also selects the one of the patch map. That way the player can switch between the two tabs and have the data they show be in sync, which I think would be good to reduce confusion.

It's already doing that but I'm afraid that this feature will confuse the new players even more. Let me make a few screenshots now and I will pretend to be a new player for a few minutes...

hhyyrylainen:
As long as the report tab shows a text saying "you are not in this patch" and the tutorial about the patch map isn't skipped, I think it would be okay.

RacerBG:
For simplicity, this are a Freebuild screenshots. I selected the Cave patch even though I'm located in the in the Vents. Then "Move to this patch" will be gray, I will press "Next" anyway assuming that I had moved to this patch and I will still be in the Vents. I will continue to do that untill I find out that this system is not working as I was thinking it should work and it will make it even more confusing because I will have no idea were actually is my current patch.

hhyyrylainen:
Ah, I see. That is definitely a problem. Even now people miss the move to this patch button... Definitely need to think of something to not reduce the usability there.

RacerBG:
> Ah, I see. That is definitely a problem. Even now people miss the move to this patch button... Definitely need to think of something to not reduce the usability there.

As I said if this is Statistics inclusive it's not a big deal but in the current situation it will add another level of confusion to the already confusing (for some players) patch map.

hhyyrylainen:
At least the reverse still should be the case so if you select something on the patch map, going back to the previous tab should show the statistics for that. But I guess the reverse of selecting a patch on the report tab shouldn't affect the patch map tab.

Kasterisk:
So to sump up:
- Dropdown should only affect what's shown on the report tab and patch map interdependency should be separated.
- Selecting a patch does not move the player to that patch

Second point relates to a more general issue with the "select here" and "move here" map mechanics that's out of scope of this PR since this PR itself falls within the purview of said system.

By the way this opens up an interesting question regarding the timeline tab, should it be worldwide or should it be patchwide? Likewise the food chain tab. Looking at long term, they would be no doubt filled with ton of events and data that a patchwide limit would make sense. Though to give more option maybe add a separate "global" option in the dropdown for showing what happens worldwide?

RacerBG:
> So to sump up:
> 
>     * Dropdown should only affect what's shown on the report tab and patch map interdependency should be separated.
> 
>     * Selecting a patch does not move the player to that patch
> 
> 
> Second point relates to a more general issue with the "select here" and "move here" map mechanics that's out of scope of this PR since this PR itself falls within the purview of said system.
> 
> By the way this opens up an interesting question regarding the timeline tab, should it be worldwide or should it be patchwide? Likewise the food chain tab. Looking at long term, they would be no doubt filled with ton of events and data that a patchwide limit would make sense. Though to give more option maybe add a separate "global" option in the dropdown for showing what happens worldwide?

I'm more in favor for a global timeline and food chain tabs but I won't refuse a filter or a switch to make them patchwide as well. Currently both the timeline and the auto-evo report are basically a huge wall of text.

Kasterisk:
Yeah, I think a filter needs to be added eventually considering amount of information in future stages.

hhyyrylainen:
> Dropdown should only affect what's shown on the report tab and patch map interdependency should be separated.

I'd say that it still should be possible to select a patch on the patch map and go back to the report tab to see the graphs for it. But yeah, the other direction, there should be no selected patch change on the map if the report tab patch is changed.

Kasterisk:
> One final concern I have is that if we have huge patch maps, then we might need to limit the selection to like the closest 20 patches to the one the player is in.

Just noticed this and this might require some kind of graph traversal or nearest neighbor algorithm. As for me, might take a little while to figure out...

hhyyrylainen:
> Just noticed this and this might require some kind of graph traversal or nearest neighbor search algorithm. As for me, might take a little while to figure out...

I think you could do some kind of recursive algorithm like:

```
int FindNeighbours(Patch patch, ISet<Patch> results, int left = 20)
if(left <= 0) 
  return 0;

foreach(var patch2 in patch.neighbours){
    results.Add(patch2);
    if(--left <= 0)
         return 0;
}

foreach(var patch2 in patch.neighbours){
    left = FindNeighbours(patch2, results, left);
}

return left;

```

Very rough code but about that kind of algorithm should work where as a kind of breadth first traversal.

Zireael07:
In my own game, I found a simple BFS (Breadth First Search) on a graph worked for figuring out neighbors ;)

Kasterisk:
Indeed, just implemented this and it seems like a perfect fit for the problem. Thanks for the suggestions :)

src/microbe_stage/Patch.cs
```diff
@@ -107,6 +107,44 @@ public HashSet<Patch> GetAllConnectedPatches()
...
+        while (queue.Count > 0 && !maxReached)
+        {
+            var vertex = queue.Dequeue();
+
+            foreach (var patch in vertex.Adjacent)
+            {
+                if (!visited.Contains(patch))
+                {
+                    queue.Enqueue(patch);
+                    visited.Add(patch);
```
hhyyrylainen, line 134: I think `Add` already returns a boolean indicating if the value was new or not, so this could be used in the if on line 131.

src/microbe_stage/PatchMap.cs
```diff
@@ -46,7 +46,10 @@ public Patch CurrentPatch
     public void AddPatch(Patch patch)
     {
         if (Patches.ContainsKey(patch.ID))
-            throw new ArgumentException("patch cannot be added to this map");
+        {
+            throw new ArgumentException(
+                "patch cannot be added to this map: already exist patch with ID " + patch.ID);
```
hhyyrylainen, line 51:
```suggestion
                "patch cannot be added to this map, the ID is already in use: " + patch.ID);
```

hhyyrylainen (changes requested):
Overall this looks pretty good. Slight code tweaking is needed.
Besides that I noticed a UI issue (which is why the patch names where autowrap enabled before):



even worse in other languages:

Kasterisk:
Looks like I have to clip the text, there's no autowrap setting for a menu button it seems.

RacerBG:
Clipping the text fixes the formatting but now the names looks, let's say not right:



The solution in my opinion is to just shorten the patch names in Weblate or make the font of the button 1.5 times or 2 times smaller.

hhyyrylainen:
I think perhaps a different GUI control / doing our own should be considered. So we could have the wrapping text of the patch, but next to it some kind of "pseudo" option button that allows selecting a different patch...
The patch names can't really be any shorter because once we have a patch map generator I'd imagine we'll have multiple parts, so the "pangonian" part which is like the area name needs to be visible as well as the scientific name of the biome. It can't be cut to be any shorter.

Kasterisk:
How does this look? I basically separated the option button and the patch name text while keeping the same look. The Patch name text is just a simple label with background theme override.


Tradeoff is the dropdown is aligned like so and not starting from the left of the text and then covering the whole width as before. Not that big of a deal though.



Maybe it'd be better if the button is vertically aligned to the bottom?

hhyyrylainen:
I think that's better than having to cut off important characters from the name.

RacerBG:
It will be fine for the time being.

RacerBG (commented):
The formatting still needs improvement for longer names as you can see. And as I said earlier, would it be possible to have the old patch names back and having a button for the drop down menus as currently? In my opinion this will be the best of both worlds.

Other than that everything else works as expected.

84634E1A607A:
Can the drop down be made a little rounder?

Kasterisk:
@RacerBG You mean removing the background on the patch name?



It results to this, the dropdown button is far to the right due to patch name text's expand horizontally size flag, which looks just weird. Unless we moved the button to the left of the text which doesn't really look intuitive either.



@84634E1A607A Sure, that'll require a change to the styling in thrive_theme.tres.

hhyyrylainen:
I think that variant with the patch selection to the left of the patch name looks the best out of all the screenshots shown so far.

Kasterisk:
Alright, will change to that.

RacerBG:
> Alright, will change to that.

Exactly! :smiley:

RacerBG (approved):

src/microbe_stage/editor/MicrobeEditorGUI.cs
```diff
@@ -2145,6 +2166,15 @@ private void UpdateShownPatchDetails()
 
         // Enable move to patch button if this is a valid move
         moveToPatchButton.Disabled = !editor.IsPatchMoveValid(patch);
+
+        reportTabPatchSelector.Select(reportTabPatchSelector.GetItemIndex(patch.ID));
+    }
+
+    private void OnReportTabPatchListSelected(int index)
+    {
+        var patch = editor.CurrentGame.GameWorld.Map.Patches[index];
```
hhyyrylainen, line 2175: This is not correct, the patches find order shouldn't be assumed to be the same.

This should instead get the item ID based on the index from the patch selector, and then from the map get the patch by the ID.

hhyyrylainen (changes requested):

src/microbe_stage/editor/MicrobeEditorGUI.cs
Kasterisk: Fixed.

hhyyrylainen (approved):
Seems to work perfectly fine now and I have no concerns about the code left.

PR #2966, opened 2021-12-15
Turned help screen into a window dialog

Kasterisk:

Just realized it makes most sense to turn this into a window dialog given its typical "content and close button" structure.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/locale/bg.po
+++ b/locale/bg.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2021-12-12 15:38+0700\n"
+"POT-Creation-Date: 2021-12-15 22:05+0700\n"
 "PO-Revision-Date: 2021-12-11 08:35+0000\n"
 "Last-Translator: Georgi Georgiev (RacerBG) <g.georgiev.shumen@gmail.com>\n"
 "Language-Team: Bulgarian <https://translate.revolutionarygamesstudio.com/"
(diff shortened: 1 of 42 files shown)


RacerBG:
I haven't tested it yet but it looks nice from your screenshot. The old design reminds me of the compounds and environment popup windows and it wasn't bad as well. The new design is consistent with the more recent changes and also the cell processes window.

I'm going to approve it anyway after testing it's functionality.

Kasterisk:
This dialog based frame is closer to how I imagine Thrivepedia would look as well so this change made the help screen look a bit closer to that at least.

hhyyrylainen:
If the help menu is getting this treatment, I'm not sure we have many things left in the game that *shouldn't* be popup windows...

hhyyrylainen:
Btw does this has any impact on https://github.com/Revolutionary-Games/Thrive/pull/2877 ?

RacerBG (approved):
No unexpected behaviour and new errors. Everything looks fine.

Kasterisk:
The help menu is part of pause menu and will probably stacked on top of the theoretically popup-based PauseMenu on pop up so it shouldn't impact it too much I think.

hhyyrylainen:
That's good. Btw could you maybe help get that PR 2877 finished anytime soon? I'd like to see it get finished relatively soon, so that once I manage to make the crash reported system work again, it would be in 0.5.6.2.

hhyyrylainen:
To me this kinda doesn't look that good, like the pause menu is already a fullscreen overlay, and it has submenus, so I think this reduces the consistency in how it functions, like it's no longer a fullfledged thing on it's own but instead needs to open extra popups on top of it. This change would make more sense if the help menu could be opened independently without the pause menu being open...

RacerBG:
> To me this kinda doesn't look that good, like the pause menu is already a fullscreen overlay, and it has submenus, so I think this reduces the consistency in how it functions, like it's no longer a fullfledged thing on it's own but instead needs to open extra popups on top of it. This change would make more sense if the help menu could be opened independently without the pause menu being open...

Removing the button from the pause menu should help, I think?

hhyyrylainen:
Technically it is implemented so that it can't be really well open while the pause menu isn't open. Also I think our tutorials mention that you can find help in the pause menu. So that's not a very good option, I don't think.

Kasterisk:
How would we present Thrivepedia? Should its frame be non-window dialog based and should it be accessible only through main/pause menu or through other means?

hhyyrylainen:
I think it should be in the pause menu. I think we can perfectly fine have frames / elements like that in the pause menu. It's just when literally a core part of the pause menu is implemented as a window dialog that can be dragged around and has a titlebar, is what feels wrong to me.

PR #2972, opened 2021-12-18
Adjust save max length

### LINKED ISSUE #2969: Add max length to the savename
Linux and windows has a max filename of 255, which means 244 characters are usable (without the `.thrivesave` ending). The user can enter longer filenames and the saving crashes.

84634E1A607A:
Save length adjusted to 100 chars; Load screen save name clipped to 40 chars.

**Related Issues**
Fixes #2969 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Couldn't the save screen just have line wrapping enabled instead of clipping? I think that saves with long names taking up extra vertical space would be fine.

84634E1A607A:
image

Like this? Oh... Impressive! (This is an empty file so no screenshot is shown)

84634E1A607A:
image

Personally think this looks better.

FuJa0815:
Personally I think clipping is better here

Kasterisk:
If we're going for clipping, Godot 4 supports [text trimming](https://github.com/godotengine/godot/pull/51178) that automatically adds ellipsis as shown here, so this need to be noted somewhere in the code for when we finally move to godot 4.

hhyyrylainen:
Wouldn't the user still want to see their cool filename they selected in its entirety? Even if the layout is not optimized for it. I think it's wrong to allow the user to enter a long name that then later gets clipped because for example the player might use a long name with a suffix like "_2" or "_3" and then try to differentiate the files in the load menu.

84634E1A607A:
Is there a function to get the predicted string display width?
A string with a lot of Ms is significantly longer than normal.

The place can hold about 20 Ms, but 40 normal characters.

Kasterisk:
Sounds like [Font.GetStringSize()](https://docs.godotengine.org/en/stable/classes/class_font.html#class-font-method-get-string-size) is what you're looking for.

84634E1A607A:
Then I think we can remove from the last 5 chars, until the text fits.

Like 'djhehhhfhrhjudhh......hdo_4'

Kasterisk:
How about we clip the text but show the full name in a tooltip when the player hovers the name?

hhyyrylainen:
Does it *really* look that bad to just enable the text wrapping on the save name label? That's a 5 second change (well a bit more if you count committing to Github).

84634E1A607A:
Okay...

Btw, we can restrict long save name when saving, but not when they manually change it in filesystem. And 100 chars may be too long if you set the save folder to a dir with long paths (maybe with symbol links). So we may add a function to determine how long exactly the save name can be.

84634E1A607A:
Okay...

Btw, we can restrict long save name when saving, but not when they manually change it in filesystem. And 100 chars may be too long if you set the save folder to a dir with long paths (maybe with symbol links). So we may add a function to determine how long exactly the save name can be.

hhyyrylainen:
As far as I know symbolic links have no effect on the path length limit. As long as there isn't some code that tries to resolve the absolute path and then use that. Otherwise the OS is just given the path with symbolic links in it and the file system will internally resolve it when opening the file.

So anyway manually named saves can bypass any limit we set? I really think we should just add wrapping to that save name label, and make one PR of just that change and I'd accept that right away (I might just soon do that myself if no one else does it).

84634E1A607A:
Well I am talking about adding a manual check And if file.open failed We use a tooltip instead of exception dialog.

RacerBG (changes requested):
The current implementation doesn't look right to me. Having a horizontal scrolling on the Saving/Loading screen is not the best idea in my opinion. Also, as was commented earlier a manual edit of the file name can cause even longer save names and this causes even bigger horizontal scrolling.

84634E1A607A:
The problem addressed above has been fixed by #2988 but not yet merged to this pr.

84634E1A607A:
I think the biggest problem this needs to solve is that if user haI typed in a loooong name exceeding the path length limit, the saving fails and the name typed in is not preserved.

84634E1A607A:
If this isn't regarded as a problem, then this pr can be closed.

PR #2985, opened 2021-12-21
Boosted FPS by seperating compound cloud calculation in its own process loop

FuJa0815:

Boosted the FPS on my laptop in battery mode from 30FPS to about 70FPS. The only thing that lags on weak hardware are the compound clouds, not the whole game.

Needs testing on fast hardware to verify that this doesn't affect the compound cloud system on them.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/CompoundCloudPlane.cs
+++ b/src/microbe_stage/CompoundCloudPlane.cs
@@ -299,7 +299,7 @@ public class CompoundCloudPlane : CSGMesh, ISaveLoadedTracked
     /// <summary>
     ///   Updates the cloud in parallel.
     /// </summary>
-    public void QueueUpdateCloud(float delta, List<Task> queue)
+    public IEnumerable<Task> QueueUpdateCloud(float delta)
     {
         // The diffusion rate seems to have a bigger effect
         delta *= 100.0f;
@@ -316,7 +316,7 @@ public class CompoundCloudPlane : CSGMesh, ISaveLoadedTracked
                     Size / Constants.CLOUD_SQUARES_PER_SIDE,
                     Size / Constants.CLOUD_SQUARES_PER_SIDE,
                     delta, pos));
-                queue.Add(task);
+                yield return task;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


hhyyrylainen (changes requested):
Doesn't this just split compound cloud updates accross 3 frames?
That's simply wrong because microbes can absorb stuff in the meantime.
Did you try the performance option of only updating the clouds like very 100ms, in the performance options? That already makes the full cloud simulation be executed multiple times fewer per second.

RacerBG (changes requested):
I'm sorry to be the bearer of bad news but this did not improve the perfomance by even 1 FPS. The game was a solid 70 FPS at the beginning but even without this branch it was at 70 FPS to begin with and the new clouds looks like a lego pieces. And for some reason with this branch my old saves just CTD without anything wrong in the log files - it just says "loading finished".

hhyyrylainen:
I think @RacerBG is using the cloud simulation increased interval already to alleviate the clouds simulation bottleneck on their system. So maybe my previous comment is accurate.

RacerBG:
> I think @RacerBG is using the cloud simulation increased interval already to alleviate the clouds simulation bottleneck on their system. So maybe my previous comment is accurate.

It is. I forgot to mention that I'm using the 250 ms interval.

FuJa0815:
> Doesn't this just split compound cloud updates accross 3 frames? That's simply wrong because microbes can absorb stuff in the meantime. Did you try the performance option of only updating the clouds like very 100ms, in the performance options? That already makes the full cloud simulation be executed multiple times fewer per second.

Yes, it splits the compound cloud updates to 3 or more frames. I haven't greatly tested the implications of this but eating compounds still works. I have noticed that this code eats ~40% of the frame time for me. I saw that some threads are made but the same _Process call waits for all of them to finish. I've tested this with 40ms cloud interval.



> I'm sorry to be the bearer of bad news but this did not improve the perfomance by even 1 FPS. The game was a solid 70 FPS at the beginning but even without this branch it was at 70 FPS to begin with and the new clouds looks like a lego pieces. And for some reason with this branch my old saves just CTD's without anything wrong in the log files - it just says "loading finished".

In that case it seems like the compound cloud system simply wan't the bottleneck for you.

hhyyrylainen:
> In that case it seems like the compound cloud system simply wan't the bottleneck for you.

> It is. I forgot to mention that I'm using the 250 ms interval.

When it is only ran 4 times per second, the performance gain from this is probably minimal as described.

hhyyrylainen:
I'll close this with the conclusion that having a separate slowing down system for the compound clouds is not needed.

PR #2987, opened 2021-12-23
Tweaked toxin particles visuals

Kasterisk:

Changed the toxin particles blend mode to additive to make it more visible/brighter in darker patches and increased fire toxin particles visibility AABB. The reason why it's increased is because previously the particle trail suddenly disappears when going out of visible screen (can be reproduced by shooting toxin vertically relative to the screen), increasing this should hopefully make it stay visible for a bit longer.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


FuJa0815 (changes requested):
The toxins are invisible in GLES2 mode for me

Kasterisk:
That's a problem since we first added particles into the game. Godot's Particles node only works in GLES3 unless you're using CPUParticles.

hhyyrylainen (approved):
I'll assume this also looks fine even without testing this personally.

PR #3007, opened 2022-01-08
Added Save / Load menu double-click handler

### LINKED ISSUE #2970: Make doubleclicking on a save save / load it
EDIT

When:

1. In SaveMenu double-clicking a save should display a override confirmation dialog and save the current game state using the item's name. (Current behavior: when you double-click on a save you select it then deselect)
2. In LoadMenu double-clicking a save should load it. (Current behavior: you select it then deselect)

There's a strange thing, that if you have selected a save to override, you can select one after it (appears downwards), but not one before it.

84634E1A607A:
**Related Issues**
fixes #2970

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Kasterisk (approved):

RacerBG (approved):
LGTM.

src/saving/NewSaveMenu.cs
```diff
@@ -167,4 +167,13 @@ private void OnSaveNameTextEntered(string newName)
             ToolTipManager.Instance.ShowPopup(TranslationServer.Translate("INVALID_SAVE_NAME_POPUP"), 2.5f);
         }
     }
+
+    private void OnSaveListConfirmed(SaveListItem item)
```
hhyyrylainen, line 171:
```suggestion
    private void OnSaveListItemConfirmed(SaveListItem item)
```

src/saving/NewSaveMenu.cs
```diff
@@ -167,4 +167,13 @@ private void OnSaveNameTextEntered(string newName)
...
+    {
+        saveNameBox.Text = item.SaveName;
+        overwriteConfirm.DialogText = string.Format(CultureInfo.CurrentCulture,
+            TranslationServer.Translate("THE_CHOSEN_FILENAME_ALREADY_EXISTS"),
+            item.SaveName);
+        overwriteConfirm.PopupCenteredShrink();
```
hhyyrylainen, line 177: Isn't this duplicate code? I think these should be placed into a common method that both places can use.

src/saving/SaveListItem.cs
```diff
@@ -77,6 +77,9 @@ public class SaveListItem : PanelContainer
     [Signal]
     public delegate void OnSelectedChanged();
 
+    [Signal]
+    public delegate void OnDoubleClicked(SaveListItem item);
```
hhyyrylainen, line 81: Hang on why is the signal returning `this`? I'd much prefer so that there isn't a useless parameter (an object sending a signal and saying "hey, this is me" is unnecessary.

src/saving/SaveList.cs
```diff
@@ -137,6 +140,8 @@ public override void _Process(float delta)
                 if (SelectableItems)
                     item.Connect(nameof(SaveListItem.OnSelectedChanged), this, nameof(OnSubItemSelectedChanged));
 
+                item.Connect(nameof(SaveListItem.OnDoubleClicked), this, nameof(OnItemDoubleClicked));
```
hhyyrylainen, line 143: Instead put the value as a connection parameter here, similarly to how the `new Array { save }` is defined for the other connections below this one.

src/saving/SaveManagerGUI.cs
```diff
@@ -258,4 +258,9 @@ private void OnBackButton()
         GUICommon.Instance.PlayButtonPressSound();
         EmitSignal(nameof(OnBackPressed));
     }
+
+    private void OnSaveListConfirmed(SaveListItem item)
```
hhyyrylainen, line 262:
```suggestion
    private void OnSaveListItemConfirmed(SaveListItem item)
```

hhyyrylainen (changes requested):

src/saving/SaveList.cs
84634E1A607A: Not fully understand that.

hhyyrylainen:
```
                 item.Connect(nameof(SaveListItem.OnDoubleClicked), this, nameof(OnItemDoubleClicked), new Array { item });
```

hhyyrylainen: and remove the "this" from the call when the event is emitted.

src/saving/SaveListItem.cs
hhyyrylainen, line 81: This parameter here still is not removed.

hhyyrylainen (changes requested):

src/saving/SaveListItem.cs
hhyyrylainen: I did this fix myself.

hhyyrylainen (approved):
Basic functionality seems to still work and I have no further code comments.

PR #3011, opened 2022-01-09
Added microbe movement ambient loop sound

Kasterisk:

For context see https://github.com/Revolutionary-Games/Thrive/pull/2793#issuecomment-998960115. Can be enhanced by https://github.com/Revolutionary-Games/Thrive/pull/1930.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/assets/sounds/soundeffects/microbe-movement-ambience.ogg
+++ b/assets/sounds/soundeffects/microbe-movement-ambience.ogg
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:2019557f510cd204d8f05b09d62af10f633ed8fcf33304a0e78502f4143a1af0
+size 140262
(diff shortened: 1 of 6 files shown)


hhyyrylainen (approved):
I think the delay is a bit too long between when the movement start sound can play, for example when moving left, stopping and then moving right when trying to complete the tutorial I was wondering why the other direction didn't cause the sound. Turns out you would just need to wait a bit longer. I'll accept this PR now to get this done as this has been pending for a bit but I'd really like a follow up PR tweaking that aspect.

Kasterisk:
That wasn't the delay I think, the movement start sound only plays when you're mostly idle (that is little to none velocity) so quickly stopping and changing direction may not qualify as "starting" to move since there's some amount of velocity left. I agree that this probably needs tweaking.

hhyyrylainen:
That could explain it, but you also seemed to add a variable called `movementSoundCooldownTimer` I based on my comment on that seemingly being a minimum time delay between when the sound plays.

Issue #3016, opened 2022-01-14
Improve game performance with a lot of entities


hhyyrylainen:
The game performance suffers after playing for roughly an hour (or making a good cell and spamming the editor in freebuild). As a workaround I've reduced and reduced again the total entity cap in the game all the way down to 90. According to Godot engine benchmarks you should be able to have hundreds of physics entities at once, so I'm not sure what's up.

I've done profiling: https://forum.revolutionarygamesstudio.com/t/increasing-game-performance/841

and tried various things including detaching organelles from the scene tree, removing / combining the organelle physics to a single shape group. But nothing seems to give more than 5-10 FPS (when at sub 20 FPS), completely removing the visuals and collisions would technically make the game playable but I think we need to either keep the entity cap low or resolve the performance bottleneck rather than ruin the game visuals for not much gain.

Related issues:
- https://github.com/Revolutionary-Games/Thrive/issues/2910
- https://github.com/Revolutionary-Games/Thrive/issues/2909
- https://github.com/Revolutionary-Games/Thrive/issues/2587 (might need to adjust floating chunks that were spawned from cells dying to also despawn)
- https://github.com/Revolutionary-Games/Thrive/issues/1551

Here's the PR where I lowered the entity count: https://github.com/Revolutionary-Games/Thrive/pull/3017

hhyyrylainen:
That starting point actually reminds me, I think there's a place in the chunk spawning code or something that sets the shape. Actually looking at this I think I found that perhaps only the environmentally spawned chunks use the convex shapes, and not the microbes. Still that may be related to this if the shapes *are* used and I don't realize it. I'll open a separate issue about that: https://github.com/Revolutionary-Games/Thrive/issues/3022

hhyyrylainen:
I've tested with one place where I think the convex collision shapes were loaded commented out. That did have some FPS impact (as much as some of the other things I've tried) but it didn't make a save with a huge amount of lag playable.

One more thing I found when looking into this was that currently the amount of cell chunks to spawn is HexCount / CORPSE_CHUNK_DIVISOR which means that larger cells have linearly scaling more corpse chunks spawned when they die. I think we could have a max cap of 10 there and a power formula for quickly diminishing how many chunks a larger cell drops when it dies. Also a global cap for the number of cell chunks might be good (and as I mentioned in the issue description it seems that the cell chunks are not despawned when you get far away from them, seems like they maybe aren't added to the spawned group).

[...]

PR #3017, opened 2022-01-14
Lowered the max entities count a lot again

hhyyrylainen:
Now I only occasionally get drops to below 30 FPS and basically no time below 10. That seems maybe related to how many cells have died and how many floating chunks there are? That is an aspect regarding the performance I haven't investigated yet.

Doesn't impact the entity limit in existing saves.

**Related Issues**

Related to https://github.com/Revolutionary-Games/Thrive/issues/3016 (doesn't fix it)

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -462,7 +462,8 @@ public static class Constants
     public const float GLUCOSE_REDUCTION_RATE = 0.8f;
     public const float GLUCOSE_MIN = 0.0f;
 
-    public const int DEFAULT_MAX_SPAWNED_ENTITIES = 300;
+    // TODO: bump this back up once we resolve the performance bottleneck
+    public const int DEFAULT_MAX_SPAWNED_ENTITIES = 90;
     public const int MAX_SPAWNS_PER_FRAME = 1;
     public const int MAX_DESPAWNS_PER_FRAME = 1;
 


RacerBG (approved):
This definitely improve the perfomance of Thrive by a lot. I'm in a patch with 14 species including mine and the game started losing FPS after generation 12 (it went below 30 for a brief moment). But usually it stayed at 40-50 FPS all the time which is great, considering that I was in Freebuild Mode which by definition lags more than a normal Thrive game.

But I'm even more suspicious now what is the core cause of the lag. While the FPS was improved the game was still freezing for around 1 ms (the sound and the cell movement obviously were stuttering) after just a few generations but the FPS was stable. This pattern was observed in the earlier versions of the game as well.

This PR did it's job but Thrive have deeper lag issues in my opinion.

hhyyrylainen:
I think the microbe spawning is to blame for that, I think the membrane generation code takes quite a while so each time a cell spawns it needs to recompute the membrane for it and that causes lag. I think someone suggested caching the membranes for the base states of species, which I think is probably something we should add to try to reduce the minor lag spikes.

RacerBG:
> I think the microbe spawning is to blame for that, I think the membrane generation code takes quite a while so each time a cell spawns it needs to recompute the membrane for it and that causes lag. I think someone suggested caching the membranes for the base states of species, which I think is probably something we should add to try to reduce the minor lag spikes.

If that's the case this, combined with your PR, should mitigate most of the complaints from Steam.

Kasterisk (approved):

PR #3025, opened 2022-01-16
Fix window dialog drag expanding to the opposite direction

84634E1A607A:

This PR:

https://user-images.githubusercontent.com/60734649/149648278-b3937eba-32d0-4e03-9904-b3295a375338.mp4

Master:

https://user-images.githubusercontent.com/60734649/149648282-de399a21-6e3f-48fd-96b3-2d679166fc3f.mp4

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/gui_common/CustomDialog.cs
```diff
@@ -436,31 +436,32 @@ private void HandleActiveDrag()
         else
         {
             // Handle border dragging
+            var screenSize = (GetViewport()?.GetVisibleRect().Size).GetValueOrDefault();
```
hhyyrylainen, line 439: I think we should print an error an exit this method if the viewport doesn't exist. I'm pretty sure I've done pretty hard errors in the past about missing critical objects. The viewport can only be missing if this isn't attached to the scene tree so this wouldn't be able to receive the _GuiInput event anyway in that case.

hhyyrylainen (changes requested):

src/gui_common/CustomDialog.cs
84634E1A607A: I just copied this from the function below. So maybe we can just remove it.

hhyyrylainen: Probably. Might also want to move this to a helper method as this is pretty dense logic that doesn't make much sense if it is edited in one place and not the other (even if that happens accidentally).

Kasterisk (approved):
Nothing mildly concerning other than what's mentioned by hh.

RacerBG (approved):
LGTM!

84634E1A607A:
I just realized I should have used `Mathf.Clamp` instead of nesting `Mathf.Max` and `Mathf.Min`. Will change that later.

hhyyrylainen (approved):
I did a quick test and at least the auto-evo predictions popup box worked fine. Code looks fine now and I assume that the people who I requested re-reviews would be fine as well with this still even though they didn't re-review.

PR #3026, opened 2022-01-16
Increased the min horizontal size of the metabolosomes organelle tooltip title

Kasterisk:

Increased the label's min horizontal size to 210.

**Related Issues**

The recent thrive font version upgrade may have broken some labels wrapping due to the slight size difference, which in this case would be the metabolosomes tooltip title.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/tooltip/ToolTipManager.tscn
+++ b/src/gui_common/tooltip/ToolTipManager.tscn
@@ -365,7 +365,8 @@ margin_bottom = 189.0
 
 [node name="metabolosome" parent="GroupHolder/organelleSelection" instance=ExtResource( 5 )]
 visible = false
-margin_bottom = 173.0
+margin_right = 500.0
+margin_bottom = 2.0
 DisplayName = "METABOLOSOMES"
 ProcessesDescription = "METABOLOSOMES_PROCESSES_DESCRIPTION"
 Description = "METABOLOSOMES_DESCRIPTION"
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
Shouldn't all of the min widths be increased? I think there was one other organelle that had the same problem now as metabolosomes.

Kasterisk:
That was the chemosynth, yeah I guess it can get the same fix. No other labels has the text cut off abruptly other than these two.

Kasterisk:
Here's how it looks like now:

RacerBG (approved):
I tested it both in English and Bulgarian, and it looks good.

Unrelated to this PR I noticed that the tooltip in the behaviour tab have wrong coordinates relative to the mouse and a slight oversight on my part - the dropdown menu from PR "Changed report tab patch name label into a dropdown list of patches" does not disappear when a new patch is selected from it.

hhyyrylainen (approved):

Kasterisk:
> Unrelated to this PR I noticed that the tooltip in the behaviour tab have wrong coordinates relative to the mouse and a slight oversight on my part - the dropdown menu from PR "Changed report tab patch name label into a dropdown list of patches" does not disappear when a new patch is selected from it.

Those were all intended actually, the tooltips are correctly positioned at the bottom right corner relative to the ui element. The reason for the latter is in case the player wants to skim through multiple patches, so having it disappear for each selection would be slightly annoying.

PR #3049, opened 2022-01-22
Disabled value bar scroll

### LINKED ISSUE #3037: Make our custom colour picker set non-scrollable property on the colour value bars
otherwise trying to scroll over them causes the values to change.

Noticed this is especially bad in the organelle modify menu for the chemoreceptor, and causes warnings about a control that can't grab focus

84634E1A607A:
**Related Issues**

closes #3037 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/TweakedColourPicker.cs
+++ b/src/gui_common/TweakedColourPicker.cs
@@ -226,7 +226,9 @@ public class TweakedColourPicker : ColorPicker
         PresetsVisible = presetsVisible;
         OnColourChanged(Color);
 
-        // Disable value bar scroll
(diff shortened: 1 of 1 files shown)


hhyyrylainen (approved):
Seems like the correct elements no longer react to mouse scroll, should work fine for the chemoreceptor GUI issue.

84634E1A607A:
Regarding the comment update, I don't think we should export this as property. Because I think all the color pickers should be consistent on this function. Either all can scroll or all cannot.

hhyyrylainen:
It would be consistent with the sliders as you can select per element if it can be scrolled or not. If there is a colour picker that is *not* in a scrollable container, it might be nice to be able to set a different value there.

PR #3055, opened 2022-01-28
Implement Color picker pick with custom logic to workaround Godot bug

### LINKED ISSUE #2803: Godot bug: Colour picker doesn't work when window is not default sized
I think we should remove this color picker: 
image
It's broken and useless

84634E1A607A:
Rewrote color picker's pick button function. Temp fix for Godot issue https://github.com/godotengine/godot/issues/57343

**Related Issues**
Fixes #2803

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/gui_common/TweakedColourPicker.cs
```diff
@@ -222,6 +227,15 @@ public override void _Ready()
         presetsContainer = GetNode<GridContainer>("PresetContainer");
         addPresetButton = GetNode<TextureButton>("PresetButtonContainer/AddPresetButton");
 
+        // Picker button logic rewrite #3055
+        // TODO: Revert this PR once https://github.com/godotengine/godot/issues/57343 is solved.
+        baseControl = GetChild(1);
+        var customPickerButton = new ToolButton { Icon = pickerButton.Icon };
+        baseControl.RemoveChild(pickerButton);
```
hhyyrylainen, line 234: Is it possible to overwrite the default connection (so unconnect first) and then just link us to the default button? That would mean we mess with Godot's internals less here.

hhyyrylainen (changes requested):
I experienced the same kind of major lag as @RacerBG when I increased the window size a lot. I think the read from the rendered image starts taking up too long and causes a long queue of input events to process. To solve this, I think we should just sample the colour under the mouse in _Process every 100-200 milliseconds so that we aren't reading so much data from the GPU and converting that over to data readable from C#.

src/gui_common/TweakedColourPicker.cs
hhyyrylainen: I guess we don't know the connected object and method name so we can't actually call `Disconnect` here.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -42,6 +42,9 @@ public class TweakedColourPicker : ColorPicker
     private bool rawButtonEnabled = true;
     private bool presetsEditable = true;
     private bool presetsVisible = true;
+    private bool pickingColor;
+    private Color colorBeforePicking;
+    private float pickerTimeElapsed;
     private PickerMode mode;
```
hhyyrylainen, line 49:
```suggestion
    private bool presetsVisible = true;
    private PickerMode mode;

    private bool pickingColor;
    private Color colorBeforePicking;
    private float pickerTimeElapsed;

```

src/gui_common/TweakedColourPicker.cs
```diff
@@ -452,6 +500,32 @@ private void OnHtmlColourEditFocusExited()
...
+    {
+        colorBeforePicking = Color;
+        pickingColor = true;
+    }
+
+    private void HandleActiveColourPicking(float delta)
+    {
+        pickerTimeElapsed += delta;
+
+        if (pickerTimeElapsed < 0.2)
```
hhyyrylainen, line 513:
```suggestion
        if (pickerTimeElapsed < 0.2f)
```

hhyyrylainen (approved):
Seems to perform pretty well even at 4K now, which should be good enough.

Kasterisk (approved):

RacerBG (commented):
The lag is gone. Now it is around 1 FPS compared 50-70 FPS reduction in my first test.

But I noticed another problem: Now when I stop the cursor movement, while I'm still in color picker mode, the "cross" (the horizontal and vertical lines) continue to loop over the colors. Also while moving the cursor, the cross lags behind it even though the game is not lagging.

hhyyrylainen:
> But I noticed another problem: Now when I stop the cursor movement, while I'm still in color picker mode, the "cross" (the horizontal and vertical lines) continue to loop over the colors. Also while moving the cursor, the cross lags behind it even though the game is not lagging.

Isn't that just how it works anyway? If you hover over the colour picker itself it ends up updating itself and then reading the updated value causing weirdness.
The mouse lag is an unfortunate side effect of only updating 4 times per second. I suppose we could read the screen texture a few times a second and read from that whenever we want (which should also fix the lag), but that would be harder to implement and I don't think we need to spend the time to code that.

hhyyrylainen:
I'll merge this because this makes the situation much better (and I think fixes a crash). The colours going crazy is a problem with Godot's builtin colour picker as well. If someone really wants the last part of my previous comment can be done by a separate PR.

RacerBG:
> > But I noticed another problem: Now when I stop the cursor movement, while I'm still in color picker mode, the "cross" (the horizontal and vertical lines) continue to loop over the colors. Also while moving the cursor, the cross lags behind it even though the game is not lagging.
> 
> Isn't that just how it works anyway? If you hover over the colour picker itself it ends up updating itself and then reading the updated value causing weirdness. The mouse lag is an unfortunate side effect of only updating 4 times per second. I suppose we could read the screen texture a few times a second and read from that whenever we want (which should also fix the lag), but that would be harder to implement and I don't think we need to spend the time to code that.

I don't know if we are talking about the same thing. I did this:

1. Switched in to color picker mode.
2. Moved the cursor over the colors window or square in a fast manner.
3. Stopped the movement while still the cursor was in the boundaries of the colors window.
4. The cross started to loop over the current cursor x, y position and the last cursor x, y position endlessly.

hhyyrylainen:
That "oscillation" is probably caused by the colour being updated only like 5 times per second so with the fast movement it sees the previous colour, changes to that, and then the new colour ends up being sampled as the previous data was still in use, or something. The move to storing the window render buffer and updating the picked colour from it immediately should fix that, but as I said, I don't think this is a big enough problem to warrant a lot of effort to fix. The oscillation stops if you move the cursor over another element than the colour picker, right?

RacerBG:
> That "oscillation" is probably caused by the colour being updated only like 5 times per second so with the fast movement it sees the previous colour, changes to that, and then the new colour ends up being sampled as the previous data was still in use, or something. The move to storing the window render buffer and updating the picked colour from it immediately should fix that, but as I said, I don't think this is a big enough problem to warrant a lot of effort to fix. The oscillation stops if you move the cursor over another element than the colour picker, right?

Right. And the same behaviour doesn't happen when I try to select a color from the cell surroundings for example.

PR #3065, opened 2022-02-01
Fixed changing organelle rotation in place

### LINKED ISSUE #2993: Organelle move's same-position check doesn't take into account organelle rotation
Doing a "move" action for a hex but not exactly moving it, however with a different rotation just outright cancels the action when you finishes the move action. The common expectation would be for it to be accepted as a valid move since the rotation has changed (and costs no MP as per the code documentation).

Kasterisk:

Added rotation equals check in MoveOrganelle's anti-exploit position equals check so that rotating an organelle in place actually rotates the organelle into that particular angle when you finishes the action.

Steps to reproduce the original issue
- Place pilus
- Start organelle move action
- Rotate it in place in a different angle
- Place it
- It reverts back to the previous angle

**Related Issues**

Fixes #2993 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/MicrobeEditor.cs
+++ b/src/microbe_stage/editor/MicrobeEditor.cs
@@ -2113,10 +2113,11 @@ public class MicrobeEditor : NodeWithInput, ILoadableGameState, IGodotEarlyNodeR
             return false;
         }
 
-        // Don't register the action if the final location is the same as previous. This is so the player can't exploit
-        // the MovedThisSession flag allowing them to freely move an organelle that was placed in another session
-        // while on zero mutation points. Also it makes more sense to not count that organelle as moved either way.
-        if (oldLocation == newLocation)
+        // Don't register the action if the final location and rotation is the same as previous. This is so the player
+        // can't exploit the MovedThisSession flag allowing them to freely move an organelle that was placed in another
+        // session while on zero mutation points. Also it makes more sense to not count that organelle as moved either
+        // way.
+        if (oldLocation == newLocation && oldRotation == newRotation)
         {
             CancelCurrentAction();
 


hhyyrylainen (changes requested):
This allows a free move exploit.

If I move an organelle to the exact same position and rotation, then it doesn't take MP and the next move takes -5MP like usually, however if I rotate the organelle and place it at the current location, this makes the next move free 0MP, even though it should cost MP.

Kasterisk:
I don't know this seems like an underlying flaw with the free move logic itself. So as far as I understand it, a move action is free if it satisfies these three conditions: an organelle is moved within current session, added within current session or rotated in place (as per the docs). Looks like the former and latter is inherently exploitable, while I've fixed the first one being exploitable in #2400, this bugfix-fix seems to unmask the problem with the last.

hhyyrylainen:
The logic should be that placing the organelle at the exact place it is at, should be free (and basically cancel the move action). And any other move (the first time when done in an editor session for an organelle) should cost MP. Perhaps rotating in-place should also cost MP? This PR probably needs reworking once https://github.com/Revolutionary-Games/Thrive/pull/2445 is done.

Kasterisk:
> Perhaps rotating in-place should also cost MP?

No problem for me.

> This PR probably needs reworking once https://github.com/Revolutionary-Games/Thrive/pull/2445 is done.

Well in that case, shouldn't it be better to fix the issue this PR is trying to solve in that PR (or the free move issue in general)? This can be closed instead.

hhyyrylainen:
Probably yeah, if you are up for looking at this again afterwards.
Btw it looks like currently the new PR will ignore in-place rotation when determining costs: https://github.com/Revolutionary-Games/Thrive/pull/2445/files#diff-b68cb07ad85aa6847e8f3075164777676512f49a460148d05e08701f87f9af4dR27

Kasterisk:
Yet another thing to consider in the code @FuJa0815. I'll close this in the meantime.

PR #3068, opened 2022-02-04
Moved to using nullable references

### LINKED ISSUE #3052: Convert code to use nullable reference types and fix warnings
Nullable reference types is a compiler option to make the C# compiler distinguish between reference types that can and can't be null.
https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references

After trying it out with the DevCenter it's a bit of a pain to setup but in terms of code robustness knowing when something can be null is a huge benefit, I think.

hhyyrylainen:

Nullable references is a C# 8.0 and later feature where you can specify if something can be null or not. This is very useful when defining methods so the language syntax will already tell the programmer what parameters are allowed to be null and what must always be provided.

I'd especially like feedback on the styleguide and the cases where I decided to suppress initial value nullness with `null!`.

Also includes a bunch of small style fixes that I did while basically having to look over the warnings in each of the game files.

**Related Issues**

closes #3052

**Progress Checklist**

      break existing features:
      members reviewing this PR)


doc/style_guide.md
```diff
@@ -144,6 +144,35 @@ Code style rules
   reading the code that they couldn't see directly from the method
   signature.
 
+- For the main Thrive.csproj 
+  [nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references) 
+  feature is turned on. All nullable reference warnings need to be
```
FuJa0815, line 149:
```suggestion
  are enabled. All nullable reference warnings need to be 
```

src/engine/StartupActions.cs
```diff
@@ -25,7 +26,15 @@ private StartupActions()
             " latest log is 'log.txt'");
 
         // Load settings here, to make sure locales etc. are applied to the main loaded and autoloaded scenes
-        if (Settings.Instance == null)
-            GD.PrintErr("Failed to initialize settings.");
+        try
+        {
+            // We just want to do something to ensure settings instance is fine
+            // ReSharper disable once UnusedVariable
+            var hashCode = Settings.Instance.GetHashCode();
```
FuJa0815: Isn't it possible to just call the method and not save the return value in any variable to avoid that UnusedVariable warning?

src/engine/input/key_mapping/InputActionItem.cs
```diff
@@ -86,6 +93,15 @@ public InputGroupList GroupList
     /// </summary>
     public override void _Ready()
     {
+        if (string.IsNullOrEmpty(InputName))
+            throw new InvalidOperationException($"{nameof(DisplayName)} can't be empty");
```
FuJa0815:
```suggestion
            throw new InvalidOperationException($"{nameof(InputName)} can't be empty");
```

FuJa0815 (changes requested):
Didn't look at everything yet, but I like the overall style.

src/engine/StartupActions.cs
hhyyrylainen: No, that gives a different warning. (return value of pure function is not used)

src/engine/input/key_mapping/InputActionItem.cs
hhyyrylainen: I'll fix that.

doc/style_guide.md
hhyyrylainen: Any other thoughts on this from someone else? I don't think strictly speaking what I wrote is incorrect grammatically or conceptually.

hhyyrylainen:
Fixed the crashes I found with this.

doc/style_guide.md
adQuid: "is turned on" is more colloquial, but nobody will misunderstand it. I'd lean slightly towards "are enabled", but it's small potatos.

simulation_parameters/SimulationParameters.cs
```diff
@@ -314,7 +314,7 @@ private static string ReadJSONFile(string path)
         return result;
     }
 
-    private Dictionary<string, T> LoadRegistry<T>(string path, JsonConverter[] extraConverters = null)
+    private Dictionary<string, T> LoadRegistry<T>(string path, JsonConverter[]? extraConverters = null)
```
adQuid: Would empty lists work here? This seems like a case for a non-nullable parameters.

Also applies to methods below.

hhyyrylainen: I would guess it would, but I had over 2000 warnings to fix so I couldn't spend any time thinking about clear cut cases like this where the parameter is obviously meant to be nullable as `null` was specified as the default value.

PR #3090, opened 2022-02-12
Added a cloud brightness multiplier that is used in brighter biomes

### LINKED ISSUE #2561: Increase compound cloud brigthness in brighter biomes
Increase the compound cloud brightness values in the brighter biomes as a stopgap until the issue to make clouds look really good is done (https://github.com/Revolutionary-Games/Thrive/issues/1701)

Can probably put a brightness multiplier in biomes.json that is then on patch change passed to the cloud system for it to multiply things when creating the texture data. One slight difficulty may be dealing with saving, but I think the cloud system already saves some info so adding the multiplier to also be saved should be doable.

Another approach could be to have per-biome compound colours as a fix for this.

hhyyrylainen:
so that clouds are easier to see there

Before:

After:

I manually eyeballed just some values for the biomes, it'll be just editing the biomes json file if these values need further tweaks either now or in the future.

**Related Issues**

closes #2561

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Kasterisk (approved):
A welcome change.

RacerBG:
> A welcome change.

Indeed but I want to run a few test cases first. I have a suspicion that a few clouds might need to be brighter.

RacerBG (changes requested):
The other biomes needs some rework as a result of those changes.

hhyyrylainen:
How so they are unaffected by multiplying things by 1?
If there are suggestions for values to put in the json I'll gladly accept them.

hhyyrylainen:
I'll leave further tweaking to someone else, at least this will address the long term complain about the brighter patches being pretty unplayable in terms of cloud finding.

Issue #3035, opened 2022-01-21
Chemoreceptor sometimes doesn't detect a nearby cloud


hhyyrylainen:
Might be related to: https://github.com/Revolutionary-Games/Thrive/issues/1924

I noticed that with certain positioning the chemoreceptor doesn't detect a cloud that is on screen, but when moving a bit then it starts happily pointing to it.

PR #3093, opened 2022-02-13
Ai uses chemoreceptors

### LINKED ISSUE #3033: Add some benefit for AI cells from a chemoreceptor
currently the AI gets no benefit from the cells having a chemoreceptor

adQuid:

Allows AI to use chemoreceptors to seek out valuable compounds. I went with a pretty subtle behavior change, and added a new directive for spreading out. This should help both over-swarming due to lots of microbes rushing the same cloud, and finally give high activity species a distinct behavior.

closes #3033

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Microbe.cs
+++ b/src/microbe_stage/Microbe.cs
@@ -585,7 +585,7 @@ public partial class Microbe : RigidBody, ISpawned, IProcessable, IMicrobeAI, IS
 
             if (detectedCompound != null)
             {
-                detections.Add((compound, colour, (Vector3)detectedCompound));
+                detections.Add((compound, colour, detectedCompound.Value));
             }
         }
 
--- a/src/microbe_stage/Microbe.cs
+++ b/src/microbe_stage/Microbe.cs
@@ -567,6 +567,15 @@ public partial class Microbe : RigidBody, ISpawned, IProcessable, IMicrobeAI, IS
         }
     }
 
+    /// <summary>
+    ///   Returns a list of tuples, representing all possible compound targets. These are not all clouds that the microbe
+    ///   can smell; only the best candidate of each compound type.
+    /// </summary>
+    /// <param name="clouds">CompoundCloudSystem to scan</param>
+    /// <returns>
+    ///   A list of tuples. Each tuple contains the type of compound, the color of the line (if any needs to be drawn),
+    ///   and the location where the compound is located.
+    /// </returns>
     public List<(Compound Compound, Color Colour, Vector3 Target)> GetDetectedCompounds(CompoundCloudSystem clouds)
     {
         var detections = new List<(Compound Compound, Color Colour, Vector3 Target)>();
--- a/src/microbe_stage/Microbe.cs
+++ b/src/microbe_stage/Microbe.cs
@@ -567,6 +567,22 @@ public partial class Microbe : RigidBody, ISpawned, IProcessable, IMicrobeAI, IS
         }
     }
 
+    public List<(Compound Compound, Color Colour, Vector3 target)> GetDetectedCompounds(CompoundCloudSystem clouds)
+    {
+        var detections = new List<(Compound Compound, Color Colour, Vector3 target)>();
+        foreach (var (compound, range, minAmount, colour) in activeCompoundDetections)
+        {
+            var detectedCompound = clouds.FindCompoundNearPoint(Translation, compound, range, minAmount);
+
+            if (detectedCompound != null)
+            {
+                detections.Add((compound, colour, (Vector3)detectedCompound));
+            }
+        }
+
+        return detections;
+    }
+
     /// <summary>
     ///   This method calculates the relative rotation and translation this microbe should have to its microbe parent.
     ///   <a href="https://randomthrivefiles.b-cdn.net/documentation/fixed_colony_rotation_explanation_image.png">
--- a/src/microbe_stage/Microbe.cs
+++ b/src/microbe_stage/Microbe.cs
@@ -568,8 +568,8 @@ public partial class Microbe : RigidBody, ISpawned, IProcessable, IMicrobeAI, IS
     }
 
     /// <summary>
-    ///   Returns a list of tuples, representing all possible compound targets. These are not all clouds that the microbe
-    ///   can smell; only the best candidate of each compound type.
+    ///   Returns a list of tuples, representing all possible compound targets. These are not all clouds that the
+    ///   microbe can smell; only the best candidate of each compound type.
     /// </summary>
     /// <param name="clouds">CompoundCloudSystem to scan</param>
     /// <returns>
--- a/src/microbe_stage/Microbe.cs
+++ b/src/microbe_stage/Microbe.cs
@@ -567,9 +567,9 @@ public partial class Microbe : RigidBody, ISpawned, IProcessable, IMicrobeAI, IS
         }
     }
 
-    public List<(Compound Compound, Color Colour, Vector3 target)> GetDetectedCompounds(CompoundCloudSystem clouds)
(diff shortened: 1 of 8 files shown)


GodOfKnowledge:
Thought of this while playing, glad to see it being worked on.

hhyyrylainen:
If I'm reading the diff correctly then this PR (accidentally) downgrades the used RubySetupSystem version.

src/microbe_stage/MicrobeAISystem.cs
```diff
@@ -10,9 +10,12 @@ public class MicrobeAISystem
 
     private readonly Node worldRoot;
 
-    public MicrobeAISystem(Node worldRoot)
+    private readonly CompoundCloudSystem clouds;
```
hhyyrylainen, line 13: Could you add a cautionary comment here as the AI is run in threaded mode, the AI may *not* call any state changing methods on the cloud system (only data reading methods are allowed).

src/microbe_stage/Microbe.cs
```diff
@@ -567,6 +567,31 @@ public override void _IntegrateForces(PhysicsDirectBodyState physicsState)
...
+    public List<(Compound Compound, Color Colour, Vector3 Target)> GetDetectedCompounds(CompoundCloudSystem clouds)
+    {
+        var detections = new List<(Compound Compound, Color Colour, Vector3 Target)>();
+        foreach (var (compound, range, minAmount, colour) in activeCompoundDetections)
+        {
+            var detectedCompound = clouds.FindCompoundNearPoint(Translation, compound, range, minAmount);
+
+            if (detectedCompound != null)
+            {
+                detections.Add((compound, colour, (Vector3)detectedCompound));
```
hhyyrylainen, line 588:
```suggestion
                detections.Add((compound, colour, detectedCompound.Value));
```
Sorry, if this suggestion is wrong, I'm making this suggestion assuming the type here is `Vector3?`

src/microbe_stage/MicrobeAI.cs
```diff
@@ -395,6 +398,60 @@ private void EngagePrey(Vector3 target, Random random, bool engulf)
...
+                SmellForComounds(data);
+            }
+        }
+        else
+        {
+            RunAndTumble(random);
+        }
+    }
+
+    private void SmellForComounds(MicrobeAICommonData data)
```
hhyyrylainen, line 436:
```suggestion
    private void SmellForCompounds(MicrobeAICommonData data)
```

src/microbe_stage/MicrobeAI.cs
```diff
@@ -395,6 +398,60 @@ private void EngagePrey(Vector3 target, Random random, bool engulf)
...
+    {
+        ComputeCompoundsSearchWeights();
+
+        var detections = microbe.GetDetectedCompounds(data.Clouds)
```
hhyyrylainen, line 440: Shouldn't the AI have some kind of knowledge of when it *doesn't* have a chemoreceptor to not attempt to use it?

adQuid: I would argue that this is better separation of concerns. The Microbe class, which owns the detections, is capable of answering the question, "what do you smell?" by running a loop zero times and returning a list of zero items. Adding a check here would just be an extra place to break that doesn't protect against anything. You will notice that, so far, the AI makes no checks about the microbe's organelles.

adQuid:
> If I'm reading the diff correctly then this PR (accidentally) downgrades the used RubySetupSystem version.

huh, I'll work on that when I get some time.

src/microbe_stage/MicrobeAI.cs
hhyyrylainen: *cough* *cough* I just finished the signaling agents, and they *do* check a property on the Microbe to see if it has signaling agents. I think such a check is similar to checking if the memberane set on the Microbe allows engulfing or not, which is directly checked in the AI. I think that's less clear separation of concerns, a cleaner way would be for the AI to ask the Microbe class if it can engulf instead of reading the membrane type and comparing that.

adQuid: I agree, and maybe that would be a good future refactor. This code asks the Microbe class if it has any active detections instead of reading the organelle list and comparing that. Detections are currently only created when there are chemoreceptors and they smell something. If later on that changes, the change won't break this code.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -462,6 +465,61 @@ private void EngagePrey(Vector3 target, Random random, bool engulf)
...
+        if (lastSmelledCompoundPosition != null)
+        {
+            if (DistanceFromMe(lastSmelledCompoundPosition.Value) > 3.0f)
+            {
+                targetPosition = lastSmelledCompoundPosition.Value;
+                microbe.LookAtPoint = targetPosition;
+            }
+            else
+            {
+                SmellForCompounds(data);
```
adQuid, line 494:
```suggestion
            {
                SetMoveSpeed(0.0f);
                SmellForCompounds(data);
```

adQuid:
> It's pretty cool to see the AI roam around with real aim and intent, great job!
> 
> Two things I noticed:
> 
> 1. The AI occasionally seemed to have a bit of a delay between their usual wandering and noticing that there's a compound they want. (Even after they had passed it) I tried playing around with behavior sliders to see if they would respond quicker but to seemingly no avail. This is pretty minor though.
> 2. The AI will occasionally spin in place on a cloud for a while until something else gets their attention. It's kind of funny and cute in a way. We could probably avoid this from happening for too long by making cells move on if they have enough compounds.

Point number 1 is definitely the result of the code not smelling for compounds every tick with the AI to save on computations. I have considered refactoring the chemorecepter algorithm to use ray tracing, but that would expand the scope of this PR by a good deal, so I was trying to avoid that.

I made a quick remote change to try and fix number 2. Let me know how it works if you have time, but I won't get a chance to test it for a while.

hhyyrylainen:
> I have considered refactoring the chemorecepter algorithm to use ray tracing, but that would expand the scope of this PR by a good deal, so I was trying to avoid that.

I was first like "wait what?" when I read this, but now thinking about it, this does actually make sense and probably could improve chemoreceptor performance quite a lot (and maybe even fix https://github.com/Revolutionary-Games/Thrive/issues/3035).

buckly90:
Nice change, AI feels a little smarter now.

I did notice that a cell ended up stuck on an iron chunk as it was trying to reach a cloud on the other side... I think that could be remedied by making cells back up and turn when they collide against objects based on responsiveness or something. But maybe that's best left for another PR.

hhyyrylainen:
I think perhaps the cells should only some of the time follow the signaling agents anyway, otherwise I think it breaks away a bit too far from the real science run and tumble behaviour...

adQuid:
Alright, I think I've made a compromise now. Now the AI uses similar logic with chasing compounds as with chasing prey, "if I'm not still getting closer, time to give up" I kind of want to add this logic to chunk-seeking too, but that will be technically tougher and beyond this PR. This doesn't implement true pathing; microbes will still bump into stuff, and don't know what is or isn't a wall until they try to swim through it. Personally, I like that approach more, and on a practical level pathing will require pretty much scrapping most of this class.

HH, I think you'll find that the microbes currently are pretty inconsistent with actually going after every chemoreceptor signal (which Buckly mentioned not liking, actually). Remember that right now microbes only smell every 6 seconds on average. Adding a chemoreceptor to a microbe certainly makes a species feel more "aware" and intelligent, but they still seem pretty dumb and primative compared to the player, which seems right to me.

hhyyrylainen (approved):
Code is now fine, I guess we also had good enough results from playtesting to merge this now.

PR #3097, opened 2022-02-17
Added signaling agents

### LINKED ISSUE #2364: Add signaling agents
Signaling agents is a new organelle that allows the player to give simple commands to their species.
https://forum.revolutionarygamesstudio.com/t/make-agent-binding-available-at-a-distance/797/2

- [x] placeable organelle added (with model and icon, moved these to a separate issue)
- [x] key added that opens a command menu in the microbe stage where the command can be selected
- [x] microbe AI follows the commands a species member of them is giving

hhyyrylainen:

adds signaling agents

requires someone else to make an icon and model + texture, but I'll merge even without those

**Related Issues**

closes #2364

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/MicrobeAI.cs
+++ b/src/microbe_stage/MicrobeAI.cs
@@ -53,6 +53,15 @@ public class MicrobeAI
     [JsonIgnore]
     private Dictionary<Compound, float> compoundsSearchWeights;
 
+    [JsonIgnore]
+    private float timeSinceSignalSniffing;
+
+    [JsonIgnore]
+    private EntityReference<Microbe> lastFoundSignalEmitter = new();
+
+    [JsonIgnore]
+    private MicrobeSignalCommand receivedCommand = MicrobeSignalCommand.None;
+
     public MicrobeAI(Microbe microbe)
     {
         this.microbe = microbe ?? throw new ArgumentException("no microbe given", nameof(microbe));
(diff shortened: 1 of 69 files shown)


src/microbe_stage/MicrobeAI.cs
```diff
@@ -67,21 +76,42 @@ public MicrobeAI(Microbe microbe)
...
-    private float SpeciesFear => microbe.Species.Behaviour.Fear;
-    private float SpeciesActivity => microbe.Species.Behaviour.Activity;
+    private float SpeciesAggression => microbe.Species.Behaviour.Aggression *
+        (receivedCommand == MicrobeSignalCommand.BecomeAggressive ? 1.5f : 1.0f);
+
+    private float SpeciesFear => microbe.Species.Behaviour.Fear *
+        (receivedCommand == MicrobeSignalCommand.BecomeAggressive ? 0.75f : 1.0f);
+
+    private float SpeciesActivity => microbe.Species.Behaviour.Activity *
+        (receivedCommand == MicrobeSignalCommand.BecomeAggressive ? 1.25f : 1.0f);
```
adQuid: While I'm still not a big fan of such strong signaling agents, if you want to go with the paradigm of "when under a command, the microbe ALWAYS follows it", my guess is you will need to go way bigger with your multipliers. If a player has a non-aggressive and fearful species, it's not going to attack something bigger than itself unless you multiply and divide by 10 or more. 

Even then, keep in mind that the minimum of each stat is 0, so if a player slides all the way to the left signaling agents won't matter. I don't know if that's intended functionality.

src/microbe_stage/MicrobeAISystem.cs
```diff
@@ -76,8 +76,11 @@ private void RunAIFor(IMicrobeAI? ai, float delta, Random random, MicrobeAICommo
         if (ai.TimeUntilNextAIUpdate > 0)
             return;
 
+        // TODO: would be nice to add a tiny bit of randomness to the times here so that not all cells think at once
         ai.TimeUntilNextAIUpdate = Constants.MICROBE_AI_THINK_INTERVAL;
 
-        ai.AIThink(delta, random, data);
+        // As the AI think interval is made constant, we pass that value as the delta to make time keeping be actually
+        // (mostly) consistent in the AI code
+        ai.AIThink(Constants.MICROBE_AI_THINK_INTERVAL, random, data);
```
adQuid: I'm unsure that this will be accurate. We only have assurance that time since last interval is at least MICROBE_AI_THINK_INTERVAL, but it could have been significantly more, meaning an accurate time delta would be higher. I don't think that the AI actually uses that delta yet, but if we're passing it in it should be accurate.

hhyyrylainen: It does now for my signaling agents calculation.
I just made it better. Before it passed the delta for the *previous* frame. Which is way, way more incorrect here. I realize my solution is not perfect, and I did write that this is mostly right. As long as we don't have massive lag spikes the time elapsed should be consistent enough.

PR #3099, opened 2022-02-19
Try to trigger other cell disposed exception much earlier

hhyyrylainen:
in the binding process to try to finally stop partial binds from glitching out

**Related Issues**

related to the still happening binding issues.
https://www.reddit.com/r/thrive/comments/svcy7c/v057_bug_with_binding/

**Progress Checklist**

      break existing features:
      members reviewing this PR)


buckly90:
I tried my hardest to break something with binding agents but failed to do so. So at the very least it seems you didn't mess anything up.

hhyyrylainen:
Shape error indexes happen a lot anyway, it's most likely a bug with how Godot engine integrates the bullet physics (which they are not going to fix ever, instead they are writing their own physics engine for 4.0).

I'll consider it good enough that things aren't worse now. As is usually the case we'll just need to wait for the next release to see if the binding fixes are actually working or if we get more problem reports.

PR #3117, opened 2022-03-06
Optimize Cloud Search

adQuid:

Updates code to find compound clouds based on polar coordinates. It's possible for a chemoreceptor to fail to spot a very small cloud very far away if it hasn't had time to disperse yet, but that seems accurate to how smell really works.

This shouldn't noticeably change behavior, but the chemoreceptor is now performant at scale: https://www.youtube.com/watch?v=RYGc2CZ7DZI

**Related Issues**

(not sure if there are related issues to this, I'll check)

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -332,33 +330,25 @@ public void GetAllAvailableAt(Vector3 worldPosition, Dictionary<Compound, float>
             // relative origin point is calculated and that is restricted by
             // checking if the point is within the circle before grabbing
 
-            // And apparently there isn't an algorithm to generate the points with closest ones first? so
-            // this goes through everything and keeps the closest found point
-
-            int xEnd = (int)Mathf.Round(cloudRelativeX + localRadius);
-            int yEnd = (int)Mathf.Round(cloudRelativeY + localRadius);
+            // TODO: pull these to a constant to not constantly recalculate them
```
hhyyrylainen, line 333: TODO comment spotted

src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -332,33 +330,25 @@ public void GetAllAvailableAt(Vector3 worldPosition, Dictionary<Compound, float>
...
+            double fullCircle = Math.PI * 2;
+            double searchArc = fullCircle / 24.0;
 
-            for (int x = (int)Mathf.Round(cloudRelativeX - localRadius);
-                 x <= xEnd;
-                 x += 1)
+            for (int radious = 1;
```
hhyyrylainen, line 337: typo in `radius`?

src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -332,33 +330,25 @@ public void GetAllAvailableAt(Vector3 worldPosition, Dictionary<Compound, float>
...
+                 radious < localRadius;
+                 radious += 1)
             {
-                for (int y = (int)Mathf.Round(cloudRelativeY - localRadius);
-                     y <= yEnd;
-                     y += 1)
+                for (double theta = 0;
```
hhyyrylainen, line 341: Are doubles *actually* needed or would this work with floats? I guess either way this won't have much of an effect (maybe) as the float variant of math methods just convert to and back from doubles anyway...

adQuid: When dividing 2 pi by 24 (and someone may up that granularity in the future), then doing trig out at long distances, I think floats will start making some trouble.

hhyyrylainen: I suppose this is better to use it this way, then.

src/microbe_stage/CompoundCloudSystem.cs
adQuid: Is there anywhere else you want to move this to? It turned out to be unneeded for performance, so I'd just drop the TODO and leave it as is.

hhyyrylainen: Constants or MathUtils.

simulation_parameters/Constants.cs
```diff
@@ -362,6 +362,9 @@ public static class Constants
     public const float CHEMORECEPTOR_COMPOUND_UPDATE_INTERVAL = 0.25f;
     public const string CHEMORECEPTOR_DEFAULT_COMPOUND_NAME = "glucose";
 
+    // Size, in radians, of the gaps between directions the chemoreceptor checks for compounds
```
adQuid, line 365:
```suggestion
    /// <summary>
    ///     Size, in radians, of the gaps between directions the chemoreceptor checks for compounds
    /// </summary>
```

simulation_parameters/Constants.cs
```diff
@@ -362,6 +362,11 @@ public static class Constants
     public const float CHEMORECEPTOR_COMPOUND_UPDATE_INTERVAL = 0.25f;
     public const string CHEMORECEPTOR_DEFAULT_COMPOUND_NAME = "glucose";
 
+    /// <summary>
+    ///     Size, in radians, of the gaps between directions the chemoreceptor checks for compounds
```
hhyyrylainen, line 366:
```suggestion
    ///   Size, in radians, of the gaps between directions the chemoreceptor checks for compounds
```

hhyyrylainen:
Code looks fine now.

I have a slight dislike for how now the chemoreceptor line flickers on and off pretty quickly when moving around with small clouds around you. Maybe the check angles can be made smaller? I asked on discord if we can get more people to test this.

Also would the math be more efficient if you calculate the angles in the upper loop and then loop until the search distance (or maybe up to half the search distance first, and in a second loop up to the final distance so that checking can end early if something is found) that way you just need some float multiplies. Instead now you need to call the trigonometry functions once for each search unit distance, which to me sounds much more math that the CPU needs to process, even if this code can sometimes exit early by finding a nearby cloud. I think as quite often the chemoreceptor doesn't detect anything the case where it needs to search up to the max distance might be more common than being able to exit the loop way early.

Kemikal1:
I think its fine as is, didnt test much but functionality wise i cant see any difference from before, and performance is indeed increased.

simulation_parameters/Constants.cs
```diff
@@ -362,6 +362,11 @@ public static class Constants
     public const float CHEMORECEPTOR_COMPOUND_UPDATE_INTERVAL = 0.25f;
     public const string CHEMORECEPTOR_DEFAULT_COMPOUND_NAME = "glucose";
 
+    /// <summary>
+    ///   Size, in radians, of the gaps between directions the chemoreceptor checks for compounds
+    /// </summary>
+    public const double CHEMORECEPTOR_ARC_SIZE = Math.PI / 12.0;
```
adQuid, line 368:
```suggestion
    public const double CHEMORECEPTOR_ARC_SIZE = Math.PI / 24.0;
```

adQuid:
Do these changes satisfy you @hhyyrylainen?

hhyyrylainen (approved):
It's better now. I'd have liked to get a bit more testing from more people but I guess our testing people are unavailable currently. Let's just get this merged, people can then complain when 0.5.8 comes out if the chemoreceptor is not behaving as well as before.

hhyyrylainen:
https://github.com/Revolutionary-Games/Thrive/issues/3125

PR #3118, opened 2022-03-08
Patch Name Generator

### LINKED ISSUE #2153: Add place name generator to be used by the patch map generator
For the procedural map: https://github.com/Revolutionary-Games/Thrive/issues/682 we need procedural names.
For now just an English implementation is needed, but in the future we could have language specific generators.

See: https://github.com/Revolutionary-Games/Thrive/pull/2151#issuecomment-808895448

Kemikal1:

Adds random patch names to the game

**Related Issues**
closes #2153 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


simulation_parameters/microbe_stage/patch_syllabels.json
```diff
@@ -0,0 +1,42 @@
+{
+	"syllabels": [
```
adQuid, line 2:
```suggestion
	"syllables": [
```

simulation_parameters/SimulationParameters.cs
```diff
@@ -39,6 +39,7 @@ public class SimulationParameters : Node
     public AutoEvoConfiguration AutoEvoConfiguration => autoEvoConfiguration;
 
     public NameGenerator NameGenerator { get; private set; } = null!;
+    public PatchMapNameGenerator PatcMapNameGenerator { get; private set; } = null!;
```
adQuid, line 42:
```suggestion
    public PatchMapNameGenerator PatchMapNameGenerator { get; private set; } = null!;
```

src/microbe_stage/PatchMapNameGenerator.cs
```diff
@@ -0,0 +1,42 @@
...
+    private string patchName = null!;
+    [JsonRequired]
+    private List<string> syllabels = null!;
+    [JsonRequired]
+    private List<string> sufixes = null!;
+    public string InternalName{ get; set; } = null!;
+    public PatchMapNameGenerator(int lowerLimit, int higherLimit)
+    {
+        syllabelsLowerLimit = lowerLimit;
+        syllabelsHigherLimit = higherLimit;
```
adQuid, line 20:
```suggestion
        syllablesLowerLimit = lowerLimit;
        syllablesHigherLimit = higherLimit;
```

adQuid (dismissed):
Spelling: the low fruit of code reviews.

src/microbe_stage/PatchMapNameGenerator.cs
Kemikal1: as you can probably guess i really thought this was the correct form, i kept spelling it in my head but coudlnt figure it out

Kemikal1:
you mean the localized stringbuilder
?

hhyyrylainen:
Yes that's what I mean, LocalizedStringBuilder.

locale/en.po
```diff
@@ -538,7 +538,7 @@ msgstr "Big Iron Chunk"
 
 #: ../simulation_parameters/microbe_stage/biomes.json:148
 msgid "VOLCANIC_VENT"
-msgstr "Volcanic vent"
+msgstr "Volcanic Vent"
```
hhyyrylainen, line 541: Do we *need* to change these? I kinda don't think this is necessary. Anyone else have thoughts on this?

simulation_parameters/SimulationParameters.cs
```diff
@@ -94,6 +95,9 @@ public override void _Ready()
         gameCredits =
             LoadDirectObject<GameCredits>("res://simulation_parameters/common/credits.json");
 
+        PatchMapNameGenerator = LoadDirectObject<PatchMapNameGenerator>(
+            "res://simulation_parameters/microbe_stage/patch_syllabels.json");
```
hhyyrylainen, line 99: "syllable"

simulation_parameters/microbe_stage/patch_syllabels.json
```diff
@@ -0,0 +1,46 @@
...
+        "bre",
+        "ble",
+        "ca",
+        "che",
+        "cre",
+        "cle",
+        "ga",
+        "go"
+    ],
+    "sufixes": [
```
hhyyrylainen, line 38:
```suggestion
    "suffixes": [
```

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -190,4 +205,14 @@ private static void TranslatePatchNames()
         _ = TranslationServer.Translate("PATCH_ICE_SHELF");
         _ = TranslationServer.Translate("PATCH_PANGONIAN_SEAFLOOR");
     }
+
+    private static LocalizedStringBuilder GetPatchLocalizedName(string name, string localizedBiome)
```
hhyyrylainen, line 209: localizedBiome is not actually localized here, it should be named something like `biomeTranslationKey`

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -190,4 +205,14 @@ private static void TranslatePatchNames()
...
+    {
+        var localizedNames = new LocalizedStringBuilder();
+        localizedNames.Append(name);
+        localizedNames.Append(" ");
+        localizedNames.Append(new LocalizedString(localizedBiome));
```
hhyyrylainen, line 214: This part should be one entire localized string so that different languages can override this.
So this should be text like "PATCH_NAME" with the English translation having "{0} {1}" in it.

src/microbe_stage/PatchMapNameGenerator.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        // Convert first letter to uppercase
+        char[] charName = name.ToCharArray();
+        charName[0] = char.ToUpper(name[0]);
+        name = new string(charName);
+
+        return name;
+    }
+
+    // TODO
```
hhyyrylainen, line 77: TODO: comments should be removed.
This should actually check that syllablesHigherLimit is higher than lower limit and that the suffixes is not empty etc. conditions.

src/microbe_stage/PatchMapNameGenerator.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    public string GetPatchName()
+    {
+        return patchName;
+    }
+
+    /// <summary>
+    ///   Generates and returns a new patch name
+    /// </summary>
+    public string Next()
```
hhyyrylainen, line 39: This should allow passing in Random as a parameter so that one random can be used for the entire new game generation (so that we can move to allowing new save seeds at some point).

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -136,7 +150,8 @@ public static PatchMap Generate(WorldGenerationSettings settings, Species defaul
         };
         map.AddPatch(iceShelf);
 
-        var seafloor = new Patch("PATCH_PANGONIAN_SEAFLOOR", 10,
+        localizedName = GetPatchLocalizedName(name, "SEA_FLOOR").ToString();
```
hhyyrylainen, line 153: Converting this to string here will lose information. I think you should also modify the Patch class so that the name is a localized string in it. That way when game language is changed the "volcanic vents" part can correctly react to translations change.

This current approach also unnecessarily passes this string to the translation system so even if the above thing I mentioned is not done, that needs cleanup.

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -14,8 +14,13 @@ public static PatchMap Generate(WorldGenerationSettings settings, Species defaul
 
         var map = new PatchMap();
 
+        // Generate random name for patch
+        var nameGenerator = SimulationParameters.Instance.GetPatchMapNameGenerator();
+        var name = nameGenerator.Next();
+
         // Predefined patches
-        var vents = new Patch("PATCH_PANGONIAN_VENTS", 0,
+        string localizedName = GetPatchLocalizedName(name, "VOLCANIC_VENT").ToString();
```
hhyyrylainen, line 22: Instead of using an intermediate variable for the localized name I think you should just put the call in the patch constructor.

src/microbe_stage/PatchMapGenerator.cs
Kemikal1: I dont understand what  you mean

Kemikal1: you mean to make a single string and append it?

hhyyrylainen: `new LocalizedString("PATCH_NAME", name, new LocalizedString(localizedBiome))`
something like that. I think that'll work to make the order of the name parts work, and then there's no need to have the string builder here but can instead just directly return that new localized string from this method.

hhyyrylainen: In existing code it's used in for example `new LocalizedString("COMPOUND_FOOD_SOURCE", compound.Name);`

Kemikal1: now that i think about it, why would i do that since the only situation in the codebase the localized string is used with the string builder?

Kemikal1: using localizedStr.TroStr yields the translation key without the arguments, i think there needs to be used a localized Str builder nonetheles.

hhyyrylainen: String builder is only if you want to append multiple things. A single localized string can have a key (which is used to look up the text) and any number of string placeholder parameters. In this case like I showed (https://github.com/Revolutionary-Games/Thrive/pull/3118#discussion_r841255735) this can be implemented as a single LocalizedString.

hhyyrylainen: > using localizedStr.TroStr yields the translation key without the arguments, i think there needs to be used a localized Str builder nonetheles.

That can't be right. LocalizedString's ToString method should look up the translation and perform any string placeholder substitutions.

https://github.com/Revolutionary-Games/Thrive/blob/master/src/engine/LocalizedString.cs#L37

simulation_parameters/SimulationParameters.cs
hhyyrylainen: The typo is still here? Why was this marked as resolved?

Kemikal1: huh

hhyyrylainen (approved):
Did all the changes I still wanted. Also fixed loading older saves, and as far as I can tell all is fine now, but I'll wait for a bit to give time for other people to test.

Issue #3043, opened 2022-01-21
Make chemoreceptor upgrade auto select the compound's colour


hhyyrylainen:
could make the chemoreceptor auto select the compound's colour when selecting one, this would make it faster to configure multiple chemoreceptors for different compounds to look for.

Just needs a callback for the option button to detect when it changes value, and then just looking up the compound colour and setting that on the colour picker. Could maybe only change the colour if it is white or one of the other compound's colour. That would make it less annoying for players as if they first selected a custom colour and then selected a different compound, they'd lose that colour without that check.

84634E1A607A:
What about getting compound colors to presets?

hhyyrylainen:
I guess that could be an alternative, though I didn't add a separate preset group for the chemoreceptor. That's because I thought it makes more sense for the player to be able to carry over the presets they set on the membrane tab to the chemoreceptor colour picker.

PR #3131, opened 2022-03-27
Make it so that when you click on a compounds color the color matches the compound

harchrisdytian:

closes[ #3043](https://github.com/Revolutionary-Games/Thrive/issues/3043)

this makes it so that when you select a different component it automatically switch the color of the components

**Related Issues**

**Progress Checklist**

### DIFF
--- a/src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
+++ b/src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
@@ -22,13 +22,15 @@ public class ChemoreceptorUpgradeGUI : VBoxContainer, IOrganelleUpgrader
 
     private List<Compound>? shownChoices;
     private OrganelleTemplate? storedOrganelle;
-
+    //setting list for saved colors
+    private List<Color> shownSavedColour;
     public override void _Ready()
     {
         compounds = GetNode<OptionButton>(CompoundsPath);
         maximumDistance = GetNode<Slider>(MaximumDistancePath);
         minimumAmount = GetNode<Slider>(MinimumAmountPath);
         colour = GetNode<TweakedColourPicker>(ColourPath);
+
 
         compounds.Clear();
 
@@ -43,10 +45,12 @@ public class ChemoreceptorUpgradeGUI : VBoxContainer, IOrganelleUpgrader
     {
         storedOrganelle = organelle;
         shownChoices = SimulationParameters.Instance.GetCloudCompounds();
-
+        
+        
         foreach (var choice in shownChoices)
         {
             compounds.AddItem(choice.Name);
+                
         }
 
         // Select glucose by default
@@ -88,5 +92,12 @@ public class ChemoreceptorUpgradeGUI : VBoxContainer, IOrganelleUpgrader
         // TODO: make an undoable action
         storedOrganelle.SetCustomUpgradeObject(new ChemoreceptorUpgrades(shownChoices[compounds.Selected],
             (float)maximumDistance.Value, (float)minimumAmount.Value, colour.Color));
+    }
+    public void CompoundChanged(int index)
+    {
+        if (shownChoices[index] != null)
+
+            colour.Color = shownChoices[index].Colour; // when the color changes then it changes it to match the color
+        }
     }
 }
--- a/src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
+++ b/src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
@@ -92,22 +92,25 @@ public class ChemoreceptorUpgradeGUI : VBoxContainer, IOrganelleUpgrader
 
     public void CompoundChanged(int index)
     {
-        if (shownChoices?[index] != null)
+        if (shownChoices == null)
         {
-            // If the color is in the shownChoices list don't change the color
-            bool isColorInCompundList = false;
-            foreach (Compound compound in shownChoices)
-            {
-                if (colour.Color == compound.Colour || colour.Color == Colors.White)
-                {
-                    isColorInCompundList = true;
-                }
-            }
+            GD.PrintErr("Shown choices are not set up correctly");
+            return;
+        }
 
-            if (isColorInCompundList)
+        // If the color is in the shownChoices list don't change the color
+        bool isColorInCompundList = false;
+        foreach (Compound compound in shownChoices)
+        {
+            if (colour.Color == compound.Colour || colour.Color == Colors.White)
             {
-                colour.Color = shownChoices[index].Colour;
+                isColorInCompundList = true;
             }
+        }
+
+        if (isColorInCompundList)
+        {
+            colour.Color = shownChoices[index].Colour;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.tscn
```diff
@@ -88,5 +85,7 @@ text = "LINE_COLOUR"
 margin_left = 0.0
 margin_top = 200.0
 margin_right = 1280.0
-margin_bottom = 794.0
+margin_bottom = 759.0
```
adQuid, line 88: Intentional change?

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -89,4 +92,11 @@ public void ApplyChanges(MicrobeEditor editor)
         storedOrganelle.SetCustomUpgradeObject(new ChemoreceptorUpgrades(shownChoices[compounds.Selected],
             (float)maximumDistance.Value, (float)minimumAmount.Value, colour.Color));
     }
+    public void CompoundChanged(int index)
```
adQuid, line 95:
```suggestion

    public void CompoundChanged(int index)
```

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -43,10 +44,12 @@ public void OnStartFor(OrganelleTemplate organelle)
     {
         storedOrganelle = organelle;
         shownChoices = SimulationParameters.Instance.GetCloudCompounds();
-
+        
+        
         foreach (var choice in shownChoices)
         {
             compounds.AddItem(choice.Name);
+
```
adQuid, line 52:
```suggestion
```

hhyyrylainen:
You should use more descriptive names for pull requests (but the actual thing it does in the name instead of just an issue reference).

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -89,4 +89,12 @@ public void ApplyChanges(MicrobeEditor editor)
         storedOrganelle.SetCustomUpgradeObject(new ChemoreceptorUpgrades(shownChoices[compounds.Selected],
             (float)maximumDistance.Value, (float)minimumAmount.Value, colour.Color));
     }
+
+    public void CompoundChanged(int index)
+    {
+        if (shownChoices != null)
+        {
+            colour.Color = shownChoices[index].Colour; // when the color changes then it changes it to match the color
```
hhyyrylainen, line 97: Using trailing comments isn't really the Thrive style so you should put this comment on a line by itself (and start it with a capital letter).

Also there's one part of the issue that's left unimplemented here:
> Could maybe only change the colour if it is white or one of the other compound's colour. That would make it less annoying for players as if they first selected a custom colour and then selected a different compound, they'd lose that colour without that check.

So this should first check if the current colour is *any* of the compound colours and only *then* update the colour, so that if the player has selected a custom colour before changing the compound type, they don't lose that selection.

hhyyrylainen (changes requested):
This looks like a good start, though I didn't playtest this. There's a small style issue here and I'd like to see the second part of the issue implemented here.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -89,4 +89,25 @@ public void ApplyChanges(MicrobeEditor editor)
         storedOrganelle.SetCustomUpgradeObject(new ChemoreceptorUpgrades(shownChoices[compounds.Selected],
             (float)maximumDistance.Value, (float)minimumAmount.Value, colour.Color));
     }
+
+    public void CompoundChanged(int index)
+    {
+        if (shownChoices?[index] != null && shownChoices != null)
+        {
+            // If the color is in the shownChoices list don't change the color
```
hhyyrylainen, line 97: Shouldn't this comment be the other way around?

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -89,4 +89,25 @@ public void ApplyChanges(MicrobeEditor editor)
...
+            bool isColorInCompundList = false;
+            foreach (Compound compound in shownChoices)
+            {
+                if (colour.Color == compound.Colour || colour.Color == Colors.White)
```
hhyyrylainen, line 101: Instead of having Colors.White here, I think it would be better to change the other place in this file where Colors.White is used to use the glucose colour, that way we shouldn't need the explicit check with white here.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -89,4 +89,25 @@ public void ApplyChanges(MicrobeEditor editor)
...
             (float)maximumDistance.Value, (float)minimumAmount.Value, colour.Color));
     }
+
+    public void CompoundChanged(int index)
+    {
+        if (shownChoices?[index] != null && shownChoices != null)
+        {
+            // If the color is in the shownChoices list don't change the color
+            bool isColorInCompundList = false;
+            foreach (Compound compound in shownChoices)
```
hhyyrylainen, line 99: I think this could be pretty easily be converted to a single LINQ statement which are preferred over these short explicit loops, something like (I wrote this on github so it might not compilte): `bool isCompoundsColourSelected = shownChoices.Any(c => c.Colour == colour.Color)`

hhyyrylainen (changes requested):
Approach looks fine, a few comments about cleaning up the code a bit.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -89,4 +90,17 @@ public void ApplyChanges(MicrobeEditor editor)
         storedOrganelle.SetCustomUpgradeObject(new ChemoreceptorUpgrades(shownChoices[compounds.Selected],
             (float)maximumDistance.Value, (float)minimumAmount.Value, colour.Color));
     }
+
+    public void CompoundChanged(int index)
+    {
+        if (shownChoices != null)
```
hhyyrylainen: I think something like `if(shownChoices?.Any() == true))` is what the linter wants.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -89,4 +90,18 @@ public void ApplyChanges(MicrobeEditor editor)
         storedOrganelle.SetCustomUpgradeObject(new ChemoreceptorUpgrades(shownChoices[compounds.Selected],
             (float)maximumDistance.Value, (float)minimumAmount.Value, colour.Color));
     }
+
+    public void CompoundChanged(int index)
+    {
+        if (shownChoices?[index] != null && shownChoices?.Any() == true)
```
hhyyrylainen, line 96: This is not correct. Do you want me to code this right to show how this should be done (I understand it can be a bit cryptic sometimes what jetbrains wants if you aren't using their IDE)?

hhyyrylainen (approved):
I fixed up the code and tested in game and at least basic functionality with this doesn't have any issues.

PR #3148, opened 2022-04-05
Environment in fitness score

### LINKED ISSUE #3132: Auto-evo should consider actual efficiency of processes and not reward impeded processes for fitness
Apparently, auto-evo result seems not to be affected by the actual concentration of compounds. 

I guess the issue is that, in the code below (from `FoodSource.cs`), `RunnableProcesses` do not take into account the environment in which the process runs.

```
foreach (var process in organelle.Definition.RunnableProcesses)
                {
                    if (process.Process.Inputs.ContainsKey(compound))
                    {
                        if (process.Process.Outputs.ContainsKey(glucose))
                        {
                            energyCreationScore += process.Process.Outputs[glucose]
                                / process.Process.Inputs[compound] * Constants.AUTO_EVO_GLUCOSE_USE_SCORE_MULTIPLIER;
                        }

                        if (process.Process.Outputs.ContainsKey(atp))
                        {
                            energyCreationScore += process.Process.Outputs[atp]
                                / process.Process.Inputs[compound] * Constants.AUTO_EVO_ATP_USE_SCORE_MULTIPLIER;
                        }
                    }
                }
```

Maxonovien:

This PR mitigates energy score generation by integrating process efficiency, derived from compound concentrations in the patch.

**Related Issues**

Closes #3132 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/simulation/food_source/ChunkFoodSource.cs
+++ b/src/auto-evo/simulation/food_source/ChunkFoodSource.cs
@@ -57,7 +57,7 @@
 
             // We ponder the score for each compound by its amount, leading to pondering in proportion of total
             // quantity, with a constant factor that will be eliminated when making ratios of scores for this niche.
-            var score = energyCompounds.Sum(c => EnergyGenerationScore(microbeSpecies, c.Key) * c.Value);
+            var score = energyCompounds.Sum(c => EnergyGenerationScore(microbeSpecies, c.Key, patch) * c.Value);
 
             score *= chunkEaterSpeed * species.Behaviour.Activity;
 
--- a/src/auto-evo/simulation/food_source/CompoundFoodSource.cs
+++ b/src/auto-evo/simulation/food_source/CompoundFoodSource.cs
@@ -26,7 +26,7 @@
         {
             var microbeSpecies = (MicrobeSpecies)species;
 
-            var compoundUseScore = EnergyGenerationScore(microbeSpecies, compound);
+            var compoundUseScore = EnergyGenerationScore(microbeSpecies, compound, patch);
 
             var energyCost = simulationCache
                 .GetEnergyBalanceForSpecies(microbeSpecies, patch)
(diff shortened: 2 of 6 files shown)


buckly90:
I started the game (Not freebuild) and immediately replaced my lone cytoplasm to get an increased population score of 1307
I did the same in the 0.5.7 build and got 1653. So that's good. Especially compared to the other autotrophic parts which increase the score much more substantially.

It seems that autoevo still favors thylakoids over cytoplasm though, so it's possible that some thylakoid-bearing prokaryotes might still come about in the dark environments.

Whatever the case, so far it seems that this is a good change to autoevo. I'll continue playing with it a bit before submitting my review proper.

hhyyrylainen:
What about thylakoids? Do they show up in the auto-evo explanation as energy sources with sunlight in 0.5.7, and is that then gone when testing this?

hhyyrylainen (approved):
I think the code looks fine, the changes seem minimal (other than making that one private method public causing all of its lines to be marked changes). Unless this impacts auto-evo running time, I don't see any problem with this, though I didn't test in-game to see if the gameplay change is actually good or not)

buckly90:
> What about thylakoids? Do they show up in the auto-evo explanation as energy sources with sunlight in 0.5.7, and is that then gone when testing this?

It doesn't seem to take in sunlight in either version. That being said, I've done some more testing now and it seems that thylakoids are indeed far less attractive in dark patches

The only exception is a single hex thylakoid organism, but there are still more attractive options for autoevo regardless.

This is good so I'm approving it!

buckly90 (approved):

Maxonovien:
Just for the record, this
> It seems that autoevo still favors thylakoids over cytoplasm though, so it's possible that some thylakoid-bearing prokaryotes might still come about in the dark environments.

It is probably due to cytoplasm having a higher base (=glycolysis) output of ATP. That's precisely what happens for metabolosomes. Since the total amount of energy for each source (here glucose) is not affected by microbe species and their actual processes, this results in lower `population` = `total amount/individual output`. 

Gave me some idea, I'll try to drop a PR when I can.

PR #3149, opened 2022-04-06
Balance Pass for 0.5.8

buckly90:
* Decreased mass of photosynthesis so that it matches the mass of three thylakoids
* Increased production of ammonia from nitrogen-fixation
* Decreased MP cost of nucleus to 70 to allow for small changes after it is placed
* Decreased MP cost of toxin vacuole to 55
* Decreased mass of signaling agent and chemoreceptor

**Progress Checklist**

      break existing features:
      members reviewing this PR)


adQuid:
I playtested it a little bit just in case, but I don't see any way to break these changes to test out.

badideaguy:
This may have the side effect of among auto evo consider a nucleus more 
Since auto evo would also be able to make those small changes, helping the nucleus not being absolutely useless and wasteful to auto evo

hhyyrylainen (approved):
I guess I'll merge this tomorrow in time for the RC, unless there's something significant that still needs changing.

buckly90:
I just need some feedback if possible

hhyyrylainen:
My feedback is that I don't much see the point of the nucleus being reduced in price. It still takes up basically all of your MP so with the leftover you can't really fix your cell so anyway you need to prepare for the nucleus.

PR #3175, opened 2022-04-09
update_localization.rb Remove trailing space

### LINKED ISSUE #2467: Make check_formatting.rb remove trailing space in `# ` in locale files
> The .po starting comments include one line that is `# ` though it seems that sometimes the 
> space after `#` is automatically removed. I'm not sure what causes that, so I've just let that be.
> _Originally posted by @hhyyrylainen in https://github.com/Revolutionary-Games/Thrive/issues/2459#issuecomment-894018838_

I can't! Every time I use update_localization.rb I have to manually delete all those trailing spaces! Pls make it automatic.

(Along with starting with BOM, I really don't know how I can manually add BOM to files)

84634E1A607A:
**Related Issues**
fixes #2467

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/scripts/update_localization.rb
+++ b/scripts/update_localization.rb
@@ -1,6 +1,7 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+
 # List of locales, edit this to add new ones:
 LOCALES = %w[ar af bg ca cs da de el en eo es_AR es et fi fr frm he hu id ka ko
              la lb_LU it nl nl_BE pl pt_BR pt_PT ro ru si_LK sr_Cyrl sr_Latn sv
@@ -12,6 +13,7 @@ LOCALES = %w[ar af bg ca cs da de el en eo es_AR es et fi fr frm he hu id ka ko
 LINE_WRAP_SETTINGS = ['--no-wrap'].freeze
 
 require 'optparse'
+require "os"
 require_relative '../bootstrap_rubysetupsystem'
 require_relative '../RubySetupSystem/RubyCommon'
 
@@ -62,6 +64,7 @@ Dir.chdir(LOCALE_FOLDER) do
 
     if File.exist? target
       puts "Extracting #{locale}.po"
+
       runOpen3Checked 'msgmerge', '--update', '--backup=none', *LINE_WRAP_SETTINGS,
                       target,
                       "messages#{@options[:pot_suffix]}"
@@ -73,6 +76,15 @@ Dir.chdir(LOCALE_FOLDER) do
                       "messages#{@options[:pot_suffix]}"
     end
   end
+
+  # Remove trailing space
+  OS.windows?
+    info "Windows detected. Removing trailing space..."
+
+    LOCALES.each do |locale|
+      target = locale + @options[:po_suffix]
+      File.write(target, File.read(target).gsub(/# \n/,"#\n"))
+    end
 end
 
 success 'Done extracting .po files'
--- a/scripts/update_localization.rb
+++ b/scripts/update_localization.rb
@@ -44,7 +44,7 @@ puts "Detected Thrive root folder: #{ROOT_FOLDER}"
 
 Dir.chdir(LOCALE_FOLDER) do
   puts 'Extracting .pot file'
-=begin
+
   runOpen3Checked 'pybabel', 'extract', '-F', File.join(LOCALE_FOLDER, 'babelrc'), '-k',
                   'LineEdit', '-k', 'text', '-k', 'DisplayName', '-k', 'Description', '-k',
                   'ProcessesDescription', '-k', 'window_title', '-k', 'dialog_text', '-k',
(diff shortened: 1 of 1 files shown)


scripts/update_localization.rb
```diff
@@ -73,6 +76,15 @@
                       "messages#{@options[:pot_suffix]}"
     end
   end
+
+  # Remove trailing space
+  OS.windows?
+    info "Windows detected. Removing trailing space..."
```
hhyyrylainen, line 82: Trailing spaces should be stripped even on non-windows. Ruby string strip should probably remove the trailing spaces. Did we already have check disallowing trailing spaces for CI?

scripts/update_localization.rb
```diff
@@ -73,6 +76,15 @@
...
+
+    LOCALES.each do |locale|
+      target = locale + @options[:po_suffix]
+      File.write(target, File.read(target).gsub(/# \n/,"#\n"))
```
hhyyrylainen, line 86: I think you should switch the code to use a line by line approach to find the trailing whitespace, and *only* write the new contents to disk if there is a change.

hhyyrylainen, line 86: Also the file read problem in CI is because you need to specify that it should be read in utf8 encoding.

scripts/update_localization.rb
84634E1A607A: We have, so every time I run update_localization.rb CI fails.

The cause for me to add an OS.windows? is that **every** time I use it on Windows it adds trailing space, but **none** have I experienced the strange behaviour on Linux.

scripts/update_localization.rb
84634E1A607A: As above.

scripts/update_localization.rb
hhyyrylainen: If someone manually edits stuff on Linux, they might make a mistake so for that reason I think it would be nice to have the trailing spaces strip on any platform.

scripts/update_localization.rb
hhyyrylainen: Other people won't be happy if their disk is uselessly worn down by rewriting 30 files that don't change.

scripts/update_localization.rb
84634E1A607A: Then we should add this in check_formatting.rb (?)

hhyyrylainen: As you said there's already a check there against this problem, and if the problem only occurs when changing the locales, then I think it's fine to have here. I won't accept any kind of extra mess in check_formatting.rb, it's kind of incidental that some of the checks there actually fix what they detect. If I had a choice in that check_formatting.rb would *not* do any modifications, but due to the tools it needs to call, that is not possible.

84634E1A607A:
No no this checks for ` \n`
(btw should I check for `( *)\n`?

hhyyrylainen:
There are more types of whitespace than just the space character...

84634E1A607A:
Emm, this deals with the auto-generated .po files. (Pybabel won't add a tab at the end of a line, will it?)

hhyyrylainen:
It might, and anyway as I said, if someone manually edits a file, I think gettext might preserve those, so better to just be safe and handle any whitespace. I'm very careful about always handling any type of whitespace and not just spaces because it will cause problems at some point, so just easier to always assume whitespace can be anything.

84634E1A607A:
Then what are they? Apart from " ", "\t"?

hhyyrylainen:
Unicode has various whitespace, and even before unicode there are things like *vertical* tabs. There's just a bunch of whitespace things, that's why you need to use the strip method or regex whitespace class selector, you can't list them all.

scripts/update_localization.rb
```diff
@@ -42,7 +44,7 @@
 
 Dir.chdir(LOCALE_FOLDER) do
   puts 'Extracting .pot file'
-
+=begin
```
hhyyrylainen, line 47: What's this?

scripts/update_localization.rb
```diff
@@ -73,6 +76,24 @@
...
   end
+=end
+  # Remove trailing space
+  info "Removing trailing space..."
+
+  LOCALES.each do |locale|
+    target = locale + @options[:po_suffix]
+    content = File.read(target, :encoding => "utf-8")
+    trimmed = ""
+    content.split("\n").each do |line|
```
hhyyrylainen, line 87: I think there is an inbuilt method to split things into lines. There's actually a way to read files line by line which is what I've preferred to use. So you could have code like:
```
for each line in file{|line|
stripped = line.strip
if stripped != line
  changes = true
end
lines.append stripped
}

if changes
  write each line to the file
end

```

scripts/update_localization.rb
84634E1A607A: ... Debug leftovers

scripts/update_localization.rb
```diff
@@ -12,6 +13,7 @@
 LINE_WRAP_SETTINGS = ['--no-wrap'].freeze
 
 require 'optparse'
+require "os"
```
hhyyrylainen, line 16: Unrequired now

scripts/update_localization.rb
```diff
@@ -73,6 +76,26 @@
                       "messages#{@options[:pot_suffix]}"
     end
   end
+
+  # Remove trailing space
+  info "Removing trailing space..."
```
hhyyrylainen, line 81:
```suggestion
  info "Removing trailing whitespace..."
```

scripts/update_localization.rb
```diff
@@ -73,6 +76,26 @@
...
+      trimmed = line.rstrip() + "\n"
+      if line != trimmed then changed = true; end
+      trimmedtext += trimmed
+    end
+
+    # Babel generates identically formatted files. So when the first one doesn't have trailing space, nor will the others.
+    if !changed then break; end
+
+    File.write(target, trimmedtext, :encoding => "utf-8")
+    puts target
```
hhyyrylainen, line 97: This should say something like "removed trailing whitespace in #{target}"

scripts/update_localization.rb
```diff
@@ -73,6 +76,26 @@
...
+    trimmedtext = ""
+    changed = false
+    File.readlines(target, :encoding => "utf-8").each do |line|
+      trimmed = line.rstrip() + "\n"
+      if line != trimmed then changed = true; end
+      trimmedtext += trimmed
+    end
+
+    # Babel generates identically formatted files. So when the first one doesn't have trailing space, nor will the others.
+    if !changed then break; end
```
hhyyrylainen, line 94: Pretty sure this isn't what rubocop wants so could you run rubocop auto fix on this file? So I don't have to.

hhyyrylainen (approved):
Fixed the formatting of the file myself.

PR #3176, opened 2022-04-10
Adds the ability to click off the Cell Species Name Editor 

GameDungeon:

In the microbe editor, in the past, if you started editing the name line editor it would gain keyboard focus and would not unfocus until you hit enter. This became a problem if you forgot to hit enter and you later rotated something while editing the cell. In this pull if you click on something other than the name editor it loses focus.

**Related Issues**

As far as I can tell, there is no currently opened issue for this.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

PS: This is my first pull in this project, so if I've messed anything up I'm sorry, and please let me know :)


src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -696,6 +699,17 @@ public override void _Process(float delta)
         }
     }
 
+    public override void _Input(InputEvent @event)
+    {
+        if (@event is InputEventMouseButton buttonEvent)
+        {
+            if (buttonEvent.Pressed)
+            {
+                EmitSignal("Clicked");
```
Kasterisk, line 708: Use `nameof(Clicked)`.

Kasterisk (dismissed):
Functionality seems good, just a small nitpick. Also I think this PR is small enough to get into 0.5.8.

src/microbe_stage/editor/EditorComponentBottomLeftButtons.cs
```diff
@@ -241,8 +252,6 @@ private void OnNameTextEntered(string newText)
                 speciesNameEdit.GetNode<AnimationPlayer>("AnimationPlayer").Play("invalidSpeciesNameFlash");
             }
         }
-
-        EmitSignal(nameof(OnNameSet), newText);
```
hhyyrylainen, line 245: I feel like this is a bit of a problem because for example the multicell editor has UseSpeciesNameValidation set to false and it wants to do its own validation so not having any signals triggered from this method means that it can't be made to react to this.

src/microbe_stage/editor/EditorComponentBottomLeftButtons.cs
```diff
@@ -225,11 +225,22 @@ private void OnNameTextChanged(string newText)
...
+        ValidationPopup(newText);
+        EmitSignal(nameof(OnNameSet), newText);
+    }
+
+    private void OnClickedOffName()
+    {
+        ValidationPopup(speciesNameEdit.Text);
+    }
+
+    private void ValidationPopup(string text)
```
hhyyrylainen, line 238: `PerformValidation`

src/microbe_stage/editor/EditorComponentBottomLeftButtons.tscn
```diff
@@ -170,6 +170,7 @@ size_flags_vertical = 4
 custom_colors/font_color = Color( 1, 1, 1, 1 )
 max_length = 100
 placeholder_text = "SPECIES_NAME_DOT_DOT_DOT"
+caret_blink = true
```
hhyyrylainen, line 173: Is this a changed option or is this like the default value? Godot editor usually when you reopen it, it will remove default values.

Kasterisk: I think this is a changed option, personally I like if the caret blinks so this is a welcome change from me.

hhyyrylainen: I suppose a blinking caret is the more common way it usually works.

GameDungeon: This is a value I changed. I think it looks far better. If people disagree that's fine, I can change it.

src/microbe_stage/editor/EditorComponentBottomLeftButtons.cs
GameDungeon: The validation method itself does not have it, but the ones that call it (other than the one that does not change the text) have it.

src/microbe_stage/editor/CellEditorComponent.cs
GameDungeon:
```suggestion
                EmitSignal(nameof(Clicked));
```

src/microbe_stage/editor/EditorComponentBottomLeftButtons.cs
GameDungeon: Ok, I'll go change this.

GameDungeon: fixed

src/microbe_stage/editor/EditorComponentBottomLeftButtons.cs
GameDungeon: A problem I do see, is it won't release focus with UseSpeciesNameValidation false. I guess I could fix that bug too.

src/microbe_stage/editor/EditorComponentBottomLeftButtons.cs
```diff
@@ -226,16 +226,16 @@ private void OnNameTextChanged(string newText)
 
     private void OnNameTextEntered(string newText)
     {
-        ValidationPopup(newText);
+        PreformValidation(newText);
```
hhyyrylainen, line 229: Typo...

GameDungeon: oops. Fixed

RacerBG:
There is a problem here. The main idea works no doubts about it but there are a few quirks. I will edit this comment further if I find more things:

1. Try to delete "Thrivium" to make the name red and the focus will remain then try to generate a random name (which will be valid) but the font will remain red. It will turn white only after a second press to the name generator
2. Pressing "Esc" should unfocus it IMO.


PS: Not related to this PR but even with an invalid name you can still continue to play the game and when the name is invalid you shouldn't be able to do anything else (like placing organelles) but to fix it first. If this is not an existing problem a new issue could be opened.

hhyyrylainen:
At least the unfocus callback seems to be triggered in the right place? I was worried it wouldn't trigger when clicking on an organelle selection button or some other interactive GUI element as I thought that might stop the event bubbling up to the base of the GUI hierarchy where the on click handler was added.

hhyyrylainen:
I think that means this PR is basically ready, unless that first issue you found is new in this PR and doesn't happen on master branch. Could you confirm if that issue is new in this branch? If it is new then I'd like to get that fixed before merge, otherwise this is pretty close to being done.

hhyyrylainen:
Well the prototype doesn't have name validation like currently so I doubt this is broken there. 
But yeah as it seems this introduces a new bug, I'd like to see that fixed in this PR before merging this.

GameDungeon:
Do you wish me to fix 2 and the ps here? I'll fix the concern when I get home again, but I should be able to fix the other things as well.

hhyyrylainen:
The second issue is not required, but I don't think it would bloat this PR too much, though the priority between canceling active actions, menus, etc. with the ESC input needs to be handled correctly so that the right place only receives the ESC press so that might be a bit of a mess if you aren't familiar with our input system and the priority levels in it yet.

RacerBG:
> Do you wish me to fix 2 and the ps here as well? I'll fix the concern when I get home again, but I should be able to fix the other things as well.

As you wish. :smiley: As I said, only 1 and 2 are relevant for this PR but you are more than welcome to fix PS as well (probably in another PR though).

GameDungeon:
Ok I'll look at it and if I can't figure it out I'll leave it.

GameDungeon:
I've mostly fixed the pull, but I'm having a strange issue. I posted it in the discord as it will be easiest to solve there.

GameDungeon:
So I fixed the bug and also the PS as it turned out to be a trivial fix. Now that you can click off the name editor, IMO the ESC key should be reserved for the pause menu.

GameDungeon:
As far as I can tell that failed check is not my fault right?

hhyyrylainen:
> IMO the ESC key should be reserved for the pause menu. It would also be annoying to be unable to pause the game, due to an incorrect name.

It already serves multiple purposes as it allows you to cancel the current action in the editor. This is handled by the input system priorities which allows the higher priority key press callbacks to check first if they should do something and only consume the key press if they did something. But I agree that it'd be better to make a separate PR and not try to add it in here.

@RacerBG could you retest this please?

src/microbe_stage/editor/EditorComponentBottomLeftButtons.cs
```diff
@@ -279,4 +297,14 @@ private void UpdateRandomButtonVisibility()
             randomizeNameButton.Visible = showRandomizeButton;
         }
     }
+
+    private void OnControlMouseEntered()
```
hhyyrylainen, line 301: I kinda don't like this solution. Isn't there some other way to make sure the validation is for example performed *again* when clicking the random button or something to fix the problem?

hhyyrylainen:
> As far as I can tell that failed check is not my fault right?

Yeah, it isn't. Recently there's been some problems with the CI executors not managing to keep a connection open.

src/microbe_stage/editor/EditorComponentBottomLeftButtons.cs
GameDungeon: I don't like the solution, but as far as I can tell, there is no other good option. I know "validation is for example performed again when clicking the random button" has a couple of issues that would stop it from working.

hhyyrylainen: Care to elaborate? I think it anyway would make sense that the random button would retrigger the validation.
Oh and now that I'm thinking, shouldn't the Clicked callback ensure it doesn't trigger the validation unnecessarily a bunch of times? It should only happen *once* after the text has been changed in the name box.

hhyyrylainen (approved):
I had a go myself but couldn't really figure out a good way around the hovered callbacks. So I just did some cleanup and a slight logic change on this.

Issue #2055, opened 2021-01-24
Rework spawn system to be encounter based


hhyyrylainen:
basically instead of random spawning everything, focus should be shifted to spawning interesting encounters.
https://forum.revolutionarygamesstudio.com/t/the-microbe-spawn-system/767

Though maybe keeping the environmental toxins also randomly spawning to give the world a bit more stuff in it would be fine.

hhyyrylainen:
Related discussion: https://forum.revolutionarygamesstudio.com/t/overhauling-spawn-system-architecture/976

hhyyrylainen:
This seems unlikely to ever be done at this point...

PR #3235, opened 2022-04-18
Rework Spawn System

adQuid:

Reworks the spawn system to only spawn once per area around the player, excluding microbe spawn which is continuous. Each "sector" is about halfway to the edge of the screen when zoomed out, and new entities appear 2 or more sectors away from the player. 

Because there are 12 sectors possibly spawning microbes now, the Microbe Spawner has been modified to spawn them less often. There is also an AI change to make sure newly spawned AI is headed towards the player instead of just swimming around offscreen waiting to despawn. Because of both of these, I also took out the fancier colony spawn patterns to save on some computation time.

I also increased the cloud spawn radius to prevent clouds despawning too soon, and tied the chemoreceptor max range to that value, so that detection lines wont mysteriously go away because the cloud despawned (and to prevent excess computation)

Finally, I added just a little phosphates that always spawns at the start of the game, so that the early game isn't completely empty and to give a new player something obvious and visible to interact with.

There is still an issue with the tiny purple circle that appears then fades away sometimes, but that issue is already in master branch, and I don't know where to start fixing that.

**Related Issues**

I don't see any from a cursory scan of the issues list, but let me know if I'm wrong.

Note: This probably renders https://github.com/Revolutionary-Games/Thrive/pull/2913 obsolete

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/SpawnSystem.cs
```diff
@@ -225,143 +207,103 @@ private int HandleQueuedSpawns(int spawnsLeftThisFrame)
...
+                        Mathf.Sin(angle) * Constants.SPAWN_SECTOR_SIZE * 2));
             }
         }
-
-        estimateEntityCountInSpawnRadius += spawned;
     }
 
     /// <summary>
-    ///   Does a single spawn with a spawner
+    ///   Does a single spawn with a spawner, or enqueues a spawn if over the frame limit
```
adQuid, line 284:
```suggestion
    ///   Does a single spawn with a spawner
```

buckly90:
Incredible change! It makes the world feel far from barren and actually alive.

Unfortunately, there is a glaring performance issue after a few generations. Cells would just pile up and pile up until I got to about 20 fps, and sometimes even lower. And that's pretty scary with my setup.

Strangely enough, even dying, which usually despawns everything, did not fix the performance issue. Does everything despawn as intended? Is there some kind of memory leak?

src/microbe_stage/MicrobeAI.cs
```diff
@@ -223,6 +223,17 @@ private void ChooseActions(Random random, MicrobeAICommonData data, Microbe? sig
                 break;
         }
 
+        // If I'm very far from the player, get on stage
+        var player = data.AllMicrobes.Where(otherMicrobe => otherMicrobe.IsPlayerMicrobe).FirstOrDefault();
```
hhyyrylainen, line 227: Do we want to lump this in with the spawn system change which is already a huge difference?

simulation_parameters/Constants.cs
```diff
@@ -19,6 +19,9 @@ public static class Constants
     public const float MIN_SPAWN_DISTANCE = -5000.0f;
     public const float MAX_SPAWN_DISTANCE = 5000.0f;
 
+    // Size of "chunks" used for spawning entities
```
hhyyrylainen, line 22: XML needed

hhyyrylainen:
Unless the code is hidden pretty well, this seems to now ignore the maximum spawn limit of stuff, which would mean that I'd also expect the performance to tank a lot because the entity limit is really the only thing keeping the game from falling to a couple of FPS...

src/microbe_stage/MicrobeAI.cs
adQuid: I'm pretty confident this is now essential. With microbes now spawning in a ring far from the player, without this change you would need to spawn almost 5 times as many to get the same number to appear on screen.

Oliveriver:
Had a go with this since I was curious. It does indeed make the world much more dynamic and much more consistent*.

However, that had the knock-on effect of making it much easier. I had no issue at all collecting enough compounds almost immediately every generation. The spawn system is consistent, but it's consistently high. If there is a simple way to turn down this dial, that would be great, especially since that would provide one option for difficulty variation.

I also ran into performance issues pretty quickly. Hopefully reimplementing the entity limit fixes that.

* The exception was when I entered a new patch for the first time. It was often completely barren for long enough that I ran out of compounds and died. After that though, all was well.

src/microbe_stage/MicrobeAI.cs
hhyyrylainen: Wasn't the idea with sector spawns that sectors contain specific amount of stuff, which would even load in and out as the player moves? Did this PR fundamentally change from the first spawn system PR?

adQuid: I'm not intimately familiar with the other PR, but based on what you're describing this is a much simpler solution. All this change does is take the "Biased circle" spawn pattern and instead track what areas have and have not spawned something in them. It then relies on existing despawn functionality to remove things too far from the player.

This works fine for chunks and clouds which don't move, but most microbes won't just sit around and wait to be discovered; they will start moving, usually out of the perimeter for depsawning and never enter the game properly. My suspicion is that this is already an issue in master, but now that the Microbes are spawning a little further out (so they don't appear onscreen) it's a more noticeable problem.

adQuid:
> If there is a simple way to turn down this dial, that would be great, especially since that would provide one option for difficulty variation.

This variable (which I just moved into the constants file) would be the easiest way to do that, and not a bad way to practice messing with the code :) https://github.com/Revolutionary-Games/Thrive/pull/3235/files#diff-1866e717ab370a48530eaa34ab83ade4adf2d08d2a63befb6adb2d8c5e951211R27

src/microbe_stage/MicrobeAI.cs
hhyyrylainen: Ah okay, you should rename this PR to not contain the "sector" word in the name then.

hhyyrylainen:
@adQuid you should still use the `Constants.DEFAULT_MAX_SPAWNED_ENTITIES` constant for the entity limit, or rename / remove that previous one if it isn't used anymore.

hhyyrylainen:
I think a pretty huge finding has been made regarding the spawn system, so I think this PR should wait for that to be complete and then we should re-evaluate this: https://github.com/Revolutionary-Games/Thrive/pull/3243

adQuid:
Understood. That works out great for me since I won't have much time to work on this until the weekend anyway.

adQuid:
There is still an unfortunate tendency for islands of compound clouds to form, I think due to limitations in the current compound cloud system, but the lag spikes seem to be fixed and I like the microbe spawn rates a lot better now.

Kemikal1:
Did some testing and to me the compund islands werent so obvious. It does spawn more microbes indeed maybe around 70-100 entitites from 20-30 normally so more lively definetly.

buckly90:
Performance is much better now. My fps never dropped below the 30-40 range which still might not be good given I have a pretty good PC, but it's at least playable now.

For some reason it feels like compounds spawns are far less consistent since I last tested. There would be periods of time where I would just swim and swim and encounter maybe one or two clouds. This issue seemingly became more frequent as time went on. I even had to double check to make sure I was on the right branch because it felt a bit too similar to what we already have. Microbe spawns are at least rather frequent so I was able to survive off of predation.

RacerBG:
> Finally, I added just a little phosphates that always spawns at the start of the game, so that the early game isn't completely empty and to give a new player something obvious and visible to interact with.

Is it a good idea though? If the tutorial is enabled, it should guide the player towards the nearest cloud of glucose.

Otherwise, I agree with the changes you made but broader testing is needed.

adQuid:
> Is it a good idea though? If the tutorial is enabled, it should guide the player towards the nearest cloud of glucose.

I'm really confident this will be important. A side effect of the new spawn system not spawning stuff in view of the player is that the screen starts out blank save for one microbe. If the player follows the tutorial and just swims in circles or chooses to skip it (I assume most do, people like to skip those) there are very few early clues about what the gameplay is about, and a player could reasonably wonder if the cell is moving at all or the background is just drifting. Look up any talk on video game tutorials, and you'll hear the same thing: don't throw a wall of text at the player, give them something to experiment with. The start of the game is a great time to have one cell, one cloud directly to the right, and a quick demonstration of what the first 10 minutes of the game are like.

And if I'm wrong, it's one purple puff that fills up half the phosphate bar.

adQuid:
@buckly90 Having confirmed that the issue I was talking about on Discord was a fluke, I now think that this is in about the best state I can get it into without a massive compound cloud refactor, which I just don't have the time to commit to right now.

simulation_parameters/Constants.cs
```diff
@@ -381,7 +388,7 @@ public static class Constants
     public const float CELL_REQUIRED_DRAG_BEFORE_APPLY = 0.0033f;
 
     public const float CHEMORECEPTOR_RANGE_MIN = 2;
-    public const float CHEMORECEPTOR_RANGE_MAX = 700;
+    public const float CHEMORECEPTOR_RANGE_MAX = CLOUD_SPAWN_RADIUS;
```
hhyyrylainen, line 391: If we assume that cloud spawn isn't equal to the compound cloud plane distance, restricting this distance doesn't seem like the best idea to me...

src/microbe_stage/MicrobeAI.cs
```diff
@@ -73,6 +73,9 @@ public class MicrobeAI
     [JsonIgnore]
     private MicrobeSignalCommand receivedCommand = MicrobeSignalCommand.None;
 
+    [JsonIgnore]
+    private bool hasBeenNearPlayer;
```
hhyyrylainen, line 77: Shouldn't this be saved? If an old save doesn't have it, this will be false. But for newer saves this seems like a good thing to save.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -849,6 +857,7 @@ private void HandlePlayerRespawn()
         else
         {
             // Player is not extinct, so can respawn
+            spawner.DespawnAll();
```
hhyyrylainen, line 860: This should not be required with the despawn fixes... if this is required please add a comment detailing why this is needed. I think this'll make things visually pop out of existance on camera so I think things will look better if we don't do this and instead rely on the more powerful despawn system to be able to get things done.

Leinourdian:
I did some testing out of curiosity. I like the feel of the new system a lot. It does indeed feel livelier. I also like that things spawn in already before you swim into a new area so I don't have to swim in big circles to get more spawns. My only problem at the moment is the performance which is worse than in 5.8.1. I think it's caused by two things: Firstly, there are more entities in general. Secondly, entities can spawn in quick bursts.

When it comes to the first one, I could sometimes have up to 140 entities (the entity limit), but I'd only see 5 to 30 entities on the screen with max zoom out. That means that about 80-95% of entities were off-screen where the player doesn't really care about them. I'm wondering if it'd be possible to do something to reduce the amount entities off-screen? Or perhaps off-screen entities could be made less performance heavy because they probably don't have to be fully simulated. That's probably easier said than done but it could help performance a lot.

As for the second one, I think that the spawn system can at worst spawn up to dozens of entities in a single frame. You can sometimes see the entity count jump up really fast. I saw some smallish lag spikes that were probably caused by it. I think this is because `spawnsLeftThisFrame` is not taken into consideration inside `SpawnEntities` or the methods used by it like `SpawnWithSpawner` unlike how it used to be.

I noticed the despawning of all entities on player death by the way. I couldn't usually see it but a low framerate sometimes made it visible for a fraction of a second. It didn't look that weird but it's something to take into consideration.

`Denstity` should probably be `Density`.
https://github.com/Revolutionary-Games/Thrive/blob/127c12282e8881290f7ee82e90209f56da88ce7d/src/microbe_stage/Spawner.cs#L30

hhyyrylainen:
> As for the second one, I think that the spawn system can at worst spawn up to dozens of entities in a single frame. You can sometimes see the entity count jump up really fast. I saw some smallish lag spikes that were probably caused by it. I think this is because spawnsLeftThisFrame is not taken into consideration inside SpawnEntities or the methods used by it like SpawnWithSpawner unlike how it used to be.

Sounds like a new bug with this PR. The current spawn system definitely should respect the spawns per frame.

Leinourdian:
> Sounds like a new bug with this PR. The current spawn system definitely should respect the spawns per frame.

Yeah, it's new to this PR.

simulation_parameters/Constants.cs
adQuid: This really was an attempt to fix a related but unnecessary issue anyway.

Oliveriver:
I think my feedback remains about the same as it was: a much more consistent and livelier gameplay experience, but still too easy and still hitting some major performance issues.

I tried passing the same save between master and this branch a few times. In master, I never saw more than 40 entities, and almost never dropped frames. In this branch, I consistently had 80+ entities, sometimes hitting 130, and had both lag spikes and prolonged periods of 30-40 FPS.

Kemikal1:
i think the fact that 130 entities create such a lag is a problem in itself.

hhyyrylainen:
If someone can fix the performance problems with many entities, I would be extremely pleased. In the meantime we must limit the number of entities we allow in the game.

src/microbe_stage/MicrobeStage.cs
adQuid: I've moved this in order to keep the entities from disappearing while the player might see, but the despawn definitely is needed with this spawning system, since without it the game will hold twice as many entities as the max for a couple seconds.

buckly90:
No noticeable drops in framerate, and the world looks livelier than ever!
I never experienced any sudden fields of nothing, there was always a resource nearby, useful or not.

Unfortunately, it seems the density of ammonia and phosphate is a bit too low. It spawns often and consistently enough, but in amounts that contribute little to progress. The result is a massively stretched amount of time between generations. This wouldn't actually be a problem if there was more engagement in the gameplay phase, but for now my suggestion is to tune up the density. You can leave that up to me if you like.

You've done a wonderful job here, this PR functions beautifully from a gameplay standpoint.

Oliveriver:
Misclicked while looking at this on mobile and accidentally merged master into this PR. Sorry about that...

adQuid:
meh, I was probably about to do that anyway

simulation_parameters/Constants.cs
```diff
@@ -19,6 +19,13 @@ public static class Constants
     public const float MIN_SPAWN_DISTANCE = -5000.0f;
     public const float MAX_SPAWN_DISTANCE = 5000.0f;
 
+    /// <summary>
+    ///     Size of "chunks" used for spawning entities
```
hhyyrylainen, line 23:
```suggestion
    ///   Size of "chunks" used for spawning entities
```

src/microbe_stage/PatchManager.cs
```diff
@@ -158,7 +159,7 @@ private void HandleCellSpawns(Patch patch)
 
         GD.Print("Number of species in this patch = ", patch.SpeciesInPatch.Count);
 
-        foreach (var entry in patch.SpeciesInPatch)
+        foreach (var entry in patch.SpeciesInPatch.OrderBy(entry => -entry.Value))
```
hhyyrylainen, line 162: `OrderByDescending`?

src/microbe_stage/Spawners.cs
```diff
@@ -102,109 +102,21 @@ public static void GiveFullyGrownChanceForMulticellular(Microbe microbe, Random
     // TODO: this is likely a huge cause of lag. Would be nice to be able
     // to spawn these so that only one per tick is spawned.
     public static IEnumerable<Microbe> SpawnBacteriaColony(Species species, Vector3 location,
-        Vector3 playerPosition, Node worldRoot, PackedScene microbeScene,
```
hhyyrylainen, line 105: With the playerPosition no longer being in here, does this method still make sure it doesn't spawn the colony members so far away that they can despawn instantly?

src/microbe_stage/PatchManager.cs
```diff
@@ -215,15 +214,6 @@ private void HandleCellSpawns(Patch patch)
         if (existing != null)
         {
             existing.Marked = true;
-
-            float oldFrequency = existing.Spawner.SpawnFrequency;
```
hhyyrylainen, line 219: What happened to this spawn frequency updating code?

src/microbe_stage/MicrobeStage.cs
```diff
@@ -379,6 +387,8 @@ public void SpawnPlayer()
 
         Camera.ObjectToFollow = Player;
 
+        spawner.DespawnAll();
```
hhyyrylainen, line 390: Is there a chance that this crashes the game? There was that one random guy who tried to crank up the despawn rate a ton and seemed to get more crashes that way. Anyone experienced crashes in testing?

src/microbe_stage/Spawners.cs
adQuid: No, that PR was a victim of bad timing. Colonies now have a max size of 2, and spawn well within the despawn radius.

src/microbe_stage/PatchManager.cs
adQuid: I admit I don't recall exactly since It's been a while since I made this change, but I do know that frequency and density were combined into one value, since they were just scalar multiples of each other anyway.

src/microbe_stage/MicrobeStage.cs
adQuid: @buckly90 @Nunez2196 @Oliveriver ?

hhyyrylainen: It *might* help a bit to use the `Invoke.Instance.Queue` to trigger this on the next frame as on this frame the player is already spawned, so that would spread things out a bit over different frames, though I'm not sure if that has any effect on anything...

Oliveriver: Nothing close to a crash for me.

adQuid: I could certainly add more stuff here if you want me to, but obviously with no crashes so far I can't prove that it's doing anything.

Nunez2196: i didn't experience any crashes either

src/microbe_stage/PatchManager.cs
hhyyrylainen: Are they still correctly updated, though? This even has the logging for the spawn frequency change, so my worry is that your changes actually break the spawners from actually updating...

adQuid: It is not, which is probably what accounted for the overrepresentation Nunz noticed. I've pushed a fix to correct.

hhyyrylainen: Okay so we found a bug here. @Nunez2196 can you re-check that now?

Oliveriver:
I don't want to be the bearer of bad news, but I just tested this again and ended up in a patch with zero compound clouds. Plenty of other cells, but no clouds at all for at least six death and respawn cycles. I can provide logs if needed.

adQuid:
@Oliveriver Found it. I failed to pass on a strange multiplication we do (and I contributed to, the cloud system as a whole is due for a refactor I won't attempt), resulting in cloud density decreasing by 10,000 every generation.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -253,148 +245,117 @@ private int HandleQueuedSpawns(int spawnsLeftThisFrame)
...
+            var enumerable = spawnType.Spawn(worldRoot, location);
 
-            // Check if we are out of quota for this frame
+            if (enumerable == null)
+                return spawns;
 
-            // TODO: this is a bit awkward if this stops compound clouds from spawning as well...
-            if (spawned + existing >= Constants.DEFAULT_MAX_SPAWNED_ENTITIES)
+            using var spawner = enumerable.GetEnumerator();
+            while (spawner.MoveNext())
```
hhyyrylainen, line 347: There seems to be a massive bug here, this no longer limits the spawns per frame properly here, at least I can't see this loop exiting if there are no spawns left. Was the lag spikes the reason why you removed the colony spawns or? If so then, here's the bug which started to cause the lag spikes.

Also now that you've broken the queued spawns system, the left in code to spawn queued things, is completely useless... so I very strongly say you should fix the queued spawns system to ensure no minor stuttering happens from spawning more than a couple of things per frame.

hhyyrylainen (changes requested):
I've now reviewed all the files and found some things to comment on in the SpawnSystem.cs file, including probably a bug that this broke the max spawns per frame system.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -253,148 +245,117 @@ private int HandleQueuedSpawns(int spawnsLeftThisFrame)
...
+        sectorsToSpawn.Add(new Tuple<int, int>(playerCoordinatePoint.Item1 - 2, playerCoordinatePoint.Item2 + 1));
+        sectorsToSpawn.Add(new Tuple<int, int>(playerCoordinatePoint.Item1 + 2, playerCoordinatePoint.Item2 - 1));
+        sectorsToSpawn.Add(new Tuple<int, int>(playerCoordinatePoint.Item1 + 2, playerCoordinatePoint.Item2));
+        sectorsToSpawn.Add(new Tuple<int, int>(playerCoordinatePoint.Item1 + 2, playerCoordinatePoint.Item2 + 1));
+        sectorsToSpawn.Add(new Tuple<int, int>(playerCoordinatePoint.Item1 - 1, playerCoordinatePoint.Item2 - 2));
+        sectorsToSpawn.Add(new Tuple<int, int>(playerCoordinatePoint.Item1, playerCoordinatePoint.Item2 - 2));
+        sectorsToSpawn.Add(new Tuple<int, int>(playerCoordinatePoint.Item1 + 1, playerCoordinatePoint.Item2 - 2));
+        sectorsToSpawn.Add(new Tuple<int, int>(playerCoordinatePoint.Item1 - 1, playerCoordinatePoint.Item2 + 2));
+        sectorsToSpawn.Add(new Tuple<int, int>(playerCoordinatePoint.Item1, playerCoordinatePoint.Item2 + 2));
+        sectorsToSpawn.Add(new Tuple<int, int>(playerCoordinatePoint.Item1 + 1, playerCoordinatePoint.Item2 + 2));
```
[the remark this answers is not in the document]
adQuid: I don't know an intuitive pair of nested loops that returns this very unusual pattern of 4 lines of squares, but preseting the list size makes sense.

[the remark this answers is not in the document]
hhyyrylainen: Maybe then this could be made into 4 loops? That'd still reduce the repetition a tiny bit...

src/microbe_stage/SpawnSystem.cs
```diff
@@ -65,18 +52,14 @@ public class SpawnSystem
     /// </summary>
     private int estimateEntityCount;
 
+    private int spawnsLeftThisFrame;
```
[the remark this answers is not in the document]
adQuid: I think the fate of this variable will depend on our conclusion about queues.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -253,148 +245,117 @@ private int HandleQueuedSpawns(int spawnsLeftThisFrame)
...
+
+        SpawnMicrobesAroundPlayer(playerPosition);
+    }
+
+    /// <summary>
+    ///   Handles all spawning for this section of the play area, as it will look when the player enters. Does NOT
+    ///   handle recording that the sector was spawned.
+    /// </summary>
+    /// <param name="sector">X/Y coordiates of the sector to be spawned, in SECTOR_SIZE units</param>
+    private void SpawnInSector(Tuple<int, int> sector)
```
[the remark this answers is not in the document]
adQuid: Just didn't know about Int2. Fixed

src/microbe_stage/SpawnSystem.cs
hhyyrylainen: I fail to understand why you think that using the spawn queue would impact the sectors. The logic should be:
- If there are queued spawns, handle those
- If out of spawns this frame, skip
- Otherwise enter into the spawn sectors code. When spawning there, spawn up to spawnsLeftThisFrame things and if there are more stuff to spawn, then save that as the queue



It's simply not doable to spawn more things than spawnsLeftThisFrame, it's a massive source of small lag spikes, which I don't want to put back into the game. If you have changed the logic so that multiple spawners are allowed to trigger per frame, you need to also change the spawn queueing system to allow multiple spawners to have active queues at once, then no spawns that are wanted will be forgotten.

adQuid: Unless I'm deeply misunderstanding something (and if so educate me now) queuedSpawns is not a list of QueuedSpawn, it's a single nullable instance. That means the queue has a max length of 1. That's not enough if multiple things can spawn per frame.

hhyyrylainen: It is a nullable *enumerator* and an enumerator can return as many items as it wants. If you do `foreach(var thing in list)` that uses an enumerator. The list is converted into an enumerator and then the foreach loops it. So basically we have a nullable reference to a thing that promises to generate us one or more items we can read one at a time.

adQuid: I see how that works within a single spawntype (spawning a colony), but this code: https://github.com/Revolutionary-Games/Thrive/blob/master/src/microbe_stage/SpawnSystem.cs#L391 is not appending anything, it's replacing whatever was in the variable. This would be very hard to test, but I'm pretty sure that during a lag spike that will remove the queues from any spawn types but the last one to be run.

hhyyrylainen: I'm 99% certain that that code cannot run if the previous spawner still has stuff. Look at like 193 in Process. And unless there's a bug in HandleQueuedSpawns, it won't report that there are any spawns left if it didn't manage to spawn the entire queue.

Kasterisk:
Does this close https://github.com/Revolutionary-Games/Thrive/issues/2055 or is that a completely different approach?

hhyyrylainen:
Yeah, this does not address #2055 so that's still something to consider in the future even after merging this.

adQuid:
I've modified the spawn queue to be able to support multiple spawn types. I've also got a linter error that seems incorrect, which is next for me to look at.

PR #3236, opened 2022-04-18
Delete old backups

### LINKED ISSUE #2794: Clear old saves button should delete save upgrade backups
to not leave them hanging around

KJ002:

This PR adds functionality to the delete old saves button to delete save files that are of an older game version and are also a backup.

**Related Issues**

closes #2794

**Progress Checklist**

      break existing features:
      members reviewing this PR)


KJ002:
I think it is definitely worth checking that I have updated locales properly, specifically with removing `DELETE_ALL_OLD_SAVE_WARNING` and adding `DELETE_ALL_OLD_SAVE_WARNING2`

hhyyrylainen:
> I think it is definitely worth checking that I have updated locales properly, specifically with removing DELETE_ALL_OLD_SAVE_WARNING and adding DELETE_ALL_OLD_SAVE_WARNING2. I also noticed that 

You should use the update localization script to take care of things and a gettext supporting editor to write the English translation. That way there can't be mistakes.

src/saving/SaveHelper.cs
```diff
@@ -265,6 +265,37 @@ public static (int Count, ulong DiskSpace) CountSaves(string? nameStartsWith = n
...
+                {
+                    GD.PrintErr("Can't read size of save file: ", save);
+                    continue;
+                }
+
+                ++count;
+                totalSize += file.GetLen();
+            }
+        }
+
```
Kemikal1, line 295: I personally would merge this with the other CountSaves() method. 
Maybe you tried to keep the logic of the other method clean but i usually try to keep the code lines as few as possible.

src/saving/SaveManagerGUI.cs
```diff
@@ -162,17 +167,21 @@ private void RefreshSaveCounts()
         getTotalSaveCountTask = new Task<(int Count, ulong DiskSpace)>(() => SaveHelper.CountSaves());
         getAutoSaveCountTask = new Task<(int Count, ulong DiskSpace)>(() => SaveHelper.CountSaves("auto_save"));
         getQuickSaveCountTask = new Task<(int Count, ulong DiskSpace)>(() => SaveHelper.CountSaves("quick_save"));
+        getOldBackupCountTask = new Task<(int Count, ulong DiskSpace)>(() => SaveHelper.CountOldBackupSaves());
```
Kemikal1, line 170: If you merge the method as i said you could have the ".backup" here as an argument , and change the CountSaves(nameStartsWith) argument name to "nameContains" inside the function itself.

Kemikal1 (changes requested):
This is how i would approach the problem here, maybe henri, the lead agrees with you.
Everything else looks squeaky clean.

src/saving/SaveHelper.cs
Kemikal1: Also maybe update the xml comment of that function to say what kind of saves it can count if u merge.

KJ002:
> Why did you write the warning with "2" at the end?? i think you couldve just replaced it.

I was just doing what I was told here https://github.com/Revolutionary-Games/Thrive/issues/2794#issuecomment-1101499112 but if you think that would be better I would be happy to change it.

Kemikal1:
Alright, then leave as is.

src/saving/SaveHelper.cs
KJ002: I just didn't feel like messing all that much with the function because it would likely mean adding some sort of boolean parameter to clean old save file.

src/saving/SaveManagerGUI.cs
KJ002: I but we might also have time implement some logic to check versions too, we don't want to go deleting backups with the same version as the game

src/saving/SaveHelper.cs
Kemikal1: it could be like 
`if (nameStartsWith == null || save.StartsWith(nameContains, StringComparison.CurrentCulture || (nameContains.Contains( ".backup" ) &&  SaveUpgrader.IsSaveABackup(save) &&  VersionUtils.Compare(saveInfo.ThriveVersion, Constants.Version) < 0 ) )`

hhyyrylainen:
I was actually saying that the suffix should be "_2" to make it more clear and not just "2"

src/saving/SaveHelper.cs
hhyyrylainen: Purely from memory I seem to remember we already had a save counting method, so it would make sense to adjust it slightly to be able to count old backups rather than needing a new method that looks almost the same.

I'd rework the conditions to be opposite, to fit in the extra logic to check if the name ends with something instead. I don't both options should be usable at the same time. And separate out the counting outside the if, and just leave it there to set a boolean flag to make it clean. Very rough code:

```
            if (nameStartsWith != null && !save.StartsWith(nameStartsWith, StringComparison.CurrentCulture))
            {
                 continue
             }
             if(requiredSuffix != null && doesn't match suffix)
                continue;

             if (file.Open(Path.Combine(Constants.SAVE_FOLDER, save), File.ModeFlags.Read) != Error.Ok)
              {
                    GD.PrintErr("Can't read size of save file: ", save);
                    continue;
                }

                ++count;
               totalSize += file.GetLen();

```

src/saving/SaveManagerGUI.cs
hhyyrylainen: > But we might also have time implement some logic to check versions too, we don't want to go deleting backups with the same version as the game

I would not worry about that at all. If someone has played a newer version of the game and then goes back is a very convoluted use case. Anyway if they had cleaned the old files while in the new version, their backup saves would be already gone. Too much complexity for no gain in my opinion in trying to handle that.

src/saving/SaveHelper.cs
KJ002: It might to adjust the parameters, instead of `string? nameStartsWith = null` I don't see why this couldn't be `string? nameContains = null` which does what it says on the tin

hhyyrylainen: That is not the same. If someone writes in their save in a random spot the text that identifies the different save types, then the save counting would not work. I was lazy when initially writing this so I just checked the name prefixes and assumed the save type number would be after that.
I guess this might be a case where regex is suitable (changing the method to take in a nullable regex object as a parameter), that way both cases could be properly handled.

Kemikal1: you can also name the save "quick_save_1", wouldnt that also break the count?

hhyyrylainen: Yeah, but I didn't consider it a problem if the player intentionally named saves like that as it would be pretty unlikely they do so by accident. Including the word "backup" anywhere in the save name is much more likely.

Kemikal1: I think its ".backup",meh, to mee it seems unlikely

KJ002: Would it be possible to allocate a separate directory for backups

KJ002: This should be fine to see if it contains `Constants.SAVE_BACKUP_SUFFIX` because it is defined as `public const string SAVE_BACKUP_SUFFIX = ".backup" + SAVE_EXTENSION_WITH_DOT;` which is `.backup.` and the player shouldn't be able to save files containing a dot

KJ002: > Yeah, but I didn't consider it a problem if the player intentionally named saves like that as it would be pretty unlikely they do so by accident. Including the word "backup" anywhere in the save name is much more likely.

would it be worth implementing the same suffix functionality for quick saves and auto saves, I get it's not a critical feature but it would remove this and I would be happy to work on it.

hhyyrylainen: I see regex as the only thing that can support both cases in a flexible enough way.

hhyyrylainen:
> Would it be possible to allocate a separate directory for backups

I got email about that but don't see the comment anywhere so I'll reply like this...
That would be quite a bit of work, first you'd need to modify all the save counting etc. methods to also check the extra folder. And secondly Steam cloud saves need reconfiguring to look in the sub folder.

src/saving/SaveHelper.cs
hhyyrylainen: > which is .backup. and the player shouldn't be able to save files containing a dot


As far as I know, there's nothing preventing the player from putting as many `.`s in the filename as they want.

KJ002: Shouldn't we just make the dot an illegal character for the name as a save file, or at least remove from the name for creating the savefile

KJ002: Alright, so I just need to adjust DeleteSaveOfType to take a regex expression instread

hhyyrylainen: I'll go with "yes", not having recently read through all of the save counting and deleting code so I can't exactly remember.

KJ002: I understand regex expressions need to be compiled, would it be better to take precompiled regex expression of type `Regex` or take a or take a valid regex expression with type `String`? It may be better to store these regex expressions as constants and as the `Regex` type as to avoid recompiling the expression

hhyyrylainen: 100% take in a Regex object to just need to compile the regex once and not each time the method is called, or even worse, for each different save that is checked.

KJ002:
> > Would it be possible to allocate a separate directory for backups
> 
> I got email about that but don't see the comment anywhere so I'll reply like this... That would be quite a bit of work, first you'd need to modify all the save counting etc. methods to also check the extra folder. And secondly Steam cloud saves need reconfiguring to look in the sub folder.

Yeah, I decided to delete my comment because I thought it wasn't a great idea

simulation_parameters/Constants.cs
```diff
@@ -774,6 +775,13 @@ public static class Constants
         "user://mods",
     };
 
+    /// <summary>
+    ///   Regex expresions to categorize different file types.
+    /// </summary>
+    public static readonly Regex BackupRegex = new Regex(@"^.*\.backup\.thrivesave$");
```
hhyyrylainen, line 781: These should rely on the `SAVE_EXTENSION` constant to not hard code that part.

KJ002: I wasn't quite sure how to implement this because I would have to escape "." and any other symbols that happen to be valid regex instructions

hhyyrylainen: That's why I'd use SAVE_EXTENSION as it doesn't include the dot and still shows intent that this is tightly linked to another constant.

src/saving/SaveHelper.cs
```diff
@@ -239,17 +240,17 @@ public static List<string> CreateListOfSaves(SaveOrder order = SaveOrder.LastMod
     }
 
     /// <summary>
-    ///   Counts the total number of saves and how many bytes they take up
+    ///   Counts the total number of saves with the given regex expression and how many bytes they take up
```
hhyyrylainen, line 243:
```suggestion
    ///   Counts the total number of saves with the given regular expression and how many bytes they take up
```

src/saving/SaveHelper.cs
```diff
@@ -239,17 +240,17 @@ public static List<string> CreateListOfSaves(SaveOrder order = SaveOrder.LastMod
...
     /// </summary>
-    public static (int Count, ulong DiskSpace) CountSaves(string? nameStartsWith = null)
+    public static (int Count, ulong DiskSpace) CountSaves(Regex? nameContains = null)
```
hhyyrylainen, line 245:
```suggestion
    public static (int Count, ulong DiskSpace) CountSaves(Regex? nameMatches = null)
```

as you use the "IsMatch" method and don't search if the regex is matched anywhere in the string, I think this should be renamed.

src/saving/SaveHelper.cs
```diff
@@ -295,20 +296,20 @@ public static void DeleteExcessSaves(string nameStartsWith, int maximumCount)
     }
 
     /// <summary>
-    ///   Deletes all saves with the given prefix except the latest one and returns the list of saves deleted
+    ///   Deletes all saves with the given regex expression except the
+    ///   latest one if deleteLatest is false and returns the list of saves deleted
```
hhyyrylainen, line 300: This line really begs to be separate parameter and `<returns></returns>` xml elements.

src/saving/SaveManagerGUI.cs
```diff
@@ -218,11 +230,12 @@ private void DeleteOldButtonPressed()
     {
         int autoSavesToDeleteCount = (currentAutoSaveCount - 1).Clamp(0, Settings.Instance.MaxAutoSaves);
         int quickSavesToDeleteCount = (currentQuickSaveCount - 1).Clamp(0, Settings.Instance.MaxQuickSaves);
+        int oldBackupsToDeleteCount = currentBackupCount.Clamp(0, Settings.Instance.MaxQuickSaves);
 
         deleteOldConfirmDialog.DialogText =
             string.Format(CultureInfo.CurrentCulture,
-                TranslationServer.Translate("DELETE_ALL_OLD_SAVE_WARNING"),
-                autoSavesToDeleteCount, quickSavesToDeleteCount);
+                TranslationServer.Translate("DELETE_ALL_OLD_SAVE_WARNING2"),
```
hhyyrylainen, line 237:
```suggestion
                TranslationServer.Translate("DELETE_ALL_OLD_SAVE_WARNING_2"),
```

src/saving/SaveManagerGUI.cs
```diff
@@ -218,11 +230,12 @@ private void DeleteOldButtonPressed()
     {
         int autoSavesToDeleteCount = (currentAutoSaveCount - 1).Clamp(0, Settings.Instance.MaxAutoSaves);
         int quickSavesToDeleteCount = (currentQuickSaveCount - 1).Clamp(0, Settings.Instance.MaxQuickSaves);
+        int oldBackupsToDeleteCount = currentBackupCount.Clamp(0, Settings.Instance.MaxQuickSaves);
```
hhyyrylainen, line 233: Why is this using MaxQuickSaves here? All backups should be deleted when cleaning out old things. (or a separate variable added for that, but that would be quite a bit of work for no real benefit).

KJ002: I don't really know to be honest, I saw it in the auto save and quick save and because this is just the numbers that are being displayed it does no harm so I just copied it over. I don't really see why there is a Max for auto or quick saves either to be honest.

hhyyrylainen: These are the numbers we are showing the player about what they are going to true. Not showing the actual count of things to delete is a major issue.

KJ002: So we should instead just not be clamping it and set the value to zero if the value is less than zero otherwise leave it untouched.

hhyyrylainen: I'd just use the current backup count as is. It can't be below zero anyway? I probably used Clamp above as I subtracted 1 so it could end up negative in those cases.

KJ002: The count will below zero in cases where there are 0 auto saves to delete and we are subtracting 1 which accounts for not deleting the most recent auto save.

hhyyrylainen: Yes? But how does that affect the count of backups? That's the line we are talking about here. I'm confused about what you are confused about...

hhyyrylainen: > that there will always be a number of auto saves that is less than or equal to this max clamp 

That is an unsafe assumption. The player can crank up the numbers, make a bunch of saves, and then lower them a lot. Meaning that the actual save count is unrelated to the setting.

KJ002: if thats the case we should clamp the number of auto saves deleted or something because that would mean that it unsynced with the actual number of saves deleted because we will __always__ delete all but the most recent save

hhyyrylainen: I suppose the code should actually be something like `Math.Max(currentAutoSaveCount - 1, 0)` and the same for currentQuickSaveCount. But doing that change needs to also verify that the actual deletion part doesn't have a bug, that's all I can say without having to read the code and understand it again.

simulation_parameters/Constants.cs
```diff
@@ -774,6 +775,14 @@ public static class Constants
         "user://mods",
     };
 
+    /// <summary>
+    ///   Regex expresions to categorize different file types.
+    /// </summary>
```
hhyyrylainen, line 780: This should be a normal comment as this is related to 3 variables and not just this one.

simulation_parameters/Constants.cs
```diff
@@ -774,6 +775,14 @@ public static class Constants
...
+    public static readonly Regex BackupRegex = new(@"^.*\.backup\.thrivesave$");
```
hhyyrylainen, line 781: This still doesn't use the other constant for thrivesave part.

src/saving/SaveHelper.cs
```diff
@@ -239,17 +240,17 @@ public static List<string> CreateListOfSaves(SaveOrder order = SaveOrder.LastMod
...
+    public static (int Count, ulong DiskSpace) CountSaves(Regex? nameMatches = null)
     {
         int count = 0;
         ulong totalSize = 0;
 
         using var file = new File();
         foreach (var save in CreateListOfSaves())
         {
-            if (nameStartsWith == null || save.StartsWith(nameStartsWith, StringComparison.CurrentCulture))
+            if (nameMatches?.IsMatch(save) == null)
```
hhyyrylainen, line 253: Isn't this logic wrong? That's because IsMatch returns "true" when it matches, which is not equal to null.

KJ002: I am fairly new to this syntax in C# coming from more of a C background but if I were comparing two values like true and null I would expect to to give me a compiler build error and this is also exactly what I saw relating to a stack overflow post on it. The previous way it was done fails the jetbrains CI.

simulation_parameters/Constants.cs
KJ002: My bad, I got caught up in working on other things.

src/saving/SaveHelper.cs
hhyyrylainen: What you are basically checking for here is equivalent to `nameMatches == null || nameMatches.IsMatch(save) == null` I think. Whereas you really should be checking for `nameMatches == null || nameMatches.IsMatch(save) == true` Which I think translates to (I usually let my IDE do this): `nameMatches?.IsMatch(save) != false`

hhyyrylainen: That should be correct because null doesn't equal false so we are checking that the value is not null, and if the nullable call returned true, that also doesn't equal false.

KJ002: Alright I shall look into tonight when I have some time

Kasterisk:
What's the status on this PR?

KJ002:
> What's the status on this PR?

I think it's good to go but if there are any changes that need to be made I am happy to make them.

hhyyrylainen:
If Github is showing me things correctly, there's been no new commits after my latest comment: https://github.com/Revolutionary-Games/Thrive/pull/3236#commitcomment-72465024

KJ002:
Is this a time where I should be squashing commits? Never really done it before.

hhyyrylainen:
That's unnecessary we do squash merges on Github so we don't require PR commit histories to be clean. What you should do is handle my latest comment by either changing the code or discussing it with why you think the current way is the way it should be.

hhyyrylainen:
I'll finish the code myself to get this out of the way so I can get to helping finishing the big features that should get done before 0.5.9

hhyyrylainen (approved):
I did the changes I still wanted to see and verified myself with a simple playtest that this seems to work correctly in the normal use case.

Kasterisk (approved):

PR #3243, opened 2022-04-21
Despawn fixing and better tools to monitor it

hhyyrylainen:
Adds a new metrics display to see game entity counts
Fixes some pretty glaring problems in the despawning system

**Related Issues**

https://community.revolutionarygamesstudio.com/t/issues-related-to-despawning/4638/2

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -84,7 +84,7 @@ public static class Constants
     ///   Extra radius added to the spawn radius of things to allow them to move in the "wrong" direction a bit
     ///   without causing them to despawn instantly
     /// </summary>
-    public const int DESPAWN_RADIUS_OFFSET_SQUARED = 150;
+    public const int DESPAWN_RADIUS_OFFSET_SQUARED = 2500;
 
     public const float STARTING_SPAWN_DENSITY = 70000.0f;
     public const float MAX_SPAWN_DENSITY = 20000.0f;
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -504,7 +504,12 @@ public static class Constants
     /// <summary>
     ///   Delete a max of this many entities per step to reduce lag from deleting tons of entities at once.
     /// </summary>
-    public const int MAX_DESPAWNS_PER_FRAME = 1;
+    public const int MAX_DESPAWNS_PER_FRAME = 2;
+
+    /// <summary>
+    ///   How often despawns happen on top of the normal despawns that are part of the spawn cycle
+    /// </summary>
+    public const float DESPAWN_INTERVAL = 0.08f;
 
     public const float CHANCE_MULTICELLULAR_SPAWNS_GROWN = 0.1f;
     public const float CHANCE_MULTICELLULAR_SPAWNS_PARTLY_GROWN = 0.3f;
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -80,6 +80,12 @@ public static class Constants
     public const int MICROBE_SPAWN_RADIUS = 170;
     public const int CLOUD_SPAWN_RADIUS = 170;
 
+    /// <summary>
+    ///   Extra radius added to the spawn radius of things to allow them to move in the "wrong" direction a bit
+    ///   without causing them to despawn instantly
+    /// </summary>
+    public const int DESPAWN_RADIUS_OFFSET_SQUARED = 150;
+
     public const float STARTING_SPAWN_DENSITY = 70000.0f;
     public const float MAX_SPAWN_DENSITY = 20000.0f;
     public const float MIN_SPAWN_RADIUS_RATIO = 0.95f;
--- a/src/microbe_stage/SpawnSystem.cs
+++ b/src/microbe_stage/SpawnSystem.cs
@@ -92,7 +92,7 @@ public class SpawnSystem
     public static void AddEntityToTrack(ISpawned entity,
         float radius = Constants.MICROBE_SPAWN_RADIUS)
     {
-        entity.DespawnRadiusSquared = (int)(radius * radius);
+        entity.DespawnRadiusSquared = (int)(radius * radius) + Constants.DESPAWN_RADIUS_OFFSET_SQUARED;
         entity.EntityNode.AddToGroup(Constants.SPAWNED_GROUP);
     }
 
@@ -444,11 +444,7 @@ public class SpawnSystem
     /// </summary>
     private void ProcessSpawnedEntity(ISpawned entity, Spawner spawnType)
     {
-        // I don't understand why the same
-        // value is used for spawning and
-        // despawning, but apparently it works
-        // just fine
-        entity.DespawnRadiusSquared = spawnType.SpawnRadiusSquared;
+        entity.DespawnRadiusSquared = spawnType.SpawnRadiusSquared + Constants.DESPAWN_RADIUS_OFFSET_SQUARED;
 
         entity.EntityNode.AddToGroup(Constants.SPAWNED_GROUP);
     }
--- a/src/microbe_stage/SpawnSystem.cs
+++ b/src/microbe_stage/SpawnSystem.cs
@@ -145,13 +145,21 @@ public class SpawnSystem
     {
         queuedSpawns = null;
 
+        int despawned = 0;
+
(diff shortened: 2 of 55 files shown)


src/microbe_stage/SpawnSystem.cs
```diff
@@ -444,11 +444,7 @@ private int DespawnEntities(Vector3 playerPosition)
     /// </summary>
     private void ProcessSpawnedEntity(ISpawned entity, Spawner spawnType)
     {
-        // I don't understand why the same
-        // value is used for spawning and
-        // despawning, but apparently it works
-        // just fine
-        entity.DespawnRadiusSquared = spawnType.SpawnRadiusSquared;
+        entity.DespawnRadiusSquared = spawnType.SpawnRadiusSquared + Constants.DESPAWN_RADIUS_OFFSET_SQUARED;
```
Leinourdian, line 447: This (and line 95) is technically incorrect mathwise. In math, (a + b)^2 = a^2 + 2ab + b^2, not a^2 + b^2. Of course this works too if you just figure out a fitting number for `Constants.DESPAWN_RADIUS_OFFSET_SQUARED`, but it may be harder to estimate a good value, and I think the math would make the offset effectively shrink if the spawn radius were increased. This could be solved by changing the formula or by adding the offset before squaring. From my testing it seems that the offset unsquared should be at least about 50 to prevent parts of bacterial colonies from spawning outside the despawn radius.

simulation_parameters/Constants.cs
```diff
@@ -504,7 +504,12 @@ public static class Constants
     /// <summary>
     ///   Delete a max of this many entities per step to reduce lag from deleting tons of entities at once.
     /// </summary>
-    public const int MAX_DESPAWNS_PER_FRAME = 1;
+    public const int MAX_DESPAWNS_PER_FRAME = 2;
```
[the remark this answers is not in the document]
hhyyrylainen: Migrating happens maybe once every 5 minutes, and during the editor exiting loading screen. So that causing a lag spike there is not a problem. However during gameplay random lag spikes should be avoided.

Unless loading older saves the entity count shouldn't get out of hand in the first place.

src/microbe_stage/SpawnSystem.cs
hhyyrylainen: I guess if  you want to consider the squared distance to be calculated from the final distance. I just added an extra offset on top of the calculated value to make despawning not happen immediately. Though perhaps I should increase the offset a bit?

simulation_parameters/Constants.cs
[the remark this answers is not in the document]
Leinourdian: All right. How much lag does despawning cause? In my previous testing I experienced some jittering in FPS but I couldn't tell if it was from spawning or despawning.

I'll try testing this branch and see how it feels.

src/microbe_stage/SpawnSystem.cs
hhyyrylainen: I upped the offset to that.

Leinourdian:
- After testing I can say the despawning rate felt good. The amount of entities never got very high. When i forced the entity count high with cheats and then left the area, it still went down in a few seconds and felt fine. If spawning rate ever gets increased significantly enough to make despawning struggle to keep up with it, then this can be revisited, but this is good for now in my opinion. 
I also tested despawning in another branch and despawning a lot of entities at once can indeed cause a small lag spike,
so I think your approach is good.

- The new metrics display seems to be the foremost GUI element. It even goes in front of the "Confirm Exit" window.
This means that it can happen to block you from exiting the game since you can't press blocked buttons and you can't interact with the metrics window when the exit window is open. You can still exit with ESC luckily.


- I did some more testing with the despawn radius offset and it turns out it's still low enough for some cells to spawn outside of it. In my previous testing I used a reduced spawn radius to observe the despawning directly and it seemed fine. Then I did some more math and confirmed in testing that increasing the spawn radius effectively decreases the despawn offset a lot.
This can be tested in-game by sitting still, turning AI off and watching the metrics to see if anything despawns. If it does, then it means that it spawned outside the despawn radius. The value 2500 is definitely better than nothing, but to get the effect I was looking for you'd have to replace the value with 50 * 50 + 50 * 170 * 2 = 19500. With that value I didn't observe anything despawning immediately after spawning (although I think it's still technically possible, just super rare). It matches exactly with the unsquared offset of 50 and is still low enough that it's easy to despawn cells by swimming away.
I would personally replace the squared offset with the unsquared offset of 50 and move the math elsewhere. This way it would be easy to change the offset value in the future, it'd be more readable and changing the spawning radius wouldn't change the effective offset. This is actually very simple to code: just go the two places where the `DespawnRadiusSquared` is calculated, calculate despawn radius = spawn radius + offset, and then square it.

And that's that. I hope I'm not being too pedantic or wordy. You can ignore me if that's case. :)
Overall I really like the changes and additions.

hhyyrylainen:
> I would personally replace the squared offset with the unsquared offset of 50 and move the math elsewhere. This way it would be easy to change the offset value in the future, it'd be more readable and changing the spawning radius wouldn't change the effective offset. This is actually very simple to code: just go the two places where the DespawnRadiusSquared is calculated, calculate despawn radius = spawn radius + offset, and then square it.

I don't want to do a unnecessary square root call. That's the reason why squared distances are used in the first place... though I admit in this case it would be just one call per spawned entity and not for each entity each spawn cycle, but still the idea of wasting performance in a game seems wrong to me.

What was the squared value you suggested using as the offset? I can't tell from your math which value I'd need to use for it.

Leinourdian:
> I don't want to do a unnecessary square root call. That's the reason why squared distances are used in the first place... though I admit in this case it would be just one call per spawned entity and not for each entity each spawn cycle, but still the idea of wasting performance in a game seems wrong to me.

You don't need to take a square root. I meant that _despawn radius squared = (spawn radius + offset) * (spawn radius + offset)_. You can find the spawn radius without needing to take a square root (`spawnType.SpawnRadius` or `Constants.MICROBE_SPAWN_RADIUS`) and you can change the offset to 50.

> What was the squared value you suggested using as the offset? I can't tell from your math which value I'd need to use for it.

19500

hhyyrylainen:
> You don't need to take a square root. I meant that despawn radius squared = (spawn radius + offset) * (spawn radius + offset). You can find the spawn radius without needing to take a square root (spawnType.SpawnRadius or Constants.MICROBE_SPAWN_RADIUS) and you can change the offset to 50.

There is code where only the squared spawn radius variable is available:

https://github.com/Revolutionary-Games/Thrive/blob/7932b9e32fc8f1de1545d9a0c1cc844819b1c3a0/src/microbe_stage/SpawnSystem.cs#L447

Also 19500 is already 140 when you take the square root. With the default spawn distance of 170, that seems extremely high margin to me... I think I'll just open a separate issue about making sure bacteria colony spawns don't despawn instantly.

Leinourdian:
> There is code where only the squared spawn radius variable is available:

You can change `spawnType.SpawnRadiusSquared` to `spawnType.SpawnRadius`. That worked for me at least.

hhyyrylainen:
Hmm yeah, that's the spawner object, I thought it was the spawned object, which doesn't have that property in the interface. Anyway I'll leave it as is, I opened that another issue about fixing the colony spawns. I don't want to increase testing burden again on this issue with greatly increasing the despawn radius which may make the game world feel much more empty.

Leinourdian:
Ok. The despawn radius isn't increased that much (comparing the offset 50 to the spawn radius 170), and I don't think this makes the world less lively since it just makes entities despawn a bit further away. But this isn't urgent and we can well leave this to the other issue.

Now I consider this PR good to go. Do you want feedback from other people?

hhyyrylainen:
I already got quite a bit from other developers on discord, which is why I really didn't majorly want to rework the radius. The testing was actually very good, couple of people got almost 50% performance improvement in late game.

Leinourdian (approved):
Good to go! This can be merged.

Also, awesome to hear that it's really helping with performance!

RacerBG (approved):

Kasterisk:
Can we change the performance metrics content's font to `Lato-Regular-AlmostSmall.tres`? I'm a bit irritated by the size of the texts.

hhyyrylainen:
You can (I assume it won't become *really* tiny) in a separate PR as this is already merged.

PR #3254, opened 2022-04-22
Removed full screen mode

84634E1A607A:

Removed full screen mode in the project file.

The reason I do this is that on Windows, if you open this option, the initializing period is full-screened which is annoying; and the window is sized to nearly full screen (instead of 1280x720, default project window size)

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -58,7 +58,6 @@ settings/fps/force_fps=400
 
 window/size/width=1280
 window/size/height=720
-window/size/fullscreen=true
 window/dpi/allow_hidpi=true
 mouse_cursor/custom_image="res://assets/textures/gui/cursors/default.png"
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
The reason I set the game to default to fullscreen is that it's much more likely to run on dedicated graphics than integrated graphics. It is a bit annoying for me as well, but that's the sacrifice I think we need to make.

84634E1A607A:
What's the difference though? You can always enter full screen with F11

hhyyrylainen:
Once the graphics driver loader in the OS sees that the game has started in windowed mode and has picked intel integrated graphics, it is no longer possible to switch to dedicated nvidia graphics. So that's why it's very important which mode the game starts in by default.

84634E1A607A:
Okay. I'll leave this as is.

PR #3265, opened 2022-04-24
Prevent player getting stuck when leaving multicellular editor.

adQuid:

Uses a very simple algorithm to pull the parent away from the player. The player may still be stuck between cells if the species uses very big cells.

**Related Issues**

(I think there's an issue for this, but I couldn't find it)

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/MicrobeStage.cs
+++ b/src/microbe_stage/MicrobeStage.cs
@@ -671,7 +671,9 @@ public class MicrobeStage : NodeWithInput, IReturnableGameState, IGodotEarlyNode
         {
             daughter.BecomeFullyGrownMulticellularColony();
 
-            // TODO: add more extra offset between the player and the divided cell
+            // Back the daughter away so the player isn't stuck inside
+            while (daughter.Colony!.ColonyMembers.Any(member => member.GlobalTransform.origin.DistanceSquaredTo(Player.GlobalTransform.origin) < 50.0f))
+                daughter.Translation += new Vector3(0.0f, 0, -1.0f);
         }
 
         // Update the player's cell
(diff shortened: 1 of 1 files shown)


src/microbe_stage/MicrobeStage.cs
```diff
@@ -671,7 +671,12 @@ public void OnReturnFromEditor()
         {
             daughter.BecomeFullyGrownMulticellularColony();
 
-            // TODO: add more extra offset between the player and the divided cell
+            // Back the daughter away so the player isn't stuck inside
+            while (daughter.Colony!.ColonyMembers.Any(member =>
```
hhyyrylainen, line 675: I think this logic should probably go in Microbe.Divide where I already did custom separation logic for colonies.

adQuid: That won't work because this method calls BecomeFullyGrownMulticellularColony() is called after divide.

hhyyrylainen: What about the comment few lines above: `// This is done first to ensure that the player colony is still intact for spawn separation calculation`

My new code for colony split separation definitely gets called, after all if you compare the RC to 0.5.8 the separation is obviously different.

So, really I'd like you to improve this bit of code (found in Microbe.Divide):
```
            // When in a colony we approximate a much higher separation distance
            var colonyRadius = separation.x;

            foreach (var colonyMember in Colony.ColonyMembers)
            {
                if (colonyMember == this)
                    continue;

                var radius = colonyMember.Radius + Constants.COLONY_DIVIDE_EXTRA_DAUGHTER_OFFSET;

                // TODO: switch this to something else if this is too slow for large colonies
                var positionInColony = colonyMember.GlobalTransform.origin - currentPosition;

                var outerRadius = Math.Max(Math.Abs(positionInColony.x) + radius,
                    Math.Abs(positionInColony.z) + radius);

                if (outerRadius > colonyRadius)
                    colonyRadius = outerRadius;
            }

            separation = new Vector3(colonyRadius + Constants.COLONY_DIVIDE_EXTRA_DAUGHTER_OFFSET, 0, 0);
```
and preferably you would come up with an algorithm that isn't a bunch of steps * the number of colony members of complexity.

PR #3271, opened 2022-04-26
Fix colony spawn

### LINKED ISSUE #3245: Fix microbe colony spawns being able to spawn cells so far that they despawn instantly
we need to either adjust the despawn radius of the farther away cells that are spawned or adjust the logic to not be able to spawn in such a wide area when the colony spawn is decided to make sure the edges aren't too far away.

Leinourdian:

This pull request aims to fix bacterial colony members spawning outside of the despawn radius.  It does three things:
1. Refactor the despawn radius calculation slightly so that it's easier to work with.
2. Increase the despawn radius so that its 170 + 50. This makes it much rarer for colonies to get "cut off".
3. Add a check to bacterial colony spawning that skips spawning cells if they'd spawn outside of the despawn radius. This required giving its spawn method access to the player's position.

After these changes, spawning and despawning looks like this:
Spawn and despawn radius

Observations I made testing these changes:

- It works.
- The amount of existing entities is increased a bit, by about 5 to 10. This makes sense since fewer colonies get cut off and there's more room for cells.
- The density of entities during gameplay seems to be about the same. Which is what I expected, since if I understand correctly, existing entities don't affect spawning unless you've hit the entity limit or if you're sitting still.
- Increased despawn radius has a nice side effect of giving floating chunks a slightly better sense of object permanence.

**Related Issues**
fixes #3245

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -84,7 +84,7 @@ public static class Constants
     ///   Extra radius added to the spawn radius of things to allow them to move in the "wrong" direction a bit
     ///   without causing them to despawn instantly
     /// </summary>
-    public const int DESPAWN_RADIUS_OFFSET = 7;
+    public const int DESPAWN_RADIUS_OFFSET = 50;
 
     public const float STARTING_SPAWN_DENSITY = 70000.0f;
     public const float MAX_SPAWN_DENSITY = 20000.0f;
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -84,7 +84,7 @@ public static class Constants
     ///   Extra radius added to the spawn radius of things to allow them to move in the "wrong" direction a bit
     ///   without causing them to despawn instantly
     /// </summary>
-    public const int DESPAWN_RADIUS_OFFSET_SQUARED = 2500;
+    public const int DESPAWN_RADIUS_OFFSET = 7;
 
     public const float STARTING_SPAWN_DENSITY = 70000.0f;
     public const float MAX_SPAWN_DENSITY = 20000.0f;
--- a/src/microbe_stage/SpawnSystem.cs
+++ b/src/microbe_stage/SpawnSystem.cs
@@ -90,9 +90,9 @@ public class SpawnSystem
     ///   Adds an externally spawned entity to be despawned
     /// </summary>
     public static void AddEntityToTrack(ISpawned entity,
-        float radius = Constants.MICROBE_SPAWN_RADIUS)
+        float radius = Constants.MICROBE_SPAWN_RADIUS + Constants.DESPAWN_RADIUS_OFFSET)
     {
-        entity.DespawnRadiusSquared = (int)(radius * radius) + Constants.DESPAWN_RADIUS_OFFSET_SQUARED;
+        entity.DespawnRadiusSquared = (int)(radius * radius);
         entity.EntityNode.AddToGroup(Constants.SPAWNED_GROUP);
     }
 
@@ -444,7 +444,8 @@ public class SpawnSystem
     /// </summary>
     private void ProcessSpawnedEntity(ISpawned entity, Spawner spawnType)
     {
-        entity.DespawnRadiusSquared = spawnType.SpawnRadiusSquared + Constants.DESPAWN_RADIUS_OFFSET_SQUARED;
+        float radius = spawnType.SpawnRadius + Constants.DESPAWN_RADIUS_OFFSET;
+        entity.DespawnRadiusSquared = (int)(radius * radius);
 
         entity.EntityNode.AddToGroup(Constants.SPAWNED_GROUP);
     }
--- a/src/microbe_stage/SpawnSystem.cs
+++ b/src/microbe_stage/SpawnSystem.cs
@@ -332,7 +332,7 @@ public class SpawnSystem
                         squaredDistance >= spawnType.MinSpawnRadiusSquared)
                     {
                         // Second condition passed. Spawn the entity.
-                        if (SpawnWithSpawner(spawnType, playerPosition + displacement, existing,
+                        if (SpawnWithSpawner(spawnType, playerPosition + displacement, playerPosition, existing,
                                 ref spawnsLeftThisFrame, ref spawned))
                         {
                             estimateEntityCountInSpawnRadius += spawned;
@@ -356,10 +356,10 @@ public class SpawnSystem
     ///   Does a single spawn with a spawner
     /// </summary>
     /// <returns>True if we have exceeded the spawn limit and no further spawns should be done this frame</returns>
-    private bool SpawnWithSpawner(Spawner spawnType, Vector3 location, int existing, ref int spawnsLeftThisFrame,
-        ref int spawned)
+    private bool SpawnWithSpawner(Spawner spawnType, Vector3 location, Vector3 playerPosition, int existing,
+        ref int spawnsLeftThisFrame, ref int spawned)
     {
-        var enumerable = spawnType.Spawn(worldRoot, location);
+        var enumerable = spawnType.Spawn(worldRoot, location, playerPosition);
 
         if (enumerable == null)
             return false;
--- a/src/microbe_stage/Spawners.cs
+++ b/src/microbe_stage/Spawners.cs
@@ -103,8 +103,8 @@ public static class SpawnHelpers
     // TODO: this is likely a huge cause of lag. Would be nice to be able
     // to spawn these so that only one per tick is spawned.
     public static IEnumerable<Microbe> SpawnBacteriaColony(Species species, Vector3 location,
-        Node worldRoot, PackedScene microbeScene, CompoundCloudSystem cloudSystem,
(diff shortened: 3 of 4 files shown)


src/microbe_stage/SpawnSystem.cs
```diff
@@ -90,9 +90,9 @@ public SpawnSystem(Node root)
     ///   Adds an externally spawned entity to be despawned
     /// </summary>
     public static void AddEntityToTrack(ISpawned entity,
-        float radius = Constants.MICROBE_SPAWN_RADIUS)
+        float radius = Constants.MICROBE_SPAWN_RADIUS + Constants.DESPAWN_RADIUS_OFFSET)
     {
-        entity.DespawnRadiusSquared = (int)(radius * radius) + Constants.DESPAWN_RADIUS_OFFSET_SQUARED;
+        entity.DespawnRadiusSquared = (int)(radius * radius);
```
Maxonovien, line 95: Wouldn't a power function be a bit clearer there? (and some other places around)

src/microbe_stage/Spawners.cs
```diff
@@ -103,8 +103,8 @@ public static void GiveFullyGrownChanceForMulticellular(Microbe microbe, Random
     // TODO: this is likely a huge cause of lag. Would be nice to be able
     // to spawn these so that only one per tick is spawned.
     public static IEnumerable<Microbe> SpawnBacteriaColony(Species species, Vector3 location,
-        Node worldRoot, PackedScene microbeScene, CompoundCloudSystem cloudSystem,
-        GameProperties currentGame, Random random)
+        Vector3 playerPosition, Node worldRoot, PackedScene microbeScene,
+        CompoundCloudSystem cloudSystem, GameProperties currentGame, Random random)
     {
         var curSpawn = new Vector3(random.Next(1, 8), 0, random.Next(1, 8));
```
Maxonovien, line 109: I know this isn't your code, but maybe it could help to rename this variable to something more explicit (does 'cur' stand for 'current'?)

Maxonovien (commented):
Just quickly scanned through for popup remarks, nothing shocking otherwise. But it should be given a closer look anyways.

src/microbe_stage/SpawnSystem.cs
Leinourdian: Perhaps, I just used what was already there. I can change that if wanted.

src/microbe_stage/Spawners.cs
Leinourdian: Yeah, it stands for 'current'. Would currentSpawn be good or something else?

RacerBG:
I like it, the game world still feels alive and there are a lot of microbes around. The spawn system does it's job pretty well and except for one lag spike my FPS stayed above 30 while in Freebuild mode.

RacerBG (approved):

src/microbe_stage/SpawnSystem.cs
```diff
@@ -90,9 +90,9 @@ public SpawnSystem(Node root)
     ///   Adds an externally spawned entity to be despawned
     /// </summary>
     public static void AddEntityToTrack(ISpawned entity,
-        float radius = Constants.MICROBE_SPAWN_RADIUS)
+        float radius = Constants.MICROBE_SPAWN_RADIUS + Constants.DESPAWN_RADIUS_OFFSET)
     {
-        entity.DespawnRadiusSquared = (int)(radius * radius) + Constants.DESPAWN_RADIUS_OFFSET_SQUARED;
+        entity.DespawnRadiusSquared = (int)Mathf.Pow(radius, 2.0f);
```
hhyyrylainen, line 95: Someone may have commented on this already, but I'd vastly prefer to keep the "radius * radius" code because I fear how inefficient calling Mathf.Pow here would be, because it converts all floats to doubles and might not have an optimized path for when the exponent is an integer.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -444,7 +444,8 @@ private int DespawnEntities(Vector3 playerPosition)
     /// </summary>
     private void ProcessSpawnedEntity(ISpawned entity, Spawner spawnType)
     {
-        entity.DespawnRadiusSquared = spawnType.SpawnRadiusSquared + Constants.DESPAWN_RADIUS_OFFSET_SQUARED;
+        float radius = spawnType.SpawnRadius + Constants.DESPAWN_RADIUS_OFFSET;
+        entity.DespawnRadiusSquared = (int)Mathf.Pow(radius, 2.0f);
```
hhyyrylainen, line 448: and here

src/microbe_stage/Spawners.cs
```diff
@@ -115,12 +115,18 @@ public static void GiveFullyGrownChanceForMulticellular(Microbe microbe, Random
...
                      Constants.MAX_BACTERIAL_COLONY_SIZE + 1); i++)
             {
-                // Dont spawn them on top of each other because it
-                // causes them to bounce around and lag
-                yield return SpawnMicrobe(species, location + curSpawn, worldRoot, microbeScene, true,
-                    cloudSystem, currentGame);
+                // Skip spawning this microbe if it would spawn outside of the despawn radius
+                float distanceSquared = (playerPosition - (location + locationOffset)).LengthSquared();
+                float despawnRadius = Constants.MICROBE_SPAWN_RADIUS + Constants.DESPAWN_RADIUS_OFFSET;
+                if (distanceSquared < Mathf.Pow(despawnRadius, 2.0f))
```
hhyyrylainen, line 121: I'd make `Mathf.Pow(despawnRadius, 2.0f)` into a new constant in Constants to make sure it isn't calculated during runtime.

src/microbe_stage/SpawnSystem.cs
Leinourdian: Gotcha, I'll do that to this and the other places.

src/microbe_stage/Spawners.cs
```diff
@@ -301,17 +314,23 @@ private static IEnumerable<Microbe> MicrobeColonySpawnHelper(ColonySpawnInfo col
...
+                colony.LocationOffset.z += colony.Random.Next(5, 8);
+                colony.LocationOffset.x += colony.Random.Next(-2, 3);
             }
 
-            yield return SpawnMicrobe(colony.Species, location + colony.CurSpawn, colony.WorldRoot,
-                colony.MicrobeScene, true, colony.CloudSystem, colony.CurrentGame);
+            // Skip spawning this microbe if it would spawn outside of the despawn radius
+            float distanceSquared = (playerPosition - (location + colony.LocationOffset)).LengthSquared();
+            float despawnRadius = Constants.MICROBE_SPAWN_RADIUS + Constants.DESPAWN_RADIUS_OFFSET;
+            if (distanceSquared < Mathf.Pow(despawnRadius, 2.0f))
```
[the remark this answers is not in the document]
Leinourdian: I can move the math to the beginning of the method so it's calculated just once. I think that would be simpler than making a helper method.

src/microbe_stage/Spawners.cs
hhyyrylainen: If the value is used a lot in a squared manner, yes I'd like to see that be a constant.

src/microbe_stage/Spawners.cs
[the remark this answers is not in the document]
hhyyrylainen: I thought I saw this being the exact same in 2 methods... but if it is actually just one, then  that would be fine as well.

src/microbe_stage/Spawners.cs
Leinourdian: ok

src/microbe_stage/Spawners.cs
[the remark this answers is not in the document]
Leinourdian: Oh right, it's in two methods and also the distanceSquared updates for each microbe... Maybe I'll go with the helper method.

simulation_parameters/Constants.cs
```diff
@@ -84,7 +84,17 @@ public static class Constants
...
     /// </summary>
-    public const int DESPAWN_RADIUS_OFFSET_SQUARED = 2500;
+    public const int DESPAWN_RADIUS_OFFSET = 50;
+
+    /// <summary>
+    ///   Squared radius. Microbes despawn if they are outside it.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     The value should be (MICROBE_SPAWN_RADIUS + DESPAWN_RADIUS_OFFSET)^2
```
hhyyrylainen, line 94: Why not make this math then be that? As long as those variables are also const, you can put the calculation in here so that it cannot get out of sync.

Leinourdian: Right, that makes more sense. I'll do that. I think I can remove the summary then because the formula and names explain the constant.

hhyyrylainen: "Microbes despawn if they are outside it" is pretty useful comment, so putting that on the despawn radius if it is missing from there would be nice.

hhyyrylainen (approved):

PR #3278, opened 2022-04-27
Procedural patchmap

### LINKED ISSUE #682: Add procedural patch map generator
This is a pretty general issue for making some kind of procedural generation of the patch map.
This can be pretty fancy if it takes into account the procedurally generated solar system like: surface temperature etc.

Kemikal1:

Adds random patch-map
**Related Issues**
closes #682 

**Progress Checklist**

- [x]  Add all patch types and connect the whole map
- [x] Zoom  
- [x] translation on map
- [x] Fix lines and fancier drawing.
- [x] better current patch highlighting and focus
- [x] link the new game settings in
- [ ]  PR author has checked that this PR works as intended and doesn't
      break existing features:
      members reviewing this PR)


Kemikal1:
Pretty big PR by itself already so, i think ill leave the zoom for another round.

hhyyrylainen:
Before reviews and testing properly begins, could you fix the merge conflicts and CI issues so that this is in a cleaner state for people to take a look at?

Kemikal1:
loading doesnt work so if people test this keep it in mind.

Oliveriver:
@Kemikal1 What's left to do for this PR? I could have a go at helping at some point soon to unblock my PR.

GameDungeon:
Honestly running a game to see how this looks, at least in its current state, I really don't like it. It will need more work before it should be merged.

Kemikal1:
@Oliveriver It needs to be anble to save and load, right now the loading is broken.

Kemikal1:
Ready for review

Oliveriver:
This doesn't yet respect the map-related new game settings. Specifically, there's a setting for starting location, and there's a setting for switching between a procedural map and the predefined map (the UI option for the latter is currently invisible, but this PR should make it visible and functional). Neither of those affect anything in this branch yet.

Other than that, my only real issue is navigation, i.e. the lack of zooming or clicking and dragging. But I know you said you wanted to leave those out of this PR. I'd be willing to have a go at it anyway (but I suspect @Kasterisk might be better suited for this kind of advanced UI issue).

Current patch highlighting and scroll centering do make a lot of difference though, good work with those.

hhyyrylainen:
> Other than that, my only real issue is navigation, i.e. the lack of zooming or clicking and dragging. But I know you said you wanted to leave those out of this PR. I'd be willing to have a go at it anyway (but I suspect @Kasterisk might be better suited for this kind of advanced UI issue).

Did we at least get the on screen instructions for shift scrolling etc. which I said I'd be fine with initially?

Kemikal1:
No i didnt get those instructions, i think those should be added after better map movement.

hhyyrylainen:
We have 2 weeks until release. People will *have* to be able to play with this as is, because I'm not very helpful that a saviour would swoop in at the last moment to fix things up for the release.

Kasterisk:
I'm available for help regarding that if need be hopefully.

Oliveriver:
I added click-and-drag navigation and a button to re-centre on the current patch, which I feel go a long way to improving navigation. I couldn't get zooming to work though, I think I'll have to defer to someone else if people still think it's necessary.

84634E1A607A:
Some UI related suggestions:

 - The border is a little bit too thick (I think)
 - The flashing green border is irritating. You can use a darker colour or reduce the frequency it flashes
 - You may use another color of the border to mark where you currently can move to. I feel overwhelmed by the current navigating system...
 - Some lines between patches don't make sense to me. You may separate lines that really isn't linked together.
 (Such as this one, it seems to be connected but actually it isn't image )

Kemikal1:
- I think that the border is 1px more wide than the lines connecting regions. Do you think it would look better if it was the same thickness?
-Darker green is already used, ill try to pulse it less then.
-what color are you suggesting for marking moveable patches? It has to be the same color as the light highlighting.
- planar graph drawing is a research topic, even though the orthogonal approach i used looks less complicated i tried to make it as nice as possible without me trying to learn some algos that i dont have the time currently.

hhyyrylainen:
> I couldn't get zooming to work though, I think I'll have to leave that to someone else if people still think it's necessary.

I think we can leave zooming for later. I imagine the drag by mouse to move around view would also be useful for tech webs etc. as well where zooming in and out will be a bit more important.'

Also excellent naming, DraggableScrollContainer would have been exactly what I'd have named it.

Oliveriver:
> Added zooming.

Nicely done! Patch map navigation is now seamless. However this has broken the scroll wheel for everything else in the editor (organelle list, patch info box, etc.). Also the scrollbars never seem to show up even when they should.

hhyyrylainen:
> However this has broken the scroll wheel for everything else in the editor (organelle list, patch info box, etc.)

https://github.com/Revolutionary-Games/Thrive/pull/3278#commitcomment-77454979

Oliveriver:
There should probably also be max and min zoom amounts.

hhyyrylainen:
> There should probably also be max and min zoom amounts.

Sounds like a good idea. (I recently added min and max zoom for the 3D editor camera)

Kasterisk:
Will fix these soon.

src/general/NewGameSettings.tscn
```diff
@@ -761,7 +761,6 @@ __meta__ = {
 }
 
 [node name="HBoxContainer" type="HBoxContainer" parent="CenterContainer/VBoxContainer/AdvancedOptions/Planet/VBoxContainer"]
-visible = false
```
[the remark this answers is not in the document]
Oliveriver: Yes. I added a button to switch map types in the new game settings PR but made it invisible as it didn't yet do anything. This PR makes it functional and hence visible.

src/microbe_stage/PatchMap.cs
```diff
@@ -17,6 +27,12 @@ public class PatchMap
     [JsonProperty]
     public Dictionary<int, Patch> Patches { get; private set; } = new();
 
+    [JsonProperty]
+    public Dictionary<int, PatchRegion> Regions { get; private set; } = new();
+
+    [JsonProperty]
+    public Dictionary<int, PatchRegion> SpecialRegions { get; private set; } = new();
```
hhyyrylainen, line 34: Please add a comment explaining what's so "special" about this. There's two very similarly named properties next to each other.

src/microbe_stage/PatchMap.cs
```diff
@@ -7,8 +7,18 @@
 /// <summary>
 ///   A container for patches that are joined together
 /// </summary>
-public class PatchMap
+[UseThriveSerializer]
+public class PatchMap : ISaveLoadable
 {
+    [JsonProperty]
+    public Random Seed = null!;
```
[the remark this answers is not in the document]
hhyyrylainen: Also is it useful to allow changing the seed *after* creating a map. I get the feeling also that this should really be a property, the map is after all a complex class.

src/microbe_stage/PatchMap.cs
```diff
@@ -7,8 +7,18 @@
...
+
+    [JsonProperty]
+    public bool NeedsVent = true;
+
+    [JsonProperty]
+    public bool NeedsTidepool = true;
```
hhyyrylainen, line 20: Also should be properties probably. Or what these are needed for? I feel like these are much more about the map generation and **should not** be stored in the actual finished map.

src/microbe_stage/PatchMap.cs
```diff
@@ -40,6 +56,12 @@ public class PatchMap
         }
     }
 
+    [JsonProperty]
+    private List<(int Id1, int Id2)> PatchesAdjacencies { get; set; } = new();
```
hhyyrylainen, line 60:
```suggestion
    private List<(int Id1, int Id2)> PatchAdjacencies { get; set; } = new();
```

src/microbe_stage/PatchMap.cs
```diff
@@ -54,6 +76,34 @@ public void AddPatch(Patch patch)
...
 
+    /// <summary>
+    ///   Adds a new region to the map. Throws if can't add
+    /// </summary>
+    public void AddRegion(PatchRegion region)
+    {
+        if (Regions.ContainsKey(region.ID))
+        {
+            throw new ArgumentException(
+                "region cannot be added to this map, the ID is already in use: " + region.ID);
```
hhyyrylainen, line 87:
```suggestion
                $"region cannot be added to this map, the ID is already in use: {region.ID}");
```

src/microbe_stage/PatchMap.cs
```diff
@@ -54,6 +76,34 @@ public void AddPatch(Patch patch)
...
+        }
+
+        Regions[region.ID] = region;
+    }
+
+    /// <summary>
+    ///   Adds a new special region to the map. Throws if can't add
```
hhyyrylainen, line 94: This should also explain what's so different what a special region is...

src/microbe_stage/PatchMap.cs
```diff
@@ -54,6 +76,34 @@ public void AddPatch(Patch patch)
...
+    /// </summary>
+    public void AddSpecialRegion(PatchRegion specialRegion)
+    {
+        if (SpecialRegions.ContainsKey(specialRegion.ID))
+        {
+            throw new ArgumentException(
+                "region cannot be added to this map, the ID is already in use: " + specialRegion.ID);
```
hhyyrylainen, line 101: Also should use string interpolation here

src/microbe_stage/PatchMap.cs
```diff
@@ -320,4 +370,127 @@ public void ReplaceSpecies(Species old, Species newSpecies)
             patch.ReplaceSpecies(old, newSpecies);
         }
     }
+
+    public void SetSeed(Random random)
```
hhyyrylainen, line 374: Setter for a **public** field is super useless as people can just do the incorrect thing and modify it directly.

src/microbe_stage/PatchMap.cs
```diff
@@ -320,4 +370,127 @@ public void ReplaceSpecies(Species old, Species newSpecies)
...
+        }
+    }
+
+    public void ConnectPatchesBetweenRegions()
+    {
+        foreach (var region in Regions)
+        {
+            region.Value.ConnectPatchesBetweenRegions(Seed);
+        }
+    }
```
hhyyrylainen, line 420: These seem like code that should be somewhere completely different. Like I said before the patch map is in my opinion a class for holding the *result* not intermediate data while building a map.

src/microbe_stage/PatchMap.cs
```diff
@@ -320,4 +370,127 @@ public void ReplaceSpecies(Species old, Species newSpecies)
...
+
+    public bool ContainsPatchAdjacency(int id1, int id2)
```
hhyyrylainen, line 422: This should have a comment / the name should be adjusted to tell that this looks for links in both directions. It's technically possible to have one direction links which is a feature I'll fight to keep possible code-wise for the future.

src/microbe_stage/PatchMap.cs
```diff
@@ -320,4 +370,127 @@ public void ReplaceSpecies(Species old, Species newSpecies)
...
+    {
+        if (RegionAdjacencies.Contains((id1, id2)) || RegionAdjacencies.Contains((id2, id1)))
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    public void CreateAdjacencies()
```
hhyyrylainen, line 442:
```suggestion
    public void CreateAdjacenciesFromPatchData()
```

src/microbe_stage/PatchMap.cs
```diff
@@ -320,4 +370,127 @@ public void ReplaceSpecies(Species old, Species newSpecies)
...
+            var patch1 = Patches[id1];
+            var patch2 = Patches[id2];
+            patch1.AddNeighbour(patch2);
+            patch2.AddNeighbour(patch1);
+        }
+
+        foreach (var (id1, id2) in RegionAdjacencies)
+        {
+            PatchRegion region1;
+            if (id1 > -1)
```
hhyyrylainen, line 476: Braces style is not followed here

src/microbe_stage/PatchMap.cs
```diff
@@ -320,4 +370,127 @@ public void ReplaceSpecies(Species old, Species newSpecies)
...
+                region1 = Regions[id1];
+            else
+                region1 = SpecialRegions[id1];
+
+            PatchRegion region2;
+            if (id2 > -1)
```
hhyyrylainen, line 482: Here as well.

src/microbe_stage/PatchMap.cs
hhyyrylainen, line 476: Why can an id be less than -1? This seems really arbitrary to me...

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -9,21 +9,37 @@ public class PatchMapNode : MarginContainer
     [Export]
     public NodePath IconPath = null!;
 
+    // Selected patch node
```
hhyyrylainen, line 12: XML comments on parts of code, please. There's a ton of these in this file so I'll only comment once.

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -75,6 +91,16 @@ public bool Marked
         }
     }
 
+    public bool SelectionAdjacent
```
hhyyrylainen, line 94: "SelectionAdjacent" doesn't tell me what this is. This should be probably renamed and explanation comment added. Is this a bool for when the patch is adjacent to the selected patch? If so this should be named "AdjacentToSelectedPatch"
also the related variables should be updated.

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -96,11 +123,23 @@ public override void _GuiInput(InputEvent @event)
                 Pressed: true, ButtonIndex: (int)ButtonList.Left or (int)ButtonList.Right,
             })
         {
+            ((PatchMapDrawer)GetParent()).MarkDirty();
```
hhyyrylainen, line 126: Explicitly referring to the parent node is an antipattern, especially casting to a special type. There should instead be emitted a signal in this case, which the parent node sets up.

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -9,21 +9,37 @@ public class PatchMapNode : MarginContainer
...
+    // currently selected node
     private bool selected;
+
+    // current player node
     private bool marked;
 
+    // node adjacent to the selected node
+    private bool selectionAdjacent;
     private Texture? patchIcon;
+    private float fadeTime = 0.33f;
```
hhyyrylainen, line 41: This seems a configuration thing, so probably should be a public export field?

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -96,11 +123,23 @@ public override void _GuiInput(InputEvent @event)
...
         {
+            ((PatchMapDrawer)GetParent()).MarkDirty();
             OnSelect();
             GetTree().SetInputAsHandled();
         }
     }
 
+    public override void _Process(float delta)
+    {
+        currentFadeTime += delta;
```
hhyyrylainen, line 134: Does this do the blinking of the currently selected patch? If so these related variables are badly named. Also put a comment explaining what these timer lines do, or put in to a separate method with a descriptive name.

src/microbe_stage/PatchRegion.cs
```diff
@@ -0,0 +1,304 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Godot;
+using Newtonsoft.Json;
+
+/// <summary>
+///   A region is a something like a continent/ocean that contains multiple biomes(patches).
```
hhyyrylainen, line 8:
```suggestion
///   A region is a something like a continent/ocean that contains multiple patches.
```

src/microbe_stage/PatchRegion.cs
```diff
@@ -0,0 +1,304 @@
...
+    public string RegionType;
+
+    [JsonProperty]
+    public float Height;
+
+    [JsonProperty]
+    public float Width;
+
+    [JsonProperty]
+    public List<Patch> Patches;
```
hhyyrylainen, line 41: I think these should be all properties (and the readonlies should have private setters).

src/microbe_stage/PatchRegion.cs
```diff
@@ -0,0 +1,304 @@
...
+    [JsonProperty]
+    public LocalizedString Name { get; private set; }
+
+    /// <summary>
+    ///   Coordinates this region is to be displayed in the GUI
+    /// </summary>
+    [JsonProperty]
+    public Vector2 ScreenCoordinates { get; set; }
+
+    public void BuildPatches(Random random)
```
hhyyrylainen, line 76: This looks like patch generation code, so should be moved to the actual patch generator.

src/microbe_stage/PatchRegion.cs
```diff
@@ -0,0 +1,304 @@
...
+                Patches[0].Depth[0] = Patches[0].Adjacent.First().Depth[0];
+                Patches[0].Depth[1] = Patches[0].Adjacent.First().Depth[1];
+            }
+        }
+    }
+
+    public void BuildRegion()
+    {
+        // Region size configuration
+        Width += 64f + 2 * PatchMargin + RegionLineWidth;
```
hhyyrylainen, line 176: I'll also comment on this once, I do not want to see any floats with just a "f" suffix. They always need to have a decimal point like "64.0f".
Once I get around to it I will write a code check that will not let any number followed by f through.

src/microbe_stage/PatchRegion.cs
```diff
@@ -0,0 +1,304 @@
...
+                    ScreenCoordinates.y + regionMargin);
+
+                // Caves or vents are the same depth as the adjacent patch
+                Patches[0].Depth[0] = Patches[0].Adjacent.First().Depth[0];
+                Patches[0].Depth[1] = Patches[0].Adjacent.First().Depth[1];
+            }
+        }
+    }
+
+    public void BuildRegion()
```
hhyyrylainen, line 173: This also seems like map generator logic.

src/microbe_stage/PatchRegion.cs
```diff
@@ -0,0 +1,304 @@
...
+                    ScreenCoordinates = rightSide;
+                }
+
+                ScreenCoordinates = new Vector2(ScreenCoordinates.x,
+                    adjacentPatch.ScreenCoordinates.y - PatchMargin - RegionLineWidth);
+            }
+        }
+    }
+
+    public void ConnectPatchesBetweenRegions(Random random)
```
hhyyrylainen, line 236: And this seems to be the final bit of generation logic.

src/microbe_stage/PatchRegion.cs
```diff
@@ -0,0 +1,304 @@
...
+    {
+        if (RegionType is "ocean" or "sea")
```
hhyyrylainen, line 238: I'll comment this here, which is not probably the best but it seems like there's a **ton** of comparing region type with strings, to me this strongly smells like the region type should be an enum.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -20,10 +21,25 @@ public class PatchMapDrawer : Control
...
+
+    [Export]
+    public Color HighlightedConnectionColor = new(0f, 1f, 1f, 1f);
+
+    [Export]
+    public float RegionLineWidth = 4f;
 
     private readonly List<PatchMapNode> nodes = new();
 
+    // The representation of connections between regions
```
hhyyrylainen, line 40: Should this be a summary?

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -81,6 +97,10 @@ public class PatchMapDrawer : Control
     public override void _Ready()
     {
         nodeScene = GD.Load<PackedScene>("res://src/microbe_stage/editor/PatchMapNode.tscn");
+        connections = new Dictionary<Tuple<int, int>, Tuple<Vector2, Vector2, Vector2>>();
```
hhyyrylainen, line 100: These tuples are too much for readability, they should be changed to be class types. For the left side Int2 might be good.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -107,27 +128,391 @@ public override void _Draw()
...
+    }
+
+    public void CenterScroll()
+    {
+        var parent = (DraggableScrollContainer)GetParent();
+        var coords = PlayerPatch!.ScreenCoordinates - parent.GetRect().End / 2f;
+        parent.ScrollHorizontal = (int)coords.x;
+        parent.ScrollVertical = (int)coords.y;
+        parent.ResetZoom();
+    }
```
hhyyrylainen, line 194: This seems 100% like functionality that belongs in DraggableScrollContainer, this scene should just emit a signal when this functionality is wanted and the parent should then react.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -107,27 +128,391 @@ public override void _Draw()
...
 
     private void DrawNodeLink(Vector2 center1, Vector2 center2)
     {
-        DrawLine(center1, center2, ConnectionColour, ConnectionLineWidth, true);
+        DrawLine(center1, center2, ConnectionColor, ConnectionLineWidth, true);
+    }
+
+    private PatchMapNode GetNode(Patch patch)
+    {
+        foreach (var node in nodes)
```
hhyyrylainen, line 214: nodes should probably be a dictionary as lookups like this are needed.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -107,27 +128,391 @@ public override void _Draw()
...
+
+        if (firstIntermediateIntersections > secondIntermediateIntersections)
+            intermediate = intermediate2;
+        else
+            intermediate = intermediate1;
+
+        return intermediate;
+    }
+
+    // This function associates a connection the points the connection line has to go through
```
hhyyrylainen, line 413: Should be a summary

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -168,6 +553,9 @@ private void UpdateNodeSelections()
         {
             node.Selected = node.Patch == selectedPatch;
             node.Marked = node.Patch == playerPatch;
+
+            if (selectedPatch != null)
+                node.SelectionAdjacent = selectedPatch.Adjacent.Contains(node.Patch!);
```
hhyyrylainen, line 558: Aren't nodes always associated with a patch, so why is the nullability on patch needed to be overridden here?

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -8,63 +10,386 @@
...
+
     public static PatchMap Generate(WorldGenerationSettings settings, Species defaultSpecies, Random? random = null)
     {
         var map = new PatchMap();
-
+        var predefinedMap = new PatchMap();
         random ??= new Random(settings.Seed);
 
+        // Return the classic map if settings require it, otherwise use it to draw the procedural map
+        predefinedMap = PredefinedMap(predefinedMap, TranslationServer.Translate("PANGONIAN"));
```
hhyyrylainen, line 33: Is there a reason this is done like this? I'd really like to have the code be so that we don't create a unnecessary patch map object when it won't be used. Do we need a general method to grab the world generation settings into the map we will end up using to get around this?

Also the translation key should be "PANGONIAN_REGION_NAME" to be clear to translators.

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -8,63 +10,386 @@
...
+        int vertexNr = random.Next(6, 10);
+        int edgeNr = random.Next(vertexNr, 2 * vertexNr - 4);
+        int minDistance = 180;
+
+        var currentPatchId = 0;
+        var specialRegionsId = -1;
+
+        // Potential starting patches, which must be set by the end of the generating process
+        Patch vents = null!;
+        Patch tidepool = null!;
```
hhyyrylainen, line 55: Seems like incorrect nullability use to me here as well.

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -8,63 +10,386 @@
...
+        }
+
         var nameGenerator = SimulationParameters.Instance.GetPatchMapNameGenerator();
 
-        string areaName = string.Empty;
-        switch (settings.MapType)
+        // Initialize the graphs random parameters
+        var regionCoords = new List<Vector2>();
+        int[,] graph = new int[100, 100];
+        int vertexNr = random.Next(6, 10);
```
hhyyrylainen, line 46:
```suggestion
        int vertexNumber = random.Next(6, 10);
```

No, non approved abbreviations please

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -8,63 +10,386 @@
...
+            {
+                for (j = 0; j < vertexNr && edgeToDelete != 0 && j <= i; j++)
+                {
+                    if (graph[i, j] == 1)
+                        edgeToDelete--;
+                }
+            }
+
+            i--;
+            j--;
```
hhyyrylainen, line 238:
```suggestion
            --i;
            --j;
```

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -8,63 +10,386 @@
...
+
+            // Check if the graph stays connected after subtracting the edge
+            // otherwise, leave the edge as is.
+            graph[i, j] = graph[j, i] = 0;
+            if (!CheckConnectivity(graph, vertexNr))
+                graph[i, j] = graph[j, i] = 1;
+            else
```
hhyyrylainen, line 245: Incorrect braces style

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -8,63 +10,386 @@
...
+        {
+            graph[triangles[i], triangles[i + 1]] = graph[triangles[i + 1], triangles[i]] = 1;
+            graph[triangles[i + 1], triangles[i + 2]] = graph[triangles[i + 2], triangles[i + 1]] = 1;
+            graph[triangles[i], triangles[i + 2]] = graph[triangles[i + 2], triangles[i]] = 1;
+        }
+
+        return graph;
+    }
+
+    // DFS graph search
```
hhyyrylainen, line 267: Bunch of stuff that should be proper summary comments

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -8,63 +10,386 @@
...
+    private static int[] Dfs(int[,] graph, int vertexNr, int point, int[] visited)
```
hhyyrylainen, line 268: Dfs is unapproved abbreviation.

hhyyrylainen (changes requested):
I got through most of the code but ran out of energy reviewing the PatchMapGenerator.cs file.

src/general/NewGameSettings.tscn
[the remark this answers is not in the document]
hhyyrylainen: Okay, that's good.

src/gui_common/DraggableScrollContainer.cs
```diff
@@ -0,0 +1,184 @@
...
+    {
+        base._Ready();
+
+        if (ContentPath == null)
+            throw new InvalidOperationException("Path to child control is expected");
+
+        content = GetNode<Control>(ContentPath);
+
+        GetVScrollbar().Connect("scrolling", this, nameof(OnScrollStarted));
+        GetHScrollbar().Connect("scrolling", this, nameof(OnScrollStarted));
```
Kasterisk, line 66: Actually it seems like `scrolling` is also never emitted, I put a debug print in `OnScrollStarted` and it never got called.

Kasterisk: Changed it to use `values_changed` instead.

Kasterisk:
Fixed zooming.

src/microbe_stage/PatchMap.cs
Kemikal1: New game settings. depending on new game settings you might need that or not.

Kemikal1:
Nice , good job guys

84634E1A607A:
I have to say that this looks not good.
image

src/microbe_stage/editor/PatchMapDrawer.cs
Kasterisk: Fixed.

src/microbe_stage/PatchMap.cs
84634E1A607A: I think Kemikal1 made `SpecialRegions`' id negative.

84634E1A607A:
Should these fields related to Godot Control use nullable or nullable supression?

`private Panel highlightPanel = null!;` or `private Panel? highlightPanel;`

Kasterisk:
Depends whether you expect it to be null or not, if you can guarantee it'll never be null, use `null!`.

src/microbe_stage/PatchMap.cs
hhyyrylainen: New game settings should be passed to the map generator, which then can look at this info. No need to save this in the *created* map.

src/microbe_stage/PatchMap.cs
hhyyrylainen: I still don't even know what's so "special" about those regions. Unless they are *really* special I'll just recommend making them normal regions with a bool in them saying they are special or a more accurate name to whatever "special" really represents in a region.

hhyyrylainen:
> Should these fields related to Godot Control use nullable or nullable suppression?

@Kasterisk is right. If you never need to check that the child control is null, then use the null suppression. If for example you need to check if the child object reference is not set yet (for example we have a bunch of various small scenes that apply changes to their child objects only after initialization) then use a nullable reference.

src/microbe_stage/PatchMapGenerator.cs
84634E1A607A: I think @Kemikal1 reused those patches generated in the predefined map. (Although I don't like this)

Kemikal1: Thats correct. If you wanted to make a certain each time look at the code for generating a default map, now duplicate it 2-3 times. almost 1000 lines of code.

Kemikal1 (commented):
I think you shouldve let this as is. I had another type of algorithm in mind for non overlaping regions that id have implemented after this PR....

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -37,7 +37,7 @@ public class PatchMapDrawer : Control
     /// <summary>
     ///   The representation of connections between regions, so we won't draw the same connection multiple times
     /// </summary>
-    private readonly Dictionary<Int2, Tuple<Vector2, Vector2, Vector2>> connections = new();
+    private readonly Dictionary<Int2, Vector2[]> connections = new();
```
Kemikal1, line 41: Here

84634E1A607A: Alright

Kemikal1: Ill leave as is now. Its not that much anyway.

84634E1A607A:
> I had another type of algorithm in mind

I look forward to it.

Kemikal1:
> > I had another type of algorithm in mind
> 
> I look forward to it.
> 
> I also have an algorithm in mind.
> 
> This one

the 1-9 connection never happens because the graph is made from a delaunay triangulation by removing its edges.
I think ive thought of your approach when making the patch map. It gets kinda complicated when you have more than 2 connection per region side and switching some connection lines might then cross other regions in the process.

My approach is to not allow obtuse angles in the delaunay triangles or just delete the connections that go over other regions.

src/microbe_stage/PatchMap.cs
hhyyrylainen: If they are just drawing subdivisions, I think it would make to sense model things with the region class so that it can have sub groupings in it. That way there doesn't need to be this extra complexity in the patch map class.

src/microbe_stage/PatchMapGenerator.cs
hhyyrylainen: What I don't get is why the generation code even needs to run for the default patches.
I still haven't looked through the code but I imagine it would make total sense to have the map property setting as a separate method that could be passed the finished classic map, which should be hard coded in terms of the regions etc. it contains, and be done with it. Basically that way there is no code duplication and no need to have multiple map objects be created.

84634E1A607A: I refactored this.

84634E1A607A:
@Kasterisk Tooltip shouldn't zoom with the container
image

84634E1A607A:
@Kemikal1 you can refer to https://github.com/Revolutionary-Games/thrive-prototypes/blob/master/patch_map/nodemap2.py

Oliveriver:
I merged master and fixed several bugs arising from conflicts between this and the patchwide population system. The only issue I couldn't fix was the sizing of the patch map in the extinction screen. As you can see, it's too big, and when the classic map is selected, some patches are cut off and I can't pan to them.





@84634E1A607A I had a look at fixing the tooltip scaling but couldn't make any progress. Unless someone can fix it quickly, I don't think it's a major enough bug to warrant more time.

Kasterisk:
I fixed the navigation tooltip, fortunately we have a custom tooltip system.

src/microbe_stage/PatchRegion.cs
Kemikal1: why?

hhyyrylainen: Because I like uniform code style. I have added the code check already so next time master is merged into this branch, it should start complaining about these things.

Kemikal1: :|
I dont know whats so uniform about adding a zero when its not needed.

src/microbe_stage/PatchMap.cs
Kemikal1: This will add complexity to other parts of the code where you need to search each region for its subgroupings.

hhyyrylainen: Using negative id values is anyway an extra complication. I really don't like that, so pick either the sub regions approach or make all ids > 0.

src/microbe_stage/PatchRegion.cs
hhyyrylainen: In C++ I would write `1.f` but that's not valid syntax. So I write `1.0f` instead. And as I've written / reviewed most of the game's code, that convention is followed (now that I added that check and fixed a few spots I missed) everywhere.

src/microbe_stage/Patch.cs
```diff
@@ -15,14 +15,16 @@ public class Patch
...
 
     [JsonProperty]
     public readonly Biome BiomeTemplate;
 
     [JsonProperty]
-    public readonly int[] Depth = new int[2] { -1, -1 };
+    public readonly int[] Depth = { -1, -1 };
+
+    [JsonProperty]
+    public PatchRegion Region;
```
hhyyrylainen, line 27:
```suggestion
    public readonly PatchRegion Region;
```

84634E1A607A:
> I couldn't fix was the sizing of the patch map in the extinction screen

MapPanel should be child of MarginContainer, and MarginContainer's custom margin should be set in line with the one below.

src/microbe_stage/PatchMap.cs
Kemikal1: I dont know how to do it with ids > 0 and the ones with the subregions im not really in favor of or gonna do.

Kemikal1:
Well it sucks then haha, i knew this for some time but then i realised that i cant do anything to make it better.

hhyyrylainen:
I suspect if someone did profiling with the longer taking auto-evo runs now, there would be 1 or 2 immediately obvious places where caching could be used. Last time I did that I think I got more than 50% performance improvement by moving one very common calculation into a cache class.

hhyyrylainen:
> The patch map nodes aren't processing (literally, _Process isn't being called), so the current patch isn't blinking and no hover highlights are shown. These things were definitely working before, so something's caused it to change.

Probably just a matter of switching one of the parent nodes for that to be "Process" processing stop mode, that should make all normally setup nodes underneath it run the _Process logic.

> For clarity, these are only problems with the patch map in the patch extinction box. The patch map in the editor is still fine (although the classic map issue may be hiding underneath, you just don't notice because the container is bigger).

Is the classic map not scrollable for some reason?

hhyyrylainen:
I tested and setting the process mode (at least in the parent extinction container), doesn't help, so the problem seems to be somewhere else.

hhyyrylainen:
Opened an issue about my dragging problems: https://github.com/Revolutionary-Games/Thrive/issues/3484

hhyyrylainen (approved):
Did as much fixing as I think I dare in this PR. I'll squash merge to get a cleaner base to fix some still present serious problems.

PR #3279, opened 2022-04-27
Add cost factor to editor for reduced organelle cost in multicellular

### LINKED ISSUE #3219: Cell editor in multicellular should have an MP discount
to make it more viable to make a lot of different cell types. Perhaps something like a 50% discount to start off with?

astral-grayson:

Adds a constant factor affecting the MP cost of all cell editor actions. In normal microbe stage this is 1, so there is no effect. In the multicelluar editor, the value is 0.5 (this is a simulation parameter constant and can be changed). Values are multiplied and cast to (int) so that a player can always perform 2 actions where they would have been able to perform one, since switching to fractional MP costs sounded annoying (and we might need to handle precision issues since floats seem to be the paradigm).

**Related Issues**

Closes #3219

**Progress Checklist**

      break existing features:
      members reviewing this PR)


astral-grayson:
Should organelle/membrane tooltips display the reduced MP cost (in the upper left corner) or the original? If it's the original, this PR is done.

hhyyrylainen:
Surely they should display the *actual* cost? The player wouldn't know what the discount is and if it is ever something like 21%, no one is going to do the math in their head.

RacerBG (changes requested):
The tooltip still displays the original cost, only the text above the organelles is updated which is not right in my opinion. On another note, maybe the organelles should be even cheaper?

hhyyrylainen:
Agreed on the tooltip. For now I think we can just go with 50% and see what the players think after a release with this PR goes out, it'll be pretty easy to tweak later.

Kasterisk (changes requested):
The MP cost in the membrane type and rigidity slider tooltips still show the original costs.

astral-grayson:
I think this is good to go now.

Kasterisk (approved):
Looks good. Thanks!

simulation_parameters/Constants.cs
```diff
@@ -564,6 +564,11 @@ public static class Constants
     /// </summary>
     public const float MULTICELLULAR_EDITOR_PREVIEW_PLACEHOLDER_SCALE = 0.18f;
 
+    /// <summary>
+    ///   Multiplier for editor actions in multicellular
```
hhyyrylainen, line 568:
```suggestion
    ///   Multiplier for cell editor actions in the multicellular editor
```

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -185,6 +185,9 @@ public partial class CellEditorComponent :
     private OrganelleDefinition nucleus = null!;
     private OrganelleDefinition bindingAgent = null!;
 
+    // Controls MP discounts (for multicelluar)
```
hhyyrylainen, line 188:
```suggestion
    /// <summary>
    ///   Controls MP discounts (for multicelluar)
    /// </summary>
```

src/microbe_stage/editor/OrganellePopupMenu.cs
```diff
@@ -30,6 +30,7 @@ public class OrganellePopupMenu : PopupPanel
     private bool enableDelete = true;
     private bool enableMove = true;
     private bool enableModify;
+    private float editorCostFactor = 1f;
```
hhyyrylainen, line 33:
```suggestion
    private float editorCostFactor = 1.0f;
```

src/gui_common/tooltip/microbe_editor/SelectionMenuToolTip.cs
```diff
@@ -48,6 +48,7 @@ public class SelectionMenuToolTip : Control, ICustomToolTip
     private string? description;
     private string processesDescription = string.Empty;
     private int mpCost;
+    private float editorCostFactor = 1f;
```
hhyyrylainen, line 51:
```suggestion
    private float editorCostFactor = 1.0f;
```

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1309,17 +1325,20 @@ private bool AddOrganelle(OrganelleTemplate organelle)
...
         // If the organelle was already moved this session, added (placed) this session,
         // or not moved (but can be rotated), then moving it is free
         bool isFreeToMove = organelle.MovedThisSession || oldLocation == newLocation || organelle.PlacedThisSession;
-        int cost = isFreeToMove ? 0 : Constants.ORGANELLE_MOVE_COST;
+        int moveCost = (int)(Constants.ORGANELLE_MOVE_COST * editorCostFactor);
+        int cost = isFreeToMove ? 0 : moveCost;
 
         // Too low mutation points, cancel move
-        if (!isFreeToMove && Editor.MutationPoints < Constants.ORGANELLE_MOVE_COST)
+        if (!isFreeToMove && Editor.MutationPoints < moveCost)
```
hhyyrylainen, line 1341: I know you are just updating existing code here, but wouldn't it make more sense to use the `cost` variable on this line? So this if would be: `if (Editor.MutationPoints < cost)` as cost is set to 0 anyway if isFreeToMove is true.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1309,17 +1325,20 @@ private bool AddOrganelle(OrganelleTemplate organelle)
     private bool MoveOrganelle(OrganelleTemplate organelle, Hex oldLocation, Hex newLocation, int oldRotation,
         int newRotation)
     {
+        // TODO: consider allowing rotation inplace
```
hhyyrylainen, line 1328: Could you link to the relevant github issue here?

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -509,6 +516,10 @@ public override void Init(ICellEditorData owningEditor, bool fresh)
             behaviourTabButton.Visible = false;
             behaviourEditor.Visible = false;
         }
+
+        // after the if multicellular check so it resets if this is a single-cell editor
```
hhyyrylainen, line 520:
```suggestion
        // After the if multicellular check so it resets if this is a single-cell editor
```

src/microbe_stage/editor/CellEditorComponent.cs
Kasterisk: Hint, typo in "multicelluar".

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -509,6 +516,10 @@ public override void Init(ICellEditorData owningEditor, bool fresh)
...
+        // but the previously opened editor was multicell (ie save and switch to new file)
+        UpdateTooltipMPCostFactors();
```
hhyyrylainen, line 522: So this is also required to run for the multicellular editor? If not, I'd place this in an else. If this is required to run anyway, I'd change the above comment to be something like "Make sure we have up to date cost factors set on tooltips" to emphasize that fact.

hhyyrylainen (changes requested):
I think the overall approach is fine and that was already approved by other people. I have just some code cleanup related concerns to improve consistency with existing code and clarity.

RacerBG (approved):
Everything looks fine now.

src/microbe_stage/editor/CellEditorComponent.cs
astral-grayson: The reason I did it this way (and also the first change you requested with the extra variable mpDisplay) was because otherwise the reduced costs were persistent on exiting to menu and opening a new or saved game.

I don't think tooltip objects are garbage collected when exiting the editor and re-created, they're still in the tooltip manager when the main gameplay is running. I didn't want to make major changes to the whole tooltip system in this PR so this was my solution - as well as the separate mpDisplay variable because I wanted the original cost to only ever be updated from the JSON so opening the wrong sequence of editors couldn't lead to a state where the MP cost in tooltips was incorrect (more of a future proofing than anything).

Let me know what your thoughts are!

src/gui_common/tooltip/microbe_editor/SelectionMenuToolTip.cs
```diff
@@ -291,7 +303,8 @@ private void UpdateMpCost()
         if (mpLabel == null)
             return;
 
-        mpLabel.Text = MutationPointCost.ToString(CultureInfo.CurrentCulture);
+        int mpDisplay = (int)(mpCost * editorCostFactor);
```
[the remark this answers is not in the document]
astral-grayson: See response to final review comment.

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: I meant in the source code. You can either link in this TODO to the issue or you can remove this comment that was added in this PR.

src/gui_common/tooltip/microbe_editor/SelectionMenuToolTip.cs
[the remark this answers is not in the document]
hhyyrylainen: This is just literally a local variable... and has no effect if it was inlined. Let me know if I'm majorly misreading the code here.

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: Yes, the tooltips are in a separate manager object, I'm not criticising the need to update the tooltips always here. I'm suggesting a simpler explanation comment. Mainly because to me this current wording is misleading regarding how the editor object is handled. The editor object tree is in fact deleted each time the editor is exited so that can't have invalid state, but as I said the tooltip manager owns the tooltips which persist throughout the game.

hhyyrylainen (approved):
The comment wording is still not perfectly clear to me, but let's go with this for now.

Issue #3274, opened 2022-04-26
Check that old species spawner is removed when species becomes multicellular


hhyyrylainen:
it's starting to seem like that the old player species still keeps spawning cells even after converting the species. I think this might be a bug with the spawner related to the old species not getting cleared and a new spawner not being added, so I think there maybe needs to be a call to update the spawners in the current patch when the player enters the multicellular stage.

PR #3280, opened 2022-04-28
Prevent cannibalism on transition to multicell

### LINKED ISSUE #3155: Prevent unintentional cannibalism when species type changes to multicellular
cannibalism is possible while there exists cells of the player's previous species as the microbe species instance is not equal to the early multicellular species instance

astral-grayson:

On changing to early multicellular mode, generate an array of all existing microbes of the player's species, and apply the new multicellular species to them. This prevents cannibalism and the immediate death of the player to their former colony.

**Related Issues**

Closes #3155

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I'm unsure about this approach... this does seem pretty elegant way to get around the species objects not comparing equally. I was thinking that I would fix this issue myself by overloading .Equals for the Species types to just compare IDs, only caveat with that is that quite many comparisons need to be found and updated, so this approach gets around that.
I was going to comment my worry that converting the existing cells to multicellular will allow them to start growing colonies, but maybe that isn't the case as the species only has a single cell when applied? But then again when exiting the editor the species referenced by the cells will have additional cells in the layout.

Anyone else have thoughts on this? I'm sort of leaning towards accepting this approach as the one needing less overall code changes.

astral-grayson:
When you exit the editor, each microbe is in the basic single cell bud state. I feel like having the single cell and multicell versions evaluate equal would make less sense over time as the player species diverges from the single celled form with successive evolutions. Maybe we should be explicit about the difference and turn the old player species into a branch with a different name? I also entertained the idea of just despawning some or all of the nearby player microbes, if we don't want that many copies of the player's species around.

hhyyrylainen:
>  sense over time as the player species diverges from the single celled form with successive evolutions.

The single celled version of the player species is removed immediately when the player becomes multicellular. The earlier species hanging around is a bug.

Related bug that I freshly opened when noticing that the old species still kept spawning (which is not intended): https://github.com/Revolutionary-Games/Thrive/issues/3274

src/microbe_stage/Microbe.cs
```diff
@@ -395,6 +395,9 @@ public void ApplySpecies(Species species)
             MulticellularCellType = earlyMulticellularSpecies.Cells[0].CellType;
 
             cachedMulticellularSpecies = earlyMulticellularSpecies;
+
+            if (IsPlayerMicrobe)
+                GD.Print("Applying multicellular species type to player");
```
hhyyrylainen, line 400: This print is not necessary. If you want to confirm that the player also gets applied the new species add a check in MoveToMulticellular that prints an error if the player wasn't processed.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -716,6 +720,24 @@ public void OnFinishTransitioning()
             TutorialEventType.EnteredMicrobeStage, new CallbackEventArgs(HUD.PopupPatchInfo), this);
     }
 
+    /// <summary>
+    ///   Helper function for transition to multicellular
+    ///   Returns array of all microbes of Player's species
```
hhyyrylainen, line 725: This line could be a separate `<returns>` xml block

src/microbe_stage/MicrobeStage.cs
```diff
@@ -716,6 +720,24 @@ public void OnFinishTransitioning()
...
+    private Microbe[] GetAllPlayerSpeciesMicrobes()
+    {
+        if (Player == null)
+        {
+            GD.PrintErr("Could not get player species microbes: no Player node");
+            return new Microbe[0];
+        }
+
+        var microbes = rootOfDynamicallySpawned.GetTree().GetNodesInGroup(Constants.AI_TAG_MICROBE).Cast<Microbe>()
+            .ToArray();
```
hhyyrylainen, line 736: This ToArray is also completely unnecessary, you can just chain LINQ as much as you want (well within reason, we do have a styleguide which doesn't recommend infinite LINQ).

hhyyrylainen (approved):
Did the refactoring I wanted so that I can merge this in time for 0.5.8.1

astral-grayson:
Cool, thanks :) I spent the day on my coursework instead, haha.

Issue #3174, opened 2022-04-09
Resolve the underlying issue with disposed microbes in engulf check


hhyyrylainen:
I added a guard against crashing here: 
https://github.com/Revolutionary-Games/Thrive/blob/55b3d8361b5d5669fdb8dda90089cbba5a35eadb/src/microbe_stage/Microbe.Contact.cs#L357

We should actually get to the underlying issue and solve that here, seems to be easy to trigger by trying to engulf cells of your species while they are dying.

hhyyrylainen:
This print happens a *lot*, so we might need to remove it to not spam so much. Optimally a fix would be found first before we have to resort to that...

PR #3292, opened 2022-04-30
Added custom Disposed field

84634E1A607A:
So that we can check for a bool instead of asking Godot for an unused property every time.

**Related Issues**
Dealing with #3174 - not fully resolved, but now it no longer spams.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I thought I had already made a change regarding this, but apparently I forgot: https://github.com/Revolutionary-Games/Thrive/pull/3293 I'll just merge that immediately now.

84634E1A607A:
But I think this PR makes it unnecessary...

hhyyrylainen:
Maybe.

This needs a ton of testing to ensure that there's no crash potential here anymore.

Also if a cell is despawned it will be disposed *but not dead* so if we have the underlying bug that the touched microbes is not updated when a microbe is disposed, this will make the game be able to crash if a cell is despawned and something nearby it is trying to engulf it.

84634E1A607A:
Then what about make it dead then?

hhyyrylainen:
If a microbe is killed (takes enough damage to have 0 health), then it is dead, and will be disposed and removed from the world after the death animation plays.

84634E1A607A:
I mean what about make despawned microbes "dead"?

hhyyrylainen:
As far as I remember, despawned microbes will be just removed. They are not marked dead. So this change to no longer check if something is disposed *may* cause crashes regarding despawned microbes because they would not be marked as dead.

84634E1A607A:
We can just mark them dead, or we can add a custom disposed bool to track its state instead of calling Godot property every check.

hhyyrylainen:
I don't like the idea of pretending that despawned things are made dead. We are anyway working around Godot physics callback bug here so intentionally causing the disposed exception is the best thing for that I think, because we are working around a bug.
Adding the dead check as a separate thing first would be fine though to avoid cases where something that is dead is attempted to be engulfed which should be disallowed anyway.

hhyyrylainen (changes requested):
I don't like this way to check if something is disposed. Godot already does that. Detecting if something is disposed is an anti pattern as far as I've read about it in C#.

hhyyrylainen:
Could you make a PR *just* for adding the cell is not dead check in the engulf starting method? I'd accept that easily.

84634E1A607A:
> Could you make a PR _just_ for adding the cell is not dead check in the engulf starting method? I'd accept that easily.

Let me explain this:

When engulfing is coded, the code has ensured that all dead cells won't contact with living ones, so dead cells should never be checked if it can be engulfed. This one is due to some Godot problem or some strange problem, so I think we should we should solve the problem from the source, that is, to ensure that the list doesn't contain any dead cell.

> Detecting if something is disposed is an antipattern

This is equal to adding a bool indicating the cell has been destroyed. So I really don't think this is somehow troblesome.

hhyyrylainen:
> This one is due to some Godot problem or some strange problem, so I think we should we should solve the problem from the source

The source is a Godot engine issue. If you want to fix the source of the issue the engine needs to be fixed.

> This is equal to adding a bool indicating the cell has been destroyed. So I really don't think this is somehow troublesome.

Then check if the cell is destroyed in the engulfing code???

hhyyrylainen (changes requested):
Due to performance concerns I'd instead edit the existing loop rather than add a new one, something like this:

```
        foreach (var microbe in touchedMicrobes)
        {
           if(microbe.destroyed){
            touchedMicrobes.Remove(disposedMicrobe);
            GD.Print($"Removed destroyed microbe from {nameof(touchedMicrobes)}");
            touchedMicrobes.Remove(microbe);
            break; // this is important to not crash on the next loop in the foreach
            }
            if (!attemptingToEngulf.Contains(microbe) && CanEngulf(microbe))
            {
                StartEngulfingTarget(microbe);
                attemptingToEngulf.Add(microbe);
            }
        }
```

84634E1A607A:
No, no, no. You remember the issue I said which arises due to local debugging changes? If you place it inside the loop, you modify the hashset inside enumerator, which causes a modified exception. This is also the reason I added a ToList after the Linq expression.

hhyyrylainen:
Please read my suggested code again, especially `break; // this is important to not crash on the next loop in the foreach` which I'm 99% sure work correctly in C#, ie. the loop is exited before it is looped again avoiding the problem. This has the tiny drawback that only one destroyed microbe is removed per check but I think that the engulf check is ran on each frame for each microbe anyway so at most it causes a couple of frame delay which should be perfectly safe to ignore.

84634E1A607A:
I see

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -1304,6 +1304,13 @@ private void CheckStartEngulfingOnCandidates()
         // foreach (var microbe in touchedMicrobes.Concat(otherMicrobesInEngulfRange))
         foreach (var microbe in touchedMicrobes)
         {
+            if (microbe.destroyed)
+            {
+                GD.Print($"Removed destroyed microbe from {nameof(touchedMicrobes)}");
+                touchedMicrobes.Remove(microbe);
+                break; // this is important to not crash on the next loop in the foreach
```
hhyyrylainen, line 1311: I didn't mean to literally include my explanation comment like this. If you want to keep this please put it on a separate line.

hhyyrylainen (changes requested):
One slight style problem.

hhyyrylainen (approved):

84634E1A607A:
I wonder if we can remove the disposed checks now that no disposed microbe can be accessed.

hhyyrylainen:
Just to be safe I would not remove it... we'd need a ton of testing to make sure the game hasn't become unstable... Maybe we need a system to let players report any exceptions or error prints they experience? That way we could actually confidently remove some old disposed checks in future releases.

PR #3295, opened 2022-05-01
Debug Overlay

### LINKED ISSUE #3260: Add button to toggle physics debug and entity labels
inspired by the discussion here: https://forum.revolutionarygamesstudio.com/t/disposed-microbe-problem/878/5

Should probably be created as an extra autoload that owns the labels to be drawn on screen on above the entities, and toggled with CTRL+F3 by default.

84634E1A607A:

Added a debug panel and rearranged previous debugging tools.

**Related Issues**
Closes #3260 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/Microbe.cs
```diff
@@ -371,6 +377,13 @@ public override void _Ready()
             SetMembraneFromSpecies();
         }
 
+        DebugPanel.Instance.Connect(nameof(DebugPanel.ToggleEntityLabel), this, nameof(OnToggleEntityLabel));
+        nameLabel = new Label();
+        AddChild(nameLabel);
+        SetInternalName();
+        nameLabel.Text = internalName;
+        nameLabel.Visible = DebugPanel.ShowEntityLabel;
```
Kasterisk, line 385: I prefer handling these in a single autoloaded class so that other entities like floating chunks, etc can also easily have labels on them. Doing these separately here for each entities however can lead to a hard to maintain code.

Kasterisk (changes requested):
I can't get the panel to open without rebinding the toggle key.

src/microbe_stage/Microbe.cs
hhyyrylainen: I agree with this.

84634E1A607A:
Okay. So you are suggesting I connect the node_added signal to debug panel so that we can filter all nodes and add custom working logic in one separate class? I'll try later.

84634E1A607A:
My current concern is that whether we may have more than one camera in tree. Currently I just hard coded Microbe camera's node path in it; but if we have more than one camera, how can we know which camera to use in tree to unproject the entity? Or if we have more than one camera pointing to one entity (e.g. different sides of view of an organelle), how can we know where to add the label?

Kasterisk:
One way of knowing it is probably to get the current active 3D camera from viewport using [`Viewport.get_camera()`](https://docs.godotengine.org/en/stable/classes/class_viewport.html#class-viewport-method-get-camera). Regarding your concern, I doubt we'll ever going to use multiple active cameras and even if so it'll be rare in normal gameplay.

84634E1A607A:
Another thing. How do we detect active camera change other than calling get_camera() every frame? I don't find a signal to inform me, and I really don't get it how to search for it on bing or google. All results are irrelevant.

hhyyrylainen:
For the damage numbers mod, I just check that the active camera entity is valid and exists and if it is changed from previous frame process based on that. I think that'll be a fine approach for the debug panel as it's kind of a tacked on part of the game.

84634E1A607A:
> I can't get the panel to open without rebinding the toggle key.

The current key binding for debug panel is F3 (previously for FPS display), so you need to restore it to defaults to use the updated key binding.

hhyyrylainen:
The debug panel should default to CTRL+F3 so that F3 can be the basic FPS display.

src/engine/debugging/DebugOverlay.cs
```diff
@@ -0,0 +1,243 @@
...
+        base._Ready();
+    }
+
+    public override void _EnterTree()
+    {
+        base._EnterTree();
+
+        Show();
+        InputManager.RegisterReceiver(this);
+        rootTree = GetTree();
```
Kasterisk, line 69: I think it's fine to just directly call `GetTree()` without caching it.

src/engine/debugging/DebugOverlay.tscn
```diff
@@ -0,0 +1,95 @@
...
+margin_bottom = 80.0
+custom_fonts/font = ExtResource( 4 )
+text = "ENTITY_LABEL"
+
+[node name="VBoxContainer" type="HBoxContainer" parent="DebugPanel/VBoxContainer"]
+margin_top = 84.0
+margin_right = 191.0
+margin_bottom = 101.0
+
+[node name="RigidityLabel" type="Label" parent="DebugPanel/VBoxContainer/VBoxContainer"]
```
Kasterisk, line 74: This shouldn't be named rigidity.

src/microbe_stage/FloatingChunk.cs
```diff
@@ -51,6 +51,8 @@ public class FloatingChunk : RigidBody, ISpawned, ISaveLoadedTracked
     [JsonProperty]
     private float elapsedSinceProcess;
 
+    private string debugName = null!;
```
Kasterisk, line 54: I don't see why this must be its own instance variable.

src/microbe_stage/FloatingChunk.cs
```diff
@@ -225,6 +229,11 @@ public override void _Ready()
         InitPhysics();
     }
 
+    public override string ToString()
+    {
+        return debugName;
+    }
```
Kasterisk, line 235: Perhaps we can just directly return the formatted debug name here?

src/microbe_stage/Microbe.cs
```diff
@@ -1015,4 +1026,15 @@ private void CountHexes()
...
+    private void SetDebugName()
+    {
+        if (Species == null!)
+        {
+            debugName = string.Empty;
+            return;
+        }
+
+        debugName = $"[{Species.Genus[0]}.{Species.Epithet.Substring(0, 4)}:{GetInstanceId()}]";
+    }
```
Kasterisk, line 1039: Again, the same here. Also if the species is null I think it's better to just return the debug name without it, so it'd just be the instance id.

locale/en.po
```diff
@@ -1042,21 +1046,46 @@ msgstr "run status:"
...
+msgid "FPS_DISPLAY"
+msgstr "FPS display"
+
+#: ../src/engine/debugging/DebugOverlay.tscn:53
+msgid "PERFORMANCE_METRICS"
+msgstr "Performance Metrics"
+
+#: ../src/engine/debugging/DebugOverlay.tscn:60
+msgid "COLLISION_SHAPE"
+msgstr "Collision shapes"
```
Kasterisk, line 1067: As it turn out that the collision shapes only show up after entity is spawned, I think it makes more sense to reword this to  "Spawn entity with collision shapes".

Kasterisk (changes requested):

.gitignore
```diff
@@ -67,6 +67,7 @@ _ReSharper*/
 $tf*/
 [Bb]in
 [Dd]ebug*/
+!src/engine/debugging
```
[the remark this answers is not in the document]
84634E1A607A: Er, Just at first the debugging folder is ignored by the previous line.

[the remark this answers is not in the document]
Kasterisk: I see...

src/engine/debugging/DebugOverlay.cs
```diff
@@ -0,0 +1,243 @@
+﻿using System.Collections.Generic;
+using Godot;
+
+public class DebugOverlay : Control
+{
+    private readonly Dictionary<Microbe, Label> microbeLabels = new();
+    private readonly Dictionary<FloatingChunk, Label> floatingChunkLabels = new();
```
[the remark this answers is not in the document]
84634E1A607A: I'm having another one added (which isn't IEntity)

src/microbe_stage/Microbe.cs
84634E1A607A: Then I'll need more work to check if the species is properly applied and more runtime cost.

src/engine/debugging/DebugOverlay.cs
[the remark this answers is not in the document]
Kasterisk: Which one? Maybe it'd make sense to also make it inherit IEntity since this is entity labels after all.

src/microbe_stage/FloatingChunk.cs
84634E1A607A: that's true.

src/engine/debugging/DebugOverlay.cs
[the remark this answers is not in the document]
84634E1A607A: Agent Projectile

.gitignore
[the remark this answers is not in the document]
hhyyrylainen: I still wouldn't like this to be added to git ignore...

src/engine/debugging/DebugOverlay.cs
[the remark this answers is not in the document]
hhyyrylainen: I think agent projectile would make sense to also be an IEntity

src/microbe_stage/Microbe.cs
Kasterisk: I think it won't performance too much, should be very cheap to warrant to be run each frame. On a related note would nullables work in a string? Like I think it's more efficient to write `$"[Species?.name] ..."`.

src/engine/debugging/DebugOverlay.cs
hhyyrylainen: Yeah, if it is called just once per frame, I would not see that as a problem.

src/microbe_stage/FloatingChunk.cs
hhyyrylainen: I would much prefer to not need to have special code in each entity related to this feature.

src/microbe_stage/FloatingChunk.cs
84634E1A607A: chunktype.Name is not stored inside FloatingChunk class, so I can't get the name if elsewhere.

Kasterisk: Add a new Name property to FloatingChunk and assign it with `chunkType.Name` in the Init. Not sure why it's missing in the first place...

84634E1A607A: Maybe because Node has a Name property

hhyyrylainen: I guess there's never been a need for chunks to store that information so I never added such a field. One thing I wonder though is that there's that method that makes a chunk configuration from an existing chunk somewhere and I think that copies over the name somehow.

84634E1A607A: Name is not included.

Kasterisk: > Maybe because Node has a Name property

Ah yeah that's true, then this should have more explicit naming to `ChunkName`.

> Name is not included.

Then there is high possibility that the Name is not copied over in that method.

src/engine/debugging/PerformanceMetrics.cs
```diff
@@ -41,6 +41,9 @@ private PerformanceMetrics()
         instance = this;
     }
 
+    [Signal]
+    public delegate void OnHidden();
```
Kasterisk, line 45: I just remembered that popups have `popup_hide` signal, so that should be used instead of adding a new one unless there's a reason why a separate is needed.

src/microbe_stage/FloatingChunk.cs
```diff
@@ -106,6 +106,8 @@ public class FloatingChunk : RigidBody, ISpawned, ISaveLoadedTracked
     /// </summary>
     public string DamageType { get; set; } = "chunk";
 
+    public new string Name { get; set; } = string.Empty;
```
Kasterisk, line 109: Shouldn't be overriden, instead use a more explicit name like `ChunkName`.

src/microbe_stage/Microbe.cs
```diff
@@ -763,6 +763,13 @@ public override void _IntegrateForces(PhysicsDirectBodyState physicsState)
         }
     }
 
+    public override string ToString()
+    {
+        return Species != null! ?
+            $"[{Species.Genus[0]}.{Species.Epithet.Substring(0, 4)}:{GetInstanceId()}]" :
+            string.Empty;
```
Kasterisk, line 770: When null, only return the instance id and for good measure also add the node Name to both strings.

Kasterisk (changes requested):

src/microbe_stage/FloatingChunk.cs
```diff
@@ -225,6 +229,11 @@ public override void _Ready()
         InitPhysics();
     }
 
+    public override string ToString()
+    {
+        return $"[{Name}:{GetInstanceId()}]";
```
Kasterisk, line 234: Including the actual Node name could also be good here.

Kasterisk (changes requested):

src/microbe_stage/FloatingChunk.cs
84634E1A607A: PerformanceMetrics is a Control

src/microbe_stage/Microbe.cs
84634E1A607A: Name is too long and of little use.

src/engine/debugging/PerformanceMetrics.cs
84634E1A607A: PerformanceMetrics is a Control...

src/microbe_stage/Microbe.cs
Kasterisk: I can see myself comparing the nodes in the remote scene tree with the ones in the game screen while debugging, so this could be of some use.

src/engine/debugging/PerformanceMetrics.cs
Kasterisk: Ah yeah I missed that.

src/microbe_stage/Microbe.cs
84634E1A607A: > in the remote scene

It looks like I've missed some important debugging tools. What is that? I'll change InstanceId to that later.

hhyyrylainen: When you run Thrive from Godot, in the left panel in the Godot editor you can swap between the local and remote scene trees. When you switch to the remote view you can see all the Nodes the game has created and even edit them a bit in realtime. I've used it myself a few times and it's been pretty handy to see how things actually are in the game to confirm where I've made a mistake.

84634E1A607A:
To find debugging tools more easily, I merged all tools into class DebugOverlay and rearranged signals.

src/engine/DebugOverlay.EntityLabel.cs
```diff
@@ -0,0 +1,158 @@
...
+    private void EntityLabelProcess()
+    {
+        if (activeCamera is not { Current: true })
+            activeCamera = GetViewport().GetCamera();
+
+        if (showEntityLabels)
+        {
+            UpdateEntityLabels();
+        }
+    }
```
Kasterisk, line 54: I think these can just be in the main class without the need for separate methods. The microbe class for example already handles all its init for its partial classes in the main class.

Kasterisk (changes requested):

src/engine/DebugOverlay.EntityLabel.cs
84634E1A607A: I think it looks better if I can find all relevant logics in a single file. Being a simple function, the compiler may as well merge this into the main one.

src/engine/DebugOverlay.EntityLabel.cs
```diff
@@ -0,0 +1,158 @@
...
+                microbe.OnDeath += UpdateLabelOnMicrobeDeath;
+                break;
+            }
+
+            case FloatingChunk:
+            case AgentProjectile:
+            {
+                label.AddFontOverride("font", smallerFont);
+                break;
+            }
```
Kasterisk, line 128: Should simply use `default` keyword as it doesn't matter whether this is FloatingChunk or AgentProjectile. Or you could even use this font for all entity types, not sure why Microbe is excluded.

src/engine/DebugOverlay.EntityLabel.cs
Kasterisk: It looks more tangled to me having these in the partial classes. In my view, only the distinct logics should be in the partial classes. EnterTree, ExitTree, Ready or Process logics are not distinct and is associated with the DebugOverlay node itself should be kept in the main DebugOverlay class where it makes sense.

src/engine/DebugOverlay.cs
```diff
@@ -0,0 +1,145 @@
+﻿using Godot;
+
+/// <summary>
+///   Main script for debugging.
+///   Partial class: Debug panel
+/// </summary>
+public partial class DebugOverlay : Control
```
Kasterisk, line 7: Since this contain many overlays, should be named `DebugOverlays`.

src/microbe_stage/AgentProjectile.cs
```diff
@@ -16,6 +16,10 @@ public class AgentProjectile : RigidBody, ITimedLife
     public AgentProperties? Properties { get; set; }
     public EntityReference<IEntity> Emitter { get; set; } = new();
 
+    public Node EntityNode { get; private set; } = null!;
```
Kasterisk, line 19: Use lambda expression, `public Node EntityNode => this;`.

Kasterisk (changes requested):

src/engine/DebugOverlay.EntityLabel.cs
84634E1A607A: It's just because floating chunks always overlap each other, making it very difficult to read. Microbes are separated.

src/engine/DebugOverlay.EntityLabel.cs
```diff
@@ -0,0 +1,158 @@
+﻿using System.Collections.Generic;
+using Godot;
+
+/// <summary>
+///   Partial class: Entity label
+/// </summary>
+public partial class DebugOverlay
+{
+    private readonly Dictionary<RigidBody, Label> entityLabels = new();
```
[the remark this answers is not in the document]
hhyyrylainen: Should this use maybe EntityReference as well to guard against destroyed entities?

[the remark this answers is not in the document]
84634E1A607A: Not necessary. All nodes in this dict are still in tree (Unless Godot have a strange bug not triggering node_removed signal when a node is removed).

src/engine/DebugOverlay.EntityLabel.cs
```diff
@@ -0,0 +1,158 @@
...
+    private void UpdateLabelOnMicrobeDeath(Microbe microbe)
+    {
+        if (entityLabels.TryGetValue(microbe, out var label))
+            label.Set("custom_colors/font_color", new Color(1.0f, 0.3f, 0.3f));
+    }
+
+    private void OnNodeAdded(Node node)
+    {
+        if (node is not RigidBody body)
+            return;
```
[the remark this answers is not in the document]
84634E1A607A: Will an entity be anything other than Spatial? Node doesn't have GlobalTransform or Transform. If not, I'll make IEntity.EntityNode Spatial so that I can access the Transform property without a cast.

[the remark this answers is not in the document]
hhyyrylainen: I think it's probably safe to assume any entities will be at least a Spatial.

src/engine/DebugOverlays.cs
```diff
@@ -0,0 +1,213 @@
...
+            while (despawnHistory.Count > SpawnHistoryLength)
+                despawnHistory.RemoveFromFront();
+
+            currentSpawned = 0;
+            currentDespawned = 0;
+        }
+
+        // FPS counter
+        if (fpsCounter.Visible)
+            fpsDisplayLabel.Text = new LocalizedString("FPS", Engine.GetFramesPerSecond()).ToString();
```
Kasterisk, line 151: So does this.

src/engine/DebugOverlays.cs
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    [RunOnKeyDown("toggle_metrics", OnlyUnhandled = false)]
+    public void OnPerformanceMetricsToggled()
```
Kasterisk, line 155: Is it possible to move overlay-specific callbacks to the partial classes?

src/microbe_stage/AgentProjectile.cs
```diff
@@ -47,7 +51,12 @@ public override void _Process(float delta)
 
         FadeTimeRemaining -= delta;
         if (FadeTimeRemaining <= 0)
-            Destroy();
+            OnDestroyed();
+    }
+
+    public void OnDestroyed()
+    {
+        this.DetachAndQueueFree();
```
Kasterisk, line 59: Shouldn't this be setting AliveMarker to false instead? I think this could possibly conflict with `DestroyDetachAndQueueFree`.

src/microbe_stage/AgentProjectile.cs
```diff
@@ -47,7 +51,12 @@ public override void _Process(float delta)
 
         FadeTimeRemaining -= delta;
         if (FadeTimeRemaining <= 0)
-            Destroy();
+            OnDestroyed();
```
Kasterisk, line 54: This can simply call `this.DestroyDetachAndQueueFree`.

Kasterisk (changes requested):

src/microbe_stage/AgentProjectile.cs
hhyyrylainen: It's very wrong to call DetachAndQueueFree in a OnDestroyed method.

src/engine/DebugOverlays.cs
84634E1A607A: This function is actually DebugPanel (the dialog) callback, which is in this file.

Kasterisk: Oh right, I failed to notice that. This is good to go then.

Kasterisk (commented):
Looks good and works good but is there any way for the entity labels to not get drawn in front of the pause menu?

84634E1A607A:
I don't think there is. (however, I am curious why a pop-up should have been placed below a control by Godot.)

84634E1A607A:
Plus I'm wondering if I should make transparency work for EntityLabel to.

Kasterisk:
> however, I am curious why a pop-up should have been placed below a control by Godot.

I just remembered why. This is because a CanvasLayer is used for the screen overlays, its draw order is set so high that it's on top of everything so it doesn't matter if a popup or else is on screen.

> Plus I'm wondering if I should make transparency work for EntityLabel to.

I don't think it's worth the effort.

Kasterisk (approved):
In any case this is now looks good to me.

src/engine/DebugOverlays.cs
```diff
@@ -0,0 +1,164 @@
...
+    {
+        base._EnterTree();
+
+        Show();
+        InputManager.RegisterReceiver(this);
+
+        // Entity label
+        var rootTree = GetTree();
+        rootTree.Connect("node_added", this, nameof(OnNodeAdded));
+        rootTree.Connect("node_removed", this, nameof(OnNodeRemoved));
```
hhyyrylainen, line 53: This should really only be connected while debugging is active. I do not want the debugging features to use *any* performance when not in use.

src/engine/DebugOverlays.cs
```diff
@@ -0,0 +1,164 @@
...
+    public static DebugOverlays Instance => instance ?? throw new InstanceNotLoadedYetException();
+
+    public override void _EnterTree()
+    {
+        base._EnterTree();
+
+        Show();
+        InputManager.RegisterReceiver(this);
+
+        // Entity label
```
hhyyrylainen, line 50: This comment doesn't tell me anything...

src/engine/DebugOverlays.cs
```diff
@@ -0,0 +1,164 @@
...
+    public override void _ExitTree()
+    {
+        InputManager.UnregisterReceiver(this);
+
+        // Entity label
+        var rootTree = GetTree();
+        rootTree.Disconnect("node_added", this, nameof(OnNodeAdded));
+        rootTree.Disconnect("node_removed", this, nameof(OnNodeRemoved));
+
+        base._ExitTree();
```
hhyyrylainen, line 65: Base method calls need to be at the start of a method unless specifically that causes a problem.

hhyyrylainen (changes requested):
This looks and works mostly fine to me, but I'd like a little bit of code tweaking done before merge.

locale/en.po
```diff
@@ -1097,10 +1097,35 @@ msgstr ""
...
+msgid "DEBUG_PANEL"
+msgstr "Debug Panel"
+
+#: ../src/engine/DebugOverlays.tscn:96
+msgid "FPS_DISPLAY"
+msgstr "FPS display"
+
+#: ../src/engine/DebugOverlays.tscn:110
+msgid "COLLISION_SHAPE"
+msgstr "Spawn entity with collision shapes"
```
hhyyrylainen, line 1115:
```suggestion
msgstr "Spawn entities with collision shapes"
```

src/engine/DebugOverlays.cs
84634E1A607A: But this means we need to know all the entities when entity label is turned on, which requires much time.

hhyyrylainen: In my opinion that's better than requiring all the tens of thousands of players to have a slower experience who never turn on the debug overlay.

84634E1A607A:
Fortunately, it doesn't affect performance too much.

hhyyrylainen (approved):

hhyyrylainen:
I'll fix the translation merge conflicts and then merge this PR.

84634E1A607A:
My current concern is that whether we may have more than one camera in tree. Currently I just hard coded Microbe camera's node path in it; but if we have more than one camera, how can we know which camera to use in tree to unproject the entity? Or if we have more than one camera pointing to one entity (e.g. different sides of view of an organelle), how can we know where to add the label?

PR #3299, opened 2022-05-03
Fade pause menu background on opening and closing

Kasterisk:

Pause menu background now fades in and out. Cleaned up the pause menu's open and close logic to reduce confusion.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/PauseMenu.tscn
+++ b/src/general/PauseMenu.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=8 format=2]
+[gd_scene load_steps=11 format=2]
 
 [ext_resource path="res://src/general/PauseMenu.cs" type="Script" id=1]
 [ext_resource path="res://src/gui_common/thrive_theme.tres" type="Theme" id=2]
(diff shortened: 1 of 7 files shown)


hhyyrylainen:
Doesn't this need changes to the early multicellular editor scene as that also contains the pause menu in it?

Kasterisk:
Fixed two signal connections relating to menu opening and closing in there.

Kasterisk:
No idea why the build failed, never got those particular warnings before.

hhyyrylainen:
Probably a newer version of the checking tools is used for the first time on a file you modified, so it now finds more things to request fixes for.

Kasterisk:
What should be done? I don't think we should manually dispose Godot objects and it's just weird that this warning only occurs in this file and not the others. I think I'll probably just suppress it.

hhyyrylainen:
I guess for some reason the checks detect that the type is ultimately derived from IDisposable? But then again I*m not sure why it isn't triggered for other classes as we have a bunch deriving from Godot types. I also thought it might be caused by the class inheriting something where we have written a Dispose method but that doesn't seem to be the case either. So I guess ignoring the warning is the right way to go here...

src/general/PauseMenu.tscn
```diff
@@ -8,8 +8,128 @@
...
+tracks/2/keys = {
+"times": PoolRealArray( 0 ),
+"transitions": PoolRealArray( 1 ),
+"update": 0,
+"values": [ true ]
+}
+
 [node name="PauseMenu" type="Popup"]
 pause_mode = 2
+visible = true
```
hhyyrylainen, line 132: I think this might be a mistake...

Kasterisk: That's weird, it shows me that the node is hidden in the Godot editor however... I might have to remove this manually.

hhyyrylainen (approved):
Looks fine to me. I noticed one problem but it was probably a problem also with the previous code, which is that if you press the pause button in the microbe stage and then open the pause menu, closing the pause menu then unpauses the game incorrectly.

hhyyrylainen:
Opened the issue: https://github.com/Revolutionary-Games/Thrive/issues/3319

@Kasterisk if you could take a look at that, that would be great as it was maybe caused by this PR.

Kasterisk:
It's indeed a regression from my previous PR, this should be the fix: #3324.

PR #3303, opened 2022-05-03
Improved the frame delta translation

hhyyrylainen:
so that it is more obvious as to what it is

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/locale/en.po
+++ b/locale/en.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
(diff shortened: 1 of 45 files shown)


locale/en.po
```diff
@@ -1082,6 +1081,10 @@ msgstr ""
 msgid "METRICS"
 msgstr "Performance Metrics"
 
+#: ../src/engine/PerformanceMetrics.tscn:66
+msgid "FRAME_DELTA"
+msgstr "Delta: {0}"
+
```
Kasterisk, line 1087: There's still the previous translation key here, is this intended?

hhyyrylainen: That's unintended. I'll clear this from the scene file.

Kasterisk (approved):

Issue #1573, opened 2020-08-31
Make an auto-evo exploring tool


hhyyrylainen:
https://forum.revolutionarygamesstudio.com/t/a-roadmap-for-implementing-auto-evo/703

In short a new GUI page accessible from the main menu where you can setup an auto-evo simulation and step it forward observing the results, should be made.

A key feature for developing auto-evo with this tool would be to have a mode to run like 100 auto-evo runs in a row and collect some overall statistics. I was thinking that an option like "auto reset after 50 generations" and a toggle button to "run generations" would get the first part done.

An optional nice to have feature would be to have a button to enter the game in freebuild mode in the world that was generated by the tool to explore around.

PR #3306, opened 2022-05-03
Rework predation in auto evo

Maxonovien:

This PR modifies predation in auto-evo, by
- [x] incentizing sessility (lower energy cost; causes chasable producers)
- [x] limiting size bonuses to species big enough to engulf (more energy from source for actual predators and less for the various microbe species)
- [x] add a linear bonus to speed (incentive for speed, penalizing species definitely too slow; also see bonuses to size, tweakings in progress)

Work done. This was intended as a small tweak to auto-evo computations. 

Digestion will be added in another PR as the logic is shared with chunks. And similarly for predator prey relationships.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Deus-Codes:
Played for roughly 40 minutes.

While I can't really say I saw increased sessile behavior, I did notice that many more prey items were able to be caught, and that fewer cells as a whole were tiny "zoomers" like we usually see. I usually am not able to really catch anything until I have 2 or 3 flagella as a eukaryote with mitochondria, but now I could catch prey even as a eukaryote because they weren't flying around all over the place. I also noticed that different types of cell membranes were more common; it seems like your updates have de-incentivized speed and has atleast allowed more diverse options now that speed isn't such a huge factor.

While prokaryotes aren't really known predators and predation should be made difficult for them, it isn't due to prokaryotes being too fast in real life but instead or other reasons. So overall, I think this is a great change.

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -38,15 +38,24 @@ public override float FitnessScore(Species species, SimulationCache simulationCa
...
             var engulfScore = 0.0f;
             if (microbeSpeciesHexSize / preyHexSize >
                 Constants.ENGULF_SIZE_RATIO_REQ && !microbeSpecies.MembraneType.CellWall)
             {
                 engulfScore = Constants.AUTO_EVO_ENGULF_PREDATION_SCORE;
-            }
 
-            engulfScore *= predatorSpeed > preySpeed ? 1.0f : Constants.AUTO_EVO_ENGULF_LUCKY_CATCH_PROBABILITY;
+                // The faster the better
+                engulfScore *= predatorSpeed / preySpeed;
```
adQuid, line 49: I question this division here. The prey speed will be the same for all perspective predators, so my assumption would be that the only effect here would be to incentivize engulfing-based predation for slower prey, and disincentivize it for faster prey, rather than give any further advantage to a faster predator of the same prey.

If that was the plan, then the comment is a little misleading (but props for adding the comment)

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -38,15 +38,24 @@ public override float FitnessScore(Species species, SimulationCache simulationCa
...
+
+                // Allow for some degree of lucky engulfment
+                engulfScore *= predatorSpeed > preySpeed ? 1.0f : Constants.AUTO_EVO_ENGULF_LUCKY_CATCH_PROBABILITY;
+
+                // Favor larger microbes, as if they stand more chances in a fight/can digest prey faster.
+                // Here it is artificially after speed comparison for gamey incentive.
+                // Compensating lower speed requires at least PREDATION_SCORE/LUCKY_CATCH_PROBABILITY more hexes.
+                engulfScore += microbeSpeciesHexSize;
```
adQuid, line 57: I would lean towards multiplication here instead of addition, since I don't really have a sense of scale for how many points of speed is equal to one more hex size. This also means that a species with zero speed could be very effective at chasing down prey simply by being huge.

src/microbe_stage/ProcessSystem.cs
```diff
@@ -128,7 +128,7 @@ public ProcessSystem(Node worldRoot)
         // Add movement consumption together
         result.BaseMovement = Constants.BASE_MOVEMENT_ATP_COST * hexCount;
         result.AddConsumption("baseMovement", result.BaseMovement);
-        var totalMovementConsumption = movementATPConsumption + result.BaseMovement;
+        result.TotalMovement = movementATPConsumption + result.BaseMovement;
```
adQuid, line 131: A bug you found while working on these changes I assume?

adQuid (commented):
At this point auto-evo is more of an art than a science, so if this works from the perspective of the playtester I'm cool with it, but there were two things in the predator calculations that raised an eyebrow for me.

hhyyrylainen:
> At this point auto-evo is more of an art than a science

Don't mind me, just linking giving this random link for no reason: https://github.com/Revolutionary-Games/Thrive/issues/1573

src/microbe_stage/ProcessSystem.cs
hhyyrylainen: I didn't look at the code, but I could have sworn that TotalMovement in the result *is* set correctly. This change would indicate that that variable is not initialized at all...

Maxonovien: Coming back a bit late, but indeed, it was never initialized at all... Which is why I had a few bugs at first.

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
Maxonovien: Addition is basically intended to keep the same bonus as before, but restrict it to engulfers. On the other hand, you are right on the pitfall. I'll try that, thanks.

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
Maxonovien: You're right, in that `preySpeed` definitely incentizes predation for slower preys. I guess you clarified it much better.

hhyyrylainen (approved):
The code change looks fine to me. Let's see if someone will comment on the gameplay effects...

buckly90:
Seems to work pretty great for sessile cells. I saw alot more of them show up than ever before. There were still a healthy amount of motile cells as well, which is great. An idle observation: All sessile cells that emerged (with only a single species as exception) seemed to still chase prey when they wandered nearby, but was otherwise stationary. 

There is a negative though; Species don't seem to prioritize speed at all. Not a single flagella seemed to have made an appearance in all the times I tested, which is mildly concerning. The resulting species were all standoffish slugs covered in spikes, lobbing toxins, or both.

nickthenick:
Just playtested this again. I used the auto-evo tool to run 10 generations, and then I jumped into an ocean biome with 6 species. I didn't notice last time how big of a difference this PR makes. It definitely improves the tendency of evolution to evolve predators a lot. I was preyed on 2 times within seconds of spawning before I was finally able to survive more than 10 seconds the third time. And that's really good cause we need that kind of challenge. 

I also noticed that cells more clearly evolved into different roles. I don't know if this was intended or just circumstantial. For example some cells stayed small (which is very good to see), which I haven't seen in the latest official builds. I loved seeing all the different cell types on screen.

image

The only downside I can think of is that no cells seemed to evolve nuclei, and thus there were no eukaryotic predators. Once I became a eukaryote I was suddenly too large for any prokaryote to consume and lost all threat of being hunted. But this might be something that needs to be better addressed in its own PR.

nickthenick (approved):

hhyyrylainen:
We need to decide if this makes things better than before. Today's about the last day to decide which new features to put in 0.5.10

hhyyrylainen:
I guess we are merging this for 0.5.10. At least for the RC, we can do further adjustment / revert this if it turns out to have too many problems when public testing is done.

PR #3310, opened 2022-05-04
Mini alfa revert images

MiniAlfa:
I messed up by merging the main Thrive repo with mine, then discovered i couldn't pull with the images present.
Couldn't revert locally because the I have to pull the changes first (that is not possible because of the images)


MiniAlfa:
Wrong place

PR #3317, opened 2022-05-05
Make Nucleus Reduce Incoming Damage by Half

adQuid:

Pursuant to https://forum.revolutionarygamesstudio.com/t/gameplay-discussion-benefits-of-being-a-larger-cell/881/6, this makes eukaryotes take half damage from all sources.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Microbe.Contact.cs
+++ b/src/microbe_stage/Microbe.Contact.cs
@@ -331,6 +331,11 @@ public partial class Microbe
             amount /= CellTypeProperties.MembraneType.PhysicalResistance;
         }
 
+        if (!CellTypeProperties.IsBacteria)
+        {
+            amount /= 2;
+        }
+
         Hitpoints -= amount;
 
         ModLoader.ModInterface.TriggerOnDamageReceived(this, amount, IsPlayerMicrobe);


Deus-Codes:
Tested this in free build since it was taking a pretty long time for predatory adaptations to evolve in the AI, so it might be an imperfect representation of gameplay experience.

It definitely had an effect with prokaryotes that have a posterior pilus, which is a pretty common build to evolve in traditional gameplay. It doesn't feel like you get your health absolutely drained. It's still a bit extreme if you encounter a cell which likes to rotate a lot due to a lack of damage delay, a pretty common behavior for cells with an anterior pilus.

I still think we should have health scaling with size though. From a gameplay perspective, I already mentioned the sense of progression that comes with gaining extra health, but also, when playing this, I only really cared about the damage debuff if I adapted a more sturdy cell membrane - at 60 health, even halved damage can be risky because each health point is so valuable. I also feel like it's intuitive; halved damage from a nucleus seems a tad arbitrary, but increased health from mass makes sense.

I think we should consider adding mass-based health along with a damage debuff from bacterium from the nucleus, although I'd say at a rate less than half; halved damage and health increasing with mass might be too much of a buff.

adQuid:
Closing as we didn't seem to come to a conclusion here.

hhyyrylainen:
I would have merged this if this hadn't also cut starving damage in half...

thrivingdeveloper:
I can fix this to remove cutting starving damage in half. I can make a new pull request if you want

hhyyrylainen:
I'd appreciate that so I don't have to do that myself.

PR #3325, opened 2022-05-07
Disallow quick save/load during screen transition

### LINKED ISSUE #2191: Quickloading during a cutscene freezes the game
When pressing F9 during the "new game"-cutscene the game loads the savegame, but a frame of the cutscene still takes up the screen

Kasterisk:

Added property in SaveHelper to allow and disallow quick save/load which is then used in TransitionManager to disallow performing those actions while transition is in progress.

**Related Issues**

Fixes #2191

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/saving/SaveHelper.cs
```diff
@@ -41,6 +41,27 @@ public enum SaveOrder
         FileSystem,
     }
 
+    /// <summary>
+    ///   The error that is returned after performing quick load.
```
hhyyrylainen, line 45:
```suggestion
    ///   The error that is returned after trying to perform a quick load.
```
?

src/saving/SaveHelper.cs
```diff
@@ -41,6 +41,27 @@ public enum SaveOrder
...
+    /// </summary>
+    public enum QuickLoadError
+    {
+        /// <summary>
+        ///   No error, quick load is successful (or error is not handled).
+        /// </summary>
+        None,
+
+        /// <summary>
+        ///   The loaded version do not match the current game version.
```
hhyyrylainen, line 55:
```suggestion
        ///   The loaded version does not match the current game version.
```

src/saving/SaveHelper.cs
```diff
@@ -41,6 +41,27 @@ public enum SaveOrder
...
+        /// </summary>
+        VersionMismatch,
+
+        /// <summary>
+        ///   Quick load is made not allowed to be performed.
```
hhyyrylainen, line 60:
```suggestion
        ///   Quick load is currently prevented by the current state of the game.
```
I think this is slightly better wording, but not perfect either

src/saving/SaveHelper.cs
```diff
@@ -52,6 +73,8 @@ public enum SaveOrder
     /// <returns>True if the last save is still recent, false if otherwise.</returns>
     public static bool SavedRecently => lastSave != null ? DateTime.Now - lastSave < Constants.RecentSaveTime : false;
 
+    public static bool AllowQuickSavingAndLoading { get; set; } = true;
```
hhyyrylainen, line 76: Please also document here what this property does.

src/saving/SaveHelper.cs
```diff
@@ -81,6 +104,9 @@ public static void Save(string name, MicrobeEditor state)
     /// <param name="state">Data to include in save</param>
     public static void QuickSave(MicrobeStage state)
     {
+        if (!AllowQuickSavingAndLoading)
```
hhyyrylainen, line 107: Maybe InternalSaveHelper should check this? As we'll end up with 16 quick save variants (at least, one per stage and editor, and with much more editors later in the game, we might end up with a ton of these checks being duplicated).

Kasterisk: Yeah that's preferrable, I actually didn't saw that there's a SaveType parameter that you can use to distinguish saves, that was why I done it like this.

RacerBG (approved):
Nothing happens when I press F5 or F9 during the cutscene, I believe this is the expected behaviour of the game.

src/saving/SaveHelper.cs
```diff
@@ -398,6 +427,9 @@ public static void ClearLastSaveTime()
     private static void InternalSaveHelper(SaveInformation.SaveType type, MainGameState gameState,
         Action<Save> copyInfoToSave, Func<Node> stateRoot, string? saveName = null)
     {
+        if (type == SaveInformation.SaveType.QuickSave && !AllowQuickSavingAndLoading)
+            return;
```
hhyyrylainen, line 431: One more tiny thing, could you add a print here warning the user why they couldn't save? That way if ever AllowQuickSavingAndLoading variable has a bug and isn't reset, we can see from the game logs why someone complains they can't save.

hhyyrylainen (changes requested):

src/saving/SaveHelper.cs
hhyyrylainen: To clarify it should be something like `GD.Print("Can't save due to quick save being currently suppressed");`

hhyyrylainen (approved):

PR #3345, opened 2022-05-19
Added Cilia

### LINKED ISSUE #174: Add cilia
Another movement organelle, but fundamentally different as they apply tangential force and torque.

hhyyrylainen:

Adds cilia as a basic organelle implementation in the game.

Needs rotation speed and ATP cost balancing, which I'm hoping someone else will do.

**Related Issues**

closes #174

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/Thrive.csproj
+++ b/Thrive.csproj
@@ -244,6 +244,7 @@
     <Compile Include="src\microbe_stage\CompoundBalance.cs" />
     <Compile Include="src\microbe_stage\CompoundCloudPlane.cs" />
     <Compile Include="src\microbe_stage\MicrobeSystem.cs" />
+    <Compile Include="src\microbe_stage\organelle_components\CiliaComponent.cs" />
     <Compile Include="src\microbe_stage\organelle_components\EmptyOrganelleComponent.cs" />
     <Compile Include="src\microbe_stage\PatchMapNameGenerator.cs" />
     <Compile Include="src\microbe_stage\editor\BarHelper.cs" />
--- a/assets/models/organelles/Cilia.anim
+++ b/assets/models/organelles/Cilia.anim
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:4b0502a496674c30be732f527681aadab8cd31bb7b52e13b83c2a68009a168c6
+size 14257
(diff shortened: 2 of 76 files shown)


Oliveriver:
Took this for a short _spin_.

I noticed `requiresNucleus` is set to true but `prokaryoteChance` is non-zero. These should align, and in my opinion, cilia should be available for cells without a nucleus.

I did unfortunately hit some performance issues, though it's hard to tell if that was cilia-related. This was only in rather extreme cases too, where I built a big cell covered in cilia in freebuild.

In terms of gameplay, I found this a nicely refreshing change. Other than the controls being frustrating at times since the cell wouldn't quite go where I wanted, it was very enjoyable. I discovered lack of cilia nerfs the pilus, in my opinion a good and long overdue change.

hhyyrylainen:
> I noticed requiresNucleus is set to true but prokaryoteChance is non-zero. These should align, and in my opinion, cilia should be available for cells without a nucleus.

I think that makes sense. I didn't touch the requires a nucleus option for the cilia, it was already set on...

> I did unfortunately hit some performance issues, though it's hard to tell if that was cilia-related. This was only in rather extreme cases too, where I built a big cell covered in cilia in freebuild.

I'm pretty certain cilia should not impact the game performance in a significant way. In terms of how much extra computation they cause they should actually be a bit better than the flagella. Unless of course the much more complicated animation setup trips up Godot in some unexpected way...

hhyyrylainen:
Changed to allowing cilia for prokaryotes and I added a check to make sure that kind of invalid organelle data is not made in the future.

I also did a quick profiling run and could not find any evidence that cilia use any more CPU power than (at least in our C# code) flagellum and other externally positioned organelles.

hhyyrylainen:
I just remembered one thing I didn't do yet, colonies need handling for cilia. I'll do that tomorrow.

buckly90:
* Cilia animation does not stop on cell death, minor issue but looks funny.
* Cilia can be used by prokaryotes. On Earth, it's a feature only seen on eukaryotes so it's currently non-LAWK. Again, minor issue.
* I feel like rotation speed decreases just slightly too quickly which can lead to unreasonable rotation speeds pretty soon. This is bad because it might make players feel forced to evolve cilia. We'll either need to set the minimum speed to somewhere around 10, or slightly decrease the rotation speed penalty.

hhyyrylainen:
> Cilia animation does not stop on cell death, minor issue but looks funny.

I think the same applies to the flagella as well. I put it on my list of issues I need to open soon (Make organelle (cilia, flagella) animations drop when dropped as cell corpse chunks or marine snow)

> Cilia can be used by prokaryotes. On Earth, it's a feature only seen on eukaryotes so it's currently non-LAWK. Again, minor issue.

Changed that.

PR #3348, opened 2022-05-20
Refactored pausing to use named pause locks to solve pause stuck after saving

hhyyrylainen:
instead of juggling the single paused state flag which started to become unwieldy and error prone

**Related Issues**

Should fix pausing getting stuck on with the changes in: https://github.com/Revolutionary-Games/Thrive/commit/a6a39d071f10a2dbc1a520c342da3b15fbef359f

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/MicrobeStage.tscn
+++ b/src/microbe_stage/MicrobeStage.tscn
@@ -2252,7 +2252,6 @@ custom_styles/separator = SubResource( 36 )
 anims/FadeInOut = SubResource( 37 )
 
 [node name="PauseMenu" parent="." instance=ExtResource( 11 )]
-visible = true
 
 [connection signal="pressed" from="MicrobeHUD/BottomLeft/LeftPanels/EnvironmentPanel/VBoxContainer/Header/HBoxContainer/EnvironmentExpandButton" to="MicrobeHUD" method="OnEnvironmentPanelSizeButtonPressed" binds= [ "expand" ]]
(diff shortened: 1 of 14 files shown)


src/microbe_stage/MicrobeStage.tscn
```diff
@@ -2252,6 +2252,7 @@ custom_styles/separator = SubResource( 36 )
 anims/FadeInOut = SubResource( 37 )
 
 [node name="PauseMenu" parent="." instance=ExtResource( 11 )]
+visible = true
```
Kasterisk, line 2255: Godot _really_ insist for this to be visible, last time I got this the actual menu is not shown and this value is basically just saved here for no reason. Could be wrong in this case but I guess we can just let this be?

Kasterisk (changes requested):

src/microbe_stage/MicrobeStage.tscn
hhyyrylainen: Good catch. I'll fix this in a bit.

hhyyrylainen: Looks like truly this change could not be reverted with the Godot editor so I had to manually do this...

Kasterisk: I know right.

Kasterisk (approved):

PR #3360, opened 2022-05-20
Cilia Balance

buckly90:
Adjusts loss of turning speed from larger cells, as well as the impact of cilia in the wake of implementation.

As always, I would like some input from others about if it feels right to them or not.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -72,9 +72,9 @@ public static class Constants
 
     public const float CELL_BASE_ROTATION = 0.2f;
     public const float CELL_MAX_ROTATION = 0.40f;
-    public const float CELL_MIN_ROTATION = 0.0001f;
-    public const float CELL_MOMENT_OF_INERTIA_DISTANCE_MULTIPLIER = 0.8f;
-    public const float CILIA_ROTATION_FACTOR = 0.035f;
+    public const float CELL_MIN_ROTATION = 0.01f;
+    public const float CELL_MOMENT_OF_INERTIA_DISTANCE_MULTIPLIER = 0.2f;
+    public const float CILIA_ROTATION_FACTOR = 0.020f;
     public const float CILIA_RADIUS_FACTOR_MULTIPLIER = 0.7f;
 
     public const float CELL_COLONY_MAX_ROTATION_MULTIPLIER = 2.5f;
@@ -82,9 +82,9 @@ public static class Constants
     public const float CELL_COLONY_MAX_ROTATION_HELP = 2.5f;
     public const float CELL_COLONY_MEMBER_ROTATION_FACTOR_MULTIPLIER = 45.0f;
 
-    public const float CILIA_ENERGY_COST = 3.0f;
+    public const float CILIA_ENERGY_COST = 2.0f;
     public const float CILIA_ROTATION_NEEDED_FOR_ATP_COST = 0.03f;
-    public const float CILIA_ROTATION_ENERGY_BASE_MULTIPLIER = 6.0f;
+    public const float CILIA_ROTATION_ENERGY_BASE_MULTIPLIER = 4.0f;
 
     public const float CILIA_DEFAULT_ANIMATION_SPEED = 0.3f;
     public const float CILIA_MIN_ANIMATION_SPEED = 0.15f;
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -72,9 +72,9 @@ public static class Constants
 
     public const float CELL_BASE_ROTATION = 0.2f;
     public const float CELL_MAX_ROTATION = 0.40f;
-    public const float CELL_MIN_ROTATION = 0.01f;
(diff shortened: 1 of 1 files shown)


simulation_parameters/Constants.cs
```diff
@@ -72,19 +72,19 @@ public static class Constants
...
     public const float CELL_COLONY_MAX_ROTATION_MULTIPLIER = 2.5f;
     public const float CELL_COLONY_MIN_ROTATION_MULTIPLIER = 0.05f;
     public const float CELL_COLONY_MAX_ROTATION_HELP = 2.5f;
     public const float CELL_COLONY_MEMBER_ROTATION_FACTOR_MULTIPLIER = 45.0f;
 
-    public const float CILIA_ENERGY_COST = 3.0f;
+    public const float CILIA_ENERGY_COST = 2.0f;
     public const float CILIA_ROTATION_NEEDED_FOR_ATP_COST = 0.03f;
-    public const float CILIA_ROTATION_ENERGY_BASE_MULTIPLIER = 6.0f;
+    public const float CILIA_ROTATION_ENERGY_BASE_MULTIPLIER = 4.0f;
```
hhyyrylainen, line 87: Does this specific change have an effect? This is really only used to calculate how fast a cell needs to turn before it consumes CILIA_ENERGY_COST. So by lowering this cells need to turn much faster before cilia consumes the full ATP.

buckly90: Yeah, it was very easy to consume ATP by just barely turning prior, which would mean it was easy to drain ATP even if you weren't really planning on turning at all.

hhyyrylainen: I suppose that's fine, then. I did test that if you literally move your mouse a few pixels *that* didn't consume much ATP (at least visibly).

buckly90: I didn't actually thoroughly test that specific change though, I tweaked it last night before calling it quits, and forgot to actually test it before uploading today

hhyyrylainen: Testing it would definitely be a good thing.

Deus-Codes:
Good call to increase non-ciliate rotation. I personally think setting cell-min-rotation to .005 and cell-moment-of-inertia to .4 works a bit better for balance as it gives cilia a bit more function. Before I could reasonably neglect cilia, but after, cilia is a bit more useful especially for heterotrophs. Thoughts? 

I would like to also say that ideally, I think we'd want cilia to also include a small speed buff. Whereas flagella represents a strong force in one direction, cilia represents a smaller force in multiple directions, not granting as much speed but giving much more maneuverability and flexibility. We can then increase ATP cost of cilia to make it a lot more of a choice between flagella and cilia - most life forms aren't really both flagellate and ciliate, though we shouldn't outright eliminate that possibility but reflect its difficulty I think. 

But for this implementation of cilia and since hh pulled this long-awaited feature out of nowhere, it's good for now. If I find the bravery, I'll perhaps try to make a prototype.

buckly90:
> Good call to increase non-ciliate rotation. I personally think setting cell-min-rotation to .005 and cell-moment-of-inertia to .4 works a bit better for balance as it gives cilia a bit more function. Before I could reasonably neglect cilia, but after, cilia is a bit more useful especially for heterotrophs. Thoughts?

My goal was to make it so that you could reasonably forgo the use of cilia with a decently designed cell, and more importantly, ease up on any frustration players might face from the new factor of rotation speed that wasn't present before. 
Many players build inefficient and bloated cells, so they will certainly feel the pressure of a constantly slowing turn-speed. With the current set-up, players will still be rewarded for building more streamlined cells, but won't overly suffer if they don't entirely understand how to do so.  I'll reconsider this stance.

 I feel that a turn-speed of around 10 is the minimum acceptable value, where anything below that begins to become unbearably slow for the player. Thus I decided that a minimum of 5 is a good compromise between punishment and leniency. That way if a player is just really really bad at building their cell, they won't be completely immobilized, but still feel the pressure of their choices.


> I would like to also say that ideally, I think we'd want cilia to also include a small speed buff. Whereas flagella represents a strong force in one direction, cilia represents a smaller force in multiple directions, not granting as much speed but giving much more maneuverability and flexibility. We can then increase ATP cost of cilia to make it a lot more of a choice between flagella and cilia - most life forms aren't really both flagellate and ciliate, though we shouldn't outright eliminate that possibility but reflect its difficulty I think.

Not a bad idea, I had considered giving flagella a small turning speed buff as well. It's something we will have to experiment with sometime, I'm sure it won't be too hard to implement.


Thank you for your input, I'll experiment with the changes you suggested and see how it feels.

buckly90:
Seems I was actually a bit too lenient. I've adjusted accordingly.

Deus-Codes:
I think these numbers are good feel-wise. If we do bump into some community grievances, we can adjust with 0.5.9.1 rather easily. Do you want further testing on ATP consumption?

Deus-Codes (approved):

PR #3362, opened 2022-05-20
Add warning that organelles require nucleus in tooltips

Oliveriver:

Displays a red warning in the tooltips for organelles which require a nucleus.

Since cilia are now in the game and require a nucleus but aren't clearly indicated as such (unlike other organelles which feature in the same panel section as the nucleus), I thought it would be a good idea to add a warning to the relevant tooltips to avoid confusion. Ideally I would like it if the warning disappeared once the user had placed a nucleus, but I couldn't figure out where in the code it would be best to add that. Suggestions welcome.

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/tooltip/microbe_editor/SelectionMenuToolTip.cs
+++ b/src/gui_common/tooltip/microbe_editor/SelectionMenuToolTip.cs
@@ -327,7 +327,7 @@ public class SelectionMenuToolTip : Control, ICustomToolTip
     {
         if (requiresNucleusLabel == null)
             return;
-        
+
         requiresNucleusLabel.Text = TranslationServer.Translate("REQUIRES_NUCLEUS");
         requiresNucleusLabel.Visible = requiresNucleus;
     }
--- a/src/gui_common/tooltip/microbe_editor/SelectionMenuToolTip.cs
+++ b/src/gui_common/tooltip/microbe_editor/SelectionMenuToolTip.cs
@@ -17,6 +17,9 @@ public class SelectionMenuToolTip : Control, ICustomToolTip
     public NodePath MpLabelPath = null!;
 
     [Export]
+    public NodePath RequiresNucleusPath = null!;
+
+    [Export]
     public NodePath DescriptionLabelPath = null!;
 
     [Export]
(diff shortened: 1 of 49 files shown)


hhyyrylainen:
Does the warning go away once the player has a nucleus? I think that would be a pretty important thing.

Oliveriver:
> Does the warning go away once the player has a nucleus? I think that would be a pretty important thing.

As I said, I would have liked that, but wasn’t sure of the best way to go about it. I’ll give it another go in a few days.

hhyyrylainen:
Best way is probably the same way as the editor discount is applied to the tooltips. @Kasterisk might have some other idea as well.

Kasterisk:
I think the simplest solution would be to fetch all of the eukaryote-specific organelles tooltip and then set their `RequiresNucleus` property (since this also controls whether the label should be visible) to false when a nucleus is placed. This can be done by calling a new or existing method like [CellEditorComponent.UpdatePartsAvailability](https://github.com/Revolutionary-Games/Thrive/blob/c97e936ae6c15a003bd78f2fd789e80d19e49dc7/src/microbe_stage/editor/CellEditorComponent.cs#L1588) in `CellEditorComponent.OnOrganellesChanged` and to check if a nucleus has been placed, `CellEditorComponent.HasNucleus` can be used.

hhyyrylainen:
One more thing to consider is that loading a save made in the editor needs to work correctly so the initial state needs to be ensured is correct for the tooltips.

Oliveriver:
Ready for review.

The label now appears/disappears as appropriate to reflect the existence of a nucleus. Seems robust after some testing. Visual suggestions (font, size) are welcome as I don't really know the intended style.

src/gui_common/tooltip/microbe_editor/SelectionMenuToolTip.cs
```diff
@@ -306,6 +323,15 @@ private void UpdateMpCost()
         mpLabel.Text = ((int)(mpCost * editorCostFactor)).ToString(CultureInfo.CurrentCulture);
     }
 
+    private void UpdateRequiresNucleus()
+    {
+        if (requiresNucleusLabel == null)
+            return;
+
+        requiresNucleusLabel.Text = TranslationServer.Translate("REQUIRES_NUCLEUS");
```
hhyyrylainen, line 331: One small comment from me:
The label text seems to be always the same, so why isn't it set in the scene file?

hhyyrylainen (approved):
Code looks fine to me now, let's see if I can manage to get someone to playtest this next...

Kasterisk (approved):
Works fine the last time I tested this, no problem loading save too, should also works fine now as there are no major code changes.

Issue #3108, opened 2022-02-21
Add a hotkey to toggle the microbe stage pause feature


hhyyrylainen:
this will make it easier to pause the game while trying to mouse over a cell.

This should be easy just need to:
- add key to godot project
- to our JSON config
- add translations
- add the code to the microbe stage that uses our input system to detect when it is pressed and forward that to where the clicks on the pause button go

Kasterisk:
I'd suggest <kbd>P</kbd> for the key as usual in most games but that's already taken by the phosphate cloud spawn cheat... so maybe backtick button? And if we end up adding command console then it can just use the tilde button, that is <kbd>Shift</kbd>+<kbd>`</kbd>

hhyyrylainen:
Might be good to try the new physical keys here as backticks and other punctuation that could be used here varies by keyboard layout. Then again maybe it makes more sense for the cheat keys to get moved to the weirder keys.

PR #3363, opened 2022-05-20
Pause hotkey

Oliveriver:

Adds ability to pause the game with the spacebar, with a visual indicator.

**Related Issues**

Fixes https://github.com/Revolutionary-Games/Thrive/issues/3108

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Seeing as this seems to be a WIP PR, I converted this into a draft.

Blisto91:
If jumping or flying or similar ever became a thing in some later 3D stages then i would probably imagine space being a odd default choice for pause. At least since i imagine it should be kept consistent between stages.
Space as pause is common in grand strategy games and top down so in that way it makes sense i guess. But in games where there is 3D movement in a 3D environment it might be a bit odd.

Tho that is ofc probably far into the future so might not be an issue worth considering at all yet if it will even be a thing.

hhyyrylainen:
> If jumping or flying or similar ever became a thing in some later 3D stages then i would probably imagine space being a odd default choice for pause. At least since i imagine it should be kept consistent between stages.
Space as pause is common in grand strategy games and top down so in that way it makes sense i guess. But in games where there is 3D movement in a 3D environment it might be a bit odd.

Well with just 2 or maybe 3 stages out of all of the stages would have jumping so space being pause would be used in the majority of the game. So I think there's precedent from strategy games to use it as the pause button when not needed for jumping.

Blisto91:
Ye but i would imagine the same default button would be kept through out the game for good UX and less complicated code.

Also i realized after i had posted the comment it like a total non issue to think about now since that is so far of anyway and changing the default key at that time would be easy as pie 😁 so ignore me 👀

Oliveriver:
I've been instinctively pressing space to pause Thrive even when it isn't part of the game. To me it just feels the most natural.

I agree that it could cause conflict once jumping/flying/etc. exist, but that's not a problem to discuss now. If such a day comes, we'll just have to re-map the key and tell everyone to get used to it.

hhyyrylainen:
> I agree that it could cause conflict once jumping/flying/etc. exist, but that's not a problem to discuss now. If such a day comes, we'll just have to re-map the key and tell everyone to get used to it.

I did not bring up that before because the key mapping system allows specifying which parts of the game a key applies to. So there's no problem really with space being both the jump action in stages where that is used and pause in stages that don't have the jump action.

buckly90:
> > I agree that it could cause conflict once jumping/flying/etc. exist, but that's not a problem to discuss now. If such a day comes, we'll just have to re-map the key and tell everyone to get used to it.
> 
> I did not bring up that before because the key mapping system allows specifying which parts of the game a key applies to. So there's no problem really with space being both the jump action in stages where that is used and pause in stages that don't have the jump action.

I disagree, if we use space as a pause-key in an earlier stage and then a jump key in later stages, players might get a little jumbled with trying to pause the game only to jump instead. It's a relatively minor issue but it will disrupt the flow. I recommend using an alternative key, like P. It's currently only used for a cheat right now, and I say the bindings of cheats are lower priority.

hhyyrylainen:
If the strategy stages of Thrive end up being closer to a tactics game where pausing often is a pretty important thing (or many peoples prefer that kind of playstyle) space should be the pause key to follow practices of that genre. I think we should assign keys based on the expectations of a genre rather than avoiding keys doing different things in different stages. Action games get away with remapping up/down when you hop in a plane as that's what flying games have as a convention.

buckly90:
Hmmm, that's a fair point

Perhaps we could get away with that in later stages since the player wouldn't be controlling a specific character, thus they would have no expectations to jump/ascend/etc with the space key as opposed to stages where you are in direct control, of a character.

hhyyrylainen:
> Perhaps we could get away with that in later stages since the player wouldn't be controlling a specific character, thus they would have no expectations to jump/ascend/etc with the space key as opposed to stages where you are in direct control, of a character.

Before the switch to strategy view, basically the only parts that applies to are the microbe stage and early multicellular.

Oliveriver:
Ready for review.

There are two sticking points. First, toggling pause is currently only disabled when the menu is visible. There are other screens that maybe should disable pause too (multicellular popup, tutorial popup) but I wasn't sure.

Second, I added the pause hotkey setting under camera in the options menu. Not sure if it should go here or in misc.

Kasterisk:
I think toggling pause should be disabled on those screens too, it feels buggy otherwise. The pause hotkey setting should go in misc as some other hotkeys were already there, so I think it makes sense for this one to be there too.

On another note, the hotkey conflicts with controls' [focus](https://docs.godotengine.org/en/stable/tutorials/ui/gui_navigation.html) behavior. This means pressing space along with toggling pause also presses any last focused control, this can be any interactable elements. The way I'd go to fix this would be to release the focus of the focused control in the code where the hotkey is handled. [`get_focus_owner`](https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-method-get-focus-owner) and [`release_focus`](https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-method-release-focus) would be useful for this.

Another thing that I noticed was that the space button on the pause info was too small which could be hard discern. Changing this would involve tweaking the underlying custom RTL code which I could handle in another PR.

hhyyrylainen:
> On another note, the hotkey conflicts with controls' [focus](https://docs.godotengine.org/en/stable/tutorials/ui/gui_navigation.html) behavior. This means pressing space along with toggling pause also presses any last focused control, this can be any interactable elements. The way I'd go to fix this would be to release the focus of the focused control in the code where the hotkey is handled. [get_focus_owner](https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-method-get-focus-owner) and [release_focus](https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-method-release-focus) would be useful for this.

I think this warrants a bit more consideration as GUI navigation needs to be considered, so if a button is navigated to by the keyboard pressing space needs to still properly click it. We may need to consider when the HUD GUI needs to be entirely unfocused.

hhyyrylainen:
> pause hotkey, so it won't be propagated to the focused control but I'm not sure the way how to go about it.

I think there's already places in the code where we mark input as used, which I hope would also mean the GUI knows that something was consumed.

We already have `GetTree().SetInputAsHandled();` in the InputManager.

Oliveriver:
From what I can tell, checking whether `GetOwnerFocus()` is null does enough. It covers the cases for the multicellular and tutorial popups, as well as anything else that takes keyboard focus. I've made the change so let me know what you think.

buckly90:
Seems to function without error, everything went fine.

Having the paused game overlay is a very nice touch, great work!

hhyyrylainen:
@Kasterisk could you re-check this PR?

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -729,6 +742,36 @@ public void UpdateEnvironmentalBars(BiomeConditions biome)
         // TODO: pressure?
     }
 
+    public void PauseButtonPressed()
+    {
+        if (!menu.Visible)
+        {
```
Kasterisk, line 748: Use early return.

Kasterisk:
> From what I can tell, checking whether GetFocusOwner() is null does enough. It covers the cases for the multicellular and tutorial popups, as well as anything else that takes keyboard focus. I've made the change so let me know what you think.

I guess this is a good enough solution.

PR #3370, opened 2022-05-23
Refactored `TransitionManager` to do away with Godot-based signals

Kasterisk:

Refactored TransitionManager and add fades to the loading screen so there'll be no more abrupt screen change from the loading screen to the cell editor or vice versa. Also moved TransitionManager to ScreenOverlays after LoadingScreen.

https://user-images.githubusercontent.com/54026083/169830859-523b1d44-f7f7-4dd4-82cb-df8526662865.mp4

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (changes requested):
When I enter the freebuild editor I get a screen fade that is happening and then presumably as the editor gets ready that fade cuts in abruptly so it doesn't look good.
In other cases I didn't notice any screen fade problems.

Kasterisk:
I'm assuming from what you've described, was that the fades looked something like this

https://user-images.githubusercontent.com/54026083/170658598-60ac5528-95b1-420d-b310-7a8db25403e4.mp4

I think this is tolerable, but what do you have in mind?

hhyyrylainen:
That video doesn't look like what it does for me, for me there is a specific cut where the transition skips to a different point. Your video looks fine to me. I can try to record what it looks like to me later today.

hhyyrylainen:
Here's the video (I don't know why but the picture quality got destroyed by OBS):

https://user-images.githubusercontent.com/3796411/170708610-9403e2c9-f9d2-41f1-8cd3-2598b014d619.mp4

If you compare the end of my video to yours, it's clearly visible how there's one extra attempted fade there that cuts off a previous one abruptly.

src/gui_common/TransitionManager.cs
```diff
@@ -1,167 +1,201 @@
...
 /// <summary>
 ///   Manages the screen transitions, usually used for when
 ///   switching scenes. This is autoloaded
 /// </summary>
-public class TransitionManager : NodeWithInput
+public class TransitionManager : ControlWithInput
 {
-    private static TransitionManager? instance;
+    private static readonly PackedScene ScreenFadeScene = GD.Load<PackedScene>("res://src/gui_common/ScreenFade.tscn");
+    private static readonly PackedScene CutsceneScene = GD.Load<PackedScene>("res://src/gui_common/Cutscene.tscn");
```
hhyyrylainen, line 12: I think it would be better to load these in _Ready or the constructor like was done before, I like avoiding Godot resource loading that happens on assembly load time.

hhyyrylainen (changes requested):
A few tiny code comments. In testing I only found one more problem (and everything else seems to be fine): when quick loading the transition to the loading screen has a brief flicker at the start, like the loading screen fades in but then once it has faded in there's a really quick flash before the loading screen then shows normally.

Kasterisk:
I did found the cause for that, it was a duplicate fade in on the editor enter, removing it fixed it, seems to be a problem in microbe editor only. Now with that fixed, one more popped up, somehow the editor confirm button stops working after loading a save. I have a slight hunch this bug is not specific to this PR but I'll check.

src/gui_common/TransitionManager.cs
Kasterisk: Static readonlies can only be initialized this way or in a static constructor.

hhyyrylainen: I'd make this an instance specified readonly field then, as that's why preferred way here.

hhyyrylainen:
One thing I was thinking might help is to track the if the game is fading in our out and just basically skip the transitions if a duplicate fade out or fade in is added.

Kasterisk:
That's a good idea, I just made `AddSequence` skips all previous sequences (optional but it defaults to that), this should make it behave similar to the previous system.

hhyyrylainen:
I feel like that may still experience this problem. I was more thinking that you'd only allow a fade in to happen if a fade out has happened before it. That way no double fade in or double fade out sequences could be caused.

Also the way to add the sequences starts being pretty complex so I think it would simplify a lot of code if you added an overload that takes just a single ITransition as the first parameter.

Kasterisk:
> I feel like that may still experience this problem. I was more thinking that you'd only allow a fade in to happen if a fade out has happened before it. That way no double fade in or double fade out sequences could be caused.

Yeah that will be a tricky but I'll see what I can do.

Kasterisk:
Did just that, in the simplest and least code to write albeit in a not so clean way as I don't want to write a full fledged general purpose transitions order checker that probably won't be used much beyond screen fading.

src/general/EditorBase.cs
```diff
@@ -586,7 +590,10 @@ protected virtual void InitEditor(bool fresh)
             // Make sure non-default tab button is highlighted right if we loaded a save where the tab was changed
             editorTabSelector?.SetCurrentTab(selectedEditorTab);
 
-            FadeIn();
+            // Just assume that a transition is finished (even though one may still be running after save load is
+            // complete).  This should be fine as it will just be skipped if the player immediately exits the editor
```
hhyyrylainen, line 594:
```suggestion
            // complete). This should be fine as it will just be skipped if the player immediately exits the editor
```

src/gui_common/Cutscene.cs
```diff
@@ -3,53 +3,84 @@
...
+        UpdateVideoPlayer();
+        Hide();
     }
 
-    public void OnStarted()
+    public void Begin()
     {
+        if (cutsceneVideoPlayer == null)
+        {
+            GD.Print("Video player missing, can't play cutscene");
```
hhyyrylainen, line 55:
```suggestion
            GD.PrintErr("Video player missing, can't play cutscene");
```

src/gui_common/TransitionManager.cs
```diff
@@ -30,138 +31,235 @@ private TransitionManager()
...
-    public bool CancelTransitionPressed()
-    {
-        if (!HasQueuedTransitions)
-            return false;
+    public bool HasQueuedTransitions => queuedSequences.Count > 0;
 
-        CancelQueuedTransitions();
-        return true;
-    }
+    public ScreenFade.FadeType? LastFadedType { get; set; }
```
hhyyrylainen, line 36: This should be private

hhyyrylainen (changes requested):
I was about to merge this (hence why I merged 2 typo fixes), but I noticed one more problem: when exiting the editor back to the stage from a save that was loaded where the game is in the editor, the stage fade in doesn't play for some reason, in all other cases exiting the editor seems to work correctly.

Kasterisk:
Should be fixed, was just a flawed check in the stage enter transition call for loaded saves.

hhyyrylainen (approved):
I think this is fine now.

PR #3376, opened 2022-06-01
Add symmetry support to the editor hex grid

Kasterisk:

Should've been added way earlier.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/HexEditorComponentBase.cs
+++ b/src/general/HexEditorComponentBase.cs
@@ -23,9 +23,6 @@ public abstract class
     public NodePath EditorArrowPath = null!;
 
     [Export]
-    public NodePath EditorGridPath = null!;
-
-    [Export]
     public NodePath CameraFollowPath = null!;
 
     [Export]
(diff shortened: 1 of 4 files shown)


hhyyrylainen:
Code seems fine to me, could someone playtest this? There's probably no issue but I'd also like to have the multicell editor checked that it still works with this change. There it would also be nice if the symmetry hex grids worked there as it *should* have symmetry enabled editing there.

Kasterisk:
Still works.

hhyyrylainen (changes requested):
One small typo I noticed in the code. Otherwise it looks good.

In playtesting I noticed just one problem, there's a slight graphical glitch when multiple hex grids overlap (the lines look very jagged):



Compare to everything being smooth with just one grid:

Kasterisk:
There must be a shader magic to workaround this overlapping glitch that I can't quite put my finger on.

hhyyrylainen:
One solution would be to not allow the hex grids to display too close to each other to make sure they don't overlap. My graphics programming skills aren't enough to try to solve this, though I suppose the solution might be similar as for the compound clouds: a single large shader covering the screen could be used that is given all the coordinates around which the hexes should be rendered.

hhyyrylainen:

PR #3378, opened 2022-06-03
Fixed remaining cases where mutation code could get auto-evo stuck

### LINKED ISSUE #3373: Auto-evo can get stuck in organelle mutations calculation
Seems like my fix to the islands fixing code can still get stuck in some cases.

My guess as to how this could be fixed: what if the code just doesn’t calculate all of the hexes to move at once? Instead just calculating the move direction for each island organelle separately and move that in the needed direction. That way the symmetry related not knowing how to move code shouldn't be able to happen.

https://community.revolutionarygamesstudio.com/t/auto-evolve-issue/4843/7?u=hhyyrylainen

Leinourdian:

I found two problems that caused the issue. First, the code didn't always recognize island hexes correctly meaning that they could be next to main hexes. Second, the code could calculate the movement path incorrectly if `minSubHex.Q == minSubHex.R`. This PR fixes both of them. I tested this quite a bit and think that auto-evo should now never get stuck.

**Related Issues**

Fixes #3373

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/HexLayout.cs
+++ b/src/general/HexLayout.cs
@@ -220,7 +220,7 @@ public abstract class HexLayout<T> : ICollection<T>
         var hexesWithNeighbours = new List<Hex> { initHex };
 
         // These are all of the existing hexes, that if there are no islands will all be visited
-        var shouldBeVisited = existingHexes.Select(p => p.Position).ToList();
+        var shouldBeVisited = ComputeHexCache();
 
         CheckmarkNeighbors(hexesWithNeighbours);
 
--- a/src/general/Mutations.cs
+++ b/src/general/Mutations.cs
@@ -282,14 +282,10 @@ public class Mutations
                     break;
             }
 
-            minSubHex.Q = (int)(minSubHex.Q * (minDistance - 1.0) / minDistance);
-            minSubHex.R = (int)(minSubHex.R * (minDistance - 1.0) / minDistance);
+            if (minSubHex.Q != minSubHex.R)
+                minSubHex.Q = (int)(minSubHex.Q * (minDistance - 1.0) / minDistance);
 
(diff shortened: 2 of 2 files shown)


src/general/HexLayout.cs
```diff
@@ -220,7 +220,7 @@ public List<Hex> GetIslandHexes()
         var hexesWithNeighbours = new List<Hex> { initHex };
 
         // These are all of the existing hexes, that if there are no islands will all be visited
-        var shouldBeVisited = existingHexes.Select(p => p.Position).ToList();
+        var shouldBeVisited = ComputeHexCache();
```
hhyyrylainen, line 223: I think I see. The bug here was that as there are many organelles that are multiple hexes large, this didn't properly loop *all* hexes instead only the first hex of an organelle. Did I understand this right?

hhyyrylainen:
Seeing as the hex island fixing math got very complex looking, I'll just leave the code be, and ask for testing. Though if you could, could you add a line or two of a comment to explain the if and the new calculation there a bit?

src/general/HexLayout.cs
Leinourdian: Yes, exactly.

Leinourdian:
Sure, I added a comment. The if statement is there because otherwise the path could be too short if minSubHex is (1, 1) for example. Without it the path would become (0, 0) when it needs to be (1, 0) or (0, 1). It basically adds a default direction like your fix but it also works for negative numbers and bigger numbers.

buckly90:
Couldn't get any freezes to happen during loading, so it looks good from my end.

Kasterisk (approved):
Great fix. No more stuck issue even by changing the mutation constants.

PR #3380, opened 2022-06-05
Add support for specifying image size to the key prompt bbcode

### LINKED ISSUE #2434: Remove CustomRichTextLabel compound lookup fallback
This should be done in a few months maybe in like October or November of 2021?
CustomRichTextLabel has code to fallback to old style compound names, which should be removed for clarity.

Kasterisk:

Follow up to #3363. Added `size` and `ascent` attribute to the [Input](https://wiki.revolutionarygamesstudio.com/wiki/Translation_Guidelines#Input) tag to allow increasing the Space key prompt's size. Also took the time to remove the long overdue fallback removal of #2434.

**Related Issues**

Closes #2434

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/gui_common/CustomRichTextLabel.cs
```diff
@@ -323,12 +314,25 @@ private static string BuildTemplateForTag(string input, ThriveBbCode bbcode, Lis
...
+                if (pairs.TryGetValue("size", out string sizeInput))
+                    size = sizeInput.ToInt();
+
+                if (pairs.TryGetValue("ascent", out string ascentInput))
+                    ascent = ascentInput.ToInt();
+
+                var ascentFont = $"res://src/gui_common/fonts/BBCode-Image-VerticalCenterAlign-{ascent}.tres";
+
+                if (!ResourceLoader.Exists(ascentFont))
+                    GD.PrintErr($"Input Action: No ascent font found for {ascent}, fallback to using default");
```
hhyyrylainen, line 333: This doesn't seem to actually fallback to anything, just prints this error?

hhyyrylainen (changes requested):
Looks good ingame but I think I noticed an improperly handled error condition in the code, other than that the code looks good as well.

Oliveriver:
I can’t test this until tomorrow, and can’t seem to comment on the diff on mobile. So I’ll mention what I noticed here.

The size of the g_pause icon has been adjusted to make space the right size, but what happens if the input key for g_pause is changed? Do you get a bigger icon for the new key, which we wouldn’t want?

src/gui_common/CustomRichTextLabel.cs
Kasterisk: `ascent` defaults to 9 in line 322.

hhyyrylainen: But on line 335, you use the `ascentFont` string anyway, which now has an invalid value?

Kasterisk:
I actually haven't considered that but yes the icon did get bigger... It can't be tweaked at runtime since the size specifier is inside the locale file. Perhaps not increasing the size is the right way?

hhyyrylainen:
I think the icons need to automatically adjust to the size of the text that goes in them. Otherwise we'll have problems at some point.

src/gui_common/CustomRichTextLabel.cs
Kasterisk: I just realized what you're saying, yes `ascent` needs to be reassigned.

Kasterisk:
I don't think that's necessary, font sizes hardly change at runtime especially the way how we use it now. Besides the bbcode inline image support in Godot 3.4 is kind of limited, I have to workaround adding vertical alignment for example. If that's really needed I think it's best to wait for 4.0, in the meantime customizing the size in the locale file is enough.

On another note, I left the size of the spacebar key icon as is. The new size doesn't play well other keys but I'm open for suggestions.

src/gui_common/CustomRichTextLabel.cs
Kasterisk: Nevermind, I end up doing another approach for this.

hhyyrylainen:
> I don't think that's necessary, font sizes hardly change at runtime especially

Translations are a thing. If the key displayed properly in Finnish it would say "Välilyönti" on the button which is much longer.

Kasterisk:
But the texts on the key prompts themselves are part of the texture and don't change according to language?

hhyyrylainen:
> But the texts on the key prompts themselves are part of the texture and don't change according to language?

I was somewhat sure that was the case, so I was confused by the earlier talk about the font size...
What's the actual situation now?

Kasterisk:
Seems like there was confusion on what the text is supposed to be, I realized now you're talking about the texts _inside_ the key prompts while I thought of regular texts in between it, outside of the key prompts.

The situation was that, increasing the size of spacebar key prompt used in the pause prompt info to make it more readable turned out to be a not so optimal solution due to other keys not scaling up to how we want it because of differing aspect ratios.

This looks okay:


While this looks out of proportion:


And relative to the whole screen:


In the end I decided to not scale up the key prompt and leave it as is. So now this PR just add a way to change the size of inline images for the custom bbcodes in general.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -262,25 +262,54 @@ private static string ParseCustomTagsString(string extendedBbcode)
     /// <param name="input">The string enclosed by the custom tags</param>
     /// <param name="bbcode">Custom Thrive bbcode-styled tags</param>
     /// <param name="attributes">Attributes specifying additional functionalities to the bbcode.</param>
-    private static string BuildTemplateForTag(string input, ThriveBbCode bbcode, List<string> attributes)
+    private string BuildTemplateForTag(string input, ThriveBbCode bbcode, List<string> attributes)
```
hhyyrylainen, line 265: What was the change that means this can no longer be static?

Kasterisk: I've come to the conclusion that this being static is not necessary, it's private thus won't be used in any other places, secondly in the future it might benefit knowing its current rich text label instance to do operation/accessing information regarding them for use in some additional tags.

hhyyrylainen: This being static makes the code easier to reason about. This being static said "I won't touch the calling object" so you could just call that method and not have to think if it changed the current state of the object. This no longer being static increases the mental load on future programmers (and me who is trying to review this code).

Kasterisk: Alright then, static it is.

Oliveriver (commented):
I agree that it's better to have space small than other keys big. Considering this changed to a different feature, I don't think I have anything else to add.

PR #3388, opened 2022-06-08
Add GUI Control Highlighting for editor button available tutorial

### LINKED ISSUE #3104: Modify editor button available tutorial to use GUI control highlighting
so that the button is very obviously pointed out to the player and they can't ignore going to the editor.

I have slight concern that not in all cases the player likes being forced to not play the stage anymore and go to the editor (though perhaps it is unlikely that the player is fighting other cells before getting to the editor once).

andeason:

PR adds GUI control highlighting for the tutorial when the player has enough to enter the cell editor for the first time. 

**Related Issues**

Closes #3104 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/tutorial/microbe_stage/MicrobeTutorialGUI.cs
```diff
@@ -73,6 +76,7 @@ public class MicrobeTutorialGUI : Control, ITutorialGUI
 
     public Node GUINode => this;
 
+    public ControlHighlight? PressEditorButtonHighlight { get; private set; }
     public bool IsClosingAutomatically { get; set; }
```
Kasterisk, line 80: A small nitpick, add space between these two.

Kasterisk (approved):
Looks good and works fine.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -821,7 +821,10 @@ private void OnPlayerReproductionStatusChanged(Microbe player, bool ready)
         if (ready && (player.Colony == null || player.IsMulticellular))
         {
             if (!player.IsMulticellular)
+            {
+                HUD.SetupPressEditorTutorial(TutorialState);
```
hhyyrylainen, line 825: This should be done in the tutorial setup for the stage, not just before emitting the event, because that is not flexible, if the trigger condition is changed this code also needs to move so that's why this is not futureproof. This should be done similarly to how the undo/redo buttons are sent to the tutorial, which is when the editor GUI is initialized: https://github.com/andeason/Thrive/blob/070fa612f641efdd582130e639349dff8708c03a/src/microbe_stage/editor/MicrobeEditor.cs#L167

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -775,6 +775,11 @@ public void PauseButtonPressed()
         }
     }
 
+    public void SetupPressEditorTutorial(TutorialState tutorialState)
```
hhyyrylainen, line 778:
```suggestion
    public void SendEditorButtonToTutorial(TutorialState tutorialState)
```
To have more consistent naming.

hhyyrylainen (changes requested):
A few code comments that should be pretty easy to fix.
Additionally I noticed a bit of a bug. If the player dies while the editor button tutorial is active their screen will stay covered:

So to fix that I think the editor button tutorial should hide itself if it is visible and it receives `TutorialEventType.MicrobePlayerDied` event.

src/microbe_stage/MicrobeStage.cs
andeason: I put in the SetupStage after setting the eventreciever as that seemed like the best place for now.  I can attempt to do it in the HUD init, but from a quick glance seems that I would need to reorder the various operations to make it work for that.  Could be wrong, though.

hhyyrylainen (approved):
Code looks good now and I retested and didn't find any issues.

hhyyrylainen:
Turns out this had a bug slip through: https://github.com/Revolutionary-Games/Thrive/commit/fd36945c8fdc171d76fb2e8892a16b0c20f752ee

Issue #873, opened 2019-10-05
Add new game options


hhyyrylainen:
Add new game options when starting a game.
Should have at least planet selection.
And a LAWK switch (if thermoplast gets added).

This should hook into: https://github.com/Revolutionary-Games/Thrive/issues/872
Which should then affect the patch map generation.

The setup screen that allows to set the seed for the procedural patch map so that the seeds can be shared. Once this is done: https://github.com/Revolutionary-Games/Thrive/pull/3278

hhyyrylainen:
As for the graphics a basic "terran planet" icon is what I can think of this needing that we don't have yet. And maybe also a "copy to clipboard" icon.

hhyyrylainen:
Dev forum discussion: https://forum.revolutionarygamesstudio.com/t/implementing-new-game-settings/884

hhyyrylainen:
Split the graphics tasks into a separate issue: https://github.com/Revolutionary-Games/Thrive/issues/3425

Issue #575, opened 2018-05-31
Please implement modes like easy, normal and hard


AndreLZGava:
There are many enviroments when playing the game, it seems that many of them are harder to survive. It's too hard, sometimes resources like oxygen, amonian and glicogene aren't find soo quikly, and they are spent very fast.

crodnu:
Hi Andre!
We appreciate your suggestions, however this is not the place to submit them.
This place is used to divide work on our team on already thougth out features for the next release.
If you want to make suggestions, please join our [community forum](https://community.revolutionarygamesstudio.com) and open a new thread about it.

hhyyrylainen:
I'm willing to accept technical issues etc. here as github issues. But @crodnu is right that any suggestions etc. should be brought up somewhere else.

hhyyrylainen:
I'll open this because it has been discussed before that there should be difficulty options in the game. But I won't put this into any milestone yet. This also depends on a future issue for specifying the options for a new game.

AndreLZGava:
Ok, I created a account in the site and add it as a topic of game in the future. I hope you implements it in the game, would be great :)

PR #3385, opened 2022-06-07
Added more tooltips to game buttons

Kasterisk:
**Brief Description of What This PR Does**

Found that in old versions tooltips were quite abundant but some of them seems to be lost now. This brought back some of those tooltips, no reason not to have them now. Tooltips readded for: some of the menu buttons, pause menu buttons and editor button. 

Also tweaked the version number and store user tag label's font and size.

**Progress Checklist**

Note: before starting this checklist the PR should be marked as non-draft.

- [x] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [x] Initial code review passed (this and further items should not be checked by the PR author)
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author. Merging must follow our
[styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md#git).

PR #3389, opened 2022-06-08
New game settings

Oliveriver:

Adds a new game settings screen to the main menu, with basic and advanced views, and a multitude of options which affect in-game content.

See: https://forum.revolutionarygamesstudio.com/t/implementing-new-game-settings/884

**Related Issues**

Closes https://github.com/Revolutionary-Games/Thrive/issues/873
Closes https://github.com/Revolutionary-Games/Thrive/issues/575

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Oliveriver:
Note that this PR is now blocked by https://github.com/Revolutionary-Games/Thrive/pull/3278 and https://github.com/Revolutionary-Games/Thrive/pull/3235.

Kasterisk:
I haven't checked in this PR but this discussion https://github.com/Revolutionary-Games/Thrive/pull/3385#discussion_r890938068 relating to the new game button tooltip could be applied here if not yet.

Oliveriver:
> I haven't checked in this PR but this discussion [#3385 (comment)](https://github.com/Revolutionary-Games/Thrive/pull/3385#discussion_r890938068) relating to the new game button tooltip could be applied here if not yet.

Ah, I haven’t added tooltips yet. I’ll keep that in mind when I get to it.

Oliveriver:
I've decided that, with https://github.com/Revolutionary-Games/Thrive/pull/3235 done, this PR can go ahead without https://github.com/Revolutionary-Games/Thrive/pull/3278. I've therefore made the procedural/classic map selection invisible, with the intention of making it visible as part of the procedural patch map PR (same for Easter eggs).

So other than the spawn system weirdness I mentioned on Discord, I believe this PR is ready for review and testing. Testing should cover both checking the settings change what they say they change and deciding what values they should take for difficulty presets.

I doubt we'll get much effective feedback on the latter until we open this up to the community, seeing as we as devs have a skewed experience of the game's difficulty. So I'd like to request that we get this into a devbuild sooner rather than later (this is why I've changed course to not depend on the procedural patch map).

src/general/NewGameSettings.cs
```diff
@@ -0,0 +1,638 @@
...
+
+    private void OnAdvancedPressed()
+    {
+        ProcessAdvancedSelection(true);
+    }
+
+    private void ProcessAdvancedSelection(bool playSound)
+    {
+        if (playSound)
+            GUICommon.Instance.PlayButtonPressSound();
```
Kasterisk, line 348: The usual non-written way we do it is to only call the play button press sound from the callbacks, so that the actual logic containing methods doesn't have to know whether to play the sound or not.

src/general/NewGameSettings.cs
```diff
@@ -0,0 +1,638 @@
...
+        settings.FreeGlucoseCloud = freeGlucoseCloudButton.Pressed;
+
+        settings.MapType = MapTypeIndexToValue(mapTypeButton.Selected);
+
+        settings.IncludeMulticellular = includeMulticellularButton.Pressed;
+        settings.EasterEggs = easterEggsButton.Pressed;
+
+        var scene = GD.Load<PackedScene>("res://src/general/MainMenu.tscn");
+        var mainMenu = (MainMenu)scene.Instance();
+        mainMenu.NewGameSetupDone(settings);
```
Kasterisk, line 398: This doesn't seem very good, there's now duplicate main menu instances and since this one isn't added to the scene tree, it will be orphaned, potentially causing a memory leak. Instead, I think it's better to just move the logic in `MainMenu.NewGameSetupDone` to here. It's pertaining to the new game settings anyways.

src/general/NewGameSettings.cs
```diff
@@ -0,0 +1,638 @@
...
+        UpdateDifficultyPreset();
+    }
+
+    private void OnLifeOriginSelected(int index)
+    {
+        // Set both buttons here as we only received a signal from one of them
+        lifeOriginButton.Selected = index;
+        lifeOriginButtonAdvanced.Selected = index;
+
+        settings.Origin = LifeOriginIndexToValue(index);
```
Kasterisk, line 568: Wouldn't it be far more efficient to cast these indexes directly to enum?
```csharp
settings.Origin = (WorldGenerationSettings.LifeOrigin)index;
```

Kasterisk (dismissed):

Kasterisk:
Really like how the UI works. I left some reviews for things I think could be improved codewise.

src/general/WorldGenerationSettings.cs
```diff
@@ -1,6 +1,175 @@
...
+    public const double MIN_AI_MUTATION_RATE = 0.5;
+    public const double MAX_AI_MUTATION_RATE = 5;
+    public const double MIN_COMPOUND_DENSITY = 0.2;
+    public const double MAX_COMPOUND_DENSITY = 2;
+    public const double MIN_PLAYER_DEATH_POPULATION_PENALTY = 1;
+    public const double MAX_PLAYER_DEATH_POPULATION_PENALTY = 5;
+    public const double MIN_GLUCOSE_DECAY = 0.3;
+    public const double MAX_GLUCOSE_DECAY = 0.95;
+    public const double MIN_OSMOREGULATION_MULTIPLIER = 0.2;
+    public const double MAX_OSMOREGULATION_MULTIPLIER = 2;
```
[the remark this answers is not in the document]
Oliveriver: I thought it might be cleaner to keep these constants with the rest of the world generation settings, since they're exclusively applicable here. But if the consensus says otherwise, I'll move them.

src/general/NewGameSettings.cs
Oliveriver: Did not know this was possible. Will definitely change that.

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -8,17 +8,28 @@
 [SuppressMessage("ReSharper", "StringLiteralTypo", Justification = "Patch names aren't proper words")]
 public static class PatchMapGenerator
 {
+    public static WorldGenerationSettings WorldSettings = new();
```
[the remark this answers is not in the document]
Oliveriver: All the methods in this class are static, so this has to be static.

RacerBG (commented):
I'm still in the early stage of testing this PR so I will give my final verdict later. So far I'm very impressed with the new UI, it looks professional, nice work!

I would suggest having a tooltip over the random seed button though.

"Advanced view" and "Basic view" could be renamed in to "Basic" and "Advanced" but it's up to you.

I'm not so sure about the text below the multicellular stage because saving and loading should be working quite soon in my opinion.

I had tested the values from "Advanced view" and they seem to work just fine but I would like to spend a few hours on them tomorrow and then give my final review. :smiley:

src/general/WorldGenerationSettings.cs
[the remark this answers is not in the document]
hhyyrylainen: I think maybe in this case the huge dump of difficulty option ranges should maybe be in the separate file. At least if they were moved to constants they'd need much more descriptive names, and I imagine (I haven't looked through this PR yet) that only the world generation really should use these. So the design is a bit clearer if other places in the code aren't able to access these and instead need to read the values that the current game has stored.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -168,6 +168,11 @@ microbe editor.
     /// </summary>
     public bool Unique;
 
+    /// <summary>
+    ///   Determines whether this organelle appears in LAWK-only games
+    /// </summary>
+    public bool Lawk;
```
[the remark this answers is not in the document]
hhyyrylainen: I think naming this to LAWK would make more sense as it is not a normal word (need to also add to the jetbrains dotsettings file so that it doesn't say this is mispelled).

src/microbe_stage/PatchMapGenerator.cs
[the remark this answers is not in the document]
hhyyrylainen: Why are the settings static here? Shouldn't they be a generation run specific?

src/general/GameWorld.cs
```diff
@@ -17,6 +17,8 @@
 [UseThriveSerializer]
 public class GameWorld : ISaveLoadable
 {
+    public WorldGenerationSettings WorldSettings = new();
```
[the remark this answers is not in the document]
hhyyrylainen: Yeah I agree. Only one of them should have the variable. And well I think I can see the use of storing the world generation settings, but in my mind I was imagining that this object would *only* exist while generating the world. But then again later on in the game we'll also do world generation so I guess storing this somewhere with the current game data is the right design.

assets/concept_art/planet_generation.jpeg.import
```diff
@@ -0,0 +1,35 @@
+[remap]
```
hhyyrylainen, line 1: For uniformity I think this should be a .jpg file (we do have one .jpeg file in the game, but that's more of an oversight).

simulation_parameters/SimulationParameters.cs
```diff
@@ -222,34 +222,48 @@ public GameCredits GetCredits()
...
+    public OrganelleDefinition GetRandomProkaryoticOrganelle(Random random, bool lawkOnly)
     {
         float valueLeft = random.Next(0.0f, prokaryoticOrganellesTotalChance);
 
-        foreach (var organelle in prokaryoticOrganelles)
+        // Filter to only LAWK organelles if necessary
+        var usedOrganelles = prokaryoticOrganelles;
+        if (lawkOnly)
+        {
+            usedOrganelles = usedOrganelles.Where(o => o.Lawk).ToList();
```
hhyyrylainen, line 233: This ToList seems to unnecessarily create a temporary list here. If you change `var` in used organelle to `IEnumerable<OrganelleDefinition>` I think this code will compile without the ToList call here.

simulation_parameters/SimulationParameters.cs
```diff
@@ -222,34 +222,48 @@ public GameCredits GetCredits()
...
+    public OrganelleDefinition GetRandomEukaryoticOrganelle(Random random, bool lawkOnly)
     {
         float valueLeft = random.Next(0.0f, eukaryoticOrganellesChance);
 
-        foreach (var organelle in eukaryoticOrganelles)
+        // Filter to only LAWK organelles if necessary
+        var usedOrganelles = eukaryoticOrganelles;
+        if (lawkOnly)
+        {
+            usedOrganelles = usedOrganelles.Where(o => o.Lawk).ToList();
```
hhyyrylainen, line 255: Same here

src/auto-evo/steps/FindBestMutation.cs
```diff
@@ -17,12 +17,16 @@ public class FindBestMutation : VariantTryingStep
 
         private readonly Mutations mutations = new();
 
-        public FindBestMutation(AutoEvoConfiguration configuration, PatchMap map, Species species,
+        private WorldGenerationSettings settings;
+
+        public FindBestMutation(AutoEvoConfiguration configuration,
+            WorldGenerationSettings settings, PatchMap map, Species species,
```
hhyyrylainen, line 23: I think here the use of WorldGenerationSettings is pretty unclear why it is needed here. I think the design would be clearer if you create some interface to store the difficulty data or the world configuration that is needed here. Something like IWorldConfiguration or IDifficultyOption or something like that, or maybe even passing in the current game here?

src/microbe_stage/PatchMapGenerator.cs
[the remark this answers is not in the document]
Oliveriver: As I said, everything in this class is static for some reason, so an instance variable wouldn’t work.

Although now that I think about it, there’s probably no need to store them as a member variable here since they’re passed in every time. I’ll change that to resolve this.

src/auto-evo/steps/FindBestMutation.cs
Oliveriver: Passing in the current game sounds simplest to me.

src/microbe_stage/PatchMapGenerator.cs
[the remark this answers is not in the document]
hhyyrylainen: If the design starts to look like each method needs to pass the current settings all around, I think converting this to a non-static class would start to make a lot of sense so that the settings can be passed just once to the constructor.

src/general/WorldGenerationSettings.cs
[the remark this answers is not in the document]
hhyyrylainen: I think they might make sense here (I haven't read the code for this PR yet), another place might be the new game GUI, since these look like to be the intervals the new game options need to be within. That way it would be super clear that only that new game GUI should use this (especially if these could be made private in there), and all other places in the code need to access the actual current game variables.

simulation_parameters/SimulationParameters.cs
Oliveriver: Attempting to do this is giving me a PossibleMultipleEnumeration error during style checking. Any suggestions or should I keep it as is?

simulation_parameters/SimulationParameters.cs
```diff
@@ -222,34 +222,48 @@ public GameCredits GetCredits()
...
     {
         float valueLeft = random.Next(0.0f, prokaryoticOrganellesTotalChance);
 
-        foreach (var organelle in prokaryoticOrganelles)
+        // Filter to only LAWK organelles if necessary
+        IEnumerable<OrganelleDefinition> usedOrganelles = prokaryoticOrganelles;
+        if (lawkOnly)
+            usedOrganelles = usedOrganelles.Where(o => o.Lawk);
+
+        OrganelleDefinition chosenOrganelle = null!;
```
hhyyrylainen, line 234: Don't do nullability suppression like this. It is only meant for constructor-like resources (like Godot _Ready). In this case ~~I think you can just leave this variable uninitialized, or~~ just assign null (you need to make the variable nullable first `OrganelleDefinition?`). And then after the loop if the variable is null throw an exception as it's invalid that there wasn't any organelles.

src/auto-evo/steps/FindBestMutation.cs
```diff
@@ -17,12 +17,16 @@ public class FindBestMutation : VariantTryingStep
 
         private readonly Mutations mutations = new();
 
-        public FindBestMutation(AutoEvoConfiguration configuration, PatchMap map, Species species,
+        private GameWorld gameWorld;
+
+        public FindBestMutation(AutoEvoConfiguration configuration,
```
hhyyrylainen, line 22: It seems that the game world is used pretty often so maybe AutoEvoConfiguration should have the world as a mandatory constructor parameter so it'd be available everywhere in auto-evo or maybe at least the world settings should be in there. As only needing world settings will be easier to use for auto-evo prediction and the auto-evo exploring tool, I imagine, when that is done.

src/general/MainMenu.cs
```diff
@@ -68,15 +66,19 @@ public class MainMenu : NodeWithInput
...
-
-    private Button newGameButton = null!;
     private Button freebuildButton = null!;
 
     private Label storeLoggedInDisplay = null!;
 
     private CustomConfirmationDialog gles2Popup = null!;
     private ErrorDialog modLoadFailures = null!;
 
+    public static void OnEnteringGame()
```
hhyyrylainen, line 76: I'd make this a non-static method to be able to put this later in the class where the other new game methods are, to not have this be the first piece of code someone sees when they try to read the main menu file.

hhyyrylainen: Okay, I read now this entire file... I guess it's not worth trying to think of a different design for this as this is needed to be called from the new game menu as there's other stuff to fix in this PR...

src/general/Mutations.cs
```diff
@@ -164,7 +165,7 @@ private void MutateBehaviour(Species parent, Species mutated)
     ///   Creates a mutated version of parentOrganelles in organelles
     /// </summary>
     private void MutateMicrobeOrganelles(OrganelleLayout<OrganelleTemplate> parentOrganelles,
-        OrganelleLayout<OrganelleTemplate> mutatedOrganelles, bool isBacteria)
+        OrganelleLayout<OrganelleTemplate> mutatedOrganelles, bool isBacteria, double creationRate, bool lawkOnly)
```
hhyyrylainen, line 168: Why is the creationRate a double here?

src/general/WorldGenerationSettings.cs
```diff
@@ -1,6 +1,158 @@
...
+
+    public enum PatchMapType
+    {
+        Procedural,
+        Classic,
+    }
+
+    /*
+    Values for this particular object
+    */
```
hhyyrylainen, line 23:
```suggestion
    // Values for this particular object
```
This is the comment style used in the code, C-style comments are not used.

Oliveriver: I used this style because it turned up a few times here, so maybe this should be changed too: https://github.com/Revolutionary-Games/Thrive/blob/master/src/general/OptionsMenu.cs

src/general/WorldGenerationSettings.cs
```diff
@@ -1,6 +1,158 @@
...
+
+    public bool Lawk { get; set; }
```
hhyyrylainen, line 25: Most of these are pretty self explanatory, but could you add a short summary for each? These are pretty widely affecting things so similarly to the game settings object, I think all of these properties should be documented.

Also I think this field should be named LAWK.

Oliveriver: Code styling had a problem with the name LAWK so I changed it to this.

src/general/WorldGenerationSettings.cs
hhyyrylainen: Opened a PR: https://github.com/Revolutionary-Games/Thrive/pull/3422

src/general/WorldGenerationSettings.cs
hhyyrylainen: You need to add the LAWK as an abbreviation so that it knows that it is a single word and stops complaining.

hhyyrylainen: Thrive.sln.DotSettings around line 460

src/general/WorldGenerationSettings.cs
```diff
@@ -1,6 +1,158 @@
...
+    public bool FreeGlucoseCloud { get; set; }
+    public PatchMapType MapType { get; set; } = PatchMapType.Procedural;
+    public bool IncludeMulticellular { get; set; } = true;
+    public bool EasterEggs { get; set; } = true;
+
+    /*
+    Static values for each difficulty preset
+    */
+
+    public static float GetMPMultiplier(DifficultyPreset preset)
```
hhyyrylainen, line 44: I think it would be a cleaner design if there was a Difficulty class which already had all of the right properties initialized for the difficulty level, that way these constant lookups would not be needed, and those difficulty presets could be loaded from a JSON file for easy tweaking by @buckly90 for balancing.

src/auto-evo/steps/FindBestMutation.cs
hhyyrylainen: You are right, I was thinking of `SimulationConfiguration` which doesn't seem to be passed as a parameter here so the data can't be added there to easily use here.

RacerBG (approved):
Alright, part II of my review and my approval of course:

I don't have a strong opinion about the values affected by the difficulty they are looking fine in my opinion but you are free to change them as you like. :) The overall experience is pleasant, my earlier comments were mostly adressed and I can't find any game breaking bugs either.

src/auto-evo/steps/IncreaseBiodiversity.cs
```diff
@@ -20,8 +20,12 @@ public class IncreaseBiodiversity : IRunStep
         private bool tryCurrentPatch = true;
         private bool createdASpecies;
 
-        public IncreaseBiodiversity(AutoEvoConfiguration configuration, PatchMap map, Patch patch, Random random)
+        private WorldGenerationSettings settings;
+
+        public IncreaseBiodiversity(AutoEvoConfiguration configuration,
+            WorldGenerationSettings settings, PatchMap map, Patch patch, Random random)
```
hhyyrylainen, line 26: I'd suggest naming `settings` `worldSettings` here so that there aren't two very similarly named things, configuration and settings.

simulation_parameters/SimulationParameters.cs
```diff
@@ -222,34 +226,69 @@ public GameCredits GetCredits()
...
 
-    public OrganelleDefinition GetRandomProkaryoticOrganelle(Random random)
+    public DifficultyPreset GetDifficultyPreset(string name)
+    {
+        return difficultyPresets[name];
+    }
+
+    public DifficultyPreset GetDifficultyPresetByIndex(int index)
+    {
+        return difficultyPresets.Values.Where(p => p.Index == index).First();
```
hhyyrylainen, line 236:
```suggestion
        return difficultyPresets.Values.First(p => p.Index == index);
```
I think that's valid LINQ.

src/general/DifficultyPreset.cs
```diff
@@ -0,0 +1,130 @@
+﻿using System;
+using Newtonsoft.Json;
+
+/// <summary>
+///   Definition for a difficulty preset
+/// </summary>
+/// <remarks>
+///   <para>
+///     Values for each difficulty preset, as given in difficulty_preset.json
```
hhyyrylainen, line 9:
```suggestion
///     Values for each difficulty preset, as given in difficulty_presets.json
```
typo

src/microbe_stage/MicrobeStage.cs
```diff
@@ -117,6 +119,14 @@ public class MicrobeStage : NodeWithInput, IReturnableGameState, IGodotEarlyNode
     [JsonIgnore]
     public PlayerHoverInfo HoverInfo { get; private set; } = null!;
 
+    [JsonProperty]
+    public WorldGenerationSettings WorldSettings
```
hhyyrylainen, line 123: Shouldn't this be json ignore and just have a getter for getting the object from the gameworld? Similarly to how the gameworld is just a helper property to read it from the current game.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -300,6 +310,8 @@ public void SetupStage()
             }
         }
 
+        GD.Print(WorldSettings);
```
hhyyrylainen, line 313: Doesn't this print a ton of info? I think maybe always logging this isn't the best to keep log clutter down a bit...

src/microbe_stage/MicrobeStage.cs
```diff
@@ -343,7 +355,7 @@ public void OnFinishLoading()
 
     public void StartNewGame()
     {
-        CurrentGame = GameProperties.StartNewMicrobeGame();
+        CurrentGame = GameProperties.StartNewMicrobeGame(WorldSettings);
```
hhyyrylainen, line 358: This should just create a new world settings as this is called (I think) only when playing the scene directly from Godot.

src/microbe_stage/MicrobeStage.cs
Oliveriver: I wouldn't say it logs a lot, only three lines.

image

I definitely think this needs to be logged somewhere, as otherwise we wouldn't have a record of the player's world settings if we just receive their logs. But if you'd prefer it to be logged somewhere different than main output, I can change that.

src/microbe_stage/ProcessSystem.cs
```diff
@@ -134,6 +139,12 @@ public ProcessSystem(Node worldRoot)
         result.Osmoregulation = Constants.ATP_COST_FOR_OSMOREGULATION * hexCount *
             membrane.OsmoregulationFactor;
 
+        // World settings may be null if this is the first generation
```
hhyyrylainen, line 142: Would it be possible to fix the code structure so that the process system is not given invalid data? I'd much prefer a "proper" design where this workaround here could be avoided.

src/microbe_stage/MicrobeStage.cs
hhyyrylainen: Well we don't log the players overall settings either... But fine, let's have the settings in the output for now, I can always remove the registry load prints to reduce the amount of logs a debugging session prints.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -285,6 +285,31 @@ private void UpdateOrganelleUnlockTooltips()
         }
     }
 
+    private void UpdateOrganelleLAWKSettings()
+    {
+        // Don't use placeablePartSelectionElements as the thermoplast isn't placeable yet but is LAWK-dependent
+        foreach (var organelle in allPartSelectionElements.Keys)
+        {
+            var control = allPartSelectionElements[organelle];
```
hhyyrylainen, line 293: This kind of loop isn't very uniform with style elsewhere.
Could you reformat this and the next few loops to be like:
```
foreach (var entry in allPartSelectionElements)
...
```
or `foreach (var item in allPartSelectionElements.Items)`?

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -285,6 +285,31 @@ private void UpdateOrganelleUnlockTooltips()
...
+            control.Visible = !Editor.CurrentGame.GameWorld.WorldSettings.LAWK || organelle.LAWK;
+        }
+    }
+
+    private void UpdateDifficultyAdjustedMPCost()
```
hhyyrylainen, line 298: What's the difference to `UpdateTooltipMPCostFactors` seems like there's now two methods that basically do the same thing... I think these should be combined to reduce (my) confusion.

src/general/NewGameSettings.cs
```diff
@@ -0,0 +1,676 @@
...
+    [Export]
+    public NodePath IncludeMulticellularButtonPath = null!;
+
+    [Export]
+    public NodePath EasterEggsButtonPath = null!;
+
+    [Export]
+    public NodePath ConfirmButtonPath = null!;
+
+    private PanelContainer basicOptions = null!;
```
hhyyrylainen, line 116: This is a bit long one long block of variables. Usually our scenes have "logical" grouping (for example based on the subtab) for variables like this to make it easier to read and find related variables.

src/microbe_stage/MicrobeStage.cs
Oliveriver: The issue then is adding the world settings to the MicrobeStage object, which I did in NewGameSettings.OnConfirmPressed(). As it stands, trying to set the WorldSettings variable from GameWorld in the same place throws an error since CurrentGame is null at that point.

src/general/NewGameSettings.cs
```diff
@@ -0,0 +1,676 @@
...
+        glucoseDecayRate.MaxValue = Constants.MAX_GLUCOSE_DECAY * 100;
+        osmoregulationMultiplier.MinValue = Constants.MIN_OSMOREGULATION_MULTIPLIER;
+        osmoregulationMultiplier.MaxValue = Constants.MAX_OSMOREGULATION_MULTIPLIER;
+
+        settings = new WorldGenerationSettings();
+        difficultyPresets = SimulationParameters.Instance.GetAllDifficultyPresets();
+        normal = SimulationParameters.Instance.GetDifficultyPreset("normal");
+        custom = SimulationParameters.Instance.GetDifficultyPreset("custom");
+
+        foreach (DifficultyPreset preset in difficultyPresets)
```
hhyyrylainen, line 231: I think a safety OrderBy(p => p.Index) should be added. Or does the simulation parameters already sort and ensure no duplicate index values exist?

src/general/NewGameSettings.cs
```diff
@@ -0,0 +1,676 @@
...
+        compoundDensity.Value = preset.CompoundDensity;
+        playerDeathPopulationPenalty.Value = preset.PlayerDeathPopulationPenalty;
+        glucoseDecayRate.Value = preset.GlucoseDecay * 100;
+        osmoregulationMultiplier.Value = preset.OsmoregulationMultiplier;
+        freeGlucoseCloudButton.Pressed = preset.FreeGlucoseCloud;
+
+        UpdateDifficultyPreset();
+    }
+
+    private void UpdateDifficultyPreset()
```
hhyyrylainen, line 490:
```suggestion
    private void UpdateSelectedDifficultyPresetControl()
```
I was a bit confused what this method did at once, so I suggest a bit more descriptive name.

src/general/NewGameSettings.cs
```diff
@@ -0,0 +1,676 @@
...
+        settings.MapType = MapTypeIndexToValue(index);
+    }
+
+    private WorldGenerationSettings.PatchMapType MapTypeIndexToValue(int index)
+    {
+        switch (index)
+        {
+            case 1:
+                return WorldGenerationSettings.PatchMapType.Classic;
+            default:
```
hhyyrylainen, line 614: I think the default case should print an error (but then return procedural), so I'd like to see that and `case 0` added.

src/general/NewGameSettings.cs
```diff
@@ -0,0 +1,676 @@
...
+        if (lawk && lifeOriginButton.Selected == panspermiaIndex)
+        {
+            lifeOriginButton.Selected = (int)WorldGenerationSettings.LifeOrigin.Vent;
+            lifeOriginButtonAdvanced.Selected = (int)WorldGenerationSettings.LifeOrigin.Vent;
+        }
+    }
+
+    private void OnGameSeedChangedFromBasic(string text)
+    {
+        gameSeedAdvanced.Text = text;
```
hhyyrylainen, line 647: Can't this and OnGameSeedChangedFromAdvanced be combined? Same as most of the other callbacks or is there a reason why this is this way?

src/general/NewGameSettings.cs
```diff
@@ -0,0 +1,676 @@
...
+            difficultyPresetAdvancedButton.AddItem(preset.Name);
+        }
+
+        // Do this in case default values in NewGameSettings.tscn don't match the normal preset
+        InitialiseToPreset(normal);
+
+        var seed = GenerateNewRandomSeed();
+        gameSeed.Text = seed;
+        gameSeedAdvanced.Text = seed;
+        SetSeed(seed);
```
hhyyrylainen, line 243: Why doesn't SetSeed update gameSeed and gameSeedAdvanced automatically (assuming the given new seed is valid)?

hhyyrylainen:
I've now finished reviewing all the changes in this PR, so those were all of my findings.

src/microbe_stage/MicrobeStage.cs
Oliveriver: CurrentGame is not initialised at the time we exit the game settings screen. It is only added to MicrobeStage in SetupStage(), which is called after we switch to the scene. So the key problem is that we can't just pass the world settings to CurrentGame at this point.

I tried passing the settings to the new game _after_ switching the scene, but then many things (starting patch, cloud densities, etc.) initialised without the correct world settings.

I agree that GameWorld should be the source of truth for the world settings, but I can't figure out a way of initialising everything correctly without passing a WGS object to MicrobeStage itself. So I've changed this variable to InitialWorldSettings, used only for this initial setup. Everything else should use the WorldSettings object in GameWorld.

src/general/NewGameSettings.cs
Oliveriver: Ah, after making this change I've realised why I couldn't do that. Since a combined method sets the text of both text inputs, it would in particular overwrite the text the user has just entered. Although it overwrites to the correct value, it places the caret back at the start of the box and won't let you change the caret position. So unless you have a suggestion for that, I'll revert to the way it was.

src/microbe_stage/MicrobeStage.cs
Oliveriver: I discovered while testing InitialWorldSettings that we do need to pass the settings object here, even in the normal game mode. So I've kept it.

hhyyrylainen: I think this is related to: https://github.com/Revolutionary-Games/Thrive/pull/3389#discussion_r902677928

src/microbe_stage/MicrobeStage.cs
hhyyrylainen: I couldn't help myself so I went ahead and programmed and committed this kind of change. Also added 2 more words to the dictionary to suppress some warnings.

hhyyrylainen:
I just playtested this for the first time, and I really think the "basic view" and "advanced view" should just be named "basic" and "advanced", they are clear enough from the context and I think adding the "view" to the end sounds worse.

There is some kind of weird flickering going on when hovering with the mouse on the random seed where the cursor flickers between the caret and normal. I've not noticed this happening with any other text boxes in the game.

Also made another commit to fix the place that allowed null world settings and based on that I found that some parts of auto-evo didn't take the world settings into account, so I fixed those places as well.

Besides that "basic view" and the things I fixed myself, I didn't notice problems.

Oliveriver:
Alright, thanks for that. I was working on those changes myself and had pretty much exactly the same changes to all the Auto-Evo code, but I'll defer to your commit.

I can't reproduce the flickering issue. Do you think it's a big enough problem to investigate further?

hhyyrylainen:
The flickering might be platform specific (linux only) or might only happen in windowed mode (if you are playing in fullscreen). If it can't be solved, I'll open an issue for fixing it later.

Oliveriver:
I've been playing in fullscreen, but having just switched to windowed mode, I still can't see an issue. So maybe it is Linux-specific.

hhyyrylainen (approved):
As far as I recall, this is now fully done. I'll merge this tomorrow morning unless someone has anything still to say.

PR #3390, opened 2022-06-09
Ai mutation setting

adQuid:
A non-master PR. 

I was so interested in my own idea https://forum.revolutionarygamesstudio.com/t/implementing-new-game-settings/884/14 that I decided to make a "yes and" PR into Oliveriver's branch. I have confirmed it works, but I still don't have the working ruby environment to add the labels, so if you're interested in incorporating this @Oliveriver you'll have to add those yourself.


Oliveriver:
Neat, I was hoping someone would do something like this. I can confirm the mutation rate slider is having a noticeable effect, and I'm actually in favour of pushing up the default values for everything as it felt more interesting to have more developed cells around.

I've added labels with localisation and fixed a few minor issues. Let me know if you have anything to add, otherwise I'll merge this into my branch later.

Oliveriver (approved):

PR #3391, opened 2022-06-11
Slider history

### LINKED ISSUE #1496: Combine subsequent rigidity changes in the editor to a single action
to make undoing easier

### LINKED ISSUE #3346: Combine subsequent behaviour slider changes into a single action
so that undoing is easier.
Probably need to tweak the general action part as when new actions are added GetInterferenceModeWithGuaranteed doesn't get called so there's currently no logic to combine the just added action into the latest action in the editor history, which is the way I think this issue should be solved.

This is basically the same as: https://github.com/Revolutionary-Games/Thrive/issues/1496 but for a different set of sliders

84634E1A607A:

Improved sliders handling. 

**Related Issues**
Fixes #1496 
Fixes #3346 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

Works as intended, but I'm not quite sure whether merge should only be applied to SingleEditorAction, and if all SingleEditorActions may be merged.

And I am thinking of adding a Mergeable choice in the Interference enum to indicate that the two actions can be merged.

### DIFF
--- a/src/general/CombinableActionData.cs
+++ b/src/general/CombinableActionData.cs
@@ -14,6 +14,11 @@ public abstract class CombinableActionData
     public virtual bool ResetsHistory => false;
 
     /// <summary>
+    ///   Should this action be merged with the previous one if possible?
+    /// </summary>
+    public virtual bool WantsMerge => false;
+
+    /// <summary>
     ///   How does this action interfere with the <paramref name="other"/> action?
     /// </summary>
     /// <returns>
@@ -43,6 +48,21 @@ public abstract class CombinableActionData
         return CombineGuaranteed(other);
     }
 
+    /// <summary>
+    ///   Merge the other data into this if possible.
+    /// </summary>
+    /// <param name="other">The action to merge into this</param>
+    /// <returns>If a merge has been conducted</returns>
+    public virtual bool TryMerge(CombinableActionData other)
+    {
+        var mode = GetInterferenceModeWith(other);
+        if (mode != ActionInterferenceMode.Combinable && mode != ActionInterferenceMode.CancelsOut)
+            return false;
+
+        MergeGuaranteed(other);
+        return true;
+    }
+
     protected abstract ActionInterferenceMode GetInterferenceModeWithGuaranteed(CombinableActionData other);
 
     /// <summary>
@@ -51,4 +71,13 @@ public abstract class CombinableActionData
     /// <param name="other">The action this should be combined with. Guaranteed to be combinable</param>
     /// <returns>Returns the combined action</returns>
     protected abstract CombinableActionData CombineGuaranteed(CombinableActionData other);
+
+    /// <summary>
+    ///   Merges the other data into this
+    /// </summary>
+    /// <param name="other">The action to merge into this. Guaranteed to be mergeable</param>
+    protected virtual void MergeGuaranteed(CombinableActionData other)
+    {
+        throw new NotSupportedException();
+    }
 }
--- a/src/general/CombinableActionData.cs
+++ b/src/general/CombinableActionData.cs
@@ -46,13 +46,14 @@ public abstract class CombinableActionData
     /// <summary>
     ///   Should this action be merged with the previous one if possible?
     /// </summary>
+    /// <returns>True if this action wants to be merged with <see cref="other"/></returns>
     public virtual bool WantsMergeWith(CombinableActionData other)
     {
         return false;
     }
 
     /// <summary>
-    ///   Merge the other data into this if possible.
+    ///   Merge the other data into this if possible
     /// </summary>
     /// <param name="other">The action to merge into this</param>
     /// <returns>True if a merge has been conducted</returns>
@@ -75,9 +76,12 @@ public abstract class CombinableActionData
     protected abstract CombinableActionData CombineGuaranteed(CombinableActionData other);
 
     /// <summary>
-    ///   Merges the other data into this
+    ///   Merges the other data into this action
     /// </summary>
-    /// <param name="other">The action to merge into this. Guaranteed to be mergeable</param>
+    /// <param name="other">
+    ///   The action to merge into this. Guaranteed to be mergeable by the called.
+    ///   <see cref="GetInterferenceModeWith"/> has been called to make sure this can be combined.
+    /// </param>
     protected virtual void MergeGuaranteed(CombinableActionData other)
     {
         throw new NotSupportedException();
--- a/src/general/CombinableActionData.cs
+++ b/src/general/CombinableActionData.cs
@@ -56,13 +56,17 @@ public abstract class CombinableActionData
     ///   Merge the other data into this if possible
     /// </summary>
     /// <param name="other">The action to merge into this</param>
+    /// <param name="otherIsNewer">
+    ///   If the other data is newer than the current one, use only when the two cancels out, to make sure it goes right
+    /// </param>
     /// <returns>True if a merge has been conducted</returns>
-    public virtual bool TryMerge(CombinableActionData other)
+    public virtual bool TryMerge(CombinableActionData other, bool? otherIsNewer = null)
     {
-        if (GetInterferenceModeWith(other) != ActionInterferenceMode.Combinable)
+        var mode = GetInterferenceModeWith(other);
+        if (mode != ActionInterferenceMode.Combinable && mode != ActionInterferenceMode.CancelsOut)
             return false;
 
-        MergeGuaranteed(other);
+        MergeGuaranteed(other, otherIsNewer);
         return true;
     }
 
@@ -82,7 +86,8 @@ public abstract class CombinableActionData
     ///   The action to merge into this. Guaranteed to be mergeable by the called.
     ///   <see cref="GetInterferenceModeWith"/> has been called to make sure this can be combined.
     /// </param>
-    protected virtual void MergeGuaranteed(CombinableActionData other)
+    /// <param name="otherIsNewer">If the other data is newer than the current one</param>
+    protected virtual void MergeGuaranteed(CombinableActionData other, bool? otherIsNewer = null)
     {
         throw new NotSupportedException();
     }
--- a/src/general/CombinableActionData.cs
+++ b/src/general/CombinableActionData.cs
@@ -14,11 +14,6 @@ public abstract class CombinableActionData
     public virtual bool ResetsHistory => false;
 
     /// <summary>
-    ///   Should this action be merged with the previous one if possible?
-    /// </summary>
-    public virtual bool WantsMerge => false;
-
-    /// <summary>
     ///   How does this action interfere with the <paramref name="other"/> action?
     /// </summary>
     /// <returns>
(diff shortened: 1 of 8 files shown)


84634E1A607A:
Needs ADVICE regarding code resign.

hhyyrylainen:
I think you are going the wrong direction with the new `MergeGuaranteed` as I said before I think the existing combine functionality should be leveraged which causes new data instances to be created. Those instances can be used to replace the data in history. I think that way there is a minimal need for new code. Just an extra method to ask the data class if it wants to replace the other in undo history (we want quite many actions to not combine).

84634E1A607A:
However, only **sequent** slider action should be merged. If I program the code like what you said, I wonder if I can get CombineGuaranteed know whether the two actions are sequent without messing up with the other logic.

84634E1A607A:
And, I don't get it how *to replace the data in history*, because you can't access `SingleEditorAction<T>`'s `SingleData` property without enumerating all `<T>`s; and if you just modify the `ActionData` itself, it makes no sense to create a redundant one.

hhyyrylainen:
> However, only sequent slider action should be merged. If I program the code like what you said, I wonder if I can get CombineGuaranteed know whether the two actions are sequent without messing up with the other logic.

That's why I suggested putting the logic in the "add action to history" method. That way it can check *just* the previous action with the new one if they want to combine or not. Even in that case things like the player just placing something and deleting it immediately *should not* combine to not mess with the edit history. That's why I suggested adding a new method for the data to tell the system that it wants to combine.

> And, I don't get it how to replace the data in history, because you can't access SingleEditorAction<T>'s SingleData property without enumerating all <T>s; and if you just modify the ActionData itself, it makes no sense to create a redundant one.

Early multicell editor code:
```c#
        // Combine the topmost action in the stack with this new one to make sure finishing editing a cell doesn't
        // cause a separate step
        var action = new CombinedEditorAction(history.PopTopAction(),
            new SingleEditorAction<EndCellTypeEditActionData>(DoEndCellTypeEditAction, UndoEndCellTypeEditAction,
                new EndCellTypeEditActionData(selectedCellTypeToEdit)));

        // We need to do this here to free up the MP that is now in the undone action, otherwise it won't succeed in
        // all cases
        DirtyMutationPointsCache();

        if (!EnqueueAction(action))
            GD.PrintErr("Combined action, with 0 cost added cell edit finish, could not be performed again");
```

In this case the new action should be created something like (probably needs slight adjusting):
```c#
if(actions should combine){
var action = history.PopTopAction();
var combined = action.CombineGuaranteed(newAction);
Add(action);
return;
}
normal action add logic goes here
```

This is basically as much as I can say without going and literally coding this entire thing myself.

84634E1A607A:
> In this case the new action should be created something like (probably needs slight adjusting):

That **is** my first thought. But `CombineGuaranteed` returns a new `ActionData` instead of a new `Action`. As I mentioned above, to create an `Action` (`SingleEditorAction`), you need to know exactly what kind of ActionData `<T>` is.

hhyyrylainen:
I think it wouldn't be super wrong to just make it into a combined action, then? Or perhaps would be cleaner design to have the ActionData have a method that creates the single action out of it for this case (and types it is supported for)?

84634E1A607A:
> I think it wouldn't be super wrong to just make it into a combined action, then?

But it makes little sense, as it should be a single action. I'm actually thinking that we can make `SingleEditorAction<T>` `SingleEditorAction`, as `<T>` is actually not that necessary.

> Or perhaps would be cleaner design to have the ActionData have a method that creates the single action out of it for this case (and types it is supported for)?

An `ActionData` doesn't know how to Undo and Redo itself.

hhyyrylainen:
> An ActionData doesn't know how to Undo and Redo itself.

Okay so modify SingleEditorAction base class to have a method for extending data into it, which could have all of the checks to make sure the other action has compatible data and merge it then, and it could return a bool to indicate if it merged the action or not.

84634E1A607A:
> Okay so modify SingleEditorAction base class to have a method

If you want to call a `SingleEditorAction<T>` method, you must know `<T>` .................

I'll try to convert `SingleEditorAction<T>` to separate classes `SingleEditorAction` and `SingleEditorAction<T>` (like `GetNode()` and `GetNode<T>()`)

hhyyrylainen:
That's why I said to extend the *base* class of the single editor action so you can call it without having to cast to any kind of SingleEditorAction.

84634E1A607A:
This is exactly what I had done (?) I had extended base class 'action history' and directly call Merge (?)

hhyyrylainen:
Okay, I see it's mostly there, though two points:
- Merge should return a boolean and not throw for an invalid operation. And it should also be overloaded per action type so that that place and remove sequence is not merged, but multiple slider actions in a row are
- And AddAction needs work, especially that type specific logic should be thrown out, that's not proper OOP design (see my above point for how it should be)

src/general/CombinableActionData.cs
```diff
@@ -43,6 +48,21 @@ public virtual CombinableActionData Combine(CombinableActionData other)
         return CombineGuaranteed(other);
     }
 
+    /// <summary>
+    ///   Merge the other data into this if possible.
+    /// </summary>
+    /// <param name="other">The action to merge into this</param>
+    /// <returns>If a merge has been conducted</returns>
```
hhyyrylainen, line 55:
```suggestion
    /// <returns>True if a merge has been conducted</returns>
```

src/general/CombinableActionData.cs
```diff
@@ -43,6 +48,21 @@ public virtual CombinableActionData Combine(CombinableActionData other)
...
+    public virtual bool TryMerge(CombinableActionData other)
+    {
+        var mode = GetInterferenceModeWith(other);
+        if (mode != ActionInterferenceMode.Combinable && mode != ActionInterferenceMode.CancelsOut)
```
hhyyrylainen, line 59: I think this should instead check if the mode is combinable or not rather than this way because this logic will bite us if new action interference modes are ever added.

src/general/CombinableActionData.cs
```diff
@@ -13,6 +13,11 @@ public abstract class CombinableActionData
     /// </summary>
     public virtual bool ResetsHistory => false;
 
+    /// <summary>
+    ///   Should this action be merged with the previous one if possible?
+    /// </summary>
+    public virtual bool WantsMerge => false;
```
hhyyrylainen, line 19: I think this should be a method like:
```c#
public virtual bool WantsMergeWith(CombinadleActionData other){
return false;
}
```
then child classes that support merging can return true if the other object is of the right type. This will give much more flexibility to the design which I foresee the chance of needing.

src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -138,8 +138,16 @@ public override bool Undo()
 
     public override void AddAction(TAction action)
     {
-        // TODO: check if the action can be combined (for example behaviour or rigidity slider subsequent edits should
+        // Check if the action can be combined (for example behaviour or rigidity slider subsequent edits should
         // combine) in a single step for undo
+        if (ActionIndex > 0 && action.Data.FirstOrDefault()!.WantsMerge)
```
hhyyrylainen, line 143:
```suggestion
        if (ActionIndex > 0 && action.Data.First().WantsMergeWith(action.Data)
```

Also I'm slightly suspicious if you got the ActionIndex correct. Is it the case that we have the action index pointing *after* the latest performed action? Just to remind me. If that's the case then the logic is right here.

src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -138,8 +138,16 @@ public override bool Undo()
...
+        {
+            EditorCombinableActionData currentActionData = action.Data.FirstOrDefault()!;
+            EditorCombinableActionData lastActionData = Actions[ActionIndex - 1].Data.FirstOrDefault()!;
```
hhyyrylainen, line 146:
```suggestion
            EditorCombinableActionData currentActionData = action.Data.First();
            EditorCombinableActionData lastActionData = Actions[ActionIndex - 1].Data.First();
```

First returns a non-nullable value (or throws) so it is much better to be explicit about that than suppress nullability of the version that can return null (which would then hit a much less helpful null pointer exception).

src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -138,8 +138,16 @@ public override bool Undo()
...
+
+            if (currentActionData.TryMerge(lastActionData))
+                PopTopAction();
```
hhyyrylainen, line 149: Wouldn't it be easier to reason about this if this was flipped. So that the already performed action is the one that gets merged with the subsequent action? This is because that's the order in which MP computation does it, so if this does it the same way around, it'll be less error prone for future programmers to work on this.

src/microbe_stage/editor/action_data/BehaviourActionData.cs
```diff
@@ -46,4 +48,16 @@ protected override CombinableActionData CombineGuaranteed(CombinableActionData o
 
         return new BehaviourActionData(behaviourChangeActionData.NewValue, OldValue, Type);
     }
+
+    protected override void MergeGuaranteed(CombinableActionData other)
+    {
+        var behaviourChangeActionData = (BehaviourActionData)other;
```
hhyyrylainen, line 54: Would it be possible to leverage the existing combining code? Even if this was changed so that MergeGuaranteed needed to return a CombinableActionData, I think that would be better than every action needing to program the dynamic MP merging *and* this merging method.

hhyyrylainen (changes requested):
I think this approach is doable. I have a few ideas on how to tweak this a bit further.

src/general/CombinableActionData.cs
84634E1A607A: If just combineable is checked, you get a new action every time you put the slider to the original place. Drag around original place slowly for several times and you'll see.

src/microbe_stage/editor/EditorActionHistory.cs
84634E1A607A: It's correct.

src/general/CombinableActionData.cs
84634E1A607A: It does cause some problems.

src/microbe_stage/editor/action_data/BehaviourActionData.cs
84634E1A607A: Can't figure out how.

src/general/CombinableActionData.cs
hhyyrylainen: I more meant like you should *reverse* this if check and not *only* check for combinable value. I did this in my refactoring (see my next comment).

src/microbe_stage/editor/action_data/BehaviourActionData.cs
hhyyrylainen: Okay so I couldn't either, but I found a pretty major not handled case, rigidity changes in subsequent multicellular cell edits, which have this kind of really ~~nasty~~ nice structure:

Kuvakaappaus - 2022-06-14 18-42-48

So let me know what you think. If no one can suggest a further simplified design, I'm happy enough with the code now.

84634E1A607A: I don't even know there's such data in CombinedEditorAction.

84634E1A607A:
> Isn't the data that is merged with always newer 

We can't guarantee. Maybe this can be removed and add some comments at the method summary.

> What sequence of actions leads to triggering the problem you are trying to solve here, so that I could test it as well?

You drag it rightwards, and then drag it back. With your design, this accidentally changes the original place.

> Also the cancels out type merging seems a bit suspicious to me now that it doesn't remove the action in the history.

At first I thought that mergeable data only exists in Single Data, but now that it also exists in Combinable Data, we can't just undo the action.

hhyyrylainen:
@84634E1A607A I did one more bit of refactoring. Thoughts on that?

hhyyrylainen (approved):
I think the code is good enough now.

84634E1A607A:
A bit more testing...

84634E1A607A:
Fine.

PR #3399, opened 2022-06-12
Patch innacurate compound cloud brightness

### LINKED ISSUE #1229: Anti-intuition brightness in cloud rendering
TIM截图20200510214409
From the image above, we can see that the circled part (where I put my mouse) has total compounds of 3675.3.
TIM图片20200510214418

And from this one, the total compounds is only 817.2, but appears several times brighter than the first image.
Seems like the brightness of the layers simply added up in the rendering algorithm and which shouldn't be happening. 
OS: Win10
Graphic Card: Intel HD 6000
Thrive Version: 0.5.0.0 beta

I did not find any log file under the folder, this may because I did not use a launcher.

MiniAlfa:

This PR patches the issue mentioned in [#1229](https://github.com/Revolutionary-Games/Thrive/issues/1229)
Without flicker issues

**Related Issues**
Fixes #1229

**Progress Checklist**

      break existing features:
      members reviewing this PR)


MiniAlfa:
Only thing is that compound planes are now drawn behind each other.
But I think that is unavoidable without creating more issues

MiniAlfa:
This PR changes the look of compound clouds. So needs verification on whether the new look is ok.

shaders/CompoundCloudPlane.shader
```diff
@@ -59,7 +59,7 @@ void fragment(){
         + // fourth
         colour4 * cloud4;
     
-    ALPHA = min(cloud1 + cloud2 + cloud3 + cloud4, 0.9f);
+    ALPHA = min(cloud1 + cloud2 + cloud3 + cloud4,  1.0f)*0.7f;
```
hhyyrylainen, line 62:
```suggestion
    ALPHA = min(cloud1 + cloud2 + cloud3 + cloud4,  1.0f) * 0.7f;
```

src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -119,7 +119,7 @@ public void UpdatePosition(Int2 newPosition)
     /// <summary>
     ///   Initializes this cloud. cloud2 onwards can be null
     /// </summary>
-    public void Init(FluidSystem fluidSystem, Compound cloud1, Compound? cloud2,
+    public void Init(FluidSystem fluidSystem, int currentIndex, Compound cloud1, Compound? cloud2,
```
hhyyrylainen, line 122: Please think of a better name for currentIndex, probably renderPriority would be a much better name? Because pure "index" is used as the compound type index within a cloud very heavily.

src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -91,16 +91,20 @@ public void Init(FluidSystem fluidSystem)
...
         {
             foreach (var cloud in clouds)
             {
                 // Re-init with potentially changed compounds
                 // TODO: special handling is needed if the compounds actually changed
-                cloud.Init(fluidSystem, cloud.Compounds[0]!, cloud.Compounds[1], cloud.Compounds[2],
+                cloud.Init(fluidSystem, currentIndex, cloud.Compounds[0]!, cloud.Compounds[1], cloud.Compounds[2],
                     cloud.Compounds[3]);
 
+                currentIndex += 1;
```
hhyyrylainen, line 106:
```suggestion
                ++currentIndex;
```

src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -128,7 +132,8 @@ public void Init(FluidSystem fluidSystem)
             if (startOffset + 3 < allCloudCompounds.Count)
                 cloud4 = allCloudCompounds[startOffset + 3];
 
-            clouds[i].Init(fluidSystem, cloud1, cloud2, cloud3, cloud4);
+            clouds[i].Init(fluidSystem, currentIndex, cloud1, cloud2, cloud3, cloud4);
+            currentIndex += 1;
```
hhyyrylainen, line 136:
```suggestion
            ++currentIndex;
```

buckly90:
image
Unfortunately the in-game background becomes a white void for me. It's fine in the editor though, which is odd.

MiniAlfa:
> Unfortunately the in-game background becomes a white void for me. It's fine in the editor though, which is odd.

I tried testing it, but i have some wifi issues right now, and can't get to clone the rep anymore . I'll come back to it in a few days once I resolve that.

hhyyrylainen:
I got the same problem. Seems like a shader syntax error: `SHADER ERROR: Expected ';' at: (null) (:19)`

shaders/CompoundCloudPlane.shader
```diff
@@ -13,6 +13,7 @@ uniform vec4 colour4 : hint_color = vec4(0, 0, 0, 0);
 uniform vec2 UVOffset = vec2(0, 0);
 
 uniform float BrightnessMultiplier = 1.0f;
+uniform float CloudAlpha = 0.7f
```
hhyyrylainen, line 16: This looks like this is the line with a missing semicolon.

MiniAlfa: sorry for this, tested it and it should work now.

Kasterisk (approved):
I somehow can't tell the difference with how the clouds look, but this doesn't make it worse so looks good to me.

src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -119,7 +119,7 @@ public void UpdatePosition(Int2 newPosition)
     /// <summary>
     ///   Initializes this cloud. cloud2 onwards can be null
     /// </summary>
-    public void Init(FluidSystem fluidSystem, Compound cloud1, Compound? cloud2,
+    public void Init(FluidSystem fluidSystem, int cloudPlaneIndex, Compound cloud1, Compound? cloud2,
```
hhyyrylainen, line 122: This is still confusingly named. Could you rename this to just `renderPriority` that's at least clear to anyone familiar with godot (transparent) materials.

hhyyrylainen (changes requested):

src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -91,16 +91,20 @@ public void Init(FluidSystem fluidSystem)
             clouds.Remove(cloud);
         }
 
+        int renderPriority = -1;
```
hhyyrylainen, line 94: Do positive values not work or why these are negatives? We already use positive render priority values for organelles so using exact opposite values here will reduce code uniformity.

hhyyrylainen: Okay, then this is the right way then. Could you add a comment explaining this here to future programmers?

RacerBG (approved):
I can't tell the difference but everything works and the bug mentioned above is not reproducible anymore.

hhyyrylainen (approved):
I fixed a typo in this and I now think this is ready for merging. I also didn't notice much different from quick testing.

PR #3412, opened 2022-06-15
Entitylabel bug fix

84634E1A607A:
1. Found that multicellular microbes may not have Species, causing a crash.
2. Commented out NaN PrintErr because it costs too much time. (15FPS before -> 60FPS after)

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/DebugOverlays.EntityLabel.cs
+++ b/src/engine/DebugOverlays.EntityLabel.cs
@@ -47,7 +47,8 @@ public partial class DebugOverlays
                 {
                     if (microbe.Species != null)
                     {
-                        label.Text = $"[{microbe.Name}:{microbe.Species.Genus[0]}.{microbe.Species.Epithet.Left(4)}]";
+                        label.Text =
+                            $"[{microbe.Name}:{microbe.Species.Genus.Left(1)}.{microbe.Species.Epithet.Left(4)}]";
                     }
 
                     break;
--- a/src/microbe_stage/CompoundBag.cs
+++ b/src/microbe_stage/CompoundBag.cs
@@ -2,7 +2,6 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using Godot;
 using Newtonsoft.Json;
 
 /// <summary>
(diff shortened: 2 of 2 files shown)


src/engine/DebugOverlays.EntityLabel.cs
```diff
@@ -47,7 +47,8 @@ private void UpdateEntityLabels()
                 {
                     if (microbe.Species != null)
                     {
-                        label.Text = $"[{microbe.Name}:{microbe.Species.Genus[0]}.{microbe.Species.Epithet.Left(4)}]";
+                        label.Text =
+                            $"[{microbe.Name}:{microbe.Species.Genus.Left(1)}.{microbe.Species.Epithet.Left(4)}]";
```
Kasterisk, line 51: Why does adding `Left` fixes the crash?

84634E1A607A: Directly use [0] triggers an exception, but Left(1) returns string.empty if there isn't any char.

84634E1A607A: When I coded this part, I assumed that the species may be less than four characters, so I used left(4), But I assume that there will be at least one character in genus, so I directly used [0].

Kasterisk: Ah okay.

hhyyrylainen: How does there exist species with no genus? Isn't that *very* invalid data?

84634E1A607A: In multicellular editor, the cell on display.

hhyyrylainen: Ah okay, I suppose that is going to be different. I guess I'll accept this then here as this is a debugging feature.

hhyyrylainen (approved):
Seems fine to me. I commented on the nan compounds issue that this PR comments out the error print and anyone wanting to debug the issue should uncomment the code.

PR #3431, opened 2022-06-18
Added functionality for configurable entity limit in performance.

### LINKED ISSUE #3071: Add configurable entity limit level in performance options
to let people select a higher or lower global entity limit based on how fast their computer is

andeason:

Adds an option in the settings in the performance tab for a custom entity limit.  

**Related Issues**
closes #3071

**Progress Checklist**

      break existing features:
      members reviewing this PR)


RacerBG (changes requested):
A lifesaver PR! However I would recommend a drop-down menu similar to the one used for the Maximum FPS value.

Oliveriver:
Definitely functional and seems to have the intended performance impact. This does have some gameplay implications though - setting the limit to its minimum of 25 meant the world was a bit barren and objects kept popping into existence in my view. While it would obviously be better to improve performance by other means, if this is what we have to do for now, so be it.

A couple of minor suggestions:

* The max value for the entity count is 1000, which seems way too high to ever be useful. Maybe set it to 150 or something.
* I agree that there needs to be a different way of entering the value. Either a dropdown or a slider.
* The text in the options menu has each word capitalised (Maximum Number of Entities) but for consistency with the other options, it shouldn't (Maximum number of entities).

hhyyrylainen:
I'm also in favour of a dropdown to select it, I think these numbers are a bit too high currently but some time ago I planned roughly these kind of options to select from: tiny (50), small (80), smaller (110), normal (140), larger (150), large (170), huge (190), extreme (250), due to latest changes these numbers should probably all be halved, then they would probably work pretty well.

RacerBG:
> I'm also in favour of a dropdown to select it, I think these numbers are a bit too high currently but some time ago I planned roughly these kind of options to select from: tiny (50), small (80), smaller (110), normal (140), larger (150), large (170), huge (190), extreme (250), due to latest changes these numbers should probably all be halved, then they would probably work pretty well.

Starting from 25 or 30 would be a good compromise in my opinion and we should increment them in even steps, for example (30, 60, 90, 120, 150). I would personally avoid anything bigger than 150 for the time being.

hhyyrylainen:
The numbers shouldn't be even shown to the player, only the descriptive names, as the entity limits would naturally also be different per stage, so it isn't useful to show them. Also 70 is the current entity limit so "normal" should be exactly 70.

RacerBG:
> he descriptive name



> The numbers shouldn't be even shown to the player, only the descriptive names, as the entity limits would naturally also be different per stage, so it isn't useful to show them. Also 70 is the current entity limit so "normal" should be exactly 70.

I agree that the numbers should be hidden. Actually, I was going to suggest similar thing for the Auto-Evo prediction values as well...

hhyyrylainen:
> I agree that the numbers should be hidden. Actually, I was going to suggest similar thing for the Auto-Evo prediction values as well...

Auto-evo prediction values are the same numbers as shown in the report. I guess percentages, how much change there is, would work as well, but that discussion is better saved somewhere else.

RacerBG (approved):
LGTM

src/engine/Settings.cs
```diff
@@ -199,6 +199,11 @@ private Settings()
     /// </summary>
     public SettingValue<int> ThreadCount { get; set; } = new(4);
 
+    /// <summary>
+    ///   Sets the maximum number of entities that can be can exist as well as their overall spawn rate
```
hhyyrylainen, line 203: > as well as their overall spawn rate

Seems suspicious to me. Do I misunderstand our spawn system or? To me the system does not adjust the spawn rates, it will only skip spawning stuff if the limit is reached, so to me this comment is not correct.

src/general/OptionsMenu.cs
```diff
@@ -752,6 +758,58 @@ private int CloudIndexToResolution(int index)
         }
     }
 
+    private int MaxEntitiesIndexToValue(int index)
+    {
+        switch (index)
+        {
+            case 0:
+                return 25;
```
hhyyrylainen, line 766: These amounts should be constants in Constants.

hhyyrylainen (changes requested):
I didn't playtest yet but from the look of things the options menu is probably fine so I just have two small code related concerns.

src/engine/Settings.cs
andeason: Mistake on my part as I didn't realize the comment in Constants was describing more than one thing.  I separated and added a separate comment for MAX_SPAWNS_PER_FRAME as a result.

simulation_parameters/Constants.cs
```diff
@@ -553,9 +553,20 @@ public static class Constants
     public const float GLUCOSE_REDUCTION_RATE = 0.8f;
     public const float GLUCOSE_MIN = 0.0f;
 
-    // These control how many game entities can exist at once and how fast they are allowed to spawn / despawn
+    // These control how many game entities can exist at once
     // TODO: bump this back up once we resolve the performance bottleneck
```
hhyyrylainen, line 557: This comment should be moved reworded to be "bump **these** back up"

simulation_parameters/Constants.cs
```diff
@@ -553,9 +553,20 @@ public static class Constants
...
+    public const int VERY_SMALL_MAX_SPAWNED_ENTITIES = 40;
+    public const int SMALL_MAX_SPAWNED_ENTITIES = 55;
+    public const int NORMAL_MAX_SPAWNED_ENTITIES = 70;
+    public const int LARGE_MAX_SPAWNED_ENTITIES = 85;
+    public const int VERY_LARGE_MAX_SPAWNED_ENTITIES = 100;
+    public const int HUGE_MAX_SPAWNED_ENTITIES = 115;
+    public const int EXTREME_MAX_SPAWNED_ENTITIES = 130;
+
+    /// <summary>
+    ///     Controls how fast entities are allowed to spawn / despawn
```
hhyyrylainen, line 568:
```suggestion
    ///   Controls how fast entities are allowed to spawn
```
Incorrect indentation and despawn is controlled by the next variable in this file.

hhyyrylainen (approved):

PR #3433, opened 2022-06-18
Update exit entity label logic

84634E1A607A:
Found a bug when testing engulfment revamp.

**NOTICE: Because the CallbackConverter.cs fails to handle static class callback, save will fail when you turn entity label on.**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/DebugOverlays.EntityLabel.cs
+++ b/src/engine/DebugOverlays.EntityLabel.cs
@@ -23,11 +23,14 @@ public partial class DebugOverlays
         }
     }
 
-    private void UpdateLabelColour(IEntity iEntity, Spatial entity, Label label)
+    private void UpdateLabelColour(IEntity iEntity, Label label)
     {
+        var entity = iEntity.EntityNode;
+
         if (!iEntity.AliveMarker.Alive)
         {
-            label.Set("custom_colors/font_color", new Color(1.0f, 0.3f, 0.3f));
+            label.AddColorOverride("font_color", new Color(1.0f, 0.3f, 0.3f));
+            return;
         }
 
         switch (entity)
@@ -38,25 +41,25 @@ public partial class DebugOverlays
                 {
                     case Microbe.MicrobeState.Binding:
                     {
-                        label.Set("custom_colors/font_color", new Color(0.2f, 0.5f, 0.0f));
+                        label.AddColorOverride("font_color", new Color(0.2f, 0.5f, 0.0f));
                         break;
                     }
 
                     case Microbe.MicrobeState.Engulf:
                     {
-                        label.Set("custom_colors/font_color", new Color(0.2f, 0.5f, 1.0f));
+                        label.AddColorOverride("font_color", new Color(0.2f, 0.5f, 1.0f));
                         break;
                     }
 
                     case Microbe.MicrobeState.Unbinding:
                     {
-                        label.Set("custom_colors/font_color", new Color(1.0f, 0.5f, 0.2f));
+                        label.AddColorOverride("font_color", new Color(1.0f, 0.5f, 0.2f));
                         break;
                     }
 
                     default:
                     {
-                        label.Set("custom_colors/font_color", new Color(1.0f, 1.0f, 1.0f));
+                        label.AddColorOverride("font_color", new Color(1.0f, 1.0f, 1.0f));
                         break;
                     }
                 }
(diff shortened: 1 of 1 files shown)


Kasterisk:
So just to confirm, this save error only happens when you turn the entity labels on but none otherwise?

> NOTICE: Because the CallbackConverter.cs fails to handle static class callback, save will fail when you turn entity label on.

Well that's pretty interesting, I think we should either directly fix the source that is to make CallbackConverter support callbacks in static classes or until that's done omit the use of callbacks in `DebugOverlays.EntityLabel` for the time being.

84634E1A607A:
Previously I forgot to remove the delegate when entity label was turned off. This PR fix that, So once you close the entity label the save will be OK.

84634E1A607A:
Well having thought for a while, I think debug overlays should never interfere the game content. I will make another pull request, maybe it tomorrow to remove that code. Checking the status of entities every frame would not be too costy.

hhyyrylainen:
I'll try to find the time to make a PR soon to make CallbackConverter fail on save when trying to save a delegate that points to a static method.

84634E1A607A:
Should be okay.

src/engine/DebugOverlays.EntityLabel.cs
```diff
@@ -22,6 +23,49 @@ private bool ShowEntityLabels
         }
     }
 
+    private void UpdateLabelColour(IEntity iEntity, Spatial entity, Label label)
```
Kasterisk, line 26: Why not access the Spatial node from IEntity's `EntityNode`? That way there don't have to be these 2 similar params.

src/engine/DebugOverlays.EntityLabel.cs
```diff
@@ -22,6 +23,49 @@ private bool ShowEntityLabels
...
+    {
+        if (!iEntity.AliveMarker.Alive)
+        {
+            label.Set("custom_colors/font_color", new Color(1.0f, 0.3f, 0.3f));
```
Kasterisk, line 30: Prefer using the explicit `AddColorOverride` as it's most often used in the codebase.

Kasterisk (changes requested):
Can confirm no more save failure when testing this branch on top of the engulf revamp branch. Some suggestions:

src/engine/DebugOverlays.EntityLabel.cs
```diff
@@ -22,6 +23,52 @@ private bool ShowEntityLabels
         }
     }
 
+    private void UpdateLabelColour(IEntity iEntity, Label label)
```
Kasterisk, line 26:
```suggestion
    private void UpdateLabelColour(IEntity entity, Label label)
```

src/engine/DebugOverlays.EntityLabel.cs
```diff
@@ -22,6 +23,52 @@ private bool ShowEntityLabels
...
+    {
+        var entity = iEntity.EntityNode;
```
Kasterisk, line 28:
```suggestion
        var node = iEntity.EntityNode;
```

src/engine/DebugOverlays.EntityLabel.cs
```diff
@@ -32,11 +79,14 @@ private void UpdateEntityLabels()
 
         foreach (var pair in entityLabels)
         {
-            var entity = pair.Key.EntityNode;
+            var iEntity = pair.Key;
+            var entity = iEntity.EntityNode;
```
Kasterisk, line 83: No longer needed.

src/engine/DebugOverlays.EntityLabel.cs
```diff
@@ -32,11 +79,14 @@ private void UpdateEntityLabels()
 
         foreach (var pair in entityLabels)
         {
-            var entity = pair.Key.EntityNode;
+            var iEntity = pair.Key;
```
Kasterisk, line 82:
```suggestion
            var entity = pair.Key;
```

Kasterisk (changes requested):

src/engine/DebugOverlays.EntityLabel.cs
84634E1A607A: What? Why?

Kasterisk: Ah wait, I see that this is still used in line 86. Nevermind this.

Kasterisk (approved):
LGTM

hhyyrylainen (approved):
Seems good that the debug panel no longer affects the game microbe objects.

PR #3435, opened 2022-06-18
Spawn density fixes

Oliveriver:

Adjusts cloud spawning for balance and sensibility.

* Added validation to biome conditions to check we never create a spawner with density greater than 1.
* Cut cloud amount in half and doubled cloud density in every biome (relative to where these values were prior to the spawn system rework in the case of the vents).
* Added cloud amount constant to tweak amount of compounds in each cloud.
* Added some randomness in the amount of compounds in each cloud.

I've playtested this a fair bit and I think it's roughly balanced for normal difficulty. My intention was to create more clouds but have less compound amount in each cloud, as I feel this creates a more interesting way of playing than getting all the compounds you need from one cloud.

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -24,7 +24,15 @@ public static class Constants
     /// </summary>
     public const float SPAWN_SECTOR_SIZE = 120.0f;
 
-    public const int CLOUD_SPAWN_SCALE_FACTOR = 10000;
+    /// <summary>
+    ///   Scale factor for density of compound cloud spawns
+    /// </summary>
+    public const int CLOUD_SPAWN_DENSITY_SCALE_FACTOR = 10000;
+
+    /// <summary>
+    ///   Scale factor for amount of compound in each spawned cloud
+    /// </summary>
+    public const float CLOUD_SPAWN_AMOUNT_SCALE_FACTOR = 0.25f;
 
     /// <summary>
     ///   The (default) size of the hexagons, used in calculations. Don't change this.
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -32,7 +32,7 @@ public static class Constants
     /// <summary>
     ///   Scale factor for amount of compound in each spawned cloud
     /// </summary>
-    public const float CLOUD_SPAWN_AMOUNT_SCALE_FACTOR = 0.25f;
+    public const float CLOUD_SPAWN_AMOUNT_SCALE_FACTOR = 0.75f;
 
     /// <summary>
     ///   The (default) size of the hexagons, used in calculations. Don't change this.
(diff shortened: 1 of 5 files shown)


simulation_parameters/Constants.cs
```diff
@@ -24,7 +24,15 @@ public static class Constants
...
-    public const int CLOUD_SPAWN_SCALE_FACTOR = 10000;
+    /// <summary>
+    ///   Scale factor for density of compound cloud spawns
+    /// </summary>
+    public const int CLOUD_SPAWN_DENSITY_SCALE_FACTOR = 10000;
+
+    /// <summary>
+    ///   Scale factor for amount of compound in each spawned cloud
+    /// </summary>
+    public const float CLOUD_SPAWN_AMOUNT_SCALE_FACTOR = 0.75f;
```
hhyyrylainen, line 35: I got the feeling that this variable is kind of useless? Because the cloud amounts are all specified in the JSON file, which you also modified in this PR.

Oliveriver: The cloud densities and amounts are used together to calculate two things: clouds themselves, and the GUI display for amount of compounds in each patch. Given that these aren't the same units, I thought it would be sensible to have an easily visible constant to translate between the two.

hhyyrylainen: Still I feel like the extra conversion factor is pretty unnecessary here. I think the display in the GUI might already have a multiplier there, so to keep the previous GUI value showing even with the different amounts actually spawning, that GUI display multiplier would be the one I think should be changed.

hhyyrylainen: I guess I can live with this change that adds one extra step to understanding what the values in biomes.json mean, so that we can focus on #3389

hhyyrylainen (approved):
Code is fine. Let's wait a bit more time to see if anyone can playtest this.

RacerBG (approved):

simulation_parameters/Constants.cs
Oliveriver: Turns out there is no other constant for the GUI - it displays the raw `amount * density` value. So I definitely think keeping this constant makes sense.

Oliveriver:
I've adjusted the compound hover thresholds as requested by @RacerBG. Now the abundance class is more useful as a measure, since you can actually differentiate 'a lot' from 'a lot a lot'.

I wanted to make further changes to biomes.json, e.g. changing density to match the actual spawn density to do away with the CLOUD_SPAWN_DENSITY_SCALE_FACTOR constant, but it turns out those values are used in Auto-Evo calculations among others, so I decided multiplying or dividing those values by 10,000 wasn't a sensible thing to do for now.

If @hhyyrylainen approves, this is ready for merging.

hhyyrylainen (approved):

PR #3436, opened 2022-06-18
Option to turn on background particles

### LINKED ISSUE #1296: Add option to turn off the environmental particles

MiniAlfa:

This pr adds an option in the options menu to turn on background particles (turned on by default).

**Related Issues**

Resolves #1296

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Kasterisk:
This looks pretty good to me as is, can you make the editor have the particles turned on by default also? There's now a preference for that.

hhyyrylainen:
Are there separate options for the stage versus the editor? Seems a bit pointless to me, but I guess we don't have *too* many graphics options yet so taking up a bit unnecessary space is not a problem yet...

MiniAlfa:
> Are there separate options for the stage versus the editor?

The editor option was added before, and completely disables particles (from being created). If you feel this is unnecessary, it can be removed.

src/microbe_stage/MicrobeCamera.cs
```diff
@@ -119,6 +119,9 @@ public override void _Ready()
 
         materialToUpdate = (ShaderMaterial)material;
 
+        OnDisplayBackgroundParticlesChanged(Settings.Instance.DisplayBackgroundParticles);
```
MiniAlfa, line 122: Is the length of the variable name "OnDisplayBackgroundParticlesChanged" an issue?
An alternative is replace background with decor.

hhyyrylainen: Is it? Doesn't look too bad to me.

hhyyrylainen:
> The editor option was added before, and completely disables particles (from being created). If you feel this is unnecessary, it can be removed.

I'm a bit confused, then. What's the functionality otherwise? I really think the option to disable particles actually should not create particles anyway, so there wouldn't be other difference than whether you want the particles in the stage and whether you want the particles in the editor. I don't see much difference between those so just a single checkbox could be used to control both.

MiniAlfa:
How does it work with the .po files? I only ran update_localization.rb, and filled in msgstr for English.

hhyyrylainen:
> and filled in msgstr for English.

gettext guesses text for new strings. Strings marked as "fuzzy" are actually untranslated and they aren't used.

Kasterisk:
> How does it work with the .po files? I only ran update_localization.rb, and filled in msgstr for English.

I recommend [Poedit](https://poedit.net) if you haven't already.

84634E1A607A (approved):
Fine.

PR #3437, opened 2022-06-19
Fix broken resulting MP calculation in cell editor

### LINKED ISSUE #3397: Editor: Mutation point bar segment no longer animating (broken resulting MP calculation)
Didn't see a related issue exist yet so I opened a new one. So originally the MP bar have a segment that shows visually how much the to-be-placed organelle would cost, this would animate the bar segment to the resulting mutation point. This now seems to be broken.

Kasterisk:

~Removed `MouseHoverPositions` overwrite in `CellEditorComponent._Process`, don't know why it's there? `CellBodyPlanEditorComponent` doesn't do that and yet its resulting MP is working, I figured that removing this one line should do the same and fix the problem and it does. I haven't yet found any issues from this change.~

https://github.com/Revolutionary-Games/Thrive/pull/3437#issuecomment-1159689346

**Related Issues**

Fixes #3397 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/early_multicellular_stage/editor/CellBodyPlanEditorComponent.cs
+++ b/src/early_multicellular_stage/editor/CellBodyPlanEditorComponent.cs
@@ -473,13 +473,13 @@ public partial class CellBodyPlanEditorComponent :
 
         if (MovingPlacedHex == null)
         {
-            moveOccupancies = GetMultiActionWithOccupancies(positions, cellTemplates, true);
+            moveOccupancies = GetMultiActionWithOccupancies(positions, cellTemplates, false);
         }
         else
         {
             moveOccupancies = GetMultiActionWithOccupancies(positions.Take(1).ToList(),
                 new List<HexWithData<CellTemplate>>
-                    { MovingPlacedHex }, false);
+                    { MovingPlacedHex }, true);
         }
 
         return Editor.WhatWouldActionsCost(moveOccupancies.Data);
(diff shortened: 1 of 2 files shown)


hhyyrylainen (changes requested):
It seems you've removed the *only* place where that is updated, so to me this change is obviously wrong. Instead you should look into the MP calculation logic why it goes wrong if the actual hover hex list is provided. Without the hover hexes the code falls back to a safety calculation of just MP cost * symmetry, which is only for safety and not meant to be used in normal case. So either remove the mouse hover hexes list entirely as it is never updated *or* (which is what I really think should be done) is that the logic for MP calculation is fixed, because that is supposed to correctly detect the moved hex to a previous position etc. things with the dynamic MP system.

Kasterisk:
Found the source of the problem, was an incorrect parameter input all along where a moving flag is set to true when we aren't really moving an organelle. Fixing that solves all of the issues.

hhyyrylainen (approved):
Flipping the boolean the right way around makes sense to me. I'll give this some time to see if anyone can playtest this.

RacerBG:
So which issues are fixed by this PR? I had observed the speed difference a lot before, does this PR fixes it as well?

hhyyrylainen:
This only addresses: https://github.com/Revolutionary-Games/Thrive/issues/3397 (AFAIK, Kasterisk can clarify if I got that wrong).

Kasterisk:
That's right.

RacerBG (approved):
The animation seems to work indeed.

PR #3450, opened 2022-06-27
Limited organelle chunk spawning to number of existing organelles

### LINKED ISSUE #3428: Microbes shouldn't drop multiple corpse chunks of organelles they have just one of
for example the nucleus model should not be able to be selected as the model for multiple cell corpse chunks

HyperionSkunk:

Fixed an issue with more organelle chunks spawning on death than the total number of that organelle present

**Related Issues**

Closes #3428

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/Microbe.Contact.cs
```diff
@@ -509,11 +509,20 @@ public void Kill()
...
         for (int i = 0; i < chunksToSpawn; ++i)
         {
             // Amount of compound in one chunk
             float amount = HexCount / Constants.CORPSE_CHUNK_AMOUNT_DIVISOR;
 
+            // If all possible organelles have been spawned or none available, end loop
+            if (organellesAvailable.Count <= 0)
+            {
+                break;
+            }
```
84634E1A607A, line 524: This should NEVER happen because crunksToSpawn is divided by 3 on line 508; And if it DOES happen, this will cause the remaining compounds vanishing.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -548,18 +557,20 @@ public void Kill()
...
                 }
 
                 if (!string.IsNullOrEmpty(organelle.Definition.DisplayScene))
                 {
                     sceneToUse.LoadedScene = organelle.Definition.LoadedScene;
                     sceneToUse.SceneModelPath = organelle.Definition.DisplaySceneModelPath;
+                    organellesAvailable.Remove(organelle);
                     break;
                 }
             }
```
84634E1A607A, line 576: Remove all unusable organelles to gain a little more performance.

84634E1A607A (changes requested):
Functionality is good. A few code related suggentions.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -548,18 +557,20 @@ public void Kill()
...
                     sceneToUse.LoadedScene = organelle.Definition.LoadedCorpseChunkScene;
+                    organellesAvailable.Remove(organelle);
                     break;
                 }
 
                 if (!string.IsNullOrEmpty(organelle.Definition.DisplayScene))
                 {
                     sceneToUse.LoadedScene = organelle.Definition.LoadedScene;
                     sceneToUse.SceneModelPath = organelle.Definition.DisplaySceneModelPath;
+                    organellesAvailable.Remove(organelle);
```
hhyyrylainen, line 573: If this is ever hit and the loop runs again, won't this actually hit an exception and not be able to continue looping?

84634E1A607A: It's a list. It's fine.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -505,10 +505,15 @@ public void Kill()
             }
         }
 
-        int chunksToSpawn = Math.Max(1, HexCount / Constants.CORPSE_CHUNK_DIVISOR);
+        // queues either 1 corpse chunk or a factor of the organelles
+        int chunksToSpawn = Math.Max(1, (organelles.Count + 1) / Constants.CORPSE_CHUNK_DIVISOR -
+            (1 / Constants.CORPSE_CHUNK_DIVISOR));
 
         var chunkScene = SpawnHelpers.LoadChunkScene();
```
HyperionSkunk, line 512: I changed the chunksToSpawn algorithm to use the number of organelles instead of hexes as fixing [this issue](https://github.com/Revolutionary-Games/Thrive/pull/3450#discussion_r907005849) resulted in the game crashing when a cell with large organelles dies, as it tries to spawn multiple chunks even though there's only one organelle

HyperionSkunk: Also trying to avoid divide by zero by factoring in/out a 1

84634E1A607A: How? I don't understand...

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -548,20 +553,24 @@ public void Kill()
...
 
                 if (!string.IsNullOrEmpty(organelle.Definition.DisplayScene))
                 {
                     sceneToUse.LoadedScene = organelle.Definition.LoadedScene;
                     sceneToUse.SceneModelPath = organelle.Definition.DisplaySceneModelPath;
+                    organellesAvailable.Remove(organelle);
                     break;
                 }
+
+                organellesAvailable.RemoveAll(item => item == organelle);
```
84634E1A607A, line 573: Emm ... I guess I'll just take back what I said before. We'd better delete this line, because if this line is executed you'll certainly get a ModifiedException the next loop. If we really want to delete without exceptions, we may need more time to do that.

src/microbe_stage/Microbe.Contact.cs
hhyyrylainen: Here's a new idea I've gotten after sleeping on this: why don't we randomize the order when creating organellesAvailable list, and then create an enumerator from it which we just step once for each time a new chunk is needed? That avoids having to modify the list at all after creating it.

HyperionSkunk: Oops, my bad regarding div/0 - should be fine without factoring, will revise

Otherwise, while trying to spawn more chunks than organelles available it may continue with the loop and try to make a chunk without entering the foreach loop, throwing the generic exception `Chunk scene has not been loaded even though it should be loaded here` in Spawners.cs line 136

HyperionSkunk: > How? I don't understand...

div/0 is not possible, my bad. Should be resolved now.

> Here's a new idea I've gotten after sleeping on this: why don't we randomize the order when creating organellesAvailable list, and then create an enumerator from it which we just step once for each time a new chunk is needed? That avoids having to modify the list at all after creating it.

That is definitely a better solution.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -547,22 +552,24 @@ public void Kill()
...
+
+                if (!string.IsNullOrEmpty(organellesAvailableEnumerator?.Current?.Definition.CorpseChunkScene))
                 {
-                    sceneToUse.LoadedScene = organelle.Definition.LoadedCorpseChunkScene;
-                    break;
+                    sceneToUse.LoadedScene = organellesAvailableEnumerator?.Current?.Definition.LoadedCorpseChunkScene;
                 }
 
-                if (!string.IsNullOrEmpty(organelle.Definition.DisplayScene))
+                if (!string.IsNullOrEmpty(organellesAvailableEnumerator?.Current?.Definition.DisplayScene))
```
84634E1A607A, line 565: else if is welcome

84634E1A607A (approved):

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -505,14 +505,19 @@ public void Kill()
             }
         }
 
-        int chunksToSpawn = Math.Max(1, HexCount / Constants.CORPSE_CHUNK_DIVISOR);
+        // Queues either 1 corpse chunk or a factor of the organelles
+        int chunksToSpawn = Math.Max(1, organelles.Count / Constants.CORPSE_CHUNK_DIVISOR);
```
hhyyrylainen, line 509: As this uses organelle count now, CORPSE_CHUNK_DIVISOR should be updated to reflect, probably should make that half as big as it is now, to more (roughly) match the existing functionalites.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -505,14 +505,19 @@ public void Kill()
...
 
         var chunkScene = SpawnHelpers.LoadChunkScene();
 
+        // Duplicates the organelle list in a random order and creates an enumerator from it
```
hhyyrylainen, line 513: You should write in the comment *why* something is done, not just what is being done here because someone reading this code will be able to figure out what this does but they can't figure out why it was done.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -505,14 +505,19 @@ public void Kill()
...
+        var organellesAvailable = organelles.ToList().OrderBy(_ => random.Next());
+        IEnumerator<PlacedOrganelle> organellesAvailableEnumerator = organellesAvailable.GetEnumerator();
```
hhyyrylainen, line 515:
```suggestion
        var organellesAvailableEnumerator = organellesAvailable.GetEnumerator();
```

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -505,14 +505,19 @@ public void Kill()
...
         }
 
-        int chunksToSpawn = Math.Max(1, HexCount / Constants.CORPSE_CHUNK_DIVISOR);
+        // Queues either 1 corpse chunk or a factor of the organelles
+        int chunksToSpawn = Math.Max(1, organelles.Count / Constants.CORPSE_CHUNK_DIVISOR);
 
         var chunkScene = SpawnHelpers.LoadChunkScene();
 
+        // Duplicates the organelle list in a random order and creates an enumerator from it
+        var organellesAvailable = organelles.ToList().OrderBy(_ => random.Next());
```
hhyyrylainen, line 514:
```suggestion
        var organellesAvailable = organelles.OrderBy(_ => random.Next()).GetEnumerator();
```
As the list is no longer updated, wouldn't this work? I mean not needing to make an intermediate list is going to be a benefit. You might need to slightly adjust how the enumerator call is done to compile.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -547,22 +552,26 @@ public void Kill()
 
             var sceneToUse = new ChunkConfiguration.ChunkScene();
 
-            // Try all organelles in random order and use the first one with a scene for model
-            foreach (var organelle in organelles.OrderBy(_ => random.Next()))
+            // Searches through the organellesAvailable list for a valid chunk to spawn
+            do
```
hhyyrylainen, line 556: Instead of using a `do` loop here, just checking if the enumerator can still move forward should be fine, as the enumerator not being valid is a permanent error that can't be fixed by looping more, so that should trigger the error condition print already.

hhyyrylainen: Or is this because organelles can have no model. In that case I think we still want to just run to the end of the enumerator once, and if we are out of organelle types to spawn, we should probably fallback to the cytoplasm model. If this fallback is used then changing HexCount to organelles.Count would also be unnecessary as the fallback would make sure that we can spawn as many chunks as is needed.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -581,6 +590,8 @@ public void Kill()
             ModLoader.ModInterface.TriggerOnChunkSpawned(chunk, false);
         }
 
+        organellesAvailableEnumerator?.Dispose();
```
hhyyrylainen, line 593: Wouldn't an using statement on the organelle enumerator work?

hhyyrylainen (changes requested):
I ended up posting my review as separate comments, see above.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -505,80 +505,88 @@ public void Kill()
...
+
+                // Sets the default chunk to be the cytoplasm model incase a chunk can't be spawned
+                var sceneToUse = new ChunkConfiguration.ChunkScene();
+                sceneToUse.LoadedScene = SimulationParameters.Instance
+                    .GetOrganelleType(Constants.DEFAULT_CHUNK_MODEL_NAME).LoadedCorpseChunkScene;
+
+                // Will only loop if there are still organelles available
+                organellesAvailableEnumerator?.MoveNext();
+                if (i < organelles.Count && organellesAvailableEnumerator?.Current != null)
                 {
```
HyperionSkunk, line 560: There may be a better way to do this but this is the best solution I came up with

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -505,80 +505,88 @@ public void Kill()
...
+                    chunkType.Compounds[entry.Key] = compoundValue;
                 }
 
-                if (!string.IsNullOrEmpty(organelle.Definition.DisplayScene))
+                chunkType.Meshes = new List<ChunkConfiguration.ChunkScene>();
+
+                // Sets the default chunk to be the cytoplasm model incase a chunk can't be spawned
+                var sceneToUse = new ChunkConfiguration.ChunkScene();
+                sceneToUse.LoadedScene = SimulationParameters.Instance
+                    .GetOrganelleType(Constants.DEFAULT_CHUNK_MODEL_NAME).LoadedCorpseChunkScene;
```
HyperionSkunk, line 555: This too, assigning it preemptively is not very elegant

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -505,80 +505,88 @@ public void Kill()
             }
         }
 
+        // Queues either 1 corpse chunk or a factor of the hexes
         int chunksToSpawn = Math.Max(1, HexCount / Constants.CORPSE_CHUNK_DIVISOR);
```
HyperionSkunk, line 509: Reverted to HexCount

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -505,80 +505,88 @@ public void Kill()
...
 
         var chunkScene = SpawnHelpers.LoadChunkScene();
 
-        for (int i = 0; i < chunksToSpawn; ++i)
+        // An enumerator to step through all available organelles in a random order when making chunks
+        using (var organellesAvailableEnumerator = organelles.OrderBy(_ => random.Next()).GetEnumerator())
         {
```
HyperionSkunk, line 515: Adding a using loop marked more changes than I actually made- most is just indentation

hhyyrylainen: I switched this to a using statements, which we prefer in Thrive for exactly the reason of not needing to add extra indentation.

hhyyrylainen (approved):
Did my own refactoring to help get this finished, I think I was able to simplify the code quite a bit by making the enumerator not nullable, and changed the loop condition that didn't make much sense to me. Please check my changes carefully for future reference.

PR #3452, opened 2022-06-28
Rigidity slider no longer jumps to mouse

### LINKED ISSUE #1533: Inactive slider doesn't detect mouse up
might be a Godot bug:
- Use all MP, try to move the rigidity bar
- Keep mouse held down while moving the mouse off the slider to the undo button
- Undo changes
- Move cursor back to the rigidity bar and see the slider value jump to the mouse without pressing anything

Cowman9002:

The PR changes the rigidity slider so that once it becomes locked at an edge, resetting it through a undo will no longer cause it to snap to the position of the mouse when the mouse is hovered.

**Related Issues**

closes #1533

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/CellEditorComponent.GUI.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.GUI.cs
@@ -202,7 +202,7 @@ public partial class CellEditorComponent
 
     private void UpdateRigiditySliderState(int mutationPoints)
     {
-        Godot.Texture grabberDisableIcon = rigiditySlider.GetIcon("grabber_disabled");
+        Texture grabberDisableIcon = rigiditySlider.GetIcon("grabber_disabled");
 
         int costPerStep = (int)Math.Min(Constants.MEMBRANE_RIGIDITY_COST_PER_STEP * CostMultiplier, 100);
         if (mutationPoints >= costPerStep && MovingPlacedHex == null)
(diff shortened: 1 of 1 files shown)


Cowman9002:
I'm not sure about the change events being emitted, but the slider is still unable to move without MP due to this code:
```
int cost = Math.Abs(rigidity - intRigidity) * costPerStep;

if (cost > Editor.MutationPoints)
{
    ...
    rigidity = intRigidity > rigidity ? intRigidity - stepsLeft : intRigidity + stepsLeft;
}
```

Which can be found in the file src\microbe_stage\editor\CellEditorComponent.cs starting on line 890.

84634E1A607A:
Once when I was tackling #3991 I tried to delete this code and make dynamic MP decide if the slider can still move (but failed). I prefer that we never disable the sliders but fix that logic mentioned above
> but the slider is still unable to move without MP due to this code

hhyyrylainen:
> Once when I was tackling #3991 I tried to delete this code and make dynamic MP decide if the slider can still move (but failed). I prefer that we never disable the sliders but fix that logic mentioned above

I suppose, yes, that'll actually make sense. We just need a way to prevent dragging it in a direction that would cause negative MP, but allow dragging in a direction that gives back MP.

Cowman9002:
I'm not sure if this is exactly what you mean, but I was also playing around with the code and found out that if you write:
```
int cost = (Math.Abs(rigidity) - Math.Abs(intRigidity)) * costPerStep;
```
The slider can not go any farther into its domain (positive to farther positive) but can still go back even to the opposite domain as long as the player has enough MP to do so.

I didn't put it in the PR because I didn't want to do any extra that what the issue reported.

PR #3453, opened 2022-06-29
Auto evo exploring tool

### LINKED ISSUE #1573: Make an auto-evo exploring tool

84634E1A607A:
Still a work in progress.

**Related Issues**

Closes #1573 

Forum discussion thread: https://forum.revolutionarygamesstudio.com/t/auto-evo-exploring-tool/897/20

**Progress Checklist**

- [x] Implement auto-evo controls
- [x] Implement auto-evo config editor
- [x] Implement auto-evo history viewer
- [x] Implement species viewer
- [x] Implement hex view of cells
- [x] Merge two history selections together
- [x] Use a light background (Help needed)
- [x] Translation
- [x] Code style
- [x] Play with current setting
- [x] Sound
      break existing features:
      members reviewing this PR)


84634E1A607A:
I found a problem, that when `PhotoStudio.UseBackgroundSceneLoad = true` and auto-evo running in full speed, PhotoStudio stops giving photos. So you can't get previews when running auto-evo. What makes matter worse, PhotoStudio don't have an Abort method, so all queued tasks must be finished even if they are no longer needed.

hhyyrylainen:
> I found a problem, that when PhotoStudio.UseBackgroundSceneLoad = true

How does it get stuck? Is it just a lot slower or are we hitting an engine bug here? The reason I added that variable in the first place is that I was afraid that we'd hit an engine bug and would need to turn that off.

> PhotoStudio don't have an Abort method, so all queued tasks must be finished even if they are no longer needed.

It would be easy enough to add a cancel method to the tasks. The reason I did not add that is because it's easier to use the photostudio from all over the code base if the photo tasks cannot fail. If the cancel flag is added each place using the photostudio needs to make sure they handle canceled tasks. And a bit more work on top of that would be to make the components cancel photostudio tasks in a reasonable way.

84634E1A607A:
> Is it just a lot slower

It is. The TaskExecutor may need to take priority into account. The current problem is that a fullspeed auto-evo adds too many tasks to thee executor that it prevents any newly added task to complete until auto-evo is finished.

hhyyrylainen:
Ah, so there is still no sign of a engine bug, which is good. I haven't checked but it might be possible to push tasks to the front of the queue, if such a method doesn't exist, then some other kind of solution is needed.

Kasterisk:
`EvolutionaryTreeNode.tscn` is missing.

src/microbe_stage/editor/MicrobeEditorPatchMap.tscn
```diff
@@ -108,8 +108,6 @@ margin_right = 1500.0
 margin_bottom = 1500.0
 rect_min_size = Vector2( 1500, 1500 )
 DrawDefaultMapIfEmpty = false
-PatchNodeWidth = 64.0
```
hhyyrylainen, line 111: Unintended changes?

84634E1A607A: I think I deleted this export value and moved it to Constants.cs in my last few prs so this is automatically deleted.

Surely this is unintended, but this is the way (

hhyyrylainen: I suppose it is, I can't find a variable named PatchNodeWidth anymore.

84634E1A607A:
I'll leave the evolutionary tree performance problem for later. I'm not quite sure if I can solve it and I may mess up.

src/microbe_stage/MicrobeSpecies.cs
```diff
@@ -183,4 +187,21 @@ public override int GetVisualHashCode()
...
+        // We need to call _Ready here as the object may not be attached to the scene yet by the photo studio
+        microbe._Ready();
+
+        microbe.ApplySpecies(this);
+    }
+
+    public float CalculatePhotographDistance(Spatial instancedScene)
+    {
+        return PhotoStudio.CameraDistanceFromRadiusOfObject(((Microbe)instancedScene).Radius *
+            Constants.PHOTO_STUDIO_CELL_RADIUS_MULTIPLIER);
```
hhyyrylainen, line 205: This is duplicated code from MicrobeSpecies. MicrobeSpecies should be refactored to call these methods like this:
```
    public void ApplySceneParameters(Spatial instancedScene)
    {
       ((Microbe)instancedScene).ApplySceneParameters(instancedScene);
    }
```

hhyyrylainen: No wait sorry this is almost exactly duplicated code from CellType. That's the place these should be unified with as much as possible, especially bad is the long technical comment that is in multiple places.

src/gui_common/SpeciesPreview.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    private void UpdatePreviewSpecies()
+    {
+        if (previewSpecies is MicrobeSpecies microbeSpecies)
+        {
+            task = new ImageTask(microbeSpecies);
+            PhotoStudio.Instance.SubmitTask(task);
+        }
```
hhyyrylainen, line 48: I'd add an else here with an error print that an unknown species type was attempted to be previewed.

src/gui_common/DraggableScrollContainer.cs
```diff
@@ -79,8 +79,7 @@ public override void _Ready()
     {
         base._Ready();
 
-        if (ContentPath == null)
-            throw new InvalidOperationException("Path to child control is expected");
+        ContentPath ??= GetChild(0).GetPath();
```
hhyyrylainen, line 82: Could you write documentation on the ContentPath variable that if it is null then the path of the first child is used?

src/gui_common/CustomDropDown.cs
```diff
@@ -86,6 +86,14 @@ public void AddItemSection(string name)
         return item;
     }
 
+    /// <summary>
+    ///   Helper for clearing the items in the dictionary.
```
hhyyrylainen, line 90: Could you add a clarification here that this doesn't cause rebuild of the actual child controls?

src/general/GameWorld.cs
```diff
@@ -82,6 +82,9 @@ public GameWorld()
     [JsonProperty]
     public Species PlayerSpecies { get; private set; } = null!;
 
+    [JsonIgnore]
+    public Dictionary<uint, Species> Species => worldSpecies;
```
hhyyrylainen, line 86: Would it be possible to return a read only dictionary interface? It would be extremely bad if some external code modified the existing species data through this.

src/auto-evo/EditorAutoEvoRun.cs
```diff
@@ -6,10 +6,14 @@
...
 {
-    public EditorAutoEvoRun(GameWorld world, Species originalEditedSpecies, Species modifiedProperties) : base(world)
+    private readonly AutoEvoConfiguration configuration;
+
+    public EditorAutoEvoRun(GameWorld world, Species originalEditedSpecies, Species modifiedProperties,
+        AutoEvoConfiguration? configuration = null) : base(world, configuration)
     {
         OriginalEditedSpecies = originalEditedSpecies;
         ModifiedProperties = modifiedProperties;
+        this.configuration = configuration ?? SimulationParameters.Instance.AutoEvoConfiguration;
```
hhyyrylainen, line 16: The configuration is already passed to the base class on line 12, and unless I misremember this fallback logic already exists in there, so isn't this line unnecessary?

src/auto-evo/AutoEvoConfiguration.cs
```diff
@@ -1,48 +1,49 @@
-﻿using Newtonsoft.Json;
+﻿using System;
+using Newtonsoft.Json;
 
-public class AutoEvoConfiguration : IRegistryType
+public class AutoEvoConfiguration : IRegistryType, ICloneable
 {
     [JsonProperty]
-    public int MutationsPerSpecies { get; private set; }
+    public int MutationsPerSpecies { get; set; }
```
hhyyrylainen, line 7: Now that everything is publicly settable here, could you add a comment on this class like:
```
/// <summary>
///   Settings for auto-evo
/// </summary>
/// <remarks>
///   <para>
///      Like all registry types this should not be modified during runtime at all. The reason the properties are modifiable is to allow auto-evo exploring tool to modify its own copy of the settings
...
```

src/microbe_stage/MicrobeSpecies.cs
84634E1A607A: Wait, how do I apply species without telling Microbe what its species is?

84634E1A607A: Because of that, I can't make Microbe photograph itself...

hhyyrylainen: Looking at the code, it looks like CellType creates a temporary microbe species object, so moving as much logic as possible to the MicrobeSpecies photograph would be optimal, then the CellType could just call those methods without having code duplication.

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -0,0 +1,222 @@
...
+        // Draw extinct species name
+        foreach (var extinctedSpecies in latestNodes.Values.Where(n => n.LastGeneration))
+        {
+            DrawString(latoSmall, new Vector2(
+                extinctedSpecies.RectPosition.x + extinctedSpecies.RectSize.x + SPECIES_NAME_OFFSET,
+                extinctedSpecies.Center.y), speciesNames[extinctedSpecies.SpeciesID], Colors.DarkRed);
+        }
+    }
+
+    public void Init(Species player)
```
hhyyrylainen, line 102: I'd put this just after _Ready (before _Draw) as logically these are both about making this class ready to use.

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -0,0 +1,222 @@
...
+    {
+        SetupTreeNode(player, null, 0);
```
hhyyrylainen, line 104: Btw, why does this need the player species specifically? Wouldn't it make more sense that this is given the "luca" species and starts from there? I guess if you don't do any special player stuff, just renaming this variable would be enough.

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -0,0 +1,222 @@
...
+
+        speciesOrigin.Add(player.ID, (uint.MaxValue, 0));
```
hhyyrylainen, line 106: Are the zeros in this method the generation or something? If so I'd check that the given `Species.Generation == 0` otherwise the logic might break at a surprising time.

src/auto-evo/RunResults.cs
```diff
@@ -987,6 +988,17 @@ public void LogResultsToTimeline(GameWorld world, List<ExternalEffect>? effects
             }
         }
 
+        public IEnumerator GetEnumerator()
```
hhyyrylainen, line 991: I'm fine with making this an IEnumerable, but could you add a comment here warning that if auto-evo is running while enumerating this, that's going to randomly cause collection modified exceptions, so don't enumerate this while auto-evo is not ready yet?

src/gui_common/CellHexPhotoBuilder.cs
```diff
@@ -0,0 +1,135 @@
+﻿using System;
+using Godot;
+
+public class CellHexPhotoBuilder : Spatial, IPhotographable
```
hhyyrylainen, line 4: I'd move this class into the microbe stage folder as photographing the hexes of a cell are not a common operation, instead it sounds like a microbe specific thing.

src/gui_common/CellHexPhotoBuilder.cs
```diff
@@ -0,0 +1,135 @@
...
+            return;
+        }
+
+        float farthest = 0;
+
+        foreach (var organelle in species.Organelles)
+        {
+            var position = organelle.Position;
+            var itemHexes = organelle.RotatedHexes;
+            foreach (var hex in itemHexes)
```
hhyyrylainen, line 124:
```suggestion
            foreach (var hex in organelle.RotatedHexes)
```

src/gui_common/CellHexPhotoBuilder.cs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    private void BuildHexStruct()
+    {
+        var hexScene = GD.Load<PackedScene>("res://src/microbe_stage/editor/EditorHex.tscn");
+        var validMaterial = GD.Load<Material>("res://src/microbe_stage/editor/ValidHex.material");
+        var modelScene = GD.Load<PackedScene>("res://src/general/SceneDisplayer.tscn");
+
+        foreach (Node node in GetChildren())
+            node.DetachAndQueueFree();
```
hhyyrylainen, line 53: I'm pretty sure there's an extension method to free all child nodes in NodeHelpers or something.

src/auto-evo/RunResults.cs
84634E1A607A: results is a concurrent dict, it shouldn't?

src/gui_common/CellHexPhotoBuilder.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        if (Species == null)
+            throw new InvalidOperationException("Species is ont initialized");
+
+        var organelleLayout = Species.Organelles;
+        foreach (var organelle in organelleLayout.Organelles)
+        {
+            var position = organelle.Position;
+            var itemHexes = organelle.RotatedHexes;
+            foreach (var hex in itemHexes)
```
hhyyrylainen, line 63:
```suggestion
            foreach (var hex in organelle.RotatedHexes)
```

src/gui_common/CellHexPhotoBuilder.cs
```diff
@@ -0,0 +1,135 @@
...
+                var hexNode = (MeshInstance)hexScene.Instance();
+                AddChild(hexNode);
+                hexNode.MaterialOverride = validMaterial;
+                hexNode.Translation = pos;
+            }
+        }
+
+        foreach (var organelle in organelleLayout)
+        {
+            // Hexes are handled by UpdateAlreadyPlacedHexes
```
hhyyrylainen, line 76: This seems like copied code from the hex editor...

I'd really like to avoid this much logic duplication because changing this in the future will be harder.
I think a few static methods in the organelle editor that do stuff like instance and place the scenes in the right places would be good so that this and the editor could share the same core logic, which is complex code.

src/auto-evo/RunResults.cs
hhyyrylainen: Eh, I use manual locking anyway in this class (which might be unnecessary, but it still exists), so I'd at least warn that this bypasses that locking so using this while auto-evo is running is not suitable.

src/auto-evo/EvolutionaryTree.cs
84634E1A607A: This generation has nothing to do with Species.Generation. This actually means "0 Myrs"

src/gui_common/CellHexPhotoBuilder.cs
84634E1A607A: They are somehow different because this builder simplifies the logic greatly. I cannot merge them together...

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -0,0 +1,222 @@
...
+        DrawLine(new Vector2(0, TIMELINE_LINE_Y), new Vector2(RectSize.x, TIMELINE_LINE_Y), Colors.DarkCyan,
+            TIMELINE_LINE_THICKNESS, true);
+
+        for (int i = 0; i <= latestGeneration; i++)
+        {
+            DrawLine(new Vector2(i * GENERATION_SEPARATION + treeNodeSize.x / 2, TIMELINE_LINE_Y),
+                new Vector2(i * GENERATION_SEPARATION + treeNodeSize.x / 2, TIMELINE_LINE_Y + TIMELINE_MARK_SIZE),
+                Colors.DarkCyan, TIMELINE_LINE_THICKNESS, true);
+
+            var localizedText = i + " " + TranslationServer.Translate("MEGA_YEARS");
```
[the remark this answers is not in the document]
84634E1A607A: That value is double, this one is int...

src/gui_common/CellHexPhotoBuilder.cs
hhyyrylainen: I was thinking the general logic would work like so that you give it a single organelle, along with callbacks it can use to ask for things like the graphics holders. I think while that would be a bit complicated to make, it would still allow 80-90% code sharing, and all of the complex math could be shared.

src/auto-evo/EvolutionaryTree.cs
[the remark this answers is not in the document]
hhyyrylainen: That format also applies to ints as well, right? Because for Finnish for example it inserts spaces every 3 zeros in the number, just converting an int to a string, not even using the user's locale here, is not good enough.

src/gui_common/CustomDropDown.cs
```diff
@@ -86,6 +86,19 @@ public void AddItemSection(string name)
...
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     This method doesn't cause rebuilt of the popup.
```
hhyyrylainen, line 94:
```suggestion
    ///     This method doesn't cause rebuild of the popup.
```
typo

src/general/EditorBase.cs
```diff
@@ -789,26 +789,7 @@ private void ApplyAutoEvoResults()
...
-            entry.Value.RecordSnapshot(true);
-        }
-
-        var extinct = CurrentGame.GameWorld.Map.RemoveExtinctSpecies(FreeBuilding);
-
-        foreach (var species in extinct)
-        {
-            CurrentGame.GameWorld.RemoveSpecies(species);
-        }
+        run.ApplyAllEffects(FreeBuilding);
```
hhyyrylainen, line 792: I'd rename ApplyAllEffects to ApplyAllResultsAndEffects to make it clear that that is what does everything auto-evo needs to do after a run.

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -0,0 +1,224 @@
...
+    {
+        var node = treeNodeScene.Instance<EvolutionaryTreeNode>();
+        node.Generation = generation;
+        node.SpeciesID = species.ID;
+        node.LastGeneration = false;
+        node.ParentNode = parent;
+        node.RectPosition = new Vector2(generation * GENERATION_SEPARATION, TIMELINE_HEIGHT);
+        node.LastGeneration = isLastGeneration;
+        node.Group = nodesGroup;
+        node.Connect("button_down", this, nameof(OnTreeNodeSelected), new Array { node });
```
hhyyrylainen, line 168: Shouldn't this be on the clicked event as the tree node is a button so it will emit the button event whenever things that constitute clicking a button happen?

src/auto-evo/AutoEvoExploringTool.tscn
```diff
@@ -0,0 +1,763 @@
...
+margin_right = 62.0
+margin_bottom = 35.0
+text = "BACK"
+
+[node name="ExitConfirmationDialog" parent="." instance=ExtResource( 11 )]
+rect_min_size = Vector2( 400, 0 )
+WindowTitle = "CONFIRM_EXIT"
+DialogText = "RETURN_TO_MENU_WARNING"
+
+[connection signal="pressed" from="GUI/VBoxContainer/TabButtons/ConfigButton" to="." method="ChangeTab" binds= [ 0 ]]
```
hhyyrylainen, line 751: Elsewhere we use the approach where the selected tab is an enum and the binds is a valid enum value like "Structure" for the editor structure tab. I think the same pattern should be used here for codebase consistency reasons.

src/auto-evo/AutoEvoRun.cs
```diff
@@ -123,17 +126,18 @@ public string Status
...
 
             if (total > 0)
             {
                 var percentage = CompletionFraction * 100;
 
-                // {0:F1}% done. {1:n0}/{2:n0} steps.
-                return TranslationServer.Translate("AUTO-EVO_STEPS_DONE").FormatSafe(percentage, CompleteSteps, total);
+                // {0:F1}% done. {1:n0}/{2:n0} steps. [Paused.]
+                return TranslationServer.Translate("AUTO-EVO_STEPS_DONE").FormatSafe(percentage, CompleteSteps, total)
+                    + (Running ? string.Empty : " " + TranslationServer.Translate("PAUSED_DOT"));
```
hhyyrylainen, line 140: I'd rename "PAUSED_DOT" key to "OPERATION_PAUSED_TAG" , and change the text to "[Paused]"

src/auto-evo/AutoEvoRun.cs
```diff
@@ -174,6 +178,44 @@ public void Start()
...
+        {
+            if (Step())
+                Finished = true;
+        }
+        catch (Exception e)
+        {
+            Aborted = true;
+            GD.PrintErr("Auto-evo failed with an exception: ", e);
+        }
+
```
hhyyrylainen, line 203: missing `timer.Stop` here?

hhyyrylainen: I see I did the same thing elsewhere, so you are free to ignore my comment here.

src/auto-evo/AutoEvoRun.cs
```diff
@@ -174,6 +178,44 @@ public void Start()
...
+        Running = false;
+
+        RunDuration += timer.Elapsed;
+    }
+
+    public void Continue()
+    {
+        if (Running)
```
hhyyrylainen, line 211: Isn't there a possible race condition here? As in Running won't be set to true until another thread does that, so if this method is called twice very quickly in a row it can result in multiple auto-evo starts.
Maybe it's enough to set `Running = true` on line 213?

src/auto-evo/AutoEvoRun.cs
```diff
@@ -619,4 +624,67 @@ private void RunSingleStepToCompletion(IRunStep step)
...
+            entry.Value.RecordSnapshot(true);
+        }
+
+        var extinct = Parameters.World.Map.RemoveExtinctSpecies(playerCantGoExtinct);
+
+        foreach (var species in extinct)
+        {
+            Parameters.World.RemoveSpecies(species);
+
+            GD.Print("Species ", species.FormattedName, " has gone extinct from the world.");
```
hhyyrylainen, line 648: I'd remove this logging, timeline should include this info already or be extended to include events like this. I plan on removing all GD.Prints coming from auto-evo as I see it now as a bit of a problematic thing, so it'd help me if this could be removed in this PR.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
+﻿using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using AutoEvo;
+using Godot;
+using HarmonyLib;
+
+public class AutoEvoExploringTool : NodeWithInput
```
hhyyrylainen, line 8: This is a pretty big feature and this is the main class, so could you add a brief summary here what the whole exploring tool is about?

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+
+        // Init game
+        gameProperties = GameProperties.StartNewMicrobeGame(new WorldGenerationSettings());
+        autoEvoConfiguration = (AutoEvoConfiguration)SimulationParameters.Instance.AutoEvoConfiguration.Clone();
+
+        // Init generation 0 (LUCA)
+        runResultsList.Add(new LocalizedStringBuilder());
+        speciesHistoryList.Add(new Dictionary<uint, Species>
+        {
+            { gameProperties.GameWorld.PlayerSpecies.ID, gameProperties.GameWorld.PlayerSpecies },
```
hhyyrylainen, line 285: I guess this is fine for now, but could you add a TODO comment that it would be nice if we could remove the player species entirely and have an NPC species as the LUCA?

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+            autoEvoConfiguration.NewBiodiversityIncreasingSpeciesPopulation;
+        protectMigrationsFromSpeciesCapCheckBox.Pressed = autoEvoConfiguration.ProtectMigrationsFromSpeciesCap;
+        protectNewCellsFromSpeciesCapCheckBox.Pressed = autoEvoConfiguration.ProtectNewCellsFromSpeciesCap;
+        refundMigrationsInExtinctionsCheckBox.Pressed = autoEvoConfiguration.RefundMigrationsInExtinctions;
+        strictNicheCompetitionCheckBox.Pressed = autoEvoConfiguration.StrictNicheCompetition;
+        speciesSplitByMutationThresholdPopulationAmountSpinBox.Value =
+            autoEvoConfiguration.SpeciesSplitByMutationThresholdPopulationAmount;
+        speciesSplitByMutationThresholdPopulationFractionSpinBox.Value =
+            autoEvoConfiguration.SpeciesSplitByMutationThresholdPopulationFraction;
+        useBiodiversityForceSplitCheckBox.Pressed = autoEvoConfiguration.UseBiodiversityForceSplit;
```
hhyyrylainen, line 314: The _Ready method is pretty long so I'd put this in a separate method.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+        runResultsList.Add(new LocalizedStringBuilder());
+        speciesHistoryList.Add(new Dictionary<uint, Species>
+        {
+            { gameProperties.GameWorld.PlayerSpecies.ID, gameProperties.GameWorld.PlayerSpecies },
+        });
+        historyListMenu.AddItem("0", false, Colors.White);
+        historyListMenu.CreateElements();
+        HistoryListMenuIndexChanged(0);
+
+        evolutionaryTree.Init(gameProperties.GameWorld.PlayerSpecies);
```
hhyyrylainen, line 291: And this part as well.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+                speciesSelectPanel.Visible = true;
+                historyReportSplit.Visible = true;
+                viewerTab.Visible = true;
+                break;
+            }
+        }
+    }
+
+    /// <summary>
+    ///   This function updates all configurations in a row to avoid adding numerous separate callback functions.
```
hhyyrylainen, line 406: This is a bit lazy, but I'll allow it as this isn't a core gameplay component that should have the best quality code.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+    /// <summary>
+    ///   Apply auto-evo effects and set up related controls
+    /// </summary>
+    private void ApplyAutoEvoRun()
+    {
+        // Apply the results
+        autoEvoRun!.ApplyAllEffects(true);
+
+        // Add run results
+        RunResults results = autoEvoRun.Results!;
```
hhyyrylainen, line 477:
```suggestion
        var results = autoEvoRun.Results!;
```
?

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+        evolutionaryTree.UpdateEvolutionaryTreeWithRunResults(results, ++currentGeneration);
+        runResultsList.Add(results.MakeSummary(gameProperties.GameWorld.Map, true));
+        speciesHistoryList.Add(
+            gameProperties.GameWorld.Species.ToDictionary(pair => pair.Key, pair => (Species)pair.Value.Clone()));
+
+        // Add check box to history container
```
hhyyrylainen, line 483:
```suggestion
        // Add checkbox to history container
```

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+    /// </summary>
+    private void ApplyAutoEvoRun()
+    {
+        // Apply the results
+        autoEvoRun!.ApplyAllEffects(true);
+
+        // Add run results
+        RunResults results = autoEvoRun.Results!;
+        evolutionaryTree.UpdateEvolutionaryTreeWithRunResults(results, ++currentGeneration);
+        runResultsList.Add(results.MakeSummary(gameProperties.GameWorld.Map, true));
```
hhyyrylainen, line 479: Calling MakeSummary after applying auto-evo effects doesn't result in the correct summary text.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+
+        // Add check box to history container
+        historyListMenu.AddItem(currentGeneration.ToString(), false, Colors.White);
+        historyListMenu.CreateElements();
+
+        currentGenerationLabel.Text = currentGeneration.ToString();
+    }
+
+    /// <summary>
+    ///   Run one step sync
```
hhyyrylainen, line 491:
```suggestion
    ///   Run one step synchronously
```

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+    {
+        GUICommon.Instance.PlayButtonPressSound();
+
+        if (autoEvoRun?.Aborted != false || autoEvoRun.Finished)
+        {
+            autoEvoRun = new AutoEvoRun(gameProperties.GameWorld, autoEvoConfiguration);
+        }
+
+        // To avoid concurrent steps
+        autoEvoRun.FullSpeed = false;
```
hhyyrylainen, line 503: I'm not convinced that setting full speed to false here actually prevents concurrent steps. Shouldn't this wait for currently executing auto-evo step tasks to stop? Or can this method only be called when auto-evo hasn't been running in the background previously?

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+            return;
+
+        UpdateSpeciesPreview(species);
+    }
+
+    private void UpdateSpeciesPreview(Species species)
+    {
+        speciesListMenu.Text = species.FormattedName;
+        speciesPreview.PreviewSpecies = species;
+        hexPreview.PreviewSpecies = species as MicrobeSpecies;
```
hhyyrylainen, line 569: I'd add an if here for the cast and print an error if the species is an unknown type to preview.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+                speciesDetailsLabel.ExtendedBbcode += $"[b]Stage:[/b]\n  Microbe\n" +
+                    $"[b]Membrane Type:[/b]\n  {microbeSpecies.MembraneType.Name}\n" +
+                    $"[b]Membrane Rigidity:[/b]\n  {microbeSpecies.MembraneRigidity}\n" +
+                    $"[b]Base Speed:[/b]\n  {microbeSpecies.BaseSpeed}\n" +
+                    $"[b]Base Rotation Speed[/b]\n  {microbeSpecies.BaseRotationSpeed}\n" +
+                    $"[b]Base Hex Size[/b]\n  {microbeSpecies.BaseHexSize}";
+                break;
+            }
+        }
+        */
```
hhyyrylainen, line 601: This should be removed.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+                speciesDetailsLabel.ExtendedBbcode += string.Format(CultureInfo.CurrentCulture,
+                    TranslationServer.Translate("MICROBE_SPECIES_DETAIL_TEXT"), microbeSpecies.MembraneType.Name,
+                    microbeSpecies.MembraneRigidity, microbeSpecies.BaseSpeed, microbeSpecies.BaseRotationSpeed,
+                    microbeSpecies.BaseHexSize);
+                break;
+            }
+        }
+    }
+
+    private void PlayWithCurrentSettingPressed()
```
hhyyrylainen, line 621: Ah, you implemented the enter the game feature, I guess then my previous comment about adding a TODO about the player should be ignored. Instead could you add a comment on the luca variable that it'd be nice to unmark the species as the player species while in the tool, because being the player species gives it an advantage, and probably stops it from evolving?

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+
+        // Disable the button to prevent it being executed again.
+        playWithCurrentSettingButton.Disabled = true;
+
+        TransitionManager.Instance.AddSequence(ScreenFade.FadeType.FadeOut, 0.1f, () =>
+        {
+            // Instantiate a new editor scene
+            var editor = (MicrobeEditor)SceneManager.Instance.LoadScene(MainGameState.MicrobeEditor).Instance();
+
+            // Start freebuild game
```
hhyyrylainen, line 633:
```suggestion
            // Copy our currently setup game to the editor
```

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,640 @@
...
+        speciesPreview = GetNode<SpeciesPreview>(SpeciesPreviewPath);
+        hexPreview = GetNode<CellHexPreview>(HexPreviewPath);
+        speciesListMenu = GetNode<CustomDropDown>(SpeciesListMenuPath);
+        speciesDetailsLabel = GetNode<CustomRichTextLabel>(SpeciesDetailsLabelPath);
+        evolutionaryTree = GetNode<EvolutionaryTree>(EvolutionaryTreePath);
+
+        exitConfirmationDialog = GetNode<CustomConfirmationDialog>(ExitConfirmationDialogPath);
+
+        // Init game
+        gameProperties = GameProperties.StartNewMicrobeGame(new WorldGenerationSettings());
```
hhyyrylainen, line 278:
```suggestion
        gameProperties = GameProperties.StartNewMicrobeGame(new WorldGenerationSettings(), true);
```
That true should be passed here so that the game is in freebuild mode, if the player decides to play in the world.

hhyyrylainen (changes requested):
See my previous comments for my code related concerns. Basically just one code duplication thing I'd like to resolve, other stuff is really minor typos and slight refactoring.

 - [x] (By Ajax) As for gameplay testing, I noticed the confirm exit popup seems to be incorrectly sized with a bunch of space above and below the "Are you sure you want to exit..." text.

 - [x] (This is because the left margin being 0, Ajax) When going to the Viewer without doing anything the evolutionary tree has the "0 Myr" text label slightly cutoff on the left side.

I got one weird run where other species refused to appear, but that may have had more to do with the random map that got generated or something. (actually seems like if the split gain population is increased from 100 it basically prevents other species from appearing, which might be a bug in auto-evo).

 - [x] Also now that I played this, I realized that the paused text is literally "Paused." so instead of my earlier comment, I'd say that translation key should be "OPERATION_PAUSED_DOT" with the current content, just so if any language has a need to know what is paused to translate it correctly.

src/auto-evo/AutoEvoRun.cs
hhyyrylainen: See my overall review for an update on this.

hhyyrylainen:
Once this is merged I'll open separate issues about the evolution tree performance and the multi generation runs with a single button.

locale/en.po
```diff
@@ -983,115 +983,319 @@ msgstr "Cilia"
...
+msgid "STATUS_COLON"
+msgstr "Status:"
+
+#: ../src/auto-evo/AutoEvoExploringTool.tscn:431
+msgid "READY"
+msgstr "Ready"
+
+#: ../src/auto-evo/AutoEvoExploringTool.tscn:440
+msgid "FINISH_ONE_GENERATION"
+msgstr "Finish One Gegeration"
```
adQuid, line 1115:
```suggestion
msgstr "Finish One Generation"
```

src/auto-evo/AutoEvoRun.cs
84634E1A607A: I want to at least comment all those prints too ...

src/auto-evo/AutoEvoExploringTool.cs
84634E1A607A: I'm not sure whether GameWorld will go on without a player species...

src/auto-evo/AutoEvoExploringTool.cs
84634E1A607A: This doesn't consume much performance but saves quite a bit code and in case more config will be there.

src/auto-evo/AutoEvoExploringTool.cs
84634E1A607A: This button is enabled only when auto-evo is paused or not running.

src/auto-evo/AutoEvoExploringTool.cs
84634E1A607A: I think this makes the strange formatSafe clearer.

src/auto-evo/AutoEvoExploringTool.cs
84634E1A607A: Copied from freebuild editor (

src/auto-evo/AutoEvoExploringTool.cs
84634E1A607A: Can't do this, freebuild generates random species that make the tree a real mess.

84634E1A607A:
This is as far as I can reduce the duplication...

src/auto-evo/AutoEvoExploringTool.cs
84634E1A607A: There's no option to unset player species.

84634E1A607A:
> Once this is merged I'll open separate issues about the evolution tree performance
Sure

> and the multi generation runs with a single button.
Implemented.

hhyyrylainen:
It's a nice addition to run many generations at once, but I actually meant this part of the auto-evo tool issue:
> A key feature for developing auto-evo with this tool would be to have a mode to run like 100 auto-evo runs in a row and collect some overall statistics. I was thinking that an option like "auto reset after 50 generations" and a toggle button to "run generations" would get the first part done.

So basically instead of just running a single game world for some number of steps. You'd run one world for 50 generations, collect some statistics, then start a new world and run for 50 generations. And you'd do that for like a total of 50 to 100 times to get some real statistics to see how well a change to auto-evo works.

84634E1A607A:
What's the "overall stat" then? For me 50 generation takes at least 5 min

src/auto-evo/AutoEvoExploringTool.cs
hhyyrylainen: See my later comments that make this specific review comment irrelevant. If this wasn't irrelevant I would have asked you to write such methods, but turns out we don't need that.

src/auto-evo/AutoEvoExploringTool.cs
hhyyrylainen: Still other places in the game have bothered (for example the options menu) to set individual callbacks for controls, but as I said I'll allow it here but I wanted to say I don't consider this a very clean coding practice.

hhyyrylainen:
> What's the "overall stat" then? For me 50 generation takes at least 5 min

Things like average number of species per patch, how many organelles the largest species has, how diverse energy sources are for species. Basically anything that might make it easy to say "yes, this change makes auto-evo work better".
It's a lot of work to design those and you haven't started on them which is the reason why I suggested I'll open a separate issue for that.

src/auto-evo/AutoEvoExploringTool.cs
hhyyrylainen: I have not been keeping up anywhere with showing the actual text that a translation uses. Only a few places one older developer who I think is now inactive used. I think that's a huge antipattern because the text is changed on weblate so no one is going to update the code comments which will become outdated, and outdated comments are worse than no comments.

src/auto-evo/AutoEvoExploringTool.cs
hhyyrylainen: Well this is the wrong comment anyway to be in this context.

This is the code there for reference:
```
            // Start freebuild game
            editor.CurrentGame = GameProperties.StartNewMicrobeGame(new WorldGenerationSettings(), true);
```
How you are using the code is *completely* different, it doesn't make sense to keep the comment as is here as it is incorrect.

src/auto-evo/AutoEvoExploringTool.cs
hhyyrylainen: Okay, then, you need to transform the game into a freebuild state later either before entering the game as the player or just after this line. Might need to add a new method if there isn't one already, and that should be commented that that method specifically won't create the freebuild starting species.

hhyyrylainen: I see you did this already in a later commit.

84634E1A607A:
> Things like average number of species per patch, how many organelles the largest species has, how diverse energy sources are for species. Basically anything that might make it easy to say "yes, this change makes auto-evo work better". It's a lot of work to design those and you haven't started on them which is the reason why I suggested I'll open a separate issue for that.

Let's leave it for later.

src/gui_common/PhotographPreview.cs
```diff
@@ -0,0 +1,78 @@
...
+///       <item>
+///         When to update preview by calling UpdatePreview()
+///       </item>
+///       <item>
+///         What to preview by overriding SetupImageTask()
+///       </item>
+///     </list>
+///   </para>
+/// </remarks>
+public abstract class PhotographPreview : Control
```
hhyyrylainen, line 19:
```suggestion
public abstract class PhotographablePreview : Control
```
I'd rename it like that to make it clear that this is not for any "photo" previews but previews for things the game considers "photographable"

src/auto-evo/EvolutionaryTree.cs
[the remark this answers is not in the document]
hhyyrylainen: The fix you did here seems to have gotten reverted...

src/microbe_stage/CellHexPhotoBuilder.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System;
+using Godot;
+
+public class CellHexPhotoBuilder : Spatial, IPhotographable
```
hhyyrylainen, line 4:
```suggestion
public class CellHexesPhotoBuilder : Spatial, IPhotographable
```
I think that kind of name would be a bit clearer that it's not a "cell hex" but "cell hexes" (the hexes the cell consists of) photo builder.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -417,6 +417,29 @@ public int MicrobeHexSize
         (IsMulticellularEditor ? Constants.MULTICELLULAR_EDITOR_COST_FACTOR : 1.0f) *
         Editor.CurrentGame.GameWorld.WorldSettings.MPMultiplier;
 
+    public static void InitOrganelleModel(SceneDisplayer organelleModel, OrganelleTemplate organelle)
```
hhyyrylainen, line 420: This doesn't literally init the model builder, this just sets the position and scale for it.

hhyyrylainen, line 420:
```suggestion
    public static void UpdateOrganelleDisplayerTransform(SceneDisplayer displayer OrganelleTemplate organelle)
```

src/microbe_stage/CellHexPhotoBuilder.cs
```diff
@@ -0,0 +1,112 @@
...
+        var hexScene = GD.Load<PackedScene>("res://src/microbe_stage/editor/EditorHex.tscn");
+        var hexMaterial = GD.Load<Material>("res://src/microbe_stage/editor/ValidHex.material");
+        var modelScene = GD.Load<PackedScene>("res://src/general/SceneDisplayer.tscn");
+
+        this.QueueFreeChildren();
+
+        if (Species == null)
+            throw new InvalidOperationException("Species is ont initialized");
+
+        var organelleLayout = Species.Organelles.Organelles;
```
hhyyrylainen, line 57: The duplication reduction is still not perfect so could you at least add a TODO comment here pointing out the duplication with the other place? And also please add a comment in that editor about the duplication that if the code there is changed CellHexesPhotoBuilder also needs those changes.

Kasterisk (changes requested):
Some general things.

The gap between this bar need to have uniformity, it's mildly OCD inducing.



Upon entering a game, the dataset for species population graph seems broken, it looks like multiple points are bunching up on the 0 myr mark.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -0,0 +1,732 @@
...
+        Ready,
+        Running,
+        Paused,
+    }
+
+    public override void _Ready()
+    {
+        base._Ready();
+
+        TransitionManager.Instance.AddSequence(ScreenFade.FadeType.FadeIn, 0.1f, null, false);
```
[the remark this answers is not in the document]
hhyyrylainen: Well, this switches to a separate scene entirely, so I thought this is a fine enough way to do it.

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -0,0 +1,229 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using AutoEvo;
+using Godot;
+using Godot.Collections;
```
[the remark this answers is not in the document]
hhyyrylainen: Isn't that only if the Array name would conflict? I think here is the system collections is not imported, there's no name conflict so that kind of using statement is not needed. Well I let Rider handle that but that might be the case why that code is not generated here.

src/general/MainMenu.cs
```diff
@@ -340,6 +345,16 @@ private void FreebuildEditorPressed()
...
     }
 
+    private void OnAutoEvoExploringPressed()
+    {
+        GUICommon.Instance.PlayButtonPressSound();
+
+        autoEvoExploringButton.Disabled = true;
+
+        TransitionManager.Instance.AddSequence(ScreenFade.FadeType.FadeOut, 0.1f,
+            () => { SceneManager.Instance.SwitchToScene("res://src/auto-evo/AutoEvoExploringTool.tscn"); }, false);
```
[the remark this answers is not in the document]
hhyyrylainen: I think the exploring tool is a kind of a separate thing, so at least to me it makes a bit more sense to have it separately from the main menu, after all entering the freebuild editor is kind of the same thing of jumping to a different game state. And also I think it'll make the main menu a fractionally lighter scene to not have to put the auto-evo tool in there, which I am in favour of because it's not really something I think most players would find that useful.

src/gui_common/DraggableScrollContainer.cs
```diff
@@ -7,7 +7,7 @@
 /// </summary>
 public class DraggableScrollContainer : ScrollContainer
 {
-    [Export]
+    [Export(hintString: "If set to null, the first child is used.")]
```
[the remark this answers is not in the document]
hhyyrylainen: What does it do then?

src/microbe_stage/CellHexesPhotoBuilder.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        get => species;
+        set
+        {
+            species = value;
+            radiusDirty = true;
+        }
+    }
+
+    public void ApplySceneParameters(Spatial instancedScene)
```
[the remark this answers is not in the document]
hhyyrylainen: instanced sounds fine to me, this is probably a copy of what photographable classes have. Though now I'm a bit wondering why this is nor marked "override".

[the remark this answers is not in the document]
hhyyrylainen: Oh yeah, IPhotographable is an interface so of course this doesn't have the override keyword. This should definitely use the same parameter naming as the interface this implements.

src/auto-evo/EvolutionaryTree.cs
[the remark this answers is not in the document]
Kasterisk: I see, but a couple lines below there is a redundant qualifier for some reason.

e.g:
`private readonly System.Collections.Generic.Dictionary<uint, string> speciesNames = new();`

src/microbe_stage/CellHexesPhotoBuilder.cs
```diff
@@ -0,0 +1,114 @@
...
+    private void BuildHexStruct()
+    {
+        var hexScene = GD.Load<PackedScene>("res://src/microbe_stage/editor/EditorHex.tscn");
+        var hexMaterial = GD.Load<Material>("res://src/microbe_stage/editor/ValidHex.material");
+        var modelScene = GD.Load<PackedScene>("res://src/general/SceneDisplayer.tscn");
+
+        this.QueueFreeChildren();
+
+        if (Species == null)
+            throw new InvalidOperationException("Species is ont initialized");
```
hhyyrylainen, line 55:
```suggestion
            throw new InvalidOperationException("Species is not initialized");
```
one more typo I spotted

src/auto-evo/EvolutionaryTree.cs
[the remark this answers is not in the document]
hhyyrylainen: I think Rider picks the style that results in the least amount of long qualifiers. So if Godot.Array is used more than once, it would pick to use the long name for the system collection as that results in overall less characters. At least that's my understanding of the logic.

src/gui_common/DraggableScrollContainer.cs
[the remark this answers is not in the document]
Kasterisk: Very little documentation on it, here's the only mention of it I can find on the docs: [link](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_exports.html#grouping-properties). From that I can only infer it has no relation to tooltips. Furthermore, this functionality was just only recently [added to 4.0](https://godotengine.org/qa/313/tooltip-documentation-custom-exported-variables-gdscript?show=90601#a90601).

src/auto-evo/EvolutionaryTree.cs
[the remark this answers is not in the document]
84634E1A607A: I don't know, but this is what reSharper and CI wants

src/general/MainMenu.cs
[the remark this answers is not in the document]
84634E1A607A: Yeah, and with this code I don't need to manually clean exploring tool when it exits.

src/microbe_stage/CellHexesPhotoBuilder.cs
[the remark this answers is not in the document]
Kasterisk: The interface also has the same naming it looks like. Notwithstanding, instance as a verb in the future engine versions would officially be instantiate so it would be less work to use this terminology now than later.

Though this is bike-shedding so I digress, I see you're fine with it so I guess both terms can be used interchangeably going forward.

src/general/MainMenu.cs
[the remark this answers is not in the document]
hhyyrylainen: Actually related to a recent bug: https://community.revolutionarygamesstudio.com/t/error-when-trying-to-load-save/5139/16 could you make it so that if an auto-evo run is in progress when the auto-evo tool exits the tree (put code in tree exit callback) the auto-evo run is aborted automatically?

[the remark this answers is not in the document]
84634E1A607A: Sure.

src/microbe_stage/CellHexesPhotoBuilder.cs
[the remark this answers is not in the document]
hhyyrylainen: I personally always will end up writing instanced, it just sounds better to me but I'll also allow the other term to be used. If someone really wants to nitpick there might be some meaning difference but I'm not going to really care.

Kasterisk:
Another feature that would be really nice to have for the record: the selected generation affects the evolutionary tree so we can see through the evolution of the tree from the first generation to the last.

84634E1A607A:
Can't you figure that out just by looking at the final tree? It is a TREE.

84634E1A607A:
And I think that may cause confusions, that when you selected 0 for example, you can't select anything later on the tree. Besides, selecting a tree node will cause generation and species selection to update.

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -0,0 +1,226 @@
...
+            TIMELINE_LINE_THICKNESS, true);
+
+        for (int i = 0; i <= latestGeneration; i++)
+        {
+            DrawLine(new Vector2(LEFT_MARGIN + i * GENERATION_SEPARATION + treeNodeSize.x / 2, TIMELINE_LINE_Y),
+                new Vector2(LEFT_MARGIN + i * GENERATION_SEPARATION + treeNodeSize.x / 2,
+                    TIMELINE_LINE_Y + TIMELINE_MARK_SIZE),
+                Colors.Cyan, TIMELINE_LINE_THICKNESS, true);
+
+            var localizedText = i + " " + TranslationServer.Translate("MEGA_YEARS");
```
hhyyrylainen, line 77: This line is still the wrong version, this is my last code related comment. Everything else has now been fixed.

hhyyrylainen (changes requested):
I did some more playtesting and looked at the code again, I think one of my earlier comments that was fixed already was reverted back. So from my side just need to fix that one thing and I think this is good enough to merge, further improvements and the extra features for auto-evo we can leave for later.

src/auto-evo/EvolutionaryTree.cs
84634E1A607A: I'm not familiar with the number format you use in the code, so I cannot use it like that without doing some 'i * 1e6' thing.

If you know how to make it work correctly without multiplying 1e6, you may change that yourself. It wouldn't be a tough thing.

locale/en.po
```diff
@@ -1791,71 +1999,66 @@ msgstr "{0} concentrations have decreased by {1}"
...
-#: ../src/general/MainMenu.tscn:459 ../src/general/NewGameSettings.tscn:1012
-#: ../src/general/OptionsMenu.tscn:1261 ../src/general/PauseMenu.tscn:263
-#: ../src/microbe_stage/gui/ExtinctionBox.tscn:76
-#: ../src/modding/ModManager.tscn:81 ../src/saving/NewSaveMenu.tscn:110
-#: ../src/saving/SaveManagerGUI.tscn:120
-msgid "BACK"
-msgstr "Back"
+#: ../src/general/MainMenu.tscn:293
+msgid "AUTO_EVO_EXPLORING_TOOL"
+msgstr "Auto-evo Exploring Tool"
```
Oliveriver, line 2061: One small thing: 'evo' should probably be capitalised here, so this should be 'Auto-Evo Exploring Tool'.

src/auto-evo/EvolutionaryTree.cs
hhyyrylainen: It shows correctly in the editor, I'm pretty sure this format specifier removes the scientific notation:
```
Text = string.Format(CultureInfo.CurrentCulture, "{0:#,##0,,}", data.TimePeriod) + " "
                    + TranslationServer.Translate("MEGA_YEARS"),
```
`{0:#,##0,,}` is the part you need. I think @Kasterisk wrote that initially and can clarify.

84634E1A607A: Anyway.

hhyyrylainen: Hang on now that I read about this, you are using `i` as the generation number? That doesn't map to our years fully. I think maybe you should just use `i.ToString(CultureInfo.CurrentCulture)` without any suffix or something like that? If you add a suffix it should be like "G" for "generation" as `i` is the generation number and not a year as I thought before.

84634E1A607A: https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-numeric-format-strings

Ah, generation i is 100 * i Myr

hhyyrylainen (changes requested):
A bit better but this will break the second someone decides that the editor cycle is no longer 100 million years. Also the left side being cut off problem is back. I think really having the numbers like 0, 1, 2 would be fine enough, because getting the elapsed time from the game world would be a bit tricky.

src/auto-evo/EvolutionaryTree.cs
hhyyrylainen: Currently, but not necessarily forever, changing the game world time elapse code will happen at some point and then the auto-evo tool will once again show incorrect values.

84634E1A607A: Note that evolutionary tree will be a main game thing, so the world time will eventually be applied.

hhyyrylainen: I guess yeah. Well the solution where the 0 is printed as "000 Myr" is not very nice looking to me.

84634E1A607A: That's purely accident...

hhyyrylainen (approved):
The number is good enough now.

Kasterisk (approved):
Looks good! Thanks for working on this.

PR #3456, opened 2022-06-29
Fix membrane color not applying in multicell editor

### LINKED ISSUE #3367: Multicellular membrane colour edit doesn't apply if only it was changed
seems to be a regression from the dynamic MP.

Kasterisk:

Put membrane color changing through the editor action system so this change registers to the editor and work with undo/redo.

**Related Issues**

Fixes #3367 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/gui_common/TweakedColourPicker.cs
```diff
@@ -194,7 +194,6 @@ public new Color Color
                 return;
 
             base.Color = value;
-            EmitSignal("color_changed", value);
```
Kasterisk, line 197: This had to be removed as this causes repeated signal emissions, causing problems. Also redundant because it should already be emitted under the hood.

hhyyrylainen: @84634E1A607A I think you made the tweaked colour picker, so do you have anything to say about this change?

84634E1A607A: I recall I added this because sometimes the native color_changed signal is not emitted. (Perhaps when you change the Color property directly)

src/microbe_stage/editor/action_data/ColourActionData.cs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public override bool WantsMergeWith(CombinableActionData other)
+    {
+        return other is ColourActionData;
+    }
+
+    protected override int CalculateCostInternal()
+    {
+        // Changing membrane color has no cost
```
hhyyrylainen, line 22:
```suggestion
        // Changing membrane colour has no cost
```

src/microbe_stage/editor/action_data/ColourActionData.cs
```diff
@@ -0,0 +1,77 @@
...
+            }
+
+            PreviousColour = colourChangeActionData.PreviousColour;
+            return;
+        }
+
+        NewColour = colourChangeActionData.NewColour;
+    }
+
+    protected override ActionInterferenceMode GetInterferenceModeWithGuaranteed(CombinableActionData other)
```
hhyyrylainen, line 56: These methods look like they are in different order than other actions, please reorder them to be in a consistent order.

hhyyrylainen (changes requested):
Besides the need for playtesting to ensure the "color_changed" signal removal doesn't cause issues, I have just one tiny nitpick about the code style (and one typo in a comment).

src/microbe_stage/editor/action_data/ColourActionData.cs
Kasterisk: Right, was put this way by my IDE.

src/gui_common/TweakedColourPicker.cs
Kasterisk: Hm yeah, just isolated and tested this and it indeed doesn't automatically actually get emitted....

Kasterisk: Actually maybe it's better to not override this property, instead separate it to a new method as we sometimes want Color to not emit a signal?

hhyyrylainen: Probably yeah, but then it should be called here as well... or?

Kasterisk: Yes, so we will use the base `Color` and a `SetColor` method that sets `Color` and emits signal, this property should always be used in TweakedColorPicker, `Color` can be for other places.

@84634E1A607A What was actually the reason for emitting the signal by the way?

84634E1A607A: > I think it should since that also calls this overriden property.

Yes. (I said once, when this property is not overriden.) And this is why I chose to emit a signal here.

src/microbe_stage/editor/action_data/ColourActionData.cs
Kasterisk: Fixed.

src/gui_common/TweakedColourPicker.cs
84634E1A607A: Yes, but I didn't playtest so I'm not sure all functions are still correct.

84634E1A607A: I'd prefer a `if (Color == colour) return;` added

Oliveriver:
Tested and it seems to work fine in both unicellular and multicellular settings for me.

hhyyrylainen (approved):
This is probably fine now, but I'll give a bit more time for testing.

Issue #3357, opened 2022-05-20
Make organelle chunks stop organelle animations


hhyyrylainen:
Make organelle (cilia, flagella) animations stop when dropped as cell corpse chunks or marine snow

hhyyrylainen:
I fixed this along with the ECS refactor.

PR #3472, opened 2022-07-04
Cell spawn Easter egg

Oliveriver:

Adds a small chance for Spore-like cells to spawn if the player has enabled Easter eggs.

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)


assets/models/easter_eggs/SporeCell.tscn
```diff
@@ -0,0 +1,505 @@
...
+"blend_shape_data": [  ],
+"format": 2194903,
+"index_count": 1296,
+"material": ExtResource( 2 ),
+"primitive": 4,
+"skeleton_aabb": [ AABB( -1.46845, -1.40214, -2.25429, 2.93691, 2.80427, 2.30923 ), AABB( -1.46845, -1.40214, -2.25429, 2.93691, 2.80427, 3.14247 ), AABB( -1.24065, -1.36023, -0.88309, 2.48131, 2.65776, 3.04678 ), AABB( -0.472767, -0.412523, 1.91071, 0.472777, 0.825046, 0.252986 ), AABB( 0, 0, 0, -1, -1, -1 ), AABB( 0, -0.412523, 1.91071, 0.472767, 0.825046, 0.252976 ), AABB( 0, 0, 0, -1, -1, -1 ), AABB( -1.74808, -1.40214, -2.09403, 2.34498, 2.80427, 4.08859 ), AABB( -2.09104, -1.0902, -1.62873, 1.40571, 2.45044, 2.71561 ), AABB( -2.61336, -0.36634, 0.315248, 1.00556, 0.670542, 0.842363 ), AABB( 0, 0, 0, -1, -1, -1 ), AABB( -0.455002, -1.40214, -2.00178, 2.20308, 2.80427, 3.99635 ), AABB( 0.685331, -1.36023, -0.993029, 1.40571, 2.72047, 2.07991 ), AABB( 1.6078, -0.36634, 0.315248, 1.00556, 0.670541, 0.842363 ), AABB( 0, 0, 0, -1, -1, -1 ), AABB( -0.907669, -1.23504, -2.28006, 2.14519, 2.47007, 0.670185 ) ],
```
hhyyrylainen, line 19: CI is complaining about this embedded data, basically the data needs to be in a separate .mesh or a .glb file for Git LFS to work.

Oliveriver:
I managed to switch to using an external resource. Making the animation still play was a slight pain, but all good now.

I've also tested and could find a Spore cell after a few minutes using the super speed cheat to spawn as much as possible. So they are spawning, and they're spawning rarely, which I think is about right.

adQuid:
As painful as it would be, you might want to try a couple hours of playing at normal speed, since max cheat speed is just 5X normal IIRC. The joke might get less funny if one shows up every 20-30 minutes of gameplay.

Oliveriver:
> The joke might get less funny if one shows up every 20-30 minutes of gameplay.

Are you saying that's too often or not often enough? Because one Spore cell every 30 minutes sounds fine to me.

hhyyrylainen:
I think seeing an easter egg once or twice per hour is about right. At least when considering the later stages as well, we should eventually end up with easter eggs for all of them that would appear about once per hour.

hhyyrylainen:
Would it be possible to rename the assets/models/easter_eggs/Material.material etc. files so that they are specific to the model in question? I guess if it isn't that's the downside of the .glb file, and probably means it needs to be split up with the individual parts saved as separate resources (to have git LFS work). Or maybe the graphics team could do  that?

simulation_parameters/microbe_stage/biomes.json
```diff
@@ -97,6 +97,22 @@
                             "amount": 60000
                         }
                     }
+                },
+                "sporeCell": {
+                    "name": "SPORE_CELL",
```
hhyyrylainen, line 102: I think maybe we should avoid the word "Spore" being in the game at all, so this could be something like "GOOGLY_EYED_CELL" or some kind of nickname like that.

simulation_parameters/microbe_stage/biomes.json
```diff
@@ -97,6 +97,22 @@
...
+                    "meshes": [
+                        {
+                            "scenePath": "res://assets/models/easter_eggs/SporeCell.tscn",
```
hhyyrylainen, line 105: And that would also apply to the file names.

src/microbe_stage/FloatingChunk.cs
```diff
@@ -221,6 +233,12 @@ public override void _Ready()
             chunkMesh = graphicsNode.GetNode<MeshInstance>(ModelNodePath);
         }
 
+        if (!string.IsNullOrEmpty(AnimationPath))
+        {
+            animation = graphicsNode.GetNode<AnimationPlayer>(AnimationPath);
+            animation.GetAnimation(animation.CurrentAnimation).Loop = true;
```
hhyyrylainen, line 239: For cilia chunks, I'd actually think it should not loop, but this is the default behaviour now, so no need to specifically code that now.
But if you actually have to break up the glb file, you should be able to edit and force the animation to autoplay in looping mode, so this like could be commented out with a TODO comment to make type specific code to stop animations of specific chunks. https://github.com/Revolutionary-Games/Thrive/issues/3357

hhyyrylainen (changes requested):
Overall the approach looks fine. Here's just comments to some of the specifics. Also I mentioned before but being able to name Material_001, Material_002 etc. something more specific would be extremely good. The second we add a second model that wants to use materials with those names we'll be in trouble anyway, so would be good to fix those too generic names now.

Oliveriver:
> Would it be possible to rename the assets/models/easter_eggs/Material.material etc. files so that they are specific to the model in question? I guess if it isn't that's the downside of the .glb file, and probably means it needs to be split up with the individual parts saved as separate resources (to have git LFS work). Or maybe the graphics team could do that?

I was able to do this when I converted the .glb to a .tscn, but if we're using the .glb as an external resource, I don't think I can without editing the .glb itself. I'll take a look at that.

hhyyrylainen:
> without editing the .glb itself

Yeah, that's probably the key. In the future we probably need to ask artists to set more sensible names for the materials they use in Godot.

Oliveriver:
Alright, I think I've covered everything now. I had to fix a bug caused by the new engulfment system in FloatingChunk, and I noticed a small tweak I'd missed in the new game settings in PatchManager. So maybe give those a quick look.

Otherwise, I propose leaving more substantial testing to the release candidate. We should hopefully get a clear view of how often people are finding these with a much larger search run in parallel while testing other things.

hhyyrylainen (approved):
The code now looks good to me.

hhyyrylainen:
I guess I'll merge this now and we'll leave more indepth testing for later.

PR #3480, opened 2022-07-07
Basic cilia attraction mechanism

Kasterisk:

Partially based on Thim's prototype. Each cilia now have their own "force field" that can pull things towards it, this field is made not too big to make it not too OP.

Related parameters are freely accessible in Constants for easier balancing.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


RacerBG:
> At least the cilia tooltip would need to explain this new surprise mechanic. Which we'll need to edit out then afterwards when this is properly done, which means we waste translators time.

I can update the tooltip, don't worry about that. Personally, I'm in favor of this feature.

84634E1A607A:
My concern:
- When cilia grabs a cell with a lot of pilus, will I die quickly of pilus damage?
- When cilia crabs a cell bigger than you, will you eventually get engulfed quickly?

Kasterisk:
> When cilia grabs a cell with a lot of pilus, will I die quickly of pilus damage?

Yes.

> When cilia crabs a cell bigger than you, will you eventually get engulfed quickly?

Probably, if the cilia can manage to pull it.

> If this bioprocess consumes atp, can I disable it somehow in the stage?

Does not cost ATP.

Oliveriver:
I'm of the opinion that this should wait until after 0.5.9. It's a cool feature, but cilia on their own are a cool feature for everyone who hasn't had access to them for months (i.e. not us). Adding this now probably adds too many layers of confusion for new players.

Oliveriver:
Is this still planned for 0.5.10? Seems a silly thing to miss considering how close it was for 0.5.9.

hhyyrylainen:
Well I no longer have time to dedicate to making the part upgrade GUI part for this... so someone else will need to work on that if this is to be done. This might get pushed back to the release after 0.5.10... I'll switch this to that milestone once I create it.

MayeulC:
> > When cilia crabs a cell bigger than you, will you eventually get engulfed quickly?
> 
> Probably, if the cilia can manage to pull it.

Out of curiosity, does this respect Newton's third law? ie, do you get pulled towards it as well, with the more massive cell acquiring less velocity (due to the second law).

Kasterisk:
I actually just saw your comment now so sorry for the delay... The answer is no, player's microbe is completely ignored in the process currently. Even if it's the case, technically you should only be able to pull objects equal or less than your mass I guess? I don't know, the physics would need some balancing.

I do think that would allow cilia being made as some sort of propulsion mechanism for little cells though, not saying it shouldn't be possible, the upcoming complementary organelle upgrades feature would allow specialized cilia anyway so the cilia's role would not just for sucking things in.

hhyyrylainen:
I'll actually probably have time next week to start on the general organelle selectable upgrades system, so this PR might finally get merged then.

MayeulC:
> The answer is no, player's microbe is completely ignored in the process currently (for gameplay reasons).

Aww, that's a bit of a bummer, I like double-edged swords.

> Even if it's the case, technically you should only be able to pull objects equal or less than your mass I guess? I don't know, the physics would need some balancing.

Not really, if you exert a force, technically they exert the same force on you (question of perspective). The less massive object is going to be pulled more, but that's just because it is faster to accelerate.
That means: it's OK for you to make something sticky, but don't complain if you get stuck to something bigger than you.

hhyyrylainen:
@Kasterisk what's with the collision layer changes in this PR? I tried to read the code but I didn't manage to understand why that is changed. Obviously a new weapon the player can use, should also be usable by the AI and for that reason I find it highly suspicious that a collision layer specifically for the player was added.

Kasterisk:
> player's microbe is completely ignored in the process currently (for gameplay reasons).

That was the reason if I recall my own changes correctly.

hhyyrylainen:
I'm planning on starting working on the organelle upgrades now, and as the AI cannot upgrade their organelles they won't spontaneously evolve it (so it can only appear after the player has added it). So I think it'd be better to remove that limitation from this to make this a lot smaller change.

hhyyrylainen:
Can you make those changes, or do you mind if I go pretty heavy handed on this now and then merge?

Kasterisk:
Yes, just a moment.

Kasterisk:
Okay done.

src/microbe_stage/organelle_components/CiliaComponent.cs
```diff
@@ -1,6 +1,9 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
 using Godot;
 
+[SuppressMessage("Design", "CA1001:Types that own disposable fields should be disposable",
```
hhyyrylainen, line 5: Oh and also now this suppression should be removed and the variables ignored similarly to all the other code, which I did in a big PR yesterday).

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -1500,17 +1500,8 @@ private void IngestEngulfable(IEngulfable target, float animationSpeed = 2.0f)
         target.HostileEngulfer.Value = this;
         target.PhagocytosisStep = PhagocytosisPhase.Ingestion;
 
-        // Disable collisions
-        body.CollisionLayer = 0;
-        body.CollisionMask = 0;
-
         body.ReParentWithTransform(this);
 
-        var originalRenderPriority = target.RenderPriority;
```
hhyyrylainen, line 1509: It seems the original render priority no longer gets stored and this logic is also different. Why does this need to change to make cilia pull work?

src/microbe_stage/organelle_components/CiliaComponent.cs
Kasterisk: And done.

src/microbe_stage/Microbe.Contact.cs
Kasterisk: Have you actually seen this?
https://github.com/Revolutionary-Games/Thrive/blob/006237d9908dd276daefc0c16b1854343b83c410/src/microbe_stage/Microbe.Contact.cs#L1549-L1561
The logic does not change. Just the ordering of statements are sorted properly now and the original render priority are now _stored_ in the EngulfedObject instance and to make the following possible:

https://github.com/Revolutionary-Games/Thrive/blob/006237d9908dd276daefc0c16b1854343b83c410/src/microbe_stage/Microbe.Contact.cs#L1879-L1881

hhyyrylainen: I see, now that you point these out, I think I understand this change now.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -1954,5 +1954,7 @@ public EngulfedObject(IEngulfable @object, Endosome phagosome)
         public (Vector3 Translation, Vector3 Scale, Vector3 EndosomeScale) InitialValuesToLerp { get; set; }
         public Vector3 OriginalScale { get; set; }
         public int OriginalRenderPriority { get; set; }
+        public uint OriginalCollisionLayer { get; set; }
+        public uint OriginalCollisionMask { get; set; }
```
hhyyrylainen, line 1958: Isn't there a potential save incompatibility thing here?
I think these should have default values set to what makes sense for microbes (and a comment added explaining that the values are here for save compatibility).

src/microbe_stage/organelle_components/CiliaComponent.cs
```diff
@@ -96,6 +114,30 @@ protected override void CustomAttach()
...
+            GravityDistanceScale = Constants.CILIA_PULLING_FORCE_FALLOFF_FACTOR,
+            Gravity = Constants.CILIA_PULLING_FORCE,
+            CollisionLayer = 0,
+            CollisionMask = microbe.CollisionMask,
+        };
+
+        var sphereShape = new SphereShape { Radius = Constants.CILIA_PULLING_FORCE_FIELD_RADIUS };
+        attractorAreaShape = new CollisionShape { Shape = sphereShape };
+        attractorArea.AddChild(attractorAreaShape);
+        organelle.OrganelleGraphics.AddChild(attractorArea);
```
hhyyrylainen, line 133: Is it safe in Godot to add an Area to a scaled Node? Adding sub collision shapes or rigid bodies will break stuff. So I worry that this might be a problem as well. To workaround that this area would need to be parented to the base microbe (with potentially an offset if the area is wanted to match where the cilia is actually at).

src/microbe_stage/organelle_components/CiliaComponent.cs
```diff
@@ -81,6 +94,11 @@ public override void UpdateSync()
             // better to do things correctly here as this is newer code...
             SetSpeedFactor(targetSpeed);
         }
+
+        if (attractorAreaShape != null)
+        {
+            attractorAreaShape.Disabled = organelle!.ParentMicrobe!.State != Microbe.MicrobeState.Engulf;
```
hhyyrylainen, line 100: Should the pull effect be disabled if this microbe this cilia is on gets engulfed? Otherwise wouldn't the engulfing cell move around with the extra gravity pull area causing quite many potential issues?

src/microbe_stage/organelle_components/CiliaComponent.cs
Kasterisk: This has a nice effect that the Area's radius scales as the organelle grows, perhaps we can keep this behavior even after moving the Area out of OrganelleGraphics?

hhyyrylainen: It would require manual updating code for it, but I'm not against having that as a feature. Though the effect is very temporary as organelles grow pretty fast, that it isn't super important (and with the eventual change to organelles no longer duplicating over time, this feature will just disappear).

src/microbe_stage/organelle_components/CiliaComponent.cs
Kasterisk: I think cells have their engulfment mode automatically disabled (thus disables pulling) when they're engulfed so this conditional kind of kills two birds with one stone already, but an extra check there would still be nice.

hhyyrylainen: Oh yeah, that's probably true. An explicit check might be better anyway, as we just found out I did not immediately think about that.

Kasterisk:
I ended up refactored the attraction logic a bit which also fixed a problem where the cilia pulling no longer works after being re-enabled from deactivation.

src/microbe_stage/organelle_components/CiliaComponent.cs
Kasterisk: Done.

src/microbe_stage/organelle_components/CiliaComponent.cs
Kasterisk: I've kept this behavior in the latest commit, it's not that hard to add anyways.

hhyyrylainen (approved):
The code change looks good now. I haven't playtested this or checked if the save compatibility will actually really work now. I'll give some time for playtesting / other people checking this. But even if that doesn't happen I'll probably merge this first thing on Monday morning when I start working again.

Issue #3432, opened 2022-06-18
Flagella ATP cost inconsistencies between editor sessions


Oliveriver:
Discovered this while testing another PR. If I add flagella to my cell, the ATP cost displayed is different to the ATP cost displayed the next time I enter the editor, even with no changes. Even weirder, this only appears to happen with certain cell builds.

Steps to reproduce:

* Start a new game and enter the editor.
* Replace the starting cytoplasm with a metabolosome and add a flagellum. ATP balance should be 8/8 and give an ATP too low warning.
* Return to the environment, then enter the editor again.
* ATP balance is now 8/4 with no ATP too low warning, and the flagellum cost is gone from the balance bars.

![image](https://user-images.githubusercontent.com/7430433/174436628-6d37d327-fe52-4ab9-a302-c20e63c2ab25.png)

![image](https://user-images.githubusercontent.com/7430433/174436635-7248282c-4452-4e8b-988a-fe8621f49577.png)

hhyyrylainen:
Did you check if the hexes in your species moved between the editor cycles? Hexes are repositioned if the center of your organism is not exactly at 0,0 when you exit. If that was the case, then this is a duplicate of: https://github.com/Revolutionary-Games/Thrive/issues/2094

Oliveriver:
This has nothing to do with speed calculations. The ATP cost of the flagellum just disappeared.

hhyyrylainen:
Ah, the *cost*. I didn't pay enough attention. 
In the second screenshot it looks the flagellum is entirely missing, which I missed entirely. Could you provide a save and/or screenshots of the species before and after the problem?
I still think that something like flagella orientation after the automatic repositioning could be the cause. That is because the flagella position affects if it can push the cell forward or not. If the flagellum could push the cell forward before, it'll have an ATP cost, but later it got into a position where it can't push forwards, it wouldn't be in the ATP cost.

Oliveriver:
Here's a save that should have the offending cell in the editor: https://drive.google.com/file/d/1eeuoaXvlu1-4VdQUmF86FmSYI3Mvxr4Z/view?usp=sharing

You're probably on the right track about positioning - if I move the flagellum anywhere else during the editor session above, the editor suddenly realises what the correct ATP balance should be. Pressing undo however doesn't revert that.

hhyyrylainen:
[...]

Issue #3109, opened 2022-02-22
Bias external organelle visual position based on editor rotation when at origin


hhyyrylainen:
seems like when the pilus is at 0,0 coordinate which switches over to a predetermined angle (instead of trying to calculate it) for it, which, because the feature was originally made for the flagellum, points it backwards. So this seems like a place where possible confusion for the player can occur if they place a pilus at 0,0 pointing forward in the editor but when they enter the game, the pilus then suddenly points backwards

https://community.revolutionarygamesstudio.com/t/0-5-7-feedback-thread/4226/5?u=hhyyrylainen

hhyyrylainen:
An alternative approach to do this: https://github.com/Revolutionary-Games/Thrive/issues/3620

hhyyrylainen:
Also maybe should bias the orientation even when not at the origin to for example make sideways pointing flagella less likely to appear when not intended.

PR #2243, opened 2021-05-19
Editor speed, flagellum thrust based on flagellum orientation

luminarium:
* Updates the microbe's movement such that the thrust provided by each flagellum is based on its orientation (the direction it's facing), rather than its location relative to the center of the microbe. This makes the flagellum's behavior align with what one might intuitively expect, and allows making actual use of the ability to choose an organelle's orientation.
* Updates the microbe editor speed calculation to account for the above. This should also resolve the issue where changes to the actual speed resulting from the center of mass being changed aren't reflected in the editor's displayed speed. 

This may also close #2094 since with this change the speed calculation in the editor is no longer dependent on where the flagella are placed.

PR #3481, opened 2022-07-07
Fix flagella direction calculation

Oliveriver:

Fixes the calculation of flagella directions to remove ATP calculation inconsistencies.

**Related Issues**

Fixes https://github.com/Revolutionary-Games/Thrive/issues/3432

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
This changes the flagella calculation to use the orientation it is set to in the editor? Which means that this also isn't fully correct, in fact I think this might be slightly less accurate than the previous math, as that at least depended on the angle to the center of the cell, which while not perfect still kind of mimicked the angle that the generated membrane would result in.

See this old PR that had a similar idea: https://github.com/Revolutionary-Games/Thrive/pull/2243

hhyyrylainen:
For speed there's explicit code to assume flagella point backwards in that case. Though it's still not perfect: https://github.com/Revolutionary-Games/Thrive/issues/3109

Oliveriver:
I've switched to using the existing direction calculation, but I'm not 100% happy with it. I think this deserves more refactoring, but that's probably too risky a thing to do in the run up to release.

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -33,7 +33,8 @@ public static Vector3 MaximumSpeedDirection(IEnumerable<OrganelleTemplate> organ
 
     public static Vector3 GetOrganelleDirection(OrganelleTemplate organelle)
     {
-        return (Hex.AxialToCartesian(new Hex(0, 0)) - Hex.AxialToCartesian(organelle.Position)).Normalized();
+        // TODO: calculate this for organelles other than flagella
+        return MovementComponent.CalculateForce(organelle.Position, organelle.Definition.Components.Movement!.Momentum);
```
hhyyrylainen, line 37: I don't really like how the *force* is used here. I don't actually see how this can even fix the hex layout being offset, unless this is meant to fix something else...

Also a static helper method somewhere that is not as flagella specific and one that doesn't use momentum, should be done.

hhyyrylainen:
I'll remove this from any release milestones as this is pretty well stuck...

hhyyrylainen:

PR #3492, opened 2022-07-09
Deleted special region and merge it into the normal regions

### LINKED ISSUE #3505: Seed `69` causes an infinite loop when generating patch map

84634E1A607A:
Plus fixes #3505

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
It seems you made this PR against a slightly outdated version of master, so you'll need to merge master into this and solve the conflicts...

84634E1A607A:
I know. I wrote this a few hours ago. This pr **breaks** save compatibility and still have a few things to be done (but the few things basically doesn't affect save compatibility and gameplay experience.)

src/microbe_stage/Patch.cs
```diff
@@ -39,6 +40,32 @@ public Patch(LocalizedString name, int id, Biome biomeTemplate, PatchSnapshot cu
...
+        Bathypelagic = 5,
+        Abyssopelagic = 6,
+        Seafloor = 7,
+        Cave = 8,
+        IceShelf = 9,
+        Vents = 10,
+    }
+
+    [JsonProperty]
+    public BiomeTypes BiomeType { get; private set; }
```
hhyyrylainen, line 67: I think this should be moved to be after the Region property to better follow the order of constructor parameters (that has slight performance impact on json deserialization now). And makes more sense to me to have the ID and other identification properties before this.

84634E1A607A:
Functionality completely done.

RacerBG:
I noticed that if you are using the classical patch map the path towards the next patch is not highlighted. Otherwise it looks good.

84634E1A607A:
Isn't it always looking like this?

84634E1A607A:
Wait, this doesn't break save compatibility. Although the saves made in rc1 looks bad, it can still load and function.

src/microbe_stage/PatchMap.cs
```diff
@@ -25,15 +25,7 @@ public class PatchMap : ISaveLoadable
...
-    public IEnumerable<KeyValuePair<int, PatchRegion>> NormallyDrawnRegions =>
-        Regions.Where(r => !r.Value.UsesSpecialLinking);
-
-    /// <summary>
-    ///   These regions use special logic on how they are linked and positioned to other regions (and drawn)
-    /// </summary>
-    [JsonIgnore]
-    public IEnumerable<KeyValuePair<int, PatchRegion>> SpeciallyDrawnRegions =>
-        Regions.Where(r => r.Value.UsesSpecialLinking);
+    public Vector2 Center { get; private set; }
```
hhyyrylainen, line 28: I see that this is basically used only in one place? If so I think this should be auto computed property, and then in the map drawer you should just use a normal variable to compute this value once when needed. I think that'll be easier in the future as I suspect our maps won't be static forever so that just stops a bug where this is out of date.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -332,117 +270,258 @@ private Vector2 GetRightCornerPointOnMap()
...
+    ///   </para>
+    /// </remarks>
+    private int IntersectionCountWithPriority((Vector2[] Path, int Priority) tuple)
+    {
+        var path = tuple.Path;
+        var priority = tuple.Priority;
+
+        var count = 0;
+
+        for (var i = 1; i < path.Length; ++i)
```
hhyyrylainen, line 494:
```suggestion
        for (int i = 1; i < path.Length; ++i)
```
primitive types when used like this (var count is also the same above) should in my preference say the type. I see this a bunch all over this PR.

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -47,132 +30,101 @@ public static PatchMap Generate(WorldGenerationSettings settings, Species defaul
...
-
         // Potential starting patches, which must be set by the end of the generating process
         Patch? vents = null;
         Patch? tidepool = null;
 
         // Create the graph's random regions
         for (int i = 0; i < vertexCount; ++i)
         {
-            var patchName = nameGenerator.Next(random);
+            var regionName = nameGenerator.Next(random);
```
hhyyrylainen, line 44: Unless I*m wrong the namegenerator next has documentation that it returns the next patch name so this variable rename conflicts with what the name generator says it does.

src/microbe_stage/PatchMap.cs
84634E1A607A: I think I update this every time a new region is added.

src/microbe_stage/editor/PatchMapDrawer.cs
84634E1A607A: Emm, rider suggests this. I'll change it back then.

src/microbe_stage/PatchMapGenerator.cs
84634E1A607A: But what it is used for is actually region name? Or is this saying all patches in a region have the same name?

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -234,34 +184,34 @@ private static void LinkRegions(PatchRegion region1, PatchRegion region2)
...
+        var currentEdgeNumber = CurrentEdgeNumber(ref graph, vertexNumber);
 
         // Subtract edges until we reach the desired edge count.
         while (currentEdgeNumber > edgeNumber)
         {
-            int edgeToDelete = random.Next(1, currentEdgeNumber);
-            int i;
-            int k;
+            var edgeToDelete = random.Next(1, currentEdgeNumber + 1);
+            int i, k;
```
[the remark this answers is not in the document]
84634E1A607A: I think putting the two one-letter circulation variable in separate lines looks worse...

src/microbe_stage/PatchMapGenerator.cs
hhyyrylainen: Probably a misunderstanding. I think maybe the variable is misnamed elsewhere. But I'm certain that this code versus documentation comment problem needs to be resolved.

src/microbe_stage/PatchMapGenerator.cs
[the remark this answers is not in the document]
hhyyrylainen: At least in C++ it is a source of confusing to read bugs, that's why I'm disallowing it.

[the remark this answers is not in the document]
hhyyrylainen: Also our style guide says: "Don't declare multiple local variables on the same line, instead place each declaration on its own line"

So please follow the style guide.

src/microbe_stage/PatchMap.cs
```diff
@@ -445,7 +431,6 @@ private void RecreateAdjacencies()
         {
             var patch1 = Patches[id1];
             var patch2 = Patches[id2];
-
```
hhyyrylainen, line 448: Please leave this be, I prefer extra spacing, in fact I added either this or the one below this myself.

src/microbe_stage/PatchMap.cs
```diff
@@ -454,7 +439,6 @@ private void RecreateAdjacencies()
         {
             var region1 = Regions[id1];
             var region2 = Regions[id2];
-
```
hhyyrylainen, line 457: This is the other place.

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -623,27 +579,32 @@ private static void BuildPatchesInDrawingRegions(PatchMap map, Random random)
...
         }
     }
 
     /// <summary>
-    ///   Returns a predefined patch with default values
+    ///   Returns a predefined patch with default values.
     /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Note that after calling this, region and patch has already been linked together.
```
hhyyrylainen, line 593:
```suggestion
    ///     Note that after calling this, regions and patches have already been linked together.
```

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -623,27 +579,32 @@ private static void BuildPatchesInDrawingRegions(PatchMap map, Random random)
...
+    ///   </para>
+    /// </remarks>
     /// <param name="biome">The requested biome</param>
     /// <param name="id">ID of this patch</param>
     /// <param name="region">Region this patch belongs to</param>
-    /// <param name="regionName">Name of the region</param>
+    /// <param name="patchName">Name of the region</param>
```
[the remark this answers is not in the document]
hhyyrylainen: Actually looking at the results this *is* the region name as the patches in a region share the region name. I think instead of doing the regionName to patchName rename, PatchMapNameGenerator should be updated so that `Next`'s documentation says that it returns a region name (used as the base part of patch names in a region).

src/microbe_stage/PatchMap.cs
84634E1A607A: Actually, i've already changed that...

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -8,35 +8,18 @@
...
-        Vents = 10,
-    }
-
     public static PatchMap Generate(WorldGenerationSettings settings, Species defaultSpecies, Random? random = null)
     {
         random ??= new Random(settings.Seed);
 
         if (settings.MapType == WorldGenerationSettings.PatchMapType.Classic)
         {
             // Return the classic map if settings ask for it it
```
Kasterisk, line 17:
```suggestion
            // Return the classic map if settings ask for it
```

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -364,242 +311,228 @@ private static bool CheckIfRegionsIntersect(PatchRegion region1, PatchRegion reg
...
-                    var depth = region.Patches[i].Depth;
-                    var seafloor = region.Patches[i + 1];
-                    region.Patches[i].Depth[1] = random.Next(depth[0] + 1, depth[1] - 10);
-
-                    seafloor.Depth[0] = region.Patches[i].Depth[1];
-                    seafloor.Depth[1] = region.Patches[i].Depth[1] + 10;
+                    caveLinkedTo = random.Next(0, waterPatchCount);
+                    LinkPatches(cave, region.Patches[caveLinkedTo]);
+                    cave.Depth[0] = region.Patches[caveLinkedTo].Depth[0];
+                    cave.Depth[1] = region.Patches[caveLinkedTo].Depth[1];
```
Kasterisk, line 443: I'd add a separation between this 2 line with the above.

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -771,26 +748,28 @@ private static Patch NewPredefinedPatch(PredefinedBiome biome, int id, PatchRegi
         return newPatch;
     }
 
-    private static PatchMap PredefinedMap(PatchMap map, string areaName, Random random)
+    private static PatchMap GeneratePredefinedMap()
     {
+        var map = new PatchMap();
+        var areaName = TranslationServer.Translate("PANGONIAN_REGION_NAME");
```
Kasterisk, line 754:
```suggestion
        var regionName = TranslationServer.Translate("PANGONIAN_REGION_NAME");
```

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -332,117 +250,260 @@ private Vector2 GetRightCornerPointOnMap()
...
+    ///   <para>
+    ///     Note that priority should be within (-10, 10), and the return value is relative.
+    ///   </para>
+    /// </remarks>
+    private int IntersectionCountWithPriority((Vector2[] Path, int Priority) tuple)
+    {
+        var path = tuple.Path;
+        var priority = tuple.Priority;
+
+        int count = 0;
```
Kasterisk, line 474: `count` needs to be explained.

src/microbe_stage/Patch.cs
```diff
@@ -39,6 +40,29 @@ public Patch(LocalizedString name, int id, Biome biomeTemplate, PatchSnapshot cu
         this.currentSnapshot = currentSnapshot;
     }
 
+    /// <summary>
+    ///   This contains predefined biome types
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Explicitly stated values are used as patch ids in classic map
```
Kasterisk, line 48:
```suggestion
    ///     Explicitly stated values are used as patch IDs in classic map
```

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -332,117 +250,260 @@ private Vector2 GetRightCornerPointOnMap()
...
+
+                connections.Add(int2, pathToAdjacent);
             }
         }
 
-        foreach (var entry in map.SpeciallyDrawnRegions)
+        AdjustPathEndpoints();
+    }
+
+    private Vector2[] GetLeastIntersectingPath(PatchRegion start, PatchRegion end)
```
Kasterisk, line 280: Needs documentation for what seems to be a heavy logic containing method.

src/microbe_stage/Patch.cs
```diff
@@ -39,6 +40,29 @@ public Patch(LocalizedString name, int id, Biome biomeTemplate, PatchSnapshot cu
...
+    ///   </para>
+    /// </remarks>
+    public enum BiomeTypes
```
Kasterisk, line 51: This feels a bit unorderly to be in the Patch class, might need to be in its own file. Also this can probably be used not just as patch IDs in the classic map so the explicit values can be irrelevant. Generalizing this might be a good approach.

Kasterisk (dismissed):
Classic map defaults to the following initial view:



As you can see the top part is cut off, is this because of the hard coded region size?

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -332,117 +250,260 @@ private Vector2 GetRightCornerPointOnMap()
         return point;
     }
 
-    private Vector2 GetLeastIntersectionIntermediate(PatchRegion region1, PatchRegion region2)
+    /// <summary>
+    ///   This function create least intersected drawing links according to region adjacent
```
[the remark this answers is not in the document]
84634E1A607A: I think so.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -332,117 +250,260 @@ private Vector2 GetRightCornerPointOnMap()
...
+
+        int count = 0;
+
+        for (int i = 1; i < path.Length; ++i)
+        {
+            var startPoint = path[i - 1];
+            var endPoint = path[i];
+
+            // Calculate the number of intersecting regions for each possible line path
+            foreach (var reg in map.Regions)
```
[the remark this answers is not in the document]
84634E1A607A: I think I didn't code that. (Nor did I notice that)

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -469,24 +530,42 @@ private void DrawRegionLinks()
...
     {
         // Don't draw a border if there's only one region
         if (map.Regions.Count == 1)
             return;
 
-        foreach (var entry in map.Regions)
+        foreach (var region in map.Regions.Values)
+        {
+            DrawRect(new Rect2(region.ScreenCoordinates, region.Size),
+                new Color(0.0f, 0.7f, 0.5f, 0.7f), false, Constants.PATCH_REGION_BORDER_WIDTH);
```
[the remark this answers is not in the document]
84634E1A607A: Nor did I code this...

src/microbe_stage/editor/PatchMapDrawer.cs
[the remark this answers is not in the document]
Kasterisk: It's still warranted to be improved especially when spotted in code reviews like this.

src/microbe_stage/Patch.cs
84634E1A607A: > Also this can probably be used not just as patch IDs

This already isn't and has been used widely in PatchMapGenenrator.cs
The explicit values also serves a purpose, that in patch generation logic biome types are generated using random.Next() and by using an int circulation variable.

Kasterisk: Btw I just realized shoudn't this work without explicit values? Enums start from 0 already.

84634E1A607A: It should, but I doubt if one day someone will sort this enum alphabetically then break the whole generation.

Kasterisk: Documentations are there for a reason.

84634E1A607A: I'm wondering if I can move this to Biome class

Kasterisk: Biome class is a biome type by itself and this enum should be outside of it in terms of scope.

84634E1A607A:
I think that is because the move to selected function moved the classical map.

src/microbe_stage/PatchMapNameGenerator.cs
```diff
@@ -40,7 +40,7 @@ public string PatchName
     public string InternalName { get; set; } = null!;
 
     /// <summary>
-    ///   Generates and returns a new patch name
+    ///   Generates and returns a new region name
```
hhyyrylainen, line 43: I think the variables within this method need renaming now as well. There's at least PatchName on line 84

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -221,78 +168,110 @@ private static Vector2 ClosestPoint(Vector2 comparisonPoint, Vector2 point1, Vec
...
+    /// <remarks>
+    ///   <para>
+    ///     True if intersect at endpoint. And true if the two segments are collinear and has common points.
+    ///   </para>
+    ///   <para>
+    ///     Doesn't use `Geometry.SegmentIntersectsSegment2d()` because it isn't handling intersection at endpoint well.
+    ///   </para>
+    /// </remarks>
+    /// <returns>True if intersect</returns>
+    private static bool SegmentSegmentIntersect(Vector2 segment1Start, Vector2 segment1End,
```
hhyyrylainen, line 183:
```suggestion
    private static bool SegmentSegmentIntersects(Vector2 segment1Start, Vector2 segment1End,
```

84634E1A607A:
>  see this change to be GetPatchLocalizedName(string regionName, string biomeInternalName)

It's not. It's also used to generate continent name.

hhyyrylainen:
> It's not. It's also used to generate continent name.

Could you point out where that happens? How does it make sense there as the biome name key is a required parameter?

src/microbe_stage/PatchMapNameGenerator.cs
```diff
@@ -66,8 +51,7 @@ public string Next(Random random)
         // Convert first letter to uppercase
         name[0] = char.ToUpper(name[0], CultureInfo.InvariantCulture);
 
-        // Continent name is the name without the genitive
```
hhyyrylainen, line 69: Could you put this comment back, I think it's important to keep for future understandability?

84634E1A607A:
It isn't used anywhere other than debug errorprint for now.

hhyyrylainen:
I assumed that was the case because it would have otherwise probably come up in testing.

84634E1A607A:
If we want to use the continent name, we shouldn't add the "ocean / sea / continent" afterwards. So I changed it like this.

(Is it possible to use string instead of LocalizedString?)

hhyyrylainen:
> (Is it possible to use string instead of LocalizedString?)

Maybe it would actually make sense after all that the patch region name is a string type, the continent names come from the generator so they can't be translated anyway.

PR #3510, opened 2022-07-12
Use local population for microbe spawns

Oliveriver:

Use local rather than global population to determine microbe species spawn densities.

A bug that somehow slipped under everyone's noses for several years. This explains why the player species always seemed to have such a high spawn rate. This change will also mean the player species (other than the player cell and its daughter) won't spawn if the player moves to a patch they haven't yet colonised.

@buckly90 This may need a balancing pass. I picked 25 as the divisor, but there's not a lot of reasoning behind it. It needs to be less than 50 (otherwise the player species won't spawn for the next generation as well) and I went a fair bit lower to cancel out the reduction in spawn rate caused by using local rather than global populations.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/PatchManager.cs
+++ b/src/microbe_stage/PatchManager.cs
@@ -181,6 +181,7 @@ public class PatchManager : IChildPropertiesLoadCallback
         foreach (var entry in patch.SpeciesInPatch.OrderByDescending(entry => entry.Value))
         {
             var species = entry.Key;
+            var population = entry.Value;
 
             if (species.Population <= 0)
             {
@@ -194,7 +195,7 @@ public class PatchManager : IChildPropertiesLoadCallback
                 continue;
             }
 
-            var density = Mathf.Max(Mathf.Log(species.Population / 50.0f) * 0.01f, 0.0f);
+            var density = Mathf.Max(Mathf.Log(population / 25.0f) * 0.01f, 0.0f);
 
             var name = species.ID.ToString(CultureInfo.InvariantCulture);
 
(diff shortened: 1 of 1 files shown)


src/microbe_stage/PatchManager.cs
```diff
@@ -181,6 +181,7 @@ private void HandleCellSpawns(Patch patch)
         foreach (var entry in patch.SpeciesInPatch.OrderByDescending(entry => entry.Value))
         {
             var species = entry.Key;
+            var population = entry.Value;
 
             if (species.Population <= 0)
```
hhyyrylainen, line 186: Shouldn't this be updated as well to use `population`?

RacerBG (approved):
Looks good.

1. When I move to a different patch, indeed my cell and my daughter cell are the only inhabitants of the new patch.
2. My initial population is always 50 cells, after that it easily grows beyond 2000 for 1 generation.

Oliveriver:
> Looks good.
> 
> 1. When I move to a different patch, indeed my cell and my daughter cell are the only inhabitants of the new patch.
> 2. My initial population is always 50 cells, after that it easily grows beyond 2000 for 1 generation.

What do you think of the spawn rate of cells in general?

RacerBG:
> > Looks good.
> > 
> > 1. When I move to a different patch, indeed my cell and my daughter cell are the only inhabitants of the new patch.
> > 2. My initial population is always 50 cells, after that it easily grows beyond 2000 for 1 generation.
> 
> What do you think of the spawn rate of cells in general?

It should be lower, it's getting high too fast. I'm not talking about specifically about this PR though this is my general opinion for the spawn rate.

buckly90:
I think it works perfectly fine.

Player species is no longer absolutely everywhere, and I had no trouble finding colony candidates either. Splendid job!

Kasterisk (approved):

PR #3511, opened 2022-07-12
Fixed loading saves with microbe colonies in them

hhyyrylainen:

hopefully fixes the last saving bug in 0.5.9 RC1

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/ColonyCompoundBag.cs
+++ b/src/microbe_stage/ColonyCompoundBag.cs
@@ -12,7 +12,7 @@ public class ColonyCompoundBag : ICompoundStorage
     }
 
     [JsonProperty]
-    private MicrobeColony Colony { get; set; }
+    private MicrobeColony Colony { get; }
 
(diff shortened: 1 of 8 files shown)


src/microbe_stage/ColonyCompoundBag.cs
```diff
@@ -12,7 +12,7 @@ public ColonyCompoundBag(MicrobeColony colony)
     }
 
     [JsonProperty]
-    private MicrobeColony Colony { get; set; }
+    private MicrobeColony Colony { get; }
```
Kasterisk, line 15: I thought saving won't work with properties with no setters?

hhyyrylainen: It actually works (and now that I've improved the system, I think we should move towards this style) if there is a constructor (either marked with the json constructor property or the first one) that takes a parameter of that name.

PR #3520, opened 2022-07-13
Balancepass 0.5.9

buckly90:
* Decreased overall production of photosynthesis and thermosynthesis
* Fixed rustycyanin processes being doubled for sake of consistency. (Does not change gameplay.)
* Slightly lowered pilus damage
* Occurrence of ammonia and phosphate has been increased
* Caves no longer have abnormally small amounts of iron

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/MainMenu.tscn
+++ b/src/general/MainMenu.tscn
@@ -417,7 +417,6 @@ text = "BACK"
 
 [node name="OptionsMenu" parent="." instance=ExtResource( 3 )]
 visible = false
-DisplayPartNamesTogglePath = null
 
 [node name="SaveManagerGUI" parent="." instance=ExtResource( 2 )]
 visible = false
--- a/src/general/MainMenu.tscn
+++ b/src/general/MainMenu.tscn
@@ -417,6 +417,7 @@ text = "BACK"
 
 [node name="OptionsMenu" parent="." instance=ExtResource( 3 )]
 visible = false
+DisplayPartNamesTogglePath = null
 
 [node name="SaveManagerGUI" parent="." instance=ExtResource( 2 )]
 visible = false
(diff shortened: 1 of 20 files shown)


src/general/MainMenu.tscn
```diff
@@ -417,6 +417,7 @@ text = "BACK"
 
 [node name="OptionsMenu" parent="." instance=ExtResource( 3 )]
 visible = false
+DisplayPartNamesTogglePath = null
```
hhyyrylainen, line 420: This seems a bit suspicious change...

buckly90: Uh oh, I must have accidently commited godot's changes

Oliveriver (approved):
Played for a while with these changes. While on the whole I agree with them - I think upping amounts of phosphate and ammonia was a particularly good change - I have discovered one problem.

It's surprisingly hard to be an effective predator now. The engulfment change means there are fewer options for prey as so many of them evolve protective membranes, and the lower pilus damage means stabbing takes so long you end up losing compounds in the overall hunt. So I'm in favour of increasing pilus damage, probably past what it was.

Otherwise, all seems good.

buckly90:
You might be right about that. I lowered the pilus damage a bit to help lessen the impact of what I like to call "critical strikes" when a pilus makes contact with a cell in just the right way to cause damage numerous times within a short time frame. It can sometimes kill a cell nigh instantly.

buckly90:
A good point, I'll change it back for now, and won't a change it until we can introduce a damage delay or something

hhyyrylainen (approved):
If the gameplay changes are good I don't have anything left to comment on the technical side of things.

PR #3522, opened 2022-07-14
Add GUI section to the graphics setting

Kasterisk:

For better organization.

<image width=60% src="https://user-images.githubusercontent.com/54026083/178906653-44b2b803-3b99-4ede-8fc7-693858eed946.png">

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/locale/af.po
+++ b/locale/af.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2022-07-14 12:36+0700\n"
+"POT-Creation-Date: 2022-07-14 14:21+0700\n"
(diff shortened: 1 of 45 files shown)


hhyyrylainen:
Seems like a sensible change. I think we'll eventually want a separate "UI" tab for all kinds of tweaks like that, we could maybe even make the damage numbers mod into an official feature that could be toggled there. I suspect we can come up with tons of small UI tweaks so that it'll eventually warrant its own tab.

hhyyrylainen (changes requested):
I think the only change I'd like to see is that the background particles is in the non-GUI section as it is not really related to the GUI at all, it is after all an effect part of the 3D scene like the chromatic aberration is.

Kasterisk:
Didn't saw that one.

hhyyrylainen:
Looks like the CI build is about to fail due to outdated translation string order.

hhyyrylainen (approved):
Seems fine, though I didn't go through and click every one of the options, I'll just assume that the checkbox callbacks still work fine.

PR #3532, opened 2022-07-15
Fix engulfment of colony members

### LINKED ISSUE #3528: New engulfing mechanics need to handle detaching colony members properly
from @Kasterisk's explanation it seems that the engulfing won't properly let the colony know that a member was removed from it, which probably causes various bugs. We should fix this by just letting the colony know that a member got "killed" when something is engulfed from it.

Kasterisk:

Colony members are now properly considered detached/removed if they're engulfed.

**Related Issues**

Fixes #3528

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Microbe.cs
+++ b/src/microbe_stage/Microbe.cs
@@ -1023,6 +1023,9 @@ public partial class Microbe : RigidBody, ISpawned, IProcessable, IMicrobeAI, IS
         if (Colony == null)
             return GetParent();
 
+        if (Colony.Master.HostileEngulfer.Value != null)
+            return Colony.Master.HostileEngulfer.Value.GetStageAsParent();
+
         return Colony.Master.GetParent();
     }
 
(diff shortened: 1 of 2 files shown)


src/microbe_stage/Microbe.cs
```diff
@@ -1023,6 +1023,9 @@ private Node GetStageAsParent()
         if (Colony == null)
             return GetParent();
 
+        if (Colony.Master.HostileEngulfer.Value != null)
```
hhyyrylainen, line 1026: When is this not null, doesn't the fact that when the engulfing starts the cell is removed from the colony mean that if the master is engulfed, it will just immediately break the colony? Or is this needed in this case. Could you clarify this with a comment in the source code?

hhyyrylainen:
Code change looks good now. I haven't had time to playtest this, hoping someone else might have time...

PR #3534, opened 2022-07-15
Fix multicellular editing after loading a save, and fixed already placed cells not updating

hhyyrylainen:

there was a problem where the cell template instances where different in different places (because they weren't saved as references) that caused some editing problems. Fixed already placed cell visuals no longer reacting to cell type edits.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/early_multicellular_stage/CellTemplate.cs
+++ b/src/early_multicellular_stage/CellTemplate.cs
@@ -2,6 +2,7 @@
 using Godot;
 using Newtonsoft.Json;
 
+[JsonObject(IsReference = true)]
 public class CellTemplate : IPositionedCell, ICloneable, IActionHex
 {
     private int orientation;
--- a/src/early_multicellular_stage/CellType.cs
+++ b/src/early_multicellular_stage/CellType.cs
@@ -6,6 +6,7 @@ using Newtonsoft.Json;
 /// <summary>
 ///   Type of a cell in a multicellular species. There can be multiple instances of a cell type placed at once
 /// </summary>
+[JsonObject(IsReference = true)]
 public class CellType : ICellProperties, IPhotographable, ICloneable
 {
     [JsonConstructor]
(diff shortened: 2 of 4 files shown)


Kasterisk:
Fixes what it's trying to fix. Though I found a bug not sure if it's related to this, if you edit a cell type and then return to the body plan for the first time, it wouldn't update the placed cells' visuals. Subsequent tab switches fixes the problem.

hhyyrylainen:
> Though I found a bug not sure if it's related to this, if you edit a cell type and then return to the body plan for the first time,

Do you mean if you don't confirm the changes? That's kind of working like it logically should, I mean the change is not done before you hit confirm... But I can see why it would be a bit confusing as the change will get confirmed when exiting the editor. I'm not really sure at this point which way would be more intuitive for players.

Kasterisk:
I didn't press next, what I did was changing the color and then switching to the `Editor` tab directly. What's weird was that the change is only visible to the placed cells the next time you do that so the change must have been confirmed regardless whether you press next or not. Even if it's not (changes are only confirmed when your press next), personally I think it's more intuitive if changes are confirmed immediately.

hhyyrylainen:
So you say that pressing *next* should be the same as confirming the changes?

I have a guess as to what causes the bit of weirdness (also in late multicell editor):
```
        // Only do something if the user has done any action in the past
        if (!history.CanUndo())
            return;
```

hhyyrylainen:
I opened an issue for that: https://github.com/Revolutionary-Games/Thrive/issues/3535 because I started thinking along the lines that that is a kind of unintended bug in the code.

Kasterisk:
> So you say that pressing next should be the same as confirming the changes?

By confirming, I mean committing the changes and be actualized in the Editor/body plan tab, not the confirm-and-exit multicell editor kind.

Kasterisk (approved):

PR #3546, opened 2022-07-17
Fix patch extinction animation not showing

### LINKED ISSUE #3543: Patch map in patch extinction screen doesn't animate correctly
Making this issue to make sure this isn't forgotten.

The patch map in the patch extinction box doesn't call _Process on the patch map nodes, for reasons I couldn't fathom after a while spent trying to fix it. This means the current patch border doesn't blink and hovered patch borders don't show.

84634E1A607A:
Fixes #3543 

Turns out map is rebuilt every time so there isn't a chance for them to animate.

Personally I dislike the approach to attempt to update map every frame, but adding early return causes a crash. If you think this approach is acceptable, go on and merge this.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/PatchMapDrawer.cs
+++ b/src/microbe_stage/editor/PatchMapDrawer.cs
@@ -49,7 +49,13 @@ public class PatchMapDrawer : Control
         get => map;
         set
         {
-            map = value ?? throw new ArgumentNullException(nameof(value), "setting to null not allowed");
+            if (value == null)
+                throw new ArgumentNullException(nameof(value), "setting to null not allowed");
+
+            if (map == value)
+                return;
+
+            map = value;
             dirty = true;
 
             playerPatch ??= map.CurrentPatch;
@@ -91,6 +97,8 @@ public class PatchMapDrawer : Control
 
     public override void _Ready()
     {
+        base._Ready();
+
         nodeScene = GD.Load<PackedScene>("res://src/microbe_stage/editor/PatchMapNode.tscn");
 
         if (DrawDefaultMapIfEmpty && Map == null)
@@ -102,6 +110,8 @@ public class PatchMapDrawer : Control
 
     public override void _Process(float delta)
     {
+        base._Process(delta);
+
         CheckForDirtyNodes();
 
         if (dirty)
@@ -120,6 +130,8 @@ public class PatchMapDrawer : Control
     /// </summary>
     public override void _Draw()
     {
+        base._Draw();
+
         if (Map == null)
             return;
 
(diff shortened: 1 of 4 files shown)


src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -49,7 +49,13 @@ public class PatchMapDrawer : Control
         get => map;
         set
         {
-            map = value ?? throw new ArgumentNullException(nameof(value), "setting to null not allowed");
+            if (value == null)
+                throw new ArgumentNullException(nameof(value), "setting to null not allowed");
+
+            if (map == value)
```
hhyyrylainen, line 55: This seems like good idea, though, would you or someone else be able to check why the patch extinction drawer repeatedly sets the map property? That would be a good place to also fix.

84634E1A607A:
```
PatchMapDrawer.set_Map() at D:\source\repos\Thrive\src\microbe_stage\editor\PatchMapDrawer.cs:line 53
PatchExtinctionBox.set_Map() at D:\source\repos\Thrive\src\microbe_stage\gui\PatchExtinctionBox.cs:line 27
StageHUDBase<MicrobeStage>.ShowPatchExtinctionBox()
StageBase<Microbe>.PlayerExtinctInPatch()
MicrobeStage.PlayerExtinctInPatch()
StageBase<Microbe>.HandlePlayerRespawn()
StageBase<Microbe>._Process()
MicrobeStage._Process()
```

hhyyrylainen: So the ShowPatchExtinctionBox should be modified to be called just once only if the extinction screen is not open?

84634E1A607A: I think so, but as I stated, it causes a crash.

hhyyrylainen: Could someone help out here? If it takes more than a few days, I can take a look at this problem next week.

84634E1A607A: I got it working.

hhyyrylainen: Seems like a good approach. I'll try to ask for testing on this, especially if multiple patch extinctions happen in a row to make sure there's no bugs.

Oliveriver:
Nicely done! I can confirm this works.

RacerBG (approved):
Indeed, it now blinks just like in the normal patch map.

hhyyrylainen:
I'll assume people tested the multiple patch extinctions in a row case.

hhyyrylainen (approved):
Code changes look fine.

PR #3570, opened 2022-07-20
Global exception logger

84634E1A607A:
**Progress Checklist**

      break existing features:
      members reviewing this PR)


84634E1A607A:
Crash report description needs polishing. (Feel free to change it)

hhyyrylainen:
I'm 90% certain this will not change anything as only native code crashes (which is what Breakpad detects) are reportable by the launcher. There's an issue for the launcher to make it detect exception stack traces and offer to report them from logs: https://github.com/Revolutionary-Games/Thrive-Launcher/issues/163
Looks like I maybe had forgotten to open an issue for that in the launcher, may have just been on my personal TODO list, so I had to open one now.

src/engine/StartupActions.cs
```diff
@@ -15,6 +16,13 @@ private StartupActions()
         // TODO: for devbuilds it would be nice to print the hash here
         GD.Print("This is Thrive version: ", Constants.Version);
 
+        // Add crash logger if debugger not attached
+        if (!Debugger.IsAttached)
+        {
+            GD.Print("Crash logger attached");
```
hhyyrylainen, line 22: This is not really a crash logger, I'd name this unhandled exception logger, because as I said in my comment, I think none of the reports that players have reported can have been like this. So I'd like to reword this to not confuse players reporting these with the native crashes. So I'd rename all instances of "crash" in this PR to be "Unhandled Exception"

src/engine/CrashLogger.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Godot;
+
+/// <summary>
+///   Crash logger for any unhandled Mono exceptions
+/// </summary>
+public class CrashLogger
+{
+    public static void UnhandledException(object sender, UnhandledExceptionArgs args)
+    {
+        GD.PrintErr("------------ Begin of Crash Log ------------\n" +
```
hhyyrylainen, line 10: Don't concatenate multiple strings with +, it's very inefficient in C#. Instead I think this could use just commas to separate the parts, if you don't want to use a string builder here.

hhyyrylainen (approved):
I did a few slight tweaks myself and tested this. Works as intended I think. In the future we could maybe do something like exiting to a separate scene allowing viewing and copying the exception log to make it easier to report these.

84634E1A607A:
> In the future we could maybe do something like exiting to a separate scene

I don't think it is possible, because an unhandled exception leaves the logger little time, as long as the logger function returns, game crashes.

hhyyrylainen:
> I don't think it is possible, because an unhandled exception leaves the logger little time, as long as the logger function returns, game crashes.

Oh yeah, I suppose that is true... Perhaps there's a way to tell Godot to not actually close due to the exception? Or maybe we could write the exception to a specifically named file, which the launcher could then detect and offer to submit it as a crash report?

PR #3571, opened 2022-07-20
Multiple patchmap fixes

### LINKED ISSUE #3555: Regions should not have duplicate biomes to avoid duplicate patch names
as there's no way for the player to differentiate them, also a bit weird from the perspective that patches are uniform condition areas that a species can inhabit as a whole, so having multiple patches of the same type in the same region isn't very sensical as the next to each other patches should be considered just a single larger patch.

### LINKED ISSUE #3557: Tidepool patches shouldn't directly connect to ocean patches
instead they should always require a coast patch in between

### LINKED ISSUE #3592: Loading a save with the classic patch map breaks patch connections
Start a game with the classic map, save (in the editor) and load immediately, in the editor you'll see the following problem:

84634E1A607A:

This PR does some stuff...
 - Fixes #3555
 - Fixes #3557 
 - Fixes cave connected to other region
 - Fixes #3592 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (approved):
I can confirm the classic map loading is fixed and that in the few freebuild starts I did, I didn't see any of the problems this should fix, so with that minimal testing I'll merge this as it seems no one else wants to comment on this.

PR #3573, opened 2022-07-20
Increased auto-evo performance by 60% with more caching

hhyyrylainen:
I did some profiling and found some places (mostly expensive computations that weren't cached) in auto-evo that I was able to improve the performance of with extra caching.

Overall result is about a 60% performance improvement (in my test from 40 seconds to a bit under 16).

This should bring the performance to be about on par the classical map for the procedural maps with a lot more patches.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Oliveriver:
Might have found a pretty serious bug with this. Every generation for 20 generations, there were no more than two species in any patch. I loaded the same save on master and immediately the patches filled with many more species. That alone could be the cause of the performance improvement, so it definitely needs to be fixed before more testing is done.

Just did the same thing again and got exactly the same result, except for a single patch that had four species.

hhyyrylainen:
Okay yeah, I see the same thing, looks like migrations are not working (I'll also check that mutations work as those are almost the same approach).

Oliveriver:
Yep, things seem to behave now. And performance is still good, I get about a 60% improvement too.

Might still be worth some more thorough checking that nothing about Auto-Evo's functionality has changed.

PR #3585, opened 2022-07-21
Reduced lag when opening art gallery as the image cards now load in o…

### LINKED ISSUE #3464: Art gallery shouldn't try to load all images at once
otherwise there is a lag spike when opening it. Doing something like loading a new image every 0.1 seconds passed would be probably the best, and showing the loading icon on all the other art pieces.

AaronMcGarry:
…ne every 0.1 seconds

All the cards in the first tab have a timer that determines when they should start loading, which eliminates most of the lag.

**Related Issues**
closes #3464 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


AaronMcGarry:
Not sure why its failing this test, because I didn't have any style warnings when I committed. Also, the log says "Code cleanup performed changes, please stage / check them before committing", but I'm not sure where to find the changes.

hhyyrylainen:
The jetbrains check shows the diff of what changes it wants you to perform:
```diff
--- a/src/gui_common/art_gallery/GalleryCard.cs
+++ b/src/gui_common/art_gallery/GalleryCard.cs
@@ -16,7 +16,7 @@ public class GalleryCard : Button
     private Label? titleLabel;
     private TextureRect? imagePreview;
     private Texture? thumbnail;
-    private bool isLoaded = false;
+    private bool isLoaded;
     private float time;
 
     [Signal]
```

src/gui_common/thrive_theme.tres
```diff
@@ -1,9 +1,8 @@
-[gd_resource type="Theme" load_steps=73 format=2]
+[gd_resource type="Theme" load_steps=72 format=2]
 
 [ext_resource path="res://src/gui_common/fonts/Jura-Regular-AlmostSmaller.tres" type="DynamicFont" id=1]
 [ext_resource path="res://assets/textures/gui/bevel/grab.png" type="Texture" id=2]
 [ext_resource path="res://assets/textures/gui/bevel/grabhover.png" type="Texture" id=3]
-[ext_resource path="res://assets/textures/gui/bevel/hSeparatorCentered.png" type="Texture" id=4]
```
hhyyrylainen, line 6: Probably unintended breaking changes in this file?

src/gui_common/art_gallery/GalleryCard.cs
```diff
@@ -44,6 +52,18 @@ public override void _Ready()
         UpdatePreview();
     }
 
+    public override void _Process(float delta)
+    {
+        base._Process(delta);
+
+        time += delta;
+        if (IsImageOnlyCard && !isLoaded && time >= CardIndex * LOAD_TIME_PER_CARD)
```
hhyyrylainen, line 60: How does this prevent multiple cards from loading at the same time?
I was thinking that a new ResourceManager class would make sense to create, it would work similarly to how the photostudio works in that you submit resource loading tasks to it, and then it loads one thing per frame (or even waits between frames to not tank FPS below 30 or so).

PR #3588, opened 2022-07-23
Improved rigidity slider handling again

### LINKED ISSUE #3203: Rigidity slider can become disabled while there is still MP (multicellular?)
report came in that in multicellular the rigidity slider got disabled while there were 8 MP remaining. This might be multicellular specific unless my refactoring caused the microbe editor to also break like this.

Probably can be fixed by going through the actions in the editor and make sure it probably updates the rigidity state. Also when switching to the cell editor tab needs to make sure that the logic there will set things correctly for that tab.

84634E1A607A:
**Related Issues**
Suppresses #1533
Closes #3452 
Fixes #3203

**Progress Checklist**

      break existing features:
      members reviewing this PR)
- [x] Multicellular functionality needs testing

### DIFF
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -278,6 +278,8 @@ public partial class CellEditorComponent :
 
     private bool microbePreviewMode;
 
+    private int originalRigidity;
+
     public enum SelectionMenuTab
     {
         Structure,
@@ -487,8 +489,6 @@ public partial class CellEditorComponent :
         // Send info to the GUI about the organelle effectiveness in the current patch
         CalculateOrganelleEffectivenessInPatch(Editor.CurrentPatch);
 
-        UpdateRigiditySliderState(Editor.MutationPoints);
-
         UpdateCancelButtonVisibility();
 
         if (IsMulticellularEditor)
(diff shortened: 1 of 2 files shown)


84634E1A607A:
I think this's got bugs in multicellular. Wait until I figure this out.

84634E1A607A:
It works well actually...

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -598,6 +598,7 @@ public override void OnEditorSpeciesSetup(Species species)
         // that just marks things dirty and we update our stats on the next _Process call)
         Membrane = Editor.EditedCellProperties!.MembraneType;
         Rigidity = Editor.EditedCellProperties.MembraneRigidity;
+        originalRigidity = (int)Math.Round(Rigidity * Constants.MEMBRANE_RIGIDITY_SLIDER_TO_VALUE_RATIO);
```
hhyyrylainen, line 601: Storing the original value like this goes against the dynamic MP, so having just the rigidity not follow the way everything else is handled in the editor is not the right design in my opinion.

84634E1A607A:
I tried but was defeated by the dynamic MP. I don't even know if a bug is because dynamic MP or because I wrote it. Closing this.

hhyyrylainen:
I think I'll open sometime soon a discussion topic on the dev forum about the dynamic MP and if there is enough willingness to do a system overhaul on it.

Issue #3270, opened 2022-04-26
Reduce locks in compound absorbing in multithreaded conditions if there aren't major glitches


hhyyrylainen:
Compound absorbing of microbes is one of the most computationally intensive parts of the game. So moving it to happen in a multithreaded way would probably give us a bunch of extra performance. One caveat is though that probably using locks in separate cloud layers would reduce performance we need to just choose to not care about some slightly lost reads / extra compounds being absorbed from clouds. That shouldn't hopefully be able to cause more serious data corruption so we can probably deal with that.

hhyyrylainen:
I added basic and safe multithreading but due to the locks it doesn't help that much: https://github.com/Revolutionary-Games/Thrive/pull/3288 so this issue is now about trying to remove those locks and making sure things don't explode too badly in that situation.

hhyyrylainen:
With more profiling I've actually now also realized that the absorbing is so expensive because it triggers the membrane rebuild if it is dirty. That cannot entirely happen asynchronously, see: https://github.com/Revolutionary-Games/Thrive/issues/3770

PR #3590, opened 2022-07-23
Add mucilage (slime)

Oliveriver:

Adds mucilage to the game, a new agent that a cell can secrete to jet away from predators and slow them down.

TODO

* Replace icons (mucilage compound and slime jet organelle)
* Fix model self-intersect and not jiggling

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Oliveriver:
I've converted this into ready to review since the existing parts are about there. There are still a few holes though, listed above, so help with those from knowledgeable people would be appreciated.

yowshi:
how hard would it be to have the stuff disappear entirely over time representing its breakdown? because otherwise the map is just going to be full of this stuff.

nickthenick:
Just playested with it for 10 minutes. My thoughts:

- Functionally speaking, great work! It functions exactly as I would've imagined. 
- I felt like mucilage ran out too quickly. My first instinct was that it should expel at half the current rate, so that you can spread out its usage. For example, as a Eukaryote, I think I could only "jet" for 0.5-1 seconds before it would run out, even when fully filled up.
- Not an issue with this feature, but because of the game's current performance issues, the fast bursts of speed look very jittery (they highlight the low framerate). 
- I really like that the trails of mucilage can be used to track cells. Adds an extra element of strategy to which way to swim. 
- If there is anybody on the graphics team able to implement the "clear gel emitter" that Nunz described, I think that would be much preferrable over green clouds. I feel like there are already enough clouds and colours on the screen already and we shouldn't be adding another.

yowshi:
> Just playested with it for 10 minutes. My thoughts:
> 
> * Functionally speaking, great work! It functions exactly as I would've imagined.
> * I felt like mucilage ran out too quickly. My first instinct was that it should expel at half the current rate, so that you can spread out its usage. For example, as a Eukaryote, I think I could only "jet" for 0.5-1 seconds before it would run out, even when fully filled up.
> * Not an issue with this feature, but because of the game's current performance issues, the fast bursts of speed look very jittery (they highlight the low framerate).
> * I really like that the trails of mucilage can be used to track cells. Adds an extra element of strategy to which way to swim.
> * If there is anybody on the graphics team able to implement the "clear gel emitter" that Nunz described, I think that would be much preferrable over green clouds. I feel like there are already enough clouds and colours on the screen already and we shouldn't be adding another.

honestly i only noticed a speed boost when i was a small prokaryote. i didn't notice one when i hit eukaryote.

Deus-Codes:
I played only a short bit currently but it's very, very fun, and it livens up the whole experience. Gameplay wise, we benefit a lot from essentially having the Thrive equivalent of those cellular jet things in Spore. It's a nice change of pace.

One thing I noticed is a bit of bugginess where my cell occasionally maintains its when ejecting mucilage constantly even if I am not holding down the button. In fact, this would last until I next tapped the R button, allowing me at one point to zip across the map for about 10 seconds straight without expending slime. This might indicate more bugginess overall so I won't make any substantial comments on balancing for speed, but propulsion felt decent upon first impression. I'd like to see if anyone else noticed these issues and then move on to balancing for gameplay.

Oliveriver:
> how hard would it be to have the stuff disappear entirely over time representing its breakdown? because otherwise the map is just going to be full of this stuff.

I was hoping to leave the question of what to do with environmental mucilage until a later PR, but it seems it’s too big a problem to ignore. I have a few ideas:

* Cells which produce mucilage collect it back from the environment when swimming through it.
* Mucilage clouds decay into glucose over time (though obviously a smaller amount than it took to make).

Other suggestions are welcome. I’d advise against cells collecting mucilage and converting it to glucose internally, since that’s going to introduce many headaches with the process system.

Oliveriver:
> One thing I noticed is a bit of bugginess where my cell occasionally maintains its when ejecting mucilage constantly even if I am not holding down the button. In fact, this would last until I next tapped the R button, allowing me at one point to zip across the map for about 10 seconds straight without expending slime. 

I haven’t experienced this bug myself. How reliably can you reproduce it?

yowshi:
> > how hard would it be to have the stuff disappear entirely over time representing its breakdown? because otherwise the map is just going to be full of this stuff.
> 
> I was hoping to leave the question of what to do with environmental mucilage until a later PR, but it seems it’s too big a problem to ignore. I have a few ideas:
> 
> * Cells which produce mucilage collect it back from the environment when swimming through it.
> * Mucilage clouds disappear over time.
> * Mucilage clouds decay into glucose over time (though obviously a smaller amount than it took to make).
> 
> Other suggestions are welcome. I’d advise against cells collecting mucilage and converting it to glucose internally, since that’s going to introduce many headaches with the process system.

that third one is kind of cool, has interesting possibilities for play with the new day/night cycle too.

Oliveriver:
I've now added a decay rate for mucilage clouds. Seems about balanced to me. I'm therefore in favour of keeping it like this and leaving other options until later.

Kasterisk:
I don't know if this is just me but I think the latest change made the jetting action feel less "explosive", instead it's like more gradual increase in acceleration? Unless this is intended, I feel this might make the jet less effective considering you could only do it in small quick bursts most of the time.

src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -55,6 +59,8 @@ public class CompoundCloudPlane : CSGMesh, ISaveLoadedTracked
     // Called when the node enters the scene tree for the first time.
     public override void _Ready()
     {
+        mucilage = SimulationParameters.Instance.GetCompound("mucilage");
```
Kasterisk, line 62: I think CompoundCloudPlane shouldn't know about a specific compound. Perhaps I think a better a way could be to add a new `absorbable` property (if there isn't already) to the compounds which is set false for the mucilage. This property will then be checked higher up in the microbe's own compound absorbing handling.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -327,6 +337,45 @@ public void QueueEmitToxin(Compound toxinCompound)
         queuedToxinToEmit = toxinCompound;
     }
 
+    /// <summary>
+    ///   Tries to secrete slime if possible
+    /// </summary>
+    public void SecreteSlime(float delta)
```
Kasterisk, line 343: Is it possible handling this in SlimeJetComponent instead? Just like flagellum component and other ExternallyPositionedComponents, I think we should design organelle functionalities in a way that as much as the implementation should be put in their own respective components.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -106,6 +113,9 @@ public partial class Microbe
     [JsonProperty]
     public int AgentVacuoleCount { get; private set; }
 
+    [JsonProperty]
+    public List<SlimeJetComponent> SlimeJets { get; private set; } = new();
```
Kasterisk, line 117: Then the SlimeJet components don't have to be stored in the Microbe class. IMHO it would probably quickly turn into a mess otherwise if we have to do this for many other possible components.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -471,11 +520,11 @@ public Microbe Divide()
     /// <summary>
     ///   Throws some compound out of this Microbe, up to maxAmount
     /// </summary>
-    public float EjectCompound(Compound compound, float maxAmount)
+    public float EjectCompound(Compound compound, float maxAmount, Vector3 direction, float displacement = 0)
```
Kasterisk, line 523: These new parameters should be documented for understandability.

Kasterisk (dismissed):

src/microbe_stage/Microbe.Interior.cs
hhyyrylainen: If nothing else the slime secretion logic could be put in the organelle component to spare the poor Microbe class from having even more lines of code...

src/microbe_stage/Microbe.Interior.cs
hhyyrylainen: I haven't looked at the overall design but I'll cautiously say I'm in favour of this. Perhaps a design like the flagellum could be used if multiple slime jets help with overall propulsion to avoid having each slime jet call separately into the clouds system.

Oliveriver:
> I don't know if this is just me but I think the latest change made the jetting action feel less "explosive", instead it's like more gradual increase in acceleration? Unless this is intended, I feel this might make the jet less effective considering you could only do it in small quick bursts most of the time.

If you're referring to the bug fix I made most recently, that's definitely needed. The explosiveness you're referring to might well be that bug in action. It's possible to work around this by adjusting the constant though.

src/microbe_stage/Microbe.Interior.cs
Oliveriver: I'm not sure how I can achieve that considering this method needs to know about all the slime jets in a cell, not just one.

hhyyrylainen: Then with this and the comment on `public List<SlimeJetComponent> SlimeJets { get; private set; } = new();` the approach used by toxin shooting is probably the right one.

buckly90:
Very fun feature, and very well executed. I was weary of the mucilage existing as it's own compound but I've warmed up to it! The mucilage slows targets a good amount, not too much or too little. The boosting effect is very fun with just enough boost to escape immediate danger but not too much as to become a bottle rocket. I enjoyed playing as a little single-celled squid creature.

I did not notice any bugs or errors, so I say this is good to go on the gameplay side of things, fantastic work Oliver!

hhyyrylainen:
Is this now starting to be feature complete so that programming side reviews should happen?

Oliveriver:
> Is this now starting to be feature complete so that programming side reviews should happen?

Yes, apart from the TODOs mentioned in the PR description, but I think those are entirely separate code-wise.

src/general/StageHUDBase.cs
```diff
@@ -1017,9 +1033,11 @@ protected void UpdatePanelSizing(float delta)
 
         var environmentPanelVBoxContainer = environmentPanel.GetNode<VBoxContainer>("VBoxContainer");
         var compoundsPanelVBoxContainer = compoundsPanel.GetNode<VBoxContainer>("VBoxContainer");
+        var agentsPanelVBoxContainer = agentsPanel.GetNode<VBoxContainer>("VBoxContainer");
```
Maxonovien, line 1036: I understand the consistency with previously existing code, but this should probably use exported `NodePath`s, as discussed on discord.

src/microbe_stage/Compound.cs
```diff
@@ -20,6 +20,16 @@ public class Compound : IRegistryType
 
     public bool IsCloud;
 
+    /// <summary>
+    ///   Scales the retention rate of this compound as a cloud in the environment (DO NOT set above 1!)
+    /// </summary>
+    public float DecayRate = 1.0f;
```
Maxonovien, line 26: Why isn't this in the JSON parameters files, like `IsAgent` property? 

Also, for the maximum value, perhaps it would then be adequate to use `Math.Clamp(0.0f, 1.0f)` (syntax not guaranteed) ? Or, alternatively, use exceptions if this should really not go unnoticed.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -327,6 +337,45 @@ public void QueueEmitToxin(Compound toxinCompound)
...
+
+        // Reset the jet factor before calculating to prevent accidental build up of speed
+        slimeJetFactor = Vector3.Zero;
+
+        // Secrete slime and gather overall force from directions determined by equipped jets
+        foreach (var jet in SlimeJets)
+        {
+            var slimeSecreted = EjectCompound(mucilage, Constants.COMPOUNDS_TO_VENT_PER_SECOND * delta,
+                -jet.GetDirection(), 2);
+            slimeJetFactor += slimeSecreted * jet.GetDirection();
```
Maxonovien, line 361: As `jet.GetDirection()` runs a few lines of computations, it may be relevant to store it somewhere.

Maxonovien, line 361: As `jet.GetDirection()` runs a few lines of computations, it may be relevant to store it somewhere.

Maxonovien, line 361: As `jet.GetDirection()` runs a few lines of computations, it may be relevant to store it somewhere.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -327,6 +337,45 @@ public void QueueEmitToxin(Compound toxinCompound)
...
+            totalSlimeSecreted += slimeSecreted;
+        }
+
+        // Once we hit zero, start a cooldown timer
+        if (totalSlimeSecreted < Constants.MUCILAGE_MIN_TO_VENT)
+            slimeSecretionCooldown = Constants.MUCILAGE_COOLDOWN_TIMER;
```
Maxonovien, line 367: I'm unsure about what this means: `if (totalSlimeSecreted < Constants.MUCILAGE_MIN_TO_VENT)`, shouldn't venting just not happen? It seems to me that it occurs nonetheless in the lines above with `Eject compound`.

EDIT : Oh, ok, since this runs on multiple frames I guess it is in order to just know when to stop. I kept the comment above to show my confusion, and highlight that it might need some clarification (by variable names/further comments).

Also, just note that means that if you press that with less mucilage than the min, you'll vent on one frame nonetheless I believe. Not *very impactful*, granted.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -622,6 +675,10 @@ private void HandleCompoundAbsorbing(float delta)
         cloudSystem!.AbsorbCompounds(GlobalTransform.origin, grabRadius, Compounds,
             TotalAbsorbedCompounds, delta, Membrane.Type.ResourceAbsorptionFactor);
 
+        // Cells with jets aren't affected by mucilage
+        slowedBySlime = cloudSystem.AmountAvailable(mucilage, GlobalTransform.origin, 1.0f) >
+            Constants.COMPOUND_DENSITY_CATEGORY_FAIR_AMOUNT && SlimeJets.Count < 1;
```
Maxonovien, line 680: So if I understand well, the amount of slime does not effect how much you're slowed? A bit unintuitive to me, but this is fine code-wise.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -477,11 +477,15 @@ private void FleeFromPredators(Random random, Vector3 predatorLocation)
...
+        if (DistanceFromMe(predatorLocation) < 100.0f)
         {
-            MoveWithRandomTurn(2.5f, 3.0f, random);
+            // If the predator is right on top of the microbe, there's a chance to try and swing with a pilus
+            if (RollCheck(SpeciesAggression, Constants.MAX_SPECIES_AGGRESSION, random))
+                MoveWithRandomTurn(2.5f, 3.0f, random);
+
+            // There's also a chance to jet away if we can
+            if (RollCheck(SpeciesFear, Constants.MAX_SPECIES_FEAR, random))
+                SecreteSlime();
```
Maxonovien, line 488: As you only used simple `if`s you may end up in situations where the AI both swings and jets... not sure if this should be kept.

If you want to simply shift to if/else, note that the order of conditions will be impactful, and so AI will try aggression before fear, resulting in higher rates of attacks than runaways for similar values of aggression and fear.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -515,6 +519,10 @@ private void EngagePrey(Vector3 target, Random random, bool engulf)
         {
             SetMoveSpeed(Constants.AI_BASE_MOVEMENT);
         }
+
+        // Predators can use slime jets as an ambush mechanism
+        if (RollCheck(SpeciesAggression, Constants.MAX_SPECIES_AGGRESSION, random))
+            SecreteSlime();
```
Maxonovien, line 525: Given that you are out of the previous `if/else` statement, if a predator can shoot toxins, it may end up not moving towards the prey but only jetting towards them. I'm not sure such a behavior is desired?

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -249,10 +249,13 @@ protected override bool ShouldShowAgentsPanel()
         var colony = stage!.Player!.Colony;
         if (colony == null)
         {
-            return GetPlayerUsefulCompounds()!.IsSpecificallySetUseful(oxytoxy);
+            return GetPlayerUsefulCompounds()!.IsSpecificallySetUseful(oxytoxy) ||
+                GetPlayerUsefulCompounds()!.IsSpecificallySetUseful(mucilage);
```
Maxonovien, line 253: Not sure if `GetPlayerUsefulCompounds` runs a lot of computations or not, but in the former case it might be a good idea not to call it twice.

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -1042,9 +1061,6 @@ margin_bottom = 218.0
 rect_min_size = Vector2( 252, 0 )
 size_flags_horizontal = 0
 size_flags_vertical = 0
-__meta__ = {
-"_edit_use_anchors_": false
-}
```
Maxonovien, line 1047: Is this change intended?

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -1543,9 +1550,6 @@ size_flags_horizontal = 0
 size_flags_vertical = 0
 texture = ExtResource( 57 )
 expand = true
-__meta__ = {
-"_edit_use_anchors_": false
-}
```
Maxonovien, line 1548: Something was changed about the fonts (occurs several times), but I don't understand exactly what. I'll simply bring it to people more knowledgeable on front end to tell if this is a good thing or not.

src/microbe_stage/Microbe.Interior.cs
Maxonovien, line 680: So if I understand well, the amount of slime does not effect how much you're slowed? A bit unintuitive to me, but this is fine code-wise.

Maxonovien (changes requested):
Great work overall! A few style remarks, but I question how AI will handle it.

src/general/StageHUDBase.cs
hhyyrylainen: I think in this case the consistency and the fact that the panels are all setup the same way here with their child content, means that this current approach should be used here.

src/microbe_stage/Compound.cs
hhyyrylainen: The "Check" method is meant to verify the data is good. So instead of having this in a comment it should be added there to make sure the game doesn't start with incorrect parameters.

src/general/StageHUDBase.cs
```diff
@@ -926,6 +938,10 @@ protected virtual void UpdateCompoundBars(float delta)
         oxytoxyBar.MaxValue = compounds.GetCapacityForCompound(oxytoxy);
         GUICommon.SmoothlyUpdateBar(oxytoxyBar, compounds.GetCompoundAmount(oxytoxy), delta);
         oxytoxyBar.GetNode<Label>("Value").Text = oxytoxyBar.Value + " / " + oxytoxyBar.MaxValue;
+
+        mucilageBar.MaxValue = compounds.GetCapacityForCompound(mucilage);
+        GUICommon.SmoothlyUpdateBar(mucilageBar, compounds.GetCompoundAmount(mucilage), delta);
+        mucilageBar.GetNode<Label>("Value").Text = mucilageBar.Value + " / " + mucilageBar.MaxValue;
```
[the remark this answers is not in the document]
hhyyrylainen: This has gotten incrementally worse. Really goes to show that having code reviews consistently happen is better than hit a random PR that's not really at fault here with a "should be refactored" comment...

src/microbe_stage/MicrobeStage.tscn
hhyyrylainen: I always let these changes through, I don't know what they do but Godot seems to automatically just do these randomly, and it doesn't seem to effect anything.

src/general/StageHUDBase.cs
Maxonovien: Well, this may still happen to be changed through Godot, so one would then have to open the file and manually edit all the paths... not sure I'm entirely convinced that this approach is theoretically the best (now, there's consistency playing a role as well).

src/general/StageHUDBase.cs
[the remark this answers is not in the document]
Maxonovien: Sure, but now that we're here new PRs may be an opportunity to fix that. That will allow us to avoid big file refactory PRs...

src/general/StageHUDBase.cs
hhyyrylainen: I mean the code variables are suffixed with "VBoxContainer" as well and these grab the child from the related panel variable, I don't see these being anyway really modifiable just with Godot editor anyway even if a NodePath was used. So really, unless this whole thing is refactored, I wouldn't make one of these variables act entirely differently from the rest.

src/general/StageHUDBase.cs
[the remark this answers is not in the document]
hhyyrylainen: Our feature PRs usually already get "stuck" in review for a while even if we don't require them to refactor large other parts that aren't really needed. I'll leave this up to @Oliveriver to decide whether this PR will step outside the scope of this feature and refactor this part.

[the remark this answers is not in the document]
Maxonovien: I agree. I just mention the parts that *could* be refactored and leave them up for grab. Same with the resolved part on nodepaths below.

src/microbe_stage/Microbe.Interior.cs
Kasterisk: I gave this more thought. I think `SlimeJets` can stay so that `Microbe.SecreteSlime` can access all slime jets to do the actual secretion. The thing here is to make slime jets function more granularly, so the secretion logic, cooldown handling and movement speed boost all should be contained in `SlimeJetComponent`.

For example, adding speed boost can be done like this (taken from flagella code):
https://github.com/Revolutionary-Games/Thrive/blob/57f3233fccad0dc0110f30a144bbb4cea3b6759e/src/microbe_stage/organelle_components/MovementComponent.cs#L39-L42

And finally in `Microbe.SecreteSlime`, roughly what it'll look like:
```csharp
public void SecreteSlime(float delta)
{
        foreach (var jet in SlimeJets)
        {
                jet.Secrete();

               // Or if the cooldown should still account for all slime jets then `Secrete` can just return the amount secreted here.
               // totalSlimeSecreted += jet.Secrete();
        }

         // Cooldown handling...
}
```

Let me know if something is not clear.

Oliveriver:
I've addressed most of the TODOs that were left. Adding animations to the model and a placeholder sound effect required a bit of a rewrite for the slime secreting logic, so some review comments might be outdated now. I'll get to the comments and a few extra bug fixes tomorrow.

nickthenick:
Nice work on this, really excited to see it merged. 

One quick thing though, did you have a chance to check out the tooltip when hovering over the button? It says "Secrete slime" instead of "Secrete mucilage".

Oliveriver:
> Nice work on this, really excited to see it merged.
> 
> One quick thing though, did you have a chance to check out the tooltip when hovering over the button? It says "Secrete slime" instead of "Secrete mucilage".

I’m aware. As I said, I’ll address the review comments tomorrow.

src/general/StageHUDBase.cs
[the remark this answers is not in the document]
Oliveriver: I'm going to ignore this for now. It would be a relatively simple PR on its own, but I don't think it makes sense to bundle it in with this one.

src/microbe_stage/Microbe.Interior.cs
Oliveriver: I tried this, but due to the wildly varying amounts of compounds in clouds, it's really difficult to balance. I think this approach - if the cloud is clearly visible, it slows you - is intuitive enough.

src/microbe_stage/MicrobeAI.cs
Oliveriver: Good spot. I've made this an else if and swapped the order, and it seems to create more linear paths for escaping microbes, which looks much better.

src/microbe_stage/Microbe.Interior.cs
Oliveriver: I've revamped the approach here in my recent changes, so I don't think this problem applies anymore.

src/microbe_stage/Microbe.Interior.cs
Oliveriver: @Kasterisk I've added the closest approach I could manage. Due to things like the animation and sound effect, it turned out to be easier to keep a flag of whether each jet is active, which Microbe.Interior still controls. Then each jet component will eject mucilage and calculate its own force if it's set to active.

Oliveriver:
Due to the substantial review changes to how movement is calculated, this probably needs more testing now. I think this approach is more robust. One definite change: before, jetting when stationary wouldn't cause you to move, but now it does. That actually makes more sense to me from a physics perspective.

buckly90:
Works great! This is truely a versatile part that will make gameplay much more interesting.

The boosting is not the most effective for moving, but it can effectively close the gap between predator and prey if they are otherwise able to outrun you. The mucilage itself is excellent for escaping your own predators, and can even be used to ensnare more prickly prey.

A great feature, excellent work!

simulation_parameters/microbe_stage/compounds.json
```diff
@@ -88,6 +88,22 @@
...
     },
+    "mucilage": {
+        "name": "MUCILAGE",
+        "volume": 1,
+        "isCloud": true,
+        "decayRate": 0.99,
+        "isAgent": true,
+        "isEnvironmental": false,
+        "canBeDistributed": true,
+        "digestible": true,
```
Kasterisk, line 99: Just to make sure, gaining mucilage from digested cells is intentional for gameplay right?

Kasterisk (commented):
All seems good, really like how it turned out. One thing though, the secrete slime hotkey is missing the button press handling so it won't do anything if you press it.

84634E1A607A:
What about the save compatibility thing?

simulation_parameters/microbe_stage/compounds.json
Oliveriver: Ah, I suppose I hadn't considered that. My gut feel is that it shouldn't be, but other opinions are welcome.

Oliveriver:
> What about the save compatibility thing?

Haven't managed to fix that yet. I'll have another go tonight but might need help if I can't get it done then.

Oliveriver:
Right, fixed the save and load issue properly this time.

I have noticed I still have an annoying workaround in place for the animation not looping, and the model's animation is generally a bit broken. For instance, it doesn't jiggle. I swear it did at some point, so I probably changed something to break that.

A more reliable fix all around would be to ask for a redone model, so I'll do that in Discord now. That just leaves updated icons and an updated sound effect, and this should be done.

simulation_parameters/Constants.cs
```diff
@@ -224,6 +224,26 @@ public static class Constants
...
+    /// </summary>
+    public const float MUCILAGE_IMPEDE_FACTOR = 4.0f;
+
+    /// <summary>
+    ///   How much a cell's speed is increased when secreting slime (scaling with secreted compound amount)
+    /// </summary>
+    public const float MUCILAGE_JET_FACTOR = 600.0f;
+
+    /// <summary>
+    ///  Minimum stored slime needed to start secreting
```
hhyyrylainen, line 238:
```suggestion
    ///    Minimum stored slime needed to start secreting
```

simulation_parameters/Constants.cs
```diff
@@ -224,6 +224,26 @@ public static class Constants
...
+    /// </summary>
+    public const float MUCILAGE_MIN_TO_VENT = 0.01f;
+
+    /// <summary>
+    ///  Length in seconds for slime secretion cooldown
```
hhyyrylainen, line 243: Spaces also missing here

simulation_parameters/microbe_stage/compounds.json
hhyyrylainen: So this should be changed to false here, right?

simulation_parameters/microbe_stage/compounds.json
```diff
@@ -77,17 +77,33 @@
         "name": "OXYTOXY_NT",
         "volume": 1,
         "isCloud": false,
-        "isAlwaysUseful": true,
+        "isAgent": true,
         "isEnvironmental": false,
         "canBeDistributed": true,
-        "digestible": true,
+        "digestible": false,
```
hhyyrylainen, line 83: Does this change impact anything? For example engulf damage should not be broken.

simulation_parameters/microbe_stage/organelles.json
```diff
@@ -719,6 +719,40 @@
...
+        },
+        "components": {
+            "processor": {},
+            "storage": {
+                "capacity": 0.5
+            },
+            "slimeJet": {}
+        },
+        "prokaryoteChance": 0.5,
+        "chanceToCreate": 0.5,
```
hhyyrylainen, line 745: I think the chance to create should be set to be pretty low as the AI will find these pretty useless currently, right?

src/general/StageHUDBase.cs
```diff
@@ -213,12 +220,15 @@ public abstract class StageHUDBase<TStage> : Control, IStageHUD
...
     protected GridContainer? environmentPanelBarContainer;
     protected ActionButton engulfHotkey = null!;
+    protected ActionButton secreteSlimeHotkey = null!;
     protected ActionButton signallingAgentsHotkey = null!;
 
     // Store these statefully for after player death
     protected float maxHP = 1.0f;
     protected float maxATP = 1.0f;
 
+    protected List<Compound> allAgents = new();
```
hhyyrylainen, line 230: Should this be a `readonly` variable? This seems like something that should not be reassigned.

src/microbe_stage/CompoundBag.cs
```diff
@@ -130,6 +130,11 @@ public bool IsSpecificallySetUseful(Compound compound)
         return usefulCompounds.Contains(compound);
     }
 
+    public bool AreAnySpecificallySetUseful(List<Compound> compounds)
+    {
+        return usefulCompounds.Intersect(compounds).Any();
```
hhyyrylainen, line 135: I wonder if this is more or less efficient than the way I'd write this:
```c#
return compounds.Any(c => usefulCompounds.Contains(c));
```

src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -41,6 +41,9 @@ public class CompoundCloudPlane : CSGMesh, ISaveLoadedTracked
     private ImageTexture texture = null!;
     private FluidSystem? fluidSystem;
 
+    [JsonProperty]
```
hhyyrylainen, line 44: Why is this a JSON property if this is always set in `Init`?

src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -504,8 +510,8 @@ public Vector3 ConvertToWorld(int cloudX, int cloudY)
             if (compound == null)
                 break;
 
-            // Skip if compound is non-useful
-            if (!storage.IsUseful(compound))
+            // Skip if compound is non-useful or an agent
+            if (!storage.IsUseful(compound) || compound.IsAgent)
```
hhyyrylainen, line 514: I saw earlier that you removed the always useful property from the oxytoxy and didn't add it for mucilage. Would this code change be unnecessary if we marked all agents as always useful? Or another way would be to have a new property "AlwaysAbsorbed". I don't want to have this hyper special code in the compound cloud system.

simulation_parameters/microbe_stage/organelles.json
hhyyrylainen: Reading the rest of the code I see you did address that pretty comprehensively, still I think the slime jets should maybe be a bit more of a specialist organelle so maybe reducing this just a bit to "0.4" or "0.3" would still be good.

src/microbe_stage/organelle_components/ExternallyPositionedComponent.cs
```diff
@@ -12,6 +13,7 @@ public abstract class ExternallyPositionedComponent : IOrganelleComponent
     /// </summary>
     protected static readonly Vector3 DefaultVisualPos = Vector3.Forward;
 
+    [JsonProperty]
```
hhyyrylainen, line 16: Why was this added? I feel like this change has pretty huge implications. All the components should be reattached to the microbe anyway when loading a save.

src/microbe_stage/PlayerMicrobeInput.cs
```diff
@@ -68,6 +68,12 @@ public void EmitToxin()
         stage.Player?.EmitToxin();
     }
 
+    [RunOnKey("g_secrete_slime")]
+    public void SecreteSlime(float delta)
+    {
+        stage.Player!.QueuedSlimeSecretionTime += delta;
```
hhyyrylainen, line 74: Player maybe null, this will crash the game. Use `?` instead of `!`

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -2045,6 +2107,7 @@ __meta__ = {
 }
 
 [node name="PauseMenu" parent="." instance=ExtResource( 11 )]
+visible = true
```
hhyyrylainen, line 2110: accidental change

src/microbe_stage/organelle_components/SlimeJetComponent.cs
```diff
@@ -0,0 +1,130 @@
...
+    public Vector3 CalculateMovementForce(Microbe microbe, float delta)
+    {
+        if (!Active)
+            return Vector3.Zero;
+
+        var currentCellRotation = microbe.GlobalTransform.basis.Quat();
+        var direction = GetDirection();
+
+        // Eject mucilage at the maximum rate in a direction opposite to this organelle's rotation
+        var slimeSecreted = microbe.EjectCompound(
```
hhyyrylainen, line 62: This is unsafe code to trigger from `UpdateAsync`
See: https://github.com/Revolutionary-Games/Thrive/issues/3270 (absorb and eject compound operations can't be multithreaded).

src/microbe_stage/organelle_components/SlimeJetComponent.cs
```diff
@@ -0,0 +1,130 @@
...
+        animation = organelle.OrganelleAnimation;
+
+        if (animation == null)
+        {
+            GD.PrintErr("MovementComponent's organelle has no animation player set");
+            return;
+        }
+
+        // Workaround for animation not looping in Godot scene
+        animation.GetAnimation(animation.CurrentAnimation).Loop = true;
```
hhyyrylainen, line 96: This should be fixed in the actual animation file. I saw you exported the animation as a separate file, it should be possible to edit that to fix this. I didn't have to add code like this for the flagellum.

hhyyrylainen (changes requested):
Posted my review as individual comments as I once again didn't intend to do a full review but ended up doing one anyway...
I spotted a few serious issues, but other than that most of the changes looked fine to me.

simulation_parameters/microbe_stage/compounds.json
Oliveriver: Looks like I changed digestible for the wrong compound. I'll fix that.

src/microbe_stage/CompoundBag.cs
Oliveriver: No idea. I can change it if you'd prefer.

src/microbe_stage/organelle_components/ExternallyPositionedComponent.cs
Oliveriver: That may be a holdover from the temporary save fix I made that I forgot to delete. I'll remove it and see what happens.

src/microbe_stage/organelle_components/SlimeJetComponent.cs
Oliveriver: Is that specific to the cloud system or does that problem include adding/subtracting compounds from microbes? If it's both, I don't know how to fix this, since I need to know the amount of mucilage we can eject to calculate the movement force here.

src/microbe_stage/organelle_components/ExternallyPositionedComponent.cs
Oliveriver: Yep, I think that was it. No issues with saves after deleting it, so I've done that.

src/microbe_stage/CompoundBag.cs
hhyyrylainen: I'd prefer it so that code is more uniform. If it turns out Intersect is actually faster we have a bunch of code to migrate anyway.

src/microbe_stage/organelle_components/SlimeJetComponent.cs
hhyyrylainen: Adding and subtracting compounds is 100% safe and even intended for the async operation. That is for the *current* microbe, no other microbe's state may be modified. The way I've gotten around this in the microbe class is to have a "temporary" variable hold a result the async method calculated and then applying it in the sync part, though in your case it might be easier to have the Microbe class emit the slime and each slime component just adds some queued slime to be emitted.

Oliveriver: Ok, that's what I expected. I ended up implementing the temporary variable approach anyway, so take a quick look at that to check it's what you had in mind.

src/microbe_stage/CompoundCloudPlane.cs
hhyyrylainen: I think I do really want to keep the agents knowledge out of here. Oh and I just realized you use this to *skip* the absorption. I think we may want to have damaging agents in the future that damage cells that absorb them. So anyway a general "do not absorb" flag would be nice for the compound class so we have that flexibility in the future.

src/microbe_stage/organelle_components/SlimeJetComponent.cs
hhyyrylainen: Seems fine to me.

simulation_parameters/microbe_stage/organelles.json
```diff
@@ -719,6 +719,40 @@
...
+    "chanceToCreate": 0.3,
+    "mass": 0.1,
+    "displayScene": "res://assets/models/organelles/SlimeJet.tscn",
+    "displaySceneModelPath": "Armature/Skeleton/Cube",
+    "displaySceneAnimation": "AnimationPlayer",
+    "name": "SLIME_JET",
+    "consumptionColour": "#00ffcc",
+    "iconPath": "res://assets/textures/gui/bevel/parts/SlimeJetIcon.png",
+    "editorButtonGroup": "External",
+    "editorButtonOrder": 3
```
hhyyrylainen, line 754: This number needs to be unique in the group, I can see in the github diff that the number 3 is already in use by the organelle above this one.

simulation_parameters/microbe_stage/compounds.json
```diff
@@ -77,7 +77,7 @@
     "name": "OXYTOXY_NT",
     "volume": 1,
     "isCloud": false,
-    "isAlwaysUseful": true,
```
hhyyrylainen, line 80: I think we probably should keep the isAlwaysUseful marking for oxytoxy, or?

hhyyrylainen: I'll undo this change.

src/microbe_stage/CompoundBag.cs
hhyyrylainen: I'll commit changing this to `return compounds.Any(usefulCompounds.Contains);`

hhyyrylainen (approved):
I fixed the style issues this still had and I found quite a few bugs I had to fix, some of which undoubtedly I caused with the merge from master as that was very tricky in terms of conflicts. One thing to highlight is though that I fixed up the mucilage action bar button not actually working when pressed and being focusable.

PR #3593, opened 2022-07-26
EN locale copyediting

schnautzr:
Fix grammar, spelling. Change language to use "acrasin" and "grex" instead of "binding agent" and "cellular colony" where appropriate.

### DIFF
--- a/locale/en.po
+++ b/locale/en.po
@@ -87,12 +87,12 @@ msgstr "[thrive:input]g_move_forward[/thrive:input],[thrive:input]g_move_left[/t
 #: ../simulation_parameters/common/help_texts.json:10
 #: ../simulation_parameters/common/help_texts.json:100
 msgid "MICROBE_STAGE_HELP_MESSAGE_2"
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
"Acrasin definition: a chemical produced by slime moulds" 
also googling grex doesn't come up with anything cell related neither does "grex cell". So I question the validity of changing to these terms.

schnautzr:
One path to multicellularity is thought to be similar to what slime mold amoebas do, which is to release  a sticky signaling compound pheromone known as acrasin. A grex is a "flock" of such organisms adhering to one another via acrasin. There is an article on Wikipedia about this: https://en.wikipedia.org/wiki/Grex_(biology)

RacerBG:
I'm not the best person to comment on English grammar changes but this stuff is for Weblate. Also, having capital letters on certain words is a styling choice not a rule in the English language per se.

GameDungeon:
I agree with a couple of changes here, but I really don't agree with the grex stuff.

schnautzr:
Since the changes proposed here are of more than one nature, I'm going to close this and split it into changes so we can discuss this better.

PR #3596, opened 2022-07-26
Split patch population numbers into simulation and gameplay numbers

### LINKED ISSUE #3460: Immediate population effects now have serious bug which affects auto-evo
Species.ApplyImmediatePopulationChange is now incorrect after #2623 because it affects the per patch populations now, which is what auto-evo reads. Before it only updated the global population number so there wasn't a problem before.
My initial idea to fix is to have a "TemporaryGamePopulation" for each species in a patch, that population can be adjusted then for extinction calculations, and then that should be cleared when entering the editor and the real numbers with external effects and auto-evo are calculated.

hhyyrylainen:
this protects auto-evo runs from being messed up. This also makes death penalties calculate against the current population and not the simulation previous population

This PR does some stuff...

**Related Issues**

fixes #3460

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/AutoEvoRun.cs
+++ b/src/auto-evo/AutoEvoRun.cs
@@ -279,7 +279,7 @@ public class AutoEvoRun
         foreach (var entry in ExternalEffects)
         {
             var key = (entry.Species, entry.EventType, entry.Patch);
-            var speciesPopulation = entry.Species.Population;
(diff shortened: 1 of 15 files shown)


hhyyrylainen:
I'll merge this as I need this as a base for changes I plan to do today.

PR #3601, opened 2022-07-27
Tweaked the code for save deletion a bit

hhyyrylainen:
as I thought I noticed some weirdness but it turned out the code basically looks logically fine to me but I changed it a tiny bit

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/saving/SaveHelper.cs
+++ b/src/saving/SaveHelper.cs
@@ -38,7 +38,7 @@ public static class SaveHelper
         LastModifiedFirst,
 
         /// <summary>
-        ///   The first modified (on disk) save is first
+        ///   The first modified (on disk) save is first (oldest first)
         /// </summary>
         FirstModifiedFirst,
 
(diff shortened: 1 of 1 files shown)


src/saving/SaveHelper.cs
```diff
@@ -307,7 +307,11 @@ public static (int Count, ulong DiskSpace) CountSaves(Regex? nameMatches = null)
     public static void DeleteSave(string saveName)
     {
         using var directory = new Directory();
-        directory.Remove(Path.Combine(Constants.SAVE_FOLDER, saveName));
+        var finalPath = Path.Combine(Constants.SAVE_FOLDER, saveName);
+        directory.Remove(finalPath);
```
Kasterisk, line 311: Would it be more optimal to handle its return value instead?

hhyyrylainen: We aren't deleting that many files, so I preferred the API which is much more likely to detect a problem, after all I still remember that one Godot operation that just always returned Ok on Windows...

Kasterisk: Either is fine I guess.

Kasterisk (approved):

PR #3605, opened 2022-07-27
Fix various external effect bugs, including discrepancy in graphs and summary text

### LINKED ISSUE #3128: Fix the player population number disparency in the graphs versus auto-evo results and timeline
for some reason there are very different numbers shown in the different places for the player's current population. This needs to be fixed so that the numbers match up. I've looked at this before but I couldn't notice anything immediately obvious where the different numbers would come from.

hhyyrylainen:
I found various bugs in auto-evo external effects which I've now fixed and it seems this solves the discrepancy between the summary text and the graphs, at least in my quick test where I usually saw those problems before no longer happen.

**Related Issues**

closes  #3128

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/doc/style_guide.md
+++ b/doc/style_guide.md
@@ -270,9 +270,22 @@ Code style rules
 - Avoid globals. Especially in object trees where you can easily
   enough pass the reference along.
 
+- Do not use `string.Format` with a translated format string, as
+  translation mistakes can crash the game in that case. Instead either
+  use `LocalizedString`, `LocalizedStringBuilder`, or
+  `StringUtils.FormatSafe`. Those ways will automatically catch
+  exceptions from broken translations and return the format string
+  un-formatted. `StringUtils` will likely want to be invoked as an
+  extension method on the string (`"example".FormatSafe(...)`). If the
+  format string is not user supplied, normal `string.Format` is allowed,
+  but should be passed `CultureInfo.CurrentCulture` as the first
+  parameter as we want text shown to the user in the user's selected
+  locale.
+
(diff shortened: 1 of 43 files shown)


RacerBG (changes requested):
Almost everything is perfect but I don't know why the global events in the timeline are different from the local ones and the other statistics (before someone asks - my species are located in only one patch).

Steps to reproduce (I did this twice just in case):
1. Freebuild first generation
2. Second generation.
3. Check the global events.

hhyyrylainen:
@RacerBG what exactly are you trying to show with the screenshots? I presume the global timeline shows the total populations? I didn't go through the timeline logic entirely, I just touched parts of the code that dealt with external effects.

RacerBG:
> ely, I just touched parts of the cod

Yes, it shows the global population which means in multiple patches. But my species are present in only one patch when I made the screenshot. The population count is for some reason different.

hhyyrylainen:
The local event is fine, then? I guess the global timeline logic might do something differently, I'm pretty sure I got the external effect handling parts right so there is probably another reason for that. I'll open a follow up issue about that.

RacerBG:
> The local event is fine, then?

Local events are fine indeed.

> I guess the global timeline logic might do something differently, I'm pretty sure I got the external effect handling parts right so there is probably another reason for that. I'll open a follow up issue about that.

Sure.

hhyyrylainen:
Opened the issue: https://github.com/Revolutionary-Games/Thrive/issues/3611

PR #3616, opened 2022-07-28
Wait for transition before loading

### LINKED ISSUE #1406: Try to make loading use threading
to make the loading screen animation play

### LINKED ISSUE #3566: A black screen is now shown much longer than the loading screen when loading a save
probably due to the adjusted fade times, we should do something so that we would show the actual loading screen longer than the black screen. The only idea I really have is to make it so that the save load can't progress to the JSON deserialize step before the loading screen is fully visible. Because enabling multithreading for the loading will probably cause a ton of rare random crashes.

DrRadek:

Waits for transition to finish before loading (Suggested by hhyyrylainen).
It's currently not possible to implement multi-threaded loading without major slowdown due to Godot 3 limitations (see https://github.com/godotengine/godot/issues/36793).

**Related Issues**

closes #3566

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/saving/InProgressLoad.cs
+++ b/src/saving/InProgressLoad.cs
@@ -88,10 +88,14 @@ public class InProgressLoad
                 break;
             case State.ReadingData:
             {
+                // Wait for transition to finish.
+                // Workaround for #1406
+                if (TransitionManager.Instance.HasQueuedTransitions)
+                    break;
+
                 // Start suppressing loaded node deletion
                 TemporaryLoadedNodeDeleter.Instance.AddDeletionHold(Constants.DELETION_HOLD_LOAD);
 
-                // TODO: do this in a background thread if possible
                 try
                 {
                     // Invalid is given as the target state here, because it's unknown yet.
--- a/src/saving/InProgressLoad.cs
+++ b/src/saving/InProgressLoad.cs
@@ -79,23 +79,21 @@ public class InProgressLoad
                     MainGameState.Invalid,
                     TranslationServer.Translate("READING_SAVE_DATA"));
 
-                TransitionManager.Instance.AddSequence(ScreenFade.FadeType.FadeIn, 0.5f, null, false, false);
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
Thanks for the link about the threading thing, certainly very interesting. I did already know that Godot 4.0 makes background resource loading (more) possible.

I edited your PR description to mark this as actually not closing that other issue (which I updated with a new label and comment), but the more recent too long black screen issue #3566

src/saving/InProgressLoad.cs
```diff
@@ -88,10 +88,14 @@ private void Step()
...
             {
+                // Wait for transition to finish.
+                // Workaround for #1406
+                if (TransitionManager.Instance.HasQueuedTransitions)
+                    break;
+
                 // Start suppressing loaded node deletion
                 TemporaryLoadedNodeDeleter.Instance.AddDeletionHold(Constants.DELETION_HOLD_LOAD);
 
-                // TODO: do this in a background thread if possible
```
hhyyrylainen, line 94: I'd keep this comment but reword it to 
```
// TODO: do this in a background thread once possible
// https://github.com/Revolutionary-Games/Thrive/issues/1406
```

src/saving/InProgressLoad.cs
```diff
@@ -88,10 +88,14 @@ private void Step()
                 break;
             case State.ReadingData:
             {
+                // Wait for transition to finish.
```
hhyyrylainen, line 91: This class adds the transition on line 82, so shouldn't it be possible for this to also actually register the proper callback? I think checking if there are queued transitions is a dirty way to do it and I think we've managed to avoid using that elsewhere in the codebase.

hhyyrylainen (changes requested):

src/saving/InProgressLoad.cs
DrRadek: Sure.

hhyyrylainen (approved):
Code is fine now and seems to work in game.

PR #3622, opened 2022-07-29
Add info in new game settings mentioning performance options

Kasterisk:

~~Added max spawned entities option to the new game settings as per my forum post https://forum.revolutionarygamesstudio.com/t/increasing-game-performance/841/12.~~

Added text at the bottom of basic new game settings mentioning to the player that performance can be potentially improved in the options menu.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Oliveriver:
I'm afraid I still really don't think this fits with the other new game settings. It's jarring to have to set something that's attached to the program as a whole rather than a particular game alongside many game-specific things. Plus, it makes the basic view more complicated and intimidating.

I'm still in favour of having some kind of popup (or if that's too intrusive, a text tip) in the main menu or game prompting players to set the option in the options menu, perhaps with a button linking it. It could be shown only once, or have a checkbox to not show again like the tutorial popup.

But if I am overruled, in terms of this implementation: this duplicates code from the options menu and sets the entity limit without the player giving any kind of confirmation. The text hint also spans two lines, which doesn't look great.

hhyyrylainen:
Maybe just a bit of text in the simple new game options saying that "you can change performance related options at any time in the options menu to improve game performance"?

Kasterisk:
Sounds like a good compromise.

Kasterisk:
I've now added that text to the bottom of the new game settings menu. It also links to the options menu which can be clicked to open performance options.

hhyyrylainen:
@Oliveriver can you recheck this please?

Oliveriver:
Definitely a lot better. The highlighted text link is a bit odd as I don't think we do that anywhere else, and it's annoying to click back from options and return to the main menu rather than the new game settings. But both of those are minor and probably more trouble to fix than they're worth.

hhyyrylainen:
> The highlighted text link is a bit odd 

The Steam uploader dialog at least has a clickable link to open the TOS.
Does this use the same colours etc.?

Kasterisk:
I don't recall the one used by TOS link but for this one I use aqua color.

hhyyrylainen:
Here's how it looks:

kuva

It also reacts to the cursor being over it by turning into the clickable cursor.

Kasterisk:
Hm, the text font and the buttons size in that popup really needs some tweaking... Anyway, I guess I'll change the color to that for consistency then, and yes cursor will always change to the clickable cursor automatically for any `[url]` bbcode.

hhyyrylainen:
> and yes cursor will always change to the clickable cursor automatically for any [url] bbcode.

I hadn't looked at the code so I wasn't sure yet how you'd implemented it, but it's good to hear that you used the way I would have approached it.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -55,6 +57,8 @@ public enum ThriveBbCode
             // the compound names anyway
             extendedBbcode = value;
 
+            extendedBbcodeTranslated = TranslationServer.Translate(value);
```
hhyyrylainen, line 60: Can't this have problems regarding language change? This this should only be done in the ParseCustomTags method, I think. That's not super perfect either, but I think it'll be better than translating this here.

locale/en.po
```diff
@@ -2088,246 +2088,250 @@ msgstr "Mod Loading Errors"
...
-#: ../src/general/NewGameSettings.tscn:935
+#: ../src/general/NewGameSettings.tscn:228
+#: ../src/general/NewGameSettings.tscn:973
 msgid "RANDOM_SEED_TOOLTIP"
 msgstr "The value used for generating the world, which must be a positive integer"
 
-#: ../src/general/NewGameSettings.tscn:270
+#: ../src/general/NewGameSettings.tscn:267
+msgid "NEW_GAME_SETTINGS_PERFORMANCE_OPTIONS_INFO"
+msgstr "Note: you can change performance related options at any time in the [color=#3796e1][url]options menu[/url][/color] to improve game performance"
```
hhyyrylainen, line 2159: I think the url here should start a precedent for inter-GUI element clickable links.

I'd suggest something like:
`[url=thrive://GUI/OptionsMenu/Performance]options menu[/url]`

I don't think that needs to be done anything with at the moment but I think we'll eventually use links with that kind of syntax to refer to other pages in thriveopedia or refer from tooltips to the thriveopedia

src/general/NewGameSettings.cs
```diff
@@ -703,4 +706,11 @@ private void OnEasterEggsToggled(bool pressed)
     {
         _ = pressed;
     }
+
+    private void OnMetaClicked(object meta)
```
hhyyrylainen, line 710: I think this should have a slightly more descriptive name like `PerformanceNoteLinkClicked`

hhyyrylainen (changes requested):
In game behaviour is fine. I had a few small code related concerns, though.

locale/en.po
Kasterisk: I agree, hyperlinks will be the main characteristics of thrivepedia where it will be used in abundance, so it's good to keep this in mind.

hhyyrylainen: I was maybe a bit unclear but I meant you should at least update this translation text to be like that, but you don't need to start trying to make any logic to make the URL parsing or anything.

Kasterisk: Yes, I'll add it.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -191,7 +191,7 @@ private static string ParseCustomTagsString(string extendedBbcode)
                 var bbcodeNamespace = leftHandSide[0];
 
                 // Not a thrive bbcode, don't parse this
-                if (!bbcodeNamespace.Contains("thrive"))
+                if (!bbcodeNamespace.BeginsWith("thrive"))
```
Kasterisk, line 194: Ah wait hold on, I think I introduced a huge bug with this.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -446,7 +446,9 @@ string GetResizedImage(string imagePath, int width, int height, int ascent)
     /// </summary>
     private void ParseCustomTags()
     {
-        if (extendedBbcode == null)
+        var translated = TranslationServer.Translate(extendedBbcode);
```
hhyyrylainen, line 449: What if extendedBbcode is null? Would it make sense to check for that explicitly and skip the call to Translate?

Kasterisk: I think so.

src/gui_common/CustomRichTextLabel.cs
Kasterisk: Should be fixed.

hhyyrylainen (approved):
I think everything should be fine now.

PR #3630, opened 2022-07-30
Auto-Evo Exploring Tool tweaks

### LINKED ISSUE #3624: Improve performance of the evolutionary tree view
it seems due to having a lot of Godot buttons in it it can really tank the performance to single digit FPS. We might need to custom rendering for the round buttons in it or find some way to make the Godot buttons perform better. This is kind of critical to do before the evolutionary tree is included in the main game.

### LINKED ISSUE #3628: Allow running 20 generations with a single button press in the auto-evo tool
Right now the runs at once goes only up to 10, but 20 would be a better value in my opinion because that's about the time microbe stage should take so being able to get to the "final" state of a world with a single click would be nice.

This will make https://github.com/Revolutionary-Games/Thrive/issues/3627 more useful for benchmarking as well.

### LINKED ISSUE #3629: Refactor the evolutionary tree to not assume each generation is 100 Myr long
because we'll eventually change that by either changing the base year amount or once we implement the editor representing less time in the later game. So the evolutionary tree code should be changed to get the year numbers through the game world and not assume the years based on the generation index

84634E1A607A:
Fixes #3628 - I think we can allow 100 runs at a time (default value), as long as one can abort it at any moment.

Fixes #3629 - Now each generation passes its time to the evolutionary tree.

Fixes #3624 - Disabled draw line antialiasing.

Another small performance gain - Tweaks to drawing algorithm.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/EvolutionaryTree.cs
+++ b/src/auto-evo/EvolutionaryTree.cs
@@ -23,7 +23,7 @@ public class EvolutionaryTree : Control
     private readonly System.Collections.Generic.Dictionary<uint, (uint ParentSpeciesID, int SplitGeneration)>
         speciesOrigin = new();
 
-    private readonly System.Collections.Generic.Dictionary<int, double> generationTime = new();
+    private readonly System.Collections.Generic.Dictionary<int, double> generationTimes = new();
 
     private readonly ButtonGroup nodesGroup = new();
 
@@ -58,7 +58,7 @@ public class EvolutionaryTree : Control
 
         speciesOrigin.Add(luca.ID, (uint.MaxValue, 0));
         speciesNames.Add(luca.ID, luca.FormattedName);
-        generationTime.Add(0, 0);
+        generationTimes.Add(0, 0);
     }
 
     public override void _Draw()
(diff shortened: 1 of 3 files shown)


src/auto-evo/EvolutionaryTree.cs
```diff
@@ -16,15 +16,15 @@ public class EvolutionaryTree : Control
...
+    private readonly System.Collections.Generic.Dictionary<uint, List<EvolutionaryTreeNode>> speciesNodes = new();
 
     private readonly System.Collections.Generic.Dictionary<uint, string> speciesNames = new();
 
-    private readonly System.Collections.Generic.Dictionary<uint, EvolutionaryTreeNode> latestNodes = new();
-
     private readonly System.Collections.Generic.Dictionary<uint, (uint ParentSpeciesID, int SplitGeneration)>
         speciesOrigin = new();
 
+    private readonly System.Collections.Generic.Dictionary<int, double> generationTime = new();
```
hhyyrylainen, line 26:
```suggestion
    private readonly System.Collections.Generic.Dictionary<int, double> generationTimes = new();
```

hhyyrylainen (changes requested):

hhyyrylainen (approved):

PR #3633, opened 2022-07-30
Add rendering-related information to the graphics menu

### LINKED ISSUE #2373: Add a section in the options menu saying the current used GPU
this will help users to see if they are using integrated graphics or not.

Can be queried from the VisualServer:
https://docs.godotengine.org/fi/stable/classes/class_visualserver.html#class-visualserver-method-get-video-adapter-name

Also would be nice to show the current used video memory and used renderer

https://docs.godotengine.org/fi/stable/classes/class_visualserver.html#class-visualserver-method-get-render-info

`OS.GetCurrentVideoDriver()`

trashcognito:

This PR adds three entries to the Graphics tab of the Options menu:
- The name of the currently used GPU
- The name of the currently used display driver (Gles2 / Gles3)
- The amount of video RAM in use in mebibytes

**Related Issues**

Resolves #2373 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/OptionsMenu.cs
```diff
@@ -636,6 +653,40 @@ private void DisplayResolution()
...
+    private void DisplayGpuInfo()
+    {
+        var videoAdapterName = VisualServer.GetVideoAdapterName();
+        gpuName.Text = TranslationServer.Translate("AUTO_GPU_NAME")
+            .FormatSafe(videoAdapterName);
+
+        if (OS.GetCurrentVideoDriver() == OS.VideoDriver.Gles2)
+        {
+            // Gles2 is being used
+            displayDriverName.Text = TranslationServer.Translate("AUTO_DISPLAY_DRIVER_NAME")
```
hhyyrylainen, line 668: The launcher actually allows forcing the renderer to be gles2 so I wouldn't use "AUTO" in the name here.

src/general/OptionsMenu.cs
```diff
@@ -636,6 +653,40 @@ private void DisplayResolution()
...
+                .FormatSafe(TranslationServer.Translate("GLES2"));
+        }
+        else
+        {
+            // Gles3 is being used
```
hhyyrylainen, line 673: It's not 100% certain, especially once we switch to Godot 4 so I think explicitly checking for GLES3 would be better and having an else case for "unknown" would be good.

src/general/OptionsMenu.cs
```diff
@@ -636,6 +653,40 @@ private void DisplayResolution()
...
+
+        float videoMemoryInMebibytes = VisualServer.GetRenderInfo(VisualServer.RenderInfo.VideoMemUsed);
+
+        // Convert to mebibytes
+        videoMemoryInMebibytes /= Constants.MEBIBYTE;
+
+        // Round to 2 places after the floating point
+        decimal decimalVideoMemory = Math.Round((decimal)videoMemoryInMebibytes, 2);
+
+        videoMemory.Text = TranslationServer.Translate("AUTO_VIDEO_MEMORY")
```
hhyyrylainen, line 686: This should also not be "AUTO", I recommend a name like "VIDEO_MEMORY_MIB" or something.

src/general/OptionsMenu.cs
```diff
@@ -636,6 +653,40 @@ private void DisplayResolution()
...
+                .FormatSafe(TranslationServer.Translate("GLES3"));
+        }
+
+        float videoMemoryInMebibytes = VisualServer.GetRenderInfo(VisualServer.RenderInfo.VideoMemUsed);
+
+        // Convert to mebibytes
+        videoMemoryInMebibytes /= Constants.MEBIBYTE;
+
+        // Round to 2 places after the floating point
+        decimal decimalVideoMemory = Math.Round((decimal)videoMemoryInMebibytes, 2);
```
hhyyrylainen, line 684: Probably just use `Mathf.Round` here or cast to float, `decimal` type is used nowhere else in the game so I would not use it here for code consistency's sake.

hhyyrylainen (changes requested):
Overall this looks to be pretty close to being ready already, I have just a few slight code tweaking related comments.

src/general/OptionsMenu.cs
trashcognito: I did `Mathf.Floor(videoMemoryInMebibytes * 100) / 100;` to keep the two digits after the comma, which avoids the decimal type, though that approach seems a bit dirty. That, and I realized that always rounding down would be somewhat more consistent behavior, though that probably wouldn't matter that much here.

hhyyrylainen: I'm pretty sure `Mathf.Round` exists... what's the problem with using that?

trashcognito: I thought that rounding up sometimes (lying about used memory) and down other times (lying about not using memory) would have been a problem but on second thought, 0.01 mebibytes dont matter that much. I'll use `Mathf.Round()`.

src/general/OptionsMenu.cs
```diff
@@ -636,6 +653,46 @@ private void DisplayResolution()
...
+                .FormatSafe(TranslationServer.Translate("UNKNOWN_DISPLAY_DRIVER"));
+        }
+
+        float videoMemoryInMebibytes = VisualServer.GetRenderInfo(VisualServer.RenderInfo.VideoMemUsed);
+
+        // Convert to mebibytes
+        videoMemoryInMebibytes /= Constants.MEBIBYTE;
+
+        // Round to 2 places after the floating point
+        var decimalVideoMemory = Mathf.Round(videoMemoryInMebibytes * 100) / 100;
```
hhyyrylainen, line 690: Instead of doing this complex thing, you should just use `Math.Round((float)adasdsad, 2)` I didn't remember that Mathf round method didn't support specifying how many decimal places to use.

src/general/OptionsMenu.cs
```diff
@@ -636,6 +653,46 @@ private void DisplayResolution()
...
+                .FormatSafe(TranslationServer.Translate("UNKNOWN_DISPLAY_DRIVER"));
+        }
+
+        float videoMemoryInMebibytes = VisualServer.GetRenderInfo(VisualServer.RenderInfo.VideoMemUsed);
+
+        // Convert to mebibytes
+        videoMemoryInMebibytes /= Constants.MEBIBYTE;
+
+        // Round to 2 places after the floating point
+        var decimalVideoMemory = Math.Round((float)videoMemoryInMebibytes, 2);
```
hhyyrylainen, line 690: CI is about to probably complain about the useless cast here, also the "decimalVideoMemory" variable is not pretty badly named, so I'd just inline this on line 693 and remove the explicit float cast.

hhyyrylainen (changes requested):
A few findings from playtesting. Screenshot for reference:

- You should set a max width and line break on the GPU text
- "Display Driver" should be renamed to "Used renderer" 
- "Video Memory" should be renamed to "Currently used video memory"
- And finally "Gles3" and 2 should be all upper case like "GLES3"

trashcognito:
>     * You should set a max width and line break on the GPU text

image

Would the default 1/3 split with the separator suffice? Or should the middle separator be scrapped entirely to allow the GPU name to expand all the way to the left?

hhyyrylainen:
> Would the default 1/3 split with the separator suffice? Or should the middle separator be scrapped entirely to allow the GPU name to expand all the way to the left?

I wouldn't scrap the middle separator just for that as that probably has more complications.

hhyyrylainen:
Now a new problem is that the options menu can get way too tall, I think you'll need to add a scroll container in the graphics tab to fix that.
Also looks like in your latest commit you removed the UTF8 BOM marker from the start of OptionsMenu.cs

hhyyrylainen:
Should the right side padding be modified to be inside the scroll container rather than outside, so that the content isn't literally against the scroll bar?

trashcognito:
> Should the right side padding be modified to be inside the scroll container rather than outside, so that the content isn't literally against the scroll bar?
> 
> 

Yes, I am trying to do that right now. I am currently copying over the margin logic from the Inputs tab (nested ScrollContainer with MarginContainer).

locale/en.po
```diff
@@ -2279,381 +2279,418 @@ msgstr "Basic"
...
 
-#: ../src/general/OptionsMenu.cs:988
+#: ../src/general/OptionsMenu.cs:662 ../src/general/OptionsMenu.tscn:428
+msgid "AUTO_GPU_NAME"
+msgstr "{0}"
+
+#: ../src/general/OptionsMenu.cs:668 ../src/general/OptionsMenu.cs:674
+#: ../src/general/OptionsMenu.cs:680 ../src/general/OptionsMenu.tscn:453
+msgid "CURRENT_USED_RENDERER_NAME"
+msgstr "{0}"
```
hhyyrylainen, line 2293: I think having many almost exact translation keys will be a bit confusing to the translators and will result in me getting a question about this, so as these values just directly put a technical string here, I'd want to entirely remove these translation keys and just directly assign the final text in to the places used in the GUI. For the scene you can add "PLACEHOLDER" to the editor description and then put an example GPU name. And for the renderer you can just put GLES3 there, it'll work fine as the translation key and text is fine.

hhyyrylainen (changes requested):
I didn't find more issues in playtesting, just one more comment regarding simplifying these changes for the translators of the game.

locale/en.po
trashcognito: So, like this?
`// Gles2 is being used
            usedRendererName.Text = "GLES2";`

trashcognito: I had decided against putting string literals directly in the codebase without translation as I thought that might have caused some issues with arabic or chinese character related localization. Or did you mean something more like:
`// Gles2 is being used
            usedRendererName.Text = TranslationServer.Translate("GLES2");`?

hhyyrylainen: The GLES2 part should be translatable. Just so that one layer of the containerization is removed so `usedRendererName.Text = TranslationServer.Translate("GLES2");`

hhyyrylainen: >  Or did you mean something more like:

That's exactly what I meant, and already managed to type a reply before I received an email about your second comment.

84634E1A607A: In the godot editor there's a option named description or sth. Type 'PLACEHOLDER' there.

trashcognito:
Should the `PLACEHOLDER` also apply to `VIDEO_MEMORY_MIB`? The `AUTO_RESOLUTION` label didn't have the placeholder tag  and that's the most similar thing to this option that is in the codebase right now.

hhyyrylainen:
I'm fine with just leaving it as-is I often just leave the translation key in GUI elements (when the value is dynamically set by the code) without adding a separate value to be shown in the Godot editor while working on the GUI scene.

trashcognito:
So, since `VIDEO_MEMORY_MIB` is already a translation key, that one should not have the placeholder tag?

hhyyrylainen:
> So, since VIDEO_MEMORY_MIB is already a translation key, that one should not have the placeholder tag?

It doesn't *require* a placeholder. And that's exactly as you say, because the key is used elsewhere as well, having it in the tscn file as well, does no harm.

hhyyrylainen (approved):
I fixed the last placeholder related problem.

PR #3635, opened 2022-08-01
Tweaked map connection generation logic

84634E1A607A:
So that it looks a bit better.

Before:
Before

After:
After

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/PatchMapDrawer.cs
+++ b/src/microbe_stage/editor/PatchMapDrawer.cs
@@ -430,7 +430,13 @@ public class PatchMapDrawer : Control
             probablePaths.Add((new[] { startCenter, intermediate1, intermediate2, endCenter }, -i));
         }
 
-        return probablePaths.OrderBy(CalculatePathPriority).First().Path;
+        // Choose a best path
+        return probablePaths.Select(p => (p.Path, CalculatePathPriorityTuple(p)))
+            .OrderBy(p => p.Item2.RegionIntersectionCount)
+            .ThenBy(p => p.Item2.PathIntersectionCount)
+            .ThenBy(p => p.Item2.StartPointOverlapCount)
+            .ThenByDescending(p => p.Item2.Priority)
+            .First().Path;
     }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


84634E1A607A:
And this pr is open to any code-wise suggestions about patch map generator / drawer.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -555,33 +555,38 @@ private void AdjustPathEndpoints()
...
     /// </summary>
     /// <remarks>
     ///   <para>
-    ///     Note that priority should be within (-10, 10), and the return value is relative.
+    ///     Priority should be within (-10, 10)
     ///   </para>
     /// </remarks>
-    private int IntersectionCountWithPriority((Vector2[] Path, int Priority) pathPriorityTuple)
+    /// <returns>
+    ///   Calculated priority: 8bit region intersection; 8bit path intersection;
```
hhyyrylainen, line 566:
```suggestion
    ///   Calculated priority: 8-bit region intersection; 8-bit path intersection;
```
either that way or "8 bit" is the way bitness should be written.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -555,33 +555,38 @@ private void AdjustPathEndpoints()
...
+    ///   8bit start point intersection; 8bit priority
```
hhyyrylainen, line 567: Is there a reason why all the variables are packed within a single uint? The way I see this, it's as is regionIntersectionCount just has a really high modifier, meaning that there's a huge penalty if the intersection count is over 0 because it is in the high bytes of the variable. I think explicitly calculating the score multipliers into a single int *or* using an approach of Selecting the original value + a tuple of the numbers here, and then doing sort like:
```
probablePaths.Select(p => (p, CalculatePathPriority(p)).OrderBy(p => p.Item2.Item1).ThenBy(p => p.Item2.Item2).ThenBy(p => p.Item2.Item3);
```
would actually clearly spell out what this code does. If that LINQ code is not the operation you intended, then I don't think the code as implemented does what you wanted.

84634E1A607A: I don't even know there is such a thing. However, I wonder if this costs more time.

hhyyrylainen: There should be at most hundred or so nodes? So I would suspect it doesn't. Also I hope LINQ is smart enough to bunch up all of the order by and then by operations to do it in a single sort.
If you prefer I lined out that other approach where explicit priority multipliers and using `+` operator to construct a final int, would also be fine and I think that would also convey the intention well enough.

84634E1A607A: Fine

hhyyrylainen: I found a StackOverflow about this: https://stackoverflow.com/questions/41213076/what-is-the-time-complexity-of-linq-orderby-thenby-method-sequence
and they said there that a sequence of ThenBy.ThenBy.ThenBy etc. will only result in a *single* sort operation that uses multiple sort keys

hhyyrylainen (approved):
The new connections seem fine to me, though I don't really remember there being many problems in the first place...

PR #3641, opened 2022-08-02
Fix dynamic MP logic a bit

### LINKED ISSUE #3558: Specific sequences of editor actions can result in nonsensical dynamic MP result
We have a couple of reports of some cases where the MP results don't make sense. I didn't get anyone to confirm these steps so they might not be accurate but here's at least one report:
> i removed a spike for the mouth. (which was fine, loss of mp like normal) then i added like 3 lysosomes when i should have only had enough mp for 1. same thing happened with when i was adding spikes for the outer cells and when i was putting on more mitochondria for the speed cells. i found out the signaling agent one by placing and undoing things and seeing if i lost or gained mp.

Those steps should cause:
unknown

This might be early multicellular specific. Also I think I saw a related problem in a video where placing an organelle at one specific position was 5 MP cheaper than elsewhere, even though undo was used, but this might be a different problem. This issue is basically here to call attention that the dynamic MP system has at least one bug with inconsistency.

84634E1A607A:
## Description
(This is added because the complexity of Dyn MP.)
Previously, Dyn MP uses a way similar to bubble sort to check if the action data can be merged together. Which means:
```
1. Place 1
2. Place 2
3. Remove 2
```

May be merged into
```
1. Move 2 -> 1
2. Place 2
```

This PR aims to make combine operation more optimal by adding a key function `GetLeastCostCombineTuple`.

Given the data to combine and a list of actions that may be combined to, this function calculates the cost saved by combination for each action and picks the one that saves the most. Then it returns some relevant info.

**Related Issues**

Extensive testing is needed to prove if this fixes #3558 
Workaround #3661 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -191,6 +169,37 @@ public bool HexPlacedThisSession<THex>(THex hex)
...
+    /// <param name="currentData">The data you want to combine</param>
+    /// <param name="previousData">A list of data currentData may combine with</param>
+    /// <returns>
+    ///   Three-element tuple:
+    ///   CostDelta is non-positive, indicating the amount of MP you will regain;
+    ///   MinimumCostActionData is the action data to combine with (not yet combined);
+    ///   Mode is the interference mode currentData has with MinimumCostActionData.
+    /// </returns>
+    private static (int CostDelta, EditorCombinableActionData MinimumCostActionData, ActionInterferenceMode Mode)
+        GetLeastCostCombineTuple(EditorCombinableActionData currentData,
```
hhyyrylainen, line 184:
```suggestion
        GetLeastCostCombinedTuple(EditorCombinableActionData currentData,
```

84634E1A607A: The action data returned is not combined. I think renaming like that will make confusions.

hhyyrylainen: Well the original name confused me already... Maybe I can come up with a better suggestion if you briefly explain what this method actually does? I don't think I'm yet mentally prepared to the brain melting trying to read and understand this PR in its entirety is going to cause...

84634E1A607A: Added in the description

src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -33,89 +33,73 @@ public int WhatWouldActionsCost(IEnumerable<EditorCombinableActionData> actions)
...
-                        compareIndex = copyLength;
+                        compareIndex -= 2;
                         break;
                     }
-
-                    default:
-                        throw new ArgumentOutOfRangeException();
                 }
+
+                // If mode is as follows, no more check is needed.
```
hhyyrylainen, line 100:
```suggestion
                // If mode is the following, no more checks are needed.
```

src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -191,6 +175,37 @@ public bool HexPlacedThisSession<THex>(THex hex)
         return History.OfType<HexPlacementActionData<THex>>().Any(a => a.PlacedHex == hex);
     }
 
+    /// <summary>
+    ///   Pick the data with which you can regain maximum MP to combine.
+    /// </summary>
+    /// <param name="currentData">The data you want to combine</param>
+    /// <param name="previousData">A list of data currentData may combine with</param>
```
hhyyrylainen, line 182:
```suggestion
    /// <param name="previousData">A list of data <see cref="currentData"/> may combine with</param>
```

src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -191,6 +175,37 @@ public bool HexPlacedThisSession<THex>(THex hex)
...
+        GetLeastCostCombineTuple(EditorCombinableActionData currentData,
+            IEnumerable<EditorCombinableActionData> previousData)
+    {
+        return previousData.Select(data =>
+        {
+            var interferenceMode = currentData.GetInterferenceModeWith(data);
+
+            return (interferenceMode switch
+            {
+                ActionInterferenceMode.Combinable => ((EditorCombinableActionData)currentData.Combine(data))
```
hhyyrylainen, line 199: I'm a bit suspicious about the cast here. What guarantees this doesn't fail?

src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -191,6 +175,37 @@ public bool HexPlacedThisSession<THex>(THex hex)
         return History.OfType<HexPlacementActionData<THex>>().Any(a => a.PlacedHex == hex);
     }
 
+    /// <summary>
+    ///   Pick the data with which you can regain maximum MP to combine.
```
hhyyrylainen, line 179:
```suggestion
    ///   Finds the way to combine an action with previous actions so that the maximum amount of MP is saved.
```

src/microbe_stage/editor/EditorActionHistory.cs
84634E1A607A: What doesn't fail? The cast? EditorCom... combines EditorCom... always return a EditorCom..

src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -191,6 +175,37 @@ public bool HexPlacedThisSession<THex>(THex hex)
...
+    /// <param name="currentData">The data you want to combine</param>
+    /// <param name="previousData">A list of data <see cref="currentData"/> may combine with</param>
+    /// <returns>
+    ///   Three-element tuple:
+    ///   CostDelta is non-positive, indicating the amount of MP you will regain;
+    ///   MinimumCostActionData is the action data to combine with (not yet combined);
+    ///   Mode is the interference mode currentData has with MinimumCostActionData.
+    /// </returns>
+    private static (int CostDelta, EditorCombinableActionData MinimumCostActionData, ActionInterferenceMode Mode)
+        GetLeastCostCombineTuple(EditorCombinableActionData currentData,
```
hhyyrylainen, line 190: `FindCheapestActionToCombineWith` 
I think a name like that would make this clearer as to what this does.

src/microbe_stage/editor/EditorActionHistory.cs
hhyyrylainen: I read the code in my IDE, it's safe to call this as `currentData` so we can assume Combine returns a `EditorCombinableActionData`.

src/microbe_stage/editor/EditorActionHistory.cs
84634E1A607A: Fine. Wait for a minute.

hhyyrylainen:
I just playtested this and it seems to not at least break anything in a major way so once the code is cleaned up a bit I think this can be merged.

hhyyrylainen (approved):

PR #3643, opened 2022-08-02
Time based reproduction (and passive phosphate and ammonia generation)

hhyyrylainen:

https://forum.revolutionarygamesstudio.com/t/revamping-compound-clouds/913/22

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Deus-Codes:
Obviously needs balancing, but the mechanic itself works properly and, imo, makes a fine addition to Thrive. The game felt properly paced, allowing you to reach the editor even if you don't find any phosphate/ammonia and preventing you from reaching the editor 10 seconds after you left it as a eukaryote just because you spawned near a bunch of organelles. A few observations:

1. The phosphate/ammonia bar appears somewhat glitched whenever you obtain a good amount of size, where both bars somewhat filled upon exiting the editor. Another problem related to this is the fact that the phosphate/ammonia bar would indicate that you are ready to enter the editor even when you aren't (in other words, the bars are full and turned white even though the button itself won't work). The noise indicated you are ready to reach the editor still works properly though. I'm not sure exactly what the root of this issue is, but it might be related to the extra phosphate/ammonia you spawn with if you had a particularly successful life before. We'll likely have to have the player spawn in with preset compound amounts (0 phosphate, 0 ammonia, a small but decent amount of food, etc.) and we can reassign additional phosphate/ammonia to be additional population bonuses from reproducing.
2. It sometimes seemed like the bar would fill up slower if you have a surplus of ammonia/phosphate. Then after you exit a phosphate/ammonia cloud, the bar would rapidly quicken up its pace. Could again be a GUI error.
3. The length of time it took to reproduce didn't change depending on your morphology. I wasn't taking exact time and the GUI messed with my perception, but prokaryotic and eukaryotic gameplay length felt somewhat similar in length. 
4. Tutorial changes were good I feel, though we can always do better. I wasn't paying much attention to them this time around, but I'll pay attention more as I test more.

I'm sure I'm missing some observations I made, but overall I like this direction. More testing will be needed and I'd like to understand exactly what is happening with the GUI more before balancing attempts. Great job, and I'd love more people to test this commit.

hhyyrylainen:
I've now gone ahead and fixed up the multicellular stage, tweaked various things, and I actually properly reworked the tutorial.
Also there's now the 2 new game options for controlling if passive reproduction is on and if maximum reproduction compound use is limited or not.

> Another problem related to this is the fact that the phosphate/ammonia bar would indicate that you are ready to enter the editor even when you aren't (in other words, the bars are full and turned white even though the button itself won't work). 

I found a couple of bugs with inconsistent calculation, but maybe the biggest thing was that before the bars would be `used for reproduction + current storage`. Now with the model where it takes time for the storage to convert to actual progress, that's probably too confusing, so I've changed the progress bars to just show the used compounds for reproduction and take no storage into account.

> The phosphate/ammonia bar appears somewhat glitched whenever you obtain a good amount of size, where both bars somewhat filled upon exiting the editor. 

Most likely just a factor of you having enough resources left over after the editor. With the maximum reproduction compound usage rate, this is not as powerful as it was before.

>  I'm not sure exactly what the root of this issue is, but it might be related to the extra phosphate/ammonia you spawn with if you had a particularly successful life before. We'll likely have to have the player spawn in with preset compound amounts (0 phosphate, 0 ammonia, a small but decent amount of food, etc.) and we can reassign additional phosphate/ammonia to be additional population bonuses from reproducing.

Let's leave the discussion of entirely unlinking the player's previous actions (making the editor into a get out of jail free button) for later.

> It sometimes seemed like the bar would fill up slower if you have a surplus of ammonia/phosphate. Then after you exit a phosphate/ammonia cloud, the bar would rapidly quicken up its pace. Could again be a GUI error.

It should not be really possible. I've changed things around now that half the time ammonia is preferred and half the time phosphates are preferred to make the two reproduction bars fill at a much more even rate.
Still if you have just ammonia or phosphate you will progress faster on that bar because the passive generation is lower than the amount you can actually progress per second. This still keeps it possible that acquiring compounds speed up reproduction.

> The length of time it took to reproduce didn't change depending on your morphology. I wasn't taking exact time and the GUI messed with my perception, but prokaryotic and eukaryotic gameplay length felt somewhat similar in length.

I set the base cost to 20 each, so it was 20 times higher than individual organelles. I've now lowered it to 10 so it's now just the equivalent of 10 organelles.

Oliveriver:
Only just noticed the changes to new game settings here. I have a few thoughts on those too:

* The difficulty panel is now too tall. Maybe it should be converted to a scroll container now?
* The text explaining the new options isn't particularly clear to me. I don't understand what limit growth compound usage does.
* The new options aren't logged with the rest when entering a game. Should be added to WorldGenerationSettings.ToString().
* Similarly, the values for these options in the WGS object don't seem to be updated when the difficulty preset is changed. Should be updated in the setter for WorldGenerationSettings.Difficulty.

Leinourdian:
- Passive growth feels really fast in the early game. LUCA fully grows in about a minute. This means that the player has barely enough time to read the tutorials before the cell is already ready to reproduce. It also means that the cell doesn’t need to collect any glucose to survive until reproduction, since LUCA can survive about 2 minutes and 30 seconds with just its starting resources if it’s not moving. It also makes clouds of ammonia and phosphates feel useless in the early game. I think it’d be better for LUCA to take about 3 minutes to grow with just passive growth so that there’s something that pushes the player to collect clouds. Although that may be pretty slow for big and sessile cells… If passive growth scaled with size, it would be easier to balance this stuff I think.

- Are the free compounds supposed to depict passive absorption? If that’s the case, it’s weird that it stops once you have enough resources to reproduce.

- I think I kind of prefer what Thim did in his fork https://github.com/adQuid/Thrive/commit/b02c9ac5cdaf3e7339d97cbf22c85e3f3e5f5cd3#diff-e6c61db0a04801cea02f21af2ab928629a17ba6fa451c1b48b662767b0706a5a (Microbe.Interior). There passive absorption sends the nutrients to the storage instead. Absorption also scales with size. And AI species can only reproduce once, eliminating exponential growth (although that's not so realistic).

- I like that ammonia and phosphates take time to go from the storage to growth. The transfer rate feels a bit slow for big cells though, so maybe it could speed up if you’re big or have a nucleus?

- The tutorial could perhaps encourage the player to collect clouds of ammonia and phosphates more, maybe even pointing to them like it does with glucose.

- It would be nice to get the passive growth rate as a tweakable constant in the new world settings (and maybe the base reproduction cost and storage to growth transfer rate too, unless they would add too much bloat). That way players could experiment with different values and provide more feedback.

- Personally I like passive growth quite a lot even though I do understand your worries about changing the game experience too much. I think it would be great to get this to the hands of more testers. If this still seems too radical, then maybe this can be set off by default for 0.5.10 or the values can be set low enough that most players ignore them.

hhyyrylainen:
I'll reply to a few new points, but I'm not really going to make changes based on these. I implemented this based on a dev forum discussion, so this is kind of the wrong place to bring up design issues that were talked about elsewhere up here.

A further balancing pass will be done before 0.5.10 (if no one else does I'll make sure it gets done once the release date is approaching).



> I think I kind of prefer what Thim did in his fork https://github.com/adQuid/Thrive/commit/b02c9ac5cdaf3e7339d97cbf22c85e3f3e5f5cd3#diff-e6c61db0a04801cea02f21af2ab928629a17ba6fa451c1b48b662767b0706a5a (Microbe.Interior). There passive absorption sends the nutrients to the storage instead. Absorption also scales with size. And AI species can only reproduce once, eliminating exponential growth (although that's not so realistic).

It was an explicit design decision to not allow cells to stockpile the passive reproduction bonus

> The tutorial could perhaps encourage the player to collect clouds of ammonia and phosphates more, maybe even pointing to them like it does with glucose.

With the focus moving onto more time based reproduction, I don't think this is necessary. Though I'd allow someone to make a PR with this new tutorial (and I think this maybe should be hidden behind a "hand holding tutorial" mode in the new game options).

> It would be nice to get the passive growth rate as a tweakable constant in the new world settings (and maybe the base reproduction cost and storage to growth transfer rate too, unless they would add too much bloat). That way players could experiment with different values and provide more feedback.

Well, we ran out of space without a scroll bar in the game difficulty options. I didn't want to over complicate things but it would be possible to allow tweaking the rates, but I ultimately decided that reproduction is a such core mechanic that big differences wouldn't really make that much sense. I also ended up basically making all the difficulties (except easy not having a max growth rate limit) the same.

The only reason I added the option in the first place is so that people who really don't like the new way can turn it off.

buckly90:
An excellent addition, took a bit to get used to though.

With the passive income of reproductive compounds, gathering resources feels much less stressful as I do not feel constantly pressed to find phosphate and ammonia. Instead I am reassured that in due time I will be able to progress regardless, with any ammonia and phosphates I do find being a delightful bonus rather than mandatory need.

The rate of passive income feels good to me. Not too fast as to make other sources redundant, but not too slow as to feel meaningless itself.

I do feel that the passive income being constant and not scaling with size could be an issue, since it slows down rather painfully with the moderately large cells that photosynthesizers tend to become. But I would like to keep an eye on player feedback on that matter before pursuing solutions. For now I do believe this change is just fine.

hhyyrylainen:
I'll merge this and open the issues about rebalancing.

hhyyrylainen:
The follow up issues about balancing:
https://github.com/Revolutionary-Games/Thrive/issues/3654
https://github.com/Revolutionary-Games/Thrive/issues/3655

PR #3644, opened 2022-08-03
Improved the English text for the auto-evo exploring tool

hhyyrylainen:
also fixed one string with a trailing newline Poedit complained about each time

**Related Issues**

translators, pretty fairly, complained that the English text was not clear at all in the auto-evo exploring tool, so I tried to improve that without getting too wordy

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/locale/en.po
+++ b/locale/en.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
 "POT-Creation-Date: 2022-08-02 13:55+0300\n"
-"PO-Revision-Date: 2022-07-31 20:14+0300\n"
+"PO-Revision-Date: 2022-08-03 16:32+0300\n"
 "Last-Translator: Georgi Georgiev (RacerBG) <g.georgiev.shumen@gmail.com>\n"
 "Language-Team: English <https://translate.revolutionarygamesstudio.com/projects/thrive/thrive-game/en/>\n"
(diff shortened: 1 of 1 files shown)


locale/en.po
```diff
@@ -1032,75 +1032,75 @@ msgstr "Viewer"
...
 
 #: ../src/auto-evo/AutoEvoExploringTool.tscn:308
 msgid "NEW_BIODIVERSITY_INCREASING_SPECIES_POPULATION"
-msgstr "New biodiversity increasing species population"
+msgstr "Initial population for biodiversity increasing species"
 
 #: ../src/auto-evo/AutoEvoExploringTool.tscn:322
 msgid "PROTECT_MIGRATIONS_FROM_SPECIES_CAP"
-msgstr "Protect migrations from species cap"
+msgstr "Protect just migrated populations from species cap"
```
84634E1A607A, line 1079: Is "species cap" "force extinction"? If so, I think directly say "from force extinction" is better.

hhyyrylainen: I don't want to constraint people wanting to change the auto-evo too much by being overtly specific here.
I don't particularly like forced extinctions as a concept so I'd like us to move away from that in terms of the species cap, so I'd prefer to leave this as a more general way to say this.

hhyyrylainen: I'll merge this to get these changes on weblate for translators. If there's some better wording (that isn't too much longer as we need to be careful with these as these don't have line wrapping enabled) a suggestion can be made on weblate.

PR #3651, opened 2022-08-04
Added a linebreak between Species and MicrobeSpecies detail

### LINKED ISSUE #3650: Add line separation before species stage in auto-evo exploring tool
unknown

caused by: https://github.com/Revolutionary-Games/Thrive/pull/3644

Line breaks are not allowed at the end of translation strings. Poedit gives you an error popup each time you save. The line break needs to be added some other way between those elements

84634E1A607A:
**Related Issues**
Fixes #3650 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/AutoEvoExploringTool.cs
+++ b/src/auto-evo/AutoEvoExploringTool.cs
@@ -709,6 +709,8 @@ public class AutoEvoExploringTool : NodeWithInput
             species.FormattedNameBbCode, species.ID, species.Generation, species.Population, species.Colour.ToHtml(),
             string.Join("\n  ", species.Behaviour.Select(b => b.Key + ": " + b.Value)));
 
+        speciesDetailsLabel.ExtendedBbcode += "\n";
+
         switch (species)
         {
             case MicrobeSpecies microbeSpecies:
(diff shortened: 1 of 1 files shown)


src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -709,6 +709,8 @@ private void UpdateSpeciesDetail(Species species)
             species.FormattedNameBbCode, species.ID, species.Generation, species.Population, species.Colour.ToHtml(),
             string.Join("\n  ", species.Behaviour.Select(b => b.Key + ": " + b.Value)));
 
+        speciesDetailsLabel.ExtendedBbcode += "\n";
```
hhyyrylainen, line 712: Maybe this would be better in the switch below? That way if the next bit of info is not added then the line change is not added unnecessarily.

hhyyrylainen: So something like `speciesDetailsLabel.ExtendedBbcode += "\n" + TranslationServer.Translate(`

84634E1A607A: But then we have to add a line break for each and every species (Multicellular, etc) later.

hhyyrylainen: Not really a problem, in most text appending cases that's what you need anyway because at the end of a section you don't know if a new section is going to be added or not.

hhyyrylainen (approved):

PR #3669, opened 2022-08-07
Split timeline and tree so that timeline is always shown

### LINKED ISSUE #3632: Make evolutionary tree's timeline bar float at the top of the container
As the tree is basically some kind of timetree and it's important for the timeline to stay visible. The timeline should float, this way even if you move to the bottom of the tree it won't disappear and if you zoom in/out on the tree, the timeline should lengthen or shorten accordingly so that it won't go off sync.

84634E1A607A:
**Related Issues**
Fixes #3632

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
What's the approach used here? To me it seems like maybe this now re-renders each frame the mouse drags around or what? Based on a quick look of the drag container being removed. I think maybe instead what this should do is use a scroll container that has the timeline at the top and then the bottom view would use the drag container. Then this is just a case of needing a bit of code to sync the horizontal values of the timeline and the draggable tree view.

84634E1A607A:
I'm not quite sure. The current DraggableScrollContainer seems buggy. ScrollContainer is buggy as well. The key concept here is to keep scale property unchanged but calculate position & size & etc codewide. This allows for full control to the Tree content instead of coming up with hacky workarounds for all those Godot limitations & bugs.

84634E1A607A:
Plus this seems to improve performance.

hhyyrylainen:
> Plus this seems to improve performance.

Okay, I guess I need to actually read the code to figure out how this works... some extra comments explaining how the overall approach works would be highly appreciated by me (and on the offchance that someone else also looks at this, they would benefit as well).

84634E1A607A:
> I need to actually read the code to figure out how this works...

As for why this improves performance... I don't know either.

84634E1A607A:
I am now confident that EvolutionaryTree will run 60 fps no matter how bit the tree is.

Nunez2196 (approved):

Nunez2196:
I tested this a bit while reviewing. Worked pretty good, even on my laptop.

84634E1A607A:
Line corners are now smooth.

Kasterisk (changes requested):
A couple of observations:
- I think the timeline shouldn't be affected by zooming, the numbers shouldn't shrink or grow, instead the "length" is what changes. I can't think of more good example of how this works but MS Word's ruler bar might be a good reference?
- The cursor immediately changes to the default icon everytime I mouse over the tree area.

84634E1A607A:
The cursor mode is "Drag" but somehow it isn't a hand.

84634E1A607A:
> I think the timeline shouldn't be affected by zooming

Why? Then its font size is not changed accordingly, and it may seem strange when scrolled small.

Kasterisk:
To rephrase that, I mean the visual _size_ of the timeline bar shouldn't be affected as I mentioned with the time ticks (also the line itself shrinks when you zoom out). If it were to be like this, surely nothing on the timeline is readable once you zoom out far enough right?

What I suggest is something like this:
timeline

84634E1A607A:
I understand, but I still think that it will be a little strange.

Kasterisk (changes requested):
Timeline starting to look very good now. However, the cursor for the tree area is now apparently on reverse. When hovering any of the circles, the normal icon is shown, but outside of it, the blue variant of the cursor is shown which is usually used for interactable elements (patches on patch map). Obviously this is the wrong way around.

84634E1A607A:
Seems like other buttons are using the def Arrow cursor, so I'll just set tree cursor to default.

Kasterisk (approved):
I'm satisfied with the results.

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -1,21 +1,79 @@
...
+    ///   Default size of <see cref="EvolutionaryTreeNode"/>.
+    /// </summary>
+    private static readonly Vector2 TreeNodeSize = new(30, 30);
+
+    /// <summary>
+    ///   Auxiliary vector for <see cref="DRAW_MARGIN"/>
+    /// </summary>
+    private static readonly Vector2 DrawMargin = new(DRAW_MARGIN, DRAW_MARGIN);
+
+    // ReSharper disable 4 times RedundantNameQualifier
```
hhyyrylainen, line 76: I think it's better to use the disable and then enable comment rather than this "4 times" thing.

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -1,21 +1,79 @@
...
+
+    /// <summary>
+    ///   Draws outside this margin will be omitted to improve performance.
+    /// </summary>
+    private const float DRAW_MARGIN = 30.0f;
+
+    /// <summary>
+    ///   Default size of lato-small font.
+    /// </summary>
+    private const float SMALL_FONT_SIZE = 14.0f;
```
hhyyrylainen, line 64: Would it be possible to actually get the font size from the actually used font?

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -25,14 +83,49 @@ public class EvolutionaryTree : Control
 
     private readonly System.Collections.Generic.Dictionary<int, double> generationTimes = new();
 
+    /// <summary>
+    ///   All EvolutionaryTreeNodes are in this group so that they works radio.
```
hhyyrylainen, line 87:
```suggestion
    ///   All EvolutionaryTreeNodes are in this group so that they work as radio buttons.
```

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -41,92 +134,58 @@ public class EvolutionaryTree : Control
...
 
+        timeline = GetNode<Control>(TimelinePath);
+        timeline.Connect("draw", this, nameof(TimelineDraw));
+        timeline.Connect("gui_input", this, nameof(GUIInput), new Array(true));
+        timeline.Connect("mouse_exited", this, nameof(MouseExit));
+
+        tree = GetNode<Control>(TreePath);
+        tree.Connect("draw", this, nameof(TreeDraw));
+        tree.Connect("gui_input", this, nameof(GUIInput), new Array(false));
+        tree.Connect("mouse_exited", this, nameof(MouseExit));
```
hhyyrylainen, line 152: Why are these signals not setup in Godot editor?

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -41,92 +134,58 @@ public class EvolutionaryTree : Control
...
-
-        for (int i = 0; i <= latestGeneration; i++)
+        if (dirty)
         {
-            DrawLine(new Vector2(LEFT_MARGIN + i * GENERATION_SEPARATION + treeNodeSize.x / 2, TIMELINE_LINE_Y),
-                new Vector2(LEFT_MARGIN + i * GENERATION_SEPARATION + treeNodeSize.x / 2,
-                    TIMELINE_LINE_Y + TIMELINE_MARK_SIZE),
-                Colors.Cyan, TIMELINE_LINE_THICKNESS);
+            // SizeFlagVertical is set to ShrinkEnd, so that adjusting min size adjusts actual size.
+            // timeline.RectMinSize = new Vector2(0, sizeFactor * TIMELINE_HEIGHT);
```
hhyyrylainen, line 181: What's this commented out line of code for?

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -226,18 +293,216 @@ private void BuildTree(uint id, ref uint index)
...
+
+            // If species extinct, line ends at the last node; else it ends at the right end of the tree
+            var lineEnd = lastNode.LastGeneration ?
+                lastNode.Center :
+                new Vector2(treeRightPosition, lineStart.y);
+
+            TreeDrawLine(lineStart, lineEnd);
+        }
+
+        latoSmallItalic.Set("size", sizeFactor * SMALL_FONT_SIZE);
```
hhyyrylainen, line 451: Here's the font size, so I assume with a similar call in _Ready this could get the default font size automatically and not have to assume it is 14.

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -226,18 +293,216 @@ private void BuildTree(uint id, ref uint index)
...
         else
         {
-            var mid = to - new Vector2(GENERATION_SEPARATION / 2.0f, 0);
-            DrawLine(from, new Vector2(mid.x, from.y), Colors.DarkCyan, 4.0f);
-            DrawLine(new Vector2(mid.x, from.y), new Vector2(mid.x, to.y), Colors.DarkCyan, 4.0f);
-            DrawLine(new Vector2(mid.x, to.y), to, Colors.DarkCyan, 4.0f);
+            var mid = to - new Vector2(sizeFactor * GENERATION_SEPARATION / 2.0f, 0);
+            tree.DrawLine(from, new Vector2(mid.x, from.y), Colors.DarkCyan, lineWidth);
+
+            // We draw vertical line a little longer so the turning point is good.
```
hhyyrylainen, line 501:
```suggestion
            // We draw vertical line a little longer so the turning point looks better.
```

hhyyrylainen (changes requested):
I have a few code related comments. I couldn't really spend the time to try to fully understand the changes but they looked overall fine. Especially moving more stuff into the evolutionary tree scene is good.
I noticed one slightly not intuitive part in playtesting: when I scroll with the mouse the scroll is not centered around the mouse. So for example if I zoom out all the way and then place my mouse in the upper right corner and zoom in, the evolutionary tree ends up being zoomed to the top left corner. That's now how UI zooming should work. So the zooming should be fixed to be centered around the current cursor position.
Also there's now a slight graphical problem when zooming out, the circles have the bottom few pixels cutoff so the botton sides of the cirles don't look like circles anymore. This isn't super critical but if there's an easy fix I'd like this to be fixed.

hhyyrylainen (changes requested):
Everything else except the zooming seems good now.

Here's how you can see how the zooming is still not the best. Zoom out all the way, place the cursor on one of the species nodes on the right side and start scrolling in. You'll end up zoomed in at the center of the graph instead of at the right side.

84634E1A607A:
That's because the tree size is smaller than the container horizontally. When you zoom in the tree is once again aligned to the lefttop corner. I don't think this can be fixed.

hhyyrylainen:
I kind of suspected as much, but I think GUI libraries solve this by having "virtual" size for their items that is larger than the size they take up, that way they can implement mouse zoom properly. There's anyway the issue with the non-round circle bottoms so I could open also an issue about this, if you don't think you can solve this now.

hhyyrylainen (approved):
I'll open issues for my remaining playtesting things I found wrong with this.

hhyyrylainen:
Opened follow up issues:
- https://github.com/Revolutionary-Games/Thrive/issues/3715
- https://github.com/Revolutionary-Games/Thrive/issues/3716

Issue #1085, opened 2020-04-20
Could reduce lag when changing patch in the editor


hhyyrylainen:
by background loading (if / when that is possible in godot) the images and only then applying the background for the camera.
The background setting method could be changed to take in Textures.

hhyyrylainen:
Background image loading seems a bit problematic: https://github.com/Revolutionary-Games/Thrive/issues/2078

Perhaps Godot 4.0 will improve the situation?

hhyyrylainen:
I added a bunch of scheduling code for resource loading, which should make it pretty easy to make the lag spike lessened: https://github.com/Revolutionary-Games/Thrive/pull/3682 (still can't fully move things to background thread before Godot 4.0).

PR #3673, opened 2022-08-09
Patch selection change performance

### LINKED ISSUE #3660: Improve performance when changing shown patch on the report tab
With a bunch of data that causes quite big lag spikes. In our code the biggest things seem to be https://github.com/Revolutionary-Games/Thrive/issues/3659 as well as:
Kuvakaappaus - 2022-08-04 14-48-30

and:
Kuvakaappaus - 2022-08-04 14-48-54

I think ultimately the biggest performance thing is how many Godot objects get destroyed and recreated each time. I think instead we'll need to either cache the timeline display objects and update them or make bigger rich text blocks to have fewer overall objects.

### LINKED ISSUE #3674: Improve LineChart tooltip performance
image
image
image

84634E1A607A:

Increased performance when changing shown patch.

**Related Issues**
Closes #3660 
Closes #3674

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Does this not close https://github.com/Revolutionary-Games/Thrive/issues/3660 ? If this addresses the other points but not the input remap registering, I think it's still fine to mark this as closing that issue.

84634E1A607A:
Lag is still observable although better than before.

hhyyrylainen:
Is the lag caused by https://github.com/Revolutionary-Games/Thrive/issues/1085 ? If so I think #3660 could be closed.

84634E1A607A:
Caused by #3674

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -752,7 +752,6 @@ private void RebuildMapNodes()
         }
 
         UpdateNodeSelections();
-        NotifySelectionChanged();
```
hhyyrylainen, line 755: Shouldn't this try to detect if the rebuild caused the selection to change, and if it did then run this method?
I worry about potential hard to debug bugs in the future popping up because of this when we have non-static maps.

src/microbe_stage/editor/MicrobeEditorReportComponent.tscn
```diff
@@ -368,7 +369,12 @@ size_flags_vertical = 3
 custom_constants/margin_right = 10
 custom_constants/margin_bottom = 20
 
-[node name="VBoxContainer" type="VBoxContainer" parent="MarginContainer/HSplitContainer/PanelContainer/MarginContainer/VBoxContainer/TimelinePanel/MarginContainer/ScrollContainer/MarginContainer"]
+[node name="GlobalEventsContainer" type="VBoxContainer" parent="MarginContainer/HSplitContainer/PanelContainer/MarginContainer/VBoxContainer/TimelinePanel/MarginContainer/ScrollContainer/MarginContainer"]
+margin_right = 809.0
+margin_bottom = 202.0
+
+[node name="LocalEventsContainer" type="VBoxContainer" parent="MarginContainer/HSplitContainer/PanelContainer/MarginContainer/VBoxContainer/TimelinePanel/MarginContainer/ScrollContainer/MarginContainer"]
+visible = false
```
hhyyrylainen, line 377: Isn't the current default the local view? So shouldn't this visibility wit the global and local container be the opposite? Or am I misremembering things?

src/microbe_stage/editor/PatchMapDrawer.cs
84634E1A607A: But how to?

hhyyrylainen: I think doing something like looking at the existing data, storing the selected patch if there is one. Then here check the selected patch again, and if it doesn't match the earlier value, then call this method. If there isn't already one a method to find the currently selected patch is needed to reduce code duplication.

src/microbe_stage/editor/MicrobeEditorReportComponent.tscn
84634E1A607A: It's true. Anyway visibility is dynamically changed in the code during init. (Still I'll change that)

src/microbe_stage/editor/PatchMapDrawer.cs
84634E1A607A: There's already a variable tracking selected patch.

hhyyrylainen: Then it should be made sure that that variable works correctly when the drawer is cleared and a change notification is sent correctly here.

84634E1A607A: I am sure that NotifySelectionChanged shouldn't be called here.

hhyyrylainen: Let's consider the following (future scenario):
- The player has selected a patch in a map
- The GUI shows them the patch details etc.
- Now an event happens which changes the map so rebuild map is called. This causes the selected patch to be removed from the map.
- If the selection change notify callback is not called the other GUI state is incorrect.
That's why I think it is necessary in some cases to trigger NotifySelectionChanged when rebuilding the map.

84634E1A607A: If the selected patch is removed, `SelectedPatch` property must be reassigned, which triggers `NotifySelectionChanged`

hhyyrylainen: I could not find any code which sets SelectedPatch to null, leading me to believe that the map rebuild will not function correctly in the scenario I outlined above.

84634E1A607A: We don't have this scenario currently either.

hhyyrylainen: I'm bringing this up because removing this code function call here leaves the potential for a future bug, which will be harder than necessary to figure out. That's why I think if this method call is removed now, the related potential bug should also be fixed now.

84634E1A607A: What's more, the code `NotifySelectionChanged` leads to reads `SelectedPatch` property. So if you leave it unchanged, even if you call `NotifySelectionChanged` here GUI will still be in the incorrect state.

hhyyrylainen: Yes, but my point being that seeing the NotifySelectionChanged call here, makes the person in the future changing the map rebuild logic more likely to do it correctly.

84634E1A607A: Let's comment this out (

hhyyrylainen: I went ahead and wrote the right approach myself.

src/gui_common/tooltip/ToolTipHelper.cs
```diff
@@ -18,9 +20,14 @@ public static class ToolTipHelper
     /// <summary>
     ///   Instantiates a default tooltip scene
```
hhyyrylainen, line 21: This comment is incorrect now, should say this returns a cached tooltip instance or creates a new one if the cache is empty.

src/microbe_stage/editor/PatchMapDrawer.cs
84634E1A607A: I still doubt if that line will ever be called

hhyyrylainen: It might not. But maybe once we have plate tectonics simulation and the patch map is shown in the auto-evo exploring tool, then I foresee this being able to happen.

hhyyrylainen (approved):
I playtested and didn't see any glaring bugs with this and the patch change performance was definitely better. Code's now also been hammered into shape.

PR #3702, opened 2022-08-14
Day And Night Cycle

Nunez2196:

Adds a day / night cycle to Thrive

**Related Issues**

no related issue; based on suggestion and demand

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Design discussion about this will hopefully happen on the dev forums: https://forum.revolutionarygamesstudio.com/t/day-and-night/928

src/microbe_stage/MicrobeCamera.cs
```diff
@@ -56,6 +56,9 @@ public class MicrobeCamera : Camera, IGodotEarlyNodeResolve, ISaveLoadedTracked
     [JsonProperty]
     public float InterpolateZoomSpeed = 0.3f;
 
+    [JsonIgnore]
+    public ShaderMaterial MaterialToUpdate { get { return materialToUpdate; } }
```
Oliveriver, line 60: This can just be `public ShaderMaterial MaterialToUpdate => materialToUpdate`.

hhyyrylainen: It could be.
I commented on discord that it would be much better to tell the camera "hey, here's the current light level" rather than allow any other code outside this class to mess with this material object directly.

Nunez2196: I had no idea that's what you were trying to communicate, I changed it now

Nunez2196:
should things start to become desperately stalled, this feature is completely functional, no real bugs I noticed. Code is obviously going to need to be done better in some places.

I'm more or less just polishing now.

Nunez2196:
Kasterisk suggested we should do something like this:

"PatchMap should hold all of the day night cycle information, the current time of day, etc. It applies light intensity to every patches accordingly through their BiomeConditions. PatchManager ticks (updates) those properties in some public method, accessing GameWorld in doing so. And finally, MicrobeStage calls this method every frame. Let me know what you guys think of this approach"

some gentle advice per hh:
"The updating of the day/night system (as I said before), I'd put in the microbe stage class and make it for example so that it updates the biome data once per second, and applies the background image tint to the camera each frame. That way systems that really shouldn't need to be made more complex don't need to know at all."

These people are both major reviewers so we should probably listen.

GameDungeon:
I've removed the static DayNightCycle instance and set it up to work per hh's suggestion.

GameDungeon:
Close in favor of #3742 due to branch issues.

Issue #3663, opened 2022-08-04
Entity limit can erroneously constrain cloud spawns


Oliveriver:
Inspired by [this post](https://www.reddit.com/r/thrive/comments/wfkkhx/ive_been_playing_for_an_hour_and_no_phosphate_has/) on our subreddit, I had a look into cloud spawning and found an issue that doesn't make much sense, but I can reliably reproduce it.

* Change Constants.SPAWN_SECTOR_SIZE to something small to make this more visible. I've been using 20.
* Set max entities to normal.
* Start a game and swim around a bit to get to a decent entity count.
* Change max entities to tiny.
* Swim away from the spawned objects. Nothing else will spawn in your path until they despawn. While this is correct behaviour for chunks and microbes (since we're over the entity limit), this isn't correct at all for compound clouds.
* Set max entities back to normal. Everything now spawns fine.

I can only conclude that something is preventing everything (including clouds) from being spawned when we're over the entity limit. While it's obvious with a small sector size, it's much less obvious with the default size, which may be how instances like the subreddit post have managed to slip through.

Oliveriver:
Found the culprit. At SpawnSystem line 271:

```
        // If there are already too many entities, don't spawn more
        if (existing >= Settings.Instance.MaxSpawnedEntities)
            return;
```

This check runs before we actually look at the spawner types, so it does indeed constrain everything. I think the fact this method is named SpawnEntities might have caused the confusion.

Issue #3686, opened 2022-08-10
Fix entity limit not being always respected by the game


hhyyrylainen:
we've gotten a few screenshots where the entity count goes up to like 270 for some reason. It seems like that the spawn system limiting is broken or there is some other way the game ends up making that many entities.

Also could investigate to make sure that all entities are added to the spawned group so that the spawn system sees them.

hhyyrylainen:
I think I found the issue: https://github.com/Revolutionary-Games/Thrive/pull/3754/commits/0c4f15af4a7fec67b7b54bfd319db4b39998884c

so I'll close this before the full 0.5.10 release if this issue is not reported again for the RC.

PR #3704, opened 2022-08-14
Fix cloud spawns being constrained by entity limit

Oliveriver:

Fixes cloud spawns being erroneously prevented if we're over the entity limit. Also renames a method to prevent a similar bug creeping in again.

**Related Issues**

Fixes https://github.com/Revolutionary-Games/Thrive/issues/3663

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/SpawnSystem.cs
+++ b/src/microbe_stage/SpawnSystem.cs
@@ -345,11 +345,8 @@ public class SpawnSystem : ISpawnSystem
         int spawns = 0;
         foreach (var spawnType in spawnTypes)
         {
-            if (spawnType is MicrobeSpawner)
+            if (!SpawnsBlocked(spawnType) && spawnType is MicrobeSpawner)
             {
-                if (SpawnsBlocked(spawnType))
-                    continue;
-
                 spawns += SpawnWithSpawner(spawnType,
                     playerLocation + new Vector3(Mathf.Cos(angle) * Constants.SPAWN_SECTOR_SIZE * 2, 0,
                         Mathf.Sin(angle) * Constants.SPAWN_SECTOR_SIZE * 2), ref spawnsLeftThisFrame);
@@ -367,8 +364,7 @@ public class SpawnSystem : ISpawnSystem
     /// </summary>
     private bool SpawnsBlocked(Spawner spawnType)
     {
-        return spawnType is not CompoundCloudSpawner &&
-            estimateEntityCount >= Settings.Instance.MaxSpawnedEntities.Value;
+        return spawnType.SpawnsEntities && estimateEntityCount >= Settings.Instance.MaxSpawnedEntities.Value;
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


Oliveriver:
I removed the check causing the bug and moved other checks to when we actually know the spawn types. I went back and forth trying to figure out whether these checks should be further up the chain for performance reasons, so I'm open to suggestions there.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -361,7 +363,16 @@ private void SpawnMicrobesAroundPlayer(Vector3 playerLocation, ref int spawnsLef
     }
 
     /// <summary>
-    ///   Does a single spawn with a spawner
+    ///   Checks whether we're currently blocked from spawning this type
+    /// </summary>
+    private bool SpawnsBlocked(Spawner spawnType)
+    {
+        return spawnType is not CompoundCloudSpawner &&
```
hhyyrylainen, line 370: I think a slightly cleaner thing to do would be to extend the Spawner classes to know if they spawn entities or not. And then this check would just be if the type spawns entities and the estimate is over the limit. The spawn system is purposefully written so that it wouldn't care about the concrete types of the spawners.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -372,43 +383,40 @@ private int SpawnWithSpawner(Spawner spawnType, Vector3 location, ref int spawns
             return spawns;
         }
 
-        if (spawnType is CompoundCloudSpawner || estimateEntityCount < Settings.Instance.MaxSpawnedEntities)
```
hhyyrylainen, line 375: I see we already had this code, but I think cleaning things up now would be worth it. If you aren't fully up for it, I can do that refactor the way I like.

hhyyrylainen (changes requested):
For gameplay testing I cranked down the entity limit to tiny and started a freebuild game. I was able to confirm that a lot of clouds seem to still spawn even when the game is constantly at the entity limit. So the gameplay side seems good, but I have refactoring idea on how to make this with a cleaner design.

src/microbe_stage/SpawnSystem.cs
Oliveriver: I considered that but didn't think it was worth it, since there's only one spawner not spawning entities that we can easily check. But I can definitely make that change if you'd prefer. It would certainly help in case we add further spawner types.

hhyyrylainen: That's my reasoning why it should be done. Overall is just goes against the design of the spawn system if the system knows anything about the concrete spawner types.

hhyyrylainen:
I saw you made one change to the microbe spawns method in your latest commit. Could that have any impact on https://github.com/Revolutionary-Games/Thrive/issues/3686 ? I guess with the previous overall level limit that may have already guarded stuff but I still have a slight feeling that change might do something.

Oliveriver:
> I saw you made one change to the microbe spawns method in your latest commit. Could that have any impact on #3686 ? I guess with the previous overall level limit that may have already guarded stuff but I still have a slight feeling that change might do something.

It might do something but I'm not sure. There are no spawns that weren't checked before that are checked now, but they might be checked at times closer to the actual time of spawning.

hhyyrylainen (approved):

Issue #3706, opened 2022-08-15
Auto-evo configuration is not correctly stored in save.


84634E1A607A:
Due to the fact that AutoEvoConfiguration is an IRegistryType, just internal name is saved.

84634E1A607A:
My first reaction is to serialize the config as JSON and store it in the internal name.

hhyyrylainen:
Here's my comment on the PR I made which is very related to this:
If the problem is that the auto-evo settings are a registry type, I'd suggest the following refactoring:
- Remake the current class into a base class AutoEvoConfiguration which has the settings in it
- Create a new PredefinedAutoEvoConfiguration class which is a registry type

That should solve the issue as the auto-evo configuration will then be fully serialized and deserialized in the save. Also should probably slap a `[JsonObject(IsReference = true)]` on the base class.

PR #3707, opened 2022-08-15
Quick workaround for auto-evo config save problem

84634E1A607A:

This PR does some stuff...
Load part for #3706 

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/WorldGenerationSettings.cs
+++ b/src/general/WorldGenerationSettings.cs
@@ -7,6 +7,8 @@ public class WorldGenerationSettings
 {
     private DifficultyPreset difficulty = null!;
 
+    private AutoEvoConfiguration autoEvoConfiguration = SimulationParameters.Instance.AutoEvoConfiguration;
+
     public WorldGenerationSettings()
     {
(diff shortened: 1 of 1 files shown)


Kasterisk:
Maybe the property should just be assigned in `_Ready` since that's when `AutoEvoConfiguration` is initialized?

hhyyrylainen:
If the problem is that the auto-evo settings are a registry type, I'd suggest the following refactoring:
- Remake the current class into a base class AutoEvoConfiguration which has the settings in it
- Create a new PredefinedAutoEvoConfiguration class which is a registry type

That should solve the issue as the auto-evo configuration will then be fully serialized and deserialized in the save. Also should probably slap a `[JsonObject(IsReference = true)]` on the base class.

hhyyrylainen:
I think I came up with a proper fix: https://github.com/Revolutionary-Games/Thrive/pull/3720
at least I reworked quite a lot of the fundamental saving bits.

PR #3708, opened 2022-08-15
Fixed incorrect display of engulfment values and reword engulfment invulnerability

### LINKED ISSUE #3665: Wording on the invulnerable to engulfment membrane tooltips needs improving
the wording that it is invulnerable to engulfment is confusing as it is no longer actually true, instead it is only true when the attacker doesn't have the necessary lysosomes. So the wording of that bit should be improved without making it too much longer. Or the tooltip needs to be reworked some other way.

Kasterisk:

- Fixed incorrect digestion speed and efficiency values increase shown on the lysosome tooltip.
- Fixed code comment mentioning incorrect base digestion efficiency.
- Reword "Invulnerable To Engulfment" to "Resistant to basic engulfment" on the cellulose and chitin tooltips.
- Remove rounding to int for the digestion efficiency value display on the organism stats panel as that reduces accuracy.

**Related Issues**

Closes #3665 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/Microbe.Interior.cs
```diff
@@ -1416,7 +1416,7 @@ private void HandleDigestion(float delta)
                 var amount = MicrobeInternalCalculations.CalculateDigestionSpeed(Enzymes[usedEnzyme]);
                 amount *= delta;
 
-                // Efficiency starts from 40% up to 60%. This means at least 7 lysosomes are needed to achieve
+                // Efficiency starts from 30% up to 60%. This means at least 7 lysosomes are needed to achieve
```
hhyyrylainen, line 1419: I think for future proofing this comment it might be good to say that check the actual constants / code for the real values. Or optionally all numbers could be removed from this comment.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -172,7 +172,7 @@ private void UpdateTotalDigestionSpeed(float speed)
     private void UpdateTotalDigestionEfficiency(float efficiency)
     {
         digestionEfficiencyLabel.Format = TranslationServer.Translate("PERCENTAGE_VALUE");
-        digestionEfficiencyLabel.Value = Mathf.RoundToInt(efficiency * 100);
+        digestionEfficiencyLabel.Value = efficiency * 100;
```
hhyyrylainen, line 175: I think we should round at most to 2 or 3 decimal places and not display like up to 8, that'd be way too much, so `Math.Round(efficiency * 100, 2)` should be used here I think. That's for two decimal places, also can do 1 or 3 decimal places.

hhyyrylainen (changes requested):
There isn't that much that changes here, but I think a few small improvements to the code are warranted.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
Kasterisk: I rounded it to 2 decimal places since digestion speed is also rounded likewise.

Oliveriver:
Two things I noticed:

* I still think 'Resistant to engulfment' is a bit strong. It's definitely hard to fit a good qualifier in there without becoming too long. Maybe 'Resistant to basic engulfment'?
* The speed and efficiency increases don't do what they say they do. The lysosome supposedly increases them by 10% and 15% respectively (according to the tooltip) but it looks like they're actually constant additions of 0.05 and 4.5. Which is correct?

Kasterisk:
> I still think 'Resistant to engulfment' is a bit strong. It's definitely hard to fit a good qualifier in there without becoming too long. Maybe 'Resistant to basic engulfment'?

I would say that's probably better than the former.

> The speed and efficiency increases don't do what they say they do. The lysosome supposedly increases them by 10% and 15% respectively (according to the tooltip) but it looks like they're actually constant additions of 0.05 and 4.5. Which is correct?

The increase is 10% of the current speed and 15% of the current efficiency, which makes the 2 decimals. I now realize this is too vague with how it's communicated and is probably better replacing the percentages with those...

hhyyrylainen:
I think the problem here is 15 percent or 15 percentage points, as those are very different things.

Kasterisk:
Yeah, that's the essence. As I said, I think replacing percentages increase in the lysosome tooltip with explicit values (+0.05 and +4.5) can clear the confusion.

hhyyrylainen:
@Oliveriver could you review this PR (again)?

Oliveriver:
Looks good to me.

PR #3710, opened 2022-08-15
Auto-collapse diffs for .po files

Oliveriver:

I get quite annoyned having to scroll through a load of .po files with mostly file location changes before getting to the meat of the changes in most PRs. So I thought it would be sensible to auto-collapse .po file diffs, based on the method [here.](https://thoughtbot.com/blog/github-diff-supression).

Obviously it's not possible to check this works solely from this PR, so [here's a commit I made to see it in action](https://github.com/Revolutionary-Games/Thrive/commit/b6d51eca9773ee9834a6cf35e284ecfeb72d43b6).

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,6 @@
 * text=auto
-*.po text linguist-generated
-en.po text
+*.po text linguist-generated=true
+en.po text linguist-generated=false
 *.pot text
 *.rb text
 *.sln text
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,6 @@
 * text=auto
 *.po text linguist-generated
-*.pot text
+*.pot text linguist-generated
 *.rb text
 *.sln text
 *.csproj text
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 * text=auto
-*.po text
+*.po text linguist-generated
 *.pot text
 *.rb text
 *.sln text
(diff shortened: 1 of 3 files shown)


hhyyrylainen:
Was it an intentional choice that the .pot file was not also marked?

I suppose this is a change that overall helps more than it hurts as only in the translations update PRs I want to look at all the changed translations to see if they are fine. And sometimes in normal PRs I also check through a few of the changed .po files just to see if the author had used the tools correctly and there's no weird stuff going in them.

Oliveriver:
It was intentional, but thinking about it I guess there's no reason why not. I'll add .pot files too.

84634E1A607A:
Maybe we can exempt en.po?

hhyyrylainen:
I think it might be better for the .pot file to be shown by default and as @84634E1A607A just mentioned the en.po to not be hidden by default as those are the most interesting of the translation files to see without having to click that the changes to them are fine.

Oliveriver:
Alright, that should be working now.

hhyyrylainen (approved):
Based on the example commit this seems fine to me.

hhyyrylainen:
It may have worked if "en.po" was listed before in the file as I have a hunch that the rules are applied in order so if "en.po" matched a file the following rule for "*.po" might not trigger for it...

PR #3720, opened 2022-08-16
Created customized registry types concept for auto-evo settings saving

### LINKED ISSUE #3706: Auto-evo configuration is not correctly stored in save.

hhyyrylainen:

Basically I ended up setting up a system where a customized version of a registry type can be created that can be edited freely while making saving and loading work both with the customized and default instances.

This also now makes it so that we can update difficulty related numbers and old saves will benefit.

As far as I could test this doesn't seem to introduce any save compatibility problems

**Related Issues**

closes #3706
closes #3707

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/WorldGenerationSettings.cs
+++ b/src/general/WorldGenerationSettings.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using Godot;
 using Newtonsoft.Json;
 
 /// <summary>
@@ -9,7 +10,18 @@ public class WorldGenerationSettings
     [JsonConstructor]
     public WorldGenerationSettings(IDifficulty difficulty)
     {
-        Difficulty = difficulty;
+        if (difficulty is DifficultyPreset preset && preset.InternalName ==
+            SimulationParameters.Instance.GetDifficultyPreset("custom").InternalName)
+        {
+            GD.PrintErr(
+                $"Ignoring setting custom difficulty preset object to {nameof(WorldGenerationSettings)} " +
+                "(using normal instead). This should only happen when loading older saves");
+            Difficulty = SimulationParameters.Instance.GetDifficultyPreset("normal");
+        }
+        else
+        {
+            Difficulty = difficulty;
+        }
     }
 
     public WorldGenerationSettings()
--- a/src/general/WorldGenerationSettings.cs
+++ b/src/general/WorldGenerationSettings.cs
@@ -1,11 +1,16 @@
 ﻿using System;
+using Newtonsoft.Json;
 
 /// <summary>
 ///   Player configurable options for creating the game world
 /// </summary>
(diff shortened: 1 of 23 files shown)


84634E1A607A:
Looks fine to me.

84634E1A607A (approved):

src/general/WorldGenerationSettings.cs
```diff
@@ -1,11 +1,28 @@
...
-    private DifficultyPreset difficulty = null!;
+    [JsonConstructor]
+    public WorldGenerationSettings(IDifficulty difficulty)
+    {
+        if (difficulty is DifficultyPreset preset && preset.InternalName ==
+            SimulationParameters.Instance.GetDifficultyPreset("custom").InternalName)
+        {
+            GD.PrintErr(
+                $"Ignoring setting custom difficulty preset object to {nameof(WorldGenerationSettings)} " +
+                "(using normal instead). This should only happen when loading older saves");
```
Oliveriver, line 18: So loading an earlier save with custom settings would revert those settings to normal? I guess there's no easy way to avoid that?

src/general/WorldGenerationSettings.cs
```diff
@@ -1,11 +1,28 @@
...
+using Newtonsoft.Json;
 
 /// <summary>
 ///   Player configurable options for creating the game world
 /// </summary>
 public class WorldGenerationSettings
 {
-    private DifficultyPreset difficulty = null!;
+    [JsonConstructor]
+    public WorldGenerationSettings(IDifficulty difficulty)
```
Oliveriver, line 11: Presumably the planet settings will end up using this approach at some point too? Do we want to anticipate that now to avoid compatibility issues in future, or is that so far in the future that we can assume save breakage by that point?

Oliveriver (approved):
Sensible changes and testing seems fine. No real code comments, just a few questions about save compatibility.

src/general/WorldGenerationSettings.cs
hhyyrylainen: Yes. There's no easy way to avoid that. It would be possible to write a save upgrader which grabs the difficulty settings from the save and makes that into a new custom difficulty object. So while not very hard to do, it would take some extra effort to program and test.

src/general/WorldGenerationSettings.cs
hhyyrylainen: I suppose if the planet generator uses the same kind of approach where we have presets and the possibility to customize, it would need this same treatment. This needs to only be considered once the actual planet generator parameters are added, so there's nothing that needs to be done before that to prepare.

Issue #3273, opened 2022-04-26
Fix lag spikes from organelle duplication in very large cells


hhyyrylainen:
probably caused by the approach of starting at 0,0 coordinates and spiraling outwards until an empty spot is found. That probably starts running quite long when the cell is very large in radius.

I think we have two choices to fix this:
- Have the algorithm remember where it last placed an organelle and look for free spots nearby to cut down on the attempts it needs before finding a spot
- Run the search in a background task (TaskExecutor) and then place the duplicate organelle once the task is complete and it has found a suitable position

astral-grayson:
Could we just compute the growth path when saving an organism from the editor? Or is the order of growth not deterministic.

hhyyrylainen:
That would be possible as well but all species need to have the growth path computed and it needs to be made sure that cells that were partially grown when entering the editor can continue growing properly.

astral-grayson:
Ok looking into this a little more it looks like the best solution is probably to just keep track for each cell of where the last organelle was grown to and start the spiral from there. If we compute a growth path for every species on exiting the editor I worry the extra time before getting back to gameplay might be annoying if there are many large species around.

RacerBG:
Here is a bonus from me (the largest cells which I had ever created in Freebuild mode):

[Vohin.zip](https://github.com/Revolutionary-Games/Thrive/files/8626085/Vohin.zip)
[Tuxespion_Cedata.zip](https://github.com/Revolutionary-Games/Thrive/files/8626086/Tuxespion_Cedata.zip)

hhyyrylainen:
Turns out that the membrane generation for huge cells took up a lot of time, so with https://github.com/Revolutionary-Games/Thrive/pull/4251 done this should be less of an issue. But still the logic for finding a place to put a duplicated organelle does still get very expensive computationally in truly huge cells.

hhyyrylainen:
I'll write that there is now endosymbiosis soon in the game which will create extraneous organelles, which also need positions calculated for them. So the approach of pre-calculating a growth pattern is not possible.

hhyyrylainen:
I just checked the code again and there is already a pretty comprehensive algorithm for finding good places to put organelles in that starts from the split organelle's position so this isn't as bad as it could be. But I'll still try to introduce a cache of sorts to to try and see if it is possible to make organelle duplication no longer cause any lag spikes.

PR #3722, opened 2022-08-17
Initial Balancing Pass for Timed Reproduction

### LINKED ISSUE #3654: Balance the new time based reproduction (for early multicellular)
might need to consider things like scaling up the free compounds amount based on cell size or make subsequent organelles cheaper

Deus-Codes:

Tweaked values for passive ammonia/phosphate generation to result in a more balanced gameplay loop. I essentially have done three things...

1.) Halved the Base Phosphate/Ammonia Cost - Essentially, reduced in half the reproduction cost when you start as LUCA. This makes it so that cell lifespans don't get too long too quickly as you add parts.

2.) Lowered Passive Phosphate/Ammonia Generation From .3 to .2 - Slightly reduced how much ammonia and phosphate is taken up from the environment so that lifespans are slightly lengthened in lite of tweak 1, and making it just a bit harder to get by without any serious risk of starvation.

3.) Increased Max Phosphate/Ammonia Uptake from .85 to 1.25 - Increased how quickly a cell can absorb phosphate and ammonia clouds. This strongly incentivizes the search for these limiting nutrients, but doesn't make it so that the player can just jump to the editor extremely quickly.

The result I think is a solid experience which maintains finding food as the primary objective of the player while simultaneously encouraging the player to adapt so that they may better increase their ability to gather nutrients. 

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Nunez2196:
I played about 10 generations or so, and tested the saves and different things to make sure nothing broke somehow and everything is fine. Also I personally prefer these settings compared to master but, that is only my opinion.

Oliveriver:
I played 15 generations and timed them, listed in order below. I played pretty normally with a standard big engulfer.

* 36 s
* 27 s
* 52 s
* 1 m 50 s
* 1 m 21 s
* 57 s
* 44 s
* 1 m 5 s
* 1 m 5 s
* 55 s
* 47 s
* 52 s
* 56 s
* 1 m 2 s

I think it's clear that these generations are way too fast. As I said, there was nothing unnatural about how I played, and yet I was speeding through generations without getting the chance to really do anything in the environment besides engulf the nearest two or three cells. No incentive for exploring or doing anything differently, and 3/4 of a full game completed in under 15 minutes.

I also still have a problem with how early on the cell starts duplicating its organelles. As the starter cell, I had duplicated my cytoplasm within 8 seconds just by sitting there. That's way too early, even by the standards of the rest of the speeds above.

hhyyrylainen:
I think the times @Oliveriver posted should be about double. Especially the first cycle is super important, and if the player doesn't quickly play through the movement keys tutorial, they are going to miss some tutorial that is meant to happen before the first editor cycle.

Oliveriver:
> I think the times @Oliveriver posted should be about double. Especially the first cycle is super important, and if the player doesn't quickly play through the movement keys tutorial, they are going to miss some tutorial that is meant to happen before the first editor cycle.

I might even go as far as tripling them. If 20 generations is a whole game and a generation averages 3 minutes, that gives an hour for the whole game, which sounds ideal to me. Plus, that leaves plenty of time within generations for the day/night cycle to play out.

hhyyrylainen:
One thing I would be a bit cautious about is that the players seem to like making much bigger cells than intended by us, so we should at least do the bare minimum to ensure cells with 60-80 organelles are still kind of viable.

Deus-Codes:
In light of feedback, slightly increased base phosphate and ammonia costs. Also slightly reduced maximum reproduction compound uptake to hopefully result in a bit slower and steadier progression through the microbe stage. Playtesting felt fine although I think I might have had bad luck with compound spawning and auto-evo.

hhyyrylainen:
> If this still feels off, I might go through and increase the ammonia and phosphate costs of organelles. Currently I feel like there isn't much difference felt even across 5 generations.

That was mostly my intention that small and big cells basically reproduce in the same amount of time, as I think I remember some balancing talk about that that huge cells being super slow to play wouldn't be good. That's how I picked the initial free compound amounts and base cost related to the organelle costs.

Leinourdian:
The new commit doesn't seem to change that much. Some cells still reproduce quite fast ( 30 to 60 seconds) and big non-predatory cells take long to reproduce (multiple times longer than small cells). I experimented with different values and would personally suggest the following:

- Increase `MICROBE_REPRODUCTION_COST_BASE_AMMONIA` and `MICROBE_REPRODUCTION_COST_BASE_AMMONIA ` to 18. This makes passive growth take at least 3 minutes.
-  Lower `MICROBE_REPRODUCTION_MAX_COMPOUND_USE` to 0.6. This makes it so active hunting of ammonia and phosphates can at best triple your rate of growth. On average, an active cell grows about twice as fast as with passive growth alone. This does reduce how much ammonia and phosphates you get from predation, but that seems fine to me.
- Go to the `initialComposition` in `organelles.json` and divide the values of ammonia and phosphates by 10. This makes it so adding an organelle adds 1 to 2 seconds to passive growth instead of the current 10 to 20 seconds.
- Keep the rest the same.

These changes felt pretty good imho. They made both tiny and huge cells and sessile and active cells feel decent. Collecting clouds and eating cells still sped up growth noticeably but not too much. For most cells, play time varied between 1:30 minutes and 4 minutes depending on size and activity. The values can of course be tweaked if you want to change the average time, fastest time, scaling with size, etc. And this is just my suggestion, you can ignore it if you want.

---

By the way, passive growth still causes noticeable lag, especially with huge cells. I found out that it disappeared completely after I disabled organelle splitting, so that seems to be the culprit. I don't know any simple way to fix the lag though.

hhyyrylainen:
> By the way, passive growth still causes noticeable lag, especially with huge cells. I found out that it disappeared completely after I disabled organelle splitting, so that seems to be the culprit. I don't know any simple way to fix the lag though.

That's a known issue with the organelle splitting logic: https://github.com/Revolutionary-Games/Thrive/issues/3273

How big was your cell? I put that issue in our newly made microbe roadmap right at the end for polishing (as I kind of suspected that the expected size of cells will change with balancing and new features): https://forum.revolutionarygamesstudio.com/t/finishing-the-microbe-stage-roadmap/905/20

Leinourdian:
I had 52 organelles and experienced regular lag spikes every few seconds. I also experienced some, although fewer, lag spikes as a one hex cell in free build mode, so I think that AI cells constantly growing causes some of the lag.

hhyyrylainen:
I'll change that issue to a high priority bug, then.

Oliveriver:
I was going to play a full game with @Leinourdian's settings, but I think I've discovered two quite severe bugs that make testing with those values impossible.

* AI cells can reproduce extremely quickly. Like, within seconds. This leads to the screen filling up with about a million of them right away.
* If you die and respawn, you can also reproduce extremely quickly.

I think there's an inherent bug in passive generation that's not taking the base cost into account in these two places, so using tiny organelle composition values has made it apparent, but I'm not sure.

Leinourdian:
Huh, that's weird. I can't reproduce either of those bugs. The only AI cell I saw reproduce only did so after I followed it for about three minutes. Can you reproduce them consistently or are they random? Do you have more specific instructions for reproducing them, or a video of them happening or something?

hhyyrylainen:
I think I maybe followed the multicellular approach of not having the base cost being needed after reproducing once for an AI cell. Though I can't say for sure without checking the code if that's the case.
Respawning should definitely not have a bug like that. The player entity is entirely recreated so it cannot store state through a respawn.

Oliveriver:
I seem to be able to reproduce both bugs every time when starting a standard new game on this branch after changing  `initialComposition` for cytoplasm in `organelles.json` to 0.1 for both ammonia and phosphate.

https://www.youtube.com/watch?v=gnZyVhCk7pk

The first half shows the massive amount of cells appearing, all with negligible glucose since they quickly start dying. They would have full glucose if they'd spawned normally right? Low glucose must be the result of continual divisions until hitting the entity limit.

The second half shows me dying and being able to reproduce almost immediately. I didn't press any keys after pressing suicide.

Oliveriver:
Ahhh hold on, I think the AI reproduction bug is due to the starting compounds for newly spawned cells being too high to be compatible with the new values. If I duplicate myself when I have no phosphate and ammonia, the copy doesn't reproduce, but if I turn on infinite compounds first, the copy keeps multiplying until it runs out.

Deus-Codes:
I like the numbers you came up with in playtesting @Leinourdian. I have just a bit of apprehension of the fact that, since as a simple prokaryotic cell you don't really have much to do, the early game might be just a bit bloated. But I think your values serve as a good baseline to be toyed around with. I am also bumping into the issue mentioned above as I notice my cell's organelles almost immediately get duplicated when spawning in, so that's something we need to address before finalizing balancing passes.

Leinourdian:
> I am also bumping into the issue mentioned above as I notice my cell's organelles almost immediately get duplicated when spawning in, so that's something we need to address before finalizing balancing passes.

That actually happens simply because the organelles have so low composition values that they grow and duplicate in a matter of seconds after spawning, since organelle growth happens before paying the base cost. You could reduce the effect by increasing the values but then bigger cells would take longer to grow. I wonder if there's a way to spread out organelle growth without increasing the values. Maybe cells could flip-flop between growing organelles and paying the base cost or something. Or maybe ammonia and phosphates could go straight to a new total cost that's base cost + organelle compositions, and then organelles would be grown for free based on your total reproduction progress. Idk, just spitting some random ideas.

Then there are the bugs found by @Oliveriver. I managed to reproduce the one with reproducing AI cells but it was very inconsistent for some reason. It happened about once every few games and only to some cells. It happened immediately at the start of the game so no cell had time to pay the base cost. I couldn't reproduce the other one about respawning. Both seem to be clear issues based on the videos of course. It seems that some cells can bypass having to pay the base cost somehow.

Something probably needs to be done about these points/bugs if we go with my values.

hhyyrylainen:
Is this at least better than the situation before these changes? I'm starting to feel like we should get at least this merged.

Deus-Codes:
I'll come up with values which don't heavily mess with organelle composition so that we can atleast have fair balancing. @Leinourdian's values are what I'd say the ideal situation is, but it would probably take refactoring the organelle-splitting mechanic, so we can't wait on that.

I'm a bit busy with university picking up, but I'll make these changes within the next few days in accordance to the idea that the microbe stage should last roughly an hour to 2 hours.

buckly90 (approved):
Alright, I went ahead and played as an opportunistic chemosynthesizer (I would eat things if they happened to be in reach but wouldn't go out of my way) and here's the results;

Prokaryotic:
Gen 1: 1;26
Gen 2: 1;29
Gen 3: 1;21
Gen 4: 1;32
Gen 5; 1:50
Gen 6; 1:46
Gen 7: 2;00
Eukaryotic:
Gen 8: 2;58 (I died on this generation once)
Gen 9: 1;53
Gen 10; 1;48
Gen 11: 0;30 (I ended here by becoming multicellular)

As you can see, my average time stayed pretty roughly around a minute and a half. The total time it took for me to reach the next stage was roughly 20 minutes if my lazy math was correct. Of particular interest; the one time I died seemed to considerably lengthen the time it took to progress, something we might need to consider sometime for players that die often. (Perhaps a difficulty option allowing players to carry over a percentage of reproduction progress between lives?)

It is also worth considering that while I was playing fairly leisurely and in no hurry, I remained efficient and very knowledgeable of how to reach my goal, not growing any bigger than I needed to, and managing my speed and ATP balance to ensure there was no excess. This is something that many players, new or otherwise inexperienced will not be doing. We need to anticipate that the average player will have a rather inefficient cell, and that they may die fairly often.

If I recall correctly, we were hoping to aim for an average playtime of an hour for the microbe stage, but as the stage currently stands, I feel it best that we do not try to forcibly lengthen the playtime until the gameplay itself is more engaging.

Anyway yeah, I'm ok with these changes.

hhyyrylainen (approved):
Let's get at least this balancing merged now.

PR #3728, opened 2022-08-17
Patch map in exploring tool

### LINKED ISSUE #3691: Add a patch map view to the auto-evo exploring tool
To let the user see where species are and the world layout

84634E1A607A:
**Related Issues**

Closes #3691

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/DraggableScrollContainer.cs
+++ b/src/gui_common/DraggableScrollContainer.cs
@@ -82,7 +82,7 @@ public class DraggableScrollContainer : ScrollContainer
     {
         base._Ready();
 
-        ContentPath ??= GetChild(0).GetPath();
+        ContentPath ??= GetChild(2).GetPath();
 
         content = GetNode<Control>(ContentPath);
         tween = new Tween();
--- a/src/gui_common/DraggableScrollContainer.cs
+++ b/src/gui_common/DraggableScrollContainer.cs
@@ -82,6 +82,7 @@ public class DraggableScrollContainer : ScrollContainer
     {
         base._Ready();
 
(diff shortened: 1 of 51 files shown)


src/gui_common/DraggableScrollContainer.cs
```diff
@@ -82,7 +82,7 @@ public override void _Ready()
     {
         base._Ready();
 
-        ContentPath ??= GetChild(0).GetPath();
+        ContentPath ??= GetChild(2).GetPath();
```
Maxonovien, line 85: Why this change, is it a bugfix ?

hhyyrylainen: This seems very suspicious to me, and unless the documentation etc. is changed elsewhere this is an incorrect change as this makes the comments and code behaviour different. Also it's very unexpected that this would pick the *third* child as the content if the path is not set.

Kasterisk: The index definitely shouldn't be any other than 0. ScrollContainers the way it's designed should only have one Control child.

84634E1A607A: Child 0 is the HScroll bar; 1 VScroll

Kasterisk: Ah that makes sense, this should be documented then.

hhyyrylainen (approved):
I had a look through the code and tested in game and I don't really have much to say. I'll open a separate issue about being able to hide the "move to this patch" button in the patch details panel.
Just need to fix the translation conflicts and this can be merged.

Kasterisk (approved):

PR #3730, opened 2022-08-17
Testing Blurred Backgrounds PR

nickthenick:

This PR initially just replaces the Tidepool background with a blurred version. Make sure to use the "Warm Little Pond" start setting to spawn directly in the tidepool and test it right away.

image

The intent of blurring the background is to:

- Make the player feel smaller by having the background be out of focus
- Make the background look more like water by having it be blurred
- Increase the player's ability to discern objects on the midground (player's layer)

What I have left to do when I have time:

- [x] Fix the tiling issues
- [ ] Add blur (but progressively less for each layer) to the other background layers closer to the camera
- [ ] Add blur to the other biomes

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)


nickthenick:
Tiling should be fixed. 

Could someone test to see if they notice any tiling issues in the Tidepool biome. And also how the blurred background looks/feels? If all is good I can expand this PR to have blurred versions of all backgrounds and then we can test that.

Oliveriver:
Although the idea behind this makes sense, I think that because the game has spent so long with detailed backgrounds, the rest of the visuals don't really mesh with this all that well. Foreground elements (especially clouds and cell membranes) are all quite blurred and wishy-washy too, so together you kind of end up with a big mush. It's not bad, but I definitely prefer the current backgrounds.

nickthenick:
Okay I went through and reduced the blur by 4x on the base background layer. I also added blur, but to lesser degrees, to the layers between that and the player.

hhyyrylainen:

PR #3734, opened 2022-08-18
Give additional glucose that can be gained when digesting engulfed cells

Kasterisk:

Along with organelles initial composition, digesting cells now also give bonus glucose, the amount of which is the third of that cell's compounds storage capacity.

Further balancing with this if warranted is welcome.

**Related Issues**

To address players' gripe regarding digesting cells giving too few glucose to be sustainable.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


yowshi:
i don't think engulfment whole really needed a buff. it's eating the organelles outside a cell that needs a buff.

nickthenick:
> i don't think engulfment whole really needed a buff. it's eating the organelles outside a cell that needs a buff.

The main reason is just that glucose clouds will eventually disappear, and cells who evolve engulfing as an alternative to getting glucose need to be able to get enough glucose from engulfing so that they don't have to rely on finding clouds, or photosynthesis, or some other means.

yowshi:
> > i don't think engulfment whole really needed a buff. it's eating the organelles outside a cell that needs a buff.
> 
> The main reason is just that glucose clouds will eventually disappear, and cells who evolve engulfing as an alternative to getting glucose need to be able to get enough glucose from engulfing so that they don't have to rely on finding clouds, or photosynthesis, or some other means.

they already can, i play with glucose retention as low as possible and the glucose clouds vanish very fast. and i don't starve for lack of getting glucose from engulfing whole. 

While it is harder to play pure predator this seems largely to do with the rate the player uses glucose, IE this will do nothing to fix pure predators who kill for a living because the glucose you get from eating the organelles is even less.

84634E1A607A:
My thoughts: Organelle and membrane itself consists of various elements that can be digested. (See organelles.json `initialComposition`) I think engulfing a floating chunk (or a cell) should grant this part of elements to engulfer.

Kasterisk:
I've now reduced the amount a bit and to address organelles consumption, also added the bonus glucose to the dropped organelles as it makes sense to do the same for both @yowshi.

@84634E1A607A That's already the way it works in the game?

84634E1A607A:
> @84634E1A607A That's already the way it works in the game?

Is it? I think I made many vain attempts to engulf chunks / cells, yielding no phosphate at all.

Kasterisk:
https://github.com/Revolutionary-Games/Thrive/blob/f231a6970d73a492daf4230744ee724e3f8972f1/src/microbe_stage/Microbe.Interior.cs#L601-L609

That's the code. I can't say for sure, but you can do some debugging to see what the actual compounds contained in the consumed objects + additional compounds and check whether that matches with what you get.

84634E1A607A:
So it's not included on death because the only usage of this func is in EngulfedObj

Kasterisk:
I'm assuming you mean the corpse chunks? With my recent change, those also should give approximately the same amount of compounds (minus some lost to the environment) when you engulf cells whole. Marine snows have theirs defined in json, so if you want them to have more compounds, it needs a separate balancing pass.

simulation_parameters/Constants.cs
```diff
@@ -464,6 +464,8 @@ public static class Constants
     /// </summary>
     public const float ENZYME_DIGESTION_EFFICIENCY_MAXIMUM = 0.6f;
 
+    public const float ADDITIONAL_DIGESTIBLE_GLUCOSE_AMOUNT_DENOMINATOR = 4;
```
hhyyrylainen, line 467: I think this should be a multiplier instead of a divisor, once again because float division is *much* more expensive than multiplying by `0.25f`

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -804,6 +804,11 @@ private IEnumerable<FloatingChunk> OnKilled()
             }
         }
 
+        // Give bonus glucose at third the amount of our compounds bag capacity
+        compoundsToRelease.TryGetValue(glucose, out float existingGlucose);
+        compoundsToRelease[glucose] = existingGlucose + Compounds.Capacity /
+            Constants.ADDITIONAL_DIGESTIBLE_GLUCOSE_AMOUNT_DENOMINATOR;
```
hhyyrylainen, line 810: As this logic is a bit complicated and used in 2 places this should be extracted as a method.

hhyyrylainen (changes requested):
I won't comment on the gameplay impact but here's a few technical implementation details.

nickthenick:
Just tested this and it felt like a good change. Cells definitely should yield glucose, alongside ammonia and phosphate. One day we need to look at making cells yield glucose as part of their organelle composition cost (Nutrient Cost), but for now this is a good feature. It helped make a hunter build a lot more viable.

Any other people had a chance to playtest this?

buckly90:
Works great on the gameplay so I agree with Nick in the okay.

I do anticipate possibility of needing to lower the bonus glucose as part of the future balance pass though (Less glucose consumption so there's less need for it). Switching to the 0.x multiplier as hh suggested can make this easier to understand and tweak.

yowshi:
what dev build is this listed under so i can give it a play test?

hhyyrylainen:
The latest commit on this branch is `ca2026613c2d97db4bd60337d0968b3d6b50d797` which you can find by scrolling through the event feed on this page.

hhyyrylainen:
I'll mark this for the release after 0.5.10. If nothing else I can fix the code style problems in time for that, as long as I remember to do that.

hhyyrylainen (approved):
I fixed the slight cleanup things I wanted. Also solved merge conflicts with master and did a basic test that the game doesn't seem broken after these changes. If these changes are fine by @Kasterisk I'll merge this.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -804,6 +804,9 @@ private IEnumerable<FloatingChunk> OnKilled()
             }
         }
 
+        // Give bonus glucose at third the amount of our compounds bag capacity
```
Kasterisk, line 807: Maybe can be put inside the method or as a method documentation?

Kasterisk (commented):
Looks good to me, just one comment.

src/microbe_stage/Microbe.Contact.cs
hhyyrylainen: I decided to remove that comment as it was incorrect (as the multiplier was one *fourth* and not a third). So that comment got outdated even before the PR was finished, so I decided we are better off without a comment that is such a high risk for being incorrect.

Kasterisk: Shouldn't this be removed then (as this is still shown as an addition in the diff)?

hhyyrylainen: Oh right, I didn't realize until you pointed out. I've now removed this line.

Kasterisk:
Can't approve my own PR but this is good to go!

PR #3737, opened 2022-08-20
Auto evo exploration species flagging

Maxonovien:

This PR adds filtering for species in auto-evo exploration tree, by flagging nodes in red.

WIP, but main functionality is available

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/gui_common/dialogs/FilterWindow.cs
```diff
@@ -0,0 +1,365 @@
+using System;
+using System.Collections.Generic;
+using Godot;
+
+public class FilterWindow : CustomDialog
```
84634E1A607A, line 5: Use CustomConfirmationDialog instead

src/auto-evo/EvolutionaryTreeNode.cs
```diff
@@ -63,7 +79,7 @@ public override void _Ready()
 
     private void UpdateTexture()
     {
-        if (lastGeneration)
```
84634E1A607A, line 66: What's this? The red mark was intended to convey that it is the last node of a species. Why do you just modify its usage? And, I think *red* flag is not very friendly, you can request blue textures.

hhyyrylainen: I think the highlighting should also have a custom colour different from the extinct or normal.

84634E1A607A:
I did some fixes about the dialog.

src/gui_common/dialogs/FilterWindow.cs
```diff
@@ -0,0 +1,256 @@
...
+            new Godot.Collections.Array { filters.Count });
+
+        filterContainer.AddChild(filterButton);
+        filtersContainer.AddChild(filterContainer);
+        filters.Add(filter);
+
+        dirty = true;
+    }
+
+    private void OnNewFilterCategorySelected(int categoryIndex, int filterIndex)
```
84634E1A607A, line 79: I can't get across what you want to do here. Why such a mess? (This func, the previous one and several below)

You can add scenes such as "number slider filter .tscn" or something instead of messing up with casts and multiple throws.

84634E1A607A: image

src/auto-evo/EvolutionaryTreeNode.cs
Maxonovien: Indeed, the idea was to eventually revert back to it. But I had a question to tackle: what should we do for a species that is both flagged and last generation?

84634E1A607A: Flag is of higher priority of course.

Maxonovien: It was actually suggested by Nick to use inner and outer rings of the button to mark both.

84634E1A607A: I think it's not worth the trouble.

84634E1A607A: Second thought: we can set the modulate of the unwanted (filtered out) nodes half-transparent.

src/general/Filter.cs
```diff
@@ -0,0 +1,118 @@
...
+        public float GetNumberValue()
+        {
+            try
+            {
+                return (this as NumberFilterArgument)!.Value;
+            }
+            catch (NullReferenceException)
+            {
+                throw new InvalidOperationException("Can't get number value from a non-numeric filter argument!");
+            }
```
84634E1A607A, line 73:
```suggestion
            return (this as NumberFilterArgument)?.Value ??
                throw new InvalidOperationException("Can't get number value from a non-numeric filter argument!");
```

src/gui_common/dialogs/FilterWindow.cs
```diff
@@ -0,0 +1,81 @@
...
+    [Export]
+    public NodePath FiltersContainerPath = null!;
+
+    private readonly List<Filter> filters = new();
+    /// <summary>
+    ///   If redraw is needed.
+    /// </summary>
+    private bool dirty;
+
+    private PackedScene filterScene = GD.Load<PackedScene>("res://src/gui_common/dialogs/FilterLine.tscn");
```
84634E1A607A, line 16: I don't know if this is okay, but we mostly load the scenes in _ready

Maxonovien: Well, it is okay in the sense that it is functional. But I guess it falls under the guideline considerations on nodes, which advise against what I did: 

Avoid using a constructor to setup Godot resources, usually Node derived types should mostly do Godot Node related, constructor-like operations entirely in _Ready. [...]

hhyyrylainen: Yeah, this is exactly what the style guide is trying to say, scenes should be loaded in _Ready.

Maxonovien: Alright, thanks for the confirmation; I've changed it. I must say though I find this point not very clear, as I would probably not known what it meant without knowing what I was looking for. An example or two could be useful there I'd think.

src/general/Filter.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public void ClearItems()
+    {
+        filterItems.Clear();
+    }
+
+    public Func<T, bool> ComputeFilterFunction()
+    {
+        if (!filterItems.TryGetValue(filterCategory, out var filterItem))
```
84634E1A607A, line 34: If you are to throw a KeyNotFoundException, why don't you use `filterItems[filterCategory]` instead?

src/general/FilterArgument.cs
```diff
@@ -0,0 +1,59 @@
...
+    public class NumberFilterArgument : FilterArgument
+    {
+        public readonly float MinValue;
+        public readonly float MaxValue;
+
+        public float Value;
+
+        public NumberFilterArgument(float minValue, float maxValue, float defaultValue)
+        {
+            if (defaultValue < minValue || defaultValue > maxValue)
```
84634E1A607A, line 39: I think use `Mathf.Clamp()` is better.

src/general/FilterArgument.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    public class MultipleChoiceFilterArgument : FilterArgument
+    {
+        public readonly List<string> Options;
+        public string Value;
+
+        public MultipleChoiceFilterArgument(List<string> options)
+        {
+            Options = options;
+            Value = Options.Count > 0 ? Options[0] : "--";
```
84634E1A607A, line 56: `--` is not a good default value (I think)

src/general/FilterArgument.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+}
```
84634E1A607A, line 59: New line

84634E1A607A:
I really think that this is a little bit too complex a filter func.

src/general/FilterArgument.cs
hhyyrylainen: Should at least be translatable to other languages.

Maxonovien: Yes, I intend to change that a bit afterwards. I'm currently working on multiple filters at a time (and fixing snapshots with them)

src/general/FilterArgument.cs
Maxonovien: Isn't Mathf.Clamp() used to simply alter the value? Here, if the default value is not set right it would probably be a user mistake. Which is why I use these checks to throw an exception.

hhyyrylainen: I agree that not using clamp here is the right way to check and throw the argument exception.

Maxonovien: Alright, thanks for the confirmation. I added a comment to explain the logic since there's never too many of them (almost).

src/general/Filter.cs
```diff
@@ -0,0 +1,139 @@
...
+
+    string IFilter.FilterCategory
+    {
+        get => filterCategory;
+        set => filterCategory = value;
+    }
+
+    public IEnumerable<string> FilterItemsNames => filterItems.Keys;
+
+    Dictionary<string, IFilter.IFilterItem> IFilter.FilterItems => filterItems;
```
84634E1A607A, line 20: Unify interface implementation.

src/general/IFilter.cs
```diff
@@ -0,0 +1,30 @@
+using System.Collections.Generic;
+
+public interface IFilter
```
84634E1A607A, line 3: Why interface instead of abstract class?

src/general/Filter.cs
```diff
@@ -0,0 +1,139 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+// TODO CONSIDER GENERIC
+public sealed class Filter<T> : IFilter
+{
+    private string filterCategory = "NONE";
```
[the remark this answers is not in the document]
Maxonovien: I actually consider having no default category and directly set to the first of the list... just haven't decided their fate yet.

src/general/IFilter.cs
Maxonovien: Good question. I basically wanted to hide the the type of the Filter. I spontaneously went for interface as this is what has been used for e.g. organelles, but maybe it could be better with an abstract class indeed? Haven't given it much tought.

hhyyrylainen: Out of curiosity, what "organelle interface" do you mean? Only thing I can think of is `IPositionedOrganelle` where the interface is mandatory as a Godot.Spatial derived type is needed in one place and in another we want a pure data class that has nothing to do with Godot.

Maxonovien: It was actually `IOrganelleComponent`, along with `IOrganelleComponentFactory` (although on my part I currently have the factory defined within the interface, but I'm open to create a new file for readability)

84634E1A607A (changes requested):
1. image
Press Apply and get:
```
System.Collections.Generic.KeyNotFoundException: The given key 'NONE' was not present in the dictionary.
  at System.Collections.Generic.Dictionary`2[TKey,TValue].get_Item (TKey key) [0x0001e] in <c84d94161bca414384015e66544862cf>:0 
  at Filter`1[T].ComputeFilterFunction () [0x00001] in C:\Users\Ajax\Thrive\src\general\Filter.cs:34 
  at AutoEvoExploringTool+<>c.<FlagTreeNodes>b__117_0 (Filter`1[T] f) [0x00000] in C:\Users\Ajax\Thrive\src\auto-evo\AutoEvoExploringTool.cs:698 
  at System.Linq.Enumerable+SelectListIterator`2[TSource,TResult].GetCount (System.Boolean onlyIfCheap) [0x00025] in <8dc7f7cf628440cbb575eaa126cc9c6d>:0 
  at System.Linq.Enumerable.Count[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x00029] in <8dc7f7cf628440cbb575eaa126cc9c6d>:0 
  at AutoEvoExploringTool.FlagTreeNodes () [0x0003e] in C:\Users\Ajax\Thrive\src\auto-evo\AutoEvoExploringTool.cs:700 
```

2. Conditions should be connected by either `AND` or `OR` (instead of AND only)

Maxonovien:
For 1) I've been aware of this issue for some time, but I'll probably end up deleting this default category anyways, so this should fix it.

As of 2) it only is quality of life improvement (since you can do `OR` by applying two filters at different times), so I won't do it in this PR, which has some content already, and has been dragging on for too long. I will prioritize updates of the auto-evo algorithm, especially for the day/night cycle.

But what I would envision for this is to simply assign one of a handful of colors to each filters. Filters with the same color would need to be all respected for the color to apply; that would be conjunction while filters with different colors would act as disjunctions.

84634E1A607A:
If you don't have time i may take over (but i may rewrite it 50%)

hhyyrylainen:

PR #3742, opened 2022-08-21
Add a Day and Night Cycle

GameDungeon:
This PR adds a day and night cycle to thrive. This new PR exists because the old branch broke. #3702 

**Progress Checklist**

Essentials:

- [x] Replace "at noon" string literal with translated format string
- [x] Fix Sea floor ambient light level is zero even when the maximum is non-zero
- [x] Average button in the editor needs to use the actual average
- [x] Reset light level in the editor when entering the editor or when changing patch
- [x] Change setting camera light level by BiomeType to setting camera by whether maximum sunlight in a patch is non-zero
- [x] Replace repeated use of SimulationParameters.Instance.GetCompound("sunlight") with member variable in several places
- [x] Use average sunlight in auto-evo calculations (two comments in SimulationCache)
- [x] Cache compound use for storage score (TODO comment in EnvironmentalFoodSource)
- [x] Make day/night/average buttons persist their visible toggle state like the structure/appearance/behaviour buttons
- [x] Hide day/night/average buttons if day-night cycle is disabled
- [x] Fix Process Panel Glitches

Nice to have:

- [X] Make day length player configurable in new game settings 
- [ ] Cheat to set the current time of day
- [x] Interface for current light level to remove the need to decide between ambient and maximum in each scenario

Final Checks:

- [x] Resolve code comments on PR
- [ ] Balancing


hhyyrylainen:
Using one of my canned replies:

GameDungeon:
While I have no clue what happened on the last one, it was trying to overwrite every change made to thrive since the last one was opened. I couldn't figure out how to fix it, so this seemed the best way.

hhyyrylainen:
You could have probably done the following:
- Note the hashes of each commit that should stay on the branch
- Delete your local copy of the branch
- Switch to master and update
- Create a new local branch with the old name
- Cherry-pick all of those hashes you wrote down
- Force push the branch, that should clean out everything

hhyyrylainen:
Your local copy likely had the github branch merged into it meaning that pushing it wouldn't have the similar overwriting effect. I'm sure that had you had a clean local branch (that was not merged from the upstream), you could have force pushed to in effect nullify the remote branch.

GameDungeon:
Before merging this needs a working save system, but other than that the PR should be complete and is ready for testing & review.

GameDungeon:
This now supports saving. 

Now all this PR is missing is a toggle on the main menu and autoevo support.

buckly90 (approved):
The current state of balance feels good. I was able to comfortably survive as a pure plant without any complications aside from predators. Gives more strategic emphasis on walls as well which is very welcome. 

Silica grants much less osmoregulation cost but leaves you vulnerable to engulfment so you can survive nights much more readily but at greater risk of being eaten. Chitin and cellulose are decent for osmoregulation, and exceptional for protecting yourself from harm. And of course the single and double membranes remain great options for being more predatory, but the larger osmoregulation makes it harder to rely solely on photosynthesis.

The lighting effects look fantastic, and makes the world feel more alive and dynamic which I like.

There are a few quirks in this that need ironed out eventually, but are probably not urgent such as autoevo not taking the cycle into account, and the time of day being recorded as a change in light in the conditions screen.

All in all I give this a passing review, and feel that it is of sufficient quality for merging from a gameplay standpoint.

src/general/DayNightCycle.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        float halfPercentage = LightCycleConfig.DaytimePercentage / 2;
+
+        // This converts the percentage in DaytimePercentage to the power of two needed for DayLightPercentage
+        daytimeMultiplier = (float)Math.Pow(2, 1 / halfPercentage);
+
+        if (IsEnabled)
+        {
+            AverageSunlight = EvaluateAverageSunlight(0.5f + halfPercentage)
+                - EvaluateAverageSunlight(0.5f - halfPercentage);
```
Maxonovien, line 39: These lines confuse me a lot. What I read here is that the average sunlight is a difference between average sunlight (`EvaluateAverageSunlight`), which seems off. But I think that `EvaluateAverageSunlight` actually computes the *accumulated* sunlight throughout the day (that is, the antiderivative of the sunlight value function).

Also, the fact that you (rightfully) dropped the nighttime parts for simpler computations should probably warrant a small comment stating that.

src/general/DayNightCycle.cs
```diff
@@ -0,0 +1,91 @@
...
+    private float daytimeMultiplier;
+
+    public DayNightCycle(bool isEnabled)
+    {
+        IsEnabled = isEnabled;
+
+        LightCycleConfig = SimulationParameters.Instance.GetDayNightCycleConfiguration();
+        Time = LightCycleConfig.HoursPerDay / 2;
+
+        float halfPercentage = LightCycleConfig.DaytimePercentage / 2;
```
Maxonovien, line 31: This `halfPercentage` name does not tell at all that were are dealing with the (scaled) duration of daytime. 

As a note, I also was confused by the name `DaytimePercentage` as well. I assumed it would represent the max value of sunlight at day time (which is a percentage). Perhaps the term "percentage" isn't great as it refers to a measure unit and not a "dimension"?

src/general/DayNightCycle.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// </remarks>
+    [JsonIgnore]
+    private float daytimeMultiplier;
+
+    public DayNightCycle(bool isEnabled)
+    {
+        IsEnabled = isEnabled;
+
+        LightCycleConfig = SimulationParameters.Instance.GetDayNightCycleConfiguration();
+        Time = LightCycleConfig.HoursPerDay / 2;
```
Maxonovien, line 29: So if I get it well, you just instantiate the cycle at noon right?

src/general/DayNightCycle.cs
```diff
@@ -0,0 +1,91 @@
...
+    public float AverageSunlight { get; private set; }
+
+    /// <summary>
+    ///   The current time in hours
+    /// </summary>
+    [JsonProperty]
+    public float Time { get; private set; }
+
+    [JsonIgnore]
+    public float PercentOfDayElapsed => Time / LightCycleConfig.HoursPerDay;
```
Maxonovien, line 57: If I'm not mistaken, `Time` is only used within this file. As you only use it to define `PercentOfDayElapsed`, you could get rid of it. In particular, you would avoid multiplication/division by `LightCycleConfig.HoursPerDay`.

src/general/DayNightCycle.cs
```diff
@@ -0,0 +1,91 @@
...
+    ///   The percentage of daylight you should get.
+    ///   light = max(-(PercentOfDayElapsed - 0.5)^2 * daytimeMultiplier + 1, 0)
+    ///   desmos: https://www.desmos.com/calculator/vrrk1bkac2
+    /// </summary>
+    /// <remarks>
+    ///   If this equation is changed EvaluateAverageSunlight needs to be updated.
+    /// </remarks>
+    [JsonIgnore]
+    public float DayLightPercentage =>
+        Math.Max(-(float)Math.Pow(PercentOfDayElapsed - 0.5, 2) * daytimeMultiplier + 1, 0);
```
Maxonovien, line 69: `1 - Mathf.Pow(PercentOfDayElapsed - 0.5f, 2) * [...]` would make it easier to read.

src/general/DayNightCycle.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        IsEnabled = isEnabled;
+
+        LightCycleConfig = SimulationParameters.Instance.GetDayNightCycleConfiguration();
+        Time = LightCycleConfig.HoursPerDay / 2;
+
+        float halfPercentage = LightCycleConfig.DaytimePercentage / 2;
+
+        // This converts the percentage in DaytimePercentage to the power of two needed for DayLightPercentage
+        daytimeMultiplier = (float)Math.Pow(2, 1 / halfPercentage);
```
Maxonovien, line 34: `Mathf.pow`. Also, a source motivating this formula would be neat here.

src/general/DayNightCycle.cs
```diff
@@ -0,0 +1,91 @@
...
+    [JsonIgnore]
+    public float DayLightPercentage =>
+        Math.Max(-(float)Math.Pow(PercentOfDayElapsed - 0.5, 2) * daytimeMultiplier + 1, 0);
+
+    public void Process(float delta)
+    {
+        if (IsEnabled)
+        {
+            Time = (Time + (1 / LightCycleConfig.RealTimePerDay) * LightCycleConfig.HoursPerDay * delta)
+                % LightCycleConfig.HoursPerDay;
```
Maxonovien, line 76:
```suggestion
            Time = (Time + delta  * LightCycleConfig.HoursPerDay / LightCycleConfig.RealTimePerDay)
                % LightCycleConfig.HoursPerDay;
```

But also see my suggestion for not using absolute time.

Maxonovien (changes requested):
Only reviewed the logic behind Day/Night cycle while trying to understand it for auto-evo. I didn't review its use elsewhere.

src/general/DayNightCycle.cs
GameDungeon: DaytimePercentage is the Percentage of the day that has daylight.

src/general/DayNightCycle.cs
GameDungeon: Correct. Starting at night felt weird when starting in the warm pond. Should I add a comment here?

src/general/DayNightCycle.cs
GameDungeon: I have it public as it seems like something that would be useful outside the file. For example for an indicator if we ever add one.

src/general/DayNightCycle.cs
GameDungeon: I struggled a lot with naming that function. Suggestions welcome.

src/general/DayNightCycle.cs
hhyyrylainen: To me DaytimePercentage would be how much current light is of the daytime value. I'd use a name like `DaytimeDurationPercentage` or `DaylightDurationPercentage` or something like that.

src/general/DayNightCycle.cs
Maxonovien: Yes please, I think it will help understand it when reading it for the first time.

src/general/DayNightCycle.cs
Maxonovien: Sure, but then I would save the value in `PercentOfDayElapsed`, especially since you're basically reconverting Time every time you change it.

hhyyrylainen: I also think that we should make the fraction or percent elapsed the main time tracking variable, with the variable controlling how fast that changes based on elapsed time what's tweaked to get the speed right.

src/general/DayNightCycle.cs
Maxonovien: `CumulativeSunlightValue`? Or, perhaps more visually `TotalSunlightSinceMidnight`?

src/general/DayNightCycle.cs
Maxonovien: I second this suggestion. I finally understood what DaytimePercentage meant, but as I stated above the name is indeed misleading.

`halfPercentage` could also still use a clearer name even though you only use it locally.

Maxonovien:
Added handling by auto-evo. I didn't run gameplay tests though, and didn't fine-tune the added constants.

GameDungeon:
Have you tested if things break with the day-night cycle toggled on / off?

Maxonovien:
For some reason, even before my commits I am not able to confirm after selecting "enable option"

GameDungeon:
I can't replicate that. It works just fine on my computer.

Maxonovien:
I don't know what I did, but it indeed seems to work fine now from the commit. I was able to enter & exit editor, and auto-evo exploring tool reported some plants (cytoplasms + thylakoids), with a good amount of storage so it seems (at least roughly) working.

Still some review (both code & logic) would be welcome.

src/general/DayNightCycle.cs
GameDungeon: The main reason time exists is for stuff like clocks, though I guess you are right about it being better to use PercentOfDayElapsed.

src/general/DayNightCycle.cs
GameDungeon: Being honest, I don't actually know why this works, I just played with the numbers until I saw a pattern.

Oliveriver:
Tested a few generations and discovered several bugs:

* Cloud brightness in surface patches at night is now way too high thanks to the per-patch brightness multiplier not accounting for time of day. I'm not sure what the simplest way to fix this is.

* Tooltips, the patch map and the compound balance in the editor use only the light level when you entered the editor. The patch map in particular definitely needs to be fixed to show constant lux, and see my suggestion below for the others.

* Similarly, there are inconsistencies in the process panel. The light level displayed didn't match the actual patch light level until I died, when it did update but froze again.

And a few suggestions for quality of life:

* The heading for experimental settings needs more spacing above it, since it still blends into the option list. I also think we should put the later stage prototypes option under this heading, but that's up for debate.

* I think the period of 0% light is too long, especially when 100% light only occurs for a few moments.

* It would be helpful to have an option in the debug panel to manually adjust the current light level, but that's very much a luxury and not necessary for this PR.

* I think I said this before: the editor needs to convey information about the day/night cycle to the player. My ideal setup would be two-fold:

  * A slider somewhere in the editor UI that adjusts the "time of day" shown in the editor. This would update: a) the background, b) values shown on tooltips, c) compound and ATP balances. There could also be buttons to set to maximum/minimum light level and reset the light level to what it was when you entered the editor. 

  * A display for average compound production over a day, which I think you've started but haven't finished.

  Obviously though that's quite a bit of work. There are simpler options, such as just having a button to switch between max night and max day with the requisite values updating. For me that's the minimum viable amount of information to display.

GameDungeon:
Currently, each patch has a default level for each compound and one that can change. I think I will need to add a third live one for things that change during real-time gameplay, but don't affect things outside real-time.

The curve for DayLightPercentage and the day and night lengths are set at accurate values. I feel most of the issues with it are balancing issues (No balancing has been done as of right now)

I've noticed the issues with the process panel, and honestly have no clue why they happen. If someone else could figure it out that would be great. 

Most of the UI stuff should be added in a later PR (It's the reason that this is in the Experimental Tab as of right now)

Oliveriver:
I suppose the editor UI changes can be left for a separate PR, but I'm conscious of the approaching 0.5.10 deadline. I've just worked on a proof of concept, although so far it doesn't update any displayed values, only the background.

hhyyrylainen:
> The heading for experimental settings needs more spacing above it, since it still blends into the option list. I also think we should put the later stage prototypes option under this heading, but that's up for debate.

I think the prototypes are something that most people should have enabled at all times. That's because it's just one extra button to appears and when clicked shows an extra warning anyway. So the prototypes aren't a potentially game enjoyability breaking feature.

> Currently, each patch has a default level for each compound and one that can change. I think I will need to add a third live one for things that change during real-time gameplay, but don't affect things outside real-time.

Why is there a need for third variable? I've not looked at the code but my original suggestion way back was for the day/night system to adjust the compound values in the patch conditions that's meant to be modified.
My suggestion to fix the editor is to make it so that the stage applies the normal light values when swapping to the editor scene and in the on return from editor callback it reapplies the changing light values.

> I suppose the editor UI changes can be left for a separate PR, but I'm conscious of the approaching 0.5.10 deadline. I've just worked on a proof of concept, although so far it doesn't update any displayed values, only the background.

I'm not very keen on this design because if you have a manual slider then what may happen is:
- Player makes a change to organelles
- The player now has to mouse drag all over the light slider to see if their changes are good
- If the changes are not good the player needs to undo the change and go back to step 1

Whereas a system that always shows the estimate if the player can survive the night cuts down the back and forth where the player needs to go and mouse drag a slider around out of the process of editing the species.

GameDungeon:
> Why is there a need for third variable? I've not looked at the code but my original suggestion way back was for the day/night system to adjust the compound values in the patch conditions that's meant to be modified.
> My suggestion to fix the editor is to make it so that the stage applies the normal light values when swapping to the editor scene and in the on return from editor callback it reapplies the changing light values.

That is doable. The issue with it is it is currently impossible to make any permanent changes to the light level as it will get overwritten instantly. It would have to be a change to the biome template itself.

Thrive.csproj
```diff
@@ -560,22 +563,19 @@
...
     </PackageReference>
-    <PackageReference Include="Nito.Collections.Deque">
+    <PackageReference Include="Nito.Collections.Deque" Version="1.1.1">
       <Version>1.1.1</Version>
     </PackageReference>
-    <PackageReference Include="StyleCop.Analyzers">
-      <Version>1.2.0-beta.435</Version>
-    </PackageReference>
-    <PackageReference Include="Lib.Harmony">
+    <PackageReference Include="Lib.Harmony" Version="2.2.2">
```
hhyyrylainen, line 578: These changes should be reverted.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -26,6 +27,8 @@ public SimulationCache(WorldGenerationSettings worldSettings)
             this.worldSettings = worldSettings;
         }
 
+        public bool DayNightCycleEnabled => !worldSettings.DayNightEnabled;
```
hhyyrylainen, line 30: Is the bool the wrong way around here?

src/auto-evo/simulation/food_source/RandomFoodSource.cs
```diff
@@ -0,0 +1,17 @@
+namespace AutoEvo
+{
+    using System;
+
+    /// <summary>
+    ///   A class for random-encounter based food sources.
+    /// </summary>
+    public abstract class RandomFoodSource : FoodSource
```
hhyyrylainen, line 8: Why was this renamed? I think the name RandomFoodSource is not any better than CompoundFoodSource for finding compounds.

Maxonovien: It was not renamed. CompoundFoodSource still exists. Here, randomFoodSource includes chunks, other species...

 It is just that I separated Environmental Food sources (whose gathering is not chance-based) from the rest, whose gathering is encounter-based.

src/auto-evo/simulation/SimulationCache.cs
Maxonovien: Probably, I hesitated between having enabled and disabled, so it must be a leftover... i'll correct that when I get the chance to.

src/auto-evo/simulation/food_source/RandomFoodSource.cs
hhyyrylainen: Ah, I reread the code and realized that. Still I think random is super non-descriptive name, I'd assume it would actually use just random noise as the results... Something with "encounter" in the name would be more clear. Or at least not as easily confused with random numbers.

Thrive.csproj
GameDungeon: I've tried a few times, but I revert them thrive won't load for me until it auto converts it back to this. Don't know how to fix that.

hhyyrylainen: Update everything starting from dotnet and msbuild, and make sure that Godot is configured to use the right tools.
Last time this kind of change attempted to happen it entirely broke the game for me for running through Godot. Maybe 4.0 can deal with this by using the dotnet 6 runtime but for now I don't think this change should be allowed.

Oliveriver:
I've updated this branch with the extra features I considered necessary:

* The patch map now displays the maximum light level and current light level in a patch (see below).
* Patch light levels now change in sync to represent the changing day across the whole planet. Obviously we might want to change that if geographical distances between patches are large enough for different time zones, but I for one vastly prefer the new approach to having time only start ticking once you enter a patch.
* The editor in light patches now displays day/night/average light level toggle buttons which update compound balances, tooltips and the camera to match the patch maximum/minimum/average light level respectively (see below).
* Compound cloud brightness adjusts dynamically as the day progresses. Before, it would stay at the light patches' elevated brightness all through the night which didn't look good.

There are a few bugs and the code is a mess, but I don't think it will take long to clean them up. In particular, the average button doesn't actually get the average light level yet. I need to find a way to wire that to the editor properly. I'll use whatever time I find for Thrive in the coming weeks to finish this all up.

84634E1A607A:
Excellent!

buckly90:
Pause issue seems to be resolved, no additional issues found.

You've done a wonderful job!

Oliveriver:
I ended up spending way longer on this than I wanted. I think I've dealt with most of the issues. Here's what's remaining:

* I haven't tested this with the Multicellular Stage (both early and late) at all, so someone needs to do that.
* When loading a save made in the editor, the average button doesn't use the proper average. I think the average sunlight in the editor needs to saved then.
* There _might_ be some performance issues caused by updating patch light levels every process cycle. I can't really tell. If so, it can be fixed by only updating every second or so. No need to have it every process cycle.
* There's still a TODO comment by @Maxonovien in SimulationCache. I had no idea how to resolve it.
* There are multiple errors about line charts in the editor missing data. I suspect it's due to the sunlight but I didn't have time to figure it out.

I _really_ won't have any more time to devote to this before the release now. But considering the amount of work I and others have put into this, and the expectation from fans, I want to emphasise the priority of this feature to everyone else working on the release. Please make sure it gets in.

hhyyrylainen:
I hope someone else can work on this. I haven't really read the code of this PR recently, but I have a slight feeling I could not help myself doing a major refactoring to simplify the interfaces between the rest of the game and the timed values.

Oliveriver:
I wouldn’t be opposed to that. It was a massive pain to work with the existing interfaces and I tried introducing a new one, but it doesn’t fee particularly clean.

GameDungeon:
First, great job Oliver, I'm surprised at how much got done. I can do testing, but I'm not sure I can do much else here. I'm still just not familiar enough with thrives internals to make a lot of these changes.

hhyyrylainen:
But you started this PR? Going through all places that need compound data from a patch and making it go through an interface that can provide the static, dynamic or average data seems much simpler than just hoping you find all the right places in the code and update them to use the right values.

Oliveriver:
Just realised I forgot to revert some rather extreme changes to bio_processes.json that I needed for testing something. Those should be reverted, and then this PR is ready for testing.

Oliveriver:
Yeah, I thought about balance a bit. The great thing is that now with the average sunlight button in the editor, you can easily see average glucose production over a day, which is probably the best indicator for survivability.

I would be in favour of increasing the percentage of the day that has sunlight, which is DaytimePercentage in day_night_cycle.json. As well as buffing plants a bit, it would reduce the time the player spends in pitch black conditions, which I found quite a bit less engaging than times of changing light.

A quick test shows me that setting DaytimePercentage somewhere around 0.875 is enough for a single thylakoid to break even on average. That's probably too much daylight time for fun gameplay though, so photosynthesis probably needs a process buff too.

Oliveriver:
Also, just discovered the light level in the process panel can still freeze, and seemingly does so when your cell is ready to reproduce. The actual cell process definitely still responds to the correct light level though.

hhyyrylainen:
My comment from discord regarding why I didn't start reviewing this yet:

> average sunlight not saved when saving in the editor

Sounds like a save format problem to me
Which can require rethinking the data format entirely
So that's why I considered it blocking my review as if the data classes or the way the adjusted numbers are passed around is refactored, that'll touch a ton of files

---

Also if not fixed the following issues need to be opened when merging:
- process panel freezing? Does it freeze entirely? That sounds like a pretty major regression.
- TODO comment regarding caching
- Editor charts no longer always working, another big regression
- Unnecessary performance impact of applying light changes constantly (If I understood right, there's *no* reason to apply the light level 360 times per second)

hhyyrylainen:
Re: the average values. I said it's important to fix as doing it properly *may* require rethinking the data model a bit, which would impact saving.


> It doesn't freeze entirely, just DayNightCycle-related stuff. It seems to do this randomly as it works most of the time. Really annoying bug to pin down.

I'd hazard a guess that when the process is not running, it doesn't update the environmental values stored in the process run statistics as it doesn't get so far along.

Okay so if it only affects the chloroplast, I guess we can consider it a non-blocking issue, but I swear my sanity will drop if anyone at all notices the bug and I'm the one to see the complaint without replies already.

> While it's unnecessary, I really doubt there is any real performance impact from it. It's just changing a single value that gets grabbed from other functions in every frame anyway. Updating that value changes nothing on its own, functions will just get that value when they are run. It's also not like redoing light calculations, everything graphics-wise is in a shader, which has to run every frame by its very nature.

Setting new shader parameters is probably not entirely free in terms of performance.

Thrive.csproj
```diff
@@ -580,7 +585,7 @@
     <Compile Include="test\MetaballTest.cs" />
   </ItemGroup>
   <ItemGroup>
-    <Folder Include="shaders\" />
+  <Folder Include="shaders\" />
```
hhyyrylainen, line 588: This change is wrong

Thrive.csproj
```diff
@@ -598,9 +603,6 @@
     <PackageReference Include="Nito.Collections.Deque">
       <Version>1.1.1</Version>
     </PackageReference>
-    <PackageReference Include="StyleCop.Analyzers">
-      <Version>1.2.0-beta.435</Version>
-    </PackageReference>
```
hhyyrylainen, line 603: And this is wrong as well

src/saving/serializers/RegistryTypeConverter.cs
```diff
@@ -72,6 +72,9 @@ public RegistryTypeConverter(ISaveContext context) : base(context)
         if (objectType == typeof(Enzyme))
             return Context.Simulation.GetEnzyme(name);
 
+        if (objectType == typeof(DayNightConfiguration))
+            return Context.Simulation.GetDayNightCycleConfiguration();
```
hhyyrylainen, line 76: This code should be exactly following how `IAutoEvoConfiguration` is done on lines 81-82.

src/microbe_stage/editor/PatchMapEditorComponent.cs
```diff
@@ -132,6 +132,10 @@ public override void OnValidAction()
     {
     }
 
+    public override void OnLightLevelChanged(float lightLevel)
```
hhyyrylainen, line 135: I've got a feeling that there are more editor components that won't care about the light level so I think the base class should have a virtual empty method for this instead of needing to modify all of the components when minority of the components (I think) will even react to this.

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -262,9 +267,15 @@ private void UpdatePatchDetails()
...
             unitFormat.FormatSafe(SelectedPatch.Biome.Compounds[temperature].Ambient, temperature.Unit);
         pressure.Text = unitFormat.FormatSafe(20, "bar");
-        light.Text =
-            unitFormat.FormatSafe(
-                percentageFormat.FormatSafe(GetCompoundAmount(SelectedPatch, sunlightCompound.InternalName)), "lx");
+
+        var maxLightLevel = GetCompoundAmount(SelectedPatch, sunlightCompound.InternalName, CompoundAmountType.Maximum);
+        light.Text = unitFormat.FormatSafe(percentageFormat.FormatSafe(Math.Round(
+            GetCompoundAmount(SelectedPatch, sunlightCompound.InternalName))), "lx");
+        lightMax.Text = string.Format(
```
hhyyrylainen, line 274: Safe format should be used to guard against crashing with bad translations

shaders/MicrobeBackground.shader
```diff
@@ -15,6 +18,47 @@ const vec2 speed3 = vec2(1050.0f);
...
 varying vec2 UV4;
 
+vec3 LightInfluence(float amount)
+{
+    vec3 influence = vec3(amount, amount, amount);
+    int time = 0;
+
+    if(amount > 0.5f)
+        time = 2;
+    else if(amount > 0.25f)
```
hhyyrylainen, line 28: Incorrect braces style used here.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -26,6 +27,8 @@ public SimulationCache(WorldGenerationSettings worldSettings)
             this.worldSettings = worldSettings;
         }
 
+        public bool DayNightCycleEnabled => worldSettings.DayNightCycleEnabled;
```
hhyyrylainen, line 30: Why is this property here? This is not even cached here so this looks pretty dirty to me in terms of logically accessing the world settings through a cache that doesn't even cache the value. What I mean is that this feels like the wrong place to put this. Why aren't things just directly using the world settings and getting the bool from there?

src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
```diff
@@ -15,14 +15,16 @@ public EnvironmentalFoodSource(Patch patch, Compound compound, float foodCapacit
 
             this.patch = patch;
             this.compound = compound;
-            totalEnvironmentalEnergySource = patch.Biome.Compounds[this.compound].Ambient * foodCapacityMultiplier;
+            totalEnvironmentalEnergySource = compound.InternalName == "sunlight" ?
```
hhyyrylainen, line 18: Why is sunlight a special case here? This doesn't make sense to me as if we average the values for sunlight to track how well species can survive, should we not also average all the other compounds as well, because in the future we might model like oxygen levels dropping or something like that during the night. And the same exact logic with the averages should apply to that.

src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
```diff
@@ -15,14 +15,16 @@ public EnvironmentalFoodSource(Patch patch, Compound compound, float foodCapacit
...
+                patch.Biome.Sunlight!.Average * foodCapacityMultiplier :
```
hhyyrylainen, line 19: This nullability override feels pretty wrong to me here, even though the check is above, I think this should be rewritten like the line below (access using the compound key).

src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
```diff
@@ -42,5 +44,55 @@ public override float TotalEnergyAvailable()
         {
             return totalEnvironmentalEnergySource;
         }
+
+        protected override float StorageScore(MicrobeSpecies species, Compound compound, Patch patch,
+            SimulationCache simulationCache)
+        {
+            if (simulationCache.DayNightCycleEnabled)
+            {
+                var dayNightCycleConfiguration = SimulationParameters.Instance.GetDayNightCycleConfiguration();
```
hhyyrylainen, line 53: I *think* it is wrong to access the global simulation parameters here, auto-evo needs to only use the parameters it is given through the top level methods and everything else needs to be passed down to where it is used.

src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
```diff
@@ -42,5 +44,55 @@ public override float TotalEnergyAvailable()
...
+                var nightTime = dayNightCycleConfiguration.DaytimePercentage * dayNightCycleConfiguration.HoursPerDay;
+
+                // If a species consumes a lot, it ought to store more.
+                // NB: this might artificially penalize overproducers but I'm willing to accept it for now - Maxonovien
```
hhyyrylainen, line 57: What does NB stand for?

shaders/MicrobeBackground.shader
```diff
@@ -15,6 +18,47 @@ const vec2 speed3 = vec2(1050.0f);
...
+        time = 1;
+    else
+        time = 0;
```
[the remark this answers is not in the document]
GameDungeon: The never use If statements in shaders is really old advice. The software can handle If statements now with almost no performance effect. Also, there is no way an if statement that just sets a value won't immediately get optimized by the compiler.

src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
```diff
@@ -42,5 +44,55 @@ public override float TotalEnergyAvailable()
...
+                    return minimumViableReserveScore / Constants.AUTO_EVO_NON_VIABLE_RESERVE_PENALTY;
+
+                // TODO: consider using a cap, e.g two times the viable reserve (i.e. osmoregulation + base movement.
+                return reserveScore;
+            }
+
+            return 1.0f;
+        }
+
+        // TODO CACHE IT, AS FOR ENERGY GENERATION, TO AVOID MULTIPLE LOOPS.
```
hhyyrylainen, line 74: As the TODO says this should be cached. This PR already adds caching for the storage count, so adding one more cached thing should be relatively easy.

shaders/MicrobeBackground.shader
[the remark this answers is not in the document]
hhyyrylainen: There's an if + a switch right after. Any branching operation in GPU still needs to take all of the cores in an execution group separately through all of the branches, though in this case as the value is an uniform it's not as bad as it could be. But to me the if followed by a switch statement looks just really not optimal graphics programming and triggers alarms in my head.

src/auto-evo/simulation/food_source/RandomEncounterFoodSource.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    /// <summary>
+    ///   A class for random-encounter based food sources.
+    /// </summary>
+    public abstract class RandomEncounterFoodSource : FoodSource
+    {
+        protected override float StorageScore(MicrobeSpecies species, Compound compound, Patch patch,
+            SimulationCache simulationCache)
+        {
+            // TODO compute value based on encouter chance ; this is a temporary solution for day/night cycle PR.
```
hhyyrylainen, line 11:
```suggestion
            // TODO: compute value based on encouter chance; this is a temporary solution for day/night cycle PR.
```
Also I doubt this will end up being "temporary" because if this is not done properly now, I'll predict that in like 3 years this "temporary" code will still be used.

src/early_multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -444,6 +444,15 @@ public bool ShowCellOptions()
         return true;
     }
 
+    public override void OnLightLevelChanged(float lightLevel)
+    {
+        // Normalise by maximum light level in the patch
+        var maxLightLevel = Editor.CurrentPatch.GetCompoundAmount("sunlight", CompoundAmountType.Maximum);
```
hhyyrylainen, line 450: This complex math should be put in the caller. So I suggest the light level is changed to be the following signature:
```
OnLightLevelChanged(float lightFraction, float absoluteLux)
```

src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
```diff
@@ -106,7 +106,8 @@ public override void SetEditorObjectVisibility(bool shown)
 
     public void OnCurrentPatchUpdated(Patch patch)
     {
-        cellEditorTab.CalculateOrganelleEffectivenessInPatch(patch);
+        cellEditorTab.SetPatchConditions(patch);
+        cellEditorTab.CalculateOrganelleEffectivenessInPatch();
```
hhyyrylainen, line 110: I think rather than modifying CalculateOrganelleEffectivenessInPatch to no longer take a patch (the hint is in the name that it should absolutely take a patch as a parameter) and adding a new method, one higher level method for applying patch conditions should be made and called here.

src/general/DayNightConfiguration.cs
```diff
@@ -0,0 +1,30 @@
...
+    public float HoursPerDay { get; private set; }
+
+    /// <summary>
+    ///   Percentage of the in-game day which has sunlight.
+    /// </summary>
+    [JsonProperty]
+    public float DaytimePercentage { get; private set; }
+
+    public void Check(string name)
+    {
```
hhyyrylainen, line 24: This should at least check that hours per day and the percentage are not negative. And also percentage doesn't make sense to allow it to be over 100.

src/general/DayNightConfiguration.cs
```diff
@@ -0,0 +1,30 @@
...
+    ///   Number of in-game hours per in-game day.
+    /// </summary>
+    [JsonProperty]
+    public float HoursPerDay { get; private set; }
+
+    /// <summary>
+    ///   Percentage of the in-game day which has sunlight.
+    /// </summary>
+    [JsonProperty]
+    public float DaytimePercentage { get; private set; }
```
hhyyrylainen, line 21: Is this a percentage or a fraction? Most of the game systems use fractions to be easier for computing. I'd very highly recommended that this is switched to be a fraction rather than a percentage.

src/general/DayNightCycle.cs
```diff
@@ -0,0 +1,121 @@
...
+/// </summary>
+[JsonObject(IsReference = true)]
+[UseThriveSerializer]
+public class DayNightCycle
+{
+    /// <summary>
+    ///   Configuration details for the day/night cycle.
+    /// </summary>
+    [JsonProperty]
+    public DayNightConfiguration LightCycleConfig;
```
hhyyrylainen, line 16: Do these have to be public *fields*? This is not very good encapsulation design for this class that actually does something. Public properties should only really exist in data classes (or Godot NodePaths as that style became dominant there).

src/microbe_stage/BiomeConditions.cs
```diff
@@ -58,6 +58,8 @@ public object Clone()
         {
             Compounds = new Dictionary<Compound, EnvironmentalCompoundProperties>(Compounds.Count),
             Chunks = new Dictionary<string, ChunkConfiguration>(Chunks.Count),
+            Sunlight = new DaylightProperties(
+                Sunlight?.Maximum ?? 0.0f, Sunlight?.Current ?? 0.0f, Sunlight?.Average ?? 0.0f),
```
hhyyrylainen, line 62: I'd like to see this line refactored as this is pretty crazy amount of question marks on a single line. Also this has the effect that all cloned conditions *always* have the sunlight, but in the constructor, this is not created by default, which I assumed means that patches with no sunlight would have the sunlight be null. Why is there a discrepancy?

src/general/DaylightProperties.cs
```diff
@@ -0,0 +1,17 @@
+﻿public class DaylightProperties
+{
+    public float Maximum;
```
hhyyrylainen, line 3: I can't tell which values here are allowed to change, it would be super awesome to make these properties with setters only for the values that are allowed to be changed.

src/general/EditorBase.cs
```diff
@@ -136,6 +138,21 @@ public GameProperties CurrentGame
 
     public bool NodeReferencesResolved { get; private set; }
 
+    [JsonIgnore]
```
hhyyrylainen, line 141: The editor remembers the symmetry setting (unless I'm misremembering) so why shouldn't the editor also remember the selected light level (I assume there's a button to select one)?

src/general/StageBase.cs
```diff
@@ -196,10 +199,25 @@ public virtual void ResolveNodeReferences()
         NodeReferencesResolved = true;
     }
 
+    public override void _Ready()
+    {
+        base._Ready();
+
+        // Need to check for null light cycle in case this is loaded from an old save
```
hhyyrylainen, line 206: Please use the save load notification callbacks instead of doing this in _Ready.

src/general/StageBase.cs
```diff
@@ -25,6 +25,9 @@ public abstract class StageBase<TPlayer> : NodeWithInput, IStage, IGodotEarlyNod
     protected PauseMenu pauseMenu = null!;
     protected Control hudRoot = null!;
 
+    [JsonProperty]
+    protected DayNightCycle? lightCycle;
```
hhyyrylainen, line 29: The intention seems that the light cycle is never null, so this should not be nullable and use null suppression instead.

src/general/StageHUDBase.cs
```diff
@@ -711,7 +711,7 @@ public void UpdateEnvironmentalBars(BiomeConditions biome)
         var oxygenPercentage = biome.Compounds[oxygen].Ambient * 100;
         var co2Percentage = biome.Compounds[carbondioxide].Ambient * 100;
         var nitrogenPercentage = biome.Compounds[nitrogen].Ambient * 100;
-        var sunlightPercentage = biome.Compounds[sunlight].Ambient * 100;
+        var sunlightPercentage = Math.Round(biome.Sunlight!.Current * 100, 0);
```
hhyyrylainen, line 714: This line seems to indicate that the assumption is that sunlight must always exist, so someone should really consider if the nullability is right on that field in the conditions.

src/general/WorldGenerationSettings.cs
```diff
@@ -106,6 +106,16 @@ public enum PatchMapType
...
 
+    /// <summary>
+    ///   Whether the day/night cycle in this game is enabled
+    /// </summary>
+    public bool DayNightCycleEnabled { get; set; }
+
+    /// <summary>
+    ///   Real-time length of a full day on the planet in seconds
+    /// </summary>
+    public int DayLength { get; set; } = 180;
```
hhyyrylainen, line 117: 180 should be a value in Constants for easy tweaking for balancing.

src/microbe_stage/MicrobeCamera.cs
```diff
@@ -191,6 +194,11 @@ public bool Zoom(float delta, float value)
         return true;
     }
 
+    public override void _Process(float delta)
+    {
+        materialToUpdate.SetShaderParam("lightLevel", LightLevel);
```
hhyyrylainen, line 199: LightLevel should be made into a property that only applies this value to the shader when it changes.

src/microbe_stage/Patch.cs
```diff
@@ -16,6 +16,9 @@
 [UseThriveSerializer]
 public class Patch
 {
+    [JsonIgnore]
```
hhyyrylainen, line 19: You *really* don't need json attributes on static private readonlies...

src/microbe_stage/Patch.cs
```diff
@@ -16,6 +16,9 @@
...
+    private static readonly Compound Sunlight = SimulationParameters.Instance.GetCompound("sunlight");
```
hhyyrylainen, line 20: Also I'm kind of going to say that in this case this GetCompound call should be directly in the Patch constructors. This is because static compound value retrieves are a bit bad in terms of global game state and for example modding new compounds.

src/microbe_stage/Patch.cs
```diff
@@ -404,6 +413,16 @@ public void ReplaceSpecies(Species old, Species newSpecies, bool replaceInHistor
         // TODO: can we do something about the game log here?
     }
 
+    public void UpdateAverageSunlight(DayNightCycle lightCycle)
+    {
+        Biome.Sunlight!.Average = Biome.Sunlight.Maximum * lightCycle.AverageSunlight;
```
hhyyrylainen, line 418: This Sunlight nullability override that's all over really tells me that the design on Biome.Sunlight being nullable is wrong, instead it *should not* be nullable to make other places in the code much easier to reason about.

src/microbe_stage/Patch.cs
```diff
@@ -423,6 +442,21 @@ public override string ToString()
     {
         return $"Patch \"{Name}\"";
     }
+
+    private float GetSunlightAmount(CompoundAmountType option)
+    {
+        switch (option)
+        {
+            case CompoundAmountType.Maximum:
+                return Biome.Sunlight!.Maximum * 100;
```
hhyyrylainen, line 451: The 100 here and in `GetCompoundAmountInSnapshot` should be extracted as a constant as this is now used in a bunch of places as a magic value.

src/microbe_stage/PatchManager.cs
```diff
@@ -109,12 +119,45 @@ public bool ApplyChangedPatchSettingsIfNeeded(Patch currentPatch)
 
         // Change the lighting
         UpdateLight(currentPatch.BiomeTemplate);
-
-        compoundCloudSystem.SetBrightnessModifier(currentPatch.BiomeTemplate.CompoundCloudBrightness);
+        compoundCloudBrightness = currentPatch.BiomeTemplate.CompoundCloudBrightness;
```
hhyyrylainen, line 122: I dislike how hard I need to think here "does compoundCloudBrightness variable get properly sent to the system" now with this change. Could this be done as a private property instead or maybe this variable could be removed entirely and instead all the methods would call into compoundCloudSystem directly?

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -402,8 +402,12 @@ private static void BuildPatchesInRegion(PatchRegion region, Random random)
 
                 // Build seafloor light, using 0m -> 1, 200m -> 0.01, floor to 0.01
                 var sunlightProperty = seafloor.Biome.Compounds[sunlightCompound];
-                sunlightProperty.Ambient = (int)(Mathf.Pow(0.977237220956f, seafloor.Depth[1]) * 100) / 100.0f;
+                var sunlightAmount = (int)(Mathf.Pow(0.977237220956f, seafloor.Depth[1]) * 100) / 100.0f;
+                sunlightProperty.Ambient = sunlightAmount;
+
+                // Need to initialise both to account for loading from saves
                 seafloor.Biome.Compounds[sunlightCompound] = sunlightProperty;
+                seafloor.Biome.CreateSunlight(sunlightAmount);
```
hhyyrylainen, line 410: I think based on the other code, to make the sunlight more similar to the other biome setup stuff, the sunlight property should be default constructed, and the patch modification code here would then just adjust the light properties (similarly to how the compounds dictionary doesn't need to be created here but is just modified).

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -428,6 +428,9 @@ private void UpdateGUIAfterLoadingSpecies(Species species, ICellProperties prope
         UpdateGeneration(species.Generation);
         UpdateHitpoints(CalculateHitpoints());
         UpdateStorage(CalculateStorage());
+
+        // Set the editor light level and associated GUI elements to daytime
+        SetLightLevelOption("Day");
```
hhyyrylainen, line 433: This should do something like `LightLevelOption.Day.ToString()` for futureproofing.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -784,27 +825,40 @@ public override bool CanFinishEditing(IEnumerable<EditorUserOverride> userOverri
...
     }
 
+    public void SetPatchConditions(Patch patch)
+    {
+        editorPatchConditions = (BiomeConditions)patch.Biome.Clone();
+
+        // If the editor has initialised (i.e. if this is a change of patch during an editor session), switch back to
+        // daytime light levels
+        if (Editor.Ready)
+            SetLightLevelOption("Day");
```
hhyyrylainen, line 835: Same thing here.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -784,27 +825,40 @@ public override bool CanFinishEditing(IEnumerable<EditorUserOverride> userOverri
         return true;
     }
 
+    public void SetPatchConditions(Patch patch)
+    {
+        editorPatchConditions = (BiomeConditions)patch.Biome.Clone();
+
+        // If the editor has initialised (i.e. if this is a change of patch during an editor session), switch back to
+        // daytime light levels
+        if (Editor.Ready)
```
hhyyrylainen, line 834: I think instead of detecting this, this method should take in a bool that's true when things are being initialized.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -784,27 +825,40 @@ public override bool CanFinishEditing(IEnumerable<EditorUserOverride> userOverri
...
+    }
+
     public void UpdatePatchDependentBalanceData()
     {
         // Skip if not opened in the multicellular editor
         if (IsMulticellularEditor && editedMicrobeOrganelles.Organelles.Count < 1)
             return;
 
+        topPanel.Visible = Editor.CurrentGame.GameWorld.WorldSettings.DayNightCycleEnabled &&
+            Editor.CurrentPatch.GetCompoundAmount("sunlight", CompoundAmountType.Maximum) > 0.0f;
```
hhyyrylainen, line 845: sunlight should be retrieved as a Compound object and used here instead of a string key.

src/microbe_stage/editor/CellEditorComponent.tscn
```diff
@@ -353,7 +469,6 @@ custom_styles/pressed = SubResource( 10 )
 custom_styles/disabled = SubResource( 11 )
 custom_styles/normal = SubResource( 12 )
 toggle_mode = true
-pressed = true
```
hhyyrylainen, line 356: This seems incorrect to me.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1966,6 +2038,47 @@ private void CopyEditedPropertiesToSpecies(MicrobeSpecies target)
...
+
+        GUICommon.Instance.PlayButtonPressSound();
+
+        selectedLightLevelOption = selection;
+        ApplyLightLevelOption();
+    }
+
+    private void ApplyLightLevelOption()
+    {
+        // Show selected
```
hhyyrylainen, line 2053: Shouldn't this unset the pressed values first and then set one to true? Similarly to all of the other tab-style button management already in the game.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -971,6 +1016,32 @@ public float CalculateTotalDigestionEfficiency()
         return MicrobeInternalCalculations.CalculateTotalDigestionEfficiency(editedMicrobeOrganelles);
     }
 
+    public override void OnLightLevelChanged(float lightLevel)
+    {
+        var maxLightLevel = Editor.CurrentPatch.GetCompoundAmount("sunlight", CompoundAmountType.Maximum);
+        var templateMaxLightLevel = Editor.CurrentPatch.GetCompoundAmount("sunlight", CompoundAmountType.Template);
+
+        if (maxLightLevel > 0.0f && templateMaxLightLevel > 0.0f)
```
hhyyrylainen, line 1024: Shouldn't either of these being more than 0 work? We might want to allow patches to deviate from the template in terms of max light level.

src/auto-evo/simulation/SimulationCache.cs
Oliveriver: Are you suggesting making worldSettings public in this class or passing worldSettings to the other classes here?

hhyyrylainen: World settings should be passed around where needed. It's not the cache's job to allow other parts of the code get the settings through the cache object.

src/auto-evo/simulation/food_source/EnvironmentalFoodSource.cs
Oliveriver: I would assume [nota bene](https://en.wikipedia.org/wiki/Nota_bene).

hhyyrylainen: Then this should be rewritten as `NOTE:`

src/microbe_stage/editor/CellEditorComponent.cs
Oliveriver: The issue here is that only some backgrounds have been given proper nighttime colouring, specifically those whose templates have non-zero light level. The sea floor for instance has zero sunlight in its definition but can be given non-zero sunlight as a patch instance. @Nunez2196 must have missed this and didn't provide a proper background shader, so it just turns pitch black at night without this check.

I think this should be raised as an issue to deal with in future. For now I'll definitely add a comment here.

shaders/MicrobeBackground.shader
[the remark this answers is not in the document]
Oliveriver: I believe I've addressed this but I'll leave this comment unresolved as it needs checking from someone who understands the requirements of shader code.

src/general/DayNightConfiguration.cs
Oliveriver: It is a fraction (if by that you mean a value between 0 and 1). Probably warrants a name change.

Oliveriver:
I've dealt with most of the minor issues, but I ran out of mental capacity to take on the big compound amount interface refactor and a few other things. Also aware I completely missed adding tutorial details for the day/night cycle, as discussed in Discord.

I've also done some balancing. In 100% maximum light patches, a thyloakoid now produces 0.001 glucose across a day on average, and a basic eukaryote with chloroplasts is similar. I played with these settings for a bit and found it quite fun since I had the drama of watching my glucose level decrease overnight but it always started rising again before it drained entirely. It also encouraged sessility, especially during the night.

I'm going to have to leave this to other people to sort out the final few bits. I apologise for that. Again, I'd like to emphasise the work I and others have put into this (and also [the fact it remains the most wanted suggestion](https://suggestions.revolutionarygamesstudio.com/posts/1/day-night-cycle-for-microbes-to-nerf-photosynthesis)) as evidence for it being worthy of prioritising for 0.6.0.

src/microbe_stage/Patch.cs
GameDungeon: Doesn't that 100 just convert it from a decimal percentage to a whole (for lack of a better term) percentage?

src/general/DayNightConfiguration.cs
GameDungeon: What would that name change be though? I think DaytimeFraction is even more confusing. Maybe the value should just be noted in the summary?

shaders/MicrobeBackground.shader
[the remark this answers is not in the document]
hhyyrylainen: Unless someone can math formula this out, then I *think* having just one sequence of ifs is *probably* fine enough. I think that maybe a better approach would be to pass the entire multiplier vec as an uniform. That way the c# code can do the calculation once (which should be very cheap with a couple of ifs) and all of the shader units don't need to take conditional branches then.

src/general/DayNightConfiguration.cs
hhyyrylainen: Everywhere else in the game I've enforced the convention that variables with "fraction" in the name mean the valid range is 0-1 and anything with "percentage" in the name has valid value range of 0-100. I'll make a change to the styleguide about this naming in my launcher 2.0 features branch.

src/microbe_stage/Patch.cs
hhyyrylainen: Well the method names are not clear at all what they return. And also the "normal" compounds don't seem to be multiplied by anything. I'm definitely guilty of using 100 as a suitable value multiplier without intending it to be taken as a percentage value.

hhyyrylainen: At least a summary for the methods should be added explaining what kind of values it returns.

src/general/StageBase.cs
GameDungeon: By this do you mean Setup Stage or is there something I'm missing?

GameDungeon:
We really should determine what can be moved to issues for the future, and what absolutely needs to be added to this pull. Both because I agree with Oliver about prioritizing this for 0.6.0, as well as preventing this already massive pull request from becoming much larger.

src/general/StageBase.cs
GameDungeon: As far as I can tell this needs to be here otherwise jetbrains gets mad at this check:

```
    public override void _Ready()
    {
        base._Ready();

        // Need to check for null light cycle in case this is loaded from an old save
        if (!IsLoadedFromSave || lightCycle == null)
        {
            lightCycle = new DayNightCycle(
                GameWorld.WorldSettings.DayNightCycleEnabled,
                GameWorld.WorldSettings.DayLength);
        }
    }
```

After that check it should be impossible for lightCycle to be null

src/general/StageBase.cs
hhyyrylainen: Yes you are missing use of `ISaveLoadable` interface.

src/general/StageBase.cs
hhyyrylainen: You can write the null check by putting the nullability override there, at least I'm pretty sure you can do that (`lightCycle == null!`).

hhyyrylainen:
As long as the day/night cycle is disabled by default, everything that doesn't impact the rest of the game (regressions, or bad data formats that likely require breaking save compatibility in the future) can be left for later.

---

Also I had a few more things regarding this feature that popped into my head:
- Shouldn't cells that spawn just a bit before the night have their spawn compounds adjusted so that they survive the night?
- Shouldn't the AI be updated to know to stay more sessile during the night / update auto-evo to take sessility into account when determining if a species will survive or die constantly.

We don't want to mess with the game balance (for predator and scavenger builds) due to those issues. But if this feature is not enabled by default those can be left for later.

GameDungeon:
Do you have anything that you would say counts as "things that impact the rest of the game, regressions, or bad data formats that likely require breaking save compatibility in the future" that needs to be prioritized?

hhyyrylainen:
The way the sunlight is stored separately / refactoring to use an interface adapter that gives current or template values for compounds, I think has pretty good chance to require thinking in terms of saving / loading. The other is the editor graphs breaking (which is currently unknown if the breaking also happens when the day/night cycle is not enabled).

GameDungeon:
Looking into making the interface. Could you describe it to me? What should it look like and do?

hhyyrylainen:
I think the most basic thing would be to start with it having just a method to get a compound amount. Then modifying the patch etc. classes to have methods like `GetStaticCompounds` `GetCurrentCompounds` etc. methods that return classes implementing that interface. If I were doing this I'd also remove the compound getting methods from Patch so that I could find and fix the places that use the compound values to call the right methods that make sense for them for example `GetStaticCompounds` in the context where the day/night adjusted values are not wanted.

GameDungeon:
Any suggestions for a name? So at its most basic, it's an interface for a getter for Template, Biome, and Realtime?

hhyyrylainen:
I'd suggest something like `IEnvironmentCompounds` (or perhaps `IAvailableCompounds`) or something like that to clarify in the name that it is not related to `ICompoundStorage` or the `CompoundBag`

hhyyrylainen:
I'm going to make at least one major data change: I'm going to put the hours per day etc. settings into the gameworld as I really think players will want to play on planets with different length days in the future. For now I'll use the json values as defaults. This will also allow me to undo the changes to auto-evo where the world settings and day settings need to be passed along as two separate parameters, which is something I dislike as there's already a ton of method parameters used there.

src/microbe_stage/editor/CellEditorComponent.tscn
hhyyrylainen: Note to myself to fix, this problem seems to be back...

hhyyrylainen (approved):
I've now fixed all code style and bugs I noticed in this. We are still waiting for one more feature improvement @Oliveriver wants to make for this.

Oliveriver:
I wasn't able to add the current light button because I found a bug due to the refactoring where average light levels weren't being set properly. It took some time to decide where to put the fix. I also found and fixed a minor UI bug where the press sound was playing when it shouldn't have been.

After a fair amount of testing, I haven't found any other bugs. The process panel freezing when the player is ready to reproduce is not a regression in this PR, since it happens in master as well.

As far as I can tell, that leaves the following features to add later, instead of as part of this PR:

* Tutorial for the day/night cycle.
* [Button for current light level in the editor](https://github.com/Revolutionary-Games/Thrive/issues/3881) (including changing the light level displayed in the patch map if that's the design decision taken).
* Cheat to set the time of day.
* Perhaps some further balancing. I think it's fairly well balanced now. Perhaps even slightly too easy for photosynthesizers, but not by much.
* A storage warning in the editor if the player can't survive the night.
* Reduce the update rate for light level to improve performance.
* Once all those are done, enable day/night cycle by default in new games.

GameDungeon:
Is it worth it to create an issue for them yet?

GameDungeon:
After playtesting for a while, everything works as far as I can tell. Nothing weird happened.

hhyyrylainen:
I opened issues for those things still to be done for the day/night cycle. Here's the meta issue tracking when it can be enabled by default: https://github.com/Revolutionary-Games/Thrive/issues/3895

PR #3747, opened 2022-08-26
Properly restore the "orphaned" tooltips

### LINKED ISSUE #3696: Investigate the "orphaned" tooltips
Got a HUGE amount of tool tips after several patch changes.

137 tool tip groups, many having 10+ tool tips.
image

The worst:
image

It is obvious that some code which adds a tooltip fails to remove it.

84634E1A607A:
Some tooltips are not removed correctly when related nodes are disposed.

**Related Issues**
Fixes #3696

**Progress Checklist**

      break existing features:
      members reviewing this PR)


84634E1A607A:
This further reduced the lag when changing patch, now that tooltips are correctly restored, and we don't need to request for new tooltip nodes each and every patch change.

src/gui_common/charts/line/LineChart.cs
```diff
@@ -385,6 +395,9 @@ public void ClearDataSets()
 
             // Initialize line
             var dataLine = new DataLine((LineChartData)data.Value, data.Key == defaultDataSet, isChild ? 2 : 1);
+            dataLine.Connect("tree_exited", this, nameof(RemoveDataLineTooltipOnDeletion), new Array(dataLine));
+
+            dataLineTooltips.Add(dataLine, new List<ICustomToolTip>());
```
Kasterisk, line 400: For consistency and extra safety, add check for existing keys.
https://github.com/Revolutionary-Games/Thrive/blob/0f71e7936bbc2adef1bc1480c7418231202cb2a8/src/gui_common/charts/line/LineChart.cs#L391-L392

src/gui_common/charts/line/LineChart.cs
```diff
@@ -385,6 +395,9 @@ public void ClearDataSets()
 
             // Initialize line
             var dataLine = new DataLine((LineChartData)data.Value, data.Key == defaultDataSet, isChild ? 2 : 1);
+            dataLine.Connect("tree_exited", this, nameof(RemoveDataLineTooltipOnDeletion), new Array(dataLine));
```
Kasterisk, line 398: I think this should be more general, that is to connect `LineChart`'s own `tree_exited` signal to a new method called `ClearTooltips` to make sure everything else gets destroyed. The tooltip clearing logic in `ClearChart` needs to be moved there and the former just simply call `ClearTooltips`.

src/gui_common/charts/line/LineChart.cs
```diff
@@ -620,6 +633,14 @@ public void OverrideDataPointToolTipDescription(string dataset, DataPoint datapo
         }
     }
 
+    private void RemoveDataLineTooltipOnDeletion(DataLine line)
+    {
+        foreach (var tooltip in dataLineTooltips[line])
+        {
+            ToolTipManager.Instance.RemoveToolTip(tooltip.ToolTipNode.Name, LINE_CHART_GROUP_BASE_NAME);
+        }
+    }
```
Kasterisk, line 642: It's rarely that we need to remove a specific dataset and thus each line's tooltip separately so an all in one solution seems fitting.

Kasterisk (changes requested):

Kasterisk:
Reading through the code again, it seems like my past self made each line segments have their own tooltips in the _same_ dataset. That's rather inefficient and probably needs optimization. Commented here for future records.

src/gui_common/charts/line/LineChart.cs
84634E1A607A: LineChart delete lines when data is updated. (And that was the main cause of orphaned tooltips).

Kasterisk: Yes that's because `ClearChart` is called which among other things also clears the tooltips.

Kasterisk (approved):
Well, good enough I guess.

PR #3748, opened 2022-08-28
Added cache for `DataPoint`

84634E1A607A:
Before (Using UnregisterTooltip as reference):
image

After
image

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/gui_common/charts/DataPoint.cs
```diff
@@ -40,6 +40,9 @@ public enum MarkerIcon
         Skull,
     }
 
+    public double X { get; private set; }
+    public double Y { get; private set; }
```
Kasterisk, line 44: Need to check that hash code computation is still correct, last time I made these readonlies because otherwise DataPoint dictionaries end up failing to fetch the correct entries.

Kasterisk:
Looked at the diff, I think that's a weird way to put the cache, why not let LineChart handle that?

src/gui_common/charts/DataPoint.cs
84634E1A607A: As a cache, the x and y won't be set when there are external references. If the x and y is changed but reference is still held in a dict, it is a bug that needs to be fixef.

84634E1A607A:
> Looked at the diff, I think that's a weird way to put the cache, why not let LineChart handle that?

All line chart shares data points. This minimize the chance that a new datapoint needs to be constructed.

src/gui_common/charts/DataPoint.cs
hhyyrylainen: If X and Y are used in GetHashCode, then this is pretty dangerous to have these mutable...
Maybe the design could be made safer by not overriding the hash code getting for this class, and instead in charts have an extra dictionary mapping immutable coordinates to data points? For example `Dictionary<(double X, double Y), DataPoint>`

src/gui_common/charts/DataPoint.cs
```diff
@@ -103,6 +106,34 @@ public float Size
         return !(lhs == rhs);
     }
 
+    public static DataPoint GetDataPoint(double x, double y, MarkerIcon iconType = MarkerIcon.Circle, float size = 7,
+        Color markerColour = default, bool draw = true, Vector2 coordinate = default)
```
Kasterisk, line 110: Docs please.

Kasterisk (changes requested):

Kasterisk (approved):

src/gui_common/charts/DataPoint.cs
```diff
@@ -103,6 +106,43 @@ public float Size
...
+        point.coordinate = coordinate;
+        point.Visible = true;
+        return point;
+    }
+
+    /// <summary>
+    ///   Returns a datapoint to the cache. Make sure that the point isn't used anywhere else before returning it to
+    ///   the cache.
+    /// </summary>
+    public static void RestoreDataPoint(DataPoint point)
```
hhyyrylainen, line 139:
```suggestion
    public static void ReturnDataPoint(DataPoint point)
```
I think the method name would be clearer that way and would also match the documentation text.

src/gui_common/charts/DataPoint.cs
hhyyrylainen: So reading the code it turns out that the cache doesn't care about the properties. But is there a conclusion on whether this had any impact on the chart class? I guess as these are private it is good enough even without a comment as only when getting a point from cache it can be modified.

src/gui_common/charts/DataPoint.cs
```diff
@@ -103,6 +106,43 @@ public float Size
...
+    {
+        point.tween.StopAll();
```
hhyyrylainen, line 141: Could you add a real quick sanity check here that the DataPoint parent is null (ie. it is not in the scene tree)? Just in case someone doesn't follow the advice in the summary comment about not using the Datapoint.

hhyyrylainen (approved):
I didn't playtest (let's assume that enough testing for the graphs will happen at the latest once the RC is out), but the code now looks fine.

PR #3754, opened 2022-08-30
Weight contribution to entity limit by size of cells

Oliveriver:

A cell now contributes x to the entity count, where x is the number of organelles it contains.

Made this a while ago but forgot to push it. This does appear to have a small impact on performance, but not a massive one. It does have a gameplay impact since fewer cells will spawn in general, especially if they're large.

I'd appreciate some testing and feedback from both performance and gameplay perspectives. This isn't a necessary change, but if it works, it would be nice. Especially for something so simple.

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/DebugOverlays.PerformanceMetrics.cs
+++ b/src/engine/DebugOverlays.PerformanceMetrics.cs
@@ -21,17 +21,17 @@ public partial class DebugOverlays
     // TODO: make this time based
     private const int SpawnHistoryLength = 300;
 
-    private readonly Deque<int> spawnHistory = new(SpawnHistoryLength);
-    private readonly Deque<int> despawnHistory = new(SpawnHistoryLength);
+    private readonly Deque<float> spawnHistory = new(SpawnHistoryLength);
+    private readonly Deque<float> despawnHistory = new(SpawnHistoryLength);
 
     private Label fpsLabel = null!;
     private Label deltaLabel = null!;
     private Label metricsText = null!;
 
-    private int entities;
+    private float entities;
     private int children;
-    private int currentSpawned;
-    private int currentDespawned;
+    private float currentSpawned;
+    private float currentDespawned;
 
     public bool PerformanceMetricsVisible
     {
@@ -45,18 +45,18 @@ public partial class DebugOverlays
         }
     }
 
-    public void ReportEntities(int totalEntities, int otherChildren)
+    public void ReportEntities(float totalEntities, int otherChildren)
     {
         entities = totalEntities;
         children = otherChildren;
     }
 
-    public void ReportSpawns(int newSpawns)
+    public void ReportSpawns(float newSpawns)
     {
         currentSpawned += newSpawns;
     }
 
-    public void ReportDespawns(int newDespawns)
+    public void ReportDespawns(float newDespawns)
     {
         currentDespawned += newDespawns;
     }
@@ -83,7 +83,8 @@ public partial class DebugOverlays
         metricsText.Text =
             new LocalizedString("METRICS_CONTENT", Performance.GetMonitor(Performance.Monitor.TimeProcess),
                     Performance.GetMonitor(Performance.Monitor.TimePhysicsProcess),
-                    entities, children, spawnHistory.Sum(), despawnHistory.Sum(),
+                    Math.Round(entities, 1), children,
+                    Math.Round(spawnHistory.Sum(), 1), Math.Round(despawnHistory.Sum(), 1),
                     Performance.GetMonitor(Performance.Monitor.ObjectNodeCount),
                     OS.GetName() == Constants.OS_WINDOWS_NAME ?
                         TranslationServer.Translate("UNKNOWN_ON_WINDOWS") :
@@ -99,7 +100,7 @@ public partial class DebugOverlays
                     Performance.GetMonitor(Performance.Monitor.AudioOutputLatency) * 1000, threads, processorTime)
                 .ToString();
 
-        entities = 0;
+        entities = 0.0f;
         children = 0;
 
         spawnHistory.AddToBack(currentSpawned);
@@ -111,7 +112,7 @@ public partial class DebugOverlays
         while (despawnHistory.Count > SpawnHistoryLength)
             despawnHistory.RemoveFromFront();
 
-        currentSpawned = 0;
-        currentDespawned = 0;
+        currentSpawned = 0.0f;
+        currentDespawned = 0.0f;
     }
 }
--- a/src/general/StageBase.cs
+++ b/src/general/StageBase.cs
@@ -252,9 +252,9 @@ public abstract class StageBase<TPlayer> : NodeWithInput, IStage, IGodotEarlyNod
 
         if (debugOverlay.PerformanceMetricsVisible)
         {
-            var entities = rootOfDynamicallySpawned.GetChildrenToProcess<ISpawned>(Constants.SPAWNED_GROUP).Count();
+            var entities = rootOfDynamicallySpawned.GetChildrenToProcess<ISpawned>(Constants.SPAWNED_GROUP).ToList();
             var childCount = rootOfDynamicallySpawned.GetChildCount();
-            debugOverlay.ReportEntities(entities, childCount - entities);
+            debugOverlay.ReportEntities(entities.Sum(e => e.EntityWeight), childCount - entities.Count);
         }
     }
 
--- a/src/general/StageBase.cs
+++ b/src/general/StageBase.cs
@@ -252,9 +252,17 @@ public abstract class StageBase<TPlayer> : NodeWithInput, IStage, IGodotEarlyNod
 
         if (debugOverlay.PerformanceMetricsVisible)
         {
-            var entities = rootOfDynamicallySpawned.GetChildrenToProcess<ISpawned>(Constants.SPAWNED_GROUP).ToList();
+            float totalEntityWeight = 0;
+            int totalEntityCount = 0;
+
+            foreach (var entity in rootOfDynamicallySpawned.GetChildrenToProcess<ISpawned>(Constants.SPAWNED_GROUP))
+            {
+                totalEntityWeight += entity.EntityWeight;
+                ++totalEntityCount;
+            }
+
             var childCount = rootOfDynamicallySpawned.GetChildCount();
-            debugOverlay.ReportEntities(entities.Sum(e => e.EntityWeight), childCount - entities.Count);
+            debugOverlay.ReportEntities(totalEntityWeight, childCount - totalEntityCount);
         }
     }
 
--- a/src/microbe_stage/ISpawned.cs
+++ b/src/microbe_stage/ISpawned.cs
@@ -8,4 +8,9 @@ public interface ISpawned : IEntity
     ///   greater than this, it is despawned.
     /// </summary>
     int DespawnRadiusSquared { get; set; }
+
+    /// <summary>
+    ///   How much this entity contributes to the entity limit relative to a single node
+    /// </summary>
+    float EntityWeight { get; }
 }
--- a/src/microbe_stage/Microbe.cs
+++ b/src/microbe_stage/Microbe.cs
@@ -204,6 +204,13 @@ public partial class Microbe : RigidBody, ISpawned, IProcessable, IMicrobeAI, IS
     public int DespawnRadiusSquared { get; set; }
 
     /// <summary>
+    ///   Entity weight for microbes counts all organelles with a scaling factor.
+    /// </summary>
+    [JsonIgnore]
+    public float EntityWeight => organelles?.Count * Constants.ORGANELLE_ENTITY_WEIGHT ??
+        throw new InvalidOperationException("Organelles not initialised on microbe spawn");
+
+    /// <summary>
     ///   If true this shifts the purpose of this cell for visualizations-only
     ///   (Completely stops the normal functioning of the cell).
     /// </summary>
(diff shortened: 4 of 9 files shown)


src/microbe_stage/ISpawned.cs
```diff
@@ -8,4 +8,9 @@ public interface ISpawned : IEntity
     ///   greater than this, it is despawned.
     /// </summary>
     int DespawnRadiusSquared { get; set; }
+
+    /// <summary>
+    ///   How much this entity contributes to the entity limit (based on Node complexity)
+    /// </summary>
+    int EntityWeight { get; }
```
hhyyrylainen, line 15: I have a feeling that this should really be a float because I imagine the performance impact of some entities related to other ones might be something like 30-50% instead of 100% or 200% and so on.

src/microbe_stage/Microbe.cs
```diff
@@ -203,6 +203,8 @@ public MicrobeSignalCommand SignalCommand
     [JsonProperty]
     public int DespawnRadiusSquared { get; set; }
 
+    public int EntityWeight => organelles?.Count ?? 1;
```
hhyyrylainen, line 206: Entity weights should be JsonIgnore if they are computed like this (even without a setter the json lib will write out this value uselessly into the json).

Also as a safety measure I would just throw an exception if the Microbe is not initialized yet before the entity weight is accessed, because it would be a pretty nightmare debugging situation to try to find out why the entity weight system is not working if we get reports it doesn't seem to work.

yowshi: can we get a setting to turn this off? since it doesn't have a large performance impact but would make hunting harder, it can already be difficult to hunt for those of us who need to use the lower entity limits.

yowshi:
Can we get a setting to turn this off? if it doesn't have much impact on performance but will make hunting harder. hunting can already be difficult to find prey for those of us who need to use the smaller entity limits.

hhyyrylainen:
The entire point of a weighted entity limit is to more accurately measure how much performance impact certain game entities cause. So it doesn't make much sense to add an option to disable this, unless we also had other options like "chunks don't count in entity limit", a separate chunk entity limit etc. This is not something the player should be able to change, this is purely a technical measure of how bad the performance impact of things are related to each other. If you purposefully want to play with worse performance but more entities, turning up the entity limit already works for that.

yowshi:
except that turning up the entity limit at all has in my experience the game either basically not adding more cells or adding hundreds of new entities with no in between.

hhyyrylainen:
That 100% sounds like a different bug because going from tiny to very small is just an increase from 25 to 40. It's not this ideas fault that a different bug exists. There also seems to be an elusive bug where the entity limit is not respected at all. Adding a weighting to the entity limit won't have any effect (or I should probably say any effect I can think of) on those bugs.

src/microbe_stage/ISpawned.cs
Oliveriver: How much of the existing entity counting framework should be adjusted to account for this? Entity count is an int everywhere it turns up, so this would mean either rounding these float values (which to me seems redundant) or having a float total entity count.

hhyyrylainen: I'd change the entity count in just the spawn system to store the value as a float, then it can still be compared against the limits defined as int just fine.

Oliveriver:
I've made a few substantial changes to how this works so I'll mention them here:

* Entity count is now a float, and I've added a constant as a multiplier for how much organelles inside cells contribute.
* As a result, a lot of the entity tracking and debugging code had to be rewritten to use floats. I decided this was necessary as otherwise the displayed entity counts in the debug panel become meaningless. The one part I still have questions about is the debug panel's children count - I used the base node count in calculating this rather than weighted count, so I don't know how useful it is anymore.
* In doing the above, I noticed quite a few locations where we count entities in SpawnSystem that I'd missed, so I updated those to increment by EntityWeight instead of incremeting by 1, as the others do.
* I increased the max despawns per frame constant to account for this.

It seems either I hadn't tested properly before or the places I'd missed were more important than I thought, because I've just managed to play 12 generations of a regular game - with the entity limit option set to _extreme_ - without falling below 55 FPS (except for autosaves post-editor). I'm pretty amazed how big an impact it had.

That said, the environment was noticeably emptier. Though not as much as I feared. It felt about 50% of the normal cell/chunk capacity, except the first two generations where all cells were small.

Assuming my testing wasn't a fluke, I think this is now a pretty important PR to get done. It's well-documented that our performance issues increase with environmental cell complexity, and I think this is the right way (or at least, one of them) to balance that out.

Pending more testing, the constants for the max entity count setting need to be increased to keep the environment lively. How much, I don't know. And maybe it needs some thought in regard to player's stored settings, since they seem to be reset whenever I change these constants.

Oliveriver:
400 entities, consistent 55+ FPS.

Obviously the definition of entity has changed now so this isn't a true comparison, but the environment felt perfectly lively with plenty of big cells and I did indeed have upwards of 55 FPS almost all the time. The only dips seemed to be associated with cells duplicating organelles.

I've significantly increased the max entity options, and even at the new extreme value of 800, I still never dipped below 30 FPS. That's probably near the limit though I imagine, since I struggled to get above 40 FPS.

Here's the save. On master with normal max entities, it quickly spirals to ~20 FPS. On this branch with normal max entities, I get consistent 55+ FPS.

https://drive.google.com/file/d/1CKtZQtdOpdmYsUyY1bzTFUpyPvBPPjx8/view?usp=sharing

I'd really like other people to share their experiences of this branch, because I'm hoping my testing isn't a fluke and this is a reliable way to dramatically improve performance, even if it does affect gameplay a bit.

hhyyrylainen:
> Pending more testing, the constants for the max entity count setting need to be increased to keep the environment lively (again, I was playing with extreme max entities, so everything less can only be worse). How much, I don't know. And maybe it needs some thought in regard to player's stored settings, since they seem to be reset whenever I change these constants.

What do you mean resets? The raw values for the entity limit is stored in the settings, so changing the amount constants will end up the settings menu showing probably the default value while the actual game used variable is different until the player edits that option.

Oliveriver:
> What do you mean resets? The raw values for the entity limit is stored in the settings, so changing the amount constants will end up the settings menu showing probably the default value while the actual game used variable is different until the player edits that option.

Yes, that's what I meant - the value in the options menu resets but the value used by the game does not. Is this acceptable?

hhyyrylainen:
Seeing your latest changes to make even the tiny limit pretty large, I think we'll have to fix up the settings somehow. Probably something like a post validation method that is ran after load to JSON that resets options that have non-sensible values.

src/engine/DebugOverlays.PerformanceMetrics.cs
```diff
@@ -21,17 +21,17 @@ public partial class DebugOverlays
...
     private Label fpsLabel = null!;
     private Label deltaLabel = null!;
     private Label metricsText = null!;
 
-    private int entities;
-    private int children;
-    private int currentSpawned;
-    private int currentDespawned;
+    private float entities;
+    private float children;
```
hhyyrylainen, line 32: Children should not be a float as it is literally a raw Godot Node count.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -110,14 +110,14 @@ public void Clear()
...
-        int despawned = 0;
+        float despawned = 0.0f;
 
         foreach (var spawned in worldRoot.GetChildrenToProcess<ISpawned>(Constants.SPAWNED_GROUP))
         {
             if (!spawned.EntityNode.IsQueuedForDeletion())
             {
                 spawned.DestroyDetachAndQueueFree();
-                ++despawned;
+                despawned += spawned.EntityWeight;
```
hhyyrylainen, line 120: This line should be before the Destroy call as that should be allowed to mess up the entity entirely.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -157,12 +157,12 @@ public void Process(float delta, Vector3 playerPosition)
...
         playerPosition.y = 0;
 
-        int spawnsLeftThisFrame = Constants.MAX_SPAWNS_PER_FRAME;
+        float spawnsLeftThisFrame = Constants.MAX_SPAWNS_PER_FRAME;
 
         // If we have queued spawns to do spawn those
         HandleQueuedSpawns(ref spawnsLeftThisFrame, playerPosition);
 
-        if (spawnsLeftThisFrame <= 0)
+        if (spawnsLeftThisFrame <= 0.0f)
```
hhyyrylainen, line 165:
```suggestion
        if (spawnsLeftThisFrame <= 0)
```
This compiles, right? I prefer code like this to check that a float is not negative.

src/microbe_stage/SpawnSystem.cs
```diff
@@ -203,12 +203,12 @@ public bool IsUnderEntityLimitForReproducing()
...
 
-    private void HandleQueuedSpawns(ref int spawnsLeftThisFrame, Vector3 playerPosition)
+    private void HandleQueuedSpawns(ref float spawnsLeftThisFrame, Vector3 playerPosition)
     {
-        int spawned = 0;
+        float spawned = 0.0f;
 
         // Spawn from the queue
-        while (spawnsLeftThisFrame > 0 && queuedSpawns.Count > 0)
+        while (spawnsLeftThisFrame > 0.0f && queuedSpawns.Count > 0)
```
hhyyrylainen, line 211:
```suggestion
        while (spawnsLeftThisFrame > 0 && queuedSpawns.Count > 0)
```

src/microbe_stage/SpawnSystem.cs
```diff
@@ -388,7 +389,7 @@ private int SpawnWithSpawner(Spawner spawnType, Vector3 location, ref int spawns
 
         var spawner = enumerable.GetEnumerator();
 
-        while (spawnsLeftThisFrame > 0)
+        while (spawnsLeftThisFrame > 0.0f)
```
hhyyrylainen, line 392:
```suggestion
        while (spawnsLeftThisFrame > 0)
```

src/microbe_stage/SpawnSystem.cs
```diff
@@ -422,14 +424,14 @@ private int SpawnWithSpawner(Spawner spawnType, Vector3 location, ref int spawns
     ///   Despawns entities that are far away from the player
     /// </summary>
     /// <returns>The number of alive entities, used to limit the total</returns>
-    private int DespawnEntities(Vector3 playerPosition)
+    private float DespawnEntities(Vector3 playerPosition)
```
hhyyrylainen, line 427: Is it *really* necessary to weight the despawning as well? Does it actually take that much more time to despawn a big cell as a small one? Also now with the adjusted despawn amounts, this can despawn a lot more chunks per frame (which may increase game instability).

src/general/StageBase.cs
```diff
@@ -252,9 +252,9 @@ public override void _Process(float delta)
 
         if (debugOverlay.PerformanceMetricsVisible)
         {
-            var entities = rootOfDynamicallySpawned.GetChildrenToProcess<ISpawned>(Constants.SPAWNED_GROUP).Count();
+            var entities = rootOfDynamicallySpawned.GetChildrenToProcess<ISpawned>(Constants.SPAWNED_GROUP).ToList();
```
hhyyrylainen, line 255: I'd prefer to avoid allocating memory each frame to a new list, but I suppose as this is only used when debugging, this is fine...

hhyyrylainen (changes requested):
There aren't that many major problems here other than the fact that the settings need some code to migrate, because if someone had the entity limit to tiny before this change, they are going to have only 1/4 of the intended minimum spawn amount, and *we'll* get bug reports about this a lot unless fixed. If you don't have time I can do that part myself.

src/microbe_stage/SpawnSystem.cs
Oliveriver: I don't know to be honest. `estimateEntityCount` takes its value from this method each frame before spawning happens, which is what suggested to me that it should be a float.

src/general/StageBase.cs
Oliveriver: I had a lot of trouble with the linter with this line and those around it, and this was the only configuration I found that passed. So please let me know if there is an alternative that will also pass.

Oliveriver:
> There aren't that many major problems here other than the fact that the settings need some code to migrate, because if someone had the entity limit to tiny before this change, they are going to have only 1/4 of the intended minimum spawn amount, and _we'll_ get bug reports about this a lot unless fixed. If you don't have time I can do that part myself.

I'm afraid I almost certainly won't have time for that considering this is an area I'm unfamiliar with.

src/microbe_stage/SpawnSystem.cs
hhyyrylainen: I commented on this line, but my comment was more general. In my opinion the game crashing etc. unstability can be caused by how many things are despawned, so I'd prefer to use an int limit for despawning, but this method should still return a float with the currently alive weighted entity count.

src/general/StageBase.cs
hhyyrylainen: I think that a manual loop that increments an int and adds a weight value to another variable, would be the only way I can think of that avoids allocating extra memory or looping all of the entities twice.

hhyyrylainen:
I implemented that change. For now I didn't implement a full-blown save settings versioning system, instead opting for separate code checking each setting for valid values.

I didn't fix the options menu defaulting to showing "normal" when the selected value doesn't match one of the defined values, as I think that's minor enough to not warrant fixing.

src/microbe_stage/SpawnSystem.cs
hhyyrylainen: I've changed this and the other comment I still had. I think I may have also fixed #3686 as well as I noticed a problem with the break behaviour in the despawn loop.

hhyyrylainen (approved):
Did the remaining changes I wanted. May have also fixed an unrelated bug.

Issue #289, opened 2015-02-16
Add creation loading panel to gui and code


jjonj:
The load button in the editor current does nothing.

tjblazer85:
Hey jjonj, I'm on summer vacation and I wanted to help more on Thrive, Is this thread asking for a loading screen Art?

Moopli:
Good question -- no it isn't. To clarify, this issue is asking for a modal to pop up when you click load in the editor (but only when accessed from main menu, loading a creature isn't available otherwise) which lists available microbes to load and play around with in editor. Eventually, this load button will take you to the Thrivepedia (or Thrive Content Library, or whatever) so you could load up whatever).

tjblazer85:
Ok, thanks for the reply

jjonj:
part of it is actually art, but I think oliveriver might be working on it

Oliveriver:
Yeah, I'm a bit busy at the moment but it's top of my Thrive-related priorities when I get back.

jjonj:
I now implemented saving/loading microbes using very ugly temporary gui. That should make sure we don't forget about this

hhyyrylainen:
Perhaps related to this we could have a feature I've seen people on reddit being interested multiple times: saving creatures as png files? I think including a JSON dump in the metadata or after the image data as trailing bytes in the file. If we make this feature it *might* increase interest in Thrive.

Issue #3283, opened 2022-04-29
Save files should store the mods and versions they have enabled


hhyyrylainen:
so that we can show a warning if a player loads a save that has different mods as some mods probably will make saves incompatible

hhyyrylainen:
`SaveHelper.QuickLoad` should refuse to load a save that has different list of mods enabled.

hhyyrylainen:
This is a duplicate.

PR #3755, opened 2022-08-30
Thriveopedia and Fossilisation

Oliveriver:

Adds a rudimentary Thriveopedia and the ability to fossilise species.

The Thriveopedia currently contains:

* A home page
* A museum containing all fossilised species
* A current world page displaying world generation settings for the current world
* The patch map for the current world
* The evolutionary tree for the current world

<details>
<summary>Open for screenshots</summary>

</details>

**Related Issues**

Closes https://github.com/Revolutionary-Games/Thrive/issues/289

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Oliveriver:
Changing this to a proper PR since I've finally added all the functionality I wanted to add. That's not to say this is done - there are still several bugs to fix detailed below and a whole load of cleanup/linter appeasement required - but it's at least in a state where people can reliably test it. I hope.

Remaining known bugs I'll fix in this PR:

* Gaps between panels in the Thriveopedia aren't consistent. (Easy)
* Fossilisation requires a new icon, as does expanding/collapsing the page tree. (Easy)
* If a species is fossilised in the current game, the museum in the pause menu doesn't update. (Easy)
* Various minor margin width changes. (Easy, but tedious)
* Museum card border highlights stick around after hovering. (Easy, probably)
* Saves are currently broken as the species history doesn't serialize properly. (Medium)
* The number of columns in the museum doesn't update to reflect the expansion/contraction of the page tree. (Medium)
* The evolutionary tree sometimes fails to build for no apparent reason. I suspect this is due to a species going extinct. (Hard)
* A very strange bug sometimes occurs when loading a fossilised AI-created species in freebuild. The ATP production bar and tooltip break due to missing data, causing a crash. Any leads here would be welcome. (Hard)

Remaining known bugs and enhancements I'm proposing leaving out of this PR as out of scope:

* The evolutionary tree fails to build in freebuild games.
* The evolutionary tree fails to build when loading a save made before this PR.
* The player species node in the evolutionary tree always updates to reflect the present build. Ideally, there should be one node per generation showing its progression across time.
* We could add Primum Thrivium as an already existing fossilised species to demonstrate the museum's function.
* If the player moves patches during an editor session, the patch map in the Thriveopedia doesn't update until they leave the editor. Since the Thriveopedia gets this information from the GameProperties object and this isn't updated until the player exits the editor, I don't have any idea how to fix this. It is very minor though.
* Add fossilisation buttons to species in the evolutionary tree and editor.

buckly90:
Great work! The Thriveopedia is looking sleek and clean, not that there is much to clutter it.

**Issues Found:**
1. Hitting the home button brings you to the home page as intended, but the page list still highlights the previous page as if it is selected. You cannot return to the previous page until you select another, and then select it again when this happens.
2. The close button doesn't seem to work at all if you access the pedia from the main menu. It works in-game though!

**Suggestions**:
1. If possible, replacing the fossilize icon over a species with a different one when they have already been fossilized would be nice. Just changing the color could work too.

Oliveriver:
> 1. Hitting the home button brings you to the home page as intended, but the page list still highlights the previous page as if it is selected. You cannot return to the previous page until you select another, and then select it again when this happens.

Good spot. I'll see if I can fix that.

> 2. The close button doesn't seem to work at all if you access the pedia from the main menu. It works in-game though!

I'm pretty sure this is the intermittent Godot connection issue that also sometimes happens with other screens in the main menu. I can't reproduce it at the moment.

> 1. If possible, replacing the fossilize icon over a species with a different one when they have already been fossilized would be nice. Just changing the color could work too.

Good idea.

Oliveriver:
Update: I've now fixed all the bugs from my reply above and addressed the issues Buckly found, _except_ the tree breaking due to a species going extinct. Saves from both this branch and master appear to be loading correctly.

Tomorrow I'll tidy everything up and try to fix the broken tree issue if I have time. Then this should finally be ready for review.

Oliveriver:
Right. I believe this is finally ready for review. I found and fixed a few more bugs in the page tree while cleaning things up, and according to all my testing since, everything works, aside from two things I've decided I need help with.

* The evolutionary tree in the Thriveopedia breaks whenever a species goes extinct from the world. I had a quick look into this but couldn't find anything obvious. There will be people who know more than me about this though.

* Not a bug per se, but definitely a suboptimal way of doing things. I had real trouble saving and loading the RunResults needed to build the evolutionary tree. It turns out JSON can't properly serialize dictionaries whose keys are custom types, and there's a dictionary with Patch objects as keys I needed to save. I found an existing dictionary converter in the Thrive code, but that seemingly only works for static types (compounds, enzymes) and not types such as patches, which are generated dynamically and can change. My eventual solution was to convert the dictionary to and from a list of key-value pairs on saving and loading in RunResults itself. Why not just use the list of key-value pairs everywhere instead, you ask? I tried that and ended up with so many changes in other files that I ran a serious risk of breaking something completely unrelated to this PR. In summary, if someone can write a dictionary converter for patches, I'd be very grateful.

src/general/GameWorld.cs
```diff
@@ -47,10 +51,21 @@ public class GameWorld : ISaveLoadable
...
     /// </summary>
     /// <param name="settings">Settings to generate the world with</param>
-    public GameWorld(WorldGenerationSettings settings) : this()
+    /// <param name="startingSpecies">Starting species for the player</param>
+    public GameWorld(WorldGenerationSettings settings, Species? startingSpecies = null) : this()
     {
         WorldSettings = settings;
-        PlayerSpecies = CreatePlayerSpecies();
+
+        if (startingSpecies == null)
```
84634E1A607A, line 59: Good point. Can apply to Auto-evo Exploring Tool as well. - New Issue?

84634E1A607A:
image
Yet another bug I've created.

src/gui_common/thriveopedia/ThriveopediaEvolutionaryTreePage.cs
```diff
@@ -0,0 +1,158 @@
...
+            speciesHistoryList.Clear();
+
+            evolutionaryTree.Init(CurrentGame.GameWorld.PlayerSpecies);
+            InitFirstGeneration();
+
+            // A possible next step would be to rebuild only when the Thriveopedia as a whole is opened.
+            foreach (var generation in CurrentGame.GameWorld.GenerationHistory)
+            {
+                var record = generation.Value;
+                evolutionaryTree.UpdateEvolutionaryTreeWithRunResults(
```
84634E1A607A, line 97: It looks like my previous comment has something to do with this. It looks like this makes Tree believe that a species is split from another of the same generation.

src/general/GameWorld.cs
hhyyrylainen: Well isn't the case in the exploring tool that there is *always* the player species present? Which we determined to be actually required to make the "enter game world" button to work sensibly there as there needs to be a player species to control.

hhyyrylainen:
> It turns out JSON can't properly serialize dictionaries whose keys are custom types, and there's a dictionary with Patch objects as keys I needed to save. I found an existing dictionary converter in the Thrive code, but that seemingly only works for static types (compounds, enzymes) and not types such as patches, which are generated dynamically and can change.

`Species` class is used as dictionary keys and setup to be auto string serialized and converted back. It's been a while since I've looked at that code but getting that same system to work for `Patch` class should be easy enough. If you can't figure it out I can use some of my time to re-familiarize myself with this system and figure out the parts you need to set. It might be as simple as adding `[TypeConverter(typeof(ThriveTypeConverter))]` to the `Patch` class.

src/general/GameWorld.cs
84634E1A607A: Well my first design was that you can choose any species and enter the game as it.

Thrive.csproj
```diff
@@ -140,6 +140,13 @@
     <Compile Include="src\gui_common\DismissibleNotice.cs" />
     <Compile Include="src\gui_common\FocusFlowDynamicChildrenHelper.cs" />
     <Compile Include="src\gui_common\PhotographablePreview.cs" />
+    <Compile Include="src\gui_common\thriveopedia\Thriveopedia.cs" />
```
hhyyrylainen, line 143: I think as the thriveopedia is a pretty big feature by itself (maybe even as big as some editors) that it would kind of make sense for it to be not under gui_common folder, instead being directly under "src", as it's not really a common GUI component building block. And the fossilization folder could also then put under the thriveopedia folder to group those interrelated functionalities together.

Thrive.csproj
```diff
@@ -466,6 +477,8 @@
     <Compile Include="src\modding\NewModGUI.cs" />
     <Compile Include="src\modding\WorkshopData.cs" />
     <Compile Include="src\saving\FileHelpers.cs" />
+    <Compile Include="src\saving\FossilisedSpecies.cs" />
+    <Compile Include="src\saving\FossilisedSpeciesInformation.cs" />
```
hhyyrylainen, line 481: Specific saveable types shouldn't be in the saving folder but once again in the thriveopedia subfolder I suggested above.

src/general/WorldGenerationSettings.cs
```diff
@@ -32,14 +34,22 @@ public WorldGenerationSettings()
 
     public enum LifeOrigin
     {
+        [Description("LIFE_ORIGIN_VENTS")]
```
hhyyrylainen, line 37: What are these? I don't remember seeing these before...

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -167,6 +167,17 @@ public void Init(Species luca)
...
     }
 
+    public void Clear()
+    {
+        speciesOrigin.Clear();
+        speciesNames.Clear();
+        generationTimes.Clear();
+        speciesNodes.Clear();
+
+        foreach (Node child in tree.GetChildren())
```
hhyyrylainen, line 177: There's a helper method already in the game to free all children.

src/general/GameWorld.cs
hhyyrylainen: So it does make sense in general, then to fix the case with no starting species.

src/auto-evo/GenerationRecord.cs
```diff
@@ -0,0 +1,37 @@
...
+        ///   Total in-game time elapsed since its beginning.
+        /// </summary>
+        [JsonProperty]
+        public double TimeElapsed;
+
+        /// <summary>
+        ///   Auto-Evo results for this generation.
+        /// </summary>
+        [JsonProperty]
+        public RunResults AutoEvoResult = null!;
```
hhyyrylainen, line 29: In this case as this is a normal C# class I'd prefer to have a constructor that takes in the AutoEvoResults and AllSpecies, that way the code doesn't need null suppression and is much more likely to be used correctly without causing null bugs.

You can see the Patch and region and the map classes on how I handled the JSON constructors etc. there.

Oliveriver:
> `Species` class is used as dictionary keys and setup to be auto string serialized and converted back. It's been a while since I've looked at that code but getting that same system to work for `Patch` class should be easy enough. If you can't figure it out I can use some of my time to re-familiarize myself with this system and figure out the parts you need to set. It might be as simple as adding `[TypeConverter(typeof(ThriveTypeConverter))]` to the `Patch` class.

I definitely tried adding `[TypeConverter(typeof(ThriveTypeConverter))]` to `Patch` at some point and it didn't work. Can't remember exactly what went wrong though and I may have made a mistake somewhere, so it's worth another try.

src/general/WorldGenerationSettings.cs
Oliveriver: An approach for translating enum values (see the new methods at the end of this file). There's an example in `SaveInformation` already.

src/general/WorldGenerationSettings.cs
```diff
@@ -124,4 +134,48 @@ public override string ToString()
...
+            TranslationHelper.TranslateBool(LimitReproductionCompoundUseSpeed));
+    }
+
+    /// <summary>
+    ///   Generates a formatted string containing translated planet details.
+    /// </summary>
+    public string GetTranslatedPlanetString()
+    {
+        return string.Format(CultureInfo.CurrentCulture, TranslationServer.Translate("PLANET_DETAILS_STRING"),
+            TranslationServer.Translate(MapType.GetAttribute<DescriptionAttribute>()?.Description),
```
hhyyrylainen, line 166: Doesn't this pretty much completely fail to fulfill the purpose of this method if the getattribute fails? I think a cleaner approach would be to have translation helper take in a generic enum value (if possible) and if not possible then here using an attribute get overload that throws on failure to find would be at least a bit better.

src/general/WorldGenerationSettings.cs
hhyyrylainen: I see there's already `Description` in the translation extraction script so I suppose this is fine, though this is one more thing for new people to understand about the translation system...

buckly90:
I tried my hardest, but I couldn't seem to get anything to break. Everything functioned perfectly well, superb job!

However I can't help but wonder how fossilization would handle modded entities, such as custom parts. If you fossilize a modded creature, and then remove the mod, would that crash the game when opening the thriveopedia? This could potentially be pretty bad. Unfortunately I don't know of any mods that add parts so I can't really test it.

hhyyrylainen:
I'm not really sure if we should do anything about that. There's about five billion things a mod could do that make thrive crash. I guess that when a mod is uninstalled there's just a few ways to still cause crashes, and there's an issue for save compatibility warning with different mods: https://github.com/Revolutionary-Games/Thrive/issues/3283 Though I'll leave it up to @Oliveriver whether you want to add some guard stuff to show an error when loading incompatible fossilized species instructing the user which folder to delete or reinstall mods.

But on the other hand Thrive version updates could also potentially make fossilized species incompatible, though for that it'd probably be easier to just do the same thing as with saves and put a version number into the fossilized species file and refuse to load non-matching ones.

hhyyrylainen:
> I could add a try-catch block in museum card rendering and leave it at that.

I haven't had time to read all of the code, but unless the catch is where the data is tried to be deserialized or where the species is applied, I don't think a catch is going to actually really do anything...

src/general/WorldGenerationSettings.cs
Oliveriver: I'm struggling to parse exactly what you mean here. I agree it would be better to have an approach that throws if no description is found but I can't figure out what your specific suggestion is.

hhyyrylainen: Modify `GetAttribute` to return `T` and throw if it can't find the value.

Actually it looks like the method exists already before this PR so it might be a bit tricky to modify it depending on where it is used elsewhere, but I'd really like it to look like this:
```
    public static T GetAttribute<T>(this Enum e)
        where T : Attribute
    {
        return GetAttributes<T>(e).First();
    }
```

src/thriveopedia/pages/ThriveopediaPatchMapPage.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System;
+using Godot;
+
+/// <summary>
+///   Thriveopedia page displaying the patch map for the current game.
+/// </summary>
+/// <remarks>
+///   <para>
+///     Note a lot of this functionality is duplicated from PatchMapEditorComponent.
```
hhyyrylainen, line 9: For cleaner code would it be possible to create a new class that is used by composition by both this and PatchMapEditorComponent so that most of the tricky logic could be included there (I also saw that this at least duplicates the seed display scene part, so maybe that new class should be a higher level wrapper scene on the underlying map component to provide the common functionality)?

src/thriveopedia/pages/ThriveopediaHomePage.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Godot;
+
+/// <summary>
+///   Thriveopedia page displaying welcome information and links to websites.
+/// </summary>
+public class ThriveopediaHomePage : ThriveopediaPage
+{
+    public override string PageName => "Home";
+    public override string TranslatedPageName => TranslationServer.Translate("HOME_PAGE");
```
hhyyrylainen, line 9: I think the thriveopedia pages should have "THRIVEOPEDIA_" prefix, and maybe "_TITLE" suffix for all of the page names for super clarity.

src/thriveopedia/pages/ThriveopediaHomePage.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Godot;
+
+/// <summary>
+///   Thriveopedia page displaying welcome information and links to websites.
+/// </summary>
+public class ThriveopediaHomePage : ThriveopediaPage
+{
+    public override string PageName => "Home";
```
hhyyrylainen, line 8: This is like an internal name, right? If so this should be lower camelCase to be consistent with the rest of the game.

src/general/StageHUDBase.cs
```diff
@@ -477,6 +490,9 @@ public override void _Ready()
         sunlight = SimulationParameters.Instance.GetCompound("sunlight");
         temperature = SimulationParameters.Instance.GetCompound("temperature");
 
+        fossilisationButtonLayer = GetNode<Control>(FossilisationButtonLayerPath);
```
hhyyrylainen, line 493: Don't these crash the macroscopic prototype now?

src/general/StageHUDBase.cs
```diff
@@ -716,6 +736,52 @@ public void UpdateEnvironmentalBars(BiomeConditions biome)
         // TODO: pressure?
     }
 
+    /// <summary>
+    ///   Creates and displays a fossilisation button above each on-screen microbe.
+    /// </summary>
+    public void ShowFossilisationButtons()
```
hhyyrylainen, line 742: This should *not* be in the base hud, instead should be abstract here and concrete classes should have this.

src/general/StageHUDBase.cs
```diff
@@ -716,6 +736,52 @@ public void UpdateEnvironmentalBars(BiomeConditions biome)
...
+        fossilisationButtonLayer.QueueFreeChildren();
+    }
+
+    /// <summary>
+    ///   Opens the dialog to a fossilise the species selected with a given fossilisation button.
+    /// </summary>
+    /// <param name="button">The button attached to the microbe to fossilise</param>
+    public void ShowFossilisationDialog(FossilisationButton button)
+    {
+        fossilisationDialog.SelectedSpecies = button.AttachedMicrobe.Species;
```
hhyyrylainen, line 781: I think the fossilization dialog should use the base "Species" instead of referring to microbes as this is in the base HUD which means this needs to work for all future stages as well. I'm fine with if you specifically just put a "Not implemented for later stages" message in the fossilize dialog when opened for a non-microbe species.

src/general/StageHUDBase.cs
```diff
@@ -1390,4 +1456,12 @@ private void UpdatePausePrompt()
             pauseInfo.Visible = false;
         }
     }
+
+    private void UpdateFossilisationButtons()
+    {
+        foreach (FossilisationButton button in fossilisationButtonLayer.GetChildren())
```
hhyyrylainen, line 1462: I'd prefer to have an early return in this method if the fossilizationButtonLayer is not visible.

src/general/StageHUDBase.cs
```diff
@@ -716,6 +736,52 @@ public void UpdateEnvironmentalBars(BiomeConditions biome)
...
+            if (microbe.Species is not MicrobeSpecies)
+                continue;
+
+            var button = FossilisationButtonScene.Instance<FossilisationButton>();
+            button.AttachedMicrobe = microbe;
+            button.Connect(nameof(FossilisationButton.OnFossilisationDialogOpened), this,
+                nameof(ShowFossilisationDialog));
+
+            // Display a faded button with a different hint if the species has been fossilised.
+            var alreadyFossilised = FossilisedSpecies.CreateListOfSaves().Contains(
```
hhyyrylainen, line 756: Calling CreateListOfSaves in this loop sounds like a really bad idea to me... would be better to build that list outside this loop.

src/general/WorldGenerationSettings.cs
```diff
@@ -124,4 +134,48 @@ public override string ToString()
...
+            translatedDifficulty,
+            MPMultiplier,
+            AIMutationMultiplier,
+            CompoundDensity,
+            PlayerDeathPopulationPenalty,
+            TranslationServer.Translate("PERCENTAGE_VALUE").FormatSafe(Math.Round(GlucoseDecay * 100, 1)),
+            OsmoregulationMultiplier,
+            TranslationHelper.TranslateBool(FreeGlucoseCloud),
+            TranslationHelper.TranslateBool(PassiveGainOfReproductionCompounds),
+            TranslationHelper.TranslateBool(LimitReproductionCompoundUseSpeed));
```
hhyyrylainen, line 157: I see now where the TranslateBool is used, so I'd say rename this to "TranslateFeatureFlag" and then the keys would be "FEATURE_ENABLED" and "FEATURE_DISABLED" and maybe even the English text could be, "enabled" and "disabled" to make the output more natural sounding to non-programmers.

src/general/WorldGenerationSettings.cs
```diff
@@ -124,4 +134,48 @@ public override string ToString()
...
+            TranslationServer.Translate(Origin.GetAttribute<DescriptionAttribute>()?.Description),
+            Seed);
+    }
+
+    /// <summary>
+    ///   Generates a formatted string containing translated miscellaneous details.
+    /// </summary>
+    public string GetTranslatedMiscString()
+    {
+        return string.Format(CultureInfo.CurrentCulture, TranslationServer.Translate("MISC_DETAILS_STRING"),
```
hhyyrylainen, line 177:
```suggestion
        return string.Format(CultureInfo.CurrentCulture, TranslationServer.Translate("WORLD_MISC_DETAILS_STRING"),
```

src/thriveopedia/Thriveopedia.cs
```diff
@@ -0,0 +1,338 @@
...
+
+            selectedPage = value;
+            selectedPage.Show();
+
+            // The home page is always the first in the history, so ignore it
+            backButton.Disabled = pageHistory.Count <= 1;
+            forwardButton.Disabled = pageFuture.Count == 0;
+
+            pageTitle.Text = SelectedPage.TranslatedPageName;
+            SelectInTreeWithoutEvent(selectedPage);
```
hhyyrylainen, line 91: These seem like things that also happen when next / previous is pressed. So maybe this should be extracted as a separate method?

src/thriveopedia/Thriveopedia.cs
```diff
@@ -0,0 +1,338 @@
...
+            // We only want to initialise these pages once per game
+            if (value == currentGame)
+                return;
+
+            currentGame = value;
+
+            // Add all pages associated with a game in progress
+            if (currentGame != null)
+            {
+                AddPage("CurrentWorld");
```
hhyyrylainen, line 113: Related to my earlier comment about "Home" capitalization, is it actually a node name? If so then it is fine as is, and well to avoid my future confusion you could add a comment on that base property.

src/thriveopedia/Thriveopedia.cs
```diff
@@ -0,0 +1,338 @@
...
+        // By default, assume we're navigating to this page normally
+        ChangePage(pageName, true, true);
+    }
+
+    /// <summary>
+    ///   Gets an existing page by name, or null if no page exists with that name.
+    /// </summary>
+    /// <param name="name">The name of the desired page</param>
+    /// <returns>The Thriveopedia page with the given name</returns>
+    private ThriveopediaPage? GetPage(string? name)
```
hhyyrylainen, line 202: Why can the name to look for be null?

src/thriveopedia/Thriveopedia.cs
```diff
@@ -0,0 +1,338 @@
...
+    /// <summary>
+    ///   Adds a page to the Thriveopedia
+    /// </summary>
+    /// <param name="name">The name of the page</param>
+    /// <param name="parentName">The name of the page's parent if applicable</param>
+    private void AddPage(string name, string? parentName = null)
+    {
+        // Avoid adding duplicate pages
+        if (allPages.Keys.Any(p => p.PageName == name))
+            return;
```
hhyyrylainen, line 216: Shouldn't it be a serious error to try to add a duplicate page or?

src/thriveopedia/Thriveopedia.cs
```diff
@@ -0,0 +1,338 @@
...
+        // Create and hide a blank root to avoid home being used as the root
+        pageTree.CreateItem();
+        pageTree.HideRoot = true;
+
+        // Keep a special reference to the home page
+        homePage = GetNode<ThriveopediaHomePage>(HomePagePath);
+        allPages.Add(homePage, CreateTreeItem(homePage, null));
+
+        // Add all pages not associated with a game in progress
+        AddPage("Museum");
```
hhyyrylainen, line 142: I assume this doesn't take too much time because otherwise we should think about only initializing the Thriveopedia when it becomes visible.

src/thriveopedia/Thriveopedia.cs
```diff
@@ -0,0 +1,338 @@
...
+    private void OnClosePressed()
+    {
+        GUICommon.Instance.PlayButtonPressSound();
+
+        Exit();
+    }
+
+    private void Exit()
+    {
+        // Reset to the home page for the next session
```
hhyyrylainen, line 330: When opened in-game it would be super useful if the history was kept between opening and closing the thriveopedia.

src/thriveopedia/ThriveopediaPage.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            UpdateCurrentWorldDetails();
+        }
+    }
+
+    public override void _Ready()
+    {
+        base._Ready();
+
+        backgroundPanel = GetNode<PanelContainer>(".");
```
hhyyrylainen, line 50: Isn't background panel then just "this"? So on line 54 it should just be "this.AddStyle..."?

src/thriveopedia/fossilisation/FossilisationButton.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        camera = GetViewport().GetCamera();
+    }
+
+    /// <summary>
+    ///   Update the position of this button, e.g. to reflect player zooming.
+    /// </summary>
+    public void UpdatePosition()
+    {
+        RectGlobalPosition = camera.UnprojectPosition(AttachedMicrobe.GlobalTransform.origin);
```
hhyyrylainen, line 53: I think this should have a guard here against the camera no longer being current similarly to how DebugOverlays does it.

src/thriveopedia/fossilisation/FossilisationButton.cs
```diff
@@ -0,0 +1,61 @@
...
+/// </summary>
+public class FossilisationButton : TextureButton
+{
+    [Export]
+    public Texture AlreadyFossilisedTexture = null!;
+
+    /// <summary>
+    ///   The microbe this button is attached to.
+    /// </summary>
+    public Microbe AttachedMicrobe = null!;
```
hhyyrylainen, line 14: From looking at things, it should be possible to change this to `Spatial` as this doesn't seem to need to know anything about the species itself.

src/thriveopedia/fossilisation/FossilisationDialog.cs
```diff
@@ -0,0 +1,232 @@
...
+    private CustomConfirmationDialog overwriteNameConfirmationDialog = null!;
+
+    /// <summary>
+    ///   The species currently open in the dialog.
+    /// </summary>
+    private Species selectedSpecies = null!;
+
+    /// <summary>
+    ///   True when one of our (name related) Controls is hovered. This needs to be known to know if a click happened
+    ///   outside the name editing controls, for detecting when the name needs to be validated.
```
hhyyrylainen, line 42: Can a lose focus callback from the name box be used instead? As that would work for controller input as well (and keyboard navigation).

src/thriveopedia/fossilisation/FossilisationDialog.cs
```diff
@@ -0,0 +1,232 @@
...
+
+    /// <summary>
+    ///   The species currently open in the dialog.
+    /// </summary>
+    public Species SelectedSpecies
+    {
+        get => selectedSpecies;
+        set
+        {
+            selectedSpecies = (Species)value.Clone();
```
hhyyrylainen, line 54: This clone here is a bit unexpected to me to be here, as basically always you assume a property set actually *sets* the value. Would it be possible to refactor so to keep this as-is but then create the cloned species when required?

src/thriveopedia/fossilisation/FossilisationDialog.cs
```diff
@@ -0,0 +1,232 @@
...
+
+    private void UpdateSpeciesDetails()
+    {
+        speciesDetailsLabel.ExtendedBbcode = TranslationServer.Translate("SPECIES_DETAIL_TEXT").FormatSafe(
+            SelectedSpecies.FormattedNameBbCode,
+            SelectedSpecies.ID,
+            SelectedSpecies.Generation,
+            SelectedSpecies.Population,
+            SelectedSpecies.Colour.ToHtml(),
+            string.Join("\n  ", SelectedSpecies.Behaviour.Select(b => b.Key + ": " + b.Value)));
```
hhyyrylainen, line 176: Wasn't there behaviour value translations in the auto-evo exploring tool?

hhyyrylainen: Or is this code exactly the same as in there? If so this is pretty complex so I'd say this should be made into a helper method somewhere (probably the Species class makes most sense).

src/thriveopedia/fossilisation/FossilisationDialog.cs
```diff
@@ -0,0 +1,232 @@
...
+    {
+        Hide();
+    }
+
+    private void OnFossilisePressed()
+    {
+        GD.Print("Saving species " + SelectedSpecies.FormattedName);
+
+        if (FossilisedSpecies.CreateListOfSaves()
+            .Any(s => s == SelectedSpecies.FormattedName + Constants.FOSSIL_EXTENSION_WITH_DOT))
```
hhyyrylainen, line 201: I'm seeing this logic for the second time so I'd say that FossilizedSpecies should have a method to check by *name* if something is fossilized already (and it could take an optional fossils list if called in a loop for example for more efficiency).

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -0,0 +1,243 @@
...
+using Directory = Godot.Directory;
+using File = Godot.File;
+using Path = System.IO.Path;
+
+/// <summary>
+///    A species saved by the user. Contains helper methods for saving/loading species on the disk.
+/// </summary>
+public class FossilisedSpecies
+{
+    public const string SAVE_SAVE_JSON = "save.json";
```
hhyyrylainen, line 19:
```suggestion
    public const string SAVE_FOSSIL_JSON = "fossil.json";
```
I'm going to suggest a bunch of changes to make this more distinct from the normal saves.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -0,0 +1,243 @@
...
+
+    /// <summary>
+    ///   General information about this saved species.
+    /// </summary>
+    public FossilisedSpeciesInformation Info { get; set; } = new();
+
+    /// <summary>
+    ///   The species to be saved/loaded.
+    /// </summary>
+    public Species Species { get; set; } = null!;
```
hhyyrylainen, line 35: I really think you should add a constructor to force people to set this right as in my opinion this use of nullability overwrite is an unnecessary way to get around it.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -0,0 +1,243 @@
...
+
+    /// <summary>
+    ///   Creates a list of existing fossilised species names to prevent unintended overwrites.
+    /// </summary>
+    /// <returns>A list of names of .thrivefossil files in the user's home directory</returns>
+    public static List<string> CreateListOfSaves()
```
hhyyrylainen, line 41:
```suggestion
    public static List<string> CreateListOfFossils()
```

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -0,0 +1,243 @@
...
+
+    /// <summary>
+    ///   The species to be saved/loaded.
+    /// </summary>
+    public Species Species { get; set; } = null!;
+
+    /// <summary>
+    ///   Creates a list of existing fossilised species names to prevent unintended overwrites.
+    /// </summary>
+    /// <returns>A list of names of .thrivefossil files in the user's home directory</returns>
```
hhyyrylainen, line 40:
```suggestion
    /// <returns>A list of names of .thrivefossil files in the user's fossils directory</returns>
```

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -0,0 +1,243 @@
...
+
+                result.Add(filename);
+            }
+
+            directory.ListDirEnd();
+        }
+
+        using var file = new File();
+        result = result.OrderBy(item =>
+            file.GetModifiedTime(Path.Combine(Constants.FOSSILISED_SPECIES_FOLDER, item))).ToList();
```
hhyyrylainen, line 75: Do these really need to be sorted by time? Well at least when checking if a fossil exist, this should be skipped then.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -0,0 +1,243 @@
...
+
+        return result;
+    }
+
+    /// <summary>
+    ///   Loads a fossilised species by its filename.
+    /// </summary>
+    /// <param name="saveName">The name of the .thrivefossil file (including extension)</param>
+    /// <returns>The species saved in the provided file</returns>
+    public static Species LoadSpeciesFromFile(string saveName)
```
hhyyrylainen, line 85:
```suggestion
    public static Species LoadSpeciesFromFile(string fossilName)
```

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -0,0 +1,243 @@
...
+    ///   Saves this species to disk.
+    /// </summary>
+    public void SaveToFile()
+    {
+        if (Species is not MicrobeSpecies)
+        {
+            throw new NotImplementedException("Saving non-microbe species is not yet implemented");
+        }
+
+        // For now, save everything as a microbe to aid deserialization
```
hhyyrylainen, line 103: Well actually due to the dynamic species saving already, it should be possible to really do without the Type etc. fuss here, but unless tested, I think having the NotImplementedException here is still a good idea.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -0,0 +1,243 @@
...
+        var speciesInfo = new FossilisedSpeciesInformation
+        {
+            Type = FossilisedSpeciesInformation.SpeciesType.Microbe,
+        };
+
+        WriteRawSaveDataToFile(speciesInfo, Species.StringCode, Name + Constants.FOSSIL_EXTENSION_WITH_DOT);
```
hhyyrylainen, line 109: Is there a reason why the image is not saved in the fossil. I think it would make this much more helpful format if the screenshot of the thing would also be saved. That'd make the museum load much faster, right?

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    private static void WriteRawSaveDataToFile(FossilisedSpeciesInformation speciesInfo, string saveContent,
```
hhyyrylainen, line 112: This should also be renamed, I'll now stop pointing out all the instances of the word "save" in this file...

src/thriveopedia/fossilisation/FossilisedSpeciesInformation.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// </summary>
+    public DateTime CreatedAt { get; set; } = DateTime.Now;
+
+    /// <summary>
+    ///   ID for this fossilised species.h
+    /// </summary>
+    public Guid ID { get; set; } = Guid.NewGuid();
+
+    /// <summary>
+    ///   The type of this species, e.g. microbe.
```
hhyyrylainen, line 42: Now that I think about this, I agree it's a good idea to have the stage here, so that museum can be filtered for example without having to load all of the data for each species.

src/thriveopedia/fossilisation/MuseumCard.cs
```diff
@@ -0,0 +1,58 @@
...
+    [Export]
+    public NodePath SpeciesNameLabelPath = null!;
+
+    [Export]
+    public NodePath SpeciesPreviewPath = null!;
+
+    /// <summary>
+    ///   The fossilised species associated with this card.
+    /// </summary>
+    public Species SavedSpecies = null!;
```
hhyyrylainen, line 17: This should be a nullable property that on change would also reload the preview. Similarly to a bunch of our GUI nodes which either allow setting a value before or after attaching them to a scene.

src/thriveopedia/pages/ThriveopediaEvolutionaryTreePage.cs
```diff
@@ -0,0 +1,158 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Godot;
+
+/// <summary>
+///    Thriveopedia page displaying an evolutionary tree of all species in the current game.
+/// </summary>
+/// <remarks>
+///   <para>
+///     Note a lot of this functionality is duplicated from AutoEvoExploringTool.
```
hhyyrylainen, line 10: Again here, I think we should take steps to avoid code duplication to make sure the exploring tool and thriveopedia don't end up with different implementations of things in the long run.

src/thriveopedia/pages/ThriveopediaEvolutionaryTreePage.cs
```diff
@@ -0,0 +1,158 @@
...
+    public NodePath SpeciesDetailsLabelPath = null!;
+
+    [Export]
+    public NodePath SpeciesPreviewPath = null!;
+
+    [Export]
+    public NodePath HexPreviewPath = null!;
+
+    private readonly List<Dictionary<uint, Species>> speciesHistoryList = new();
+    private VBoxContainer disabledInFreebuild = null!;
```
hhyyrylainen, line 31:
```suggestion
    private readonly List<Dictionary<uint, Species>> speciesHistoryList = new();

    private VBoxContainer disabledInFreebuild = null!;
```
I prefer blank lines between *completely* different categories of variables for easier visual parsing of what's going on here.

src/thriveopedia/pages/ThriveopediaEvolutionaryTreePage.cs
```diff
@@ -0,0 +1,158 @@
...
+    /// <summary>
+    ///   Clear and then rebuild the evolutionary tree each time we open the page. This way, we ensure the tree is
+    ///   always up to date.
+    /// </summary>
+    public void RebuildTree()
+    {
+        if (!Visible)
+            return;
+
+        // TODO fix the tree for freebuild?
```
hhyyrylainen, line 74:
```suggestion
        // TODO: fix the tree for freebuild
```

src/thriveopedia/pages/ThriveopediaEvolutionaryTreePage.cs
```diff
@@ -0,0 +1,158 @@
...
+        if (CurrentGame!.FreeBuild)
```
hhyyrylainen, line 75: I'd say do the longer thing and put a `if(CurrentGame == null) throw new invalidoperationexception` at the top of this method instead of just skip over it like this. I've done that in quite many places in the game.

src/thriveopedia/pages/ThriveopediaEvolutionaryTreePage.cs
```diff
@@ -0,0 +1,158 @@
...
+        {
+            evolutionaryTree.Visible = false;
+            disabledInFreebuild.Visible = true;
+            return;
+        }
+
+        // Building the tree relies on the existence of a full history of generations stored in the current game. Since
+        // we only started adding these in 0.6.0, it's impossible to build a tree in older saves.
+        // TODO avoid an ugly try/catch block by actually checking the original save version?
```
hhyyrylainen, line 84: Doing the TODO here would be a pretty good idea

src/thriveopedia/pages/ThriveopediaEvolutionaryTreePage.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                var record = generation.Value;
+                evolutionaryTree.UpdateEvolutionaryTreeWithRunResults(
+                    record.AutoEvoResult, record.Generation, record.TimeElapsed);
+                speciesHistoryList.Add(record.AllSpecies);
+            }
+        }
+        catch (KeyNotFoundException)
+        {
+            evolutionaryTree.Visible = false;
```
hhyyrylainen, line 104: Or at least print a TODO error here about the problem. And I haven't looked at the text but I hope the disabledInFreebuild error message also says that it may be because of an older save?

src/thriveopedia/pages/ThriveopediaEvolutionaryTreePage.cs
```diff
@@ -0,0 +1,158 @@
...
+            disabledInFreebuild.Visible = true;
+        }
+    }
+
+    private void InitFirstGeneration()
+    {
+        // TODO fix this so it shows player species progression rather than current state
```
hhyyrylainen, line 111:
```suggestion
        // TODO: fix this so it shows player species progression rather than current state
```

src/thriveopedia/pages/ThriveopediaEvolutionaryTreePage.cs
```diff
@@ -0,0 +1,158 @@
...
+        speciesHistoryList.Add(new Dictionary<uint, Species>
+        {
+            { CurrentGame!.GameWorld.PlayerSpecies.ID, CurrentGame.GameWorld.PlayerSpecies },
```
hhyyrylainen, line 114: This nullability suppression is more acceptable to me than the one above.

src/thriveopedia/pages/ThriveopediaEvolutionaryTreePage.cs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    private void EvolutionaryTreeNodeSelected(int generation, uint id)
+    {
+        UpdateSpeciesPreview(speciesHistoryList[generation][id]);
+    }
+
+    private void UpdateSpeciesDetail(Species species)
+    {
+        speciesDetailsLabel.ExtendedBbcode = TranslationServer.Translate("SPECIES_DETAIL_TEXT").FormatSafe(
```
hhyyrylainen, line 141: This is also probably shared code that'd be easy to put in the microbe class, right?

src/thriveopedia/pages/ThriveopediaMuseumPage.cs
```diff
@@ -0,0 +1,178 @@
...
+        {
+            // For now, rebuild the card list entirely each time we open the page. Very unoptimised, but it keeps
+            // the museum up to date with the player's new fossilisations in a game. A possible next step would be
+            // to rebuild only when the Thriveopedia as a whole is opened.
+            cardContainer.QueueFreeChildren();
+
+            foreach (var speciesName in FossilisedSpecies.CreateListOfSaves())
+            {
+                var card = (MuseumCard)GD.Load<PackedScene>(
+                    "res://src/thriveopedia/fossilisation/MuseumCard.tscn").Instance();
```
hhyyrylainen, line 69: Packed scene should be loaded once in _Ready

src/thriveopedia/pages/ThriveopediaMuseumPage.cs
```diff
@@ -0,0 +1,178 @@
...
+
+                var savedSpecies = FossilisedSpecies.LoadSpeciesFromFile(speciesName);
+
+                if (savedSpecies is not MicrobeSpecies)
+                {
+                    throw new NotImplementedException("Loading non-microbe species is not yet implemented");
+                }
```
hhyyrylainen, line 76: I think this throw is a bit much here, I think the card should just reduce the available functions it has if given another type of species.

src/thriveopedia/pages/ThriveopediaMuseumPage.cs
```diff
@@ -0,0 +1,178 @@
...
+        {
+            GD.PrintErr("Unknown species type to preview: ", species);
+        }
+
+        UpdateSpeciesDetail(species);
+    }
+
+    private void UpdateSpeciesDetail(Species species)
+    {
+        speciesDetailsLabel.ExtendedBbcode = TranslationServer.Translate("SPECIES_DETAIL_TEXT").FormatSafe(
```
hhyyrylainen, line 127: Now *this* is definitely duplicate code.

src/thriveopedia/pages/ThriveopediaMuseumPage.cs
```diff
@@ -0,0 +1,178 @@
...
+    {
+        GUICommon.Instance.PlayButtonPressSound();
+
+        if (speciesPreview.PreviewSpecies == null)
+            return;
+
+        // If we're opening from a game in progress, warn the player
+        if (CurrentGame != null)
+        {
+            leaveGameConfirmationDialog.DialogText = TranslationServer.Translate("OPEN_IN_FREEBUILD_WARNING");
```
hhyyrylainen, line 155:
```suggestion
            leaveGameConfirmationDialog.DialogText = TranslationServer.Translate("OPEN_FOSSIL_IN_FREEBUILD_WARNING");
```

src/thriveopedia/pages/ThriveopediaMuseumPage.cs
```diff
@@ -0,0 +1,178 @@
...
+            leaveGameConfirmationDialog.PopupCenteredShrink();
+            return;
+        }
+
+        TransitionToFreebuild();
```
hhyyrylainen, line 160: I'd say add the previewSpecies as parameter here, then the nullability suppression in that method is not necessary for cleaner code.

hhyyrylainen (changes requested):
I don't have any real major objections to any of the stuff here, just a bunch of small problems I noticed. I didn't playtest yet but I suspect this breaks the macroscopic prototype entirely, so that at least should be looked at.

src/thriveopedia/pages/ThriveopediaEvolutionaryTreePage.cs
Oliveriver: Is there a good way to go about getting this information? I had a look into this but found no leads.

hhyyrylainen: Well we would actually need a new field in the save info to preserve this info... so kind of probably not worth it (but overall wouldn't be too hard to add, but kind of the point of the save upgrade system is that old saves work as well as new ones). Maybe it'd be just better then to add proper error printing and showing right info to the player (like my next comment suggests).

src/general/WorldGenerationSettings.cs
Oliveriver: The only other place this is used throws if it returns null anyway, so this should be fine.

src/thriveopedia/Thriveopedia.cs
Oliveriver: I think I tried this as a fix for a bug I found with the page tree, found a better fix, and accidentally left this in. So yeah, I'll remove it.

src/thriveopedia/pages/ThriveopediaHomePage.cs
Oliveriver: It's used as part of the name for the scene loaded for each page, so needs to be PascalCase there.

hhyyrylainen: I did eventually get that far in the code. Saying that explicitly in the base abstract class xml documentation would be nice.

src/thriveopedia/Thriveopedia.cs
Oliveriver: If a page is created with no parent, this method gets called with null and returns null. I suppose when you put it like that it's better to avoid this path and check for null earlier.

src/thriveopedia/Thriveopedia.cs
Oliveriver: Right now it doesn't, since for the museum in particular the cards are only created when it becomes visible.

It may take longer if the list of pages grows a lot bigger.

src/thriveopedia/fossilisation/FossilisationDialog.cs
Oliveriver: I believe I copied this wholesale from EditorComponentBottomLeftButtons, so if this is changed it's probably best to change that too.

src/thriveopedia/fossilisation/FossilisationDialog.cs
Oliveriver: It's occured to me that this begs a design choice I'd implicitly answered and ignored: if a player fossilises a species with a new name, should its name change in the current game too? I added this to stop that happening, but we do have the option of using this as a method for renaming species as people have been asking for.

src/thriveopedia/pages/ThriveopediaEvolutionaryTreePage.cs
Oliveriver: Yes, maybe that's the wrong name as it covers both scenarios. I'll update that.

Oliveriver:
Addressed the many small things. I'll get to the rest tomorrow.

src/thriveopedia/fossilisation/FossilisationDialog.cs
hhyyrylainen: It does seem like it, so that would be good to update as well. I just searched for "OnClickedOffName" and it seems like you didn't attach a callback to this method, so it would seem to be that this cannot currently function correctly.

src/thriveopedia/fossilisation/FossilisationDialog.cs
hhyyrylainen: I'd disallow that. The intended future feature is anyway to attach labels to things the player names, not replace their names entirely.

src/thriveopedia/pages/ThriveopediaPatchMapPage.cs
Oliveriver: I've just spent several hours trying to get this to work and gave up. There's so much interdependence in the editor classes that I couldn't separate neatly.

I got it working but saves were broken, and any attempt to fix it created horrid spaghetti code. Honestly making such a major change to the editor would introduce so much more risk for bugs and extra review requirements that I doubt it's a good idea for me to do it here.

If you want to fix it go ahead, but I don't think I can.

src/thriveopedia/fossilisation/FossilisationDialog.cs
Oliveriver: Having looked into this, I think I can get away with removing it. The editor blocks any other input until the name is valid which is why this is needed there, but I don't need that here as I'm just disabling the fossilisation button when the name is invalid and the player is free to perform other actions.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
Oliveriver: The reason is I didn't know how to. Just tried it and it would take a significant refactor of the architecture, so I don't think it's worth it for now. There are other, easier ways to get better performance with the museum (like only loading when the Thriveopedia is opened, not when the museum page is).

Oliveriver:
Right, addressed a few more comments and also improved the museum performance by loading only when the Thriveopedia is opened. For my own benefit, here's what's left:

* Fix the tree failing to build if a species goes extinct. Again, I ran out of time to look into this.
* Extract a common class for the editor patch map and Thriveopedia patch map. After spending several hours on it going nowhere, I've decided I refuse to do it.
* Extract a common class for the exploring tool evolutionary tree and Thriveopedia evolutionary tree. I suspect this will be similar.
* Save images with fossils. Tried it and it was too complex for my brain after all the above. I think it can be safely delegated to an issue to address in future.
* Simplify the dynamic JSON deserialization of species type, which hhyyrylainen will be looking into.

Nearly there...

src/thriveopedia/fossilisation/FossilisationDialog.cs
hhyyrylainen: That sounds fine.

src/thriveopedia/pages/ThriveopediaPatchMapPage.cs
hhyyrylainen: Sounds like a pain. I feel like someone has finally experienced the pain I had with trying to setup the later stage scenes when I needed to refactor almost everything in the game and HUD scenes...

We can open a follow up issue for this, though until that is done people need to be careful not to accidentally make changes in only one place the map is shown.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
hhyyrylainen: Doesn't the fossilization dialog show a preview image of the species? That can be grabbed quite easily I think. I'll give that a go with the other saving fixes I'll try to do today.

hhyyrylainen:
I found a bug, the overwrite dialog has the title "Confirm Exit":

src/thriveopedia/fossilisation/FossilisedSpecies.cs
hhyyrylainen: I've now done this and about to commit. I had to extend the species preview system a bit to get it to save the Images when needed for writing into the fossilized files.

hhyyrylainen:
Another bug:
```
ERROR: Error calling method from signal 'visibility_changed': 'PanelContainer(ThriveopediaPatchMapPage.cs)::RebuildMap': Method not found..
```
I checked that no random scenes are modified so I'm quite sure this is a mistake in this PR.

hhyyrylainen:
Okay from playtesting I noticed just one more slight problem which is that the warning about starting a new game from the museum while in-game just talks about starting a new game, but I think the message should say starting a new game as the selected species, just to be clear what's happening. As I think some players might read starting a new game and not click ok as they are confused and thought that they did something wrong.

hhyyrylainen:
One more bug:
EarlyMulticellularEditor the statistics button doesn't work

It seems the callback exists but Godot editor *insists* on removing that, and when I reverted everything and directly debugged that scene, the button still didn't work. So could you look at that? I think it might be configured wrong somehow. All the other prototypes did have their statistics buttons working correctly.

Oliveriver:
Fixed the small bugs hhyyrylainen found.

Finally figured out what's going on with the broken evolutionary tree. Even though each of the Species in AllSpecies in GenerationRecord is cloned, the Species in AutoEvoResults in GenerationRecord are not. This means the population for every node on each rebuild was always updating to the latest population, not the population at the time. Consequently, when a species went extinct, _all_ nodes for that species were being marked as extinct and so had no non-extinct parent to connect to, causing the KeyNotFoundException.

I've written a fix and will clean it up tomorrow.

hhyyrylainen:
> the Species in AutoEvoResults in GenerationRecord are not

I did notice this, but I thought you specifically did it like that. When creating the AutoEvoResults class I did not intend it to be stored past the current auto-evo cycle, so storing it leading to various problems is not very surprising to me.

Oliveriver:
Fixed the tree breaking due to species extinction. It required a change to how generation records are stored in saves so any saves made in this branch before that might be broken. But I've tested various saves made on master and the newest version of this branch and found no issues.

Pending further comments on that fix (and one tiny extra fix I had to make to stop old saves crashing when trying to draw the tree), I'm going to call this done. There are a few things outstanding, but at this point I think they're separate enough to break into new issues.

* Add a helper to switch button order in confirmation dialogs depending on system.
* Change the tree drawing to add a node for the player species every generation so they can see their own progression. I had a go at adding this but thought it would be better scrutinised in a new PR, so I'll make that once this is merged.
* Separate common parts of PatchMapEditorComponent and ThriveopediaPatchMapPage into a new scene/class.
* Separate common parts of AutoEvoExploringTool and ThriveopediaEvolutionaryTreePage into a new scene/class.
* Fix the evolutionary tree in freebuild games and old saves (though trees from old saves will always be incomplete at the very least).
* Update the Thriveopedia patch map while in the editor if the player moves patches.
* Add fossilisation buttons to species displays in the evolutionary tree and editor.
* Add the ability to delete fossilised species.

hhyyrylainen:
One more thing we can open as separate issue is having a button to open the fossils folder, as I would guess people would like sharing those. People ask enough about how to find the save files.

Oliveriver:
> One more thing we can open as separate issue is having a button to open the fossils folder, as I would guess people would like sharing those. People ask enough about how to find the save files.

Yeah, good point. Could easily fit in the museum page.

I've made those last few changes you suggested, so this is ready to merge if you're happy.

hhyyrylainen:
I'll do a one more full read through of the code to check my comments were addressed and I don't spot any more problems, in a few days. I'll ask now if anyone else can playtest or code review this before that.

84634E1A607A:
Not sure if this is known: When you fossilize a species, the icon will not change to the fossilized state until you resume the game and pause again.

84634E1A607A:
Another bug I've written: player mutations are not taken into account. There are no proper nodes indicaitng that the player has evolved.

image

Oliveriver:
> Edit: which seems to be wanted to be merged into this branch. Is there a reason why that's a separate PR, @Oliveriver ?

I did that so it was easier to review.

I've added the changes which minimise save data size based on the findings on Discord earlier. Essentially, I've done three things:

* Replaced actual species data with species IDs for everything except the main species. Turns out recording the full data for every species a given species split from was adding massive bulk.
* Removed unnecessary fields from the saved objects, like per-patch populations which had many patches stored within them. It may be that we want to add these back in future for new features, but for now they're useless bloat.
* Removed species data for the main species if it didn't change in a given generation. This data is then recreated when building the tree using previous generations.

hhyyrylainen:
> Removed unnecessary fields from the saved objects, like per-patch populations which had many patches stored within them. It may be that we want to add these back in future for new features, but for now they're useless bloat.

I think patches where the population exist, would be good to save, and still saves over 80% compared to saving all of them.

Did you resolve the generation record having the AllSpecies property and then differently cloned species also stored through the auto-evo results? I saw you remove the generation record class but not modify the GameWorld, so I didn't see how you made that work...

Oliveriver:
I didn't remove GenerationRecord, I moved it to a new folder since there are now a few other record classes. Everything that needs to be cloned is, as far as I can tell, still being cloned properly, and species which don't need to be cloned (those that haven't changed) aren't being saved anymore.

hhyyrylainen:
I see. Then I just have one comment that I think the SpeciesRecord should probably be SpeciesRecordLite so that the "full" record is fundamentally the more important version. I thought initially that maybe a base class could be made but that doesn't really seem like it as fundamentally the types are meant for different purposes. Though they seem to share at least 2 fields, right? So an abstract base class could be made to save on the documentation string etc. duplication.

src/thriveopedia/Thriveopedia.cs
```diff
@@ -0,0 +1,359 @@
...
+
+        pageHistory.Push(homePage);
+        SelectedPage = homePage;
+    }
+
+    public override void _Notification(int what)
+    {
+        base._Notification(what);
+
+        if (what == NotificationVisibilityChanged && Visible)
```
hhyyrylainen, line 151: Well we don't have translations yet but I think some places like the navigation tree needs to be regenerated on language change, I'll open a follow up issue for this as well.

src/general/StageBase.cs
```diff
@@ -285,6 +285,9 @@ public virtual void OnReturnFromEditor()
         if (CurrentGame == null)
             throw new InvalidOperationException("Returning to stage from editor without a game setup");
 
+        // Update the generation history with the newly edited player species
+        GameWorld.GenerationHistory.Last().Value.UpdateSpeciesData(GameWorld.PlayerSpecies);
```
hhyyrylainen, line 289: This relies on the dictionary being in-order, which is not 100% guaranteed, so the history should actually be a list or what I'll do myself to fix this right now is to find the max key and use that here.

src/auto-evo/records/SpeciesRecordLite.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    ///   Species mutation and population data from a single generation, with or without the full species.
+    /// </summary>
+    public class SpeciesRecordLite : SpeciesRecord
+    {
+        [JsonConstructor]
+        public SpeciesRecordLite(Species? species,
+            long population,
+            uint? mutatedPropertiesID = null,
+            uint? splitFromID = null)
```
hhyyrylainen, line 15: "chop" formatting of function arguments is not correct style, I'll fix this.

hhyyrylainen (approved):
I did some more playtesting and looked through the code again. I did find some things to fix (and one editor made save exit editor crash bug) but I fixed those myself so now I think this is good to go.

hhyyrylainen:
Opened the follow up issues. I'll highlight just one here: https://github.com/Revolutionary-Games/Thrive/issues/3854 which I marked as high priority because we've historically needed to fix translations not applying in a bunch of places so that'd be nice to fix before anyone complains about the bug.

PR #3769, opened 2022-09-11
Show exact build commit in the game, and show more info in devbuilds

### LINKED ISSUE #1793: Make devbuilds print out the hash when running
to help identify if some log file is created when running a specific commit. I suppose it wouldn't hurt if releases also printed the hash, but those shouldn't show the hash in the main menu or other such visible way.

hhyyrylainen:

Show the build date and commit in the options menu and for devbuild show the exact commit, branch etc. info in the bottom right of the main menu.

**Related Issues**

closes  #1793

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,6 +142,7 @@ package-lock.json
 /Scripts/patrons.json
 /Scripts/translators.json
 /steam_appid.txt
+simulation_parameters/revision.json
 
 /mods
 
--- a/Scripts/BuildInfoWriter.cs
+++ b/Scripts/BuildInfoWriter.cs
@@ -1,6 +1,7 @@
 ﻿namespace Scripts;
 
 using System;
+using System.IO;
 using System.Text.Json.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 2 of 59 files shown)


RacerBG (changes requested):
With this font the main menu text look a little unreadable. Could we use a tooltip of some sort when the mouse is over the version text for example or if this is not possible just change the font?

The options menu is fine but it could use a smaller font for the information on the right.

hhyyrylainen:
> With this font the main menu text look a little unreadable. Could we use a tooltip of some sort when the mouse is over the version text for example or if this is not possible just change the font?

The entire point is that it needs to be very obvious. I've had to guide people so many times to check that they are actually playing the build they think they are playing.

I don't think it's worth the effort to split the version number display into two parts to have a different font size or font for the devbuild part. Also it would not look that consistent to have different font using text right next to each other.

hhyyrylainen:
Opened a follow up issue if anyone is up for some GUI tweaking: https://github.com/Revolutionary-Games/Thrive/issues/3775

PR #3780, opened 2022-09-13
Added patch move visibility property (#3744)

### LINKED ISSUE #3744: Add a property to hide the move to patch button in the patch details panel
for use in the auto-evo exploring tool as it doesn't make sense to always show the disabled "Move to This Patch" button there.

TheHungryBandito:
* Added MoveToPatchButtonVisible property to PatchDetailsPanel.cs for setting and getting of the button's visibility.

This PR adds a boolean property MoveToPatchButtonVisible to PatchDetailsPanel.cs to allow the getting and setting of the "Move to patch button" visibility.

**Related Issues**
closes #3744 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Oliveriver:
Should this PR also include actually using the property in the Auto-Evo tool? If it’s not covered here I’m worried it could get forgotten about.

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -154,6 +154,12 @@ public class PatchDetailsPanel : PanelContainer
 
     public bool IsPatchMoveValid { get; set; }
 
+    public bool MoveToPatchButtonVisible
```
84634E1A607A, line 157: Can be an export variable.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -772,6 +772,8 @@ private void UpdateSpeciesDetail(Species species)
 
     private void UpdatePatchDetailPanel(PatchMapDrawer drawer)
     {
+        patchDetailsPanel.MoveToPatchButtonVisible = false;
```
84634E1A607A, line 775: See my later comment, so that this can be changed directly in Godot editor.

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -154,6 +154,12 @@ public class PatchDetailsPanel : PanelContainer
...
+    {
+        get => moveToPatchButton.Visible;
+        set => moveToPatchButton.Visible = value;
```
84634E1A607A, line 160: I recall there is an HSplit? It should be hidden too.

84634E1A607A (changes requested):
Correct approach, but I have some concerns.

src/microbe_stage/gui/PatchDetailsPanel.cs
TheHungryBandito: I'm having some trouble with the Export attribute.

Problem - With the export attribute present, the property seems to try to set itself before nodes have entered the tree.

What I have tried:
 - Null check return if null (The button and HSplit are always visible the first time regardless of the editor setting but updates properly when changed during runtime)
 - Null check with GetNode to fill the null fields before setting visibility (Null reference error NodePath is null)
 
 I'm still new to Godot so I might just be missing something simple.

src/microbe_stage/gui/PatchDetailsPanel.cs
TheHungryBandito: You are correct, I will adjust it accordingly

src/microbe_stage/gui/PatchDetailsPanel.cs
84634E1A607A: That's known. Our common solution seems to be:

1. Add a corresponding field
2. Set the field when property is set; set visibility if not null
3. Explicitly set visibility in _Ready()

2 and 3 logic can be merged into an update function

84634E1A607A (approved):

hhyyrylainen:
I think the auto-evo tool should be updated to use the new feature, it's after all the reason why I opened the issue to add the feature.

TheHungryBandito:
While addressing the previous concerns, I exported the property so it can be set in the inspector, and it has been set to false in the auto-evo exploring tool scene to hide the move to patch button as well as the HSeparator above it.

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -127,6 +131,8 @@ public class PatchDetailsPanel : PanelContainer
     private Patch? targetPatch;
     private Patch? currentPatch;
 
+    private bool moveToPatchButtonVisible;
```
hhyyrylainen, line 134: The default value should be true.

TheHungryBandito: This will now have a default value of true.

hhyyrylainen (approved):

PR #3786, opened 2022-09-15
New Ambient Tracks

nickthenick:

This PR replaces the old ambient tracks with newer ones made by Twitchywhalez. 

They use a combination of effects such as low pass filters and others to create as realistic of an underwater sound as possible.

There are three different tracks that represent three different depths. Since we currently don't have the functionality to play different ambient tracks in different biomes, all three tracks will currently play randomly in all three biomes. 

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Oliveriver:
Played for a bit, no strong opinions to be honest. I think they're nice. Could maybe be a bit louder but it's hard to tell.

hhyyrylainen:
From a technical standpoint I think the changes are fine. I did have a quick listen in game and I think the ambiance tracks are a bit too overpowering now, at least with the default sound settings of max volume music and ambiance. I think all of the ambience tracks should maybe be reduced to 0.7 volume or maybe even a bit lower. Another thing to do would be to change the default ambiance volume level for the game but that wouldn't effect anyone who has played the game before...

Oliveriver:
I messed with the volume sliders a long time ago and have largely kept them the same since, so I don't really have a good handle on how the game feels with the default volume levels. Maybe we need to address that.

hhyyrylainen:
I do all my sound testing with the default options because that's how new players (and people who aren't too picky to adjust them) will experience the game. In my opinion that's so much more important that it is not really worth testing non-default sound settings in cases like this at all.

hhyyrylainen (approved):
I adjusted the volume levels way down for the ambiance tracks as I don't think they should overwhelm the music tracks at the default volume settings. I'll merge this soon unless someone objects to get these in, further volume tweaking can be done easily enough.

PR #3801, opened 2022-09-19
Adjust entity limits

Oliveriver:

Adjusts some entity limit settings downwards to improve performance.

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (approved):
On second thought these numbers seem pretty good to me, the only real problem is that as the min amount is 50, the settings value upgrade only kicks in for people who lowered the entity limit a lot, so the options menu will lie to the player until they change to some other option at least once. I'm not sure if anything should be done about that anymore.
I'll give time until at least tomorrow for someone to playtest these numbers.

PR #3803, opened 2022-09-19
Allow other cells in colonies to use slime jets

Oliveriver:

Fixes a bug where only the lead cell would use slime jets in multicellular.

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/early_multicellular_stage/Microbe.Multicellular.cs
+++ b/src/early_multicellular_stage/Microbe.Multicellular.cs
@@ -132,9 +132,9 @@ public partial class Microbe
     }
 
     /// <summary>
-    ///   Perform an action for all members of this cell's colony other than this cell.
+    ///   Perform an action for all members of this cell's colony other than this cell if this is the colony leader.
     /// </summary>
-    public void PerformForAllColonyMembers(Action<Microbe> action)
+    public void PerformForOtherColonyMembersIfWeAreLeader(Action<Microbe> action)
     {
         if (Colony?.Master == this)
         {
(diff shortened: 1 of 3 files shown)


src/early_multicellular_stage/Microbe.Multicellular.cs
```diff
@@ -131,6 +131,23 @@ public void BecomeFullyGrownMulticellularColony()
         }
     }
 
+    /// <summary>
+    ///   Perform an action for all members of this cell's colony other than this cell.
+    /// </summary>
+    public void PerformForAllColonyMembers(Action<Microbe> action)
```
hhyyrylainen, line 137:
```suggestion
    public void PerformForOtherColonyMembers(Action<Microbe> action)
```

The name having "all" in it, but the comment then clarifying it doesn't actually mean "all" is pretty confusing so I'd recommend this rename.

hhyyrylainen (changes requested):

src/early_multicellular_stage/Microbe.Multicellular.cs
hhyyrylainen: I fixed this myself and went with an even more verbose naming.

hhyyrylainen (approved):
Fixed one bug (if player cell didn't have jets but other members did have, this didn't work), and improved the code clarity for the "run on other colony members" part.

hhyyrylainen:
I'll merge this unless @Oliveriver has something to say about my changes.

Oliveriver:
Good changes, I didn’t realise I missed that bug. The method name is definitely unwieldy and I’m sure there’s a better alternative, but I don’t think it warrants too much work.

PR #3807, opened 2022-09-20
Balancepass 0.5.10

buckly90:

- Reduced glucose consumption from metabolic parts, survival is much less demanding.
- Significantly reduced iron compound amount held in small iron chunks.
- Significantly reduced glucose consumption of mucilage production.
- Reworked reproduction speed to scale with cell hex count.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
There's still two numbers that are missing the decimal point. `1f` and `2f` should be `1.0f` and `2.0f` respectively.

RacerBG (approved):
Seems good to me.

simulation_parameters/Constants.cs
```diff
@@ -355,9 +355,9 @@ public static class Constants
     public const bool ALWAYS_SHOW_STORED_COMPOUNDS_IN_REPRODUCTION_PROGRESS = false;
 
     /// <summary>
-    ///   How much total compounds can be absorbed by organelles to grow per second
+    ///   Multiplier on how much total compounds can be absorbed by organelles to grow per second.
     /// </summary>
-    public const float MICROBE_REPRODUCTION_MAX_COMPOUND_USE = 0.5f;
+    public const float MICROBE_REPRODUCTION_MAX_COMPOUND_USE = 2.0f;
```
hhyyrylainen, line 360: At a hex size of 1, doesn't this now cause the max reproduction speed to be multiple times faster? So doesn't this mean that small cells can now basically ignore the reproduction speed limit? Did other people balance test this?

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -904,20 +904,17 @@ private void HandleReproduction(float delta)
...
         {
-            // TODO: make the current patch affect this?
-            // TODO: make surface area of the cell (and colony) affect this
-            remainingFreeCompounds = Constants.MICROBE_REPRODUCTION_FREE_COMPOUNDS * delta;
-
+            remainingAllowedCompoundUse = remainingFreeCompounds * Constants.MICROBE_REPRODUCTION_MAX_COMPOUND_USE;
             if (IsMulticellular)
-                remainingFreeCompounds *= Constants.EARLY_MULTICELLULAR_REPRODUCTION_COMPOUND_MULTIPLIER;
+                remainingAllowedCompoundUse *= Constants.EARLY_MULTICELLULAR_REPRODUCTION_COMPOUND_MULTIPLIER;
         }
```
hhyyrylainen, line 918: Moving these variables around now has incorrect logic regarding LimitReproductionCompoundUseSpeed and PassiveGainOfReproductionCompounds flags. This will have very unintended side effects if people play with only just one of these options enabled. I can fix the logic (and put the colony size TODO comment back), once the math has been finalized.

simulation_parameters/Constants.cs
hhyyrylainen: Sorry, I misunderstood the change. So this is now that 2x the free compounds can be used for reproduction, which is pretty much perfect I think.

src/microbe_stage/Microbe.Interior.cs
hhyyrylainen: I think I probably fixed this in my commit.

hhyyrylainen (approved):
I cleaned up the logic error and improved the documentation a bit. I quickly tested some combinations of the advanced new game options, and I think now that this works correctly.

Oliveriver:
I know it's too late now, but I said I'd test this, and I have.

In general the balance feels mostly right. I found the first few generations a bit slow, but that was probably due to lack of gameplay interest in the era before other cells emerged.

I also think manually collecting ammonia and phosphate is now quite frustrating - you expend a lot of effort finding them and they only seem to increase your reproduction progress a small amount. Having tested with passive and/or limiting rates turned off (which seemed balanced and functional too), I think the issue here is the limiting rate is too slow. You fill your ammonia/phosphate storage quickly when you enter a cloud, but because it now takes a while to drain into your reproduction progress, you stop collecting from that cloud for a moment, making the whole thing a bit less satisfying than before.

But anyway, I'm aware this is too minor a thing to change last minute, so we'll just have to bear it in mind for the next release.

Issue #3201, opened 2022-04-13
Compound amounts can become NaN


hhyyrylainen:
sometimes compound amounts can become NaNs, seems like the problem is either from the process system, cloud collecting, or sharing resources in a colony. There's probably some very rare condition were the code accidentally divides by zero or something.

![screenshot](https://cdn.discordapp.com/attachments/713265723786395649/963739996051873802/Screenshot_24.png)

The warning regarding this is currently commented out as this problem is hit a *lot* so this is a high priority to get fixed. See here where the check was commented out: https://github.com/Revolutionary-Games/Thrive/pull/3412/files#diff-9c4bb7c030371c661d3147a924171fba7e368e5cb0ff2a0dad432cfd00666cb9

hhyyrylainen:
A save showing this problem:
[nan_resource.zip](https://github.com/Revolutionary-Games/Thrive/files/8486962/nan_resource.zip)

So maybe not colony related at all.

hhyyrylainen:
I created a mitigation where all NaN values are replaced by 0: https://github.com/Revolutionary-Games/Thrive/pull/3287
Still need to try to figure out what is the root cause of the issue.

84634E1A607A:
Loads of NaN report for Iron and Hydrogen Sulfide, every time I get into a colony.
[Colony_Crash_2.zip](https://github.com/Revolutionary-Games/Thrive/files/8596857/Colony_Crash_2.zip)

hhyyrylainen:
If someone wants to debug this they will want to uncomment the nan detection print: https://github.com/Revolutionary-Games/Thrive/pull/3412/files#diff-9c4bb7c030371c661d3147a924171fba7e368e5cb0ff2a0dad432cfd00666cb9

hhyyrylainen:
Seems like still can happen for some reason:
![nan iron](https://cdn.discordapp.com/attachments/958598553389903913/1023566308467548240/unknown.png)

hhyyrylainen:
Probably a related thing: investigate spawning in with NaN oxytoxy, maybe related to process system having so small numbers they are NaN and adding that? Is that possible? Also seems that in that case (and the iron case above) the mitigation currently in the game is not working for some reason.

hhyyrylainen:
This is probably fixed for now as this issue hasn't been reported in the past two years, I think.

PR #3826, opened 2022-10-06
NaN compound value

MiniAlfa:

If a microbe's compound capacity is zero, it prevents the compound from being spread out to all microbes. As a result, setting the compound amount to NaN is prevented (division by zero).

PS: during testing I encountered the ingested matter bar turning to NaN, this PR does might not compass the ingested matter bar, as I couldn't find a way to replicate that bug.

**Related Issues**

#3201

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/CompoundBag.cs
+++ b/src/microbe_stage/CompoundBag.cs
@@ -50,8 +50,9 @@ public class CompoundBag : ICompoundStorage
 
     public float GetCompoundAmount(Compound compound)
     {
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
I'm pretty sure that TryGetValue returns 0 when the value doesn't exist... at least it definitely does that for integer types so I assumed (in multiple places in the code) that it also does that for floats. Is that not true? If that is true then this PR should not even do anything.

PR #3827, opened 2022-10-07
prevent division by zero, leading to NaN.

MiniAlfa:

If a microbe's compound capacity is zero, it prevents the compound from being spread out to all microbes. As a result, setting the compound amount to NaN is prevented (division by zero).

PS: during testing I encountered the ingested matter bar turning to NaN, this PR does might not compass the ingested matter bar, as I couldn't find a way to replicate that bug.

**Related Issues**
#3201

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/ColonyCompoundBag.cs
+++ b/src/microbe_stage/ColonyCompoundBag.cs
@@ -33,7 +33,11 @@ public class ColonyCompoundBag : ICompoundStorage
             if (!compound.Key.CanBeDistributed)
                 continue;
 
-            var ratio = compound.Value / GetCapacityForCompound(compound.Key);
+            var capacityForCompound = GetCapacityForCompound(compound.Key);
+            if (capacityForCompound == 0)
+                continue;
+
+            var ratio = compound.Value / capacityForCompound;
 
             foreach (var bag in bags)
             {


hhyyrylainen:

Kasterisk:
> PS: during testing I encountered the ingested matter bar turning to NaN, this PR does might not compass the ingested matter bar, as I couldn't find a way to replicate that bug.

The ingested matter bar is calculated separately since it is its own logic. For reference, it is based on the the prey and predator's hex count so investigating those would be a good start.

hhyyrylainen:
I think this kind of code to ensure no division by 0 can happen would be good, and I'd be willing to merge this if you think this is good enough (by marking this as no longer a draft / ready for review).

src/microbe_stage/ColonyCompoundBag.cs
```diff
@@ -120,4 +135,9 @@ private IEnumerable<CompoundBag> GetCompoundBags()
     {
         return Colony.ColonyMembers.Select(p => p.Compounds);
     }
+
+    private bool IsNoneUseful(Compound compound)
```
hhyyrylainen, line 139: Usually programming conditions are easier to understand when they are positive, ie. this method should be the opposite (`IsAnyUseful`).

hhyyrylainen: Though as this also is used to check if *any* colony member finds this useful, I'd actually want this to be named `IsUsefulToAnyColonyMember` or maybe `IsUsefulInAnyCompoundBag`.

buckly90:
I did not notice any occurrences of NaN while testing, great job!

hhyyrylainen:
As I said on discord I doubt that unless we get like 50+ hours of people playing with these changes, we can't know for certain. But I of course appreciate you testing this to make sure that this doesn't accidentally introduce new bugs.

hhyyrylainen (approved):
I fixed various small style issues and also realized some general code improvements that could be done. Please check my latest commit for future work on Thrive. I'll merge this now after doing a quick test myself.

src/microbe_stage/ColonyCompoundBag.cs
```diff
@@ -26,34 +28,38 @@ public float GetCapacityForCompound(Compound compound)
     public void DistributeCompoundSurplus()
     {
         var bags = GetCompoundBags().ToList();
-        using var compounds = GetEnumerator();
```
MiniAlfa, line 28: The |GetEnumerator| function has been overwritten at line 75, will it still return the same value as |foreach x in this| ?

hhyyrylainen: AFAIK using foreach is the equivalent to:
```
var enumerator = value.GetEnumerator();
while(enumerator.Next())
{
   var item = enumerator.Current;
    ...
}
enumerator.Dispose();
```

With this being the foreach statement:
```
foreach(var item in value)
{
    ...
}
```

PR #3829, opened 2022-10-13
Keep alpha value on mutation

### LINKED ISSUE #3675: Auto-evo can change alpha value of cell membrane (although it is effectless)
Noticed this from Discord:

image

thelayeredmind:

During Mutation based on different Microbial parameters colors are generated randomly.
By keeping the old Alpha value from the original species we avoid mutating the Alpha value.

**Related Issues**

fixes #3675 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/Mutations.cs
+++ b/src/general/Mutations.cs
@@ -110,7 +110,7 @@ public class Mutations
         }
 
         // preserve Alpha on mutation.
-        mutated.Colour = new(colour.r, colour.g, colour.b, mutated.Colour.a);
+        mutated.Colour = new Color(colour.r, colour.g, colour.b, mutated.Colour.a);
 
         mutated.MembraneRigidity = Math.Max(Math.Min(parent.MembraneRigidity +
             random.Next(-25, 26) / 100.0f, 1), -1);
--- a/src/general/Mutations.cs
+++ b/src/general/Mutations.cs
@@ -109,7 +109,8 @@ public class Mutations
             mutated.MembraneType = parent.MembraneType;
         }
 
-        mutated.Colour = colour;
+        // preserve Alpha on mutation.
+        mutated.Colour = new(colour.r, colour.g, colour.b, mutated.Colour.a);
 
         mutated.MembraneRigidity = Math.Max(Math.Min(parent.MembraneRigidity +
             random.Next(-25, 26) / 100.0f, 1), -1);


Kasterisk:
CI complains about missing type specification, it probably wants the instantiation to look like this:
```diff
- mutated.Colour = new(colour.r, colour.g, colour.b, mutated.Colour.a);
+ mutated.Colour = new Color(colour.r, colour.g, colour.b, mutated.Colour.a);
```
Otherwise looks good to me.

84634E1A607A:
Wait. Looking at the code, it seems that the alpha value change is completely intended and @hhyyrylainen coded that. (So why is #3675 `bug` and `good first issue`?)

hhyyrylainen:
I just ported over the old code. I think I allowed that issue to be opened as someone complained something about the species colours causing weird issues in graphs or something? Maybe @Kasterisk recalls something like that. I guess removing AI species being able to camouflage with the colour should not be removed, then.

Kasterisk:
> causing weird issues in graphs or something?

Yes.

I said this before but I don't see why we should include transparency as a thing that mutates, that's not realistic and as stated in the issue, it doesn't change the actual cell's visuals either. Moreover it causes spillover effect as shown with the graph lines before.

84634E1A607A:
If so, generation logic should be altered instead.

thelayeredmind:
Do you want me to adjust the Methods for color generation to ignore Alpha instead?
I also noticed that alpha mutation was there by design, and therefore not really being a bug.

hhyyrylainen:
> I said this before but I don't see why we should include transparency as a thing that mutates, that's not realistic and as stated in 

Camouflage. Untrustedlife would roll in his grave if we has dead...

So I think I'll close the issue about the alpha value and instead open an issue about making the other places in the code take into account the fact that microbes can use transparency as camouflage, as that's been a feature since around 0.4.0, I think.

hhyyrylainen:
I closed the original issue, and instead opened: https://github.com/Revolutionary-Games/Thrive/issues/3831
so this PR should probably be closed. Sorry about the wasted work, I didn't realize this at the time when I saw that issue be opened.

Kasterisk:
Okay, do the alpha value actually even affect the membrane? If not then that should rectified.

hhyyrylainen:
It actually doesn't... well that's one more thing no one noticed broke with the engine change. Opened an issue now: https://github.com/Revolutionary-Games/Thrive/issues/3832

thelayeredmind:
> I closed the original issue, and instead opened: #3831

If you want you can assign me to it. I would like to keep at it.

And don't worry about the wasted work. It was an easy opportunity to get to know the dev environment ☺️
I will question the issues more before starting to work on them the next time 😉

84634E1A607A:
Sometimes questions never pop up until someone works on it.😂

PR #3837, opened 2022-10-14
Add GUI Color for Species that interacts with GUI

### LINKED ISSUE #3831: Fix graphs and other code that doesn't work with transparency in species colours
Species being transparent for camouflage against the player (exploiting the fact that the dying penalty the player causes will kill off species the player can see) has been feature for many years. I forgot that when I allowed an earlier issue to be kept open.

So instead here's the right way to do it all code needs to be updated to take into account the fact that species can have non-1 alpha value as their colour, so for display in graphs etc. the alpha should be forced to be 1.

thelayeredmind:

With this change we render Cells in Full Alpha (1.0) when they are rendered for the GUI / Preview.
I extended the ApplySpecies to add an indirection to a SetupSpeciesPreview and SetMembraneGUI so when new GUI requirements come in the future, it is untangled from the remaining game code. I also see how this can feel like redundant code, so I'll leave the decision up to you guys!

I tested it with the Auto Evo GUI. If there is any other relevant GUI you want me to test it on, just tell me 😄 

**Related Issues**

fixes #3831 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/Species.cs
+++ b/src/general/Species.cs
@@ -46,11 +46,8 @@ public abstract class Species : ICloneable
 
     public Color Colour { get; set; } = new(1, 1, 1);
 
-    public Color ColourGUI
-    {
-        get => new(Colour.r, Colour.g, Colour.b, 1);
-    }
-
+    public Color ColourGUI => new(Colour.r, Colour.g, Colour.b, 1);
+    
     /// <summary>
     ///   Set to true when this species has evolved to a different species class type. This is mostly used to detect
     ///   that old species that should no longer be in use, are not used. Once this has been set to true, don't set
--- a/src/general/Species.cs
+++ b/src/general/Species.cs
@@ -47,8 +47,9 @@ public abstract class Species : ICloneable
     public Color Colour { get; set; } = new(1, 1, 1);
 
     /// <summary>
-    ///   Get the color value for GUI Components affected by a Species colour.
-    ///   To add additional guards like a fully opaque alpha value.
+    ///   The colour value for GUI Components that want to show this species' colour.
+    ///   This value has additional constraints compared to plain Colour,
+    ///   for example ensuring full opacity to avoid transparency, which can cause rendering bugs.
     /// </summary>
     public Color GUIColour
     {
--- a/src/general/Species.cs
+++ b/src/general/Species.cs
@@ -46,9 +46,9 @@ public abstract class Species : ICloneable
 
     public Color Colour { get; set; } = new(1, 1, 1);
 
-    public Color ColourGUI 
+    public Color ColourGUI
     {
-        get => new Color(Colour.r, Colour.g, Colour.b, 1);
+        get => new(Colour.r, Colour.g, Colour.b, 1);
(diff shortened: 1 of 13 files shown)


hhyyrylainen:
Isn't this the exact opposite of what's needed? ie. the membrane should use the species colour as is but the graphs etc. should use a different colour.

thelayeredmind:
> Isn't this the exact opposite of what's needed? ie. the membrane should use the species colour as is but the graphs etc. should use a different colour.

I here you. But I don't really understand what graphs mean. then. I thought you meant the species that is rendered in the graph. And from what I saw it uses the PhotoStudio to render it out from a Game Scene. Can you elaborate what exactly you need then?

hhyyrylainen:
When you enter the editor, the right side of the screen has various graphs, the things with the lines representing population numbers, another graph is the compound concentrations over time.

thelayeredmind:
Damn, sorry I completely misunderstood. I did not realize that the species color is used to color UI elements themselve!
I will get on it this WE when I have time.

thelayeredmind:
Hey @hhyyrylainen I looked into the code, where Species.Colour is referenced and could only find LineChartData.
However when implementing my own change I saw that there is already a line of code setting alpha to 1.
https://github.com/Revolutionary-Games/Thrive/blob/129935df9d64e10279cfbc0a0d32740d396c5e1a/src/gui_common/charts/ChartDataSet.cs#L30

Now with my new change I implemented a GUI Color getter and put in the only place I really could find, where gui is affected by a species.

hhyyrylainen:
Okay, that sounds like the correct approach. Could you rename this PR as the name no longer makes sense?

hhyyrylainen:
You should fix the automatic code checks that are failing. I'll check this over in a day or two once I have time once the checks are made happy.

src/general/Species.cs
```diff
@@ -46,6 +46,8 @@ protected Species(uint id, string genus, string epithet)
 
     public Color Colour { get; set; } = new(1, 1, 1);
 
+    public Color ColourGUI => new(Colour.r, Colour.g, Colour.b, 1);
```
hhyyrylainen, line 49: I'd personally want to see this named `GUIColour` so that this reads naturally (while I do get that having a suffix can be better for discoverability.

Also this reads the Colour property 3 times so this results in 3 method calls on each read...
So I'd like to see this written like:
```
get
{
    var colour = Colour;
    return new Color(...);
}
```

hhyyrylainen (changes requested):

src/general/Species.cs
hhyyrylainen: Oh and one more thing, there should be an xml comment on this property explaining why it exists (and is different from Colour) and when it should be used.

hhyyrylainen (approved):
Code change looks good enough to me now. I did a quick test and looked around the code and I think this probably is used in all the right places, I also only found MicrobeEditorReportComponent.cs.

Issue #3810, opened 2022-09-22
Enable word wrap for credits dynamic text parts (especially 3 columns)


hhyyrylainen:
because if multiple people with really long names end up on the same line their names will overlap, and ridiculously long names can overlap even normal names.

If I just enable word wrap then that just completely destroyed the credits screen showing name lists correctly, so there's some tricky bit of Godot GUI logic that needs to be correctly used here.

hhyyrylainen:
I'll mark this as a bug as there's currently a long name that pushes all of the names to the right so they aren't properly aligned in the translators section.

Issue #2599, opened 2021-09-19
Add fast forward and skip buttons for the credits


hhyyrylainen:
Add for the credits: fast forward and skip (need to hold for 3 seconds) keybindings (but when opened from the main menu, ESC and the skip button, if different, should immediately exit back to the main menu)

there should be a popup text saying how long to hold the skip button for that becomes visible if the skip button becomes pressed

hhyyrylainen:
Skip button is already implemented for the credits (at least when viewed at the end of the game). A fast forward should work when used from the main menu or seen at the end of the game.

hhyyrylainen:
We have the button now but there's no GUI element added.

PR #3849, opened 2022-11-05
Fix word wrap for dynamic credits parts

Oliveriver:

Fixes long names in credit blocks causing offset columns.

Note this does make the offending column slightly taller than the others, but unless we end up with multiple super long names, it's not a big issue.

**Related Issues**

Fixes https://github.com/Revolutionary-Games/Thrive/issues/3810

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/CustomExpandingLabel.cs
+++ b/src/gui_common/CustomExpandingLabel.cs
@@ -1,4 +1,4 @@
-using Godot;
+﻿using Godot;
 
 /// <summary>
 ///   Custom label with word wrap which dynamically expands to fill space.
@@ -18,7 +18,7 @@ public class CustomExpandingLabel : Label
     public CustomExpandingLabel(string text)
     {
         Text = text;
-        Align = Label.AlignEnum.Center;
+        Align = AlignEnum.Center;
         SizeFlagsHorizontal = (int)SizeFlags.ExpandFill;
     }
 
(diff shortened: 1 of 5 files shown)


src/gui_common/CustomExpandingLabel.cs
```diff
@@ -0,0 +1,34 @@
...
+///   <para>
+///     Due to a Godot bug, ordinary labels with word wrap and the expand fill flag initialise with a width of zero,
+///     so each character ends up on a new line. This massively expands the height of the label, which doesn't reset
+///     afterwards. See https://github.com/godotengine/godot/issues/47005.
+///   </para>
+///   <para>
+///     This custom class fixes the bug by only enabling word wrap once the label becomes visible.
+///   </para>
+/// </remarks>
+public class CustomExpandingLabel : Label
```
hhyyrylainen, line 16: Shouldn't this be named closer to something like "CustomExpandingWordWrappedLabel" to explain why this exists better? As the problem is the combination of expanding and word wrapping.

buckly90:
Did not notice any issues, great work!

hhyyrylainen:
There is no fast forward, I couldn't be bothered to implement it when doing the credits, instead there's an open issue: https://github.com/Revolutionary-Games/Thrive/issues/2599

hhyyrylainen (approved):
Code looks fine and seems to work fine in game, though I didn't notice any names that got wrapped, but then again I could have missed. I did notice that the translators names section was better aligned with these changes. Thanks for taking care of this small issue so that I didn't have to.

PR #3859, opened 2022-11-13
Some tweaks for the new UI elements

Kasterisk:
- Shrink socials icon a little, moved them closer to the screen edge.
- Changed websites button container to a PopupPanel.
- More suitable font for the nonactive mods warning popup.
- Replaced default checkbox used in the nonactive mods warning popup with our custom checkbox.
- Made Thriveopedia UI make use more of the available screen space.
- Addded expand and collapse animation to the Thriveopedia's page tree.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Oliveriver:
Unfortunately I've found a few issues with the changes here.

* I personally would still like a small amount of space between the social media icons and the screen edges, but that's a very minor one.
* The pop-up container with website links has inconsistent corner shapes. The top two corners are square but the bottom two are rounded.
* The Thriveopedia animation causes some problems. It's fine on the current world, patch map and tree pages, but on the home page it causes the page text to awkwardly reposition itself throughout the animation, and on the museum page the card container stays the same size for the reverse animation, leading to the whole Thriveopedia getting too wide during the animation before it snaps back to the correct width at the end.

Kasterisk:
- Gave it a little more margin.
- Rounded all the corners off.
- I've now replaced the GridContainer used for containing cards with a HFlowContainer, basically it will make columns adjusted automatically according to the container's width. As for the page text, I don't see it much of an issue repositioning itself.

src/general/StageHUDBase.cs
```diff
@@ -757,7 +757,7 @@ public void ShowFossilisationDialog(FossilisationButton button)
         if (button.AttachedOrganism is Microbe microbe)
         {
             fossilisationDialog.SelectedSpecies = microbe.Species;
-            fossilisationDialog.Show();
+            fossilisationDialog.PopupCenteredShrink();
```
Kasterisk, line 760: Here I made the fossilization dialog be shown according to the [style guide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md#:~:text=Popups%20should%20be%20shown%20with%20PopupCenteredShrink()) as I think it's the correct way what's intended to be done here.

Past observations show that Godot popups don't usually resize dynamically, this is bad when switching between languages as text lengths differ. This is the main reason why `PopupCenteredShrink` is preferable, to keep a popup true to its approximate intended size.

src/thriveopedia/pages/ThriveopediaMuseumPage.cs
```diff
@@ -76,8 +76,7 @@ public override void UpdateCurrentWorldDetails()
 
     public override void OnNavigationPanelSizeChanged(bool collapsed)
     {
-        // Change the number of columns to reflect the space the museum takes up
-        cardContainer.Columns = collapsed ? 3 : 4;
+        _ = collapsed;
```
Oliveriver, line 79: You can probably get rid of this whole method from the base class now, since all I needed it for was adjusting the number of columns on this page. It doesn't do anything for other pages.

hhyyrylainen: I'd leave it be (or at least make it just a virtual method that does nothing by default) to allow future flexibility.

src/general/StageHUDBase.cs
Oliveriver: Definitely a good change.

Oliveriver:
Ok, the animation is better now. I'm still not a fan of the way the text rearranges itself but if other people aren't bothered by it then happy for it to be kept.

I have noticed one new really weird thing. The collapse/expand page tree button doesn't have the right clickable area when the Thriveopedia is opened from the main menu. You can't click on it if the cursor is over the icon itself. I have no idea why it might be happening, but it only happens when opened from the main menu and it's not happening on master.

I also think the museum card list should be centred in its container if possible. Another small thing though.

Kasterisk:
Fixed the clickable area problem, turns out it was just the TooltipManager scene getting in the way.

> I also think the museum card list should be centred in its container if possible. Another small thing though.

I didn't find a way to do this.

src/general/MainMenu.cs
```diff
@@ -622,7 +622,8 @@ private void OnReturnFromArtGallery()
 
     private void OnWebsitesButtonPressed()
     {
-        websiteButtonsContainer.Visible = !websiteButtonsContainer.Visible;
+        websiteButtonsContainer.ShowModal();
+        websiteButtonsContainer.RectSize = Vector2.Zero;
```
hhyyrylainen, line 626: Presumably this hack is needed because this doesn't use our custom dialog? I think a comment should be added here as to why this is needed.

src/general/MainMenu.tscn
```diff
@@ -850,22 +861,23 @@ ErrorMessage = "MOD_LOAD_ERRORS_OCCURRED"
...
-[connection signal="pressed" from="MenuContainers/SocialMedia/Icons/GitHubButton" to="." method="OnSocialMediaButtonPressed" binds= [ "https://github.com/Revolutionary-Games/Thrive" ]]
-[connection signal="pressed" from="MenuContainers/SocialMedia/Icons/YouTubeButton" to="." method="OnSocialMediaButtonPressed" binds= [ "https://www.youtube.com/c/RevolutionaryGames" ]]
-[connection signal="pressed" from="MenuContainers/SocialMedia/Icons/RedditButton" to="." method="OnSocialMediaButtonPressed" binds= [ "https://www.reddit.com/r/thrive/" ]]
-[connection signal="pressed" from="MenuContainers/SocialMedia/Icons/DiscordButton" to="." method="OnSocialMediaButtonPressed" binds= [ "https://discordapp.com/invite/ECR9E8x" ]]
-[connection signal="pressed" from="MenuContainers/SocialMedia/Icons/FacebookButton" to="." method="OnSocialMediaButtonPressed" binds= [ "https://www.facebook.com/Thrive-182887991751358/" ]]
-[connection signal="pressed" from="MenuContainers/SocialMedia/Icons/TwitterButton" to="." method="OnSocialMediaButtonPressed" binds= [ "https://twitter.com/Thrive_Game" ]]
+[connection signal="about_to_show" from="MenuContainers/WebsiteButtons" to="." method="OnWebsitesButtonShow"]
```
hhyyrylainen, line 864: Where is `OnWebsitesButtonShow` defined? I can't see that in this diff.

hhyyrylainen:
A slight problem I ran into is that this branch reuses the `gui_tweaks` name, which is also another branch I had locally that is 1400 commits behind master. So branch names shouldn't be reused, because it takes me extra 2 minutes to sort my local copy.

hhyyrylainen (changes requested):
I think the functional changes are fine, though maybe the navigation panel animation in the thriveopedia could be just a bit faster. I just had one question about what seems to be a hack around a problem in Godot in the code.

hhyyrylainen (approved):
My code related comments were resolved, I'll assume the functional behaviour has not changed so I won't playtest this again.

PR #3863, opened 2022-11-18
Pilus invulnerability frames

### LINKED ISSUE #3267: Pilus should not be able to deal a ton of damage at once
currently the pilus deals damage as physics callbacks are triggered. These can trigger *very* fast so the pilus can basically instantly kill something. Instead there needs to be some limit. We could consider a flat delay between times when the pilus can deal damage or somehow use the physics system to determine how hard the pilus hit (velocity between bodies? physics interlap distance {this needs also physics based turning}?)

Enovale:

This PR adds invulnerability frames when a cell comes in contact with a Pilus, preventing further damage for a small amount of time. This helps to prevent Pili from insta-killing other cells if you contact multiple cell collision shapes on the same frame. I did not put any effort into balancing the new damage numbers, and I had to increase the Pili base damage to account for the new invulnerability.

**Related Issues**

Closes #3267 but will need balancing.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/Microbe.Contact.cs
```diff
@@ -337,9 +351,19 @@ public void Damage(float amount, string source)
         }
         else if (source == "pilus")
         {
+            if (invulnerabilityDuration > 0)
+            {
+                GD.Print($"Invuln! {amount}");
```
hhyyrylainen, line 356: Debug prints should be removed.

src/microbe_stage/Microbe.cs
```diff
@@ -725,6 +725,8 @@ public void ProcessSync(float delta)
 
         HandleReproduction(delta);
 
+        HandleInvulnerabilityDecay(delta);
```
hhyyrylainen, line 728: This should be put in the process async part as this doesn't require processing by the main thread.

src/microbe_stage/Microbe.Contact.cs
Enovale: Removed in the latest commit

src/microbe_stage/Microbe.cs
Enovale: I moved this line to ProcessEarlyAsync... I think this is the correct place.

buckly90:
Seems to work without issue, the i-frames make pili much more predictable and consistent, which will go a long way in making for a more balanced combat. We'll need to discuss how dangerous we want pili to be before balancing it properly, but the current damage value seems acceptable.

However; there is a rather unusual issue that I have noticed. The audio for AI cells is strangely quiet and disjointed in this branch compared to others. It's very subtle and I even went as far as running two different versions of Thrive at once to identify it but it's definitely there. I'm not really sure what could be causing it.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -76,6 +76,12 @@ public partial class Microbe
     [JsonProperty]
     private int flashPriority;
 
+    /// <summary>
+    ///   This determines how much time is left (in seconds) until this cell can take damage again after becoming
+    ///   invulnerable due to a damage source. This was added to balance pili but might extend to more sources.
+    /// </summary>
+    private float invulnerabilityDuration;
```
hhyyrylainen, line 83: This should be JsonProperty so that spamming save and load doesn't let you get past the invulnerability time.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -269,6 +275,14 @@ public void ApplyMembraneWigglyness()
             Membrane.Type.MovementWigglyness) * 0.2f;
     }
 
+    /// <summary>
+    ///   Give this microbe a specified amount of invulnerability time. Overrides previous value.
+    /// </summary>
```
hhyyrylainen, line 280: Could you add a remark here that the invulnerability doesn't apply to all damage sources?

hhyyrylainen (changes requested):
A few more code comments but other than that the code is starting to look right.

hhyyrylainen:
@buckly90 as far as I read from the code, the sound playing is the same as before. The only change this can do is that the pilus sound is no longer layered multiple times on top of itself, instead it plays just once. To counteract this a bit it is probably possible to specify the pilus sound to be played at like 1.2 or even more times the default volume.

src/microbe_stage/Microbe.Contact.cs
Enovale: Added this and tested that indeed you cannot cheese the invulnerability frames now.

src/microbe_stage/Microbe.Contact.cs
Enovale: Yeah I think I mentioned it in a different comment (?) but I should have specified it more clearly

hhyyrylainen: Well my concern here is that someone will read this documentation on this method and think they can use it in some different context, but when they try to use it, then it doesn't work at all for non-pilus damage. That's why I think the documentation on this method should be clearer.

Enovale: Yeah, I agree. I added a small clarification in the latest commit.

hhyyrylainen: This overtly specific comment line (281) still exists...

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -337,9 +353,16 @@ public void Damage(float amount, string source)
         }
         else if (source == "pilus")
         {
+            if (invulnerabilityDuration > 0)
+                return;
+
             // Play the pilus sound
             PlaySoundEffect("res://assets/sounds/soundeffects/pilus_puncture_stab.ogg");
```
hhyyrylainen, line 360: Could you try adjusting the sound effect volume here to address the sound difference @buckly90 is hearing?

buckly90: Might be a good idea for someone to make sure it's not just something funky with audio mixing on my end just to be sure.

Don't want to accidentally blow out somebody's speakers with multiplied pili volume.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -997,6 +1020,17 @@ private void CheckEngulfShape()
...
     }
 
+    /// <summary>
+    ///   Decrease the remaining invulnerability time
+    /// </summary>
+    private void HandleInvulnerabilityDecay(float delta)
+    {
+        if (invulnerabilityDuration > 0)
+        {
+            invulnerabilityDuration -= delta;
```
adQuid, line 1030: this could allow invulnerabilityDuration to become less than zero, but personally I don't think it's work the extra computation time to check that, since it shouldn't be a problem.

src/microbe_stage/Microbe.Contact.cs
Enovale: What do you mean? I made the comment more specific to say that only Pili have implemented i-frames. The line 281 should specify this now.

src/microbe_stage/Microbe.Contact.cs
Enovale: I tried this myself and I have noticed an issue with the pili audio balance. in Master, the sound of a pilus hitting the player or another cell gets multiplied in volume because it gets played many times on the same couple frames. This makes it range from really quiet to really loud. Now, in my branch, it is only quiet. To boot, it needs quite a hefty volume boost to sound normal (to me, im not used to this game.) I tried 4.0 and it sounded okay. At least comparable to master.

src/microbe_stage/Microbe.Contact.cs
hhyyrylainen: I mean that this method documentation should mention that this doesn't apply to all damage sources without specifically saying the pilus is the one this applies to. Because in the future:
- Someone will update the invulnerability to also apply to some other thing
- That person will forget to update the comment here, and if I don't catch it in a code review, the outdated comment will be still here
- Someone else will be doing work on Thrive and read the documentation on this method, and read that only pilus is affected
- That someone may waste multiple hours of their time and sanity due to an outdated comment

That's why I like to avoid unnecessarily specific comments that can easily get outdated.

src/microbe_stage/Microbe.Contact.cs
hhyyrylainen: Did you push your changes? Once you've done that @buckly90 can probably listen to the changes again.

Kasterisk:
Just my thoughts, but I personally would like a physics based approach more instead of a hardcoded invulnerability. I think it will give more gameplay benefits, for example it could potentially allow pilus to deal damage dynamically, the amount of which can be calculated from how hard a pilus collides with its target. Very flexible system. This would also go hand in hand with the stamina/sprinting mechanics, where in order to damage effectively you basically need to ram cells at full force.

hhyyrylainen:
That's been looked at before and sadly Godot doesn't expose the physics contact penetration amount. And in fact with the cells bumping into each other code a hack is used as all of the contact forces are summed, which would not be very good for pilus contact as cells that are in contact with other things already would take multiplied damage from pili. Though I  guess if we rewrote the game logic in pure C# and did our own physics engine integration we could do this (https://forum.revolutionarygamesstudio.com/t/alternative-ways-to-fix-the-performance/907).

src/microbe_stage/Microbe.Contact.cs
Oliveriver: Adjusted the volume and 4.0f sounds about right. Perhaps a tiny bit too loud, but even in an ocean full of spiky creatures I certainly wasn't being deafened or anything.

src/microbe_stage/Microbe.Contact.cs
Oliveriver: The comment has been changed to refer to pili (for some reason GitHub is showing an outdated version on the issue discussion but not on the diff page). Do you still want it to be changed again?

hhyyrylainen: Well the version shown on the diff page is the one I take issue with. The one shown here on the discussion page actually looks better. For absolutely clarify I checked out the code locally and this is the not that good version I see:
kuva

Oliveriver: Looks like @Enovale has updated it now.

Nunez2196 (approved):

hhyyrylainen (approved):
Okay, the comment is now general enough. I'll merge this now so that this makes it in the 0.6.0 release.

PR #3870, opened 2022-11-23
Launcher 2.0 feature needed changes (and a few misc changes)

hhyyrylainen:
Thrive side changes needed for launcher 2.0

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/MainMenu.tscn
```diff
@@ -199,58 +200,41 @@ __meta__ = {
...
-margin_bottom = 234.0
+margin_bottom = 194.0
 texture = ExtResource( 10 )
 __meta__ = {
 "_edit_use_anchors_": false
 }
 
-[node name="Menus" type="Control" parent="MenuContainers"]
-anchor_top = 0.368
+[node name="Menus" type="CenterContainer" parent="MenuContainers"]
```
Kasterisk, line 221: This change is probably what's affecting the animation the most.

Kasterisk (commented):
Main menu is getting too crowded with buttons, no problem for now but I think we'll need a major overhaul, eventually.

src/general/MainMenu.tscn
hhyyrylainen: I found it to make more sense to make it this way as I didn't understand how it was previously even working...

Also this should hopefully be the last button ever needed in the main menu, at least I hope, which is why I added it to the main menu. If we ever need more buttons, then a different design will be needed...

Kasterisk: We both agree on a big overhaul then.

hhyyrylainen:
This should now have everything that's absolutely necessary for 0.6.0 (which means that this is still missing the planned development feed that could be shown in the main menu).

PR #3872, opened 2022-11-24
Organelle depth check

### LINKED ISSUE #1070: Check if organelle shader blendmode needs changes
It seems that the blending is not very good ATM and when the organelle draw order changes around there is visible difference when the organelles swap the order they are drawn in.

### LINKED ISSUE #3743: Improve model for the mucilage jet
Related to this feature: https://github.com/Revolutionary-Games/Thrive/pull/3590
Should look something like this: https://cdn.discordapp.com/attachments/464846906971914250/1008891677659897896/another_vase.png

Nunez2196:

This PR attempts to fix depth testing with organelles which would cause strange clipping behavior

**Related Issues**
closes #3743
closes #1070

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I guess if this works, this would also finally close https://github.com/Revolutionary-Games/Thrive/issues/1070 ?

Oliveriver:
Seems to fix all the outstanding issues with the slime jet. Nice work!

assets/models/organelles/SlimeJet.tscn
```diff
@@ -1,17 +1,14 @@
-[gd_scene load_steps=7 format=2]
+[gd_scene load_steps=6 format=2]
 
 [ext_resource path="res://assets/models/organelles/SlimeJet.glb" type="PackedScene" id=1]
-[ext_resource path="res://shaders/Organelle.shader" type="Shader" id=2]
+[ext_resource path="res://shaders/OrganelleOpaque.shader" type="Shader" id=2]
 [ext_resource path="res://assets/textures/dissolve_noise.tres" type="Texture" id=3]
 [ext_resource path="res://assets/textures/SlimeJet.png" type="Texture" id=4]
-[ext_resource path="res://assets/models/organelles/SlimeJet.anim" type="Animation" id=5]
```
hhyyrylainen, line 7: What was the slime jet animation for? @Oliveriver ? If the animation is no longer used / it wasn't good, then I think deleting the animation file as well would be good to not confuse people into thinking the animation file does something.

Nunez2196: when Rose exported the .glb file from blender, the animation was packaged with the mesh so it could just be left overs from implementing that .glb file directly.

hhyyrylainen: I can't find any reference to the anim file anywhere else, so if it is actually meant to be used removing it from here will break it entirely.

Oliveriver: The .glb file has a dependency on the .anim file. Removing the .anim file causes a crash when loading the jet model. I managed to remove the dependency in the .glb file but that meant the animation didn't play. This could be fixed if the animation can be embedded in the model itself, but I don't know how to do that, so for now, we need to leave the .anim file in.

Oliveriver (approved):

assets/models/organelles/SlimeJet.tscn
hhyyrylainen: The slime jet component *does* actually refer to an animation player, so the changes in this file regarding the removal of the animation need to be reverted. That's the last thing needed in this PR.

hhyyrylainen (changes requested):
While this doesn't crash now, one of two things are needed:
- Removal of the slime jet component's use of the animation player
- Putting back the referenced animation in the slime jet organelle file

hhyyrylainen:
I guess I'll put the removed stuff back into the organelle scene file and merge this to get this in for 0.6.0...

assets/models/organelles/SlimeJet.tscn
hhyyrylainen: I've now fixed this myself, and I guess I'll merge this now as I didn't notice that undoing these changes broke anything I could see quickly.

Issue #3548, opened 2022-07-17
Ingested Matter amount can be NaN


stoptakingusernames:
The ingested matter bar can say NaN instead of the actual digested amount. Digestion still happens as usual regardless though

**To Reproduce**
I don't know how exactly to reproduce this bug, so I'll just share what I did before I got it
1. Play the game and become multicellular 
3. Load the multicellular save again after quitting
4. Play the game as a multicellular organism (I noticed at this time that the ingested matter bar did not even appear at all)
5. Die, then ingest another cell
6. See the bug

**Expected behaviour**
Ingested amount show how much ingested material you currently hold

**Screenshots**

![c](https://user-images.githubusercontent.com/33234241/179401713-d08ef168-ebaa-4531-89f2-f7991b54773b.png)

**Additional context**
This bug only happened in the multicellular stage for me.

hhyyrylainen:
One potential lead as to what might cause it or make it more likely: loading a early multicellular save, then dying, and finally trying to engulf an iron chunk.

jacekku:
Hello there! I've encountered this bug in my play through, when going to multicellular stage. I decided to investigate and debug a bit. What I discovered was the fix that was implemented for CompoundBags with `FixNaNCompounds()` BUT, the items I was digesting didn't seem to hit that method ever, resulting in NaN's still existing in the digested body, corrupting the values on every update. I added `containedCompounds.FixNaNCompounds();` in the digesting logic and that seemed to fix it. 

Excerpt from the modified file looks like this:
```CSharp
Microbe.Interior.cs
   private void HandleDigestion(float delta)
            //...
            var containedCompounds = engulfable.Compounds;
            var additionalCompounds = engulfedObject.AdditionalEngulfableCompounds;

            var totalAmountLeft = 0.0f;

+           containedCompounds.FixNaNCompounds();

            foreach (var compound in compoundTypes.Values)
            {
                if (!compound.Digestible)
                    continue;
            //...
```

Hope this helps.

Kasterisk:
For posterity, this is now masked with the above workaround, the actual causing issue however has not been found (and fixed).

hhyyrylainen:
I guess that's good enough, so I'll remove this from the milestones to not have this hang around in there indefinitely.

hhyyrylainen:
This might no longer happen with the ECS refactor (but there is a probably related bug: https://github.com/Revolutionary-Games/Thrive/issues/4794).

[...]

PR #3877, opened 2022-11-27
Fix infinite compound ejection glitch

Oliveriver:

Fixes cells infinitely spawning compounds due to corrupted engulfed objects by setting all NaN values to zero.

Based on the suggestion [here](https://github.com/Revolutionary-Games/Thrive/issues/3548#issuecomment-1321201278). Seems to fix the issue and the related problem of ingested matter not appearing in the UI in a save in this corrupted state.

**Related Issues**

Mitigates (but might not wholly fix) https://github.com/Revolutionary-Games/Thrive/issues/3548 and https://github.com/Revolutionary-Games/Thrive/issues/3201

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Microbe.Interior.cs
+++ b/src/microbe_stage/Microbe.Interior.cs
@@ -1453,6 +1453,7 @@ public partial class Microbe
 
             var containedCompounds = engulfable.Compounds;
             var additionalCompounds = engulfedObject.AdditionalEngulfableCompounds;
+            containedCompounds.FixNaNCompounds();
 
             var totalAmountLeft = 0.0f;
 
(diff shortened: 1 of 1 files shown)


hhyyrylainen (changes requested):
Could you add a comment mentioning that this is a workaround and link the related Github issues?

hhyyrylainen:
https://github.com/Revolutionary-Games/Thrive/issues/3201 existed before the engulfing revamp so this cannot fix the root cause, which is namely that something puts NaN compound amounts into the cells.

hhyyrylainen (approved):
Code change looks fine now.

hhyyrylainen:
I guess I'll merge this without getting proper testing and we'll see after the release is out if we get reports of people running into the problem anymore.

Oliveriver:
I did test this myself with a save provided where the problem was happening. Infinite ejection stopped and ingested matter corrected itself. But yeah, wider testing would have been useful.

PR #3886, opened 2022-11-30
More precise digestion efficiency stat display

Kasterisk:
Digestion Efficiency stat now displays the efficiency for all respective enzymes present in the cell separately *if* the cell has multiple digestive enzymes present. Otherwise, the stat will display a single percentage value like before.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (approved):
I found a couple of style things in this which I fixed already. And I also found a bug / non-optimal functionality where the up/down arrow was a question mark quite often on the efficiency label, so I made it now so that the arrow points mostly up when using multiple digestion types.

PR #3914, opened 2022-12-07
Changed object.ForEach to foreach for a small performance increase.

TristanMisja:


TristanMisja:
Sorry about the duplicate PR, I'm still getting used to git.

PR #3926, opened 2022-12-09
Fix infinite compound venting due to having compounds over capacity in the cell

Oliveriver:

Prevents taking negative compounds from the cloud system, which was causing another infinite compound venting glitch. Also fixes compound venting to seemingly prevent a cell ending up with more compounds than it can contain.

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/CompoundCloudPlane.cs
+++ b/src/microbe_stage/CompoundCloudPlane.cs
@@ -400,6 +400,9 @@ public class CompoundCloudPlane : CSGMesh, ISaveLoadedTracked
     /// <returns>The amount of compound taken</returns>
     public float TakeCompound(Compound compound, int x, int y, float fraction = 1.0f)
     {
+        if (fraction < 0.0f)
+            return 0.0f;
+
         float amountInCloud = HackyAddress(Density[x, y], GetCompoundIndex(compound));
         var amountToGive = amountInCloud * fraction;
 
--- a/src/microbe_stage/CompoundCloudPlane.cs
+++ b/src/microbe_stage/CompoundCloudPlane.cs
@@ -400,6 +400,8 @@ public class CompoundCloudPlane : CSGMesh, ISaveLoadedTracked
     /// <returns>The amount of compound taken</returns>
     public float TakeCompound(Compound compound, int x, int y, float fraction = 1.0f)
     {
+        // TODO: change to throwing an error once we've fixed cells requesting negative compounds
+        // https://github.com/Revolutionary-Games/Thrive/issues/3927
         if (fraction < 0.0f)
             return 0.0f;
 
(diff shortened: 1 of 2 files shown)


src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -400,6 +400,9 @@ public void AddCloud(Compound compound, float density, int x, int y)
     /// <returns>The amount of compound taken</returns>
     public float TakeCompound(Compound compound, int x, int y, float fraction = 1.0f)
     {
+        if (fraction < 0.0f)
+            return 0.0f;
```
hhyyrylainen, line 404: I think we should rather try to fix the problem at the source, ie. throw an exception here for invalid parameters if fraction is negative.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -693,10 +693,10 @@ private void HandleCompoundVenting(float delta)
             {
                 amountToVent -= EjectCompound(type, amountToVent, Vector3.Back);
             }
-            else if (Compounds.GetCompoundAmount(type) > 2 * Compounds.Capacity)
+            else if (Compounds.GetCompoundAmount(type) > Compounds.Capacity)
```
hhyyrylainen, line 696: The multiplier by 2 was here to be kinder to cells that reproduce and don't unnecessarily cause a bunch of compound leakage. I don't think this should have anything to do with infinite compound venting...

src/microbe_stage/CompoundCloudPlane.cs
Oliveriver: Perhaps, but without the other change below, this will still be hit with fraction negative. So we should not do that for now.

hhyyrylainen: As I said on discord, I think this hack is fine to keep in the game for two weeks as long as there's a TODO, which I'm fine with merging at this last minute, but that other change to the 2x vent threshold I wouldn't want to merge.

I wrote some TODO list items for myself about a proper fix (which will need some safety time to verify it works):
- comment on the 2* vent code
- potentially quick fix for negative fraction in absorb, issue to properly change to exception and fix everywhere after release
- create a new "remaining space method" for compounds that has Max(0, Capacity - Amount)

src/microbe_stage/Microbe.Interior.cs
Oliveriver: It put the cell in a state where it had almost twice its capacity of glucose in its storage, which led to `float freeSpace = storage.Capacity - storage.GetCompoundAmount(compound)` in the compound gathering logic returning a negative.

The other change does guard against that, but the cell's stored compounds are still in the wrong state, which is why I added this change. I'll remove this one for now. In Buckly's save, the result is the infinite compound spawning is fixed, but the player cell has to use up 0.5 glucose before any decrease in its stored glucose is visible.

PR #3947, opened 2022-12-16
Add button to delete fossils

### LINKED ISSUE #3930: Add button to delete fossils in the museum
as there doesn't seem to be a way to do that other than going to the folder to delete things manually, seems like this is an oversight.

Brokemia:

This PR adds a button to each card in the museum to allow the player to delete that fossil.

image

Currently I'm using the trash button used in the saves list. There was some preference in the discussion under the issue towards having a button without a border, but I'm not sure which colors should be used for the pressed and hovered versions (and I think adding new textures requires dealing with LFS, which I don't have permissions for), so anyone who knows how to do that is welcome to swap that out.

**Related Issues**

Closes #3930 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/thriveopedia/fossilisation/MuseumCard.cs
```diff
@@ -88,19 +107,47 @@ private void UpdatePreviewImage()
...
+
         GUICommon.Instance.Tween.InterpolateProperty(speciesPreview, "modulate", null, Colors.White, 0.5f);
         GUICommon.Instance.Tween.Start();
     }
+
+    private void OnDeletePressed()
+    {
+        GUICommon.Instance.PlayButtonPressSound();
+
+        deleteConfirmationDialog.DialogText = TranslationServer.Translate("DELETE_FOSSIL_CONFIRMATION");
```
Kasterisk, line 145: Why not add the text directly through the inspector rather than code?

src/thriveopedia/pages/ThriveopediaMuseumPage.cs
```diff
@@ -165,4 +167,29 @@ private void TransitionToFreebuild(Species startingSpecies)
             SceneManager.Instance.SwitchToScene(editor);
         }, false);
     }
+
+    private void DeleteSpecies(MuseumCard card)
+    {
+        GUICommon.Instance.PlayButtonPressSound();
```
Kasterisk, line 173: Button press sounds should be played from their associated button press callbacks so that the functional methods doesn't have to know whether to play the sound or not. In this case this should be put in `OnDeleteConfirmPressed`.

Kasterisk (changes requested):
For the hovered and pressed states, you could override the respective styleboxes and just make a simple monochrome rectangle which can also be differentiated by opacity using `StyleBoxFlat`.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -130,6 +130,27 @@ public static FossilisedSpecies LoadSpeciesFromFile(string fossilName)
...
+        var target = Path.Combine(Constants.FOSSILISED_SPECIES_FOLDER, fossilName);
+
+        using (var directory = new Directory())
+        {
+            if (!directory.FileExists(target))
+                throw new ArgumentException("Fossil with the given name doesn't exist");
+
+            if (directory.Remove(target) != Error.Ok)
+            {
+                GD.PrintErr("Cannot delete file: ", target);
```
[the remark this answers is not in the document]
Brokemia: I copied that from `WriteDataToFossilFile`, should it also be changed there?

src/thriveopedia/fossilisation/MuseumCard.tscn
```diff
@@ -110,6 +119,42 @@ texture = ExtResource( 2 )
...
+margin_bottom = 149.0
+alignment = 2
+
+[node name="Delete" type="TextureButton" parent="MarginContainer/VBoxContainer/MarginContainer/HBoxContainer"]
+visible = false
+margin_left = 136.0
+margin_right = 166.0
+margin_bottom = 30.0
+rect_min_size = Vector2( 30, 30 )
+mouse_filter = 1
```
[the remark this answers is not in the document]
Brokemia: That makes mouse_exited get called when mousing over the delete button which makes the whole thing flash as it gets exited, and then re-entered when it hides it, and then re-exited when it shows it again. I had a lot of struggles with this particular issue, so if you have a better workaround, then by all means

src/thriveopedia/fossilisation/MuseumCard.cs
Brokemia: Copied how `ThriveopediaMuseumPage.OnOpenInFreebuildPressed()` did it, I can change that

src/thriveopedia/pages/ThriveopediaMuseumPage.cs
```diff
@@ -165,4 +167,29 @@ private void TransitionToFreebuild(Species startingSpecies)
...
+
+        var fossilName = card.FossilName;
+
+        if (fossilName == null)
```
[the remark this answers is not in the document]
Brokemia: This is an admittedly contrived case, but what if the file name actually is empty. I tested and the game will load a file just called `.thrivefossil`

src/thriveopedia/fossilisation/FossilisedSpecies.cs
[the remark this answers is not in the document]
Kasterisk: Yeah, I don't see why double logging is necessary.

src/thriveopedia/pages/ThriveopediaMuseumPage.cs
Brokemia: Fixed

src/thriveopedia/fossilisation/MuseumCard.cs
Brokemia: Fixed

src/thriveopedia/pages/ThriveopediaMuseumPage.cs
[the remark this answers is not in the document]
Kasterisk: I suppose that's okay if it wouldn't cause weird issues down the road.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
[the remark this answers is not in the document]
Brokemia: Fixed

src/thriveopedia/fossilisation/MuseumCard.cs
```diff
@@ -88,19 +107,47 @@ private void UpdatePreviewImage()
...
+        wasPressed = Pressed;
+
         GUICommon.Instance.PlayButtonPressSound();
         EmitSignal(nameof(OnSpeciesSelected), this);
     }
 
     private void OnMouseEnter()
     {
+        if (deleteButton != null)
+            deleteButton.Visible = true;
```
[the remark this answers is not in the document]
Brokemia: Fixed

src/thriveopedia/fossilisation/MuseumCard.tscn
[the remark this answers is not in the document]
Kasterisk: I see, I think I have encountered this exact problem before but can't recall what I did off the top of my head. I guess you can keep this in the meantime.

src/thriveopedia/fossilisation/MuseumCard.cs
```diff
@@ -88,19 +107,52 @@ private void UpdatePreviewImage()
...
+    private void OnDeletePressed()
+    {
+        GUICommon.Instance.PlayButtonPressSound();
+
+        deleteConfirmationDialog.PopupCenteredShrink();
+    }
+
+    private void OnDeleteConfirmPressed()
+    {
+        GUICommon.Instance.PlayButtonPressSound();
```
Kasterisk, line 154: I just remembered now that the CustomConfirmationDialog's confirm button already plays its own sound when pressed... Sorry for the mistake but this call should no longer be necessary.

Besides this minor thing, everything looks good to me.

Kasterisk (changes requested):

src/thriveopedia/fossilisation/MuseumCard.cs
Brokemia: Fixed

Kasterisk (approved):
Works good. Just need to fix that merge conflict. @Oliveriver might have an opinion on this, otherwise this is ready to go.

Brokemia:
There's still the button texture thing. I'm going to be a bit busy the next couple days, but can get to that later

src/thriveopedia/fossilisation/MuseumCard.tscn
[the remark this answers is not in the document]
Oliveriver: I'm not a fan of this approach but yeah, I'm struggling to think of a better solution.

src/thriveopedia/fossilisation/MuseumCard.cs
```diff
@@ -49,12 +64,16 @@ public class MuseumCard : Button
         }
     }
 
+    public string? FossilName { get; set; }
```
Oliveriver, line 67: We currently check for fossilised species overwriting by comparing file names, so I'm wary that using the species name instead for deleting could lead to unexpected behaviour if file names have changed to have duplicates. Admittedly that's a very unlikely scenario and a quick test suggests it doesn't cause any problems.

Maybe it's worth changing to always compare species names. That doesn't have to be done in this PR though.

Oliveriver (approved):
Some small points but nothing major.

Oliveriver:
I will mention though that what I meant by an icon in the top right of the card was more like the cross in the top right of our confirmation dialogs. That might avoid the problems with layered buttons turning invisible. Perhaps it doesn't convey "delete" quite as well though, so up to you.

src/thriveopedia/fossilisation/MuseumCard.cs
Brokemia: > I'm wary that using the species name instead for deleting could lead to unexpected behaviour if file names have changed to have duplicates

I'm not sure I understand. This PR specifically uses the file name for deleting. If that's unclear I could rename the property to FileName or something?

Oliveriver: Ah, my bad, I forgot FossilisedSpecies.Name refers to the file name rather than the species name. Don't worry about it.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -130,6 +130,25 @@ public static FossilisedSpecies LoadSpeciesFromFile(string fossilName)
         };
     }
 
+    /// <summary>
+    ///   Deletes a fossilised species by its filename.
+    /// </summary>
+    /// <param name="fossilName">The name of the .thrivefossil file (including extension)</param>
+    public static void DeleteFossilFile(string fossilName)
```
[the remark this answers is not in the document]
Brokemia: I tried to find that, and the worst that would happen is that the fossilization button might have the wrong texture, right? Doesn't seem too bad. Plus, that button already doesn't update after you fossilize a new species.

Brokemia:
image
This was done quickly and is consequently a little squashed, but does this look like how you were thinking? And would this appear on hover or be always there. I think if it's on hover, it'll still have the same issues with mouse_enter and mouse_exit, since I'm using the exact same method, so there's no difference to be considered there.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -130,6 +130,25 @@ public static FossilisedSpecies LoadSpeciesFromFile(string fossilName)
...
+        var target = Path.Combine(Constants.FOSSILISED_SPECIES_FOLDER, fossilName);
+
+        using var directory = new Directory();
+
+        if (!directory.FileExists(target))
+            throw new ArgumentException("Fossil with the given name doesn't exist");
+
+        if (directory.Remove(target) != Error.Ok)
+        {
+            throw new IOException("Cannot delete: " + target);
```
84634E1A607A, line 148: Try to give a user-friendly error dialog here or a GD.PrintErr() instead of violently throwing an exception here. As a game we want to avoid crashes.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -161,7 +180,6 @@ private static void WriteDataToFossilFile(string target, string justInfo, string
         using var file = new File();
         if (file.Open(target, File.ModeFlags.Write) != Error.Ok)
         {
-            GD.PrintErr("Cannot open file for writing: ", target);
             throw new IOException("Cannot open: " + target);
```
84634E1A607A, line 183: For here too (although you didn't write that).

src/thriveopedia/fossilisation/MuseumCard.cs
```diff
@@ -11,15 +11,30 @@ public class MuseumCard : Button
...
+    [Export]
+    public NodePath DeleteConfirmationDialogPath = null!;
+
     private Label? speciesNameLabel;
     private TextureRect? speciesPreview;
 
     private Species? savedSpecies;
     private Image? fossilPreviewImage;
 
+    private TextureButton? deleteButton;
```
84634E1A607A, line 26: You can use nullable suppression here (as is commonly used elsewhere) to avoid throwing all those SceneTreeAttachRequired thing.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
hhyyrylainen: Just this one place should not be refactored. The saving system shows how it is done: you first check that the file can be written (and show an error dialog if there is a problem), then write the file (which can fail with an exception). Or at least that's how I remember.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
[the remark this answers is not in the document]
Oliveriver: Yeah, I’m pretty sure it’s fine, just wanted to flag it in case anyone thought otherwise.

Oliveriver:
> I will mention though that what I meant by an icon in the top right of the card was more like the cross in the top right of our confirmation dialogs. That might avoid the problems with layered buttons turning invisible. Perhaps it doesn't convey "delete" quite as well though, so up to you.
> 
> 

The crosses are a bit big but yeah, that’s what I had in mind. They don’t have to only appear on hover.

As I said though, the design approach is up to you.

Oliveriver:
> image
> This was done quickly and is consequently a little squashed, but does this look like how you were thinking? And would this appear on hover or be always there. I think if it's on hover, it'll still have the same issues with mouse_enter and mouse_exit, since I'm using the exact same method, so there's no difference to be considered there.

The crosses are a bit big, but yeah, that’s what I had in mind. They don’t have to only appear on hover. As I said, the approach is up to you.

Kasterisk:
Either approach is fine I think, if without mouse_enter/exit handling make things much simpler then that's probably better.

hhyyrylainen:
I think it is much better to always have the delete icons enabled. And maybe the "x" style is not the best.
I bring these points up as we'll need to eventually have full controller support for all GUI, so designing something that only works with the mouse is a no go.

RacerBG:
In my opinion having something similar to the save deletion layout will be better if it's feasible:

hhyyrylainen:
I considered that option as well, but as clicking on a fossil makes the right side panel show its details that pattern of selecting multiple things doesn't fit that well in this case.

Brokemia:
> I bring these points up as we'll need to eventually have full controller support for all GUI, so designing something that only works with the mouse is a no go.
I think what would make the most sense for controller support is to just have a button on the controller you press to delete when a fossil is "hovered". So no actual button in the gui to delete it, just a prompt somewhere saying press X to delete.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
Brokemia: I'm looking through the save stuff a little. It seems to me that the methods in Save.cs throw exceptions, but SaveHelper.cs catches those exceptions and then handles them more elegantly. If that sounds right, I'm thinking FossilisedSpecies.cs should throw exceptions still, but ThriveopediaMuseumPage.cs should then catch those and display a popup instead. Let me know if I've missed something important

hhyyrylainen: I think that makes sense (though having written the saving code I might be a bit biased). It makes sense to me that the low level operation communicates with exceptions which the higher level operation can then convert into a GUI for notifying the user about the problem.

Brokemia: The other way I would think to do it would be through returning error codes, which I personally like a lot less.

hhyyrylainen: Returning error codes should be avoided. That's a pattern that is not used in the codebase.

hhyyrylainen:
As far as I can tell the situation with this is that this needs just a bit of graphics tweaking and then this is functionally ready?

Kasterisk:
> And maybe the "x" style is not the best. I bring these points up as we'll need to eventually have full controller support for all GUI, so designing something that only works with the mouse is a no go.

It could be handled similarly with how window dialogs would be handled in that case.

> As far as I can tell the situation with this is that this needs just a bit of graphics tweaking and then this is functionally ready?

That and a few code change request from Ajax about throwing exceptions.

@Brokemia Would be nice if we can get this merged soon!

Brokemia:
Hi, I'm back. School got in the way and I wasn't really feeling like working on this. Opened up the project again and refined the X button idea slightly

Is this still what we're doing?
image

Brokemia:
Also did a little error handling
image

hhyyrylainen:
I suppose so, getting at least an initial implementation done and merged would be nice.

One comment on the error: unless you really plan on having a bunch of different error messages with actual troubleshooting info like the full file path, it would be better to just use a custom confirmation dialog with the cancel button hidden to show the error.

Brokemia:
Alright, made that change to the error popup
image

Brokemia:
That's from code borrowed from the writing stuff, which predates my involvement. It uses Path.Combine, which uses the Windows separators on Windows machines

hhyyrylainen:
Well then showing it to the user is not good. You should first investigate using the path globalize method from Godot to convert it to one the user might be able to understand (i.e. converting the `user://` part to what it needs to be). That code is used in the steam client handling code at least (it isn't compiled by default so your IDE might not find it with a search).

Brokemia:
image
Globalizing it seems to leave in the mismatched slashes. Should I replace the Path.Combine with a manual string interpolation thingy?

hhyyrylainen:
Is it actually needed to display the path to the user? I don't think we do that for saves or anything like that. At least I don't recall how this problem was solved elsewhere.

Brokemia:
If deleting a save fails, it doesn't give any error at all, just silently doesn't do what you asked and puts a message in the log (which also has mismatched slashes).

hhyyrylainen:
Logging is fine as it doesn't need to look pretty for the end user. I'll investigate later and open an issue for the save deletion. For this feature I'll just recommend that you don't give the path to the file and just say that deleting the fossil file has failed.

Brokemia:
Currently the message is just the string from the exception message, should it get localized?

hhyyrylainen:
100%, this is a simple error condition. It would be very silly to not localize it. But how are you getting an exception? Godot file operations do not give exceptions.

Brokemia:
image
As we'd talked about up here, FossilisedSpecies.cs throws an exception if it gets an error and it gets caught by the calling class (either ThriveopediaMuseumPage.cs or FossilisationDialog.cs). It should be relatively simple to just change the exception creation to localize the string first

hhyyrylainen:
Don't do that. I won't allow exception strings to be localized. That is a recipe for nightmares when users report errors. Only when displaying the error condition should the user facing error message be generated with localized test.

Brokemia:
Should I attach the localization key/error type to the exception as a separate thing then? Or just give one generic error popup for every failure in deleting?

hhyyrylainen:
I'd say generic error unless you want to refactor the exception throwing side to have one exception type per error condition. I think that's overkill here. The only real error that can persist is that the user has disallowed the game process to write to the fossils folder *after* a fossil file has been created.

Brokemia:
Yeah, that's the most likely one. There's also the unlikely possibility that a user renamed a fossil file while the fossil screen was shown. Although in that case they probably know what they're doing anyway

hhyyrylainen (changes requested):
I think I now read the code for the first time and it looks pretty good. I have one architecture comment on how to improve this. And one slight exception style comment.

On top of those the CI is failing due to non updated localization files.

src/thriveopedia/fossilisation/MuseumCard.cs
```diff
@@ -11,15 +11,23 @@ public class MuseumCard : Button
     [Export]
     public NodePath SpeciesPreviewPath = null!;
 
+    [Export]
+    public NodePath DeleteConfirmationDialogPath = null!;
```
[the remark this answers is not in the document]
Brokemia: Would it be reasonable to have the ThriveopediaMuseumPage have a confirm delete dialog, then set a reference to it as a property on each museum card as they're created?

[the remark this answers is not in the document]
hhyyrylainen: That's closer. But I think better would be to use the save list approach where the individual items just signal that the delete button was pressed and the higher level component then shows the confirmation popup and proceeds with the delete if the user presses ok.

Brokemia:
I think I've addressed all the feedback

src/thriveopedia/pages/ThriveopediaMuseumPage.tscn
```diff
@@ -10,6 +10,7 @@
 [ext_resource path="res://src/gui_common/dialogs/CustomConfirmationDialog.tscn" type="PackedScene" id=8]
 
 [node name="ThriveopediaMuseumPage" instance=ExtResource( 1 )]
+rect_pivot_offset = Vector2( -469, -108 )
```
hhyyrylainen, line 13: This property needs to be unset (unless there's really something that must have this to function properly).

src/thriveopedia/pages/ThriveopediaMuseumPage.cs
```diff
@@ -26,14 +27,23 @@ public class ThriveopediaMuseumPage : ThriveopediaPage
...
     private Control welcomeLabel = null!;
     private VBoxContainer speciesPreviewContainer = null!;
     private SpeciesPreview speciesPreview = null!;
     private CellHexesPreview hexesPreview = null!;
     private CustomRichTextLabel speciesDetailsLabel = null!;
     private CustomConfirmationDialog leaveGameConfirmationDialog = null!;
+    private CustomConfirmationDialog deleteConfirmationDialog = null!;
+    private CustomConfirmationDialog deletionFailedDialog = null!;
     private PackedScene museumCardScene = null!;
+    private MuseumCard? cardToBeDeleted;
```
hhyyrylainen, line 46:
```suggestion

    private MuseumCard? cardToBeDeleted;
```

Should add a blank line between blocks of logically different variables (Godot vs non-Godot components).

hhyyrylainen:
Looks like there's still translation file conflicts. Do you need help with those? I could help with them and do a few very slight tweaks to the styling / text to get this finished today. I think I can get my current branch I'm working on finished today so I could then do the tiny bit of work left on this.

Brokemia:
I'm not exactly clear on how the merge process should go for translation files. I'm willing to try and figure it out, but if you want to fix it, that might be faster.

hhyyrylainen:
I mean the translation conflicts are just normal git file conflicts. Though it's usually faster to reset the translation files and redo the English changes than going through each and every file one by one and merging it manually.

image

I didn't end up having time today, but I'll definitely have time tomorrow as my PR I was working on today is ready for merging now.

src/thriveopedia/fossilisation/MuseumCard.tscn
```diff
@@ -88,6 +89,24 @@ __meta__ = {
...
 }
 
+[node name="HBoxContainer" type="HBoxContainer" parent="MarginContainer/VBoxContainer/PanelContainer/MarginContainer"]
+margin_left = 2.0
+margin_top = 2.0
+margin_right = 164.0
+margin_bottom = 19.0
+alignment = 2
+
+[node name="Delete" type="TextureButton" parent="MarginContainer/VBoxContainer/PanelContainer/MarginContainer/HBoxContainer"]
```
hhyyrylainen, line 99: I'll fix this myself, but now testing I noticed that there is now mouse hover style for this. So it is confusing that an interactable X doesn't react to mouse hover like in a CustomDialog.

hhyyrylainen (approved):
I did the merge conflict fixing and also tweaked various things + fixed the problem of there being no hover style for the fossil delete X button.

hhyyrylainen:
This is now merged. Thanks for contributing with this feature many people had asked for.

PR #3949, opened 2022-12-17
More fossilisation buttons

Oliveriver:

* Adds a button to open the fossils folder in the museum
* Adds fossilise buttons to species details in the Auto-Evo exploring tool and Thriveopedia evolutionary tree
* Extracts SpeciesDetailsPanel to a new scene

Note the newly extracted SpeciesDetailsPanel doesn't include the heading anymore. I personally prefer it without the heading, but it could be replaced or conditionally set.

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/thriveopedia/pages/ThriveopediaMuseumPage.tscn
```diff
@@ -175,5 +190,16 @@ margin_bottom = 599.0
 rect_min_size = Vector2( 391, 0 )
 WindowTitle = "CONFIRM_EXIT"
 
-[connection signal="pressed" from="HBoxContainer/SpeciesDetailPanel/MarginContainer/VBoxContainer/OpenInFreebuildButton" to="." method="OnOpenInFreebuildPressed"]
+[node name="NoDirectoryDialog" parent="." index="3" instance=ExtResource( 8 )]
+margin_left = 340.0
+margin_top = 360.0
+margin_right = 940.0
+margin_bottom = 360.0
+rect_min_size = Vector2( 600, 0 )
```
Kasterisk, line 198: The dialog box is too wide, I think lowering this to 300 would be good.

Kasterisk (changes requested):

src/thriveopedia/pages/ThriveopediaMuseumPage.tscn
Oliveriver: Done. It does look better.

Kasterisk (approved):
With that one thing fixed, nothing seems out of ordinary.

src/gui_common/SpeciesDetailsPanel.cs
```diff
@@ -0,0 +1,79 @@
...
+    private FossilisationDialog fossilisationDialog = null!;
+
+    private Species? previewSpecies;
+
+    public Species? PreviewSpecies
+    {
+        get => previewSpecies;
+        set
+        {
+            previewSpecies = value;
```
hhyyrylainen, line 34: This should be rewritten to work when set even before this is attached to the scene tree.

src/gui_common/SpeciesDetailsPanel.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        if (PreviewSpecies is MicrobeSpecies microbeSpecies)
+        {
+            hexesPreview.PreviewSpecies = microbeSpecies;
+            fossilisationButton.Disabled = false;
+        }
+        else
+        {
+            fossilisationButton.Disabled = true;
+            GD.PrintErr("Unknown species type to preview: ", PreviewSpecies);
```
hhyyrylainen, line 65: This print will happen when the species is null. Is it intended to allow setting the species to preview to null?

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -2218,6 +2218,7 @@ __meta__ = {
 }
 
 [node name="PauseMenu" parent="." instance=ExtResource( 11 )]
+visible = true
```
hhyyrylainen, line 2221: Accidental change here again.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -723,38 +713,16 @@ private void SpeciesListMenuIndexChanged(int index)
         var speciesName = speciesListMenu.Popup.GetItemText(index);
         var species = speciesHistoryList[generationDisplayed].Values.First(p => p.FormattedName == speciesName);
 
-        if (species == speciesPreview.PreviewSpecies)
+        if (species == speciesDetailsPanel.PreviewSpecies)
```
hhyyrylainen, line 716: This check should be put in the PreviewSpecies property so that it skips applying the GUI control changes if the new value is the same as the old one.

hhyyrylainen (changes requested):
A few comments regarding how I think the code could be improved and I spotted an accidental change as well that should be reverted.

src/gui_common/SpeciesDetailsPanel.cs
Oliveriver: This follows the existing pattern of `SpeciesPreview`.

src/gui_common/SpeciesDetailsPanel.cs
Oliveriver: I've used a `ready` bool but I seem to remember you requesting a different approach to this in the past. I can't remember what that was. Let me know if you had a better approach in mind.

hhyyrylainen: Make the control nullable, then check in the update method if it is null, and if it is then skip. And finally also add a call to that update method at the end of _Ready

hhyyrylainen (approved):
All my code related concerns have been addressed. I've not playtested.

hhyyrylainen:
This is good to merge once the translation merge conflicts are resolved.

Oliveriver:
Had some issues with submodule files being unexpectedly modified during the merge. The diff suggests I've put everything back to how it was, but you might want to double check when you merge this into master.

hhyyrylainen:
As far as github shows anything, this PR doesn't change the git submodules.

hhyyrylainen:
I noticed this PR removes the species details text entirely? What's the reason behind that. I see it used to be in the auto-evo exploring tool so @84634E1A607A might also want to take a look at this.

src/gui_common/SpeciesDetailsPanel.cs
```diff
@@ -0,0 +1,93 @@
...
+        if (previewSpecies != null)
+            UpdateSpeciesPreview();
+    }
+
+    /// <summary>
+    ///   Updates displayed species information based on the set preview species.
+    /// </summary>
+    private void UpdateSpeciesPreview()
+    {
+        if (speciesPreview == null || hexesPreview == null || fossilisationButton == null ||
```
84634E1A607A, line 63: You could just check one variable. (And I think the 'ready' approach is used commonly elsewhere?)

src/gui_common/SpeciesDetailsPanel.tscn
```diff
@@ -0,0 +1,101 @@
...
+[ext_resource path="res://src/gui_common/thrive_theme.tres" type="Theme" id=1]
+[ext_resource path="res://src/gui_common/SpeciesDetailsPanel.cs" type="Script" id=2]
+[ext_resource path="res://src/gui_common/fonts/Lato-Bold-AlmostSmaller.tres" type="DynamicFont" id=3]
+[ext_resource path="res://src/gui_common/PhotographablePreview.tscn" type="PackedScene" id=4]
+[ext_resource path="res://src/gui_common/CustomRichTextLabel.tscn" type="PackedScene" id=5]
+[ext_resource path="res://src/gui_common/SpeciesPreview.cs" type="Script" id=6]
+[ext_resource path="res://src/microbe_stage/CellHexesPreview.cs" type="Script" id=7]
+[ext_resource path="res://src/thriveopedia/fossilisation/FossilisationDialog.tscn" type="PackedScene" id=8]
+
+[node name="SpeciesDetailsPanel" type="PanelContainer"]
```
84634E1A607A, line 12: I think this should be a `Control` so that it can be added in any Container. The current design will result in fixed margin and so on. It may create GUI designing issues.

src/auto-evo/AutoEvoExploringTool.tscn
```diff
@@ -762,80 +758,7 @@ margin_bottom = 486.0
 size_flags_horizontal = 3
 size_flags_vertical = 3
 
-[node name="SpeciesDetailPanel" type="PanelContainer" parent="GUI/VBoxContainer/HistoryReportSplit/Viewer/HBoxContainer"]
```
84634E1A607A, line 765: Here you should preserve the PanelContainer, MarginContainer, VBoxContainer, MarginContainer and the Label.

src/auto-evo/AutoEvoExploringTool.tscn
```diff
@@ -762,80 +758,7 @@ margin_bottom = 486.0
...
-custom_constants/margin_left = 10
-custom_constants/margin_bottom = 10
-
-[node name="VBoxContainer" type="VBoxContainer" parent="GUI/VBoxContainer/HistoryReportSplit/Viewer/HBoxContainer/SpeciesDetailPanel/MarginContainer"]
-margin_left = 10.0
-margin_top = 10.0
-margin_right = 340.0
-margin_bottom = 496.0
-
-[node name="Label" type="Label" parent="GUI/VBoxContainer/HistoryReportSplit/Viewer/HBoxContainer/SpeciesDetailPanel/MarginContainer/VBoxContainer"]
```
84634E1A607A, line 787: This label.

84634E1A607A (changes requested):
Code logic is correct, but I have one GUI designing suggestion.

src/gui_common/SpeciesDetailsPanel.cs
hhyyrylainen: I had the same thought but didn't ask for that as I think I've checked up to two variables like this... so checking 4 seems pretty minor, but I think just making one of the fields nullable would be a slight code improvement here.

Oliveriver: Do you have a preference for which one field should be nullable then?

src/auto-evo/AutoEvoExploringTool.tscn
Oliveriver: I mentioned that label in the original post. I can leave it in if you would prefer, although I think it looks better in both places without it.

src/gui_common/SpeciesDetailsPanel.cs
hhyyrylainen: I'd personally just pick the first of these fields.

src/auto-evo/AutoEvoExploringTool.tscn
84634E1A607A: I'm unsure about it, but I suppose it is fine then. However, I still want the SpeciesDetailsPanel to be a Control.

src/auto-evo/AutoEvoExploringTool.tscn
Oliveriver: The FossilisationButton exists outside of the second MarginContainer, so I can't move that out of the SpeciesDetailsPanel (the entire point of creating it was to have common fossilisation logic). Preserving the other three seems to prevent the FossilisationDialog showing, again for reasons I don't have time to figure out.

84634E1A607A:
I'll give it a try.

hhyyrylainen:
I forgot I had already reviewed the code of this PR. So that means this is just waiting for @Oliveriver's opinion on #3982

hhyyrylainen (approved):
I think the new changes from the PR merged into this are probably fine, so just need to solve the merge conflicts and then this can be merged to master.

hhyyrylainen:
I'll merge this once the tests pass.

Issue #3944, opened 2022-12-16
Refactor Thriveopedia to no longer use Godot meta feature


hhyyrylainen:
instead use C# properties like everywhere else in the game

https://github.com/Revolutionary-Games/Thrive/issues/3930#issuecomment-1354308771

PR #3950, opened 2022-12-17
Remove use of meta in Thriveopedia

Oliveriver:

Changes indexing of pages in the Thriveopedia page tree to avoid use of Godot meta.

I think this was a holdover from before a refactor I did partway through the Thriveopedia PR. Due to the constraints on Godot's `PageTree` and `TreeItem`, I believe making a custom version of `TreeItem` isn't possible, so we can't use C# properties. Besides, I had a dictionary of Thriveopedia pages and items lying around anyway, so it's easy to use that.

**Related Issues**

Closes https://github.com/Revolutionary-Games/Thrive/issues/3944

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/thriveopedia/Thriveopedia.cs
+++ b/src/thriveopedia/Thriveopedia.cs
@@ -289,7 +289,8 @@ public class Thriveopedia : ControlWithInput
 
     private void OnPageSelectedFromPageTree()
     {
-        var name = allPages.First(p => p.Value == pageTree.GetSelected()).Key.PageName;
+        var selected = pageTree.GetSelected();
+        var name = allPages.First(p => p.Value == selected).Key.PageName;
         ChangePage(name);
     }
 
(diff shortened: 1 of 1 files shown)


84634E1A607A:
First, (My opinion), setMeta is the easiest way.

Second, I'd split the Godot native call into another line to avoid multiple identical calls and save time.

Oliveriver:
Good call on splitting out the native Godot call. I guess we'll see what the consensus is on whether this PR is needed.

hhyyrylainen:
I still think we shouldn't use Godot inbuilt feature in a couple of places. Everything else is nicely made with C# classes and properties. If we wanted to we could hack almost everything to use the meta properties in Godot for storing our custom classes data, but that wouldn't be a good idea.

Kasterisk (approved):

Issue #3861, opened 2022-11-15
Add graceful fossil error handling


Oliveriver:
Currently, if a player has an invalid .thrivefossil file in their fossils folder, the entire game crashes when opening the Thriveopedia. Not really ideal.

It might be better to catch this error and either skip the fossil in the museum list or have its museum card tell the player the fossil is corrupt. The latter would be much more work and to be honest, I don't think it's worth it. Just ignoring that fossil entirely should be fine.

PR #3951, opened 2022-12-17
Graceful fossil error handling

Oliveriver:

Ignores corrupt fossil files to prevent crashes when opening the Thriveopedia.

I went for the nuclear option of just wrapping the whole loading process in a try-catch block and printing the error. That may lead to error spam if a user with a corrupt fossil repeatedly opens and closes the Thriveopedia.

**Related Issues**

Closes https://github.com/Revolutionary-Games/Thrive/issues/3861

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/thriveopedia/fossilisation/FossilisedSpecies.cs
+++ b/src/thriveopedia/fossilisation/FossilisedSpecies.cs
@@ -118,7 +118,7 @@ public class FossilisedSpecies
     ///   Loads a fossilised species by its filename.
     /// </summary>
     /// <param name="fossilName">The name of the .thrivefossil file (including extension)</param>
-    /// <returns>The species saved in the provided file</returns>
+    /// <returns>The species saved in the provided file, or null if the file doesn't exist or is corrupt</returns>
     public static FossilisedSpecies? LoadSpeciesFromFile(string fossilName)
     {
         var target = Path.Combine(Constants.FOSSILISED_SPECIES_FOLDER, fossilName);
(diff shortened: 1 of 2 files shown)


src/thriveopedia/fossilisation/FossilisedSpecies.cs
```diff
@@ -119,15 +119,24 @@ public static bool IsSpeciesAlreadyFossilised(string name, List<string>? existin
     /// </summary>
     /// <param name="fossilName">The name of the .thrivefossil file (including extension)</param>
     /// <returns>The species saved in the provided file</returns>
```
Kasterisk, line 121: Can you update the `returns` documentation please?

Kasterisk (changes requested):
I still like the approach of allowing corrupt entries just for the sake of telling players they have corrupt fossils but, this is good enough I guess.

src/thriveopedia/fossilisation/FossilisedSpecies.cs
Oliveriver: Done.

Oliveriver:
> I still like the approach of allowing corrupt entries just for the sake of telling players they have corrupt fossils but, this is good enough I guess.

It would certainly be possible to update that in future using a null check to display something else.

Kasterisk (approved):
That's good, as for now, this looks fine to me

hhyyrylainen (approved):

Issue #3892, opened 2022-12-01
Update light levels only 60 times per second


hhyyrylainen:
(or maybe even less) to reduce the amount of useless processing that the day/night code is doing

PR #3952, opened 2022-12-17
Reduce frequency of light level updates

Oliveriver:

Reduces the frequency of light level and other biome data updates to 10 times a second.=.

**Related Issues**

Closes https://github.com/Revolutionary-Games/Thrive/issues/3892

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/MicrobeStage.cs
+++ b/src/microbe_stage/MicrobeStage.cs
@@ -44,6 +44,12 @@ public class MicrobeStage : StageBase<Microbe>
     [JsonProperty]
     private float elapsedSinceEntityPositionCheck;
 
+    /// <summary>
+    ///   Used to control how often light level is updated by the day/night cycle.
+    /// </summary>
+    [JsonProperty]
+    private float elapsedSinceLightLevelUpdate;
+
     [JsonProperty]
     private bool wonOnce;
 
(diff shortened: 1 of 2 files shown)


src/microbe_stage/MicrobeStage.cs
```diff
@@ -44,6 +44,12 @@ public class MicrobeStage : StageBase<Microbe>
     [JsonProperty]
     private float elapsedSinceEntityPositionCheck;
 
+    /// <summary>
+    ///   Used to control how often light level is updated by the day/night cycle.
+    /// </summary>
+    [JsonProperty]
```
84634E1A607A, line 50: Why not JsonIgnore?

Plus I'd prefer it to be named as timeElapsedSinceLastLightLevelUpdate

src/microbe_stage/MicrobeStage.cs
```diff
@@ -208,10 +214,19 @@ public override void _Process(float delta)
...
-        patchManager.UpdateAllPatchLightLevels();
+        elapsedSinceLightLevelUpdate += delta;
+        if (elapsedSinceLightLevelUpdate > Constants.LIGHT_LEVEL_UPDATE_INTERVAL)
+        {
+            elapsedSinceLightLevelUpdate = 0;
+
+            if (GameWorld.Map.CurrentPatch != null)
+            {
+                patchManager.UpdatePatchBiome(GameWorld.Map.CurrentPatch);
+                patchManager.UpdateAllPatchLightLevels();
```
84634E1A607A, line 225: Why is here a change of logic?

84634E1A607A (changes requested):
Good approach overall but I have two questions.

src/microbe_stage/MicrobeStage.cs
Oliveriver: Both of these match `elapsedSinceEntityPositionCheck` above, so that should be changed too if this is the decision.

src/microbe_stage/MicrobeStage.cs
Oliveriver: These were all marked as needing to be updated less frequently and, as far as I can tell, all require an existing `CurrentPatch`. The order of execution is the same as before.

src/microbe_stage/MicrobeStage.cs
84634E1A607A: Fine then.

src/microbe_stage/MicrobeStage.cs
84634E1A607A: Okay. Just to make sure because it seems that UpdateAllPatchLightLevels() will be called even there isn't a current patch.

84634E1A607A (approved):

Kasterisk (approved):

Kasterisk:
Works fine.

hhyyrylainen:
Opened a follow up issue: https://github.com/Revolutionary-Games/Thrive/issues/3970

Issue #3898, opened 2022-12-02
Process panel stops updating properly with specific changes to the player


hhyyrylainen:
in the day/night cycle PR it was there's some issue regarding the process panel stopping updating, and it was checked to also happen in the master branch (before that PR was merged).

hhyyrylainen:
Seems like dying or going to the editor (reproducing) can trigger this condition so I adjusted the title. There's probably something wrong with the game not properly keeping the player cell feeding new data to the process tracking object.

PR #3955, opened 2022-12-18
Fix process panel not updating after cell reproduces

Oliveriver:

Recreates `ProcessStatistics` when processes need refreshing to fix the process panel freezing once the player reproduces.

I'll admit, I'm not 100% sure why this works and feel there must a better way. But it's definitely something in this area. I won't be surprised if this change is rejected in favour of something else.

**Related Issues**

Fixes https://github.com/Revolutionary-Games/Thrive/issues/3898

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Microbe.cs
+++ b/src/microbe_stage/Microbe.cs
@@ -1220,6 +1220,9 @@ public partial class Microbe : RigidBody, ISpawned, IProcessable, IMicrobeAI, IS
             }
         }
 
+        // Recreate tracking for running processes
+        ProcessStatistics = new ProcessStatistics();
+
         processesDirty = false;
     }
 


hhyyrylainen:
Can't say I'm a huge fan... As this would lose collected info whenever organelles duplicate (I think). There's still plenty of time untile 0.6.1. 

Even if that is not the case my grand goal for the process statistics object was that it could even collect multi minute (for example whole day cycle worth of data) sequences of data, which would be impacted by recreating the statistics. It's probably a case of where the statistics object doesn't get updated / passed around the systems correctly anymore due to one part moving to a different object while the other side is forgotten to be updated.

That's all I have to say on this for now.

Oliveriver:
Yeah, the problem is definitely that `ProcessStatistics` isn't updating properly when `ActiveProcesses` is refreshed. How to fix that is debatable.

hhyyrylainen:
My hunch was that the right data wasn't updated anymore, and sure enough it seems like when I fixed up a bunch of equality comparisons things work now: https://github.com/Revolutionary-Games/Thrive/pull/3971

Issue #3852, opened 2022-11-07
Refactor evolutionary tree in Thriveopedia to have less duplicated code with the exploring tool


hhyyrylainen:
as a follow up to: https://github.com/Revolutionary-Games/Thrive/pull/3755

PR #3963, opened 2022-12-29
Auto-evo exploring tool part II

### LINKED ISSUE #3626: Add button to run multiple worlds in a row in auto-evo exploring tool and collect statistics
there needs to be a feature to do a multi run (run up to generation 25 or so, then restart with a new start) and general multiple run statistics for auto-evo exploring tool. 

Slightly modified description from the earlier issue about the auto-evo exploring tool:
A key feature for developing auto-evo with this tool would be to have a mode to run like 25 auto-evo runs in a row and collect some overall statistics.

To compare multiple different runs (with different code) it'd be nice to be able to save the collected statistics to a file for later viewing. That file could also have a dump of some extra data that would be too much to show in the GUI.

### LINKED ISSUE #3862: Export Auto-Evo Explorer Data
While the in-game tool is very nice, the ability to easily export the data and analyze it in other software would be beneficial. If nothing else, just being able to freely create graphs from it, without dealing with godot, would go a long way to help understand the systems we have in place.

### LINKED ISSUE #3962: Make evolutionary tree able to handle multiple root nodes
So that it works correctly in freebuild mode where multiple randomly generated species are added.

84634E1A607A:
Freebuild mode support for EvolutionaryTree & multi-world view for AutoEvoExploringTool

**Related Issues**

Fixes #3962 
Resolves #3626 
Resolves #3862 
Part of #3852 (Not much of it)

**Progress Checklist**

- [x] Added freebuild mode support to thriveopedia.
- [x] Fixed a bug introduced by Oliveriver who removes a node from evolutionary tree and queue-freed it when it can be some other one's parent (still referenced).
- [x] Add world view to auto-evo exploring tool
- [x] Add world exporting system
- [x] Add multi-world exporting system
- [x] Check all GUI logic to avoid bugs
      break existing features:
      members reviewing this PR)


src/auto-evo/EvolutionaryTree.cs
```diff
@@ -245,29 +252,33 @@ public override void _Process(float delta)
     private void SetupTreeNode(Species species, EvolutionaryTreeNode? parent, int generation,
         bool isLastGeneration = false)
     {
-        var node = treeNodeScene.Instance<EvolutionaryTreeNode>();
+        var position = new Vector2(generation * GENERATION_SEPARATION, 0);
+
+        // Make sure the next line won't end in an exception.
```
Oliveriver, line 257: This comment could be a bit clearer.

Oliveriver (dismissed):
Some small changes and suggestions, but largely this seems sensible and works well. Nice!

src/auto-evo/EvolutionaryTree.cs
84634E1A607A: That's a leftover comment that I forgot to delete.

buckly90:
Everything looks fantastic! This will hugely improve our ability to measure the preferences of autoevo and the impacts our changes make!

I've only noticed one problem with it during testing;

When running worlds, it will first run the set amount of generations in the selected pre-existing world before moving on to create new worlds.
There also does not appear to be a way to run a set amount of generations across  all worlds at once, so I must switch between each world and do it manually instead which is pretty tedious.

Other than that, excellent work! I'm very excited to see this merged soon.

Oliveriver:
I'll try and fit in a code review shortly, but here are my thoughts after testing.

* Really nice work on the freebuild support and multi-world options overall! It's great to have a full statistical view of Auto-Evo for the first time. It's already quite alarming to see several organelles just don't evolve...
* The plus button to add new worlds doesn't work for me. I get a method not found error.
* The export success panel is too narrow, and I think the window text should be 'Export Success' rather than 'Exportation Success'.
* I agree with Buckly that the logic for the run x generations and run y worlds buttons doesn't always match up.
* Not essential or anything, but it would be nice for the list of organelles to be sorted by percentage prevalance.
* 'Most Potential Species' doesn't really make sense. I'm assuming that 'Most Populous' would be better.

84634E1A607A:
> When running worlds, it will first run the set amount of generations in the selected pre-existing world before moving on to create new worlds.
> There also does not appear to be a way to run a set amount of generations across all worlds at once, so I must switch between each world and do it manually instead which is pretty tedious.

> I agree with Buckly that the logic for the run x generations and run y worlds buttons doesn't always match up.

My design is that, due to the fact that the initial generation of a world is *0*, you can directly press the *run worlds* button and it will create evenly auto-evoed worlds. The spin box aside *run generations* button will set the generations to run each world. It looks like a tool-tip is needed to clarify the design.

> The plus button to add new worlds doesn't work for me. I get a method not found error.

It should be properly connected. Maybe you should restart your Godot and run git restore.

> 'Most Potential Species' doesn't really make sense. I'm assuming that 'Most Populous' would be better.
> Not essential or anything, but it would be nice for the list of organelles to be sorted by percentage prevalance.
> The export success panel is too narrow, and I think the window text should be 'Export Success' rather than 'Exportation Success'.

I will make changes soon.

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -136,6 +136,13 @@ public class EvolutionaryTree : Control
     [Signal]
     public delegate void SpeciesSelected(int generation, uint id);
 
+    /// <summary>
+    ///   Make use of tree-generated data to help exporting.
```
hhyyrylainen, line 140:
```suggestion
    ///   Allows access to tree-generated data. This helps with data exporting, for example.
```
Would that be clearer interpretation of what this comment is saying?

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -173,6 +187,9 @@ public void Clear()
         speciesNames.Clear();
         generationTimes.Clear();
         speciesNodes.Clear();
+        maxSpeciesId = 0;
+        latestGeneration = 0;
+        dragOffset = Vector2.Zero;
```
hhyyrylainen, line 192: If the user is currently dragging should this also reset the dragging status?

src/auto-evo/EvolutionaryTree.cs
```diff
@@ -156,14 +163,21 @@ public override void _Ready()
         latoSmallRegular = (Font)GD.Load("res://src/gui_common/fonts/Lato-Regular-Small.tres");
     }
 
-    public void Init(Species luca, string? updatedLUCAName = null)
+    public void Init(IEnumerable<Species> initialSpecies, uint playerSpeciesId = 1,
+        string? updatedPlayerSpeciesName = null)
     {
-        SetupTreeNode(luca, null, 0);
+        foreach (var species in initialSpecies)
```
hhyyrylainen, line 169: What if initial species is empty? Does this tree work correctly or should this throw an exception?

project.godot
```diff
@@ -442,6 +442,10 @@ file_logging/max_log_files=8
 
 limits/message_queue/max_size_kb=8192
 
+[mono]
+
+project/assembly_name="Thrive"
```
hhyyrylainen, line 447: I think I've seen this change happen a lot. I guess it is finally time to allow this, I'll assume that this does not break anything as this seems like just the default value is now added.

src/auto-evo/EvolutionaryTree.cs
84634E1A607A: I don't think so. I think dragging status should be preserved.

src/auto-evo/EvolutionaryTree.cs
84634E1A607A: The tree will not work correctly, but I can't figure out an elegant way to check if the IEnumerable is empty.

hhyyrylainen: I recommend adding a bool called "seenASpecies = false" that gets set to true inside the loop, then after the loop throw if the variable is false.

src/auto-evo/EvolutionaryTree.cs
hhyyrylainen: But doesn't that then mostly negate the drag offset? Because if the user has dragged an entire screen, the tree is rebuilt, then the user moves the mouse again, this line basically has no effect as the drag offset will jump a ton at once.

src/auto-evo/AutoEvoExploringTool.Export.cs
```diff
@@ -0,0 +1,185 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using Godot;
+using Path = System.IO.Path;
+
+/// <summary>
+///   Partial class: Export
```
hhyyrylainen, line 9:
```suggestion
///   Partial class: Data export functionality of the auto-evo exploring tool
```

src/auto-evo/AutoEvoExploringTool.Export.cs
```diff
@@ -0,0 +1,185 @@
...
+/// </summary>
+public partial class AutoEvoExploringTool
+{
+    private void ExportWorlds()
+    {
+        var previousWorld = worldsList.FindIndex(w => ReferenceEquals(w, world));
```
hhyyrylainen, line 15: Why is reference equality used here? Why not just find by the value directly, surely the GameWorld Equals operation is suitable?

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -2,17 +2,22 @@
...
 using System.Linq;
+using AutoEvo;
 using Godot;
 
 /// <summary>
-///   Auto-evo exploring tool is a scene that contains numerous tool to debug, test, and explore auto-evo.
+///   Auto-evo exploring tool is a scene that contains numerous tools to debug, test, and explore auto-evo.
 ///   You can change the parameters to see how auto-evo responds, get a clearer view of how species evolves,
 ///   and start a game based on the simulation results.
+///   Partial class: GUI, auto-evo, world control
```
hhyyrylainen, line 12: This line should be in a remark section.

src/auto-evo/EvolutionaryTree.cs
84634E1A607A: No. The drag offset won't jump a ton at once because every mouse move directly changes DragOffset a little - The later dragging will still be smooth.

hhyyrylainen: That's good then. I was thinking that this was the variable tracking where the cursor started the drag, but that's obviously a different variable.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -398,52 +429,41 @@ public void AskExit()
...
-        biodiversitySplitIsMutatedCheckBox.Pressed = autoEvoConfiguration.BiodiversityNearbyPatchIsFreePopulation;
-        biodiversityNearbyPatchIsFreePopulationCheckBox.Pressed = autoEvoConfiguration.BiodiversitySplitIsMutated;
-        lowBiodiversityLimitSpinBox.Value = autoEvoConfiguration.LowBiodiversityLimit;
-        maximumSpeciesInPatchSpinBox.Value = autoEvoConfiguration.MaximumSpeciesInPatch;
-        moveAttemptsPerSpeciesSpinBox.Value = autoEvoConfiguration.MoveAttemptsPerSpecies;
-        mutationsPerSpeciesSpinBox.Value = autoEvoConfiguration.MutationsPerSpecies;
+        worldsList.Add(new AutoEvoExploringToolWorld(configuration));
+        WorldsListMenuIndexChanged(worldsList.Count - 1);
+
+        worldsListMenu.AddItem((worldsList.Count - 1).ToString(), false, Colors.White);
```
hhyyrylainen, line 437: CurrentCulture should probably be used in the ToString call here? I assume that's meant to be user readable.

src/auto-evo/AutoEvoExploringTool.cs
84634E1A607A: Microbe.cs partial class is written like this. https://github.com/84634E1A607A/Thrive/blob/1e68bd326b91f5d8c217277479098d25ca11a4fa/src/microbe_stage/Microbe.cs#L9

hhyyrylainen:
I can't comment directly on the submodule file, but I think this PR is downgrading the submodule. `454ce398d9ca8d81748201cbae84a2a027181477` is newer commit than in this PR.

src/auto-evo/AutoEvoExploringTool.cs
hhyyrylainen: That's also not the best style in my eyes...  but here as the summary is already so long, putting the extra info outside the summary would be even better.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -692,10 +755,53 @@ private void OnAbortButtonPressed()
...
+        world = worldsList[index];
+        worldsListMenu.Text = index.ToString();
+
+        generationDisplayed = world.CurrentGeneration;
+
+        // Rebuild history list
+        historyListMenu.ClearAllItems();
+        for (int i = 0; i <= world.CurrentGeneration; ++i)
+        {
+            historyListMenu.AddItem(i.ToString(), false, Colors.White);
```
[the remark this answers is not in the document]
84634E1A607A: There are 4 ToString s I have changed locally.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -798,14 +904,212 @@ private void PlayWithCurrentSettingPressed()
...
+            populationPerPatchStandardDeviation.ToString("F2", CultureInfo.CurrentCulture),
+            microbeSpeciesHexSizeAverage.ToString("F2", CultureInfo.CurrentCulture),
+            microbeSpeciesHexSizeStandardDeviation.ToString("F2", CultureInfo.CurrentCulture));
+
+        foreach (var organelle in SimulationParameters.Instance.GetAllOrganelles())
+        {
+            var percentage = worldsList.Average(w => w.MicrobeSpeciesOrganelleStatistics[organelle].Percentage);
+            var average = worldsList.Average(w => w.MicrobeSpeciesOrganelleStatistics[organelle].Average);
+            bbcode += TranslationServer.Translate("MICROBE_ORGANELLE_STATISTICS").FormatSafe(
+                organelle.Name.Replace('\n', ' '),
```
hhyyrylainen, line 991: I've seen this `Name.Replace(...)` in quite a few places now, so I think adding a new get only property in the organelle class named something like `NameWithoutSpecialCharacters` would be better to keep the logic in one place in case it ever needs to be updated to handle more special characters.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -798,14 +904,212 @@ private void PlayWithCurrentSettingPressed()
...
+        public AutoEvoExploringToolWorld(IAutoEvoConfiguration configuration)
+        {
+            AutoEvoConfiguration = configuration.Clone();
+            GameProperties = GameProperties.StartNewMicrobeGame(new WorldGenerationSettings
+                { AutoEvoConfiguration = AutoEvoConfiguration });
+            SpeciesHistoryList = new List<Dictionary<uint, Species>>();
+            PatchHistoryList = new List<Dictionary<int, PatchSnapshot>>();
+            RunResultsList = new List<LocalizedStringBuilder>();
+            CurrentGeneration = 0;
+            TotalTimeUsed = TimeSpan.Zero;
```
hhyyrylainen, line 1055: These initial values should be specified directly in the fields above instead of the constructor as that's cleaner in my opinion and leaves only the logic that truly needs to be in the constructor in the constructor.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -798,14 +904,212 @@ private void PlayWithCurrentSettingPressed()
...
+
+        public double PatchSpeciesCountStandardDeviation { get; private set; }
+
+        public long TotalPopulation { get; private set; }
+
+        public double MicrobeSpeciesAverageHexSize { get; private set; }
+
+        public void UpdateWorldStatistics()
+        {
+            TotalSpeciesCount = (int)SpeciesHistoryList.Last().Values.Max(s => s.ID);
```
hhyyrylainen, line 1098: Shouldn't this use "Values.Count" so that this is actually accurate? The IDs aren't absolutely guaranteed to equal the number of species, so I'd avoid making that assumption in any code.

src/auto-evo/AutoEvoExploringTool.cs
84634E1A607A: I really want to ask why toxin valcume has a line break in it. Other organelles whose names are even longer are written in one line.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -798,14 +904,212 @@ private void PlayWithCurrentSettingPressed()
...
+            CurrentSpeciesCount = SpeciesHistoryList.Last().Values.Count;
+            PatchSpeciesCountAverage = GameProperties.GameWorld.Map.Patches.Values.Average(p => p.SpeciesInPatch.Count);
+            PatchSpeciesCountStandardDeviation = Math.Sqrt(GameProperties.GameWorld.Map.Patches.Values
+                .Select(p => p.SpeciesInPatch.Count)
+                .Average(c => (c - PatchSpeciesCountAverage) * (c - PatchSpeciesCountAverage)));
```
hhyyrylainen, line 1103: I see this standard deviation calculation quite a bit. I think it might be a bit nicer if MathUtils was updated to have a helper method for calculating the standard deviation and average.

src/auto-evo/AutoEvoExploringTool.cs
84634E1A607A: The list only contains species that are still alive. Its count can be significantly less than the actual count.

src/auto-evo/AutoEvoExploringTool.cs
84634E1A607A: I actually tried to, but I failed because int, uint, float and double can be calculated.

src/auto-evo/AutoEvoExploringTool.cs
hhyyrylainen: It's purely for English text in the organelle selection buttons to look nicer. It'd be great if Godot had a space line break mode, but it doesn't (or didn't back then) so the next best thing was to force the line break at a known location.

If there's any other solutions to the GUI (that would also apply to other languages automatically) that would be great to find.

src/auto-evo/AutoEvoExploringTool.cs
hhyyrylainen: But then the maximum ID number is definitely not equal to the "count". Should the variable be perhaps named something else if it doesn't actually have the count in it?

src/auto-evo/AutoEvoExploringTool.cs
hhyyrylainen: I think the method would be just about 5-ish lines? So copy-pasting that 3 times is not too bad. I've had to use the same approach for other math helpers as C# sadly doesn't have a generic numeric support.

src/auto-evo/AutoEvoExploringTool.cs
hhyyrylainen: That might be true but fundamentally I think the ID should not be *assumed* to be incremental. The code will be much more robust if actual counting is performed here instead of an assumption that might get broken sometime in the future.

hhyyrylainen:
I did a quick playtest and everything seems mostly fine, I just have two comments:

Exiting and entering the exploring tool again from the main menu causes this error:
```
ERROR: Can't add child 'DefaultTooltip' to 'default', already has a parent 'default'.
   at: add_child (scene/main/node.cpp:1282)
```

Pressing the "+" button next to the world causes an error:
```
ERROR: Error calling method from signal 'pressed': 'Node(AutoEvoExploringTool.cs)::InitNewWorld': Method not found..
   at: emit_signal (core/object.cpp:1242)
```

also it should have a tooltip, as it is not very obvious what the button does to me

locale/en.po
```diff
@@ -1017,161 +1017,231 @@ msgstr "Cilia"
...
 msgstr "Lysosome"
 
-#: ../src/auto-evo/AutoEvoExploringTool.cs:588
+#: ../src/auto-evo/AutoEvoExploringTool.Export.cs:34
+msgid "WORLD_EXPORT_SUCCESS_MESSAGE"
+msgstr "Worlds' data have been successfully exported to {0}"
+
+#: ../src/auto-evo/AutoEvoExploringTool.cs:368
+msgid "RUN_X_WORLDS_TOOLTIP"
+msgstr "Run a certain number of worlds, the amount of generations for each world to run is determined by the previous spin."
```
hhyyrylainen, line 1026:
```suggestion
msgstr "Run a certain number of worlds, the amount of generations for each world to run is determined by the above spin box."
```

84634E1A607A:
I checkout RevolutionaryGamesCommon:main and it still have some changes listed. I don't know why.

hhyyrylainen:
> I checkout RevolutionaryGamesCommon:main and it still have some changes listed. I don't know why.

That's because Thrive master branch didn't have the newest version. I checked now and this PR has the latest common module version so this now *updates* the module, which is absolutely fine in my opinion.

84634E1A607A:
Got it working.

src/auto-evo/AutoEvoExploringTool.cs
hhyyrylainen: I opened a separate issue about doing that properly: https://github.com/Revolutionary-Games/Thrive/issues/3985

84634E1A607A:
What about the performance impact of multiple enumeration? Such as when using OrderBy or combining multiple LINQ statements? And, could multiple-enumeration crash the game?

locale/en.po
```diff
@@ -1017,161 +1017,239 @@ msgstr "Cilia"
...
+msgid "WORLD_GENERAL_STATISTICS"
+msgstr "Generic Statistics of Current World"
+
+#: ../src/auto-evo/AutoEvoExploringTool.tscn:269
+msgid "CURRENT_WORLD"
+msgstr "Current World"
+
+#: ../src/auto-evo/AutoEvoExploringTool.tscn:289
+msgid "INIT_NEW_WORLD_TOOLTIP"
+msgstr "Init a new world"
```
hhyyrylainen, line 1121: I think this should use the world "create" instead of init as this is user readable text (initialize could be an alternative but I think create would be clearer to players).

This is the last thing I want to comment on.

84634E1A607A:
```csharp
        if (count == 0)
            throw new InvalidOperationException("Sequence contains no elements");
```

System.Linq throws an InvalidOperationException when sequence is empty, so I copied that.

hhyyrylainen:
I suppose that is fine as it probably doesn't make sense to get an average for a sequence with no elements.

hhyyrylainen (approved):

hhyyrylainen:
Opened a follow up to adjust the GUI a bit in relation to my latest controller changes: https://github.com/Revolutionary-Games/Thrive/pull/3986

PR #3966, opened 2022-12-31
Print world seed when map generation fails

84634E1A607A:

Just experienced a rare crash (as no such report since patch generation system was written) where no tidepool was generated, ending in an exception (and I forgot to look at its seed). It would be good if seed is included in the exception message so that we can reproduce the bug easily.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (approved):
This code change seems perfectly sensible to me.

PR #3971, opened 2023-01-02
Fixed equality comparisons for process speed data when processes change

### LINKED ISSUE #3898: Process panel stops updating properly with specific changes to the player

hhyyrylainen:

this should fix the process panel stopping updating

**Related Issues**

fixes #3898
closes #3955

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/ChemicalEquation.cs
+++ b/src/gui_common/ChemicalEquation.cs
@@ -52,7 +52,7 @@ public class ChemicalEquation : VBoxContainer
             if (equationFromProcess == null && value == null)
                 return;
 
-            if (equationFromProcess?.Equals(value) == true)
+            if (value != null && equationFromProcess?.Equals(value) == true)
                 return;
 
             equationFromProcess = value;
--- a/src/microbe_stage/ProcessStatistics.cs
+++ b/src/microbe_stage/ProcessStatistics.cs
@@ -8,7 +8,17 @@ using Nito.Collections;
 /// </summary>
 public class ProcessStatistics
 {
-    public Dictionary<TweakedProcess, SingleProcessStatistics> Processes { get; } = new();
+    /// <summary>
+    ///   The processes and their associated speed statistics
+    /// </summary>
+    /// <remarks>
+    ///   <para>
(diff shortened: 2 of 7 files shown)


src/gui_common/ChemicalEquation.cs
```diff
@@ -52,7 +52,7 @@ public class ChemicalEquation : VBoxContainer
             if (equationFromProcess == null && value == null)
                 return;
 
-            if (equationFromProcess?.Equals(value) == true)
+            if (value != null && equationFromProcess?.Equals(value) == true)
```
84634E1A607A, line 55: Why null check here?

hhyyrylainen: The Equals method here takes in a non-nullable parameter so it causes a warning because `value` is nullable here. I think this is semantically the right way to express what the old code wanted to do.

src/microbe_stage/ProcessStatistics.cs
```diff
@@ -26,15 +36,15 @@ public void RemoveUnused()
...
+    public SingleProcessStatistics GetAndMarkUsed(BioProcess forProcess)
     {
         if (Processes.TryGetValue(forProcess, out var entry))
         {
             entry.Used = true;
             return entry;
         }
 
-        entry = new SingleProcessStatistics(forProcess.Process);
+        entry = new SingleProcessStatistics(forProcess);
```
Oliveriver, line 47: Ah, good spot simplifying this.

Oliveriver (approved):
Sensible fix. I'll go ahead and close my PR for this.

src/microbe_stage/ProcessStatistics.cs
hhyyrylainen: Well this was not really about simplifying. I initially designed this to take in a TweakedProcess in case there was a situation where multiple TweakedProcesses might exist for the same base process and their statistics should be able to exist separately. I've now decided that that is a very confusing and bad thing to allow, so I adjusted the object used as key to store the speed information.

PR #4025, opened 2023-01-17
Split fossilise button from species details panel and extended usage of the panel

84634E1A607A:
**My reason to do this:**

Species details panel is a whole part, while fossilise button is independent. More controls may be put into that panel but the fossilise button will stay the same.

This PR creates a separate, plain details panel and removes some code and tscn duplication.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I think it should be kept there... Scenes that have extended functionality that is only rarely enabled are better in my opinion than code duplication. I'm open to other opinions on this topic.

84634E1A607A:
The main problem is the fossilisation dialog itself contains a species preview part (code duplication), if the details panel is kept as-is, this part of duplication cannot get rid of (and will probably go longer and longer once we add more (such as Multicellular species preveiw) feature to species preview panel.

hhyyrylainen:
Shouldn't the species preview component be what is extracted as a scene then to reduce duplication?

84634E1A607A:
This PR does this, but I wonder if we still wants a separate scene that contains the preview panel and and the fossilisation button along with the dialog (as mentioned in my initial comment)

hhyyrylainen:
As I said, I think we do want a combined scene that just handles everything. The fossilization dialog could additionally be a separate scene if that was the place where code duplication was an issue.

84634E1A607A:
So this is the combined scene (although its name are a little bit loooong.)

hhyyrylainen:
I think the name does not need to say it has the fossilization in it. I think renaming `SpeciesDetailsPanelWithFossilisation` to `SpeciesDetailsPanel` would be completely fine and understandable. There are many scenes that contain a lot of related functionality that isn't mentioned in the name.

84634E1A607A:
But `SpeciesDetailsPanel` is the one without fossilisation. Let me explain, `SpeciesDetailsPanelWithFossilisation` has subresource `FossilisationDialog`, which has `SpeciesDetailsPanel` in it so the two cannot be merged (which will create an infinite resource loop).

hhyyrylainen:
Okay, then I'd say the species details part should be extracted as an separate scene, then that can be placed in two places in the SpeciesDetailsPanel scene. I don't have a really good name suggestion, maybe something like "SpeciesDetailsDisplayer" or "SpeciesDetailsInternal" would be good enough.

hhyyrylainen:
I'll leave reviewing this up to @Oliveriver and won't think about this PR anymore unless someone mentions me here or points out that something should be done for this.

84634E1A607A:
No one to review this?

hhyyrylainen:
I don't see any new commits after: https://github.com/Revolutionary-Games/Thrive/pull/4025#issuecomment-1385030034 which is basically all I had to say about that. And I'll leave it up to @Oliveriver to say whether this change is a good idea or not (or well if Oliver is not available anytime soon I might allow this change with that one modification I wanted anyway).

Kasterisk:
What's this file [`.attach_pid12133`](https://github.com/Revolutionary-Games/Thrive/pull/4025/files#diff-e47f0de369ec033aaaa2ab822d437f0c44b1aceb5b7e0574619e143220f849e2)?

84634E1A607A:
I guess a temp file Rider created when attaching to a process.

Kasterisk:
I see, you can remove that in future commits, we probably don't need that.

hhyyrylainen:
Adding it to also .gitignore would be nice.

84634E1A607A:
I'm not sure where to put it. gitignore is having many strange things.

hhyyrylainen:
Maybe after `.vscode` ? It seems kind of related as an ide generated temporary file.

84634E1A607A:
Still no reviewing?

hhyyrylainen:
I can do a review and accept this if my earlier comment is implemented: https://github.com/Revolutionary-Games/Thrive/pull/4025#issuecomment-1385030034 (sorry if I missed that being implemented / forgot).

hhyyrylainen:
Okay, I guess the name `SpeciesDetailsPanelWithFossilisation` can stay as I don't have a better solution to the naming problem. I'll review the code changes now, I saw at least one `#pragma warning restore` being slightly in the wrong place.

src/gui_common/SpeciesDetailsPanel.cs
```diff
@@ -12,21 +11,12 @@ public class SpeciesDetailsPanel : VBoxContainer
...
     private CustomRichTextLabel speciesDetailsLabel = null!;
-    private SpeciesPreview? speciesPreview;
+    private SpeciesPreview speciesPreview = null!;
     private CellHexesPreview hexesPreview = null!;
-    private Button fossilisationButton = null!;
-    private FossilisationDialog fossilisationDialog = null!;
-#pragma warning restore CA2213
-
     private Species? previewSpecies;
+#pragma warning restore CA2213
```
hhyyrylainen, line 19: This pragma needs to be before `previewSpecies` variable as the pragma is to be only used around Godot Node variables, so it being around this species variable is wrong.

src/gui_common/SpeciesDetailsPanel.cs
```diff
@@ -38,7 +28,7 @@ public class SpeciesDetailsPanel : VBoxContainer
 
             previewSpecies = value;
 
-            if (previewSpecies != null)
+            if (previewSpecies != null && speciesPreview != null!)
```
hhyyrylainen, line 31: You should keep `speciesPreview` as nullable because that is better than having to do this dirty null override here.

src/gui_common/SpeciesDetailsPanelWithFossilisation.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
+using Godot;
+
+public class SpeciesDetailsPanelWithFossilisation : VBoxContainer
```
hhyyrylainen, line 4: Could you add a summary comment on this class to mention that this is an extended version of the species details panel that adds the fossilization popup and button?

src/gui_common/SpeciesDetailsPanelWithFossilisation.cs
```diff
@@ -0,0 +1,75 @@
...
+        fossilisationDialog = GetNode<FossilisationDialog>(FossilisationDialogPath);
+
+        UpdateFossilisationButtonState();
+    }
+
+    protected override void Dispose(bool disposing)
+    {
+        if (disposing)
+        {
+            FossilisationButtonPath.Dispose();
```
hhyyrylainen, line 53: This is incorrect, there must be a null check here to ensure that when Godot partially loads this class, this doesn't cause an exception. If you didn't already please read my recent styleguide changes, those *should* explain this.

src/thriveopedia/fossilisation/FossilisationDialog.tscn
```diff
@@ -51,7 +47,10 @@ margin_right = 152.0
 [node name="Label" parent="VBoxContainer" index="0"]
 margin_right = 137.0
 
-[node name="HBoxContainer" parent="VBoxContainer" index="1"]
+[node name="FocusGrabber" parent="VBoxContainer" index="1"]
```
hhyyrylainen, line 50: Why is there an unconfigured focus grabber node here?

hhyyrylainen (changes requested):
I'm willing to accept this refactor now, but I noticed a few small code style / Godot usage issues.

Oliveriver:
Sorry for not getting to this. I don’t really keep track of GitHub tags very well. I can review it tonight if it’s still needed.

hhyyrylainen:
I basically just did a full review so it's not really necessary. Though once my comments are addressed it would be nice to have some help playtesting the areas of the game this changed to ensure no bugs slipped through.

src/thriveopedia/fossilisation/FossilisationDialog.tscn
84634E1A607A: For some reason I don't know (Maybe @Oliveriver knows?) `FossilisationDialog` is extended from `CustomConfirmationDialog`. Any changes in the base scene will be added automatically here.

hhyyrylainen: I see you configured the focus grabber in a commit, I'll test this now.

src/thriveopedia/fossilisation/FossilisationDialog.tscn
```diff
@@ -51,7 +47,11 @@ margin_right = 152.0
...
+margin_right = 137.0
+NodeToGiveFocusTo = NodePath("../../VBoxContainer2/HBoxContainer/CancelButton")
```
hhyyrylainen, line 52: This focus grabber doesn't work here because the normal confirmation buttons are all hidden.
Is this quick to fix or should I open a separate issue to no longer have the fossilisation dialog inherit the confirmation dialog (it should be a base custom dialog)?

Oliveriver:
> I basically just did a full review so it's not really necessary. Though once my comments are addressed it would be nice to have some help playtesting the areas of the game this changed to ensure no bugs slipped through.

Tested and it works fine in all the locations it appears. The text description of fossilisation in the dialog has moved to below the scroll container though and I'm not sure which is better.

src/thriveopedia/fossilisation/FossilisationDialog.tscn
Oliveriver: > For some reason I don't know (Maybe @Oliveriver knows?) `FossilisationDialog` is extended from `CustomConfirmationDialog`. Any changes in the base scene will be added automatically here.

Whether it's significant or not, I can't remember why this changed.

hhyyrylainen: Well I definitely added the focus grabber to the base confirmation dialog so that the cancel button is selected by default.

I commented elsewhere already but it seems the fossilisation dialog entirely just hides the usual confirmation dialog content, so it inheriting that, doesn't make much sense so I'm open to leaving things be as they are in this PR but opening a follow up issue to fix that.

hhyyrylainen (approved):

hhyyrylainen:
I opened a follow up issue for the base of the fossilisation dialog fix: https://github.com/Revolutionary-Games/Thrive/issues/4157

PR #4084, opened 2023-02-01
Slight bugfix against potential action history cache bug and a few minor cleanups

hhyyrylainen:

I think I found a potential bug where a combined action having a resetting action as the non-last action, can cause a bug with incorrect action cache being created. I also included a few totally random changes in here that are from my failed attempt at simplifying the remaining MP calculation.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/editor/EditorActionHistory.cs
```diff
@@ -135,7 +135,7 @@ public override void AddAction(TAction action)
         // Handle adding directly the action to our history cache, this saves us from having to rebuild the cache
```
84634E1A607A, line 135: Does this comment still make sense?

hhyyrylainen: It does still in the other (`else`) branch add it to the cache. I wanted to remove the potential bug without having to complicate things a lot (I could grab any actions in the to be queued thing that are after the history reset, but I think history resetting actions are rare so it doesn't require that much optimization).

hhyyrylainen: I'll add a comment on the line I changed to clarify that.

84634E1A607A (approved):

PR #4099, opened 2023-02-03
Visualize at most 3 digits for resources

### LINKED ISSUE #3902: When max ATP is really high it gets cutoff in the GUI
This is the second time we've been reported that the max ATP doesn't match what the player has stored:

screenshot

We need to either line wrap and allow the box to grow in size or come up with some thing to do like a "k" suffix to get smaller numbers.

LorenzoM1997:

There is a known issue that sometimes there are too many digits in the ATP visualization and it can't be displayed properly. Here the issue is mitigated by removing the decimal place for numbers between 100 and 999 and using the "K" with one decimal number after that.

A related issue is the display of the resources, which suffers a similar problem but only in the smaller, more compact view.

Closes #3902

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/StringUtils.cs
+++ b/src/general/StringUtils.cs
@@ -81,9 +81,11 @@ public static class StringUtils
     ///   Splits string into different chunks by whitespace.
     /// </summary>
     /// <remarks>
-    ///   Only handles a single whitespace ( ) and not tabs, multiple spaces, etc. For quoted
-    ///   substrings handling, this only considers double quotes (") and not apostrophes (').
-    ///   If there is no closing quote after an opening quote, the rest of string is considered within quotes.
+    ///  <para>
+    ///     Only handles a single whitespace ( ) and not tabs, multiple spaces, etc. For quoted
+    ///     substrings handling, this only considers double quotes (") and not apostrophes (').
+    ///     If there is no closing quote after an opening quote, the rest of string is considered within quotes.
+    ///   </para>
     /// </remarks>
     /// <param name="input">String to split.</param>
     /// <param name="ignoreWithinQuotes">Ignore whitespace within quotation marks.</param>
@@ -192,18 +194,21 @@ public static class StringUtils
     ///   Convert a value in a string with up to 3 digits.
     /// </summary>
     /// <remarks>
-    ///   If greater than 100, it should use F0, otherwise it should use F1.
-    ///   Above 1000 it should use the number formatting which uses K for
-    ///   thousands, M for millions etc.
+    ///   <para>
+    ///     If greater than 100, it should use F0, otherwise it should use F1.
+    ///     Above 1000 it should use the number formatting which uses K for
+    ///     thousands, M for millions etc. It can handle negative numbers
+    ///   </para>
     /// </remarks>
     /// <param name="value">The value to format</param>
     /// <returns>The formatted string</returns>
     public static string ThreeDigitFormat(double value)
     {
-        if (value >= 1000)
+        double absoluteValue = Math.Abs(value);
+        if (absoluteValue >= 1000)
             return FormatNumber(value);
 
-        if (value >= 100)
+        if (absoluteValue >= 100)
             return value.ToString("F0", CultureInfo.CurrentCulture);
 
         return value.ToString("F1", CultureInfo.CurrentCulture);
@@ -213,11 +218,11 @@ public static class StringUtils
     ///   Formats two numbers separated by a slash. The numbers will have
     ///   up to 3 digits each.
     /// </summary>
-    /// <param name="v1">The first number (numerator)</param>
-    /// <param name="v2">The second number (denominator)</param>
+    /// <param name="numerator">The first number (numerator)</param>
+    /// <param name="denominator">The second number (denominator)</param>
     /// <returns>The formatted string</returns>
-    public static string SlashSeparatedNumbersFormat(double v1, double v2)
+    public static string SlashSeparatedNumbersFormat(double numerator, double denominator)
     {
-        return ThreeDigitFormat(v1) + " / " + ThreeDigitFormat(v2);
+        return ThreeDigitFormat(numerator) + " / " + ThreeDigitFormat(denominator);
     }
 }
--- a/src/general/StringUtils.cs
+++ b/src/general/StringUtils.cs
@@ -189,30 +189,29 @@ public static class StringUtils
     }
 
     /// <summary>
-    // utility function to get the correct text for the ATP label, taking
-    // a number as an input. If greater than 100, it should use F0, otherwise
-    // it should use F1. Above 1000 it should use F0 and add a K.
+    ///   Convert a value in a string with up to 3 digits.
     /// </summary>
+    /// <remarks>
+    ///   If greater than 100, it should use F0, otherwise it should use F1.
+    ///   Above 1000 it should use the number formatting which uses K for
+    ///   thousands, M for millions etc.
+    /// </remarks>
     /// <param name="value">The value to format</param>
     /// <returns>The formatted string</returns>
     public static string ThreeDigitFormat(double value)
     {
         if (value >= 1000)
-            return (value / 1000).ToString("F1", CultureInfo.CurrentCulture) + "K";
-        else if (value >= 100)
-            return value.ToString("F0", CultureInfo.CurrentCulture);
-        else
-            return value.ToString("F1", CultureInfo.CurrentCulture);
-    }
+            return FormatNumber(value);
 
-    /// <summary> Implements <see cref="ThreeDigitFormat(double)"/> for floats </summary>
-    public static string ThreeDigitFormat(float value)
-    {
-        return ThreeDigitFormat((double)value);
+        if (value >= 100)
+            return value.ToString("F0", CultureInfo.CurrentCulture);
+
+        return value.ToString("F1", CultureInfo.CurrentCulture);
     }
 
     /// <summary>
-    /// Formats two numbers separated by a slash
+    ///   Formats two numbers separated by a slash. The numbers will have
+    ///   up to 3 digits each.
     /// </summary>
     /// <param name="v1">The first number (numerator)</param>
     /// <param name="v2">The second number (denominator)</param>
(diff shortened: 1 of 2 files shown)


LorenzoM1997:
Below the changes in action: the number between 100 and 999 have 3 digits, above 1000 have n.nK notation
screengrab-fix-thrive

src/general/StringUtils.cs
```diff
@@ -187,4 +187,38 @@ public static string Unescape(this string input)
...
+    // utility function to get the correct text for the ATP label, taking
+    // a number as an input. If greater than 100, it should use F0, otherwise
+    // it should use F1. Above 1000 it should use F0 and add a K.
+    /// </summary>
+    /// <param name="value">The value to format</param>
+    /// <returns>The formatted string</returns>
+    public static string ThreeDigitFormat(double value)
+    {
+        if (value >= 1000)
+            return (value / 1000).ToString("F1", CultureInfo.CurrentCulture) + "K";
```
84634E1A607A, line 201: We have a translaton for "K", see https://github.com/Revolutionary-Games/Thrive/blob/eb7b7b1b8039ceed9f7d7d9787c40320316e2fd2/src/general/StringUtils.cs#L37

84634E1A607A (changes requested):
CI says that:

```
Problems found in file src/general/StageHUDBase.cs:
Line 1052 is too long. 129 > 120
Line 1056 is too long. 129 > 120
Line 1060 is too long. 135 > 120
Line 1064 is too long. 153 > 120
Line 1072 is too long. 129 > 120
Line 1076 is too long. 132 > 120
```

And

```
Code inspection detected issues:
src\general\StageHUDBase.cs:3 Using directive is not required by the code and can be safely removed type: RedundantUsingDirective
Offending code (offset 48-75) for previous message: 'using System.Globalization;'
src\general\StringUtils.cs:191 XML comment is not placed on a valid language element type: InvalidXmlDocComment
Offending code (offset 6723-6736) for previous message: '/// <summary>'
src\general\StringUtils.cs:195 No open tag found type: InvalidXmlDocComment
Offending code (offset 6963-6973) for previous message: '</summary>'
```

Those format issues should be resolved.

I think the others are fine.

src/general/StringUtils.cs
hhyyrylainen: Also, I'm pretty sure we had existing logic for doing this somewhere...

Just looked and there is `StringUtils.FormatNumber`.

LorenzoM1997: Thanks, I didn't notice it

src/general/StringUtils.cs
```diff
@@ -187,4 +187,38 @@ public static string Unescape(this string input)
...
+        else if (value >= 100)
```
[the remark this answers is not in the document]
LorenzoM1997: That is pretty interesting. I don't get this type of warning on my setup, thanks for pointing it out

src/general/StringUtils.cs
```diff
@@ -187,4 +187,38 @@ public static string Unescape(this string input)
...
+            return value.ToString("F0", CultureInfo.CurrentCulture);
+        else
+            return value.ToString("F1", CultureInfo.CurrentCulture);
+    }
+
+    /// <summary> Implements <see cref="ThreeDigitFormat(double)"/> for floats </summary>
+    public static string ThreeDigitFormat(float value)
```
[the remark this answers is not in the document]
LorenzoM1997: yep, removed now

src/general/StringUtils.cs
```diff
@@ -189,30 +189,29 @@ public static string Unescape(this string input)
...
-    // utility function to get the correct text for the ATP label, taking
-    // a number as an input. If greater than 100, it should use F0, otherwise
-    // it should use F1. Above 1000 it should use F0 and add a K.
+    ///   Convert a value in a string with up to 3 digits.
     /// </summary>
+    /// <remarks>
+    ///   If greater than 100, it should use F0, otherwise it should use F1.
+    ///   Above 1000 it should use the number formatting which uses K for
+    ///   thousands, M for millions etc.
+    /// </remarks>
```
84634E1A607A, line 198: Check the style of `<remarks>`. And I think we should add a remainder that this doesn't handle negative value.

84634E1A607A: https://github.com/Revolutionary-Games/Thrive/blob/cf8b7746ef6542c021014dd07690be94e0daf2d7/src/general/StringUtils.cs#L57

LorenzoM1997: Ah I looked at one above and it didn't have <para>, I'll fix that as well. And will add that note.

src/general/StringUtils.cs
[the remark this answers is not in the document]
hhyyrylainen: Setup a precommit hook or manually run the formatting. These are both mentioned in the setup instructions: https://github.com/Revolutionary-Games/Thrive/blob/master/doc/setup_instructions.md#running-the-format-checks

src/general/StringUtils.cs
```diff
@@ -187,4 +187,37 @@ public static string Unescape(this string input)
 
         return result.ToString();
     }
+
+    /// <summary>
+    ///   Convert a value in a string with up to 3 digits.
+    /// </summary>
+    /// <remarks>
+    ///   If greater than 100, it should use F0, otherwise it should use F1.
```
hhyyrylainen, line 195: Remarks must contain all the text within `para` elements, as per the styleguide, which you should check if you have not already.

src/general/StringUtils.cs
hhyyrylainen: Wouldn't it be relatively easy to make this work with negative numbers as well or?

src/general/StringUtils.cs
```diff
@@ -187,4 +187,37 @@ public static string Unescape(this string input)
...
+            return value.ToString("F0", CultureInfo.CurrentCulture);
+
+        return value.ToString("F1", CultureInfo.CurrentCulture);
+    }
+
+    /// <summary>
+    ///   Formats two numbers separated by a slash. The numbers will have
+    ///   up to 3 digits each.
+    /// </summary>
+    /// <param name="v1">The first number (numerator)</param>
```
hhyyrylainen, line 216: `v` is not an approved single letter variable, please rename the v1 and v2 variables to something else, for example "value1" and "value2".

hhyyrylainen (changes requested):

src/general/StringUtils.cs
```diff
@@ -187,4 +187,37 @@ public static string Unescape(this string input)
...
+    /// <param name="v2">The second number (denominator)</param>
+    /// <returns>The formatted string</returns>
+    public static string SlashSeparatedNumbersFormat(double v1, double v2)
+    {
+        return ThreeDigitFormat(v1) + " / " + ThreeDigitFormat(v2);
```
hhyyrylainen, line 221: Oh and I think it would be a pretty cheap performance win to use the `$"{variable} / {...}"` string formatting here rather than doing 2 string concatenations.

LorenzoM1997: Does it really? I don't use much C# but a couple of places I look suggest otherwise.
[Here one example](https://www.meziantou.net/performance-string-concatenation-vs-string-format-vs-interpolated-string.htm) that I found on the internet. If you want I'll change it anyway, I don't think it should make much of a difference tbh.

src/general/StringUtils.cs
LorenzoM1997: I went with 'numerator' and 'denominator' here since this is a fraction and seems slightly more descriptive.

LorenzoM1997:
Github is not letting me re-requesting review from both of you @hhyyrylainen @84634E1A607A. I thought I could, sorry if I sent multiple requests by mistake.

src/general/StringUtils.cs
hhyyrylainen: It seems the advice I've seen that string concatenation will destroy performance doesn't apply in this case. I guess the advice only applies if you iteratively build the string i.e. don't do it in a single go.

I did do a quick test in Thrive and I got:
```
String interpolation: 00:05:03.3813094
String concat: 00:04:35.7005441
```

So a simple 3 value concat on a single line is faster, so I guess maybe it is better to just keep the string concat here?

For interest, here's how the MSIL looks (it seems like the string interpolation causes boxing on the 2 primitive values, whereas the straight concat just calls string conversions on the primitive values and then just makes one call to concat). Interpolated:
```
        IL_0022: ldstr        "{0} / {1}"
        IL_0027: ldloc.3      // baseValue
        IL_0028: box          [mscorlib]System.Single
        IL_002d: ldloc.3      // baseValue
        IL_002e: ldc.r4       5
        IL_0033: add
        IL_0034: box          [mscorlib]System.Single
        IL_0039: call         string [mscorlib]System.String::Format(string, object, object)
        IL_003e: stloc.s      'value'
```
the string concat:
```
       IL_008b: ldloca.s     baseValue_V_7
        IL_008d: call         instance string [mscorlib]System.Single::ToString()
        IL_0092: ldstr        " / "
        IL_0097: ldloc.s      baseValue_V_7
        IL_0099: ldc.r4       5
        IL_009e: add
        IL_009f: stloc.s      V_9
        IL_00a1: ldloca.s     V_9
        IL_00a3: call         instance string [mscorlib]System.Single::ToString()
        IL_00a8: call         string [mscorlib]System.String::Concat(string, string, string)
        IL_00ad: stloc.s      value_V_8
```

src/general/StringUtils.cs
```diff
@@ -187,4 +189,40 @@ public static string Unescape(this string input)
...
+    ///     Above 1000 it should use the number formatting which uses K for
+    ///     thousands, M for millions etc. It can handle negative numbers
+    ///   </para>
+    /// </remarks>
+    /// <param name="value">The value to format</param>
+    /// <returns>The formatted string</returns>
+    public static string ThreeDigitFormat(double value)
+    {
+        double absoluteValue = Math.Abs(value);
+        if (absoluteValue >= 1000)
```
hhyyrylainen, line 208: I think this should use the same approach as in `FormatNumber` i.e. check if the `value is >= 1000 or <= -1000`

hhyyrylainen (changes requested):
I have just one more code related comment on this.

84634E1A607A (approved):

hhyyrylainen (approved):
Code change looks good to me now.

PR #3880, opened 2022-11-29
Patch notes update

k-agau:
**Brief Description of What This PR Does**
This PR allows the user to display the patch notes both from the start menu and the in-game options.

Resolves #3868 

**Related Issues**

**Progress Checklist**

Note: before starting this checklist the PR should be marked as non-draft.

- [ ] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [ ] Initial code review passed (this and further items should not be checked by the PR author)
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author. Merging must follow our
[styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md#git).

Issue #4028, opened 2023-01-18
Create a system to have versioned settings files


hhyyrylainen:
so that when we introduce backwards incompatible settings we can bump the settings file version.

When no settings exist for the current version, previous version settings should be looked for and if found copied to the new file.

I noticed this as a bit of a problem with controller rebinding in that when adding controller keybindings the options file is no longer compatible with older Thrive versions.

PR #4101, opened 2023-02-03
Added showing patch notes to the main menu

### LINKED ISSUE #3868: The main menu should show latest patch notes after version update
the next time the player opens the game and it has been updated, there should be patch notes shown in the main menu, probably to the left of the main menu buttons? The patch notes should be accessible in full also from the options menu in case the player wants to check them later if they missed them. We probably need to store the patch notes in a JSON form in the repository. They *probably* shouldn't be translated as they are very technical and the chance of having timely community translations (this makes timing releases much more difficult) is not very good.

hhyyrylainen:

Added displaying patch notes to the main menu on update and added a button in options to view patch notes anytime. There's an option to disable the feature. Also contains small tweaks (and one fix) for the thrive news feed.

**Related Issues**

closes #3868
closes #3880

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/engine/LastPlayedVersion.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+using Godot;
+
+/// <summary>
+///   Loads and allows updating the last played Thrive version
+/// </summary>
+public static class LastPlayedVersion
```
84634E1A607A, line 7: Why don't we add a settings value for last played version?

hhyyrylainen: What version was last played is unrelated to any game settings. And with https://github.com/Revolutionary-Games/Thrive/issues/4028 a value stored in settings wouldn't necessarily even correlate with the last played version. Kind of the same reason really as why the game startup failure detection uses a separate file as well.

Gameboy555r:
i took a look into this and, i dunno if its just for me but i works fine.

src/gui_common/PatchNotesDisplayer.cs
```diff
@@ -0,0 +1,170 @@
...
+        {
+            // If no known last played version, this is the firs time Thrive is played and there's no need to show
+            // patch notes
+            return false;
+        }
+
+        var currentVersion = Constants.Version;
+
+        // If we aren't playing a newer version then also skip showing
+        if (VersionUtils.Compare(currentVersion, lastPlayed) <= 0)
```
84634E1A607A, line 71: If I add some invalid data into `last_played_version.txt` (To simulate that it is corrupt), Thrive will print a *LOT* of exceptions.

84634E1A607A:
Many patch note links lost its left parenthesis
image

src/general/MarkdownToBbCodeConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+///   Converts markdown to bbcode for displaying
+/// </summary>
+public static class MarkdownToBbCodeConverter
+{
+    private static readonly Regex MarkdownLink = new(@"\[([^\]]+)\]\((http[^\)]+)\)", RegexOptions.Multiline);
+
+    private static readonly Regex BasicHtmlLink =
+        new(@"<a\s+[^>]*href=\""(http\S+)\""[^>]*>\s*(.+)\s*</a\s*>", RegexOptions.Compiled | RegexOptions.Multiline);
+
+    private static readonly Regex UrlWithoutMarkup = new(@"[^=](https?:\/\/[^\s)\]}]+)");
```
84634E1A607A, line 13: This RegEx is faulty, it consumes the left parenthesis (`[^=]`).

One solution is to change this to `@"(^=)(https?:\/\/[^\s)\]}]+)"` and the conversion to `$"$1{Constants.CLICKABLE_TEXT_BBCODE}[url=$2]$2[/url]{Constants.CLICKABLE_TEXT_BBCODE_END}"`

src/general/MarkdownToBbCodeConverter.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Text.RegularExpressions;
+
+/// <summary>
+///   Converts markdown to bbcode for displaying
+/// </summary>
+public static class MarkdownToBbCodeConverter
+{
+    private static readonly Regex MarkdownLink = new(@"\[([^\]]+)\]\((http[^\)]+)\)", RegexOptions.Multiline);
```
84634E1A607A, line 8: The later line has a `Compiled` option, why not this?

src/engine/LastPlayedVersion.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    private static Lazy<string?> lastPlayed = new(ReadLastPlayedVersion);
+
+    private static bool handledCurrentVersionPlayed;
+
+    public static string? LastPlayed => lastPlayed.Value;
+
+    public static void MarkCurrentVersionAsPlayed()
```
84634E1A607A, line 15: Microbe freebuild editor doesn't call this

84634E1A607A (changes requested):
So I found a few bugs. Other than that (and the UI) I have nothing to comment on. I'm not a steam user so I'm not sure what it looks like when using steam and a packaged built.

Also I wonder if it is preferable to move all those news conversions to our web server, and make the generation script to parse all those notes to its final bbcode state. It's no point to download a number of web pages and spend time parsing them.

src/gui_common/PatchNotesDisplayer.cs
hhyyrylainen: I'll put in data validation to LastPlayedVersion loading code.

src/general/MarkdownToBbCodeConverter.cs
hhyyrylainen: Yeah, I realized this based on your previous comment and basically ended up doing this exact change without seeing your comment first.

src/general/MarkdownToBbCodeConverter.cs
hhyyrylainen: I decided to just put it on the regex that looked the most intensive as these are static properties, these objects will be created when the game starts so I thought it best not to use that extensively. Of course I didn't feel like profiling this or anything...

src/engine/LastPlayedVersion.cs
hhyyrylainen: I'll move the call to mark the version as played to `MainMenu.OnEnteringGame` to not have to put it all over (hopefully entering freebuild from auto-evo exploring tool also calls that).

84634E1A607A: Nope. Auto-evo exploring tool is a separate scene and doesn't hold any info about the main menu.

(Oh, there's also thrivopedia)

hhyyrylainen:
> Many patch note links lost its left parenthesis

Ah, I'm accidentally losing one character in the regex replace call, I'll fix that to include the character that was accidentally removed.

> Also I wonder if it is preferable to move all those news conversions to our web server, and make the generation script to parse all those notes to its final bbcode state. It's no point to download a number of web pages and spend time parsing them.

I guess that is possible. Most people use web browsers that try to understand html to display it. The launcher does the exact same thing, it parses HTML. If this was changed in a sensible way, then there'd need to be multiple variants of the data for Thrive and launcher separately. Also going that far, the markdown to bbcode conversion could also be done ahead of time.

That is possible but adds the problem that the link clickable colour will be set by the *server* so if someone wants to change the link colour in Thrive they need to also modify the server software and re-process all the already generated patch notes data.

src/engine/LastPlayedVersion.cs
hhyyrylainen: I opened an issue as it's pretty important to have all methods of starting a game trigger the same operations for consistency: https://github.com/Revolutionary-Games/Thrive/issues/4112

84634E1A607A:
Link-clickable color can be set as a placeholder (For example `#LINK_COLOR`) and use string replacement which is faster. And I'm actually saying that a script can be added at the server side to calculate and **cache** feeds data, which will reduce overall calc time, reduce calc duplication and reduce network packet length.

84634E1A607A (approved):

hhyyrylainen:
Opened follow up issues: https://github.com/Revolutionary-Games/Thrive/issues/4113
and https://github.com/Revolutionary-Games/Thrive/issues/4114

For good measure I also opened an issue for the launcher: https://github.com/Revolutionary-Games/Thrive-Launcher/issues/259

PR #4141, opened 2023-02-20
Immediately snap to current patch the first time patch map is drawn

Kasterisk:

Removes the smooth panning to the current patch when you're initially opening the patch map everytime you're in a new editor cycle, as it can get pretty annoying.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I agree that would be good that the map doesn't always start off by panning somewhere.

hhyyrylainen (approved):
Much nicer now with the patch map not always panning in the editor when going through the tabs.

PR #4144, opened 2023-02-21
Allow specialized engulfing in microbe colony

### LINKED ISSUE #3549: If primary cell in a colony can't engulf, but other members can they should be able to enter engulf mode
Right now the primary cell in a colony (at least when controlled by the player) must have a membrane allowing engulfing, otherwise no member in the colony can enter engulf mode. We should update things so that the player can enter engulf mode and is shown the hot bar icon if any cell in their colony can engulf.

Kasterisk:

Walled cells can now be in engulf state but can't act on it, this is to allow specialized engulfing in a colony.

I had to make two ways of checking whether a cell is capable to engulf, one is through `ICellProperties` and the other through `Microbe` itself, decoupled from species. The reason for this is because `ICellProperties` sometimes may not reflect the actual properties visible in a cell.

I hadn't yet tested in multicellular but this should work fine for colony.

**Related Issues**

Fixes #3549 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Microbe.Contact.cs
+++ b/src/microbe_stage/Microbe.Contact.cs
@@ -190,19 +190,12 @@ public partial class Microbe
     }
 
     /// <summary>
-    ///   Returns true if this microbe or the colony this microbe is part of has the capability to engulf.
+    ///   Just like <see cref="ICellProperties.CanEngulf"/> but decoupled from Species and is based on the local
+    ///   condition of the microbe instead.
     /// </summary>
+    /// <returns>True if this cell fills all the requirements it needs to be able to enter engulf mode.</returns>
     [JsonIgnore]
-    public bool HasEngulfCapability
-    {
-        get
-        {
-            if (Colony != null)
-                return Colony.HasEngulfCapability;
-
-            return CheckEngulfCapability();
-        }
-    }
+    public bool CanEngulf => !Membrane.Type.CellWall;
 
     /// <summary>
     ///   Returns true when this microbe can enable binding mode. Multicellular species can't attach random cells
@@ -395,7 +388,7 @@ public partial class Microbe
     /// <summary>
     ///   Returns true when this microbe can engulf the target
     /// </summary>
-    public bool CanEngulf(IEngulfable target)
+    public bool CanEngulfObject(IEngulfable target)
     {
         if (target.PhagocytosisStep != PhagocytosisPhase.None)
             return false;
@@ -449,7 +442,7 @@ public partial class Microbe
             return false;
 
         // Membranes with Cell Wall cannot engulf
-        if (!CheckEngulfCapability())
+        if (!CanEngulf)
             return false;
 
         // Godmode grants player complete engulfment invulnerability
@@ -461,13 +454,14 @@ public partial class Microbe
     }
 
     /// <summary>
-    ///   Just like <see cref="ICellProperties.CanEngulf"/> but decoupled from Species and is based on the local
-    ///   condition of the microbe instead.
+    ///   Returns true if this microbe OR the colony this microbe is part of has the capability to engulf.
     /// </summary>
-    /// <returns>True if this cell has all the requirements it needs for engulfing.</returns>
-    public bool CheckEngulfCapability()
+    public bool CanEngulfInColony()
     {
-        return !Membrane.Type.CellWall;
+        if (Colony != null)
+            return Colony.CanEngulf;
+
+        return CanEngulf;
     }
 
     public void OnAttemptedToBeEngulfed()
@@ -1121,7 +1115,7 @@ public partial class Microbe
     /// </summary>
     private void HandleEngulfing(float delta)
     {
-        if (!CheckEngulfCapability())
+        if (!CanEngulf)
             return;
 
         if (State == MicrobeState.Engulf)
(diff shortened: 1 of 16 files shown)


hhyyrylainen:
Why did you not take the approach that when a cell tries to enter engulf mode, even if it cannot do it, it would broadcast the message to the colony? That way any cells that *can* engulf in the colony would switch to engulf mode. I thought that was the pretty obvious solution to #3549

Kasterisk:
That's what I did.

https://github.com/Revolutionary-Games/Thrive/blob/ee6e5c7b5ee1bab6aee56607c955ff07f78ca53f/src/microbe_stage/Microbe.Contact.cs#L1122-L1126

hhyyrylainen:
I saw quite a lot of changes when skimming the diff. I suppose I'll look at it properly.

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -491,6 +460,16 @@ public bool CanEngulf(IEngulfable target)
         return EngulfSize > target.EngulfSize * Constants.ENGULF_SIZE_RATIO_REQ;
     }
 
+    /// <summary>
+    ///   Just like <see cref="ICellProperties.CanEngulf"/> but decoupled from Species and is based on the local
+    ///   condition of the microbe instead.
+    /// </summary>
+    /// <returns>True if this cell has all the requirements it needs for engulfing.</returns>
+    public bool CheckEngulfCapability()
```
hhyyrylainen, line 468: I think this should be a computed property, doesn't make sense to me to wrap this tiny bit of code in a method.

src/microbe_stage/MicrobeColony.cs
```diff
@@ -8,10 +8,11 @@
 [UseThriveSerializer]
 public class MicrobeColony
 {
-    private Microbe.MicrobeState state;
+    private MicrobeState state;
 
-    private bool hexCountDirty = true;
+    private bool memberDirty = true;
```
hhyyrylainen, line 13:
```suggestion
    private bool membersDirty = true;
```

src/microbe_stage/Microbe.Contact.cs
hhyyrylainen: Also why isn't this named `CanEngulf` like in the species, that seems a bit confusing to me to name the same thing multiple different things?

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -235,6 +189,21 @@ public float EngulfSize
         }
     }
 
+    /// <summary>
+    ///   Returns true if this microbe or the colony this microbe is part of has the capability to engulf.
+    /// </summary>
+    [JsonIgnore]
+    public bool HasEngulfCapability
```
hhyyrylainen, line 196: I think this *really* needs a different name, I immediately confused this with `CheckEngulfCapability` while reading the changes here. Maybe something like `MakesSenseToAllowEngulfMode` or? something like that that implies that this cell doesn't necessary have the ability to engulf but still it makes sense for it to try to do so.

Kasterisk: Renamed.

src/microbe_stage/Microbe.Contact.cs
Kasterisk: Fixed.

src/microbe_stage/MicrobeColony.cs
```diff
@@ -172,4 +187,17 @@ private void UpdateHexCount()
...
         }
     }
+
+    private void UpdateCanEngulf()
+    {
+        canEngulf = false;
+
+        foreach (var member in ColonyMembers)
+        {
+            canEngulf = member.CanEngulf;
```
hhyyrylainen, line 197: The compiler might optimize this, but I think
```c#
if (!member.CanEngulf)
    continue;
canEngulf = true;
break;
```
would be preferable to show that the canEngulf variable only really needs to be written to once.

src/microbe_stage/MicrobeAI.cs
```diff
@@ -243,7 +243,7 @@ private void ChooseActions(Random random, MicrobeAICommonData data, Microbe? sig
         }
 
         // If there are no threats, look for a chunk to eat
-        if (!microbe.CellTypeProperties.MembraneType.CellWall)
+        if (microbe.CanEngulf)
```
hhyyrylainen, line 246: You don't need to solve this now, but could you put in a TODO comment here that it'd be nice for the AI to know about colonies here and still consider engulfing things if it is in a colony that can engulf something?

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -1142,6 +1115,9 @@ private void HandleUnbinding()
     /// </summary>
     private void HandleEngulfing(float delta)
     {
+        if (!CanEngulf)
```
hhyyrylainen, line 1118: I think this early returning is too early as this skips clearing `attemptingToEngulf` and also it doesn't handle the gradual smooth turning of the engulf audio. I think the cleanest refactor would be to create a new variable `actuallyEngulfing = State == MicrobeState.Engulf && CanEngulf` and replacing the existing checks in this method with that variable.

Kasterisk: Good idea, I'll fix this in a bit.

src/microbe_stage/MicrobeAI.cs
Kasterisk: Added.

hhyyrylainen (approved):
Code change is fine now. Let's try to wait and see if anyone playtests this.

hhyyrylainen:
I playtested this now and everything seemed to work fine with the colonies and single cells couldn't engulf things when they shouldn't.

PR #4145, opened 2023-02-21
Show HUD message for why player cannot engulf something (and a few other messages)

### LINKED ISSUE #3664: If the player can't engulf something they attempted to engulf it should be clarified why
there should be some kind of text popup or something if the player cells ejects something it couldn't engulf, this would really help players understand why they couldn't engulf something

hhyyrylainen:

Shows up to 4 messages that fade over time to the player about various things.

Also included a few code maintenance things by updating some deps and dependabot updates should start working for this repo again.

**Related Issues**

closes #3664

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/locale/en.po
+++ b/locale/en.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
 "POT-Creation-Date: 2023-02-21 15:21+0200\n"
-"PO-Revision-Date: 2023-02-21 15:21+0200\n"
+"PO-Revision-Date: 2023-02-22 09:47+0200\n"
 "Last-Translator: Sur3 <sur3@gmx.de>\n"
 "Language-Team: English <https://translate.revolutionarygamesstudio.com/projects/thrive/thrive-game/en/>\n"
 "Language: en\n"
@@ -2277,7 +2277,7 @@ msgstr "HSV"
 
 #: ../src/microbe_stage/gui/HUDMessages.cs:37
 msgid "HUD_MESSAGE_MULTIPLE"
-msgstr "{0} (x {1})"
+msgstr "{0} (x{1})"
 
 #: ../simulation_parameters/microbe_stage/compounds.json:47
 #: ../src/late_multicellular_stage/MulticellularHUD.tscn:1414
@@ -3914,7 +3914,7 @@ msgstr "Nothing here"
 
 #: ../src/microbe_stage/Microbe.Contact.cs:396
 msgid "NOTICE_DAMAGED_BY_NO_ATP"
-msgstr "Damaged by having no ATP"
+msgstr "Damaged due to not having any ATP"
 
 #: ../src/microbe_stage/Microbe.Interior.cs:1511
 msgid "NOTICE_ENGULF_DAMAGE_FROM_TOXIN"
@@ -3922,15 +3922,15 @@ msgstr "Damaged by toxins in ingested matter"
 
 #: ../src/microbe_stage/Microbe.Interior.cs:1456
 msgid "NOTICE_ENGULF_MISSING_ENZYME"
-msgstr "Cannot engulf target due to missing: {0}"
+msgstr "Missing {0} to engulf target"
 
 #: ../src/microbe_stage/Microbe.Contact.cs:1778
 msgid "NOTICE_ENGULF_SIZE_TOO_SMALL"
-msgstr "Cannot engulf target due to it being too big compared to current size"
+msgstr "Current cell size too small to engulf target"
 
 #: ../src/microbe_stage/Microbe.Contact.cs:1773
 msgid "NOTICE_ENGULF_STORAGE_FULL"
-msgstr "Cannot engulf target due to ingested matter storage not having enough space"
+msgstr "Not enough space in ingested matter storage to engulf target"
 
 #: ../src/general/StageHUDBase.cs:616
 msgid "NOTICE_READY_TO_EDIT"
--- a/locale/en.po
+++ b/locale/en.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2023-02-13 20:52-0800\n"
-"PO-Revision-Date: 2023-02-13 20:55-0800\n"
+"POT-Creation-Date: 2023-02-21 15:12+0200\n"
+"PO-Revision-Date: 2023-02-21 14:04+0200\n"
 "Last-Translator: Sur3 <sur3@gmx.de>\n"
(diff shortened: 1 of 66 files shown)


hhyyrylainen:
I already opened two follow up issues for a bit tricky parts that I didn't really try to mash into this initial implementation (also I ran out of time for today anyway):
- https://github.com/Revolutionary-Games/Thrive/issues/4146
- https://github.com/Revolutionary-Games/Thrive/issues/4147

locale/en.po
```diff
@@ -2275,10 +2275,14 @@ msgstr "HP:"
 msgid "HSV"
 msgstr "HSV"
 
+#: ../src/microbe_stage/gui/HUDMessages.cs:37
+msgid "HUD_MESSAGE_MULTIPLE"
+msgstr "{0} (x {1})"
```
Kasterisk, line 2280: A bit strange to have a space after the x, isn't this usually written as (x2), (x10), and so on? We already have `SPECIES_N_TIMES` too.

locale/en.po
```diff
@@ -3903,11 +3907,35 @@ msgstr "Normal"
...
+msgid "NOTICE_ENGULF_DAMAGE_FROM_TOXIN"
+msgstr "Damaged by toxins in ingested matter"
+
+#: ../src/microbe_stage/Microbe.Interior.cs:1456
+msgid "NOTICE_ENGULF_MISSING_ENZYME"
+msgstr "Cannot engulf target due to missing: {0}"
+
+#: ../src/microbe_stage/Microbe.Contact.cs:1778
+msgid "NOTICE_ENGULF_SIZE_TOO_SMALL"
+msgstr "Cannot engulf target due to it being too big compared to current size"
```
Kasterisk, line 3929:
```suggestion
msgstr "Cannot engulf target due to it being too big compared to current cell size"
```

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -1756,6 +1768,14 @@ private void CheckStartEngulfingOnCandidate(IEngulfable engulfable)
...
             OnEngulfmentStorageFull?.Invoke(this);
+
+            OnNoticeMessage?.Invoke(this,
+                new SimpleHUDMessage(TranslationServer.Translate("NOTICE_ENGULF_STORAGE_FULL")));
+        }
+        else if (EngulfSize < engulfable.EngulfSize * Constants.ENGULF_SIZE_RATIO_REQ)
+        {
+            OnNoticeMessage?.Invoke(this,
+                new SimpleHUDMessage(TranslationServer.Translate("NOTICE_ENGULF_SIZE_TOO_SMALL")));
         }
```
Kasterisk, line 1779: This is also fired when touching friendly species, which shouldn't happen as that implies cannibalism.

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -2215,7 +2223,6 @@ __meta__ = {
 }
 
 [node name="PauseMenu" parent="." instance=ExtResource( 11 )]
-visible = false
```
Kasterisk, line 2218: Ditto.

src/microbe_stage/gui/HUDMessages.tscn
```diff
@@ -0,0 +1,13 @@
+[gd_scene load_steps=3 format=2]
+
+[ext_resource path="res://src/microbe_stage/gui/HUDMessages.cs" type="Script" id=1]
+[ext_resource path="res://src/gui_common/fonts/Lato-Bold-Small.tres" type="DynamicFont" id=2]
+
+[node name="HUDMessages" type="VBoxContainer"]
+pause_mode = 1
+anchor_right = 1.0
+anchor_bottom = 1.0
```
Kasterisk, line 9: Should be anchored to the center top.

src/microbe_stage/gui/HUDMessages.cs
```diff
@@ -0,0 +1,162 @@
...
+            SizeFlagsHorizontal = (int)SizeFlags.ExpandFill,
+            Align = Label.AlignEnum.Center,
+            Autowrap = true,
+            MouseFilter = MouseFilterEnum.Ignore,
+            Text = TextForMessage(message),
+        };
+
+        label.AddFontOverride("font", MessageFont);
+
+        AddChild(label);
```
Kasterisk, line 107: Could also have an option for messages to be stacked top-bottom here, where new messages are placed topmost first. I'd like to reuse this class for the kill feed messages in multiplayer.

Here's for example how I did it: https://github.com/Revolutionary-Games/Thrive/blob/e9d8965cdaa35f7f58f6e54ac4458ff8461ed353/src/microbe_stage/multiplayer/microbial_arena/MicrobialArenaHUD.cs#L166-L167

Kasterisk (changes requested):

locale/en.po
hhyyrylainen: Yeah, that makes sense. When quickly writing out all the English text I didn't even notice this.

src/microbe_stage/gui/HUDMessages.tscn
hhyyrylainen: Well I specifically anchored it on the whole screen as it kind of makes sense that the HUD messages can get extra space.

Kasterisk: Okay, either one works I guess.

src/microbe_stage/Microbe.Contact.cs
hhyyrylainen: I'll refactor the engulf check method to return an enum which explains why it fails so the conditions aren't repeated a bunch here.

src/microbe_stage/MicrobeStage.tscn
hhyyrylainen: Actually this is the expected result. The pause menu is hidden by default, which means Godot should not write the visible property at all (as it is the default value). I recently accepted a PR that added this explicit `visible = false` here, but it seems that Godot will just remove that automatically. So I think we should just keep this change.

src/microbe_stage/gui/HUDMessages.cs
hhyyrylainen: That'll probably need other changes as well, but I'll add a reverse order option to the HUDMessages class.

locale/en.po
```diff
@@ -2275,10 +2275,14 @@ msgstr "HP:"
 msgid "HSV"
 msgstr "HSV"
 
+#: ../src/microbe_stage/gui/HUDMessages.cs:37
+msgid "HUD_MESSAGE_MULTIPLE"
+msgstr "{0} (x{1})"
```
Kasterisk, line 2280: Should `SPECIES_N_TIMES` be replaced with this?

Kasterisk (dismissed):

locale/en.po
hhyyrylainen: I think the usage context is quite different, which is why I did not combine these. For example in Finnish if the "x2" wasn't understandable, different substitute words would make sense to use in these two different contexts.

PR #4149, opened 2023-02-22
Added despawning after player reproduction to stay roughly in entity limit

### LINKED ISSUE #3208: Spawn system should discount cells in colonies in entity limit
I recently realized that the spawn system was trying to despawn colony members. This means that the way it counts the nodes in the scene tree it will also detect colony members as microbes. This means that just a couple of really big colonies will eat up the entity limit entirely. To combat this I think we should optimally change the spawn system so that only the top level Nodes count in the limit (and are subject to despawning).

### LINKED ISSUE #3812: Do something about the gameplay and performance impact of reproduced player colonies that stick around
The player reproduction doesn't follow the entity limit, and also even if it did, it can entirely overwhelm the limit so that no other stuff spawns in the game than player cells.

My initial idea is that we should somehow track the daughter cell / colony of the player and only allow 1 or 2 of those to exist at once, if there are more they could be force despawned (if there are many other entities).

hhyyrylainen:

this should limit how badly the early multicellular prototype especially runs. In microbe stage I had to try really hard to get this to trigger.

also colony members now contribute to the entity limit

**Related Issues**

closes #3812
closes #3208

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -159,6 +159,23 @@ public static class Constants
     public const float REPRODUCTION_ALLOW_EXCEED_ENTITY_LIMIT_MULTIPLIER = 1.15f;
 
     /// <summary>
+    ///   If the entity limit is over this once the player has reproduced, force despawning will happen
+    /// </summary>
+    public const float REPRODUCTION_PLAYER_ALLOWED_ENTITY_LIMIT_EXCEED = 1.25f;
+
+    /// <summary>
+    ///   Once reproduced player copies take this much or more of the overall entity limit, they are preferred to
+    ///   despawn first.
+    /// </summary>
+    public const float PREFER_DESPAWN_PLAYER_REPRODUCED_COPY_AFTER = 0.30f;
+
+    /// <summary>
+    ///   Multiplier for how much cells in a colony contribute to the entity limit. Actually colonies seem quite a bit
+    ///   heavier than normal microbes, as such this is set pretty high.
+    /// </summary>
+    public const float MICROBE_COLONY_MEMBER_ENTITY_WEIGHT_MULTIPLIER = 1.15f;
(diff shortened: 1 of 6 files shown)


buckly90:
The changes are great, and should make long-term sessile gameplay perform much better on lower power PCs, as well as provide some chance that AI cells still spawn as the player species hogs the entity limit.

Unfortunately, I am unsure if this actually makes colony members contribute to the limit... I set my limit to tiny and spawned over 100 cells to form a colony with and AI species still spawned after the formation of my colony. The entity count in performance metrics also suggests that my cells were removed from the count after binding.

hhyyrylainen:
> Unfortunately, I am unsure if this actually makes colony members contribute to the limit... I set my limit to tiny and spawned over 100 cells to form a colony with and AI species still spawned after the formation of my colony. The entity count in performance metrics also suggests that my cells were removed from the count after binding.

Ah, well the entity limit is only updated when an entity is spawned or destroyed. Which means that forming a colony probably doesn't update things. I tested in early multicellular where it definitely looked like as the player grew the extra cells were added to the entity limit.

Trying to recalculate all of the entity weights all the time for existing entities doesn't sound that great, but I'll investigate the situation a bit more tomorrow as to why the entities would disappear from the limit but not add to the limit. Did you use the metrics panel to check the entity counts in realtime while playing?

hhyyrylainen:
I was wrong, full entity weights *are* recomputed pretty often when checking if thigns should be despawned.

What I failed to realize is that the player cell is not counted in the entity limit and thus the player cell colony is also not counted in the entity limit, I've opened a separate issue for that: https://github.com/Revolutionary-Games/Thrive/issues/4153

With that I'm pretty happy with my changes now so I think I'll merge this today.

PR #4154, opened 2023-02-23
Converted mutation point cost to float

### LINKED ISSUE #4150: Rigidity slider breaks when trying to move it at exactly 0 MP
It looks like rigidity slider does a divide by zero to end up with ridiculously wrong numbers when it tries to move when there's exactly 0 MP left. This can be triggered with a very specific MP cost multiplier to be able to place parts to reach 0 MP and then trying to move the rigidity slider.

https://community.revolutionarygamesstudio.com/t/membrane-rigidity-visual-bug/6059

Video: https://youtu.be/9c4ekW5QNI8

Athriq:

As discussed in #4150. MP is rounded to 3 fractional digits in the GUI elements.

Of course needs further testing.

**Related Issues**

Fixes #4150.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I think MP when the value is over something like 5 should be rounded to a whole number, under that something like one decimal, I  think, would be nice. Only for really small values like under 0.1 I think should we show more decimals. And the MP bar should also never show decimals unless it is really low (just to distinguish between 0 and a tiny bit of MP).

Kasterisk:
Under normal conditions (no MP discounts and no mutation cost multiplier), the MP will always seem rounded, even with decimals, the fractional digits are usually so small that it's not worth to do anything special.

Here are videos showing what I'm talking about.

Without multipliers:

https://user-images.githubusercontent.com/54026083/220969104-11dabdc1-f1d0-44e9-9bf3-eb866986023c.mp4

<br>

With multipliers:

https://user-images.githubusercontent.com/54026083/220969273-e3953f76-e34f-4a78-b975-5862df3951f6.mp4

hhyyrylainen:
I think the MP bar suddenly having decimals randomly isn't that good. I think the same logic for hiding the decimals in most cases could be used there and elsewhere for consistency.

src/general/MathUtils.cs
```diff
@@ -131,4 +131,31 @@ public static int PositiveModulo(this int val, int mod)
...
+        else if (value > 0.1f)
+        {
+            result = Math.Round(value, 1);
+        }
+        else
+        {
+            result = Math.Round(value, 3);
+        }
+
+        return (float)result;
```
hhyyrylainen, line 159: There's an extra `(float)` cast here.

hhyyrylainen: Also this method could just return the result instead of having a separate variable for it in each of the if cases.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -95,7 +95,7 @@ private void SetMembraneTooltips(MembraneType referenceMembrane)
     /// <summary>
     ///   Updates the fluidity / rigidity slider tooltip
     /// </summary>
-    private void SetRigiditySliderTooltip(int rigidity)
+    private void SetRigiditySliderTooltip(float rigidity)
```
hhyyrylainen, line 98: Shouldn't rigidity be kept as an integer? That's an unrelated change to switching to MP being a float.

src/microbe_stage/editor/tooltips/SelectionMenuToolTip.cs
```diff
@@ -334,7 +334,7 @@ private void UpdateMpCost()
         if (mpLabel == null)
             return;
 
-        mpLabel.Text = mpCost.ToString(CultureInfo.CurrentCulture);
+        mpLabel.Text = Math.Round(mpCost, 3).ToString(CultureInfo.CurrentCulture);
```
hhyyrylainen, line 337: This doesn't use that fancy math utils rounding. I think it maybe should, or?

src/microbe_stage/editor/MicrobePartSelection.cs
```diff
@@ -156,7 +156,7 @@ private void UpdateLabels()
         if (mpLabel == null || nameLabel == null)
             return;
 
-        mpLabel.Text = MPCost.ToString(CultureInfo.CurrentCulture);
+        mpLabel.Text = MathUtils.RoundWithMinimalDecimals(MPCost, 5).ToString(CultureInfo.CurrentCulture);
```
hhyyrylainen, line 159: This 5 here should *probably* be a constant as this is likely needed in multiple places at some point.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
Kasterisk: The signal actually binds a float and with the change to float MP might as well just use float here for consistency.

src/microbe_stage/editor/tooltips/SelectionMenuToolTip.cs
Kasterisk: Fixed.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
hhyyrylainen: I did notice quite a few places where it was also changed. I think for logical correctness enforcing that the rigidity slider value is a full int at certain points would make a lot of sense, I don't want there to be potential for bugs where partial steps are applied in some situations but then the slider can't go to the exact value due to the step size set in the GUI and the display gets out of sync with reality.

Kasterisk: Changing rigidity slider's logic back to int brings back #4150 in full force, where to go about this?

hhyyrylainen: I think someone needs to take a look at the math to find out why trying to create the rigidity step at no MP causes the divide by zero. I would guess that the reason it works with a float is that it ends up with a really small number instead of 0 and then all the math rounds out in display. But that isn't really desirable to me, I'd like the rigidity math to be fixed / behaviour fixed so that it just skips allowing the change to proceed when the incorrect condition is triggered.

Kasterisk: So apparently, when the cost multiplier is low enough (like 0.2), the cost per step for rigidity is rounded down to 0 and this is the source of the divide by zero problem.  Now my most logical solution in this circumstance would be to cap the rigidity's minimum cost to 1.

hhyyrylainen: That makes quite a lot of sense, though maybe with the fractional MP now being visible limiting it down to 0.5 or such a low value, would also work. I think a consistent "cap minimum price" method should be created so all parts of the editor work the same way when all game setting and multicellularity discounts are applied which could result in various things being very cheap.

hhyyrylainen:
Is the rigidity handling for when no new step is permitted due to being out of MP now handled correctly here? Those are the only files in this PR I've not marked as read by me.

Kasterisk:
Working on it.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -969,12 +969,19 @@ public void OnRigidityChanged(int desiredRigidity)
...
-            int stepsToCutOff = (int)Math.Ceiling((cost - Editor.MutationPoints) / costPerStep);
-            data.NewRigidity -= (desiredRigidity - previousRigidity > 0 ? 1 : -1) * stepsToCutOff /
-                Constants.MEMBRANE_RIGIDITY_SLIDER_TO_VALUE_RATIO;
+            // TODO: This doesn't work correctly in low editor MP with significantly reduced costs.
+            //       Supposedly unaffordable sliding action makes the tooltip indicate stats changes (even though
+            //       the grabber itself doesn't seem to move) and I can't figure out the maths to fix it - kasterisk
+            //
+            // int stepsToCutOff = (int)Math.Ceiling((cost - Editor.MutationPoints) / costPerStep);
+            // data.NewRigidity -= (desiredRigidity - previousRigidity > 0 ? 1 : -1) * stepsToCutOff /
+            //    Constants.MEMBRANE_RIGIDITY_SLIDER_TO_VALUE_RATIO;
```
Kasterisk, line 978: This code made the slider able to stop at a maximum affordable MP when you jump out of the affordable range.

I can't figure out the maths for this, this always breaks at certain cost multipliers so I commented this out and the glitchy bug is fixed at least.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1100,15 +1107,28 @@ public bool ApplyOrganelleUpgrade(OrganelleUpgradeActionData actionData)
...
+    ///   Adjusts the given cost by a multiplier and clamps it to a minimum of
+    ///   <see cref="Constants.MINIMUM_MUTATION_POINTS_COST"/> and a maximum of
+    ///   <see cref="Constants.BASE_MUTATION_POINTS"/>.
+    /// </summary>
+    /// <param name="cost">The action cost.</param>
+    /// <returns>The adjusted cost.</returns>
+    protected float GetAdjustedCost(float cost)
+    {
+        return Mathf.Clamp(
+            cost * CostMultiplier, Constants.MINIMUM_MUTATION_POINTS_COST, Constants.BASE_MUTATION_POINTS);
```
84634E1A607A, line 1120: Maybe this is the cause that breaks the mutation point math.

Kasterisk: This is required to cap MP at a minimum as hh requested, a fix in spite of this is needed.

hhyyrylainen:
I assume there's still some fixing to be done on this, @-mention me if I miss when this is actually functionally ready.

hhyyrylainen:

PR #4160, opened 2023-02-24
Throw exception if an instance is already registered in InputManager

### LINKED ISSUE #2520: Add a system to check if the object has been registered before in InputManager.RegisterReceiver()
**Describe the bug**

Once I made CheatMenu a ControlWithInput, then I can't open it by F6. Later I realized that it was registered twice, one in ControlWithInput, the other in CheatMenu itself, but there was no warning, which caused a lot of confusions.

I'm thinking if some approach can be taken to warn if the object to register has been registered before.

Kasterisk:

Which also prints out the instance's type name for easy debugging.

**Related Issues**

Closes #2520.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/input/InputManager.cs
+++ b/src/engine/input/InputManager.cs
@@ -99,6 +99,9 @@ public class InputManager : Node
                      .attributes
                      .Where(p => p.Key.Method?.DeclaringType?.IsInstanceOfType(instance) == true))
         {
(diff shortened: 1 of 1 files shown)


hhyyrylainen (approved):
I tested most of the parts of the game and this doesn't seem to break anything. Adding a duplicate call to `RegisterReceiver` I was able to confirm that the exception is thrown.

PR #4168, opened 2023-02-25
Rearranged new game settings buttons

### LINKED ISSUE #4167: Show confirmation dialog when pressing confirm in the new game settings
Once I have accidentally started the game while fiddling with the new game settings, I thought the Confirm button was a Back button. I can imagine new players doing the same mistake.

Kasterisk:

To avoid misclicks.

- Rearranged new game settings buttons
- Permanently dismissable popups now have their own custom dialog type. Should be easier to add more of them now.
- CustomConfirmationDialog now uses CustomRichTextLabel instead of plain Label.
- As a result, they're very vulnerable to being stuck. Thus I've moved the "fix" done in #3916 to the general `PopupCenteredShrink` method.

**Related Issues**

Closes #4167.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/locale/en.po
+++ b/locale/en.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2023-02-28 19:12+0700\n"
-"PO-Revision-Date: 2023-02-28 19:14+0700\n"
+"POT-Creation-Date: 2023-02-28 19:54+0700\n"
+"PO-Revision-Date: 2023-02-28 19:56+0700\n"
 "Last-Translator: Henri Hyyryläinen <hhyyrylainen@revolutionarygamesstudio.com>\n"
 "Language-Team: English <https://translate.revolutionarygamesstudio.com/projects/thrive/thrive-game/en/>\n"
 "Language: en\n"
@@ -1481,6 +1481,14 @@ msgstr ""
 "This shows how many popups have been permanently dismissed by the user.\n"
 "If some popups that are actually wanted are dismissed, the button next to this can be used to make all dismissed popups appear again."
 
+#: ../src/gui_common/dialogs/PermanentlyDismissableDialog.cs:30
+msgid "DISMISS_INFORMATION_PERMANENTLY"
+msgstr "Don't show this again"
+
+#: ../src/gui_common/dialogs/PermanentlyDismissableDialog.cs:33
+msgid "DISMISS_WARNING_PERMANENTLY"
+msgstr "Don't warn about this again"
+
 #: ../src/general/OptionsMenu.tscn:610
 msgid "DISPLAY_ABILITIES_BAR"
 msgstr "Display abilities bar"
@@ -1523,10 +1531,6 @@ msgstr "Double click to view in fullscreen"
 #: ../src/gui_common/tooltip/ToolTipManager.tscn:2726
 msgid "DOUBLE_MEMBRANE_DESCRIPTION"
 msgstr "A membrane with two layers, it has better protection against damage and takes less energy to not deform. However, it slows the cell down some and lowers the rate at which it can absorb resources."
-
-#: ../src/gui_common/dialogs/PermanentlyDismissableDialog.tscn:47
-msgid "DO_NOT_SHOW_AGAIN"
-msgstr "Don't show this again"
 
 #: ../src/engine/DebugOverlays.tscn:138
 msgid "DUMP_SCENE_TREE"
@@ -3580,11 +3584,11 @@ msgstr "Internal (Folder) Name:"
 msgid "MOD_LICENSE"
 msgstr "Mod License:"
 
-#: ../src/general/MainMenu.tscn:973
+#: ../src/general/MainMenu.tscn:974
 msgid "MOD_LOAD_ERRORS"
 msgstr "Mod Loading Errors"
 
-#: ../src/general/MainMenu.tscn:974
+#: ../src/general/MainMenu.tscn:975
 msgid "MOD_LOAD_ERRORS_OCCURRED"
 msgstr "Errors occurred when loading one or more mods. Logs may contain additional information."
 
@@ -4941,7 +4945,7 @@ msgstr "Rusticyanin is a protein able to use [thrive:compound type=\"carbondioxi
 msgid "RUSTICYANIN_PROCESSES_DESCRIPTION"
 msgstr "Turns [thrive:compound type=\"iron\"][/thrive:compound] into [thrive:compound type=\"atp\"][/thrive:compound]. Rate scales with concentration of [thrive:compound type=\"carbondioxide\"][/thrive:compound] and [thrive:compound type=\"oxygen\"][/thrive:compound]."
 
-#: ../src/general/MainMenu.tscn:969
+#: ../src/general/MainMenu.tscn:970
 msgid "SAFE_MODE_EXPLANATION"
 msgstr ""
 "Thrive has started in safe mode due to a previous start failure.\n"
@@ -4953,7 +4957,7 @@ msgstr ""
 "\n"
 "If you don't fix the issue causing the start failure you will need to restart and crash Thrive multiple times to get back to safe mode."
 
-#: ../src/general/MainMenu.tscn:967
+#: ../src/general/MainMenu.tscn:968
 msgid "SAFE_MODE_TITLE"
 msgstr "Thrive Started In Safe Mode"
 
@@ -5506,11 +5510,11 @@ msgstr "Steam is unavailable currently, please retry"
 msgid "STEAM_ERROR_UNKNOWN"
 msgstr "Unknown Steam error happened"
 
-#: ../src/general/MainMenu.tscn:987
+#: ../src/general/MainMenu.tscn:988
 msgid "STEAM_INIT_FAILED"
 msgstr "Steam Initialization Failed"
 
-#: ../src/general/MainMenu.tscn:989
+#: ../src/general/MainMenu.tscn:990
 msgid "STEAM_INIT_FAILED_DESCRIPTION"
 msgstr ""
 "Steam client library initialization has failed. Steam features will be unavailable.\n"
@@ -5667,7 +5671,7 @@ msgstr "Temp."
 msgid "TESTING_TEAM"
 msgstr "Testing Team"
 
-#: ../src/general/MainMenu.cs:225 ../src/general/MainMenu.tscn:981
+#: ../src/general/MainMenu.cs:225 ../src/general/MainMenu.tscn:982
 msgid "THANKS_FOR_BUYING_THRIVE"
 msgstr ""
 "Thank you for supporting Thrive development by purchasing this copy of Thrive!\n"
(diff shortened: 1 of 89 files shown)


hhyyrylainen:
Seeing as there's changes to the dismissible popup, I guess you made it so that the "are you sure you want to start a new game" dialog permanently dismissible?

Kasterisk:
I didn't, just decided to refactor those dialogs out of whim.

hhyyrylainen:
Last night I suddenly had a realization: what if the button was changed to say "start game" or "begin game"? Then it would be super clear what it does. That solves the problem of accidentally clicking it and doesn't bog down the new user experience with an extra click before they can start playing.

Kasterisk:
This is what I come up with:

https://user-images.githubusercontent.com/54026083/221404138-609a4d75-847a-4c8b-b838-4797cfa056cc.mp4

<br>

Basically separating out the advanced button to distinguish the Start button more. In advanced view, the Back button is replaced with Basic so users don't potentially mistake it for returning to basic view.

hhyyrylainen:
> Basically separating out the advanced button to distinguish the Start button more. In advanced view, the Back button is replaced with Basic so users don't potentially mistake it for returning to basic view.

I think that's a pretty good solution.

src/gui_common/dialogs/PermanentlyDismissableDialog.tscn
```diff
@@ -0,0 +1,84 @@
...
+margin_right = 178.0
+margin_bottom = 52.0
+
+[node name="CheckBox" parent="VBoxContainer" index="5" instance=ExtResource( 2 )]
+margin_top = 52.0
+margin_right = 178.0
+margin_bottom = 69.0
+custom_fonts/font = ExtResource( 4 )
+pressed = true
+text = "DO_NOT_SHOW_AGAIN"
```
hhyyrylainen, line 47: Removing the don't show again is not very good, the reason that text exists is that some dismissible popups are not warnings. So the text that "don't warn again" does not fit all usages.

src/gui_common/dialogs/CustomConfirmationDialog.tscn
```diff
@@ -3,11 +3,11 @@
 [ext_resource path="res://src/gui_common/dialogs/CustomConfirmationDialog.cs" type="Script" id=1]
 [ext_resource path="res://src/gui_common/thrive_theme.tres" type="Theme" id=2]
 [ext_resource path="res://src/gui_common/fonts/Lato-Bold-AlmostSmaller.tres" type="DynamicFont" id=3]
-[ext_resource path="res://src/gui_common/fonts/Lato-Regular-Smaller.tres" type="DynamicFont" id=4]
+[ext_resource path="res://src/gui_common/CustomRichTextLabel.tscn" type="PackedScene" id=4]
```
hhyyrylainen, line 6: Why does this now have a rich text label? There's no place that I'm aware of that would require rich text in a confirmation dialog.

src/gui_common/dialogs/CustomConfirmationDialog.cs
```diff
@@ -139,7 +143,17 @@ private void UpdateLabel()
         if (dialogLabel == null)
             throw new SceneTreeAttachRequired();
 
-        dialogLabel.Text = TranslationServer.Translate(dialogText);
+        var text = new StringBuilder(100);
+
+        if (CenterText)
+            text.Append("[center]");
```
hhyyrylainen, line 149: Instead of centering the text like this, I think it would be so much better to use the text alignment on the label to do this. This is because rich text is expensive operation to parse and should not be used when it is avoidable.

src/general/NewGameSettings.tscn
```diff
@@ -1190,54 +1192,54 @@ __meta__ = {
...
-margin_left = 363.0
+[node name="Start" type="Button" parent="CenterContainer/VBoxContainer/HBoxContainer"]
+margin_left = 477.0
 margin_right = 550.0
 margin_bottom = 35.0
 hint_tooltip = "CONFIRM_NEW_GAME_BUTTON_TOOLTIP"
 focus_neighbour_left = NodePath("../Advanced")
 focus_previous = NodePath("../Advanced")
-text = "CONFIRM_NORMAL"
+text = "START"
```
hhyyrylainen, line 1242: I think the key here should really be "START_GAME" to be clear what it is for, even if the English text is not as clear.

src/general/MainMenu.cs
```diff
@@ -234,11 +221,11 @@ public override void _Process(float delta)
                     GD.Print("We are most likely a store version of Thrive, showing the thanks dialog");
 
                     // The text has a store link template, so we need to update the right links into it
-                    thanksDialogText.ExtendedBbcode =
+                    thanksDialog.DialogText =
                         TranslationServer.Translate("THANKS_FOR_BUYING_THRIVE").FormatSafe(storeBuyLink);
 
                     // This isn't strictly necessary but might make the fix to this popup more robust
```
hhyyrylainen, line 227: This comment is no longer accurate.

src/general/MainMenu.cs
```diff
@@ -234,11 +221,11 @@ public override void _Process(float delta)
                     GD.Print("We are most likely a store version of Thrive, showing the thanks dialog");
 
                     // The text has a store link template, so we need to update the right links into it
-                    thanksDialogText.ExtendedBbcode =
+                    thanksDialog.DialogText =
```
hhyyrylainen, line 224: I got this far and realized why the use of rich text in the dialog is needed. Still the centering should just set the text alignment property, surely that's still better than extra complicating the markup handling?

src/general/MainMenu.tscn
```diff
@@ -1019,51 +973,14 @@ rect_min_size = Vector2( 400, 350 )
...
-margin_right = 268.0
-margin_bottom = 32.0
-size_flags_horizontal = 6
-custom_fonts/font = ExtResource( 53 )
-text = "OK"
+WindowTitle = "THANK_YOU_TITLE"
+HideCancelButton = true
+DialogText = "THANKS_FOR_BUYING_THRIVE"
+CenterText = false
+NoticeType = 1
```
hhyyrylainen, line 983: Here is the case where dismissing a warning is completely incorrect. We should not warn the user of the fact that they bought the game and support us...

src/gui_common/dialogs/PermanentlyDismissableDialog.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        checkbox = GetNode<CustomCheckBox>("VBoxContainer/CheckBox");
+    }
+
+    /// <summary>
+    ///   Shows this notice by first checking whether the <see cref="NoticeType"/> for this has been
+    ///   permanently dismissed and if so cancels popup.
+    /// </summary>
+    /// <returns><see langword="true"/> if hasn't been permanently dismissed yet.</returns>
+    public bool Popup()
```
hhyyrylainen, line 24: This doesn't seem to overwrite anything? I think this should be named `PopupIfNotDismissed` to be clear where this method is from and what it does.

src/gui_common/dialogs/PermanentlyDismissableDialog.cs
```diff
@@ -0,0 +1,40 @@
...
+        if (!Settings.Instance.IsNoticePermanentlyDismissed(NoticeType))
+        {
+            this.PopupCenteredShrink();
+            return true;
+        }
+
+        return false;
+    }
+
+    private void OnHide()
```
hhyyrylainen, line 35: Doesn't the custom dialog already have our custom callback thing for being hidden, so why is there a non-overridden method like this here?

hhyyrylainen (changes requested):
I saw a lot of small things to improve here to get the refactoring to be very good quality.

src/general/NewGameSettings.tscn
Kasterisk: I think the context is already clear, Start Game would make it too verbose.

src/general/MainMenu.cs
Kasterisk: I would if I could but there's no alignment property in RTL, everything is done through bbcode.

src/gui_common/dialogs/PermanentlyDismissableDialog.cs
Kasterisk: I guess I missed it.

src/gui_common/dialogs/PermanentlyDismissableDialog.tscn
Kasterisk: Though, it's a bit amusing how the translation key itself does not reflect the actual text 😁

src/general/NewGameSettings.tscn
hhyyrylainen: The context is not clear, in Finnish translation "start" cannot be translated without context. That's why this translation key has to be "START_GAME" and for the record I'd prefer for the English text to be "Start Game" to be super clear. This is a very important button so having longer text on it, I think is actually a benefit.

src/gui_common/dialogs/PermanentlyDismissableDialog.tscn
hhyyrylainen: That's the power of the translation keys, we can adjust the text to taste whenever we want without having to modify the code.

src/general/NewGameSettings.tscn
Kasterisk: Okay, changed.

locale/en.po
```diff
@@ -1532,9 +1524,9 @@ msgstr "Double click to view in fullscreen"
 msgid "DOUBLE_MEMBRANE_DESCRIPTION"
 msgstr "A membrane with two layers, it has better protection against damage and takes less energy to not deform. However, it slows the cell down some and lowers the rate at which it can absorb resources."
 
-#: ../src/general/MainMenu.tscn:995
+#: ../src/gui_common/dialogs/PermanentlyDismissableDialog.tscn:47
 msgid "DO_NOT_SHOW_AGAIN"
-msgstr "Do not warn about this again"
+msgstr "Don't show this again"
```
hhyyrylainen, line 1529: There's still just the one text... I very intentionally made two variants, one for "warnings" and one for general info dialogs that can be dismissed permanently.

hhyyrylainen: If you want you can rename them "DISMISS_WARNING_PERMANENTLY" and "DISMISS_INFORMATION_PERMANENTLY" to clarify the usage of the text.

src/gui_common/dialogs/CustomConfirmationDialog.cs
```diff
@@ -139,7 +143,17 @@ private void UpdateLabel()
         if (dialogLabel == null)
             throw new SceneTreeAttachRequired();
 
-        dialogLabel.Text = TranslationServer.Translate(dialogText);
+        var text = new StringBuilder(100);
+
+        if (CenterText)
```
hhyyrylainen, line 148: I think for non-centered text it would be very nice to have a higher performance variant where the StringBuilder is not used at all.

Kasterisk: Done.

hhyyrylainen:
I don't see any more problems in the code. I'll give this a playtest later today, especially to confirm the thanks for buying popup still functions correctly.

hhyyrylainen (approved):
I guess this works fine even though the smart text default font is smaller than the normal confirmation dialog font, so after these changes the no mods enabled warning popup has smaller font than other similar popups that are closed with the OK button.

hhyyrylainen:
I decided the GUI inconsistency will bother me too much to leave in so I opened a high priority issue: https://github.com/Revolutionary-Games/Thrive/issues/4184

PR #4210, opened 2023-03-05
Allow floating chunks to be detected in the inspector panel + general refactoring

### LINKED ISSUE #1710: At Cursor feature doesn't detect any toxins

Kasterisk:

For reusability, `PlayerHoverInfo` is now renamed into `PlayerInspectInfo`, it is now responsible for the detection of `IInspectableEntity` objects in the game world. This class can be extended by stage-specific derivations to implement stage-specific functions, as exemplified by `MicrobeInspectInfo` where it handle the detection of compound clouds.

Detection is done by raycasting from the camera to what the mouse is pointing at and then checking whatever it intersects. Hopefully with this, the system can be reused as much as possible in other stages for any purpose.

Complementing this entities detection change is the refactor of the hover panel/looking at panel/inspector panel to allow adding detection categories dynamically and consolidating the Node structure of the label controls for detected entities. Descriptions of detected entities can now be easily added, it can either be the amount indicator of compound clouds (_an abundance, a fair amount, etc.._), the n times indicator (_x5, x10, ..._) or any other useful texts to explain to the player what he/she is looking at and what he/she can do with it.

**Related Issues**

Closes #1710.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/StageHUDBase.cs
```diff
@@ -529,7 +498,13 @@ public override void _Process(float delta)
 
         UpdateATP(delta);
         UpdateHealth(delta);
-        UpdateHoverInfo(delta);
+
+        hoverInfoTimeElapsed += delta;
+        if (hoverInfoTimeElapsed > Constants.HOVER_PANEL_UPDATE_INTERVAL)
+        {
+            UpdateHoverInfo(delta);
```
hhyyrylainen, line 505: This should pass `hoverInfoTimeElapsed` as this is time limited so the delta is not really the delta of when this method was last called.

src/general/PlayerInspectInfo.cs
```diff
@@ -0,0 +1,71 @@
...
+/// <summary>
+///   A system that manages detecting what the player is pointing with the cursor.
+/// </summary>
+public class PlayerInspectInfo : Node
+{
+    /// <summary>
+    ///   The distance for detection.
+    /// </summary>
+    [Export]
+    public float RayastDistance = 1000;
```
hhyyrylainen, line 14: Typo

src/general/PlayerInspectInfo.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    private readonly List<RaycastResult> hits = new();
+    private readonly HashSet<RaycastResult> previousHits = new();
+
+    /// <summary>
+    ///   All inspectable entities the player is pointing.
```
hhyyrylainen, line 20:
```suggestion
    ///   All inspectable entities the player is pointing at.
```

src/general/PlayerInspectInfo.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// </summary>
+    public IReadOnlyList<IInspectableEntity> InspectableEntities =>
```
hhyyrylainen, line 22: I think this should return an enumerable and leave it up to the caller to decide when they have to make a concrete list out of it.

src/general/PlayerInspectInfo.cs
```diff
@@ -0,0 +1,71 @@
...
+        hits.Select(h => h.Collider).OfType<IInspectableEntity>().ToList();
+
+    public virtual void Process(float delta)
+    {
+        var space = GetViewport().World.DirectSpaceState;
+        var mousePos = GetViewport().GetMousePosition();
+        var camera = GetViewport().GetCamera();
```
hhyyrylainen, line 29: Viewport can be extracteed as a variable.

src/general/PlayerInspectInfo.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        var from = camera.ProjectRayOrigin(mousePos);
+        var to = from + camera.ProjectRayNormal(mousePos) * RayastDistance;
+
+        hits.Clear();
+
+        space.IntersectRay(hits, from, to);
+
+        foreach (var hit in previousHits.ToHashSet())
```
hhyyrylainen, line 38: Using the remove all filtering on the previousHits is better I think rather than having to copy the data here.

src/general/PlayerInspectInfo.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            if (hit.Collider is IInspectableEntity entity)
+            {
+                entity.OnMouseEnter(hit);
+            }
+        }
+    }
+
+    /// <summary>
+    ///   Returns the raycast data of the given raycasted inspectable entity.
```
hhyyrylainen, line 64:
```suggestion
    ///   Returns the raycast data of the given raycast inspectable entity.
```
As raycast is the  past tense for raycast.

hhyyrylainen: Also should this not try to return null when the data doesn't exists, or does it make sense to return blanked out data for non-existent hit?

src/gui_common/MouseHoverPanel.cs
```diff
@@ -0,0 +1,197 @@
...
+            firstVisibleCategory.SeparatorVisible = false;
+
+        nothingHereContainer.Visible = visibleEntriesCount <= 0;
+    }
+
+    public void AddCategory(string internalName, LocalizedString displayName)
+    {
+        if (categories.ContainsKey(internalName))
+        {
+            GD.Print("MouseHoverPanel: Category already exist for \"", internalName, "\"");
```
hhyyrylainen, line 55: Should this be an error level print?

src/gui_common/MouseHoverPanel.cs
```diff
@@ -0,0 +1,197 @@
...
+    public override void _Process(float delta)
+    {
+        var visibleEntriesCount = 0;
+
+        MouseHoverCategory? firstVisibleCategory = null;
+
+        foreach (MouseHoverCategory category in categoriesOrdered)
+        {
+            category.Visible = category.TotalEntriesCount > 0;
+            category.SeparatorVisible = true;
```
hhyyrylainen, line 38: This could be refactored to be `firstVisibleCategory != null` to avoid having to modify this variable twice (the second modification being after the loop).

src/gui_common/MouseHoverPanel.cs
```diff
@@ -0,0 +1,197 @@
...
+    [Export]
+    public NodePath NothingHereContainerPath = null!;
+
+#pragma warning disable CA2213 // Disposable fields should be disposed
+    private Container categoriesContainer = null!;
+    private Container nothingHereContainer = null!;
+#pragma warning restore CA2213 // Disposable fields should be disposed
+
+    private System.Collections.Generic.Dictionary<string, MouseHoverCategory> categories = new();
+    private Array categoriesOrdered = new();
```
hhyyrylainen, line 21:
```suggestion
    private Array categoryControls = new();
```
Would that be a more clear name? From the variable usage I got the impression that this is used for that. The current name `categoriesOrdered` sounds grammatically incorrect so it needs a new name.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -75,6 +78,19 @@ public override void _Ready()
         bindingModeHotkey = GetNode<ActionButton>(BindingModeHotkeyPath);
         unbindAllHotkey = GetNode<ActionButton>(UnbindAllHotkeyPath);
 
+        mouseHoverPanel.AddCategory("compounds", new LocalizedString("COMPOUNDS_COLON"));
+        mouseHoverPanel.AddCategory("species", new LocalizedString("SPECIES_COLON"));
+        mouseHoverPanel.AddCategory("chunk", new LocalizedString("FLOATING_CHUNK_COLON"));
```
hhyyrylainen, line 83: Shouldn't "chunks" be plural here. Also as these are used as keys here, I think either using an enum or a private const variable to hold the string would be better.

With the enum approach the enum needs values for all stages at once in it, but that shouldn't be much of a problem.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -348,6 +354,70 @@ protected override void UpdateAbilitiesHotBar()
...
+
+                category = "species";
+            }
+            else if (entity is FloatingChunk)
+            {
+                category = "chunk";
+            }
+
+            hoveredEntities.TryGetValue((category, entity.InspectableName), out int count);
+            hoveredEntities[(category, entity.InspectableName)] = count + 1;
```
hhyyrylainen, line 409: The tuple on these two lines could be a separate variable to not create the tuple twice.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -348,6 +354,70 @@ protected override void UpdateAbilitiesHotBar()
...
+        }
+
+        foreach (var hoveredEntity in hoveredEntities)
+        {
+            var item = mouseHoverPanel.AddItem(hoveredEntity.Key.Category, hoveredEntity.Key.Name);
```
hhyyrylainen, line 414: Why is item nullable? I think it'd be better to force the AddItem to succeed or throw an exception. I noticed this same thing in a few other places in this PR as well. Core code usage problems should throw exceptions to make the methods nicer to use by for example guaranteeing a return value to be non-null.

hhyyrylainen (changes requested):
Seems like a good change. I had some code clarity and style comments (I submitted a few before starting this review). I didn't test this yet but I didn't see any change to AgentProjectile so that leads me to wonder if those are not detected?

src/general/PlayerInspectInfo.cs
Kasterisk: It returns a default.

hhyyrylainen: Which is a default constructed struct class instance. Which is probably very unhelpful in many cases as knowing between 0 and unset values might be pretty good for not accidentally assuming the both cases (for example sub shape hitting which we already have many bugs about) are the same.

src/microbe_stage/MicrobeHUD.cs
Kasterisk: This is now fixed with const, enum might feel too restrictive.

Kasterisk:
I didn't make that inspectable because it would be very hard to track the projectiles with your mouse anyway, but no harm in making it so.

src/microbe_stage/AgentProjectile.cs
```diff
@@ -22,6 +22,8 @@ public class AgentProjectile : RigidBody, ITimedLife, IEntity
 
     public AliveMarker AliveMarker { get; } = new();
 
+    public string InspectableName => Properties?.ToString() ?? TranslationServer.Translate("N_A");
```
hhyyrylainen, line 25: This should probably have json ignore attribute on it. I know the other properties don't really have it that well in this class, but this is saved like other entities...

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -75,6 +83,16 @@ public override void _Ready()
         bindingModeHotkey = GetNode<ActionButton>(BindingModeHotkeyPath);
         unbindAllHotkey = GetNode<ActionButton>(UnbindAllHotkeyPath);
 
+        mouseHoverPanel.AddCategory(COMPOUNDS_CATEGORY, new LocalizedString("COMPOUNDS_COLON"));
+        mouseHoverPanel.AddCategory(SPECIES_CATEGORY, new LocalizedString("SPECIES_COLON"));
+        mouseHoverPanel.AddCategory(FLOATING_CHUNKS_CATEGORY, new LocalizedString("FLOATING_CHUNKS_COLON"));
+
+        foreach (var compound in SimulationParameters.Instance.GetCloudCompounds())
+        {
+            var hoveredCompoundControl = mouseHoverPanel.AddItem("compounds", compound.Name, compound.LoadedIcon);
```
hhyyrylainen, line 92: This doesn't use the const string?

hhyyrylainen (approved):
The code is good enough now. I'll merge this tomorrow even if the unnecessary class names in exceptions are not removed tomorrow. I'll give that much time to see if anyone has time to playtest this.

hhyyrylainen:
After merging to my branch I noticed there's a bunch of typo warnings, and some actual typos (this is what I get for not opening this PR in Rider), I'll fix those in my prototype branch which I hope to merge this week.

PR #4212, opened 2023-03-07
Fix agent projectiles not being detected in the inspector panel

### LINKED ISSUE #4211: AgentProjectile not detectable by the mouse hover panel (inspector)
Probably because the projectile is not active in any layer (as the tooltip says for the pickable to work the object needs to be on some layer). I also noticed that the AgentProjectile is not marked as pickable, I'll fix that in my branch but if someone wants to work on this before that, take that into consideration.

Here's me with debug shapes on definitely hovering over the projectile and not getting anything in the inspector:

Kasterisk:

Fixes #4211 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -453,6 +453,7 @@ g_interact={
 
 3d_physics/layer_1="Common"
 3d_physics/layer_2="Microbe"
+3d_physics/layer_3="Misc"
 
 [locale]
 
--- a/project.godot
+++ b/project.godot
@@ -453,7 +453,7 @@ g_interact={
 
 3d_physics/layer_1="Common"
 3d_physics/layer_2="Microbe"
-3d_physics/layer_32="Misc"
+3d_physics/layer_31="Misc"
 
 [locale]
 
--- a/project.godot
+++ b/project.godot
@@ -453,7 +453,7 @@ g_interact={
 
 3d_physics/layer_1="Common"
 3d_physics/layer_2="Microbe"
-3d_physics/layer_3="Misc"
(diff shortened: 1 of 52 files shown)


project.godot
```diff
@@ -453,6 +453,7 @@ g_interact={
 
 3d_physics/layer_1="Common"
 3d_physics/layer_2="Microbe"
+3d_physics/layer_3="Misc"
```
hhyyrylainen, line 456: Can't the layer be 32? It'd be better to leave a bunch of empty spaces for other more useful stuff to be earlier.

Kasterisk: Hold on, that seem to mess with the raycast collision mask.

hhyyrylainen: I was just about to comment on that. Maybe layer 32 is a special debug layer, and our last real layer should be 31?
I see in the raycast that you copied the Godot default values for the mask, that might be what excludes the last layers.

hhyyrylainen: Probably best to use the last layer that Godot allows to be detected by raycasts by default so that there's no surprising bug when someone uses a mix of our raycast and Godot raycast method, which could happen if we have a different default mask.

Kasterisk: `AgentProjectile` need layers set in 31 and 32 then?

Kasterisk: Okay, I think I get it. Layer 31 is now reserved for `Misc` instead.

hhyyrylainen (approved):
Can confirm this now fixes the bug.

PR #4216, opened 2023-03-10
Update Netanalyzers to 7.0.1 Consistently

HyperbolicHadron:

This PR updates the RevolutionaryGamesCommon submodule. This makes the Microsoft.CodeAnalysis.NetAnalyzers version consistent across the project.

**Related Issues**

This PR closes #4215  by including its commits alongside the corresponding change to the common submodule so that multiple versions of the analyzers are not used across the project.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -7,6 +7,6 @@
         <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.435" PrivateAssets="all" />
         <AdditionalFiles Include="$(MSBuildThisFileDirectory)RevolutionaryGamesCommon/stylecop.json"
                          Link="stylecop.json" />
-        <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.0" PrivateAssets="all" />
+        <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.1" PrivateAssets="all" />
     </ItemGroup>
 </Project>
--- a/third_party/Directory.Build.props
+++ b/third_party/Directory.Build.props
@@ -10,6 +10,6 @@
         <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.435" PrivateAssets="all" />
         <AdditionalFiles Include="$(MSBuildThisFileDirectory)../RevolutionaryGamesCommon/stylecop.json"
                          Link="stylecop.json" />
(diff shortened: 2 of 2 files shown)


84634E1A607A:
Thanks for the PR, but I think that #4215 (maybe?) does exactly the same thing and there should also be another PR by dependabot in revocommon module. These things are often done by hh because it may need extra work.

Anyway, thanks for contributing!

HyperbolicHadron:
> Thanks for the PR, but I think that #4215 (maybe?) does exactly the same thing and there should also be another PR by dependabot in revocommon module. These things are often done by hh because it may need extra work.
> 
> Anyway, thanks for contributing!

I thought that might be the case, but this PR would make sure that the RevolutionaryGamesCommon submodule is updated at the same time, and so there would not need to be a separate PR for updating the submodule. (This PR does the submodule update that the other PR would need to add in order to be complete)

84634E1A607A:
Yeah, but as far as I know we'll just make a commit directly to master to update the submodule, instead of another PR.

HyperbolicHadron:
> Yeah, but as far as I know we'll just make a commit directly to master to update the submodule, instead of another PR.

Would that be after the #4215 PR? I would think that the PR and the submodule update would need to be done at the same time to avoid consistency problems. This is the first time I have done a PR related to a submodule, so I may have the wrong idea about how this works.

HyperbolicHadron:
I experienced the consistency problem firsthand when I was using the commit in the PR #4215 without realizing that the submodule needed to be updated as well, and continued to have issues until I remembered to update it. I would think that it would be best practice to have the PR simultaneously update the dependent submodule as well so that there would be no possibility of an inconsistent state of the project split between two different netanalyzers versions after the PR is accepted, and before the separate update of the submodule to master.

hhyyrylainen:
> I would think that the PR and the submodule update would need to be done at the same time to avoid consistency problems. 

I would have taken at most like a couple of minutes after merging the dependabot PR. But I'll accept this PR if this does it all at once.

Due to the way Weblate works (or rather doesn't by auto breaking translation files) I have a bunch of experience with PRs where I have to immediately after merging do a quick fix commit to master.

hhyyrylainen (approved):

hhyyrylainen:
Looks like this being marked as closing dependabot's PR was a mistake: https://github.com/Revolutionary-Games/Thrive/pull/4215#issuecomment-1463355122

dependabot automatically detects when things are updated and closes its PRs without triggering that version ignore condition, though in this case it should be fine with dependabot updating it the next time there's a new version.

PR #4230, opened 2023-03-16
Make initial compounds proportional to consumption

### LINKED ISSUE #1401: Make initial compounds proportional to player's compound usage
We should be able to calculate what compounds the player cell consumes in a given amount of time, and then when the player respawns, the resource given should be taken into account. Ex: The cell uses 2 glucose, 3 iron, and 1 hydrogen in _x_ amount of time. When the player respawns, they should have 1/2 of the compounds they were given be iron, 1/3 glucose, and 1/6 hydrogen.

ottoblep:

This changes the UpdateInitialCompounds function that sets a cell's starting compounds.
Starting amounts are no longer static and hard-coded per compound but depend on consumption rates. 

Compounds with a positive net balance start at zero while compounds with a negative balance receive a fixed multiple of the consumption. 
This should have the effect that any cell can survive on the starting resources for a fixed minimum amount of time .

Using this method has some up and some downsides as far as i can tell.
From most to least important these are:
- It introduces a cross-dependency on the ProcessSystem. 
It's possible that this should not exist at all, or should be done in a different way. 
Since i'm new here i unfortunately can't tell.
- Any new compound can be introduced without changes
- Consumption rates are dependent on biomeConditions. However, the starting amounts are set only once for each species.
I assumed the default biome for the CompoundBalance calculations.
This leads to inaccuracies, making some cells survive for less time in different biomes. 
A solution could be to use the biome each species is most common in.
- It should not be easily exploitable by the player
- The new constant determining starting amounts could also be added to the difficulty menu later. 
For now i chose a conservative value of 30, which should equal 30 seconds until running out.

**Related Issues**

This feature closes #1401.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/MicrobeSpecies.cs
+++ b/src/microbe_stage/MicrobeSpecies.cs
@@ -1,4 +1,5 @@
-﻿using System.ComponentModel;
+﻿using System;
+using System.ComponentModel;
 using System.Linq;
 using Godot;
 using Newtonsoft.Json;
@@ -101,45 +102,23 @@ public class MicrobeSpecies : Species, ICellProperties, IPhotographable
         BaseRotationSpeed = MicrobeInternalCalculations.CalculateRotationSpeed(Organelles);
     }
 
-    public void SetInitialCompoundsForDefault()
-    {
-        InitialCompounds.Clear();
-        InitialCompounds.Add(SimulationParameters.Instance.GetCompound("atp"), 30);
-        InitialCompounds.Add(SimulationParameters.Instance.GetCompound("glucose"), 10);
-    }
-
-    public void SetInitialCompoundsForIron()
-    {
-        SetInitialCompoundsForDefault();
-        InitialCompounds.Add(SimulationParameters.Instance.GetCompound("iron"), 10);
-    }
-
-    public void SetInitialCompoundsForChemo()
-    {
-        SetInitialCompoundsForDefault();
-        InitialCompounds.Add(SimulationParameters.Instance.GetCompound("hydrogensulfide"), 10);
-    }
-
     public override void UpdateInitialCompounds()
     {
-        var simulation = SimulationParameters.Instance;
+        // Since the initial compounds are only set once per species they can't be calculated for each Biome.
+        // So, the compound balance calculation uses the default biome.
+        var biomeConditions = SimulationParameters.Instance.GetBiome("default").Conditions;
+        var compoundBalances = ProcessSystem.ComputeCompoundBalance(Organelles,
+            biomeConditions, CompoundAmountType.Current);
 
-        var rusticyanin = simulation.GetOrganelleType("rusticyanin");
-        var chemo = simulation.GetOrganelleType("chemoplast");
-        var chemoProtein = simulation.GetOrganelleType("chemoSynthesizingProteins");
+        InitialCompounds.Clear();
 
-        if (Organelles.Any(o => o.Definition == rusticyanin))
+        foreach (var compoundBalance in compoundBalances)
         {
-            SetInitialCompoundsForIron();
-        }
-        else if (Organelles.Any(o => o.Definition == chemo ||
-                     o.Definition == chemoProtein))
-        {
-            SetInitialCompoundsForChemo();
-        }
-        else
-        {
-            SetInitialCompoundsForDefault();
+            // The cell should survive a minimum amount of time.
+            var compoundInitialAmount = Math.Abs(compoundBalance.Value.Balance) * 30;
+            if (compoundInitialAmount > StorageCapacity)
+                compoundInitialAmount = StorageCapacity;
+            InitialCompounds.Add(compoundBalance.Key, compoundInitialAmount);
         }
     }
 


ottoblep:
I noticed the multicellular stages have separate definitions that are still TBD. I will add them here later.

hhyyrylainen (changes requested):
The code change to the functionality looks good. I'll still try to get someone to balance playtest this.

But I have one small request, could you modify the initial compounds methods for early multicell and late multicell species classes to have a TODO comment in them that they should be updated to do a similar thing as for the microbe stage to calculate the actual usage and set the initial compounds based on that?

hhyyrylainen:
> Using the current player patch for balance calculation will probably cause the least confusion.
I haven't found if/how it's possible to access the current player patch from here.

It's not really possible as it doesn't make sense. The complicated but most sensible thing would be to calculate the average of all patches a species is in and use that. That's going to be hugely complicated so I won't request that you do that as ferrying the data through auto-evo and everything that needs to apply species changes would be difficult (and freebuild species generation probably would still anyway need to do the generic patch method as the species is generated first before placing it in a patch).

ottoblep:
I added the TODO to late multicellular. Since early multicellular also starts with a single cell, i copied the same method from microbe there with slight modifications.

hhyyrylainen:
> also starts with a single cell, i copied the same method from microbe there with slight modifications.

Some people have been saying that we should actually prevent that as it is not really "multicellular", but I'm not super keen on that. Anyway, I think your approach of calculating the averages needed by all cells is a pretty good solution. Though maybe it should instead just use the original (first cell type in body layout) to determine the initial needed compounds (and maybe multiply that by 1.5 or so as I think I did that for the previous hard coded values)?

That'll work pretty well for the player. For fully spawned cell colonies its going to be tougher as a single cell obviously can't contain so many resources to sustain a full colony for a while. So that change to using the first cell type with some extra on top for now I think is the best easily doable solution.

buckly90:
Seems to work great and without any apparent errors, great job!

My only concern is that the starting glucose on a new game is rather low. This isn't a huge concern but it does not leave as much room for error as I would like for new players starting their first game.
Granting a static bonus to initial glucose on new game start could ensure new players have time to gather their bearings before finding glucose becomes a problem. This could probably just be it's own issue though.

hhyyrylainen:
I guess there's not easy way to give the player an extra injection of glucose when they first start the game... which might be fine in terms of balancing. Is the 30 seconds survival time this PR is now tweaked for good?

src/early_multicellular_stage/EarlyMulticellularSpecies.cs
```diff
@@ -79,8 +79,7 @@ public override void UpdateInitialCompounds()
                 continue;
 
             // Initial compounds should suffice for a fixed amount of time.
-            var compoundInitialAmount = (Math.Abs(compoundBalance.Value.Balance) / Cells.Count) *
-                Constants.INITIAL_COMPOUND_TIME;
+            var compoundInitialAmount = Math.Abs(compoundBalance.Value.Balance) * Constants.INITIAL_COMPOUND_TIME;
```
hhyyrylainen, line 82: Could you add another constant multiplier here to tweak the single cell value for a colony (1.5 I think would be fine)? Just to get a bit closer to what the behaviour was before and ensure that spawned colonies of the player's species don't die super fast.

ottoblep:
> My only concern is that the starting glucose on a new game is rather low. This isn't a huge concern but it does not leave as much room for error as I would like for new players starting their first game.
Some solutions would be to either give a +1 base level glucose to everyone or maybe only for small cells with less than x organelles or something.

hhyyrylainen:
> maybe only for small cells with less than x organelles.

That could actually work and would not require new features to detect first time player spawns. I like this idea.

ottoblep:
This should make it that bacteria with less than four organelles start with full glucose.

src/microbe_stage/MicrobeSpecies.cs
```diff
@@ -114,6 +114,13 @@ public override void UpdateInitialCompounds()
 
         foreach (var compoundBalance in compoundBalances)
         {
+            if (compoundBalance.Key == SimulationParameters.Instance.GetCompound("glucose") &&
```
hhyyrylainen, line 117: You should look up the compound just once at the start of the method to avoid this lookup by string key each iteration of the loop. Also the organelles count check and IsBacteria should be done before the loop. Something like `bool giveBonusGlucose = Organelles.Count <= 3 && IsBacteria`

ottoblep:
I moved the unnecessary checks out of the loop. Thanks for the heads up.

src/microbe_stage/MicrobeSpecies.cs
```diff
@@ -101,45 +102,34 @@ public void CalculateRotationSpeed()
...
-    }
-
     public override void UpdateInitialCompounds()
     {
-        var simulation = SimulationParameters.Instance;
+        // Since the initial compounds are only set once per species they can't be calculated for each Biome.
+        // So, the compound balance calculation uses the default biome.
+        var biomeConditions = SimulationParameters.Instance.GetBiome("default").Conditions;
+        var compoundBalances = ProcessSystem.ComputeCompoundBalance(Organelles,
+            biomeConditions, CompoundAmountType.Current);
```
hhyyrylainen, line 111: I spotted a bug here, this needs to use the long term compound amounts not ones adjusted for current in game time for the player as that's wrong to use in this evolutionary timescale calculation. I will fix this and convert some things to constants by myself so that I can merge this now and not have to do more reviews.

hhyyrylainen (approved):
I did various improvements and fixed a bug on this myself, I now think this is good to go so I'll merge this once checks pass.

Issue #1924, opened 2020-12-14
Weird region lines in compound cloud absorption


hhyyrylainen:
Here's a video showing what's wrong: https://mega.nz/file/9BQSlLgK#gM5e3I3g31r7Izsw3LFPSAuCaimqngeQZUEDuTJCAlc

Basically it looks like there is some kind of invisible line blocking cells absorbing compounds from the other side.

hhyyrylainen:
Reported in this thread: https://community.revolutionarygamesstudio.com/t/fatal-crash-pushing-iron-chunk-with-big-cell/3211/10?u=hhyyrylainen

hhyyrylainen:
This is more visible now with more extreme currents in some patches.

PR #4234, opened 2023-03-19
Show all chemoreceptor lines in colony

### LINKED ISSUE #3185: All chemoreceptors in a cell colony should work
so that in multicellular the player can have special "smelling" cells and they would work. Right now the guidance lines probably only read the chemoreceptors from the main player cell. Should be relatively simple to change it to look at also other cells in the player's colony if you ask the colony lead cell what chemoreceptors it has.

ottoblep:

This modifies the handling of chemoreceptor lines so that all receptors in the colony are shown, instead of just the player microbe.

**Related Issues**

This fixes #3185.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


ottoblep:
There is still a bug preventing the lines being shown in some cases.
FindCompoundNearPoint sometimes doesn't pick up on clouds for the non-player microbes in the colony.
This can happen either randomly on spawn or after evolution and lasts until restarting the map.
I haven't been able to pinpoint the exact origin of this.
I observed that the bug occurs whenever CompoundCloudPlane Translation changes from 0,0,0 to another value (after dying mainly).
Maybe this causes ContainsPositionWithRadius to evaluate false and skip the distance search of FindCompoundNearPoint.
Might be related to #3035 and #1924.

hhyyrylainen:
Did you check if the chemoreceptor properly uses world positions instead of relative ones? I kind of doubt it is that easy but it could be that simple mistake in the existing code. Also that could give a lead as to what the actual issue is even if that is not the problem.

ottoblep:
>  uses world positions instead of relative ones

Yes, that was exactly it.
GetDetectedCompounds used the node Translation directly, which for Colony cells is relative.
Thus, when the global Translation shifts they move out of the search window.
I changed it to use GlobalTranslation and now it should be fixed.

src/microbe_stage/Microbe.cs
```diff
@@ -907,7 +907,7 @@ public List<(Compound Compound, Color Colour, Vector3 Target)> GetDetectedCompou
         var detections = new List<(Compound Compound, Color Colour, Vector3 Target)>();
         foreach (var (compound, range, minAmount, colour) in activeCompoundDetections)
         {
-            var detectedCompound = clouds.FindCompoundNearPoint(Translation, compound, range, minAmount);
+            var detectedCompound = clouds.FindCompoundNearPoint(GlobalTranslation, compound, range, minAmount);
```
hhyyrylainen, line 910: GlobalTranslation might be a bit expensive to calculate as it needs to find all parent nodes up to the scene root, so having a variable to hold this outside the loop would be nice.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -866,22 +866,44 @@ private void OnPlayerNoticeMessage(Microbe player, IHUDMessage message)
             GD.PrintErr("Chemoreception data reported for non-player cell");
 
         int currentLineIndex = 0;
-        var position = microbe.GlobalTransform.origin;
+        var colonyMicrobes = microbe.Colony?.ColonyMembers ?? new List<Microbe> { microbe };
```
hhyyrylainen, line 869: This is not really what I had in mind for this method. I think a simpler way to implement this, which I think is good enough and would give better performance and less confusing overlapping lines, is that just the player cell would return the detections it wants. So reverting a bunch of the changes here and then modifying `Microbe.GetDetectedCompounds` to take into account all the non-duplicate detections from the colony would be ideal in my opinion. That has the slight drawback that the lines wouldn't becoming out of the cells that have the detectors, but I think that is actually a benefit if the player has a chemoreceptor on each cell, that will end up with a huge mess of lines going everywhere.

hhyyrylainen (changes requested):
I think I'd like to see a different overall approach to this feature, but feel free to discuss if you think your approach is better than my simplified suggestion.

src/microbe_stage/Microbe.cs
ottoblep: I moved it outside the loop.

ottoblep:
This is a much leaner version of the fix.
Everything is now handled in the 'Microbe.GetDetectedCompounds' of the player microbe.
It collects all 'Microbe.activeCompoundDetections' in the colony and then runs the search once.
There is some awkward iterating through the hashset but i don't think it's worth changing 'activeCompoundDetections'.

src/microbe_stage/Microbe.cs
```diff
@@ -904,10 +904,25 @@ public void AIThink(float delta, Random random, MicrobeAICommonData data)
...
     {
+        var colonyMicrobes = Colony?.ColonyMembers ?? new List<Microbe> { this };
+        HashSet<(Compound Compound, float Range, float MinAmount, Color Colour)> colonyCompoundDetections = new();
+
+        foreach (var colonyMicrobe in colonyMicrobes)
+        {
+            foreach (var activeCompoundDetection in colonyMicrobe.activeCompoundDetections)
+            {
+                // For multiple Range/MinAmount settings this just chooses the first one
+                if (!colonyCompoundDetections.Any(o => o.Compound == activeCompoundDetection.Compound))
```
hhyyrylainen, line 915: Why do you do this? HashSet ensures already that no duplicate items are added.

src/microbe_stage/Microbe.cs
```diff
@@ -904,10 +904,25 @@ public void AIThink(float delta, Random random, MicrobeAICommonData data)
     /// </returns>
     public List<(Compound Compound, Color Colour, Vector3 Target)> GetDetectedCompounds(CompoundCloudSystem clouds)
     {
+        var colonyMicrobes = Colony?.ColonyMembers ?? new List<Microbe> { this };
```
hhyyrylainen, line 907: The code should have a performance shortcut for non-colonies. So you should have a check if the Colony is null and then just do the simple logic and only if the colony is not null and this is the colony lead cell, then do the more complex operation.

src/microbe_stage/Microbe.cs
hhyyrylainen: That's how it works with non-colony receptors as well. It's intentional implementation by me. So I'm 100% fine with it working that way for colonies as well.

src/microbe_stage/Microbe.cs
```diff
@@ -904,10 +904,34 @@ public void AIThink(float delta, Random random, MicrobeAICommonData data)
     /// </returns>
     public List<(Compound Compound, Color Colour, Vector3 Target)> GetDetectedCompounds(CompoundCloudSystem clouds)
     {
+        HashSet<(Compound Compound, float Range, float MinAmount, Color Colour)> colonyCompoundDetections = new();
```
hhyyrylainen, line 907: This still allocates an unnecessary object for the non-colony case.

ottoblep: I moved it.

src/microbe_stage/Microbe.cs
```diff
@@ -904,10 +904,34 @@ public void AIThink(float delta, Random random, MicrobeAICommonData data)
...
+        {
+            var colonyMicrobes = Colony.ColonyMembers;
+            foreach (var colonyMicrobe in colonyMicrobes)
+            {
+                foreach (var activeCompoundDetection in colonyMicrobe.activeCompoundDetections)
+                {
+                    // This is necessary to prevent duplicates for one compound because
+                    // changing any value of the tuple in the hashset like Range/MinAmount makes the element unique.
+                    // For multiple Range/MinAmount settings this just chooses the first one.
+                    if (!colonyCompoundDetections.Any(o => o.Compound == activeCompoundDetection.Compound))
```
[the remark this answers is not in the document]
ottoblep: It now uses `UnionWith` again like originally.

src/microbe_stage/Microbe.cs
```diff
@@ -904,10 +904,28 @@ public void AIThink(float delta, Random random, MicrobeAICommonData data)
     /// </returns>
     public List<(Compound Compound, Color Colour, Vector3 Target)> GetDetectedCompounds(CompoundCloudSystem clouds)
     {
+        HashSet<(Compound Compound, float Range, float MinAmount, Color Colour)> colonyCompoundDetections;
```
hhyyrylainen, line 907: This variable doesn't actually always hold the "colony" detections, so the name should be something else to actually describe what it holds.

src/microbe_stage/Microbe.cs
```diff
@@ -904,10 +904,28 @@ public void AIThink(float delta, Random random, MicrobeAICommonData data)
...
+
+        if (Colony != null)
+        {
+            colonyCompoundDetections = new HashSet<(Compound Compound, float Range, float MinAmount, Color Colour)>();
+            var colonyMicrobes = Colony.ColonyMembers;
```
hhyyrylainen, line 912: This is an unnecessary intermediate variable.

src/microbe_stage/Microbe.cs
```diff
@@ -904,10 +904,29 @@ public void AIThink(float delta, Random random, MicrobeAICommonData data)
     /// </returns>
     public List<(Compound Compound, Color Colour, Vector3 Target)> GetDetectedCompounds(CompoundCloudSystem clouds)
     {
+        HashSet<(Compound Compound, float Range, float MinAmount, Color Colour)> collectedUniqueCompoundDetections;
+
+        if (Colony != null)
```
hhyyrylainen, line 909: This needs to be so that this *only* triggers for the colony leader. I just playtested and this'll be the last needed change.

hhyyrylainen: If you can't make this change right away, I'll make the change so that I can finish testing this and get on with my other branch.

hhyyrylainen (approved):
Thanks for working on this feature and getting this pretty highly requested thing done. If you are interested, you could officially join the team and cite your PRs as samples of your work.

Code looks fine now with my latest change and I quickly tested this ingame that this seems to work correctly. If ever anyone complains about the detections being hard to get exactly the same to have the near duplicates count as duplicates, I'll open an issue for handling that then.

src/microbe_stage/Microbe.cs
ottoblep: Should be gone now.

src/microbe_stage/Microbe.cs
ottoblep: I renamed it to `collectedUniqueCompoundDetections`.

src/microbe_stage/Microbe.cs
hhyyrylainen: Yep, I saw your commit earlier.

hhyyrylainen:
This page on our website has info on how to send the application for joining the team: https://revolutionarygamesstudio.com/application/

Issue #4195, opened 2023-03-01
Classic map is not correctly setup for day/night cycle


hhyyrylainen:
Got report that the photosynthesis ATP balance and tooltips(?) don't work correctly when using the classic map with day/night cycle enabled.

I suspect that the classic map data is just not updated and it doesn't have the new day/night cycle properties in it.

https://discord.com/channels/228300288023461893/958598553389903913/1080414996418277458

hhyyrylainen:
Closing this in favour of just removing the classic option: https://github.com/Revolutionary-Games/Thrive/issues/4200

Issue #3527, opened 2022-07-14
Classic patch map view should show all patches at default position


hhyyrylainen:
instead of having the top patch cut off like this:
![2022-07-14_17 53 40 2978](https://user-images.githubusercontent.com/3796411/179012318-5849c785-4c42-419a-bdc5-539d62352741.png)

Also the patch extinction screen should be made as good fitting as possible at the default view position.

hhyyrylainen:
Closing this in favour of just removing the classic option: https://github.com/Revolutionary-Games/Thrive/issues/4200

PR #4249, opened 2023-03-26
Removed classic patch map option, and deleted relevant code.

### LINKED ISSUE #4200: Remove the classic map option as it has extra bugs (and no one really seems to want to work on it)
just so that there's no trap in the new game settings for getting a worse experience.

This one bug was open for a really long time without anyone giving it a thought really:
- https://github.com/Revolutionary-Games/Thrive/issues/3527

No one (at least on the dev team) seems to play the classic map that often so I think it'll be overall better to just remove the option instead of hoping that we don't break it anymore with further game changes.

Creidhne86:
I removed the classic patch map from the game. Translations were also updated based on the removal.

closes #4200 

Other issues: #4195, #3527 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/BiomeType.cs
+++ b/src/microbe_stage/BiomeType.cs
@@ -3,7 +3,7 @@
 /// </summary>
 /// <remarks>
 ///   <para>
-///     The order of this enum CANNOT be changed. They are saved in the gamesave file, 
+///     The order of this enum CANNOT be changed. They are saved in the gamesave file,
 ///     and re-ordering would break existing saves
 ///   </para>
 /// </remarks>
--- a/src/microbe_stage/BiomeType.cs
+++ b/src/microbe_stage/BiomeType.cs
@@ -3,7 +3,7 @@
 /// </summary>
 /// <remarks>
 ///   <para>
-///     The order of this enum CANNOT be changed. They are saved in the gamesave file,
+///     The order of this enum CANNOT be changed. They are saved in game save files,
 ///     and re-ordering would break existing saves
 ///   </para>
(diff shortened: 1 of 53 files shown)


hhyyrylainen:
This needs to have the CI check automatically found problems fixed. Do you need help with those?
Jetbrains is complaining about an extra blank line between an XML comment and the code element it describes. Format script fails because the localization script has not been ran to update things.

src/microbe_stage/BiomeType.cs
```diff
@@ -1,12 +1,7 @@
 ﻿/// <summary>
 ///   This contains predefined biome types
 /// </summary>
-/// <remarks>
-///   <para>
-///     The order of this enum CANNOT be changed. They are used as patch IDs in classic map,
```
hhyyrylainen, line 6: This comment actually needs to be kept, but reworded to say that the order of the enum values may not be changed due to saves saving the values. So reordering the values will break all existing saves, which is very bad.

Creidhne86:
> 

I wasn't aware of the translate update.. I found this documentation and I'll ask questions if I'm not able to figure it out.

https://github.com/Revolutionary-Games/Thrive/blob/master/doc/working_with_translations.md

Creidhne86:
@hhyyrylainen seems like everything is passing now, thanks for the help!

hhyyrylainen (approved):
I tested this and didn't see any Godot errors or leftover GUI for the feature to be removed. I noticed one typo reported by my IDE which I fixed, so this should be all good to go now.

Issue #4117, opened 2023-02-09
Make multihex organelles send better positions to the membrane to improve wrapping


hhyyrylainen:
I suddenly remembered last night that someone said some time ago that the reason why some organelles can poke outside the membrane is that they only send their center coordinate to the membrane. Instead I think we'd be much better off if each organelle sent the farthest from origin hex's coordinate. That way the membrane probably would stay far enough from big organelles so that they wouldn't poke outside it.

adQuid:
If you wanna cheat here was my take on a solution: https://github.com/adQuid/Thrive/pull/74/files#diff-55af6451a36834f2255a91efffdce5b5ed4581058e45aed647ec31912584588aR422

hhyyrylainen:
Now with myofibril being a really long membrane, they still want to poke through the membrane even with the latest fix as they are just so long and thin.

PR #4251, opened 2023-03-27
Make membrane generation algorithm faster

### LINKED ISSUE #1274: Fix membrane generation for really big cells
The membrane stops working on really big cells. Especially if they are long and thin.

Leinourdian:

This PR reworks the membrane generation algorithm to make it faster. It also fixes a couple of bugs at the same time. This has been discussed on [the community forum](https://community.revolutionarygamesstudio.com/t/reducing-lag-from-membrane-updates/6237). The membranes were tweaked to look a bit wavier and a bit less smooth compared to the pictures in the post (now a single hex cell with a cell wall doesn't look like a perfect circle anymore).

**Related Issues**

Fixes #1274

Makes #4117 a  lot rarer. It can still happen e.g. if the cell is a very long and thin line of organelles.

Fixes a visible seam in cell walls.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Membrane.cs
+++ b/src/microbe_stage/Membrane.cs
@@ -9,12 +9,6 @@ using Array = Godot.Collections.Array;
 /// </summary>
 public class Membrane : MeshInstance, IComputedMembraneData
 {
-    /// <summary>
-    ///   This must be big enough that no organelle can be at this position.
-    ///   TODO: maybe switching to nullable float would be a good alternative?
-    /// </summary>
-    public const float INVALID_FOUND_ORGANELLE = -999999.0f;
-
     [Export]
     public ShaderMaterial? MaterialToEdit;
 
@@ -24,10 +18,6 @@ public class Membrane : MeshInstance, IComputedMembraneData
     ///   Stores the generated 2-Dimensional membrane. Needed for contains calculations
     /// </summary>
     private readonly List<Vector2> vertices2D = new();
-
-    // Work buffers used when generating membrane data
-    private List<Vector2> previousWorkBuffer = new();
-    private List<Vector2> nextWorkBuffer = new();
 
     private float healthFraction = 1.0f;
     private float wigglyNess = 1.0f;
@@ -328,21 +318,20 @@ public class Membrane : MeshInstance, IComputedMembraneData
     }
 
     /// <summary>
-    ///   Return the position of the closest organelle to the target point if it is less then a certain threshold away.
+    ///   Return the position of the closest organelle to the target point.
     /// </summary>
     public Vector2 FindClosestOrganelles(Vector2 target)
     {
-        // The distance we want the membrane to be from the organelles squared.
-        float closestSoFar = 4;
-        Vector2 closest = new Vector2(INVALID_FOUND_ORGANELLE, INVALID_FOUND_ORGANELLE);
+        float closestDistanceSoFar = float.MaxValue;
+        Vector2 closest = new Vector2(0.0f, 0.0f);
 
         foreach (var pos in OrganellePositions)
         {
             float lenToObject = (target - pos).LengthSquared();
 
-            if (lenToObject < 4 && lenToObject < closestSoFar)
+            if (lenToObject < closestDistanceSoFar)
             {
-                closestSoFar = lenToObject;
+                closestDistanceSoFar = lenToObject;
                 closest = pos;
             }
         }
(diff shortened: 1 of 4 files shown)


Deus-Codes:
I messed around with this a bit and I didn't notice anything broken. I made cells with funky shapes and everything looked very smooth; very nice improvement to the look of things. I didn't test on multicellular organisms, but I can imagine this messing with that much.

Thank you very much for this great addition!

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -610,14 +610,19 @@ public void SendOrganellePositionsToMembrane()
             throw new InvalidOperationException("Microbe must be initialized first");
 
         var organellePositions = new List<Vector2>();
+        bool hasMultihexOrganelles = false;
```
hhyyrylainen, line 613: Is this a workaround for https://github.com/Revolutionary-Games/Thrive/issues/4117 ? If so could you add a TODO comment here linking to that issue?

src/microbe_stage/Membrane.cs
```diff
@@ -411,52 +388,40 @@ private void InitializeMesh()
             }
         }
 
-        previousWorkBuffer.Capacity = vertices2D.Capacity;
-        nextWorkBuffer.Capacity = previousWorkBuffer.Capacity;
+        // Make the length longer to guarantee that everything fits easily inside the square
+        cellDimensions *= 100;
+
+        var startingBuffer = new List<Vector2>();
```
hhyyrylainen, line 394: To save on memory allocations, could you make this a field in this class? Like `private List<Vector2> previousWorkBuffer` used to be. And just clear it here before reusing it.

src/microbe_stage/Membrane.cs
```diff
@@ -411,52 +388,40 @@ private void InitializeMesh()
...
-        // Copy final vertex data from the work buffer
+        // Get new membrane points
         vertices2D.Clear();
-
-        // The work buffer not being pointed to as the next, is the one we should read the result from
-        vertices2D.AddRange(previousWorkBuffer);
-
-        previousWorkBuffer.Clear();
-        nextWorkBuffer.Clear();
+        vertices2D.AddRange(GenerateMembranePoints(startingBuffer));
```
hhyyrylainen, line 424: Could you refactor this as well so that GenerateMembranePoints takes in the list and adds the results directly, so that it can avoid having to allocate another temporary list?

src/microbe_stage/Microbe.Contact.cs
hhyyrylainen: If that is the case the issue should also be linked from the documentation on `HasMultihexOrganelles`

Leinourdian: Yeah, it's a workaround. With the comment, do you mean something like `// TODO: This is a workaround for https://github.com/Revolutionary-Games/Thrive/issues/4117. Check if this can be removed once that issue is fixed.`?

hhyyrylainen: Yeah, that's about fine wording for the TODO comment. (though maybe remove the `.` after the link in case it makes the link unclickable for someone as some editors might consider the dot to be part of the link incorrectly).

src/microbe_stage/Membrane.cs
```diff
@@ -25,9 +19,8 @@ public class Membrane : MeshInstance, IComputedMembraneData
     /// </summary>
     private readonly List<Vector2> vertices2D = new();
 
-    // Work buffers used when generating membrane data
-    private List<Vector2> previousWorkBuffer = new();
-    private List<Vector2> nextWorkBuffer = new();
+    // Buffer for starting points when generating membrane data
```
hhyyrylainen, line 22: This should be an xml comment as this is documenting a field

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -611,13 +611,21 @@ public void SendOrganellePositionsToMembrane()
 
         var organellePositions = new List<Vector2>();
 
+        // This is a workaround for https://github.com/Revolutionary-Games/Thrive/issues/4117
```
hhyyrylainen, line 614: This also needs to be an XML comment.

src/microbe_stage/Membrane.cs
Leinourdian: Ah, I see. Fixed.

src/microbe_stage/Microbe.Contact.cs
Leinourdian: The format checks didn't allow an XML comment for a local variable, so I left this as is.

hhyyrylainen: Oh sorry, I didn't realize that this was not a field. I thought for sure that this was a field considering the vector above `var organellePositions = new List<Vector2>();` But yeah, I see now, this needs to be basically like this.

PR #4252, opened 2023-03-28
Added a warning badge to editors' finish button.

Kasterisk:

A warning badge will be shown on the button's top right to indicate that something requires player attention before finishing.

demo

A concern was voiced about the badge's color not conforming to any existing colors. I can't think of a good alternative so unless someone comes forth with a suggestion, I'll leave it as is.

**Related Issues**

_None, this is just a random idea I had._

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/assets/textures/gui/bevel/warningBadge.png
+++ b/assets/textures/gui/bevel/warningBadge.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:22143acc00a90d2c7ebee436e8dd0dd4c52f866f3a4f4704e42bd35e248b3480
+size 3554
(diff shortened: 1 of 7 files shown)


hhyyrylainen (approved):
I made that one method protected (and slightly reordered one other thing for the method order to make more sense). Testing in game I didn't really notice any problems but I think maybe a different colour would have been better for the warning icon, but I don't really have a suggestion as to what it could be instead so I'll accept this.

Kasterisk:
Thanks for changing that, I'd do it but was pretty tied up yesterday.

> I think maybe a different colour would have been better for the warning icon, but I don't really have a suggestion as to what it could be instead so I'll accept this.

We could open an issue if you'd like.

hhyyrylainen:
Sure, I think this will be entirely forgotten otherwise.

Issue #4201, opened 2023-03-02
Art gallery sometimes opens with the popup not fitting on screen


hhyyrylainen:
causing the titlebar and the bottom buttons to not be visible (might be a general bug in the custom dialog full rect):
![2023-03-02_14 21 42 9851](https://user-images.githubusercontent.com/3796411/222427689-c9b70e2f-caef-4175-9db7-7bca248ea9a5.png)

The issue sorts itself out if the game window is resized

hhyyrylainen:
Nope, this problem still exists, this time it aligns so that the bottom buttons are visible but no content is visible, which is a bit worse even as the player will just see that it is an empty screen and might never check it again.

hhyyrylainen:
Might be gone with Godot 4 upgrade as I've not seen this anytime recently.

Issue #2695, opened 2021-10-24
Add popup open and close animations for custom windows (dialogs)


athariqk:
Add default basic open and close animations that every custom dialogs should play. Can be overridden as demonstrated by current [TutorialDialog](https://github.com/Revolutionary-Games/Thrive/blob/master/src/gui_common/dialogs/TutorialDialog.cs) class.

Some relevant discussion: https://github.com/Revolutionary-Games/Thrive/pull/2387#commitcomment-56661411

PR #4260, opened 2023-04-01
Refactor custom popups

### LINKED ISSUE #4206: Permanently dismissible popups with no cancel option should focus the OK button
Instead of the checkbox for toggling it being permanently dismissed.

### LINKED ISSUE #4207: Permanently dismissible popups when they are canceled should not dismiss permanently
If we ask the user if they want to do a harmful action, they say no, but don't uncheck the box, then the next time they attempt that action they don't get the warning. This is not desirable so the dismissible popup should ignore the checkbox status when it is canceled or closed with the X (if the cancel button is visible).

Kasterisk:

`CustomDialog` and its derivatives now inherits from one single `CustomPopup` base class which itself inherits from plain `Control`. The modality of these custom popups are managed by a separate `ModalManager` system. In short we're now in control of all the modal popup functionalities. All of the changes made makes it possible for popups to implement custom hiding behavior (closing animations). An example of which can be seen in the organelle popup menu below:

menu

And the possibilities are near limitless.
Using a plain `Popup` should now be discouraged as it _might_ conflict with the custom system.

**Related Issues**

Enables #2695.
Fixes #4206.
Fixes #4207.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Does this system handle the keyboard focus situation if multiple popups open at once and the player is expected to be able to close them one by one with the keyboard? I noticed that the inbuilt Godot system for popups seems to correctly return the focus to the second most top popup when the top one is closed.

Kasterisk:
That's bugged out currently but I think I found a fix. There's another separate but related problem regarding FocusGrabber though which I brought up in the discord chat.

84634E1A607A:
I must warn that due to the fact that Godot 4.0 changed popup logic greatly, these changes may need to be redone completely when we switch to Godot 4.

hhyyrylainen:
I haven't looked at the code yet, but has the exclusive popup feature been implemented?

src/gui_common/ModalManager.cs
```diff
@@ -0,0 +1,189 @@
+﻿using System.Collections.Generic;
+using Godot;
+using Godot.Collections;
+
+/// <summary>
+///   Handles a stack of <see cref="CustomPopup"/>s that blocks GUI inputs.
```
hhyyrylainen, line 6: Don't we have a ton of popup derived things that can be left open while other stuff is done? I feel like this and some other of the comments in this PR seem to imply that, that very common use case would no longer work.

Kasterisk: They still work, they're not mentioned anywhere here because they don't need modality. For non-modal popups, setting them just as top-level with `SetAsTopLevel` is sufficient.

Kasterisk:
> I haven't looked at the code yet, but has the exclusive popup feature been implemented?

Yes. I replicated all the built-in popup functionality that I'm aware of.

locale/en.po
```diff
@@ -6755,6 +6752,9 @@ msgstr "Zoom in"
 msgid "ZOOM_OUT"
 msgstr "Zoom out"
 
+#~ msgid "CONTINUE_THRIVING"
```
hhyyrylainen, line 6755: Why is this text deleted? I thought that this was pretty clever way to start the multicellular stage.

src/awakening_stage/gui/InventoryScreen.tscn
```diff
@@ -293,12 +293,12 @@ text = "CRAFTING_TAKE_ALL"
 anims/Flash = SubResource( 1 )
 anims/RESET = SubResource( 2 )
 
-[connection signal="Closed" from="GroundPanel" to="." method="OnGroundPanelClosed"]
-[connection signal="Closed" from="MainInventory" to="." method="OnInventoryPanelClosed"]
+[connection signal="Dismissed" from="GroundPanel" to="." method="OnGroundPanelClosed"]
```
hhyyrylainen, line 296: What's the reason why the this signal was renamed? I think "closed" is pretty simple and very clear for when a popup emits that signal.

src/general/MainMenu.tscn
```diff
@@ -903,6 +903,7 @@ margin_top = 28.0
 margin_bottom = 628.0
 
 [node name="GalleryViewer" parent="." instance=ExtResource( 17 )]
+Exclusive = false
```
hhyyrylainen, line 906: I think we have more non-exclusive popups so having the exclusive property default to false would be good in my opinion.

src/general/OptionsMenu.tscn
```diff
@@ -2011,7 +2011,6 @@ margin_top = 360.0
 margin_right = 940.0
 margin_bottom = 360.0
 rect_min_size = Vector2( 600, 0 )
-popup_exclusive = true
```
hhyyrylainen, line 2014: Shouldn't this at least be exclusive as the player *has* to do something to resolve the going back problem before they can do anything else?

src/gui_common/FocusGrabber.cs
```diff
@@ -53,7 +53,10 @@ public override void _Ready()
     {
         if (string.IsNullOrWhiteSpace(NodeToGiveFocusTo))
             throw new ArgumentException("Focus grabber must have the node to focus set");
+    }
 
+    public override void _EnterTree()
+    {
         UpdateOverrideFocusStrings();
```
hhyyrylainen, line 60: There's actually a comment within `UpdateOverrideFocusStrings` about needing to update things while inside the tree. Could you resolve that? This is now a bit of a code cleanness decreasing change as the comment in that method doesn't match reality anymore...

src/gui_common/art_gallery/GalleryViewer.cs
```diff
@@ -333,7 +333,7 @@ private void OnAssetPreviewOpened(GalleryCard item)
             return;
 
         slideScreen.CurrentSlideIndex = CurrentCards.IndexOf(item);
-        slideScreen.CustomShow();
+        slideScreen.Close();
```
hhyyrylainen, line 336: Is there a typo here? The call to `CustomShow` is replaced by `Close`.

src/awakening_stage/gui/InventoryScreen.tscn
Kasterisk: Due to ambiguity with the new `CustomPopup.Close` and for reinforcing what I had documented in this signal's xml summary.

src/engine/ScreenOverlays.tscn
```diff
@@ -13,7 +13,7 @@ pause_mode = 2
 process_priority = 2000
 
 [node name="CanvasLayer" type="CanvasLayer" parent="."]
-layer = 128
+layer = 127
```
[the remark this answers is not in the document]
Kasterisk: Yeah it's pretty tricky since the render priority of these two scenes sometimes intertwine. Perhaps the color filter could be in its own dedicated canvas layer which renders above all.

src/gui_common/CustomPopup.cs
```diff
@@ -0,0 +1,132 @@
+﻿using Godot;
+
+/// <summary>
+///   A custom popup that gives us more control over <see cref="Popup"/>.
```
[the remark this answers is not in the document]
Kasterisk: I suppose CustomWindow would be a good alternative (going by Godot 4's terminology)?

Popup, dialogs, closable panels are in a way a window. I used Popup here as an umbrella term for controls that are top level.

src/gui_common/FocusGrabber.cs
Kasterisk: I don't fully understand what that comment means, reparenting makes the node exits the tree for a bit so doesn't this change has no relation to what the comment is saying?

src/gui_common/RadialPopup.cs
```diff
@@ -55,6 +55,12 @@ public bool RadialPopupCanceled()
         return true;
     }
 
+    protected override void OnHidden()
+    {
+        base.OnHidden();
+        OnClosed();
```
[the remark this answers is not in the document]
Kasterisk: Yes, it makes sense to simplify.

Though, I see `OnCancelled(int itemId)` seem to take in one argument, but `EmitSignal(nameof(OnCanceled));` never pass any arguments, why is this and should the `itemId` parameter be removed?

src/gui_common/CustomPopup.cs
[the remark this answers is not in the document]
hhyyrylainen: I think that would be a worthwhile separation to create with the naming. Btw I just remembered that we have some stuff regarding the popups in the styleguide so that document should also be updated.

src/awakening_stage/gui/InventoryScreen.tscn
hhyyrylainen: Okay, I read the documentation there. But now I feel like in 100% of cases I've seen the X closing the dialog needs to be handled the same way as pressing the cancel button or hitting ESC. So having a separate dismiss operation that is called a bunch doesn't seems very useful. Or did you mean it as a replacement for canceling? If that's the case I think the name should be like "Canceled" for clarity.

src/engine/ScreenOverlays.tscn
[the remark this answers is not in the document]
hhyyrylainen: What's the overlay that *shouldn't* be on top of everything else?

src/gui_common/FocusGrabber.cs
hhyyrylainen: Reading the comment, I *think* I had a idea in my mind on how the focus grabber could be more performant when it needs to refresh the nodes it is pointing to. If you can't figure out any way to make the code faster for the reparent case, then the comment should be just removed.

src/gui_common/RadialPopup.cs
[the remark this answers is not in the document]
hhyyrylainen: My guess is that it is a copy-paste mistake and it is safe to remove that unused parameter.

src/engine/ScreenOverlays.tscn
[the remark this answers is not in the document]
Kasterisk: There's render hierarchy even in ScreenOverlays.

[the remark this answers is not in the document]
hhyyrylainen: But that's just an effect of the scene order they are defined, right? So it is not relevant when discussing that all-non overlays should be underneath the overlays. Except I guess exclusive popups really should be on top.

Maybe reserving layer 126 for normal overlays would be good, then exclusive popups can be 127, and finally absolutely on top of everything being effect layers would be 128?

src/gui_common/FocusGrabber.cs
Kasterisk: I guess I'll just remove it.

src/gui_common/RadialPopup.cs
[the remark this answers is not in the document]
Kasterisk: The signal itself isn't used so I guess it can be just removed entirely...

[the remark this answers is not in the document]
hhyyrylainen: I added the signal because I foresee a need for something to know if the user cancels picking a thing from an action wheel. Maybe with nested action wheels or maybe in the future the action wheel would be used in a modal way so that we need to know if the action is cancelled to clean up other state.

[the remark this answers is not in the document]
Kasterisk: Keeping this would be redundant anyway since there's already the built-in CanvasItem's `hide` signal.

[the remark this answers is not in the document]
hhyyrylainen: I guess sort of, but I think the intention is much more clear when the "user cancelled picking something" is communicated clearly. Going back to my multi level action wheel example, that *might* get implemented so that the first wheel is hidden when the top level item is selected. And that would mean that using the generic hidden signal would not work and instead the domain specific signal would be much superior.

src/engine/ScreenOverlays.tscn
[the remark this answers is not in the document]
Kasterisk: Sounds good.

src/awakening_stage/gui/InventoryScreen.tscn
Kasterisk: Keep in mind this class is one level below confirmation dialogs, so it shouldn't know what "canceling" is.

src/gui_common/CustomPopup.cs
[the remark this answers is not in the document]
Kasterisk: Done, thanks for reminding.

locale/en.po
hhyyrylainen, line 1: These localization changes can now be entirely removed from this PR as there's no actual text changes.

src/engine/ControlHelpers.cs
```diff
@@ -11,32 +11,40 @@ public static class ControlHelpers
...
             });
         }
     }
 
+    public static void MoveToCenter(this Control popup)
+    {
+        // "Refresh" control to correct its size
+        popup.RectSize = Vector2.Zero;
+
+        var parentRect = popup.GetViewport().GetVisibleRect();
```
hhyyrylainen, line 42: Isn't this incorrect? The documentation on GetVisibleRect is:
```
/// <para>Returns the visible rectangle in global screen coordinates.</para>
```

I think the correct code would be: `popup.GetViewportRect()`. I just ran into this problem myself when implementing mouse window edge panning.

src/gui_common/CustomPopupMenu.cs
```diff
@@ -0,0 +1,91 @@
+﻿using Godot;
+
+public class CustomPopupMenu : CustomWindow
```
hhyyrylainen, line 3: Documentation summary on this general GUI part class would be nice.

src/gui_common/RadialPopup.tscn
```diff
@@ -1,25 +1,29 @@
...
-[node name="RadialPopup" type="Popup"]
+[node name="RadialPopup" instance=ExtResource( 3 )]
 anchor_right = 1.0
 anchor_bottom = 1.0
-popup_exclusive = true
+margin_right = 0.0
+margin_bottom = 0.0
 script = ExtResource( 1 )
+Exclusive = true
+ExclusiveAllowCloseOnEscape = false
```
hhyyrylainen, line 14: I think it would be nice for the cancel (ESC) button to also dismiss a radial popup so it is easier to get out of one in case it gets stuck on.

src/gui_common/art_gallery/GalleryViewer.tscn
```diff
@@ -31,12 +32,12 @@ GalleryCardModelScene = ExtResource( 4 )
 GalleryCardAudioScene = ExtResource( 6 )
 GalleryDetailsToolTipScene = ExtResource( 8 )
 
-[node name="MarginContainer" type="MarginContainer" parent="."]
+[node name="MarginContainer" type="MarginContainer" parent="." index="0"]
```
hhyyrylainen, line 35: What's up with these bunch of `index=` that got added in the scenes in this PR? If you reopen Godot and save all the scenes again, do these go away?

src/gui_common/dialogs/CustomConfirmationDialog.cs
```diff
@@ -172,6 +175,6 @@ private void OnCancelPressed()
     {
         GUICommon.Instance.PlayButtonPressSound();
         Hide();
-        EmitSignal(nameof(Cancelled));
+        EmitSignal(nameof(Closed));
```
hhyyrylainen, line 178: Shouldn't "Hide" already trigger the `Closed` signal?

src/microbe_stage/ProcessPanel.cs
```diff
@@ -73,6 +64,6 @@ protected override void Dispose(bool disposing)
     private void OnClosePressed()
     {
         GUICommon.Instance.PlayButtonPressSound();
-        CustomHide();
+        Close();
```
hhyyrylainen, line 67: This kind of small method seems pretty common, maybe it would make sense for the base window to have a method that can be used as the button signal target? That way these couple of lines don't need to be duplicated in a bunch of derived classes.

src/microbe_stage/editor/HexPopupMenu.cs
```diff
@@ -163,6 +152,16 @@ public bool OnMoveKeyPressed()
...
     }
 
+    protected override void ResolveNodeReferences()
+    {
+        titleLabel = GetNode<Label>(TitleLabelPath);
+        deleteButton = GetNode<Button>(DeleteButtonPath);
+        moveButton = GetNode<Button>(MoveButtonPath);
+        modifyButton = GetNode<Button>(ModifyButtonPath);
+
+        UpdateModifyButton();
```
hhyyrylainen, line 162: Logically this line doesn't really belong in `ResolveNodeReferences` is it not possible to override _Ready in this class to have this there?

src/gui_common/art_gallery/GalleryViewer.tscn
Kasterisk: No, it does not.

src/gui_common/dialogs/CustomConfirmationDialog.cs
Kasterisk: No? the signal documentation explicitly states that it's not made for that purpose.

hhyyrylainen: That's very confusing to me... `Closed` should *always* trigger after the dialog stops being visible.

src/gui_common/art_gallery/GalleryViewer.tscn
hhyyrylainen: I guess fine, we'll just have more and more of these in the project... I just checked and only a couple of our scene files currently suffer from this problem, this PR would at least double the number of files that have this strange thing going on.

src/gui_common/dialogs/CustomConfirmationDialog.cs
Kasterisk: I think this is a name problem, choosing the name `Canceled` it is...

hhyyrylainen: Yeah, the situation will be a lot better if all of the different things have clearly named signals:
- Closed: always when closed
- Cancelled: when rejected
- Accepted or maybe Success

So closed would always be accompanied with one other signal, and the code using the system can then make the decision what it wants to do and when.

src/microbe_stage/ProcessPanel.cs
Kasterisk: Done.

src/gui_common/dialogs/CustomConfirmationDialog.cs
Kasterisk: Done.

hhyyrylainen: I feel like the cancelled signal is overused. For example:

```
[connection signal="Canceled" from="PerformanceMetrics" to="." method="OnPerformanceMetricsToggled"]
```
Why does the callback *actually* need to know that the popup was cancelled, rather than it closed for *any* reason? Wouldn't that second one be much, much more bulletproof. The code doesn't care *why* it closed, just that it did and update other GUI state.

Also the signal name should be Cancelled, with two l's as my spell checker complains about "canceled" being a typo.

Kasterisk: > I feel like the cancelled signal is overused. For example:

I don't know, this was decided by the person who made it. I just tweaked the signal's name.

> Why does the callback actually need to know that the popup was cancelled, rather than it closed for any reason?

So that it's possible to separate out `Confirmed` from the equation. An example of what I'm talking about: you set up a callback registered to a "Closed/Hidden" signal and another callback to a "Confirmed" signal, you clicked confirm and instead of seeing a single operation executed by the "Confirmed" callback, you get both "Confirmed" and "Cancelled" at the same time since "Confirm" also closes the popup most of the time.

hhyyrylainen: I agreed about naming it cancelled only for the cancel action. The generic close signal should stay named "Closed". So could you please rename it back to closed and then *only where necessary* would a new "cancelled" signal be used to separate the usage where it is truly needed.

Kasterisk: Done.

hhyyrylainen: Nice, the commit looked fine with a quick glance. I'll fully re-check this PR again soon.

src/engine/DebugOverlays.tscn
```diff
@@ -209,12 +209,12 @@ __meta__ = {
 "_editor_description_": "PLACEHOLDER"
 }
 
-[connection signal="Closed" from="PerformanceMetrics" to="." method="OnPerformanceMetricsToggled"]
+[connection signal="Cancelled" from="PerformanceMetrics" to="." method="OnPerformanceMetricsToggled"]
```
hhyyrylainen, line 212: I looked at the code for PerformanceMetrics, and it really looks like to me like it only cares about if the popup becomes hidden, it doesn't care about anything else. So this is exactly my point that the most natural sounding signal name should be used in each case. So for this and probably a lot of other cases, the cancelled signal should be switched to be the Closed one (while still keeping the cancelled signal for cases that make most sense). Due to Easter I'm running out of time this week, so I won't be checking this until next week. Then I'll experiment and help finish this PR myself if I still see potential ways the system can be made more clear (after all this is a pretty core system that all parts of the GUI will use for years to come).

Kasterisk: Sure, that would be great. Wish you a good week!

src/gui_common/ModalManager.cs
```diff
@@ -0,0 +1,189 @@
...
+        popup.Notification(Control.NotificationModalClose);
+
+        if (popup is CustomDialog dialog)
+            dialog.EmitSignal(nameof(CustomDialog.Cancelled));
+
+        return true;
+    }
+
+    /// <summary>
+    ///   Returns the top-most exclusive popup in the current Viewport's modal stack. Null if there is none.
```
hhyyrylainen, line 103: This documentation is wrong, and I don't know how to fix it... this only checks if the topmost item in the stack is an exclusive popup, it doesn't look for any exclusive popups in the stack. The same problem applies to the `IsAnyExclusivePopupActive` property, as that name is really bad now. It doesn't check for *any* exclusive popups, just if there happens to be an exclusive popup on top...

src/engine/DebugOverlays.tscn
```diff
@@ -209,12 +209,12 @@ __meta__ = {
...
 [connection signal="toggled" from="DebugPanel/VBoxContainer/Checkboxes/FPSCheckBox" to="." method="OnFpsCheckBoxToggled"]
 [connection signal="toggled" from="DebugPanel/VBoxContainer/Checkboxes/PerformanceMetricsCheckBox" to="." method="OnPerformanceMetricsCheckBoxToggled"]
 [connection signal="toggled" from="DebugPanel/VBoxContainer/Checkboxes/CollisionShapeCheckBox" to="." method="OnCollisionShapeCheckBoxToggled"]
 [connection signal="toggled" from="DebugPanel/VBoxContainer/Checkboxes/EntityLabelCheckBox" to="." method="OnEntityLabelCheckBoxToggled"]
 [connection signal="toggled" from="DebugPanel/VBoxContainer/Checkboxes/InspectorCheckBox" to="." method="OnInspectorCheckBoxToggled"]
 [connection signal="pressed" from="DebugPanel/VBoxContainer/Button" to="." method="OnDumpSceneTreeButtonPressed"]
 [connection signal="value_changed" from="DebugPanel/VBoxContainer/VBoxContainer/RigiditySlider" to="." method="OnTransparencySliderValueChanged"]
-[connection signal="Closed" from="Inspector" to="." method="OnInspectorToggled"]
+[connection signal="Cancelled" from="Inspector" to="." method="OnInspectorToggled"]
```
hhyyrylainen, line 220: Actually testing this in-game this doesn't seem to work anymore... I'll go through all of the closed vs. cancelled signals in this PR and try to figure out which should be used. Hopefully I won't make mistakes, but I'll push my changes once done to give other people a chance to check my work.

src/gui_common/ModalManager.cs
Kasterisk: Well, depends on how you interpret it. The "active" part in the member names refer to the _top-most_ exclusive popup. The only thing that's in fact wrong in that documentation is that it still mentions Viewport, I guess I forgot to change that.

hhyyrylainen: All my English language skills tell me that, that's not the correct interpretation, "top-most exclusive popup" is the "top-most" popup that is exclusive. I.e. the logic or the documentation (and the variable name) has to change as they don't match. The code "returns the top most popup if it is exclusive", whereas the documentation "returns the top most popup that is exclusive". The slight difference in wording makes them mean completely different things.

Kasterisk: Sure, my suggestion for better names would be `GetTopMostPopupIfExclusive` for the method and `IsTopMostPopupExclusive` for the property, the documentations can be slightly adjusted accordingly.

hhyyrylainen: Those sound good to me. I assume there's a reason why only the top-most popup being exclusive matters.

Kasterisk: Previously, there was no need to know whether the topmost popup be otherwise. Apart from that, no other reason really.

hhyyrylainen: I suppose that is fine. I was thinking that there could be a bug / non-optimal situation where a non-exclusive popup gets opened on top of an exclusive one, or there might be a really tricky stack of popups like:
- Non-exclusive (top)
- Exclusive
- Non-exclusive
- Exclusive

The situation where that could trigger should be very rare, but it would be nice if the system would handle it in *some* logical way.

hhyyrylainen:
Does this fix: https://github.com/Revolutionary-Games/Thrive/issues/4201 ? I was reading through the changed and noticed the full rect code has been changed.

Also related to the art gallery I noticed a new bug: when playing a slideshow pressing ESC doesn't end it. Only the top right corner x closes the slideshow, but then it shows that ESC did close the underlying art gallery.

hhyyrylainen:
I did my changes, so after that one naming thing is solved and that art gallery bug is fixed, this is good to go in my opinion.

src/gui_common/ModalManager.cs
Kasterisk: If you mean that there would be a problem with hiding popups with ESC due to that then I think there shouldn't, inputs should always be consumed (unless you intentionally made it otherwise) by the topmost popup. Lower popups whether it be exclusive or non-exclusive, shouldn't be affected in any way.

The point is, exclusive popups and the above discussed class members does _not_ affect popup hides with ESC.

Kasterisk:
> Does this fix: https://github.com/Revolutionary-Games/Thrive/issues/4201 ? I was reading through the changes and noticed the full rect code has been changed.

Not sure, probably not.

hhyyrylainen:
As I was reading the diff, I got the impression that the full rect calculation is now *much* different. So even if accidental it could be fixed... I'll comment on that other issue to keep an eye on the situation.

Kasterisk:
Alright, slidescreen bug should be fixed now.

hhyyrylainen (approved):
The one bug I found is now gone, and my code related concerns have been addressed or I fixed them.

PR #4261, opened 2023-04-01
Chemoreceptor can detect cells

### LINKED ISSUE #3072: Chemoreceptor should have option to detect other cells
some kind of selection for specific species or like predator or prey (based on some criteria).

ottoblep:

This allows chemoreceptors to detect other cells.

**Related Issues**

This closes #3072.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


ottoblep:
Tracking of cells is based on species similar to compound type. The closest member of the selected species is shown.
Functionally this works almost the exact way the compound detection does.
The only structural difference is MicrobeSystem.FindSpeciesNearPoint performing the search instead of CompoundCloudSystem.FindCompoundNearPoint.
https://github.com/Revolutionary-Games/Thrive/blob/8189a71b65129b8faa23a881170308f08983f0dd/src/microbe_stage/MicrobeSystem.cs#L67
Since tracked microbes move, they need a faster update rate than compound lines.
That's why Compound and Species detection are currently completely separated and can be tweaked independently.
https://github.com/Revolutionary-Games/Thrive/blob/8189a71b65129b8faa23a881170308f08983f0dd/simulation_parameters/Constants.cs#L668-L669

TODO:
I did not find a way to access the current species list from the chemoreceptor update gui. 
The species selection from the gui is incomplete at the moment. 
I know little about the gui or localization system so far so i mostly left this out.
https://github.com/Revolutionary-Games/Thrive/blob/8189a71b65129b8faa23a881170308f08983f0dd/src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs#L54-L55
The player should be able to select between a compound or species to be tracked.
For testing purposes all species are currently added to detection for every cell.
https://github.com/Revolutionary-Games/Thrive/blob/8189a71b65129b8faa23a881170308f08983f0dd/src/microbe_stage/Microbe.cs#L984-L990

hhyyrylainen:
> Since tracked microbes move, they need a faster update rate than compound lines.

This should use the same approach as for other cell tracking, the tracking is updated each frame but new items are only detected a few times per second. This way the lines are always up to date but the expensive search doesn't run each frame.

>  I did not find a way to access the current species list from the chemoreceptor update gui.

The upgrade GUI should probably be modified to be given the CurrentGame when starting so that it can look at the relevant info in there. Why specifically the CurrentGame is that I think there's possibility of needing more info than just a species list so this general rework will be better for the future.

ottoblep:
A save upgrader / save incompatability point might be needed since chemoreceptor now has a different configuration.

hhyyrylainen:
If the data is carefully laid out so that the old data exists, but a new nullable field for species to detect is added, the data change should not require any save data upgrading.

ottoblep:
I tested the save compatability. Everything seems to transfer fine except for the line updates being broken until the next editor phase. Thats due to this one variable being renamed. I don't know if it's worth introducing a save upgrader for this, but can do if necessary.
https://github.com/Revolutionary-Games/Thrive/blob/05c128a8e664ae72a16f3523780bea640ea9300c/src/microbe_stage/MicrobeStage.cs#L622

hhyyrylainen:
Why was the rename necessary? If the rename is necessary, it is actually possible to specify an overridden name to be used in JSON, that way an updated variable name can still load the old data.

ottoblep:
The rename is necessary. It was called ```OnCompoundChemoreceptionInfo``` before and now it handles both kinds. 
I just saw the upgrader would need to update the value as well though, since the method that is referenced has changed.

hhyyrylainen:
Okay that makes sense. 
You can use that JSON attribute to set the JSON field name to be the same as it was before to avoid the save breakage.

hhyyrylainen:

hhyyrylainen:
Any progress on this? The next release is definitely going to be a planned save breakage point, so any PRs that make it into the next release won't need to take save compatibility into account at all.

ottoblep:
> Any progress on this? The next release is definitely going to be a planned save breakage point, so any PRs that make it into the next release won't need to take save compatibility into account at all.

The save upgrader is the only thing holding this up. So if the next release is a breakage point this is ready to merge. I will do some more testing and then check the boxes.

hhyyrylainen:
There was that master branch rewrite so this needs to be remade on a clean new branch, all the commits cherry-picked from the old branch and then pushed over this to make this mergeable.

ottoblep:
> There was that master branch rewrite so this needs to be remade on a clean new branch, all the commits cherry-picked from the old branch and then pushed over this to make this mergeable.

I can do that. This needs a new pull request then right?

hhyyrylainen:
You can just force push to the same branch to clear out the old stuff.

hhyyrylainen:
Actually I had a locally already fixed up branch of that. That just leaves the translation files conflicting. So if you first force pull you should have a *much* easier time to fix this branch with that as the base.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -1,12 +1,24 @@
-﻿using System.Collections.Generic;
-using System.Linq;
+﻿using System;
+using System.Collections.Generic;
 using Godot;
 
 public class ChemoreceptorUpgradeGUI : VBoxContainer, IOrganelleUpgrader
 {
+    [Export]
+    public NodePath? TargetTypesPath;
```
hhyyrylainen, line 8: Only the first NodePath should be nullable. That's the style used in Thrive. So the other NodePaths here as well as the Dispose method need to be slightly modified to match Thrive style.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -893,32 +899,64 @@ private void UpdateLinePlayerPosition()
...
-            foreach (var chemoreceptionLine in chemoreceptionLines)
-                chemoreceptionLine.Visible = false;
+            foreach (var chemoreception in chemoreceptionLines)
+                chemoreception.Line.Visible = false;
 
             return;
         }
 
-        var position = Player!.GlobalTransform.origin;
+        var position = Player!.Translation;
```
hhyyrylainen, line 908: This is an incorrect change.

src/microbe_stage/MicrobeSystem.cs
```diff
@@ -54,4 +55,63 @@ public void Process(float delta)
...
+    public (Microbe Microbe, Vector3 Position)? FindSpeciesNearPoint(Vector3 position, Species species,
+        float searchRadius = 200)
+    {
+        if (searchRadius < 1)
+            throw new ArgumentException("searchRadius must be >= 1");
+
+        (Microbe Microbe, Vector3 Position)? closestMicrobe = null;
+        float nearestDistanceSquared = float.MaxValue;
+        var searchRadiusSquared = searchRadius * searchRadius;
+        var microbes = worldRoot.GetTree().GetNodesInGroup(Constants.RUNNABLE_MICROBE_GROUP).Cast<Microbe>()
```
hhyyrylainen, line 75: This method should not rebuild the microbe species list each time this is called... instead the caller should specify it or maybe even the microbe system could cache an internal list of microbes in the Process method in this file?

src/microbe_stage/MicrobeStage.cs
```diff
@@ -893,32 +899,64 @@ private void UpdateLinePlayerPosition()
...
 
-        foreach (var chemoreceptionLine in chemoreceptionLines)
+        foreach (var chemoreception in chemoreceptionLines)
         {
-            if (chemoreceptionLine.Visible)
-                chemoreceptionLine.LineStart = position;
+            if (chemoreception.Line.Visible == false)
```
hhyyrylainen, line 912: Is Visible a nullable type? If not it should not be explicitly compared against a boolean, that's also not the Thrive style.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.tscn
```diff
@@ -10,21 +10,45 @@ anchor_bottom = 1.0
...
 MaximumDistancePath = NodePath("Distance")
 MinimumAmountPath = NodePath("Amount")
 ColourPath = NodePath("Colour")
 
 [node name="Label" type="Label" parent="."]
 margin_right = 1280.0
 margin_bottom = 25.0
+text = "TARGET_TYPE_COLON"
+
+[node name="TargetType" type="OptionButton" parent="."]
```
hhyyrylainen, line 27: As the options are static, they should be defined here in the scene instead of the .cs file. There's a special button in the Godot editor to edit the contents of an OptionButton.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -17,102 +29,156 @@ public class ChemoreceptorUpgradeGUI : VBoxContainer, IOrganelleUpgrader
...
+        foreach (var choice in shownCompoundChoices)
         {
             compounds.AddItem(choice.Name);
         }
 
-        // Select glucose by default
-        var defaultCompoundIndex =
-            shownChoices.FindIndex(c => c.InternalName == Constants.CHEMORECEPTOR_DEFAULT_COMPOUND_NAME);
+        shownSpeciesChoices = currentGame.GameWorld.Map.FindAllSpeciesWithPopulation();
+        shownSpeciesChoices.Sort((x, y) => string.Compare(x.Genus, y.Genus, StringComparison.OrdinalIgnoreCase));
```
hhyyrylainen, line 82: Why not compare against formatted name? That way the second part of the name would also help in making sure things are alphabetical.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -17,102 +29,156 @@ public class ChemoreceptorUpgradeGUI : VBoxContainer, IOrganelleUpgrader
...
 
-        if (defaultCompoundIndex < 0)
-            defaultCompoundIndex = 0;
+        foreach (var choice in shownSpeciesChoices)
+        {
+            species.AddItem(choice.Genus + " " + choice.Epithet);
```
hhyyrylainen, line 86: Manually building the species name display is not correct usage. As I said in my previous comment the species class already has a property for the properly formatted name.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -17,102 +29,156 @@ public class ChemoreceptorUpgradeGUI : VBoxContainer, IOrganelleUpgrader
...
+        }
+
+        var defaultCompound = SimulationParameters.Instance.GetCompound(Constants.CHEMORECEPTOR_DEFAULT_COMPOUND_NAME);
+        ChemoreceptorUpgrades defaultConfiguration = new ChemoreceptorUpgrades(defaultCompound, null,
```
hhyyrylainen, line 90: `var` should be preferred.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -17,102 +29,156 @@ public class ChemoreceptorUpgradeGUI : VBoxContainer, IOrganelleUpgrader
...
         if (compounds.Selected == -1)
             compounds.Selected = 0;
+        if (species.Selected == -1)
+            species.Selected = 0;
+
+        // Only one type of object can be detected
+        Compound? choiceCompound = null;
+        Species? choiceSpecies = null;
+
+        if (targetTypes.Selected == 0)
```
hhyyrylainen, line 124: Seeing as 0 and 1 are used as magic numbers in multiple places, I think you should specify an enum like in other places where different tabs are used in the GUI to add some bulletproofing to the code.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -121,4 +187,39 @@ protected override void Dispose(bool disposing)
...
+        {
+            TypeChanged(0);
+            targetTypes.Selected = 0;
+            compounds.Selected = shownCompoundChoices.FindIndex(c => c == configuration.TargetCompound);
+        }
+        else if (configuration.TargetSpecies != null)
+        {
+            TypeChanged(1);
+            targetTypes.Selected = 1;
+            species.Selected = shownSpeciesChoices.FindIndex(c => c == configuration.TargetSpecies);
```
hhyyrylainen, line 207: Does this make a sensible selection when the species is extinct and is no longer found? My hunch is that it does, but I just want to ask to confirm.

src/microbe_stage/Microbe.cs
```diff
@@ -938,6 +947,51 @@ public List<(Compound Compound, Color Colour, Vector3 Target)> GetDetectedCompou
...
+        {
+            collectedUniqueSpeciesDetections = activeSpeciesDetections;
+        }
+
+        var detections = new List<(Microbe Microbe, Color Colour, Vector3 Target)>();
+        var position = GlobalTranslation;
+
+        foreach (var (species, range, colour) in collectedUniqueSpeciesDetections)
+        {
+            var tuple = microbeSystem.FindSpeciesNearPoint(position, species, range);
```
hhyyrylainen, line 984: Oh and this is called here in a recursive loop. This is absolutely going to wreck performance if you have a colony of 50 cells each with multiple chemoreceptor and hundreds of cells in the game, the full list of microbes in the world is going to be rebuilt so many times in that situation.

hhyyrylainen (changes requested):
I ended up doing a code review piecemeal so see my previous comments. I didn't playtest this yet as I identified a few potential performance problems and some general code style problems.

src/microbe_stage/MicrobeStage.cs
ottoblep: Visible is not nullable here. Which style would you use? I couldn't find a definitive answer in the existing code or style guide.

hhyyrylainen: `if (!chemoreception.Line.Visible)`

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
ottoblep: Yes, since the configuration saves the species/compound itself this should be robust against changes in the world. On entering the editor if the species is no longer found it sets the `OptionButton` to unselected. If it stays unselected it is set to the first species in the list on `ApplyChanges`. If the editor stage is skipped it is possible that an extinct species stays in the configuration. Then the component simply does nothing until edited again.

hhyyrylainen: That sounds about what I expected would happen, which is a fine approach.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -893,32 +899,64 @@ private void UpdateLinePlayerPosition()
...
+            if (!chemoreception.Line.Visible)
+                continue;
+
+            chemoreception.Line.LineStart = position;
+
+            // The target needs to be updated for detected microbes but not compounds.
+            if (chemoreception.TargetMicrobe != null)
+            {
+                // Use colony parent position to avoid calling GlobalTranslation
+                if (chemoreception.TargetMicrobe.ColonyParent != null)
```
hhyyrylainen, line 921: This is not accurate. ColonyParent technically isn't required t obe the colony lead cell. As you actually want the colony lead cell you need to go through the MicrobeColony object and grab the master from there.

ottoblep:
Is there a standard way to merge these translation files? It seems they can get easily messed up when the line numbers change.

hhyyrylainen:
I just accept the other side for all files except en.po which I merge manually. After that I rerun the update script and *then* commit the merge.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -1,124 +1,241 @@
...
+
+        TypeChanged((int)TargetType.Compound);
     }
 
-    public void OnStartFor(OrganelleTemplate organelle)
+    public void OnStartFor(OrganelleTemplate organelle, GameProperties currentGame)
     {
-        shownChoices = SimulationParameters.Instance.GetCloudCompounds();
+        shownCompoundChoices = SimulationParameters.Instance.GetCloudCompounds();
+        shownCompoundChoices.Sort((x, y) => string.Compare(x.Name, y.Name, StringComparison.OrdinalIgnoreCase));
```
hhyyrylainen, line 81: Isn't there a "cleaner" sort overload something like `shownCompoundChoices.Sort(i => i.Name, StringComparer.OrdinalIgnoreCase)`? That avoids any potential bugs in manually performing the compare instead of just telling the standard sort algorithm to use a specific property.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -1,124 +1,241 @@
...
+        foreach (var choice in shownCompoundChoices)
         {
             compounds.AddItem(choice.Name);
         }
 
-        // Select glucose by default
-        var defaultCompoundIndex =
-            shownChoices.FindIndex(c => c.InternalName == Constants.CHEMORECEPTOR_DEFAULT_COMPOUND_NAME);
+        shownSpeciesChoices = currentGame.GameWorld.Map.FindAllSpeciesWithPopulation();
+        shownSpeciesChoices.Sort((x, y) => string.Compare(x.FormattedName, y.FormattedName,
```
hhyyrylainen, line 89: And same here.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -1,124 +1,241 @@
...
 
-        if (defaultCompoundIndex < 0)
-            defaultCompoundIndex = 0;
+        foreach (var choice in shownSpeciesChoices)
+        {
+            species.AddItem(choice.FormattedName);
+        }
+
+        var defaultCompound = SimulationParameters.Instance.GetCompound(Constants.CHEMORECEPTOR_DEFAULT_COMPOUND_NAME);
+        var defaultConfiguration = new ChemoreceptorUpgrades(defaultCompound, null,
```
hhyyrylainen, line 98: This object should only be created when necessary, i.e. in the else branch.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -1,124 +1,241 @@
...
+        // Force some type/compound/species to be selected
+        if (targetTypes.Selected == -1)
+            targetTypes.Selected = (int)TargetType.Compound;
         if (compounds.Selected == -1)
             compounds.Selected = 0;
+        if (species.Selected == -1)
+            species.Selected = 0;
+
+        // Only one type of object can be detected
+        Compound? choiceCompound = null;
```
hhyyrylainen, line 129:
```suggestion
        Compound? compoundChoice = null;
```
The word order sounds unnatural here, so here's a suggestion to improve it. Also for the next line the same thing can be done.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -1,124 +1,241 @@
...
-using System.Linq;
+﻿using System;
+using System.Collections.Generic;
 using Godot;
 
 public class ChemoreceptorUpgradeGUI : VBoxContainer, IOrganelleUpgrader
 {
     [Export]
-    public NodePath? CompoundsPath;
+    public NodePath? TargetTypesPath;
```
hhyyrylainen, line 8: I'd like these added controls to have slightly more expressive names. Right now this is just "target types", which I'd expect to be a list of target types, when in fact this is the target type selection. So I suggest the name `TargetTypeSelectorPath` and renaming the related variables.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -1,124 +1,241 @@
...
+
+    [Export]
+    public NodePath CompoundsPath = null!;
+
+    [Export]
+    public NodePath CompoundsLabelPath = null!;
+
+    [Export]
+    public NodePath SpeciesPath = null!;
```
hhyyrylainen, line 17:
```suggestion
    public NodePath SpeciesSelectorPath = null!;
```

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -1,124 +1,241 @@
...
 using Godot;
 
 public class ChemoreceptorUpgradeGUI : VBoxContainer, IOrganelleUpgrader
 {
     [Export]
-    public NodePath? CompoundsPath;
+    public NodePath? TargetTypesPath;
+
+    [Export]
+    public NodePath CompoundsPath = null!;
```
hhyyrylainen, line 11:
```suggestion
    public NodePath CompoundSelectorPath = null!;
```

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -1,124 +1,241 @@
...
+
     [Export]
     public NodePath ColourPath = null!;
 
 #pragma warning disable CA2213
+    private OptionButton targetTypes = null!;
     private OptionButton compounds = null!;
+    private Label compoundLabel = null!;
+    private OptionButton species = null!;
+    private Label speciesLabel = null!;
```
hhyyrylainen, line 39: And these are the related variables that should also be renamed if my previous rename comments are accepted.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -1,124 +1,241 @@
...
+    ///   Applies the color of a selected target to the color picker
+    /// </summary>
+    private void ApplySelectionColour()
+    {
+        if (targetTypes.Selected == (int)TargetType.Compound && shownCompoundChoices != null
+            && compounds.Selected >= 0)
+            colour.Color = shownCompoundChoices[compounds.Selected].Colour;
+        if (targetTypes.Selected == (int)TargetType.Species && shownSpeciesChoices != null
+            && species.Selected >= 0)
+            colour.Color = shownSpeciesChoices[species.Selected].Colour;
```
hhyyrylainen, line 239:
```suggestion
        if (targetTypes.Selected == (int)TargetType.Compound && shownCompoundChoices != null
            && compounds.Selected >= 0)
        {
            colour.Color = shownCompoundChoices[compounds.Selected].Colour;
        }

        if (targetTypes.Selected == (int)TargetType.Species && shownSpeciesChoices != null
            && species.Selected >= 0)
        {
            colour.Color = shownSpeciesChoices[species.Selected].Colour;
        }
```

Multiline conditions should use braces for readability purposes.

src/microbe_stage/MicrobeSystem.cs
```diff
@@ -54,4 +57,63 @@ public void Process(float delta)
...
+            if (microbe.ColonyParent != null)
+            {
+                microbeGlobalPosition = microbe.ColonyParent.Translation;
+            }
+            else
+            {
+                microbeGlobalPosition = microbe.Translation;
+            }
+
+            // Skip candidates for performance
```
hhyyrylainen, line 97: Does this *actually* perform better? This is multiple branching operations whereas the distance check should be *just* about 20 float operations and a single comparison.

src/microbe_stage/MicrobeSystem.cs
```diff
@@ -54,4 +57,63 @@ public void Process(float delta)
...
+                continue;
+            }
+
+            var distanceSquared = (microbeGlobalPosition - position).LengthSquared();
+
+            // Check search range and prevent cell detecting itself
+            if (distanceSquared > searchRadiusSquared || distanceSquared <= 1)
+                continue;
+
+            if (distanceSquared < nearestDistanceSquared)
```
hhyyrylainen, line 110: These could be maybe swapped around to simplify the logic here. It probably won't perform worse as the microbes are very likely not in a descending order of distance to the target position. And it would entirely remove the need to check if the search radius is followed here. That can be added just at the end to return null instead of the closest microbe if it is farther than the search radius.

src/microbe_stage/MicrobeSystem.cs
```diff
@@ -54,4 +57,63 @@ public void Process(float delta)
...
+        float searchRadius = 200)
+    {
+        if (searchRadius < 1)
+            throw new ArgumentException("searchRadius must be >= 1");
+
+        (Microbe Microbe, Vector3 Position)? closestMicrobe = null;
+        float nearestDistanceSquared = float.MaxValue;
+        var searchRadiusSquared = searchRadius * searchRadius;
+        microbes ??= worldRoot.GetTree().GetNodesInGroup(Constants.RUNNABLE_MICROBE_GROUP).Cast<Microbe>()
+            .ToArray();
```
hhyyrylainen, line 78:
```suggestion
        var searchRadiusSquared = searchRadius * searchRadius;

        microbes ??= worldRoot.GetTree().GetNodesInGroup(Constants.RUNNABLE_MICROBE_GROUP).Cast<Microbe>()
            .ToArray();
```

Just a bit extra blank space to separate logically separate variables, which looks better in my opinion. You don't have to accept this suggestion.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
ottoblep: An alternative would be `shownCompoundChoices = shownCompoundChoices.OrderBy(x=>x.Name).ToList();`, but this is slower since it allocates a new list. With a quick search i couldn't find a way to shorten the `Comparison<T>`.

hhyyrylainen: Hang on, there's a bug here *anyway* as you are modifying the list that is owned by SimulationParameters. You should not do that. So the actual solution here is:

```
        shownCompoundChoices = SimulationParameters.Instance.GetCloudCompounds()
            .OrderBy(c => c.Name, StringComparer.OrdinalIgnoreCase).ToList();
```
(I had forgotten that the nice syntax is only supported by the LINQ sort and not the inbuilt list sort)

I'll make a separate change to make the simulation parameters return a unmodifiable list here.

src/microbe_stage/MicrobeSystem.cs
ottoblep: Ignoring data instructions and if multiplication takes 2 ops, this roughly saves 7 ops per miss and adds 6 ops per hit. Since the vast majority of calls will be misses i think it is faster. However the difference might be negligible.

hhyyrylainen: Isn't any branching operation slower than even division? That's my point that optimizing a bit of math is always slower on a modern CPU as branching is so expensive operation (hence why branchless programming is such a popular performance topic).

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
hhyyrylainen: As [discovered](https://github.com/Revolutionary-Games/Thrive/pull/4261/files#r1313798649), there's probably no better comparison method syntax available here.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
hhyyrylainen: The above variables were renamed, but these weren't yet to match the new names of the related variables.

src/microbe_stage/MicrobeSystem.cs
ottoblep: The branch prediction will likely prefetch the `continue` path since it is the most likely. Then we need to add the cost of flushing the pipline to the added cost on a hit (lets say 10 cycles). So the change in cycles when using the branching is\
`(1-hitrate) * -7  + hitrate * 16` which is negative for a hitrate smaller than 30%. Though some profiling is probably needed to get a real comparison.

ottoblep: tldr I think the branch is pretty deterministic and thus will have only a small performance impact.

hhyyrylainen: Well I'd be really interested to show what a benchmark done using https://github.com/dotnet/BenchmarkDotNet (I've seen this used by serious C# performance related Stackoverflow answers use) would show in this case.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
hhyyrylainen: This change is now fine. And for reference here's the other PR I created: https://github.com/Revolutionary-Games/Thrive/pull/4454

src/microbe_stage/MicrobeSystem.cs
ottoblep: So i built the [test bench](https://github.com/ottoblep/thrive-profiling/blob/master/Program.cs) (BenchmarkDotNet works really well) and got these results. If the search radius is large compared to the spawn radius and if there are many of the searched species then the NoBranch option is better otherwise the reverse is true. In the average in-game case i got from testing in the editor (FRAC=0.05, RADIUS=150) the Branch option is slightly better.
```
|                       Method | SPECIES_FRACTION | SEARCH_RADIUS  |     Mean |    Error |   StdDev |   Median |
|----------------------------- |----------------- |--------------- |---------:|---------:|---------:|---------:|
|   FindSpeciesNearPointBranch |             0.05 |            50  | 617.9 ns |  1.49 ns |  1.32 ns | 618.1 ns |
| FindSpeciesNearPointNoBranch |             0.05 |            50  | 663.6 ns | 19.81 ns | 54.24 ns | 645.8 ns |
|   FindSpeciesNearPointBranch |             0.05 |           150  | 638.7 ns | 12.24 ns | 14.09 ns | 634.6 ns |
| FindSpeciesNearPointNoBranch |             0.05 |           150  | 647.1 ns | 12.72 ns | 22.62 ns | 637.6 ns |
|   FindSpeciesNearPointBranch |             0.05 |           350  | 642.0 ns | 12.64 ns | 17.72 ns | 635.3 ns |
| FindSpeciesNearPointNoBranch |             0.05 |           350  | 647.6 ns | 12.91 ns | 21.57 ns | 645.9 ns |
|   FindSpeciesNearPointBranch |              0.2 |            50  | 629.8 ns | 10.44 ns | 12.42 ns | 624.5 ns |
| FindSpeciesNearPointNoBranch |              0.2 |            50  | 615.8 ns |  2.12 ns |  1.77 ns | 615.9 ns |
|   FindSpeciesNearPointBranch |              0.2 |           150  | 622.5 ns |  3.96 ns |  3.51 ns | 621.5 ns |
| FindSpeciesNearPointNoBranch |              0.2 |           150  | 618.0 ns |  4.67 ns |  4.37 ns | 616.6 ns |
|   FindSpeciesNearPointBranch |              0.2 |           350  | 620.2 ns |  1.60 ns |  1.33 ns | 620.2 ns |
| FindSpeciesNearPointNoBranch |              0.2 |           350  | 615.3 ns |  2.54 ns |  1.98 ns | 615.4 ns |
|   FindSpeciesNearPointBranch |              0.5 |            50  | 623.7 ns |  4.17 ns |  3.70 ns | 623.1 ns |
| FindSpeciesNearPointNoBranch |              0.5 |            50  | 614.4 ns |  0.80 ns |  0.67 ns | 614.2 ns |
|   FindSpeciesNearPointBranch |              0.5 |           150  | 621.9 ns |  3.68 ns |  3.26 ns | 621.0 ns |
| FindSpeciesNearPointNoBranch |              0.5 |           150  | 616.1 ns |  2.44 ns |  2.28 ns | 615.1 ns |
|   FindSpeciesNearPointBranch |              0.5 |           350  | 623.1 ns |  4.74 ns |  4.21 ns | 621.9 ns |
| FindSpeciesNearPointNoBranch |              0.5 |           350  | 615.7 ns |  1.31 ns |  1.09 ns | 615.1 ns |
```

hhyyrylainen: Thanks for getting the actual numbers. 

I suspect that most of the time the search radius will be quite high, and as there are planned to be roughly 10 species per patch (this is a tweak value in auto-evo), the conclusion is that the branch version is slightly faster usually but sometimes a bit slower. So I suppose this is actually a very slight optimization which means this optimization can be kept.

hhyyrylainen (approved):
If I read my list of comments right, everything is now in resolved status. So this is ready to merge now. Thanks for sticking with this PR.

Due to my refactoring branch having a bunch of changes I'll likely manually need to merge this there so I'm fine with merging this to master without testing, but later I'd like to ask you to review my branch when that is ready to confirm that I've correctly managed to resolve the merge conflicts.

hhyyrylainen:
Oh right, I just remembered before submitting the review that you'll need to resolve the translation conflicts before I can merge. The usual easy way it to accept the other side changes for everything except en.po and then running the localization update after merging to get everything up to date.

Issue #3676, opened 2022-08-09
Spawner.DespawnAll doesn't work for early multicellular grown colonies


hhyyrylainen:
it probably just despawns the colony leader and all the other cells spill out of it. Also got a report that using the despawn all cheat twice in this case results in a crash.

I think there's already a TODO in the code about how to handle the primary cell in a multicellular creature being destroyed.

hhyyrylainen:
Also when resetting the growth (for example player becoming multicellular) that needs to also properly despawn the existing colony members without detaching them.

hhyyrylainen:
This now works after the ECS refactor.

PR #4273, opened 2023-04-07
Improved divide separation for cells and colonies

### LINKED ISSUE #1435: Improve the separation of big cells on divide (when reproducing after exiting the editor)
to make it less likely that the cells might get stuck.

### LINKED ISSUE #3653: Fix multicellular colony divide offset for new cell being too low in some cases
Sometimes the offset to spawn the extra cell at doesn't work correctly and it ends up overlapping a lot on the colony it spawns inside of.
This is the code:
https://github.com/Revolutionary-Games/Thrive/blob/67ab283bf9ebb617b83da86b78e15e1eda776376/src/microbe_stage/Microbe.Interior.cs#L374

Video:
https://cdn.discordapp.com/attachments/465937305849298956/1004672047345041428/1.mp4

Leinourdian:

This PR makes divide separation more accurate for cells and colonies. It works regardless of if the divide happens during gameplay or when returning from editor. This makes it so cells never spawn inside each other when dividing, and they also don't spawn detached. Colonies never spawn cells inside each other when dividing either, but they can sometimes be a little bit detached.

**Related Issues**

Fixes #1435 by making divide more accurate and having it work when returning from editor.

Fixes #3653 by moving the daughter colony after it has split and grown.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Kasterisk (changes requested):

src/general/utils/MathUtils.cs
```diff
@@ -131,4 +131,31 @@ public static int PositiveModulo(this int val, int mod)
         double standardDeviation = Math.Sqrt(sumOfSquares / count - average * average);
         return (average, standardDeviation);
     }
+
+    public static float GetLengthInDirectionFromPoints(Vector3 direction, Vector3 referencePoint,
+        List<Vector3> listOfPoints)
```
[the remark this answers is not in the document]
Leinourdian: Does my new commit clear this up?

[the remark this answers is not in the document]
Kasterisk:
``` csharp
    ///   The maximum distance that a set of points can reach in a given direction
```
It's better, but one thing, the role of `referencePoint` seems to be left out. Explaining how it relate to the set of points would be nice.

Kasterisk (approved):
I tested with chonky cells, stick cells, all works very well. Thanks!

Kasterisk:
Final review will be left for hhyyrylainen once he's back from his trip.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -389,38 +389,34 @@ public Microbe Divide()
...
 
-                // TODO: switch this to something else if this is too slow for large colonies
-                var positionInColony = colonyMember.GlobalTransform.origin - currentPosition;
+        if (Colony != null)
+        {
+            var colonyMembers = Colony.ColonyMembers.Select(c => c.GlobalTransform.origin).ToList();
 
-                var outerRadius = Math.Max(Math.Abs(positionInColony.x) + radius,
-                    Math.Abs(positionInColony.z) + radius);
+            distanceRight += MathUtils.GetMaximumDistanceInDirection(direction, currentPosition, colonyMembers);
```
hhyyrylainen, line 405: A slight improvement could be made here, GetMaximumDistanceInDirection could be modified to take in an IEnumerable instead of a list, that way `colonyMembers` won't need to be created into an actual list a few lines up from here. With that change GetMaximumDistanceInDirection can be called in some situations while saving the memory and GC pressure needed to create a temporary list.

src/general/utils/MathUtils.cs
```diff
@@ -131,4 +132,39 @@ public static int PositiveModulo(this int val, int mod)
...
+    /// <summary>
+    ///   How far in a given direction a set of points can reach from a reference point
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     This can be useful for finding the edge of a cell from a list of its organelle positions, for instance
+    ///   </para>
+    /// </remarks>
+    public static float GetMaximumDistanceInDirection(Vector3 direction, Vector3 referencePoint,
+        List<Vector3> listOfPoints)
```
hhyyrylainen, line 145:
```suggestion
        IEnumerable<Vector3> listOfPoints)
```
So here this would be modified like this.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -389,38 +389,34 @@ public Microbe Divide()
...
+        if (CellTypeProperties.IsBacteria)
+            width *= 0.5f;
 
         // Create the one daughter cell.
-        var copyEntity = SpawnHelpers.SpawnMicrobe(Species, currentPosition + separation,
+        var copyEntity = SpawnHelpers.SpawnMicrobe(Species, currentPosition + direction * width,
             GetParent(), SpawnHelpers.LoadMicrobeScene(), true, cloudSystem!, spawnSystem!, CurrentGame);
 
+        // Since the daughter spawns right next to the cell, it should face the same way to avoid colliding
+        copyEntity.Rotation = Rotation;
```
hhyyrylainen, line 418: Using Rotation destroys the quaternion by flattening it into euler angles... This should instead modify the transform, something like this (I didn't test so this might be slightly off):

```
copyEntity.Transform = new Transform(Transform.basis, copyEntity.Translation);
```

hhyyrylainen: Note that that assumes that the scale etc. is fine in the parent cell to copy. Instead it might be better to convert the `Transform.basis.Quat()` and build a new basis from that that copies the scale from the copyEntity basis.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -518,17 +526,19 @@ public override void OnReturnFromEditor()
...
-        }
+            if (daughter.Colony != null)
+            {
+                // Add more extra offset between the player and the divided cell
+                var daughterPosition = daughter.GlobalTransform.origin;
+                var direction = (playerPosition - daughterPosition).Normalized();
 
-        // Update the player's cell
-        Player.ApplySpecies(Player.Species);
+                var colonyMembers = daughter.Colony.ColonyMembers.Select(c => c.GlobalTransform.origin).ToList();
```
hhyyrylainen, line 535: The ToList call can be saved here as well if the change to GetMaximumDistanceInDirection is done.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -507,9 +507,17 @@ public override void OnReturnFromEditor()
...
+        // Update the player's cell
+        Player!.ApplySpecies(Player.Species);
+
+        // Reset all the duplicates organelles of the player
+        Player.ResetOrganelleLayout();
+
+        var playerPosition = Player.GlobalTransform.origin;
+
         // Spawn another cell from the player species
-        // This is done first to ensure that the player colony is still intact for spawn separation calculation
```
hhyyrylainen, line 511: This isn't a request to change anything, but just so that I understand what's going on, does this mean now that the player colony is already reset before dividing now? Or does `ResetOrganelleLayout` not reset the colony? Hopefully a bug where the player colony is never reset is not present now with these changes.

hhyyrylainen: It's not really intended. It's a known problem, here's the issue I opened about despawning not working: https://github.com/Revolutionary-Games/Thrive/issues/3676 I'll comment there that resetting the organelle layout should also be made to properly despawn the colony members.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -389,38 +389,36 @@ public Microbe Divide()
...
+        if (CellTypeProperties.IsBacteria)
+            width *= 0.5f;
 
         // Create the one daughter cell.
-        var copyEntity = SpawnHelpers.SpawnMicrobe(Species, currentPosition + separation,
+        var copyEntity = SpawnHelpers.SpawnMicrobe(Species, currentPosition + direction * width,
             GetParent(), SpawnHelpers.LoadMicrobeScene(), true, cloudSystem!, spawnSystem!, CurrentGame);
 
+        // Since the daughter spawns right next to the cell, it should face the same way to avoid colliding
+        var daughterBasis = new Basis(Transform.basis.Quat().Normalized()).Scaled(copyEntity.Transform.basis.Scale);
```
hhyyrylainen, line 418:
```suggestion
        var daughterBasis = new Basis(Transform.basis.Quat())
        {
            Scale = copyEntity.Transform.basis.Scale,
        };
```

I think that would be the optimal way. I didn't originally suggest copying the full transform from the parent, *just in case* the scales are different (which probably should only happen due to a bug but I think it's good to handle that sensibly anyway).

PR #4274, opened 2023-04-07
Don't create collision shapes for preview-only cells

Kasterisk:

Found this out while testing #4273.

This helps reducing lag in early multicellular editor (or perhaps microbe editor) if the cell the player currently editing is ungodly massive.

**Related Issues**

_None, I think._

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/PlacedOrganelle.cs
+++ b/src/microbe_stage/PlacedOrganelle.cs
@@ -575,7 +575,9 @@ public class PlacedOrganelle : Spatial, IPositionedOrganelle, ISaveLoadedTracked
         //     !IsLoadedFromSave)
         //     ParentMicrobe.Colony.Master.Mass += Definition.Mass;
 
-        MakeCollisionShapes(ParentMicrobe!.Colony?.Master ?? ParentMicrobe);
+        // We don't need preview cells to be collidable (as it can lag the editor if the cell is massive).
+        if (!ParentMicrobe.IsForPreviewOnly)
+            MakeCollisionShapes(ParentMicrobe!.Colony?.Master ?? ParentMicrobe);
 
         if (Definition.Enzymes != null)
         {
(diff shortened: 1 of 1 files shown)


Leinourdian:
I tried this out and wow, this makes a huge difference for big multicellular organisms! My general FPS in the editor jumped from 5 to about 50 and adding cells was visibly faster. I didn't observe an FPS change for unicellular organisms but I think updating to preview mode got slightly faster there too.

Kasterisk:
Just goes to show how majority of the lag in later stages were probably caused by unoptimized physics.

src/microbe_stage/PlacedOrganelle.cs
```diff
@@ -575,7 +575,9 @@ private void FinishAttachToMicrobe()
         //     !IsLoadedFromSave)
         //     ParentMicrobe.Colony.Master.Mass += Definition.Mass;
 
-        MakeCollisionShapes(ParentMicrobe!.Colony?.Master ?? ParentMicrobe);
+        // We don't need preview cells to be collidable (as it can lag the editor if the cell is massive).
+        if (!ParentMicrobe.IsForPreviewOnly)
+            MakeCollisionShapes(ParentMicrobe!.Colony?.Master ?? ParentMicrobe);
```
hhyyrylainen, line 580: I think the null suppression on ParentMicrobe is really unnecessary on this line now as the if will have crashed anyway before reaching this line.

hhyyrylainen (changes requested):
Seems like a good idea. I think a bunch of the components etc. setup could also be avoided in preview mode, but I won't ask to roll those into the same PR as if this already massively improves performance that's good enough.

src/microbe_stage/PlacedOrganelle.cs
Kasterisk: Sure, copied this without thought.

hhyyrylainen (approved):

PR #4276, opened 2023-04-10
Improve light level button in the editor

### LINKED ISSUE #3881: Add button to use current light level in the editor and make it the default
add a button for "current" light level (the light level from the stage when the player switched to the editor) as the default view in the editor, and make the patch map use the current editor light level (needs to be applied to all patches for relative comparisons to work) in comparisons so that the patch map and editor visuals are consistent with each other.

### LINKED ISSUE #3919: Editor should remember the selected light level when saving and loading
right now it resets when loading a save, it should be saved like a bunch of editor GUI state is

Kasterisk:

Add "Current" button which syncs the light level in all patch in the patch map to the true daylight level and this is by default.

The editor should also now correctly loads the selected light level button in a save.

Loading old save works, nothing seems broken but more testing pleasee.

**Related Issues**

Fixes #3881.
Fixes #3919.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -1068,7 +1068,7 @@ public partial class CellEditorComponent :
 
     public override void OnLightLevelChanged(float lightLevel)
     {
-        var maxLightLevel = Editor.CurrentPatch.Biome.Compounds[sunlight].Ambient;
+        var maxLightLevel = Editor.CurrentPatch.Biome.MaximumCompounds[sunlight].Ambient;
         var minLightLevel = Editor.CurrentPatch.Biome.MinimumCompounds[sunlight].Ambient;
         var templateMaxLightLevel = Editor.CurrentPatch.Biome.Compounds[sunlight].Ambient;
 
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -1088,7 +1088,7 @@ public partial class CellEditorComponent :
 
                 // Figure out the daylight in all patches relative to the player's current patch
                 var relativeLightLevel = (lightLevel - minLightLevel) / (maxLightLevel - minLightLevel) *
-                        (targetMaxLightLevel - targetMinLightLevel) + targetMinLightLevel;
+                    (targetMaxLightLevel - targetMinLightLevel) + targetMinLightLevel;
 
                 var lightLevelAmount = new BiomeCompoundProperties { Ambient = relativeLightLevel };
 
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -1070,7 +1070,7 @@ public partial class CellEditorComponent :
     {
         var maxLightLevel = Editor.CurrentPatch.Biome.MaximumCompounds[sunlight].Ambient;
         var minLightLevel = Editor.CurrentPatch.Biome.MinimumCompounds[sunlight].Ambient;
-        var templateMaxLightLevel = Editor.CurrentPatch.Biome.Compounds[sunlight].Ambient;
+        var templateMaxLightLevel = Editor.CurrentPatch.GetCompoundAmount(sunlight, CompoundAmountType.Template);
 
         // Currently, patches whose templates have zero sunlight can be given non-zero sunlight as an instance. But
         // nighttime shaders haven't been created for these patches (specifically the sea floor) so for now we can't
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -30,6 +30,9 @@ public partial class CellEditorComponent :
     public NodePath AverageLightButtonPath = null!;
 
     [Export]
+    public NodePath CurrentLightButtonPath = null!;
+
+    [Export]
     public NodePath TabButtonsPath = null!;
 
     [Export]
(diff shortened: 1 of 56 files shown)


src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -314,6 +314,12 @@ protected override void SetupEditedSpecies()
         base.SetupEditedSpecies();
     }
 
+    protected override void ApplyComponentLightLevels()
+    {
+        base.ApplyComponentLightLevels();
+        patchMapTab.UpdateShownPatchDetails();
```
hhyyrylainen, line 320: Patch map is an editor component? Why doesn't it get given the info through the editor component light level update callback?

Kasterisk: That didn't cross my mind...

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1072,8 +1068,9 @@ public float CalculateTotalDigestionSpeed()
 
     public override void OnLightLevelChanged(float lightLevel)
     {
-        var maxLightLevel = Editor.CurrentPatch.GetCompoundAmount("sunlight", CompoundAmountType.Maximum);
-        var templateMaxLightLevel = Editor.CurrentPatch.GetCompoundAmount("sunlight", CompoundAmountType.Template);
+        var maxLightLevel = Editor.CurrentPatch.Biome.Compounds[sunlight].Ambient;
+        var minLightLevel = Editor.CurrentPatch.Biome.MinimumCompounds[sunlight].Ambient;
+        var templateMaxLightLevel = Editor.CurrentPatch.Biome.Compounds[sunlight].Ambient;
```
hhyyrylainen, line 1073: Either max or the template light level variable is wrong here. Or at least that has changed. Is this change intentional?

Kasterisk: Yes, the previous method returns percentage, which is not suitable for what we want to do later in the code.

hhyyrylainen: Well the variable name has to change then as it is not related to the template. I guess logically the change is fine as well as the player will be more interested in seeing what the light level can be in the patch, not what the archetype biome can have as the light level.

Kasterisk: Realized what you're saying was `maxLightLevel` is using `Compounds` instead of `MaximumCompounds`... let me rectify my dumb mistake.

hhyyrylainen: Also fix the fact that the "template" max light level doesn't actually have the template light level... The template light level if still actually needed can be retrieved from the Patch but not the biome conditions.

Kasterisk: Alright, figured it out, everything should be good.

hhyyrylainen (changes requested):
The code now looks correct. When playtesting I noticed one bug: when I have a patch selected, go to the cell editor tab, switch the time of day, and finally go back to the patch map tab I see the previous light level values in the right panel. The situation corrects itself when I click on a different patch and then click on the original patch again.

So there's some callback thing missing for updating the currently selected patch data when the light level is changed. That was the only problem I saw, everything else seemed to make sense.

Kasterisk:
Apparently the order of this matter now:
```csharp
    protected override IEnumerable<IEditorComponent> GetAllEditorComponents()
    {
        yield return reportTab;
        yield return patchMapTab;
        yield return cellEditorTab;
    }
```

The patch map tab must follow the cell editor tab so that it gets the correct light level. Should we reorder it or entirely move the patch light level updation out of the cell editor tab?

hhyyrylainen:
Hang on, could you explain the path the update logic takes? Why does the order matter? When the light level change is triggered, that should be sent to all of the components. So it shouldn't matter what order the components are specified in...

Kasterisk:
```csharp
    private void ApplyComponentLightLevels()
    {
        foreach (var editorComponent in GetAllEditorComponents())
        {
            editorComponent.OnLightLevelChanged(lightLevel);
        }
    }
```

Here, `OnLightLevelChanged` is first called in `MicrobeEditorReportComponent`, then in `PatchMapEditorComponent` and finally in `CellEditorComponent`.

`CellEditorComponent` is where the actual light level in all patches is updated, `PatchMapEditorComponent` just updates its details with the new light level. From this, we can deduce that the bug was caused by the light level change callback being triggered in the patch map first instead of the cell editor.

Although reordering `GetAllEditorComponents` might cause a whole lot of other unforeseen problems...

hhyyrylainen:
> CellEditorComponent is where the actual light level in all patches is updated, PatchMapEditorComponent just updates its details with the new light level. From this, we can deduce that the bug was caused by the light level change callback being triggered in the patch map first instead of the cell editor.

I think it's anyway pretty unclean that the cell editor component would update the patch map view's data. Much better for the map to handle that. That's needed anyway for multicellular editor as eventually the player should be able to change the light level in the body plan editor so it is not useful for that map update logic to reside in the cell editor component.

> Although reordering GetAllEditorComponents might cause a whole lot of other unforeseen problems...

Absolutely. My intention is that the order of that list actually matches the tabs in the GUI and logical flow of the user

Kasterisk:
> I think it's anyway pretty unclean that the cell editor component would update the patch map view's data. Much better for the map to handle that. That's needed anyway for multicellular editor as eventually the player should be able to change the light level in the body plan editor so it is not useful for that map update logic to reside in the cell editor component.

Done. That makes total sense.

hhyyrylainen (approved):
Assuming the one bug I found is now fixed, this is good to go. I can't retest this right now as I'm doing a refactor on another PR.

hhyyrylainen:
I've now re-tested this.

PR #4277, opened 2023-04-10
Make engulfed tutorial not appear if player can't be digested and make it hide after some time

### LINKED ISSUE #3649: The player being engulfed tutorial should not appear if the player can't be digested
also there should be like a limit of 30 to 45 seconds of that tutorial being visible before it is hidden so that it doesn't get stuck on screen. Also I think this tutorial might as well have the closing X on it as I don't think skipping it needs to be prevented.

Kasterisk:

Added new method similar to `CanEngulfedObject` where instead it checks for target digestibility and returns an enum. And set the tutorial's visible time to 35s.

**Related Issues**

Fixes #3649.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Microbe.Interior.cs
+++ b/src/microbe_stage/Microbe.Interior.cs
@@ -95,6 +95,12 @@ public partial class Microbe
     private bool organelleMaxRenderPriorityDirty = true;
     private int cachedOrganelleMaxRenderPriority;
 
+    public enum DigestCheckResult
+    {
+        Ok,
+        MissingEnzyme,
+    }
+
     /// <summary>
     ///   The stored compounds in this microbe
     /// </summary>
@@ -630,6 +636,19 @@ public partial class Microbe
 
         CalculateBonusDigestibleGlucose(result);
         return result;
+    }
+
+    /// <summary>
+    ///   Returns the check result whether this microbe can digest the target (has the enzyme necessary).
+    /// </summary>
+    public DigestCheckResult CanDigestObject(IEngulfable engulfable)
+    {
+        var enzyme = engulfable.RequisiteEnzymeToDigest;
+
+        if (enzyme != null && !Enzymes.ContainsKey(enzyme))
+            return DigestCheckResult.MissingEnzyme;
+
+        return DigestCheckResult.Ok;
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


src/microbe_stage/Microbe.Interior.cs
```diff
@@ -632,6 +638,19 @@ public float EjectCompound(Compound compound, float maxAmount, Vector3 direction
         return result;
     }
 
+    /// <summary>
+    ///   Returns the check result whether this microbe can digest the target (has the enzyme necessary).
+    /// </summary>
+    public DigestCheckResult CanDigestObject(IEngulfable engulfable)
```
hhyyrylainen, line 644: Isn't there already a really similar method existing for checking if something can be engulfed?

Kasterisk: Yes but it doesn't check whether something can be _digested_.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -1449,7 +1468,7 @@ private void HandleDigestion(float delta)
 
             if (engulfable.RequisiteEnzymeToDigest != null)
             {
-                if (!Enzymes.ContainsKey(engulfable.RequisiteEnzymeToDigest))
+                if (CanDigestObject(engulfable) == DigestCheckResult.MissingEnzyme)
```
hhyyrylainen, line 1471: I think a bit of defensive programming would be good here. I'd split this to two levels: the first level gets the digest check result and enters an if when it is not "ok". Then a second level check checks for the exact errors. This ensures that when new error conditions are added this code won't accidentally let those past.

src/microbe_stage/Microbe.Interior.cs
hhyyrylainen: I see, could you update the documentation on this method and that other one to refer to each other and explain the difference?

src/microbe_stage/Microbe.Interior.cs
Kasterisk: I just realized this logic needs to be tweaked. Perhaps switch case is more appropriate?

hhyyrylainen: Probably if there's potential to have a bunch of values, I wouldn't really bother with a switch if the enum is ever going to have only 3 or 4 potential values.

Kasterisk: We just want the benefit of `default` really.

src/microbe_stage/Microbe.Interior.cs
Kasterisk: Done.

hhyyrylainen (approved):
The code changes look now fine to me, I trust that you tested the actual behaviour of the tutorial in game with these changes.

PR #4281, opened 2023-04-11
Show costs with discounts applied in the organelle upgrade GUI

### LINKED ISSUE #4089: Show costs with discounts applied in the organelle upgrade GUI
I realized that applying the discount is super easy (https://github.com/Revolutionary-Games/Thrive/pull/4088) but displaying it is a bit more involved so I'm opening this issue.

Kasterisk:

Basically just passing in the editor's `CostMultiplier` to `OrganelleUpgradeGUI.OpenForOrganelle` and then have the discount applied. We don't have to worry updating it later since the upgrade GUI's content gets recreated every time you open it.

**Related Issues**

Closes #4089.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/editor/upgrades/OrganelleUpgradeGUI.cs
```diff
@@ -112,10 +113,12 @@ public override void _Ready()
 
                 var selectionButton = upgradeSelectionButtonScene.Instance<MicrobePartSelection>();
 
+                var cost = (int)Math.Min(upgrade.MPCost * costMultiplier, 100);
```
hhyyrylainen, line 116: Is this the same math when calculating the cost of the action when it is performed? If this isn't then I worry that this might be rounded differently and give a wrong reading in the GUI...

hhyyrylainen (approved):
Okay, I confirmed that this is the same math as in `EditorCombinableActionData.CalculateCost` so I'll now merge this as a pretty obvious code change that shouldn't break things...

src/microbe_stage/editor/upgrades/OrganelleUpgradeGUI.cs
Kasterisk: Sorry, totally forgot that I made this PR 🤦‍♂️

Now that I think about it, perhaps getting the cost with `Editor.WhatWouldActionsCost` would be better as it account the history?

hhyyrylainen: That would actually make a lot of sense, and would also solve the situation where some upgrades are refunded at the same time as swapping to something else.

Kasterisk: Alright, I finally opened a PR for it https://github.com/Revolutionary-Games/Thrive/pull/4284.

PR #4284, opened 2023-04-13
Improve organelle upgrade cost display and minor custom window fixes

Kasterisk:

The cost for individual upgrades in the organelle upgrade GUI now takes into account the editor action history as an improvement to #4281. Due to this, it's perfectly possible for an upgrade to have a negative MP cost, this means that such action gives more MP instead and so must be conveyed correctly in the UI.

While working on that I came across a minor bug in the modal system which I also had to fix as it affects the organelle upgrade GUI. Further extensive testing is warranted to make sure no more problems come up with this.

**Related Issues**

Continuation of #4281.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


locale/en.po
```diff
@@ -4789,10 +4791,15 @@ msgstr "population:"
...
 
-#: ../src/microbe_stage/editor/CellEditorComponent.cs:2323
+#: ../src/microbe_stage/editor/CellEditorComponent.cs:2302
 msgid "POPULATION_IN_PATCH_SHORT"
 msgstr "{0} ({1})"
 
+#: ../src/microbe_stage/editor/MicrobePartSelection.cs:165
+#: ../src/microbe_stage/editor/tooltips/SelectionMenuToolTip.cs:343
+msgid "POSITIVE_SIGNED_NUMBER"
+msgstr "+{0}"
```
hhyyrylainen, line 4801: I actually don't agree with this being made into a translation, mainly because I've used a simple "workaround" in the prototypes already (prepend + when number > 0). This is also not symmetrical with negative numbers as there's no way to override those with a translation.

src/gui_common/CustomPopupMenu.cs
```diff
@@ -44,33 +42,26 @@ public void RemapDynamicChildren()
...
         tween.TweenProperty(this, "rect_size", Vector2.Zero, 0.15f)
             .From(cachedMinSize)
             .SetTrans(Tween.TransitionType.Circ)
             .SetEase(Tween.EaseType.Out);
-        tween.TweenCallback(this, nameof(OnFolded));
-    }
-
-    protected virtual void ResolveNodeReferences()
-    {
+        tween.TweenCallback(this, "hide");
```
hhyyrylainen, line 64: Is this intentional that this now calls "hide" instead of `nameof(Hide)`?

src/gui_common/ModalManager.cs
```diff
@@ -15,7 +16,9 @@ public class ModalManager : NodeWithInput
     /// </summary>
     private readonly System.Collections.Generic.Dictionary<CustomWindow, Node> originalParents = new();
 
-    private readonly Stack<CustomWindow> modalStack = new();
+    private readonly Queue<CustomWindow> demotedModals = new();
+
+    private Stack<CustomWindow> modalStack = new();
```
hhyyrylainen, line 21: Can this be readonly as well?

src/microbe_stage/editor/upgrades/OrganelleUpgradeGUI.cs
```diff
@@ -113,7 +112,16 @@ public override void _Ready()
 
                 var selectionButton = upgradeSelectionButtonScene.Instance<MicrobePartSelection>();
 
-                var cost = (int)Math.Min(upgrade.MPCost * costMultiplier, 100);
+                var oldUpgrade = organelle.Upgrades ?? new OrganelleUpgrades();
```
hhyyrylainen, line 115: This should be outside the loop to avoid unnecessary object creation.

locale/en.po
Kasterisk: Okay. My solution before this was just to replace - with +, but then I thought about the localization strictness we had.

hhyyrylainen: For working with pure numbers, I think it doesn't make sense to add this kind of asymmetry here. Like I said, I'd do something like `"+" + Math.Abs(number).ToString(CurrentCulture)`

src/gui_common/CustomPopupMenu.cs
Kasterisk: Yes. I can't find the link to the Godot docs reference right now but I think it stated that you must use the native name when calling engine functions like this.

hhyyrylainen: If that's the intention, that is correct, but reading the existing code, this used to send the Close method call to our code. Why the change from that to the godot inbuilt method?

src/gui_common/ModalManager.cs
Kasterisk: No, is used like this just few lines down
https://github.com/Revolutionary-Games/Thrive/blob/4cb9670399c902dd7ebea2feb04bbbe543ef94a0/src/gui_common/ModalManager.cs#L212-L221

hhyyrylainen: Well if that is a common problem, I think maybe a Deque should be used then instead to allow the expected error condition to be handled correctly?

src/gui_common/CustomPopupMenu.cs
Kasterisk: Because we no longer have to set `closing` to false, `OnFolded` isn't needed, one statement only.

Kasterisk: > Or is this an animation specific use case where the hide needs to be called?

Kind of, or any other thing you'd want the window to do when you call `Close` to it really.

https://github.com/Revolutionary-Games/Thrive/blob/4cb9670399c902dd7ebea2feb04bbbe543ef94a0/src/gui_common/CustomWindow.cs#L225-L238

> "OnHideAnimationComplete"

Already covered functionally by `OnHidden`.

hhyyrylainen: Then why does this not call that `OnHidden`?

My point here is that a custom functionality we are adding to the windows, should use our own methods to control the program flow, not rely on calling a Godot inbuilt method which then through more complicated callbacks eventually does the right thing. That's my point that I'd much prefer to still call one of our own methods here rather than "hide".

src/gui_common/ModalManager.cs
Kasterisk: The situation can occur somewhere in the middle of the stack not just both ends. Not only that, we'd have to reimplement stack behavior

> Well if that is a common problem

I wouldn't say it's that common, and one line of LINQ for it shouldn't be much of a performance sap.

hhyyrylainen: Conceptually if the thing doesn't act like a stack or queue, then those data structures shouldn't be used. A deque sounds exactly what is needed here (we already have a Deque library), it supports push and pop to both front and back, while also supporting access to removing middle elements.

src/gui_common/CustomPopupMenu.cs
Kasterisk: Okay.

locale/en.po
Kasterisk: Changed.

hhyyrylainen:
I fixed the code style problems I saw, but now with playtesting I found out that https://github.com/Revolutionary-Games/Thrive/issues/4201 seems to consistently happen to me 100% of the time the first time I open the art gallery when I've restarted the game.

Also I noticed that very frequently the starting of a slideshow fails to fullscreen (I've never seen this problem before):


The slideshow plays only on the first image area and doesn't cover the entire screen.

hhyyrylainen:
Not sure if this was fixed but when trying to modify a cilia (on master branch) I got both the modify popup and the organelle popup open at the same time. I don't remember seeing that happen when quickly testing this, so I'd guess that it was fixed already?

Kasterisk:
> Not sure if this was fixed but when trying to modify a cilia (on master branch) I got both the modify popup and the organelle popup open at the same time.

That's what this PR was trying to fix :)

> Also I noticed that very frequently the starting of a slideshow fails to fullscreen (I've never seen this problem before):

That's new, I guess one of the recent commits broke it.

Kasterisk:
Alright, everything should be fixed. Tooltips for cilia upgrades should work, slidescreen and modify popup no longer stuck.

So, the cause for why the tooltips won't show was because back when I implemented the custom tooltip system, I've made it so that when tooltipable controls went out of the tree, their relationship with their associated tooltips is severed. This is of course conflicting with the new modal system, because modal windows exiting the tree is an integral part of how it works (reparenting), I fixed this by checking whether a tooltipable goes out of tree due to reparenting or other reasons.

Also made `ToolTipManager`'s canvas layer be 128 to allow tooltips to render above the modals and the screen overlays (because it comes after the screen overlays in the autoload hierarchy).

hhyyrylainen (approved):
This seems to now at least not bug all the time. I'll merge this so that I can now try to make the different approach for the reparenting.

PR #4287, opened 2023-04-15
Fix current light level in editor not working correctly

### LINKED ISSUE #4286: Current light level button doesn't always work correctly after a patch change
I was in the editor having entered it in the ice shelf. I moved to an estuary patch and the light level was correct. Then I clicked the night button and back to the current light level button, which for some reason made the light level stay as the night and all the light values were shown incorrectly on the patch map. When I exited to the stage the light jumped to 84% (as that's what it was supposed to be). Going back to the editor everything was fine then.

Kasterisk:
Fixing slightly flawed approach in #4276.

Moved `DayNightCycle` instance out of `CreatureStageBase` higher to `GameWorld` so that it's accessible to more logic that needs it. Also refactored some TODO comments regarding updating average sunlight in patches.

Editor light level is now changed to be a daylight fraction instead. This is to make it easier to apply the correct editor-specific time of day to all patches in the patch map.

DRAFT because old save needs to be loaded correctly.

Fixes #4286.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/GameWorld.cs
+++ b/src/general/GameWorld.cs
@@ -55,6 +55,7 @@ public class GameWorld : ISaveLoadable
     public GameWorld(WorldGenerationSettings settings, Species? startingSpecies = null) : this()
     {
         WorldSettings = settings;
+        LightCycle.ApplyWorldSettings(settings);
 
         if (startingSpecies == null)
         {
@@ -92,6 +93,13 @@ public class GameWorld : ISaveLoadable
         GenerationHistory.Add(0, new GenerationRecord(
             0,
             new Dictionary<uint, SpeciesRecordLite> { { PlayerSpecies.ID, initialSpeciesRecord } }));
+
+        if (WorldSettings.DayNightCycleEnabled)
+        {
+            // Make sure average light levels are computed already
+            foreach (var patch in Map.Patches)
+                patch.Value.UpdateAverageSunlight(LightCycle);
+        }
     }
 
     /// <summary>
@@ -127,6 +135,9 @@ public class GameWorld : ISaveLoadable
 
     [JsonProperty]
     public TimedWorldOperations TimedEffects { get; private set; }
+
+    [JsonProperty]
+    public DayNightCycle LightCycle { get; private set; } = new();
 
     /// <summary>
     ///   The current external effects for the current auto-evo run. This is here to allow saving to work for them.
@@ -601,6 +612,8 @@ public class GameWorld : ISaveLoadable
     {
         if (Map == null || PlayerSpecies == null)
             throw new InvalidOperationException("Map or player species was not loaded correctly for a saved world");
+
+        LightCycle.CalculateDependentLightData(WorldSettings);
     }
 
     public void BuildEvolutionaryTree(EvolutionaryTree tree)
--- a/src/general/GameWorld.cs
+++ b/src/general/GameWorld.cs
@@ -97,8 +97,7 @@ public class GameWorld : ISaveLoadable
         if (WorldSettings.DayNightCycleEnabled)
         {
             // Make sure average light levels are computed already
-            foreach (var patch in Map.Patches)
-                patch.Value.UpdateAverageSunlight(LightCycle);
+            UpdateGlobalAverageSunlight();
         }
     }
 
@@ -190,6 +189,14 @@ public class GameWorld : ISaveLoadable
             SimulationParameters.Instance.GetOrganelleType("cytoplasm"), new Hex(0, 0), 0));
 
         species.OnEdited();
+    }
+
+    /// <summary>
+    ///   Takes care of processing everything in the world.
+    /// </summary>
+    public void Process(float delta)
+    {
+        LightCycle.Process(delta);
     }
 
     /// <summary>
@@ -606,6 +613,28 @@ public class GameWorld : ISaveLoadable
         }
 
         eventsLog[TotalPassedTime].Add(new GameEventDescription(description, iconPath, highlight));
+    }
+
+    /// <summary>
+    ///   Updates the light level in all patches according to <see cref="LightCycle"/> data.
+    /// </summary>
+    public void UpdateGlobalLightLevels()
+    {
+        foreach (var patch in Map.Patches.Values)
+        {
+            patch.UpdateCurrentSunlight(LightCycle.DayLightFraction);
+        }
+    }
+
+    /// <summary>
+    ///   Updates/sets the average light level of all patches according to <see cref="LightCycle"/> data.
+    /// </summary>
+    public void UpdateGlobalAverageSunlight()
+    {
+        foreach (var patch in Map.Patches.Values)
+        {
+            patch.UpdateAverageSunlight(LightCycle.AverageSunlight);
+        }
     }
 
     public void FinishLoading(ISaveContext? context)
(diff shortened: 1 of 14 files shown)


hhyyrylainen:
As day/night cycle is an experimental feature, I'm fine with the save loading just working in a somewhat sensible way. It's not a big deal if it loses the current info.

src/general/GameWorld.cs
```diff
@@ -92,6 +93,13 @@ public GameWorld(WorldGenerationSettings settings, Species? startingSpecies = nu
         GenerationHistory.Add(0, new GenerationRecord(
             0,
             new Dictionary<uint, SpeciesRecordLite> { { PlayerSpecies.ID, initialSpeciesRecord } }));
+
+        if (WorldSettings.DayNightCycleEnabled)
+        {
+            // Make sure average light levels are computed already
+            foreach (var patch in Map.Patches)
+                patch.Value.UpdateAverageSunlight(LightCycle);
```
hhyyrylainen, line 101: I think this should be made into a general method in the patch map just for that extra bit of clarity and guarantee against accidentally modifying just one copy of this code (I saw this same kind of looking pattern in the editor).

src/general/base_stage/StageBase.cs
```diff
@@ -107,6 +113,15 @@ public override void _Process(float delta)
 
             wantsToSave = false;
         }
+
+        GameWorld.LightCycle.Process(delta);
```
hhyyrylainen, line 117: I think it would make sense to add a `GameWorld.Process` method that takes care of processing everything in the world.

src/general/base_stage/StageBase.cs
```diff
@@ -38,6 +38,12 @@ public abstract class StageBase : NodeWithInput, IStageBase, IGodotEarlyNodeReso
 
     private bool transitionFinished;
 
+    /// <summary>
+    ///   Used to control how often light level is updated by the day/night cycle.
```
hhyyrylainen, line 42:
```suggestion
    ///   Used to control how often updated light level data is read from the day/night cycle.
```
I think wording it like this will be more accurate to what is actually going on.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2204,28 +2189,28 @@ private void ApplyLightLevelOption()
...
-                Editor.LightLevel = Editor.CurrentPatch.Biome.MinimumCompounds[sunlight].Ambient;
+                Editor.DayLightFraction = 0;
                 break;
             }
 
             case LightLevelOption.Average:
             {
                 averageLightButton.Pressed = true;
-                Editor.LightLevel = Editor.CurrentPatch.Biome.AverageCompounds[sunlight].Ambient;
+                Editor.DayLightFraction = Editor.CurrentGame.GameWorld.LightCycle.AverageSunlight;
```
hhyyrylainen, line 2206: Is this always correct? Can the average sunlight fraction not vary by patch?

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1077,6 +1071,8 @@ public override void OnLightLevelChanged(float lightLevel)
         // is non-zero too.
         if (maxLightLevel > 0.0f && templateMaxLightLevel > 0.0f)
         {
+            var lightLevel = maxLightLevel * dayLightFraction;
+
             // Normalise by maximum light level in the patch
             camera!.LightLevel = lightLevel / maxLightLevel;
```
hhyyrylainen, line 1077: Doesn't the math cancel out here.

You have first `ligthLevel = maxLightLevel * dayLightFraction` and then substituting that here gives:

```
lightLevel / maxLightLevel
maxLightLevel * dayLightFraction / maxLightLevel
dayLightFraction
```

So shouldn't this line just be `camera!.LightLevel = dayLightFraction;`?

src/microbe_stage/MicrobeStage.cs
```diff
@@ -712,31 +690,21 @@ protected override void UpdatePatchSettings(bool promptPatchNameChange = true)
...
+        if (GameWorld.Map.CurrentPatch == null)
+            return;
 
-    private void UpdateBackground()
-    {
-        Camera.SetBackground(SimulationParameters.Instance.GetBackground(
-            GameWorld.Map.CurrentPatch!.BiomeTemplate.Background));
-    }
+        patchManager.UpdatePatchBiome(GameWorld.Map.CurrentPatch);
+        patchManager.UpdateAllPatchLightLevels();
```
hhyyrylainen, line 699: Doesn't have to happen now, but could you add a TODO here that it would make more sense for the GameWorld to update its patch map data based on the light cycle in it. That'd be much clearer OOP design where not everything is messing with the internals of that object.

hhyyrylainen (changes requested):
I just tested and it seems like loading an earlier save works sensibly enough for me to not need to consider a save upgrade. I do have some code structure and clean up related comments, though.

src/microbe_stage/editor/CellEditorComponent.cs
Kasterisk: This one is universal, patches just multiply `AverageSunlight` by their maximum sunlight to get their patch-specific average sunlight.

src/microbe_stage/editor/CellEditorComponent.cs
Kasterisk: That's indeed right...

src/general/GameWorld.cs
Kasterisk: I put this in `GameWorld` to reduce unnecessary passing of `DayNightCycle` in methods.

src/microbe_stage/MicrobeStage.cs
Kasterisk: Added.

src/general/base_stage/StageBase.cs
```diff
@@ -38,6 +38,11 @@ public abstract class StageBase : NodeWithInput, IStageBase, IGodotEarlyNodeReso
 
     private bool transitionFinished;
 
+    /// <summary>
+    ///   Used to control how often updated light level data is read from the day/night cycle.
+    /// </summary>
+    private float elapsedSinceLightLevelUpdate;
```
Kasterisk, line 44: I didn't make this savable as I think the value is just too insignificant to be saved, plus I think we've been making similar fields as not for saving lately.

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: It's universal *now* I was just wondering if it might not be the case in the future, though I guess it is kind of pointless to think about now if there's no ideas how this might potentially change in the future...

src/general/base_stage/StageBase.cs
hhyyrylainen: Yeah, this is so inconsequential that I wouldn't save this. Though maybe this should default to 1 to ensure that an immediate update happens?

hhyyrylainen: If you don't mind, I'll commit this change and approve this PR now. I just playtested and I think this solved all the issues I had.

Kasterisk: Sure.

hhyyrylainen (approved):
Seems to work now with a quick test (I didn't remember the exact steps I did before to get the issue to appear). And I added that one last bit of code change to ensure when loading a game the light data is immediately applied.

PR #4294, opened 2023-04-18
Fix resolution label update

### LINKED ISSUE #4061: Toggling the fullscreen option in the options menu doesn't update the resolution label
it still keeps showing the last value (for example just 720p resolution when the window is fullscreened on a much bigger monitor).

DrRadek:

Fixes a bug where resolution isn't updated. The issue was caused by using NotificationResized notification that gets called only when the node changes its size which doesn't happen when viewport size changes while maintaining same aspect ratio.
So instead I have used a signal from viewport to react to the change properly.

**Related Issues**

fixes #4061 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/OptionsMenu.cs
+++ b/src/general/OptionsMenu.cs
@@ -488,6 +488,8 @@ public class OptionsMenu : ControlWithInput
 
         deadzoneConfigurationPopup.OnDeadzonesConfirmed += OnDeadzoneConfigurationChanged;
 
+        GetViewport().Connect("size_changed", this, nameof(DisplayResolution));
+
         selectedOptionsTab = OptionsTab.Graphics;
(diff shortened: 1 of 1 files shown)


Kasterisk (approved):
Seems good, don't know why we missed this obvious fix.

hhyyrylainen (changes requested):
After my latest PR, this now puts the signal connect to the wrong place. It needs to be moved back from `ResolveNodeReferences` to `_Ready`.

hhyyrylainen (approved):
Code is fine. I don't have the conditions due to some WM and Godot interaction stuff to test this anymore, but I'll trust everyone else's testing is enough.

Issue #2069, opened 2021-01-27
Make compound balances show the situation at ATP equilibrium


hhyyrylainen:
So right now the balance just shows if all processes run at max speed, however for balancing photosynthesis in your cell, it would be more interesting to see the situation so that processes run at max speed or the speed at which ATP balance is 0. That way seeing negative glucose number would tell the player that they need a little bit more photosynthesis to not have to worry about there not being enough glucose.

Related to: https://github.com/Revolutionary-Games/Thrive/issues/2068
Whichever of these get done first, the other should be adapted to work with the one done first.

Issue #2068, opened 2021-01-27
Add tool in editor to explore ATP / compound balances without all resources


hhyyrylainen:
It would be excellent to have some buttons or something to view what happens in the editor / editor process panel if you don't have all ATP generation resources available to you.

Related to: https://github.com/Revolutionary-Games/Thrive/issues/1979

PR #4298, opened 2023-04-19
Made editor compound balance assume ATP equilibrium

Leinourdian:

This adds a new method that calculates compound balance assuming that the cell will produce at most as much ATP as it consumes. The editor uses the new method.

Notes:
-	All in-game text is kept the same, so this assumes that players will find this change intuitive.
-	Initial compounds still use the old method for compound balance. Changing them to use the new one might make sense.
-	I’m not sure, but this might be slightly different from true compound balance during gameplay if e.g. process ordering and storage capacity make different processes happen at different speeds.
-	The code of the new method `CalculateCompoundBalanceAtEquilibrium` is very similar to the old method `CalculateCompoundBalance`. If wanted, the old method could be removed, or the methods or their parts could be merged into a single method.
-	This method should be pretty simple to adapt to use different presets like not moving or lacking certain compounds. Could be useful for #2068 in the future.
-	This doesn’t show ATP. If it did, it would show either 0 or a negative value. I think it’s probably unnecessary to show ATP since the ATP balance bar exists.
-	If you consume more ATP than you produce, the method treats that as you being at equilibrium. This means that ATP consuming processes (like producing toxins) are always considered to run at full speed.

**Related Issues**

Closes [#2069](https://github.com/Revolutionary-Games/Thrive/issues/2069)

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/ProcessSystem.cs
+++ b/src/microbe_stage/ProcessSystem.cs
@@ -167,7 +167,7 @@ public class ProcessSystem
     /// </summary>
     /// <remarks>
     ///   <para>
-    ///     Ignores how much ATP the cell needs
+    ///     Assumes that all processes run at maximum speed
     ///   </para>
     /// </remarks>
     public static Dictionary<Compound, CompoundBalance> ComputeCompoundBalance(
--- a/src/microbe_stage/ProcessSystem.cs
+++ b/src/microbe_stage/ProcessSystem.cs
@@ -165,6 +165,11 @@ public class ProcessSystem
     ///   Computes the compound balances for given organelle list in a patch and at a given time during the day (or
     ///   using longer timespan values)
     /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Ignores how much ATP the cell needs
+    ///   </para>
+    /// </remarks>
     public static Dictionary<Compound, CompoundBalance> ComputeCompoundBalance(
         IEnumerable<OrganelleDefinition> organelles, BiomeConditions biome, CompoundAmountType amountType)
     {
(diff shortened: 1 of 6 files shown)


src/microbe_stage/ProcessSystem.cs
```diff
@@ -165,6 +165,11 @@ public ProcessSystem(Node worldRoot)
     ///   Computes the compound balances for given organelle list in a patch and at a given time during the day (or
     ///   using longer timespan values)
     /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Ignores how much ATP the cell needs
```
hhyyrylainen, line 170:
```suggestion
    ///     Ignores how much ATP the cell needs (i.e. assumes instead that all processes run always at max possible speed)
```

src/microbe_stage/ProcessSystem.cs
```diff
@@ -207,6 +212,77 @@ void MakeSureResultExists(Compound compound)
...
+    ///   Computes the compound balances for given organelle list in a patch and at a given time during the day (or
+    ///   using longer timespan values)
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Assumes that the cell produces at most as much ATP as it consumes
+    ///   </para>
+    /// </remarks>
+    public static Dictionary<Compound, CompoundBalance> ComputeCompoundBalanceAtEquilibrium(
+        IEnumerable<OrganelleTemplate> organelles, BiomeConditions biome, CompoundAmountType amountType,
```
hhyyrylainen, line 225: Is there a reason why this takes in organelle templates and not organelle definitions like the `ComputeCompoundBalance` method?

src/microbe_stage/ProcessSystem.cs
```diff
@@ -207,6 +212,77 @@ void MakeSureResultExists(Compound compound)
...
+        foreach (var organelle in organelles)
+        {
+            foreach (var process in organelle.Definition.RunnableProcesses)
+            {
+                var speedAdjusted = CalculateProcessMaximumSpeed(process, biome, amountType);
+
+                useRatio = false;
+
+                // If the cell produces more ATP than it needs, its ATP producing processes need to be toned down
+                if (speedAdjusted.Outputs.Select(o => o.Key).Contains(ATP) && consumptionProductionRatio < 1.0f)
```
hhyyrylainen, line 250:
```suggestion
                if (speedAdjusted.Outputs.ContainsKey(ATP) && consumptionProductionRatio < 1.0f)
```
Would that work?

hhyyrylainen:
There's just a tiny bit of changes needed for this so this should be completable in time for the RC on Saturday.

src/microbe_stage/ProcessSystem.cs
Leinourdian: Not really actually, I can change this to be consistent with the other method.

src/microbe_stage/ProcessSystem.cs
Leinourdian: Unfortunately not, `Outputs` doesn't have the `ContainsKey` method or anything similar as far as I can tell.

hhyyrylainen: I think refactoring the outputs property to be like this: `public IReadOnlyDictionary<Compound, float> Outputs => WritableOutputs;` would be a good choice here.

hhyyrylainen (approved):
The code change looks fine to me now. You Just need to resolve the one merge conflict before this can be merged. And also I'll give time at least until tomorrow for playtesting this, in case someone does it.

buckly90:
Works wonderfully and without any visible error. Outstanding job!

PR #4308, opened 2023-04-22
Added dialog window ordering support

DrRadek:

I tried to make the dialog windows to be able to change their order. It works by placing last grabbed window on top of other windows on the same layer.

 I have added a WindowReorderingSupportTest scene for testing and for showing how it works.
The change currently works on microbe stage tutorial windows and on awakening stage inventory. I didn't encounter any issues.
I need help with finding where else it would be possible to use this feature ~~and it would be nice to find a workaround around the foreach loop in CustomDialog.cs on line 283.~~ I found a solution for this.
~~I also tried to put the inventory panels as children to make them always appear on top of the inventory, but the editor always messed up its size and position so I decided to let it be.~~ I made this work while keeping the nodes on the same level in the hierarchy.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


DrRadek:
I think I am done with the system. All that is left is to add the support to more suitable places that I don't know about.

test/WindowReorderingSupportTest.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        EnableAllWindows(GetChildren());
+    }
+
+    private void EnableAllWindows(Array children)
+    {
+        foreach (var child in children)
+        {
+            (child as CustomWindow)?.Show();
+            EnableAllWindows((child as Node)!.GetChildren());
```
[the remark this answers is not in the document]
DrRadek: I can't test it rn but I see no reason why it shouldn't work so I will trust you.

src/gui_common/CustomWindow.cs
```diff
@@ -11,11 +12,38 @@
 /// </remarks>
 public class CustomWindow : Control
 {
+    [Export]
+    public Array<NodePath> WindowReorderingPaths = new();
```
hhyyrylainen, line 16: It looks like the reordering functionality is not implemented at this class level, so shouldn't all of these properties be put in `CustomDialog` instead as that's the level where these seem to be used?

DrRadek: You are right. I wanted to have them here in case other children of this class needed them but I didn't think of that they would be doing nothing in this this class.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,263 @@
...
+    /// <remarks>
+    ///   <para>
+    ///     Ignored when window reordering paths are not empty
+    ///   </para>
+    /// </remarks>
+    [Export]
+    public int AutomaticWindowReorderingDepth = 0;
+
+    [Export]
+    public bool ReorderWindows = true;
```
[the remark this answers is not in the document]
DrRadek: It can work as a middleman (to pass connections to its ancestors). Theoretically it can be used to automatically connect windows that are too deep in the hierarchy. I believe that this is an edge case where it will usually be better to set a manual path for those so if you think that this is not a good enough reason to have it then I will remove it.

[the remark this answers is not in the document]
hhyyrylainen: I think it should be removed. I saw a depth control variable as well, so reaching up a bunch would work by increasing that as well, probably?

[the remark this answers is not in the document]
DrRadek: It exists for performance reasons and to prevent it from accidentally connecting to something one doesn't want it to connect to. Letting it to try to connect to every single ancestor sounds like something that should be prevented but otherwise performance shouldn't be an issue, I think? 10 sounds like too much to me because it's rarely needed if ever. What about 5?

It's possible but I can't think of anything to improve it. For now I will try to write the comments.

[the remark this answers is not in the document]
hhyyrylainen: I think 10 is a nice round number and the performance difference shouldn't be too different from 5, I think.
And also the case anyway should be that each level of windows would have their reordering support, because missing reordering support is a sign of programming bug once this is done. I guarantee that someone at some point will notice and comment about a bug where just one window in the game doesn't go to the top of others when clicked.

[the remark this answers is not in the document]
DrRadek: Okay. What about a connection between two levels of windows (So from a reordering support node to its reordering support ancestor)? The same should apply to those too, right?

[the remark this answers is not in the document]
hhyyrylainen: Well my initial thought would be that the sibling reorder would intelligently determine which node to move even for a really deep window. But also the design which you seem to want to go for where you have multiple levels of reorder sibling nodes, is probably good. In this case as you'd usually not have the higher level sibling reorder node I think here maybe 5 would be a better search depth as most of the time the search would fail and hit the limit.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,263 @@
+﻿using System.Collections.Generic;
+using Godot;
+using Godot.Collections;
+using static CustomWindow;
+
+public class AddWindowReorderingSupportToSiblings : Control
```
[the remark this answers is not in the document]
DrRadek: I tried to write the comments. I hope that it will be more understandable now.

I can't think of anything that would be possible to cut down.

src/gui_common/CustomWindow.cs
DrRadek: I don't know why but after this change godot wants to save default values of those properties into scenes which means that changing the default values has no effect in the places where it has been saved already. Is it an issue that is just on my side or am I missing something?
obrazek

hhyyrylainen: Godot does sometimes nonsensical property saves. Usually helps to rebuild the code from Godot and then close and reopen the Godot editor after that it usually for me starts to correctly detect default values when saving scenes.

src/awakening_stage/gui/InventoryScreen.tscn
```diff
@@ -98,6 +74,7 @@ margin_left = 100.0
 margin_top = 100.0
 margin_right = -465.0
 margin_bottom = -559.0
+AutomaticWindowReorderingDepth = 0
```
hhyyrylainen, line 77: Is there a reason why this window is so special that it needs a non-default value here?

(I'm trying to work out all the smaller file's problems to get them out of the way).

src/gui_common/CustomWindow.cs
```diff
@@ -16,6 +17,13 @@ public class CustomWindow : Control
 
     private bool hasBeenRemovedFromTree;
 
+#pragma warning disable CA2213
+    private Array<AddWindowReorderingSupportToSiblings> windowReorderingNodes = new();
+#pragma warning restore CA2213
+
+    [Signal]
+    public delegate void Dragged(CustomWindow window);
```
hhyyrylainen, line 25: This class still has functionality in it that it doesn't actually support. I think still all of the changes in your PR to this CustomWindow class should all move to CustomDialog so that half of an implementation of a feature is not in this file where it doesn't actually do anything without additional parts.

src/gui_common/CustomWindow.cs
```diff
@@ -16,6 +17,13 @@ public class CustomWindow : Control
 
     private bool hasBeenRemovedFromTree;
 
+#pragma warning disable CA2213
+    private Array<AddWindowReorderingSupportToSiblings> windowReorderingNodes = new();
```
hhyyrylainen, line 21: Is this necessary to use the Godot array type here? If it is not absolutely necessary (for use in Godot calls often etc.) List<T> should always be preferred in other situations.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+/// <remarks>
+///   <para>
+///     The windows are the ones who ask to establish the connection. If the window asks for a recursive
+///     connection then the connection gets passed to window reordering nodes in ancestors too.
+///   </para>
+/// </remarks>
+public class AddWindowReorderingSupportToSiblings : Control
+{
+    /// <summary>
+    ///   Paths to window reordering nodes in ancestors.
```
hhyyrylainen, line 18: I think this should clarify that when set these override the default automatic search functionality.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+    /// </summary>
+    [Export]
+    public Array<NodePath> WindowReorderingPaths = new();
+
+    /// <summary>
+    ///   Finds first window reordering node in ancestors to connect to.
```
hhyyrylainen, line 24: This should clarify that this controls the search depth as that's the point of this variable, right now this comment is still pretty unclear.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+
+    /// <summary>
+    ///   Pairs containing window reordering nodes and what sibling of theirs is an ancestor of this class.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Used to pass the connections from windows.
+    ///   </para>
+    /// </remarks>
+    private List<(AddWindowReorderingSupportToSiblings Node, Node Sibling)> windowReorderingAncestors = new();
```
hhyyrylainen, line 44:
```suggestion
    private List<(AddWindowReorderingSupportToSiblings ReorderingNode, Node Sibling)> windowReorderingAncestors = new();
```
A slightly improved naming clarity.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+
+    /// <summary>
+    ///   A sibling that is an ancestor of a window that is currently on top of others.
```
hhyyrylainen, line 47: I think I still don't fully understand this variable. Could you add a short remarks section detailing an example or two of in which situations this variable is used?

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+    /// </summary>
+    private Node? topSibling;
+
+    private Node parent = null!;
+
+#pragma warning restore CA2213
+
+    private bool isReady;
```
hhyyrylainen, line 55: This is not related to _ready so you need to come up with a different name. Maybe something like `connectionsEstablished`?

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+    private HashSet<Node> connectedSiblings = new();
+
+    /// <summary>
+    ///   Finds window reordering nodes in ancestors, looks for manual paths if set, otherwise uses automatic search to
+    ///   find first one.
+    /// </summary>
+    /// <returns>
+    ///   Pairs containing window reordering nodes and what sibling of theirs is an ancestor of this class.
+    /// </returns>
+    public static IEnumerable<(AddWindowReorderingSupportToSiblings Node, Node Sibling)> GetWindowReorderingAncestors(
```
hhyyrylainen, line 74: `ReorderingNode` could be also used here for more clarity on the naming as using `Node` in this case is a bit too unclear in my opinion.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+        Node startingNode, int maxSearchDepth, Array<NodePath>? paths)
```
hhyyrylainen, line 75: As this is a public helper method, could you add documentation xml for each of the parameters?

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+    {
+        if (paths is null || paths.Count == 0)
+        {
+            foreach (var windowReorderingSupport in GetWindowReorderingAncestors(startingNode, maxSearchDepth))
```
hhyyrylainen, line 79: Can't this code just directly return the enumerable from the recursive call? I don't think it is actually necessary in this case to rely on manually forwarding each item one by one.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+        {
+            // Connection is recursive, connect to other window reordering nodes
+            for (int i = 0; i < windowReorderingAncestors.Count; i++)
+            {
+                windowReorderingAncestors[i].Node.ConnectWindow(
+                    window, windowReorderingAncestors[i].Sibling, recursive);
+            }
+        }
+
+        window.Connect(nameof(Dragged), this, nameof(OnWindowReorder));
```
hhyyrylainen, line 114: Is there a possibility of accidentally double registering a window? For example if a partial subtree exits the scene tree and is then added back in?

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+    {
+        // Make sure to be connected to other window reordering nodes before trying to connect the window
+        Setup();
+
+        if (recursive)
+        {
+            // Connection is recursive, connect to other window reordering nodes
+            for (int i = 0; i < windowReorderingAncestors.Count; i++)
+            {
+                windowReorderingAncestors[i].Node.ConnectWindow(
```
hhyyrylainen, line 109: Reading this bit of code, I now understand more of what the design you are going for. The higher level reordering level is handled by the window registering itself there as well. Just an observation you don't need to do anything about this.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+                yield return windowReorderingSupport;
+        }
+    }
+
+    public override void _ExitTree()
+    {
+        windowReorderingAncestors.Clear();
+        connectedWindows.Clear();
+        connectedSiblings.Clear();
+        isReady = false;
```
hhyyrylainen, line 96: Should topSibling be reset here or not?

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+
+    public void ConnectWindow(CustomWindow window, Node topNode, bool recursive)
+    {
+        // Make sure to be connected to other window reordering nodes before trying to connect the window
+        Setup();
+
+        if (recursive)
+        {
+            // Connection is recursive, connect to other window reordering nodes
+            for (int i = 0; i < windowReorderingAncestors.Count; i++)
```
hhyyrylainen, line 107: This loops should be rewritten as a foreach loop, our style is not to use manual for loops unless absolutely necessary.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+        if (topSibling is null || topSibling.GetIndex() < topNode.GetIndex())
+            topSibling = topNode;
+    }
+
+    public void DisconnectWindow(CustomWindow window, bool recursive)
+    {
+        if (recursive)
+        {
+            // Connection was recursive, disconnect from other window reordering nodes
+            for (int i = 0; i < windowReorderingAncestors.Count; i++)
```
hhyyrylainen, line 129: Same here, this seems like it could easily be a foreach loop.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+            // Top sibling is already current
+            return;
+        }
+
+        var siblings = parent.GetChildren();
+
+        // Search for the first sibling that is part of the reordering system
+        for (int i = siblings.Count - 1; i >= 0; i--)
+        {
+            var sibling = (Node)siblings[i];
```
hhyyrylainen, line 174:
```suggestion
        // Search for the first sibling that is part of the reordering system
        foreach (Node sibling in parent.GetChildren())
        {
```

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+    ///   others.
+    /// </summary>
+    public void OnWindowReorder(CustomDialog window)
+    {
+        // Get a sibling that is an ancestor of this window
+        Node targetSibling = connectedWindows[window];
+
+        UpdateTopSibling();
+
+        int topSiblingIndex = topSibling!.GetIndex();
```
hhyyrylainen, line 194: It's possible for UpdateTopSibling to not find anything so an explicit null check and exception with an error message should be performed on `topSibling`.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
hhyyrylainen: Sorry I misread the loop order, I see you reversed the loop.

I saw below that UpdateTopSibling seems to be called quite often in a situation where the last child is already on top.

So actually I suggest you do something like:
```
var childCount = parent.GetChildCount();

for (int i = childCount - 1; ...)
{
    var sibling = parent.GetChild(i);
}
```

That way a full list of children don't need t obe generated and marshalled into C# unnecessarily.

hhyyrylainen (changes requested):
Sorry I forgot to start my review early so most of my "review" comments are just single comments above this review.

I'm now now about grasping the overall functionality of this feature and I think the design is suitable and can't really suggest any simplification (I won't ask you to reimplement this function based on my original design I wanted). Instead all the comments are just about small code style things and further clarifications. I didn't playtest this yet and asked Kasterisk's opinion on one hack in this PR.

A bit more work and I think we can then start considering merging this. As long as I don't run into bugs in playtesting this.

src/gui_common/CustomWindow.cs
DrRadek: I still can't get this to function properly no matter what I do. Even cloning Thrive again didn't fix it. I currently don't need to make any scene changes (I hope) so I will let it be for now.

src/awakening_stage/gui/InventoryScreen.tscn
DrRadek: To me it makes sense for it to be always behind its panels. Another approach would be to put the panels as its children, but when I tried to do it like this the sizes of the windows got a little bit messed up.

src/gui_common/CustomWindow.cs
DrRadek: I will change it. It wasn't an intention to use it.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: Sure. I have written something that should make it clear I hope. In short it's a variable used to check if the window is already on the top and if not then its index is used to move current window on the top (and then the moved windows Ancestor becomes new topSibling).
It doesn't have to be saved, but I don't think that it hurts to save it and just update it after changes.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: Sounds good to me, I will use it.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: It can. I am pretty new to enumerables and I didn't know it. Thank you.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: From my testing I ran into issues only when I accidentally set manual connections in a wrong way. 
But it might be possible? I will add a check here and in DisconnectWindow to be sure that it won't break.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: Good observation. Not here but it should be reset after disconnecting a window that is on the top. I have added a check to the DisconnectWindow method.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: Thank you. ~~Not just than, I have also used an unnecessary post-increment.~~

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: This change is good.
It's called often but most of the time topSibling is not null. This part of the code is used only when topSibling becomes invalid (like after disconnecting a window that is on the top).

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: I don't see how this can happen. Reason: there has to be at least one sibling (an ancestor of the window that asked to be reordered). So I think that this check is not useful here. Or do you think that I am wrong?

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+    /// </returns>
+    private static IEnumerable<(AddWindowReorderingSupportToSiblings Node, Node Sibling)> GetWindowReorderingAncestors(
+        Node startingNode, int maxSearchDepth)
+    {
+        Node childOfAncestor = startingNode;
+        Node ancestor = childOfAncestor.GetParent();
+
+        for (int i = 0; i < maxSearchDepth; ++i)
+        {
+            if (ancestor is null)
```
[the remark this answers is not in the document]
DrRadek: Oh, I have missed this. Yes, I have used it in other places too. I will change it.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: Actually I think that clearing the list of windows and siblings is also not needed because they should get disconnected properly and there should be nothing left in them but I will let it be here and also reset the top sibling just in case.

src/awakening_stage/gui/InventoryScreen.tscn
hhyyrylainen: To me the inventory is the more important screen than the complimentary panels.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
hhyyrylainen: Well if you make a duplicate Connect call, Godot will yell an error at you. So we should really avoid that situation. We have a method for connecting a signal if not connected already, but I think a better approach would be to ensure that if this is called with an already registered window (to this object) an error is thrown / printed and nothing is done.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
hhyyrylainen: Exactly my thoughts, if the other data is cleared here the top window should also be cleared as the other data related to the top window won't exist anymore.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
hhyyrylainen: Well this is a public method, so someone could accidentally call this with an unrelated CustomDialog object... so a bit of defensive programming could be used here to throw a specific exception with an error message in that problem situation.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+            // This window is already on the top
+            return;
+        }
+
+        // Put the sibling on the top
+        parent.MoveChild(targetSibling, topSiblingIndex);
+
+        topSibling = targetSibling;
+
+        // For unexplained reasons this has to be here to update the order visually
```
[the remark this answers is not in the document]
hhyyrylainen: Still sounds a bit like a hack. One potential bug I realized is that this doesn't preserve the on top state. Instead as this hack is required for now the current top level should be stored and restored like:
```c#
bool isTopLevel = window.IsTopLevel(); // I hope there's a method or property like this
window.SetTopLevel(!isTopLevel);
window.SetTopLevel(isTopLevel);
```

At least that code will leave the window in the same state as it previously was. Because otherwise I don't want to have to debug the new popup modal system that has already had an absolute ton of bugs related to it.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,328 @@
...
+    ///   Pairs containing window reordering nodes and what sibling of theirs is an ancestor of this class.
+    /// </returns>
+    private static IEnumerable<(AddWindowReorderingSupportToSiblings Node, Node Sibling)> GetWindowReorderingAncestors(
+        Node startingNode, Array<NodePath> windowReorderingNodePaths)
+    {
+        Node childOfAncestor = startingNode;
+        Node ancestor = childOfAncestor.GetParent();
+
+        System.Collections.Generic.Dictionary
+            <Node, AddWindowReorderingSupportToSiblings> windowReorderingNodesWithParents = new();
```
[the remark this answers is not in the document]
hhyyrylainen: If you just need to know whether a node is underneath another path, you can get the path of the current node and then (after converting to strings) just check if that node's path starts with any of the ancestor paths to start. I've used this approach in the focus grabber system to avoid having to loop a bunch of parent nodes instead of doing just one string operation.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,375 @@
...
+
+    /// <summary>
+    ///   Pairs containing window reordering nodes and what sibling of theirs is an ancestor of this class.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Used to pass the connections from windows.
+    ///   </para>
+    /// </remarks>
+    private List<(AddWindowReorderingSupportToSiblings ReorderingNode, Node Sibling)> windowReorderingAncestors =
```
hhyyrylainen, line 54: Now as this is a list, this doesn't need to be inside the pragma, and I think this should be now able to be changed to a readonly. Otherwise this variable will break the rule that all Godot node types should be before other variables of a class.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,375 @@
...
+        // Find if another connected window has the same sibling
+        bool foundSibling = false;
+        foreach (var sibling in connectedSiblings)
+        {
+            if (sibling == windowSibling)
+            {
+                foundSibling = true;
+                break;
+            }
+        }
```
hhyyrylainen, line 188:
```suggestion
        bool foundSibling = connectedSiblings.Contains(windowSibling);
```

Isn't that the equivalent code? No need to manually search through a list.

src/gui_common/dialogs/CustomDialog.cs
```diff
@@ -65,6 +96,8 @@ public class CustomDialog : CustomWindow
     private StyleBox closeButtonHighlight = null!;
 
     private Font? titleFont;
+
+    private List<AddWindowReorderingSupportToSiblings> windowReorderingNodes = new();
```
hhyyrylainen, line 100: This shouldn't be inside the pragma. Also this could probably be readonly so that this should be placed before the pragma section with godot-type holding variables.

src/gui_common/dialogs/CustomDialog.cs
```diff
@@ -395,6 +442,38 @@ protected override void ApplyRectSettings()
...
+
+        foreach (var windowReordering in windowReorderingAncestorsIEnumerable)
+        {
+            windowReordering.ReorderingNode.ConnectWindow(
+                this, windowReordering.Sibling, AllowWindowReorderingRecursion);
+            windowReorderingNodes.Add(windowReordering.ReorderingNode);
+        }
+    }
+
+    protected void DisconnectFromWindowReorderingNodes()
```
hhyyrylainen, line 458: Shouldn't this and the above method be private as these are not meant to be called by derived classes?

src/gui_common/dialogs/CustomDialog.cs
```diff
@@ -395,6 +442,38 @@ protected override void ApplyRectSettings()
         }
     }
 
+    protected void ConnectToWindowReorderingNodes()
+    {
+        var windowReorderingAncestorsIEnumerable = AddWindowReorderingSupportToSiblings.GetWindowReorderingAncestors(
+            this, AutomaticWindowReorderingDepth, WindowReorderingPaths);
+
+        foreach (var windowReordering in windowReorderingAncestorsIEnumerable)
```
hhyyrylainen, line 450:
```suggestion
        foreach (var (reorderingNode, nodeSibling) in windowReorderingAncestorsIEnumerable)
```
I think destructuring the list of tuples should work like that...

src/gui_common/dialogs/CustomDialog.cs
```diff
@@ -395,6 +442,38 @@ protected override void ApplyRectSettings()
...
+            windowReordering.ReorderingNode.ConnectWindow(
+                this, windowReordering.Sibling, AllowWindowReorderingRecursion);
+            windowReorderingNodes.Add(windowReordering.ReorderingNode);
+        }
+    }
+
+    protected void DisconnectFromWindowReorderingNodes()
+    {
+        foreach (var node in windowReorderingNodes)
+            node.DisconnectWindow(this, AllowWindowReorderingRecursion);
```
hhyyrylainen, line 461: Maybe `AllowWindowReorderingRecursion` should have a warning that it may not be changed while this is attached to the node tree? As that will cause a bug if someone sets that to false after the connections are made and then removes this from the scene tree.

src/gui_common/CustomWindow.cs
hhyyrylainen: Once this is ready for playtesting, I'll investigate this myself.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,375 @@
...
+    ///   Used to save what siblings are part of the reordering system to prevent reordering more than is necessary.
+    /// </summary>
+    private HashSet<Node> connectedSiblings = new();
+
+    /// <summary>
+    ///   Finds window reordering nodes in ancestors, looks for manual paths if set, otherwise uses automatic search to
+    ///   find first one.
+    /// </summary>
+    /// <param name="startingNode">
+    ///   A node to start searching from. The search itself starts from the nodes parent.
```
hhyyrylainen, line 93:
```suggestion
    ///   A node to start searching from. The search itself starts from the node's parent.
```
Typo

src/gui_common/AddWindowReorderingSupportToSiblings.cs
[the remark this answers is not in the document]
DrRadek: For some reason I assumed that windows are always on the top. I wonder if this hack is necessary for windows that are not on the top too. But it will probably be better to use it for them too just in case... or to do more testing to know if it's really necessary.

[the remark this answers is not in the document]
hhyyrylainen: Only modal popups (that disallow you from interacting with the rest of the GUI) are set as top level. The relevant code is: 

https://github.com/Revolutionary-Games/Thrive/blob/f0c0322add610681556122cae3687519fefd37b7/src/gui_common/ModalManager.cs#L10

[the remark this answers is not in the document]
Kasterisk: I don't quite fully get why this is necessary. As I understand it, the visual order of the windows should already be updated when their position is changed in the tree, at least that's what I observed in `ModalManager`.

CustomWindows and its descendants should always be top level, I don't see why they would otherwise?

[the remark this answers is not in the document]
hhyyrylainen: Oh sorry, I mixed up being top level and being modal. Like usual, that PR where this was an important distinction still haunts me it seems.

But yeah, I also found this very strange which is why I said that if this code is kept we need to open an issue about trying to remove the hack at some point if we can't come up with anything for this PR.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,375 @@
...
+        int topSiblingIndex = topSibling!.GetIndex();
+
+        if (topSiblingIndex == targetSibling.GetIndex())
+        {
+            // This window is already on the top
+            return;
+        }
+
+        // Put the sibling on the top
+        parent.MoveChild(targetSibling, topSiblingIndex);
```
Kasterisk, line 245: Can't we just simply make the index be 0, as that should always move any node to the top?

hhyyrylainen: I explicitly requested that this PR wouldn't reorder non-window nodes. Setting index to 0 might put some GUI stuff before nodes that want to be ahead of the GUI, like most of our stage scenes are setup like. So this is to ensure that only the window positions will jump to each other, and not ahead of other non-window nodes.

Kasterisk: I see, I somehow thought that all reorderable windows are within the scope of AddWindowReorderingSupportToSiblings (I see there's a dedicated node for it) just like how ModalManager works.

hhyyrylainen: Well not exactly. Especially as the following case needs to work:

- Some Control
  - ReorderSiblingsNode
  - Window
  - Window
  - Control
    - ReorderSiblingsNode
    - Window
    - Window
  - Window

So if the first Window inside the sub-control is clicked then the whole sub-control tree needs to move in its parent and also the two windows in it need to swap places.

src/awakening_stage/gui/InventoryScreen.tscn
DrRadek: The panels are smaller and they easily get lost behind the inventory after moving it. Also there is this option to hide/show the panels if one doesn't need them. I will change it back if you still think that this is bad.

hhyyrylainen: I didn't bring this up before but I though that maybe when opening a window should reorder to the top? So it is less likely that you lose some small windows as when you reopen them they would come on top. Though if that's a bit complicated to add that could be another PR.

For the inventory I think the inventory should at least be on top of the ground panel by default and it should be able to be brought on top of the other panels. It'd be pretty weird to have that one window specifically not be able to go on top as I'm sure someone will notice and eventually submit a bug report.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: So should it be thrown or just printed? To me printing it sounds enough.

src/awakening_stage/gui/InventoryScreen.tscn
DrRadek: I will revert the change then. I will take a look at it and I will see. I think that it can be done easily with the signal, but the signal should be renamed to something else than Dragged I think. Maybe something like ReorderRequested?

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: This and OnWindowReorder shoud be private. I will fix it and well, it won't hurt if I add the exception here.

src/awakening_stage/gui/InventoryScreen.tscn
hhyyrylainen: Well I think a signal for drag might be useful in the future so it should be kept. Unless I'm misremembering there should already be some way to signal that the node has become visible. Maybe even the godot standard becoming visible signal could be used?

src/gui_common/AddWindowReorderingSupportToSiblings.cs
hhyyrylainen: I guess printing would be better in this case to not randomly crash if the problem ever occurs.

src/awakening_stage/gui/InventoryScreen.tscn
DrRadek: Okay. What about making a signal Opened in CustomDialog and emitting it from the OnOpen method? For CanvasItem there seem to be just hide and visibility_changed signals (I mean from relevant signals) so those signals can't be used directly anyway.

hhyyrylainen: That's about what I had in mind: using the virtual open method to emit a signal if there wasn't an already existing signal that couldn't be used.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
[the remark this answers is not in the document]
DrRadek: I wrote a new logic for that and I added checks if the path is correct. There was also a bug because the paths should be relative to the startingNode so I rewrote the comment about this propertly and added a check to automatic search to skip the first step if the startingNode is reordering node to prevent returning itself. (In case of reordering node I passed its parent and I changed it to pass itself)

src/gui_common/dialogs/CustomDialog.cs
DrRadek: You are right.

src/gui_common/dialogs/CustomDialog.cs
DrRadek: What about writing a getter and setter with a logic to prevent it from breaking?

src/gui_common/AddWindowReorderingSupportToSiblings.cs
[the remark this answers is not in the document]
hhyyrylainen: In case it helps I wrote a helper method for converting paths to absolute paths. So you could use that here as well. I think it's also a good idea to add caching when converting the paths to strings to avoid unnecessary operations. Though the cached version will need to be cleared when the node exits the tree as it might get reattached at a different place meaning that relative paths will have a different absolute results after that.

src/gui_common/dialogs/CustomDialog.cs
hhyyrylainen: That could be doable as well, but the logic to reapply the different values would be pretty complicated as unregistering and re-registering with the right recursive type would be needed.

An easier fix I also came up with was storing the original value of window recursion when registering, then that variable could be used here independently of whether AllowWindowReorderingRecursion was changed or not. If this approach is taken a comment should be added to AllowWindowReorderingRecursion to say that it doesn't apply immediately if the node is currently in use.

DrRadek: There are already methods that do the registering and unregistering so it's also simple.

It's more about how it should behave and I am not sure what is the right behaviour. Is there ever a reason to change it in the code? We can make it readonly if this is the case, otherwise your approach is probably better because in the current tree it's usually not expected for it to change the behaviour, I think.

hhyyrylainen: Well you can't have the Godot set stuff be readonly. And anyway the paths *will* change if nothing else the modal system will yank windows out from the scene tree and place them in completely different places. Though maybe that actually uncovers a bug here: the relative paths (and absolute if they are resolved) will definitely break when the window is moved to be a child of the modal popup layer when it becomes a modal.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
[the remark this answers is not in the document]
DrRadek: What helper method? If this is to skip getting the node and just getting the path then it's not useful in this case because the nodes are also needed. GetPath() is enough in this case, I think.

This code is used only after entering the tree. If you want the code to be more performant then the paths can point to the windows ancestor instead of the window reordering node. This way we can skip the manipulations with absolute paths and can just look for "..\AddWindowReorderingSupportToSiblings" from there.

I have already implemented it so you can take a look at it.

Also the only reason to convert paths to strings here is to check if the path is set correctly.
I think that I can check the nodes themselves instead of using Contains on strings so the only string would be the starting node path and each starting node is needed only once after entering the tree so there is no reason to cache it.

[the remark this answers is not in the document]
hhyyrylainen: Before I take a look, you could look at how FocusGrabber class does it, especially how skipOverridingStringConverted is generated and regenerated. Having the paths precomputed is what allows skipping parent node processing as you can just use a simple string starts with to compare if something is an ancestor of some node instead of needing to go from C# -> C++ -> C# -> C++ which is the case when needing to call a bunch of Godot methods in a row.

src/gui_common/dialogs/CustomDialog.cs
DrRadek: It worked when I set them to readonly. Godot allows to change them in the editor.
The list of paths should contain only relative paths if I am not mistaken? (I don't have any check for if they are relative or absolute) Also they shouldn't break. It skips connecting to non-existent nodes.

There isn't a way to properly change the paths though. If this is needed then I can also apply the changes after re-entering the tree and not immediately.

hhyyrylainen: If you put a `readonly` modifier on it, that's not good enough. This is C# so while the reference can't be updated to point to a different object, the list object is still modifiable by any piece of code.

>  (I don't have any check for if they are relative or absolute) Also they shouldn't break. It skips connecting to non-existent nodes.

I guess I'd need to read the code to figure out. But then why was there a need to cache the ancestor nodes for checking? If that kind of relative node check could just be done from the beginning?

Also if anyone ever uses a code approach to setup a window, they might have an absolute path set as those are easier to work with when modifying nodes in the tree as keeping relative paths up to date is much more difficult when manually manipulating things through code.

> There isn't a way to properly change the paths though. If this is needed then I can also apply the changes after re-entering the tree and not immediately.

The paths is a public property so changing its contents should work, similarly to why I wrote the FocusGrabber to redo its paths.

DrRadek: I meant readonly for the AllowWindowReorderingRecursion, but I think that applying it later is a better approach.

> I guess I'd need to read the code to figure out. But then why was there a need to cache the ancestor nodes for checking? If that kind of relative node check could just be done from the beginning?

If you mean the part with the dictionary in the GetWindowReorderingAncestors method then it was just a bad code. The complications come with finding the right sibling.

> Also if anyone ever uses a code approach to setup a window, they might have an absolute path set as those are easier to work with when modifying nodes in the tree as keeping relative paths up to date is much more difficult when manually manipulating things through code.

Well then I will have count with that too.

>The paths is a public property so changing its contents should work, similarly to why I wrote the FocusGrabber to redo its paths.

Actually there is no need for additional logic for that. The paths are not used when exiting a tree so they can be changed directly and the change will take effect after re-entering the tree.

hhyyrylainen: Guess I need to read your latest changes before I can offer useful insight. At the latest I'll have time in a couple of days to do that.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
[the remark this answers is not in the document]
DrRadek: I was wrong about what I wanted to do. StartsWith definitely makes more sense than using Contains.
I need to convert only a starting node to string and this is done before the loop, otherwise I pass a NodePath directly to StartsWith, not sure what this does under the hood.
Outside of path validity checks there is no need to convert the paths to strings, all I need is GetName on the starting node's path.

I think that the code from FocusGrabber is not needed in this case, but I am not sure if I understand it. I think that the code is good enough how it is so I will let it on you to suggest some specific changes.

src/gui_common/dialogs/CustomDialog.cs
DrRadek: After looking at the code again I came to the conclusion that there is nothing that shouldn't allow absolute paths.
So I will just change AllowWindowReorderingRecursion to update after re-entering.

src/gui_common/CustomWindow.cs
hhyyrylainen: I think I've found the reason. The CustomDialog scene needed to have these default values put in it. Then they don't appear in every scene that derives from the dialogs.

I'm doing some fixing up on your PR and will commit and push my changes later today. If I end up being able to fix everything and playtest things I may end up merging this PR as well today.

src/awakening_stage/gui/InventoryScreen.tscn
hhyyrylainen: I refactored the Opened signal to the base window as that's where the Closed signal is also so it makes sense for a symmetrical operation to also exist there.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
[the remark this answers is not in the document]
hhyyrylainen: I opened the issue about this hack: https://github.com/Revolutionary-Games/Thrive/issues/4349

hhyyrylainen:
With the window opening there is a slight bug in the inventory screen. If I place the inventory panel on top of the crafting panel and close and reopen the inventory, then the crafting panel ends up on top. This is why I suggested that the open signal putting things on top could be maybe a later PR as it has a bunch of corner cases like this where it's difficult to get working exactly right.

DrRadek:
> With the window opening there is a slight bug in the inventory screen. If I place the inventory panel on top of the crafting panel and close and reopen the inventory, then the crafting panel ends up on top. This is why I suggested that the open signal putting things on top could be maybe a later PR as it has a bunch of corner cases like this where it's difficult to get working exactly right.

I see, a later PR sounds good to me. Also I will have to take a break from this because of an upcoming exam period so I will take the PR only if it will still be here.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -0,0 +1,381 @@
...
+            return;
+        }
+
+        window.Disconnect(nameof(CustomDialog.Dragged), this, nameof(OnWindowReorder));
+        window.Disconnect(nameof(CustomWindow.Opened), this, nameof(OnWindowReorder));
+
+        var windowSibling = connectedWindows[window];
+        connectedWindows.Remove(window);
+
+        if (!connectedSiblings.Contains(windowSibling))
```
hhyyrylainen, line 192: I think this bit got so refactored that it doesn't make any sense anymore...

As far as I can see this would do absolutely nothing as this checks that something is not in a hash set and then it would remove it from the hash set, but the only way the check passes if it wasn't already in the hash set.

hhyyrylainen: I'll change this line to `if (connectedWindows.All(w => w.Value != windowSibling))` as I assume that's the intended check.

DrRadek: Yes, that's the intended check. Also I think that I made this mistake in the beginning.

hhyyrylainen:
Opened the issue: https://github.com/Revolutionary-Games/Thrive/issues/4350

PR #4329, opened 2023-04-26
Made membranes more accurate

### LINKED ISSUE #4117: Make multihex organelles send better positions to the membrane to improve wrapping

Leinourdian:

This PR makes membranes follow organelle collision shapes more accurately. This is done by scaling up the membrane textures to match UV coordinate calculations better and by making membrane generation work based on hex positions instead of just organelle positions.

This makes overlapping membranes less likely. Organelles poking out is also a bit less likely (myofibril included), but it can still happen in niche situations where the membrane wiggles hard enough. Especially if you have the myofibril close to the edge, because its model is unusually long.

example with myofibril

**Related Issues**

Closes #4117

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Nunez2196 (approved):
Nothing crazy that stuck out to me

src/microbe_stage/Membrane.cs
```diff
@@ -82,17 +83,6 @@ public bool Dirty
     /// </remarks>
     public List<Vector2> OrganellePositions { get; set; } = PreviewMembraneOrganellePositions;
```
hhyyrylainen, line 84: Should this variable be renamed now? Or is this clear enough still. I guess technically each hex position is still an "organelle position". I guess it is fine if no one wants to rename this.

hhyyrylainen (approved):
I've tested this and the change (other than the multihex fix) is pretty minor. I also clarified the code changes a tiny bit on what kind of data now exists. I guess this is fine to merge for this release as some other people also tested this and the visual change is pretty minor.

PR #4339, opened 2023-04-29
Cleanup

GameDungeon:

This was made more to teach me some things than being designed as an actual feature. It's fun though, so I cleaned it up a bit, and am submitting it here.

It adds several screen shader effect options into misc. Grayscale, Color Palleting, and pixelation make them up. 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/Thrive.csproj
+++ b/Thrive.csproj
@@ -294,6 +294,7 @@
     <Compile Include="src\gui_common\charts\line\LineChart.cs" />
     <Compile Include="src\gui_common\charts\line\LineChartData.cs" />
     <Compile Include="src\gui_common\charts\DataPoint.cs" />
+    <Compile Include="src\gui_common\ScreenFilter.cs" />
     <Compile Include="src\gui_common\ActionButton.cs" />
     <Compile Include="src\gui_common\CheatMenu.cs" />
     <Compile Include="src\gui_common\ChemicalEquation.cs" />
@@ -522,6 +523,7 @@
     <Compile Include="src\general\GameWorld.cs" />
     <Compile Include="src\general\WorldGenerationSettings.cs" />
     <Compile Include="src\general\DifficultyPreset.cs" />
+    <Compile Include="src\general\ScreenEffect.cs" />
     <Compile Include="src\microbe_stage\OrganelleLayout.cs" />
     <Compile Include="src\microbe_stage\ProcessSystem.cs" />
     <Compile Include="src\microbe_stage\IProcessable.cs" />
--- a/locale/en.po
+++ b/locale/en.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2023-03-31 16:05+0300\n"
+"POT-Creation-Date: 2023-04-29 17:47-0500\n"
 "PO-Revision-Date: 2023-03-31 15:57+0300\n"
 "Last-Translator: Henri Hyyryläinen <hhyyrylainen@revolutionarygamesstudio.com>\n"
 "Language-Team: English <https://translate.revolutionarygamesstudio.com/projects/thrive/thrive-game/en/>\n"
@@ -19,7 +19,7 @@ msgstr ""
 "X-Generator: Poedit 3.2.2\n"
 "Generated-By: Babel 2.8.0\n"
 
-#: ../src/general/OptionsMenu.tscn:1457
+#: ../src/general/OptionsMenu.tscn:1460
 msgid "2D_MOVEMENT_TYPE_SELECTION"
 msgstr "2D movement style:"
 
@@ -31,7 +31,7 @@ msgstr "3D Editor"
 msgid "3D_MOVEMENT"
 msgstr "3D Movement"
 
-#: ../src/general/OptionsMenu.tscn:1489
+#: ../src/general/OptionsMenu.tscn:1492
 msgid "3D_MOVEMENT_TYPE_SELECTION"
 msgstr "3D movement style:"
 
@@ -71,7 +71,7 @@ msgstr "Action blocked while another is in progress"
 msgid "ACTIVE"
 msgstr "Active"
 
-#: ../src/general/OptionsMenu.tscn:1092
+#: ../src/general/OptionsMenu.tscn:1095
 msgid "ACTIVE_THREAD_COUNT"
 msgstr "Current threads:"
 
@@ -102,7 +102,7 @@ msgid "AEROBIC_NITROGEN_FIXING"
 msgstr "Aerobic Nitrogen Fixation"
 
 #: ../src/late_multicellular_stage/MulticellularHUD.tscn:1400
-#: ../src/microbe_stage/MicrobeStage.tscn:1465
+#: ../src/microbe_stage/MicrobeStage.tscn:1472
 msgid "AGENTS"
 msgstr "Agents"
 
@@ -172,22 +172,22 @@ msgstr ""
 msgid "ALT"
 msgstr "Alt"
 
-#: ../src/general/OptionsMenu.tscn:739
+#: ../src/general/OptionsMenu.tscn:742
 msgid "AMBIANCE_VOLUME"
 msgstr "Ambiance volume"
 
 #: ../simulation_parameters/microbe_stage/compounds.json:17
 #: ../src/late_multicellular_stage/MulticellularHUD.tscn:1177
 #: ../src/microbe_stage/gui/PatchDetailsPanel.tscn:678
-#: ../src/microbe_stage/MicrobeStage.tscn:1207
+#: ../src/microbe_stage/MicrobeStage.tscn:1214
 msgid "AMMONIA"
 msgstr "Ammonia"
 
-#: ../src/general/OptionsMenu.tscn:1634
+#: ../src/general/OptionsMenu.tscn:1637
 msgid "AMOUNT_OF_AUTOSAVE_TO_KEEP"
 msgstr "Amount of autosaves to keep:"
 
-#: ../src/general/OptionsMenu.tscn:1662
+#: ../src/general/OptionsMenu.tscn:1665
 msgid "AMOUNT_OF_QUICKSAVE_TO_KEEP"
 msgstr "Amount of quicksaves to keep:"
 
@@ -212,11 +212,11 @@ msgstr "Apply Changes"
 msgid "APRIL"
 msgstr "April"
 
-#: ../src/general/OptionsMenu.tscn:1997
+#: ../src/general/OptionsMenu.tscn:2040
 msgid "ARE_YOU_SURE_TO_RESET_ALL_SETTINGS"
 msgstr "Are you sure you want to reset ALL settings to default values?"
 
-#: ../src/general/OptionsMenu.tscn:2006
+#: ../src/general/OptionsMenu.tscn:2049
 msgid "ARE_YOU_SURE_TO_RESET_INPUT_SETTINGS"
 msgstr "Are you sure you want to reset the input settings to default values?"
 
@@ -244,7 +244,7 @@ msgstr "Mod assembly class is required when assembly is specified"
 msgid "ASSEMBLY_REQUIRED_WITH_HARMONY"
 msgstr "Mod assembly is required when auto harmony is enabled"
 
-#: ../src/general/OptionsMenu.tscn:1106
+#: ../src/general/OptionsMenu.tscn:1109
 msgid "ASSUME_HYPERTHREADING"
 msgstr "Assume CPU has hyperthreading enabled"
 
@@ -286,7 +286,7 @@ msgstr "The attempt to write this save file has failed. The save name may be too
 msgid "AT_CURSOR"
 msgstr "At Cursor:"
 
-#: ../src/general/OptionsMenu.tscn:856
+#: ../src/general/OptionsMenu.tscn:859
 msgid "AUDIO_OUTPUT_DEVICE"
 msgstr "Audio output device:"
 
@@ -294,8 +294,8 @@ msgstr "Audio output device:"
 msgid "AUGUST"
 msgstr "August"
 
-#: ../src/general/OptionsMenu.tscn:595 ../src/general/OptionsMenu.tscn:596
-#: ../src/general/OptionsMenu.tscn:1474 ../src/general/OptionsMenu.tscn:1475
+#: ../src/general/OptionsMenu.tscn:598 ../src/general/OptionsMenu.tscn:599
+#: ../src/general/OptionsMenu.tscn:1477 ../src/general/OptionsMenu.tscn:1478
 msgid "AUTO"
 msgstr "Auto"
 
@@ -322,7 +322,7 @@ msgstr ""
 msgid "AUTO-EVO_STEPS_DONE"
 msgstr "{0:F1}% done. {1:n0}/{2:n0} steps."
 
-#: ../src/general/OptionsMenu.tscn:1623
+#: ../src/general/OptionsMenu.tscn:1626
 msgid "AUTOSAVE_DURING_THE_GAME"
 msgstr "Autosave during the game"
 
@@ -359,7 +359,7 @@ msgstr "Auto-Evo Status:"
 msgid "AUTO_MOVE_FORWARDS"
 msgstr "Auto move forwards"
 
-#: ../src/general/OptionsMenu.cs:938 ../src/general/OptionsMenu.tscn:358
+#: ../src/general/OptionsMenu.cs:954 ../src/general/OptionsMenu.tscn:361
 msgid "AUTO_RESOLUTION"
 msgstr "Auto ({0}x{1})"
 
@@ -379,7 +379,7 @@ msgstr "Aware Stage"
 #: ../src/general/MainMenu.tscn:458 ../src/general/MainMenu.tscn:554
 #: ../src/general/MainMenu.tscn:610 ../src/general/MainMenu.tscn:933
 #: ../src/general/NewGameSettings.tscn:1142
-#: ../src/general/OptionsMenu.tscn:1934 ../src/general/PauseMenu.tscn:282
+#: ../src/general/OptionsMenu.tscn:1977 ../src/general/PauseMenu.tscn:282
 #: ../src/microbe_stage/gui/ExtinctionBox.tscn:76
 #: ../src/modding/ModManager.tscn:79 ../src/saving/NewSaveMenu.tscn:122
 #: ../src/saving/SaveManagerGUI.tscn:115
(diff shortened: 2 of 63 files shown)


adQuid:
I don't think "cleanup" is the best PR title for this...

GameDungeon:
Oh lol It auto title it

Thrive.csproj
```diff
@@ -758,4 +760,4 @@
   <ItemGroup>
     <ProjectReference Include="third_party\ThirdParty.csproj" />
   </ItemGroup>
-</Project>
+</Project>
```
GameDungeon, line 763:
```suggestion
</Project>

```

shaders/screen_effects/Gameboy.gdshader
```diff
@@ -0,0 +1,22 @@
+shader_type canvas_item;
+
+const float color_count = 4.0f;
+const int pixel_count = 512;
+
+const vec3 pallet[4] = {
```
Nunez2196, line 6: nitpick: misspelled palette

shaders/screen_effects/Gameboy.gdshader
```diff
@@ -0,0 +1,22 @@
...
+    vec3(161.0f, 206.0f, 10.0f) / 255.0f
+};
+
+void fragment()
+{
+    vec2 grid_UV = (vec2(ivec2( SCREEN_UV * float(pixel_count) ))+0.5)/float(pixel_count);
+    vec4 pixelColor = textureLod(SCREEN_TEXTURE, grid_UV, 0);
+    
+    float avg = (pixelColor.r + pixelColor.g + pixelColor.b) / 3.0;
+    int color = int(floor(avg * (color_count - 1.0f) + 0.5f));
```
Nunez2196, line 19: missing some f's after some floats

Nunez2196 (approved):
shaders are fine, just nitpicks / style. nothing ridiculous (like converting to HSV just to de-saturate or do something simple).

Rest of the code, and necessary JSON seems good to go at a glance.

shaders/screen_effects/Gameboy.gdshader
hhyyrylainen, line 1: These new shaders should use the same file extension as the existing shaders in the game.

simulation_parameters/SimulationParameters.cs
```diff
@@ -332,6 +336,21 @@ public IEnumerable<DifficultyPreset> GetAllDifficultyPresets()
...
     }
 
+    public ScreenEffect GetScreenEffect(string name)
+    {
+        return screenEffects[name];
+    }
+
+    public ScreenEffect GetScreenEffectByIndex(int index)
+    {
+        return screenEffects.Values.First(p => p.Index == index);
```
hhyyrylainen, line 346: I'm sure there's a better way to do this than abuse LINQ like this...

src/engine/Settings.cs
```diff
@@ -454,6 +454,9 @@ private Settings()
         Constants.CONTROLLER_DEFAULT_DEADZONE,
     });
 
+    [JsonProperty]
+    public SettingValue<ScreenEffect?> CurrentScreenEffect { get; private set; } = new(null);
```
hhyyrylainen, line 458: This is a new graphics option, right? So this should be grouped with the other graphics options rather than appended at the end here.

src/general/ScreenEffect.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
+using Newtonsoft.Json;
+
+/// <summary>
+///   Definition for a difficulty preset
```
hhyyrylainen, line 5: Very suspicious comment.

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -1591,6 +1598,7 @@ __meta__ = {
 [node name="HUDBottomBar" parent="MicrobeHUD/BottomLeft" instance=ExtResource( 16 )]
 margin_top = -43.0
 margin_bottom = -8.0
+SuicideButtonPath = null
```
hhyyrylainen, line 1601: This seems like a suspicious change.

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -427,6 +428,12 @@ script = ExtResource( 7 )
 margin_left = -0.943848
 margin_right = -0.943848
 
+[node name="ScreenFilter" parent="." instance=ExtResource( 23 )]
```
hhyyrylainen, line 431: Okay so the screen filter only works in the microbe stage? Is there a reason why this wasn't made with the overlay system to apply to the entire game?

Thrive.csproj
```diff
@@ -522,6 +523,7 @@
     <Compile Include="src\general\GameWorld.cs" />
     <Compile Include="src\general\WorldGenerationSettings.cs" />
     <Compile Include="src\general\DifficultyPreset.cs" />
+    <Compile Include="src\general\ScreenEffect.cs" />
```
hhyyrylainen, line 526: I think this should be in the "engine" folder as I think fullscreen colour effects should probably be considered an "engine feature".

src/microbe_stage/MicrobeStage.tscn
GameDungeon: I think in this case, this was just godot being godot, and making random changes to files for no reason. I will revert it though.

simulation_parameters/SimulationParameters.cs
Kasterisk: In this code's defense, the same exact approach is used in other places e.g. the difficulty presets.

GameDungeon: Also, I may or may not have stolen that code from the difficulty presets.

hhyyrylainen: I see now, "index" isn't literally the index here but just a random value that could be whatever so this is actually kind of needed to guard against that. I'd prefer using programming terms with their usual meaning. That would require a bit of a refactor here, though (or renaming Index to Order).

locale/en.po
```diff
@@ -5312,20 +5316,40 @@ msgstr "Saving is not currently possible due to:"
...
 
-#: ../src/general/OptionsMenu.tscn:1485 ../src/general/OptionsMenu.tscn:1518
-#: ../src/general/OptionsMenu.tscn:1519
+#: ../src/general/OptionsMenu.tscn:1939
+msgid "SCREEN_EFFECT"
+msgstr "Screen Effect"
+
+#: ../simulation_parameters/common/screen_effects.json:13
+msgid "SCREEN_EFFECT_GAMEBOY"
+msgstr "Gameboy"
```
hhyyrylainen, line 5337: Just *maybe* this should be abbreviated as "GB" as the trademark on Gameboy might still be valid and messing with the big N for no reason is not the best idea.

shaders/screen_effects/Gameboy.shader
```diff
@@ -0,0 +1,22 @@
...
+const vec3 palette[4] = {
+    vec3(0.0f, 63.0f, 0.0f) / 255.0f,
+    vec3(47.0f, 114.0f, 32.0f) / 255.0f,
+    vec3(140.0f, 191.0f, 10.0f) / 255.0f,
+    vec3(161.0f, 206.0f, 10.0f) / 255.0f
+};
+
+void fragment()
+{
+    vec2 grid_UV = (vec2(ivec2( SCREEN_UV * float(pixel_count) ))+0.5f)/float(pixel_count);
```
hhyyrylainen, line 15: The spacing is inconsistent on this line, spaces around parenthesis should be removed and spaces should be around operators. This same issue is in the other shader files as well.

`(vec2(ivec2(SCREEN_UV * float(pixel_count))) + 0.5f) / float(pixel_count);`

src/engine/ScreenFilter.tscn
```diff
@@ -0,0 +1,12 @@
+[gd_scene load_steps=3 format=2]
+
+[ext_resource path="res://src/engine/ScreenFilter.cs" type="Script" id=1]
+
+[sub_resource type="ShaderMaterial" id=2]
```
hhyyrylainen, line 5: In this case it is likely not a problem but for good practice this should be marked as scene unique resource as I assume you modify the shader from the code and that will absolutely break in the case where multiple instances of this scene were loaded. Which isn't the case here but using the proper way to do here might help someone realize that that's the way to do it.

src/engine/Settings.cs
```diff
@@ -340,6 +340,12 @@ private Settings()
     public SettingValue<JSONDebug.DebugMode> JSONDebugMode { get; private set; } =
         new(JSONDebug.DebugMode.Automatic);
 
+    /// <summary>
+    ///   The screen effect currently being used
+    /// </summary>
+    [JsonProperty]
+    public SettingValue<ScreenEffect?> CurrentScreenEffect { get; private set; } = new(null);
```
hhyyrylainen, line 347: Should the screen effect be in graphics tab? It is after all a graphical thing.

src/engine/Settings.cs
```diff
@@ -478,7 +484,6 @@ private Settings()
     public SettingValue<IReadOnlyList<string>> EnabledMods { get; private set; } = new(new List<string>());
 
     // Computed properties from other settings
-
```
hhyyrylainen, line 481: Please keep this line, in my commenting style the blank line here has implications regarding the scope of the comment.

src/general/OptionsMenu.tscn
```diff
@@ -1916,11 +1919,51 @@ selected = 0
 margin_top = 667.0
 margin_right = 702.0
 margin_bottom = 671.0
+rect_pivot_offset = Vector2( -571, -253 )
```
hhyyrylainen, line 1922: Why is this property set?

src/general/OptionsMenu.cs
```diff
@@ -490,6 +494,15 @@ public override void _Ready()
 
         GetViewport().Connect("size_changed", this, nameof(DisplayResolution));
 
+        var simulationParameters = SimulationParameters.Instance;
```
hhyyrylainen, line 497: This should be its separate method like the other longer setup of menu buttons.

src/engine/ScreenFilter.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    public void UpdateEffect(ScreenEffect? currentEffect)
+    {
+        if (currentEffect?.ShaderPath == null)
+        {
+            ((ShaderMaterial)Material).Shader = null;
+            return;
+        }
+
+        Shader effectShader = (Shader)GD.Load(currentEffect.ShaderPath);
```
hhyyrylainen, line 33:
```suggestion
        var effectShader = GD.Load<Shader>(currentEffect.ShaderPath);
```

src/engine/ScreenFilter.cs
```diff
@@ -0,0 +1,37 @@
...
+        base._ExitTree();
+
+        Settings.Instance.CurrentScreenEffect.OnChanged -= UpdateEffect;
+    }
+
+    public void UpdateEffect(ScreenEffect? currentEffect)
+    {
+        if (currentEffect?.ShaderPath == null)
+        {
+            ((ShaderMaterial)Material).Shader = null;
```
hhyyrylainen, line 29: Why does this not use the `material` variable you setup in _EnterTree?

src/engine/ScreenEffect.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// </summary>
+    [TranslateFrom(nameof(untranslatedName))]
+    public string Name = null!;
+
+#pragma warning disable 169,649 // Used through reflection
+    private string? untranslatedName;
+#pragma warning restore 169,649
+
+    /// <summary>
+    ///   Index for this difficulty preset in the preset menu
```
hhyyrylainen, line 25: Unupdated comment.

locale/en.po
GameDungeon: The trademark is valid and we have 3 options:

1. Ignore it, as here it would probably come to nothing. Also I'm not a lawyer, but this feels [like fair use](https://en.wikipedia.org/wiki/Nominative_use).
2. Write something along the lines of on the bottom of the menu: "Gameboy(R) and Gameboy Color(R) are registered trademarks of Nintendo Of America Inc"
3. Change it to how you suggest

src/general/OptionsMenu.tscn
GameDungeon: Godot being godot again. It really likes changing random things. I could revert, but that does likely do nothing.

src/engine/Settings.cs
GameDungeon: While it is graphical, I was trying to keep more silly options to misc. Your choice though.

src/engine/ScreenFilter.tscn
GameDungeon: I don't know how to do this, in this context

locale/en.po
hhyyrylainen: In emulation scene most people use abbreviations, which I think is the safest route. So if anything is done, that option should be picked.

src/general/OptionsMenu.tscn
hhyyrylainen: If anyone ever applies any rotation or maybe something else can trigger this, someone is going to have a really bad time trying to figure out why that doesn't work on this one node.

src/engine/Settings.cs
hhyyrylainen: I guess then we can put it in the misc tab if there's a chance of there being non-graphical joke options.

src/engine/ScreenFilter.tscn
hhyyrylainen: kuva

check that box in the editor for that material in the scene.

src/engine/ScreenFilter.cs
GameDungeon: It doesn't like it, as material might be null. Cleaned it up.

hhyyrylainen: Well the opposite fix should be done. It's a pattern used everywhere else that you take the resources once when starting and then use them. If something is not retrieved yet / null when it shouldn't be throw an InvalidOperationException.

src/engine/ScreenEffect.cs
```diff
@@ -0,0 +1,59 @@
...
+#pragma warning restore 169,649
+
+    /// <summary>
+    ///   Index for this screen effect in the effect menu
+    /// </summary>
+    [JsonProperty]
+    public int Index { get; private set; }
+
+    [JsonProperty]
+    public string? ShaderPath { get; set; }
```
hhyyrylainen, line 31:
```suggestion
    public string? ShaderPath { get; private set; }
```
Is there a reason why the set is public?

src/general/OptionsMenu.cs
```diff
@@ -269,6 +269,9 @@ public class OptionsMenu : ControlWithInput
     [Export]
     public NodePath ScreenshotDirectoryWarningBoxPath = null!;
 
+    [Export]
+    public NodePath ScreenEffectButtonPath = null!;
```
hhyyrylainen, line 273: Variables should be grouped based on the tab (and after where this is visually).

src/general/OptionsMenu.cs
```diff
@@ -421,6 +424,7 @@ public class OptionsMenu : ControlWithInput
     private CustomCheckBox showNewPatchNotes = null!;
     private Label dismissedNoticeCount = null!;
     private OptionButton jsonDebugMode = null!;
+    private OptionButton screenEffectButton = null!;
```
hhyyrylainen, line 427: Here this grouping looks correct.

src/general/OptionsMenu.tscn
```diff
@@ -1917,10 +1920,49 @@ margin_top = 667.0
...
+text = "SCREEN_EFFECT"
+
+[node name="HSeparator" type="HSeparator" parent="CenterContainer/VBoxContainer/VBoxContainer/VBoxContainer/PanelContainer/Misc/ScrollContainer/MarginContainer/VBoxContainer/HBoxContainer9"]
+margin_left = 254.0
+margin_right = 498.0
+margin_bottom = 25.0
+size_flags_horizontal = 3
+custom_styles/separator = SubResource( 9 )
+
+[node name="ScreenEffect" type="OptionButton" parent="CenterContainer/VBoxContainer/VBoxContainer/VBoxContainer/PanelContainer/Misc/ScrollContainer/MarginContainer/VBoxContainer/HBoxContainer9"]
```
hhyyrylainen, line 1947: It looks like you haven't correctly updated the focus neighbours. JSONDebug still has unupdated `focus_neighbour_bottom` and there's probably other nodes around this control that need the focus to be updated. If they are not then controller and keyboard GUI navigation is broken.

src/general/OptionsMenu.cs
```diff
@@ -2384,4 +2404,12 @@ private void OnOpenPatchNotesPressed()
         patchNotesDisplayer.ShowLatest();
         patchNotesBox.PopupCenteredShrink();
     }
+
+    private void OnScreenEffectSelected(int index)
```
hhyyrylainen, line 2408: These methods should be in the same order as the node variables (so this should be next to the one about JSON debug mode).

hhyyrylainen:
I've now looked through all of the code and I had just small style fixing things to comment on. Once those are fixed I think this'll be fine to merge. As long as someone's tested this and made sure for example that starting the game with an effect enabled works correctly, and resetting graphics options etc. doesn't cause any unexpected issues.

simulation_parameters/common/screen_effects.json
```diff
@@ -0,0 +1,22 @@
...
+    "index": 0,
+    "ShaderPath": null
+  },
+  "greyscale": {
+    "name": "SCREEN_EFFECT_GREYSCALE",
+    "index": 1,
+    "ShaderPath": "res://shaders/screen_effects/Greyscale.shader"
+  },
+  "gameboy": {
+    "name": "SCREEN_EFFECT_GB",
```
hhyyrylainen, line 13: I think the translation key should still have the full name. The filename is after all still the full name. I think it's fine to use here, and if the full name is not used here then a translation hint will need to be added on weblate.

GameDungeon: Well if it is used, a hint not to use the full name will probably also need to be added

hhyyrylainen: I think that's a smaller problem than me having to intervene when translators are clueless and end up translating GB to Great Britain equivalent or something.

hhyyrylainen: I'll do this change and perform final testing unless you really object.

src/general/OptionsMenu.tscn
```diff
@@ -1917,10 +1922,49 @@ margin_top = 667.0
...
+margin_left = 502.0
+margin_right = 702.0
+margin_bottom = 25.0
+rect_min_size = Vector2( 200, 25 )
```
hhyyrylainen, line 1953: This has incorrect bottom focus, this prevents selecting the built at label. I'll fix this myself as I got started on that, but for future reference you should get more familiar with how the GUI focus navigation works.

simulation_parameters/common/screen_effects.json
```diff
@@ -0,0 +1,22 @@
+﻿{
+  "none": {
+    "name": "SCREEN_EFFECT_NONE",
+    "index": 0,
+    "ShaderPath": null
+  },
+  "greyscale": {
+    "name": "SCREEN_EFFECT_GREYSCALE",
+    "index": 1,
+    "ShaderPath": "res://shaders/screen_effects/Greyscale.shader"
```
hhyyrylainen, line 10: This path is wrong and fails on case sensitive filesystems (which I think exported Godot games also use). I'll fix this as well.

hhyyrylainen (approved):
I playtested and fixed the few errors I saw. Also did a few purely style related technical changes. This is now good to go by me.

GameDungeon:
The CI seems to have broken.

PR #4347, opened 2023-05-06
ScreenEffect BugFix

GameDungeon:

This should fix the bug that HH pointed out to me. I will need this tested to make sure I understand what it was and was able to resolve the bug properly.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/OptionsMenu.cs
+++ b/src/general/OptionsMenu.cs
@@ -2280,8 +2280,14 @@ public class OptionsMenu : ControlWithInput
 
     private void OnScreenEffectSelected(int index)
     {
-        Settings.Instance.CurrentScreenEffect.Value =
-            SimulationParameters.Instance.GetScreenEffectByIndex(index);
+        var currentEffect = Settings.Instance.CurrentScreenEffect;
+
+        currentEffect.Value = SimulationParameters.Instance.GetScreenEffectByIndex(index);
+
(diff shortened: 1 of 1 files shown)


src/general/OptionsMenu.cs
```diff
@@ -2280,8 +2280,14 @@ private void OnJSONDebugModeSelected(int index)
 
     private void OnScreenEffectSelected(int index)
     {
-        Settings.Instance.CurrentScreenEffect.Value =
-            SimulationParameters.Instance.GetScreenEffectByIndex(index);
+        var currentEffect = Settings.Instance.CurrentScreenEffect;
```
hhyyrylainen, line 2283: Rather than getting the effect value like this and doing things the unintended way. I'd suggest rewriting this like:

```
var effect = SimulationParameters.Instance.GetScreenEffectByIndex(index);

if (effect == SimulationParameters.Instance.GetScreenEffectByIndex(0))
            effect = null;

Settings.Instance.CurrentScreenEffect.Value = effect;
```

Because assigning the effect's value back to itself like this: `Settings.Instance.CurrentScreenEffect.Value = currentEffect.Value;` doesn't do anything. Or if it does it is actually a bug in the settings system.

hhyyrylainen (changes requested):

hhyyrylainen (approved):
I fixed the problem with multiple sets to the settings value object. Also I comprehensively now tested and I think this is now entirely working correctly.

PR #4358, opened 2023-05-13
Fix intro to new game not fading out.

Kasterisk:

Fixing that one thing that always seem to break time to time and nobody is there to notice it.

Also improved the docs.

**Related Issues**

_None._

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (approved):
Yeah, I hadn't realized this. Probably mostly because I usually have the microbe intro video turned off. I think this is a good enough fix. Maybe if this kind of bug shows up more then it might be worth it to consider making the transitions by default not skip the existing ones, but that would be a bigger change needing to revalidate that all other transitions in the game work correctly.

PR #4361, opened 2023-05-15
Overhaul impact of the nucleus on the microbe

buckly90:
This makes the nucleus provide a percentile bonus to total osmoregulation cost for microbes that possess it.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


simulation_parameters/Constants.cs
```diff
@@ -607,6 +607,7 @@ public static class Constants
     ///   Osmoregulation ATP cost per second per hex
     /// </summary>
     public const float ATP_COST_FOR_OSMOREGULATION = 1.0f;
+    public const float NUCLEUS_OSMOREGULATION_REDUCTION = 0.90f;
```
adQuid, line 610: technically this is used as a multiplier, not a reduction

Deus-Codes:
I tested this in the auto-evo explore tool. Gameplay wise, everything seemed sensible and adjusted. It's a nice idea to allow some mitigation against the pilus if your organism is large enough. I'll try a regular playthrough to see if balancing works pace wise.

What I really liked was seeing so many nucleus-bearing organisms. Somewhere around 20% of cells had the nucleus roughly 25 generations in, which would be something I would have seen previously only with boosted AI mutation rates. 

I'm excited to see this work in tandem with non-linear osmoregulation scaling when that shows up. For now, it seems like we will see even bigger cells on Discord.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1542,7 +1542,7 @@ private void StartAutoEvoPrediction()
     {
         biome ??= Editor.CurrentPatch.Biome;
 
-        var energyBalance = ProcessSystem.ComputeEnergyBalance(organelles, biome, membrane, true,
+        var energyBalance = ProcessSystem.ComputeEnergyBalance(organelles, biome, membrane, true, true,
```
hhyyrylainen, line 1545: Isn't this being always true wrong?

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -1190,6 +1190,9 @@ private void HandleOsmoregulation(float delta)
             osmoregulationCost *= 20.0f / (20.0f + Colony.ColonyMembers.Count);
         }
 
+        if (CellTypeProperties.IsBacteria == false)
```
hhyyrylainen, line 1193: Explicitly checking bools with a bool value is not good style (should be just `if (CellTypeProperties.IsBacteria)`)

src/microbe_stage/Microbe.Contact.cs
```diff
@@ -1439,6 +1439,17 @@ private void OnContactBegin(int bodyID, Node body, int bodyShape, int localShape
                     return;
 
                 var target = otherIsPilus ? thisMicrobe : touchedMicrobe;
+                var prey = otherIsPilus ? touchedMicrobe : thisMicrobe;
```
hhyyrylainen, line 1442: The target variable is already the prey. i.e. it is the cell that didn't have a pilus and is hit.

simulation_parameters/Constants.cs
```diff
@@ -603,11 +603,18 @@ public static class Constants
     /// </summary>
     public const float PILUS_INVULNERABLE_TIME = 0.25f;
 
+    /// <summary>
+    ///   How much larger a cell has to be to ignore the pilus collision of another on contact.
+    /// </summary>
+    public const float PILUS_INEFFECTIVE_RATIO = 2.5f;
```
hhyyrylainen, line 609: Wouldn't it be better to have a progressively less effective pilus with the size difference? Rather than just have a one flat out cutoff where you just can't deal with damage. I can imagine a bunch of scenarios where dealing exactly 0 damage is not going to feel good.

Leinourdian:
The osmoregulation discount feels nice to have. It could perhaps be a bit bolder even, say 80%. But I understand wanting to be conservative, at least while we don't have the scaling cost implemented.

The pilus change works differently from how it was presented. It makes it so the big cell takes a small amount of damage once and then engulfs the small cell, all the while only touching the pilus and not the main body of the cell. I actually kinda like it but it's probably a bit OP.

simulation_parameters/Constants.cs
buckly90: I may have miscommunicated. The pilus always still does damage. It's just that cells can be engulfed despite the presence of a pilus if the engulfer is significantly larger than them. They still cause damage.

This is to address some complaints about predator builds being pretty frustrating once pili become wide-spread, and to make pili less effective against what should be more significant threats. A single bacterium shouldn't be able to tear apart a eukaryote with zero effort.

That being said, I do worry about communicating this, as players are going to be in for a nasty surprise if they attempt to poke a massive cell with a pilus expecting to kill them without issue like before... And of course, the exact cutoff may be unclear to the player.

buckly90:
> The osmoregulation discount feels nice to have. It could perhaps be a bit bolder even, say 80%. But I understand wanting to be conservative, at least while we don't have the scaling cost implemented.
> 
> The pilus change works differently from how it was presented. It makes it so the big cell takes a small amount of damage once and then engulfs the small cell, all the while only touching the pilus and not the main body of the cell. I actually kinda like it but it's probably a bit OP.

What do you think should be done to rectify how powerful the change can be? A higher size requirement? Or some different condition?

src/microbe_stage/editor/CellEditorComponent.cs
buckly90: If I flip this to false, and remove the bool check from the bool in Microbe.Interior, it reverses the functionality and gives bacteria the osmoregulation bonus, and not eukaryotes.

hhyyrylainen: I didn't mean flipping this. Just that this looks like this is the editor prediction, which means that this should use the current editor state to determine if the player is currently bacteria or not.

src/microbe_stage/Microbe.Interior.cs
buckly90: So if I need the boolean to be false, what should I do instead?

An else?

hhyyrylainen: `if (!CellTypeProperties.IsBacteria)`

src/microbe_stage/editor/CellEditorComponent.cs
buckly90: I admit I don't entirely understand how the editor prediction works. I just used the Isplayer boolean as an example until it worked as I intended.

It seems to work perfectly fine, I just assumed that the true is default(?) but I do not know for sure. I'm also not entirely sure how to change it to your liking without breaking it. Sorry....

hhyyrylainen:

src/microbe_stage/editor/CellEditorComponent.cs
buckly90: Ok I *think* I have done as you asked.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -39,7 +39,7 @@ public EnergyBalanceInfo GetEnergyBalanceForSpecies(MicrobeSpecies species, Biom
 
             // Auto-evo uses the average values of compound during the course of a simulated day
             cached = ProcessSystem.ComputeEnergyBalance(species.Organelles, biomeConditions, species.MembraneType,
-                species.PlayerSpecies, worldSettings, CompoundAmountType.Average);
+                species.PlayerSpecies, species.IsBacteria, worldSettings, CompoundAmountType.Average);
```
hhyyrylainen, line 42: Hang on, you added the parameter called "hasNucleus" but here you pass in the variable `IsBacteria` should this bool not be flipped here?

src/microbe_stage/ProcessSystem.cs
```diff
@@ -86,8 +87,15 @@ public ProcessSystem(Node worldRoot)
 
         int hexCount = 0;
 
+        var nucleus = SimulationParameters.Instance.GetOrganelleType("nucleus");
+
         foreach (var organelle in organelles)
         {
+            if (organelle.Definition == nucleus)
+            {
+                hasNucleus = false;
```
hhyyrylainen, line 96: I don't understand why you take in a parameter to determine if this thing has a nucleus or not, and then you compute the value anyway (but only by **unsetting** the bool from true *if* the thing has a nucleus).

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: It's more towards the right direction now, but there's still some very unexpected parts to the change.

src/auto-evo/simulation/SimulationCache.cs
buckly90: Since simulationcache pulls it's info from species information, I assumed it would still use IsBacteria since that is used often in Species files. You're right that it should probably be flipped though.

src/microbe_stage/ProcessSystem.cs
```diff
@@ -72,11 +72,12 @@ public ProcessSystem(Node worldRoot)
     ///   movement organelles are assumed to be inactive in the balance calculation.
     /// </param>
     /// <param name="isPlayerSpecies">Whether this microbe is a member of the player's species</param>
+    /// <param name="hasNucleus">Whether this microbe does not have a nucleus and is considered prokaryotic</param>
```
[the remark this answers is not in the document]
buckly90: Alright I'll change it back

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -39,7 +39,7 @@ public EnergyBalanceInfo GetEnergyBalanceForSpecies(MicrobeSpecies species, Biom
 
             // Auto-evo uses the average values of compound during the course of a simulated day
             cached = ProcessSystem.ComputeEnergyBalance(species.Organelles, biomeConditions, species.MembraneType,
-                species.PlayerSpecies, worldSettings, CompoundAmountType.Average);
+                species.PlayerSpecies, !species.IsBacteria, worldSettings, CompoundAmountType.Average);
```
hhyyrylainen, line 42: The `!` on this line is now wrong because the parameter is `isBacteria` so negating this value when passing it in is now wrong.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1542,7 +1542,7 @@ private void StartAutoEvoPrediction()
     {
         biome ??= Editor.CurrentPatch.Biome;
 
-        var energyBalance = ProcessSystem.ComputeEnergyBalance(organelles, biome, membrane, true,
+        var energyBalance = ProcessSystem.ComputeEnergyBalance(organelles, biome, membrane, true, !HasNucleus,
```
hhyyrylainen, line 1545: Considering the parameters of this method, the nucleus should be detected from the `organelles` readonly collection.

src/microbe_stage/ProcessSystem.cs
```diff
@@ -148,6 +149,11 @@ public ProcessSystem(Node worldRoot)
             result.Osmoregulation *= worldSettings.OsmoregulationMultiplier;
         }
 
+        if (isBacteria == false)
```
hhyyrylainen, line 152:
```suggestion
        if (!isBacteria)
```

src/microbe_stage/editor/CellEditorComponent.cs
buckly90: I'm sorry, could you elaborate on what I need to do here? I'm currently using the bool as called on line 423.
`  public bool HasNucleus => PlacedUniqueOrganelles.Any(d => d == nucleus);`

hhyyrylainen: You need to do the same thing here, basically. As that variable needs to relate to the inputs of this method and not the current editor stage.

Something like:
```
    private void CalculateEnergyAndCompoundBalance(IReadOnlyCollection<OrganelleTemplate> organelles,
        MembraneType membrane, BiomeConditions? biome = null)
    {
        biome ??= Editor.CurrentPatch.Biome;

        bool hasNucleus = organelles.Any(d => d == nucleus);

        var energyBalance = ProcessSystem.ComputeEnergyBalance(organelles, biome, membrane, true, !hasNucleus,
```

hhyyrylainen:
One thing that hasn't been brought up yet is the fact that a new feature the player cannot know about unless they very carefully read patch notes is not very nice. So I think the nucleus tooltip needs to have this info in it. I think that duplicating how the "osmoregulation" etc. stats are shown in the tooltip would be the perfect way. So adding a new line showing like "-10% total osmoregulation cost" (in green) would be good. Additionally it would be super excellent to have the tooltip value be automatically set by the code so that someone can't modify Constants.cs to be out of date with the info shown to the player.

PR #4362, opened 2023-05-16
Added some documentation and error checking against window reorder incorrect usage

hhyyrylainen:
caused by not being early enough in the hierarchy

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/AddWindowReorderingSupportToSiblings.cs
+++ b/src/gui_common/AddWindowReorderingSupportToSiblings.cs
@@ -18,6 +18,10 @@ using Array = Godot.Collections.Array;
 ///     The reason for this is because the name of the node is used when establishing connections. This also means
 ///     that after adding the scene instance it may not be renamed.
 ///   </para>
+///   <para>
+///     WARNING: this node has to be in the hierarchy before any GUI nodes this is going to manage. Otherwise
+///     unregistering errors will be triggered.
+///   </para>
 /// </remarks>
 public class AddWindowReorderingSupportToSiblings : Control
 {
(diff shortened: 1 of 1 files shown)


DrRadek:
This is caused by _ExitTree, a possible solution to this is to wait for next idle_frame before disconnecting stuff there. With this the errors should never be raised, I think, but why not leave them here.

By the way you can ping me on discord if you want to be sure that I will read it.

hhyyrylainen:
> a possible solution to this is to wait for next idle_frame before disconnecting stuff there

That doesn't work when the game is shutting down as there won't be a next frame. And even in normal cases a node deleted from the scene tree gets disposed. That means that trying to do anything would result in object disposed exception if we tried to do anything.

But anyway, I assume my changes to your code seem fine?

hhyyrylainen:
I'm not sure what the ordering of _ExitTree with that signal is. I would expect that the callback and signal is triggered at the same time so it wouldn't solve the problem of things happening out of order. Though to fully know for sure reading the Godot source code, or well I guess testing by printing stuff would also work, would be necessary.

DrRadek:
_ExitTree corresponds to a "tree_exiting" signal which from my understanding is triggered before the "tree_exited" signal (see https://docs.godotengine.org/en/3.6/classes/class_node.html#class-node-method-exit-tree).
I got the same result from my testing and I haven't noticed any unexpected behaviour.

hhyyrylainen:
Ah, I see you meant to use two different signals to get slightly different timings. I'll use the approach currently here to not need to mess with that kind of potential confusion with multiple signals.

PR #4368, opened 2023-05-20
Window fixes

### LINKED ISSUE #4350: Fix multiple popup windows opening at once messing with their previous display order
With the window opening there is a slight bug in the inventory screen. If I place the inventory panel on top of the crafting panel and close and reopen the inventory, then the crafting panel ends up on top. That's a leftover problem after: https://github.com/Revolutionary-Games/Thrive/pull/4308

I think one way to fix this would be to keep a "queue" of windows that have opened and only apply the window reordering at the end of the frame if none of the opened windows was on top. If one of the opened windows is on top nothing should be done to not unnecessarily unset the order of windows that's set by the user.

### LINKED ISSUE #4365: Custom popup close X can become invisible when focused
Just noticed that in some case it seems like the close X can disappear from a window when it is focused:

Focusing another window makes the display appear again:

DrRadek:

Adds a code in a reordering system to count with multiple windows being opened at once, instead of a suggested fix in #4350 I sort the windows by their tree order and then reorder them.
Changes rendering of a close button as suggested in #4365.
Removes some leftover mess from a WindowReorderingSupportTest scene.
I have also noticed that one can drag a window in a place where a close button is so I have added a fix for that too.

**Related Issues**

fixes #4350 
fixes #4365 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/AddWindowReorderingSupportToSiblings.cs
+++ b/src/gui_common/AddWindowReorderingSupportToSiblings.cs
@@ -84,7 +84,7 @@ public class AddWindowReorderingSupportToSiblings : Control
     /// <summary>
     ///   Used to save windows that ore opened at once to preserve their order.
     /// </summary>
-    private readonly List<CustomDialog> openedWindows = new();
+    private readonly List<CustomDialog> justOpenedWindows = new();
 
 #pragma warning disable CA2213
 
@@ -402,27 +402,27 @@ public class AddWindowReorderingSupportToSiblings : Control
     private void OnWindowOpen(CustomDialog window)
     {
         // Tell the system that there is an opened window and wait in case more windows will be opened at once.
-        if (openedWindows.Count == 0)
-            CallDeferred(nameof(ReorderOpenedWindows));
+        if (justOpenedWindows.Count == 0)
+            Invoke.Instance.QueueForObject(ReorderOpenedWindows, this);
 
-        openedWindows.Add(window);
+        justOpenedWindows.Add(window);
     }
 
     private void ReorderOpenedWindows()
     {
         // Sort the windows to make sure they are updated in the right order.
-        openedWindows.Sort(delegate(CustomDialog first, CustomDialog second)
+        justOpenedWindows.Sort((first, second) =>
         {
             return connectedWindows[first].GetIndex().CompareTo(connectedWindows[second].GetIndex());
         });
 
         // Reorder the windows.
-        foreach (CustomDialog window in openedWindows)
+        foreach (CustomDialog window in justOpenedWindows)
         {
             OnWindowReorder(window);
         }
 
-        openedWindows.Clear();
+        justOpenedWindows.Clear();
     }
 
     /// <summary>
--- a/src/gui_common/AddWindowReorderingSupportToSiblings.cs
+++ b/src/gui_common/AddWindowReorderingSupportToSiblings.cs
@@ -433,7 +433,7 @@ public class AddWindowReorderingSupportToSiblings : Control
         }
         catch (Exception e)
         {
-            GD.PrintErr(e);
+            GD.PrintErr($"Exception occurred in {Name} ({this}) in {nameof(ReorderOpenedWindows)}:\n{e}");
 
             // Remove invalid windows
             justOpenedWindows.RemoveAll(window => !IsInstanceValid(window) || !connectedWindows.ContainsKey(window));
--- a/src/gui_common/AddWindowReorderingSupportToSiblings.cs
+++ b/src/gui_common/AddWindowReorderingSupportToSiblings.cs
@@ -82,7 +82,7 @@ public class AddWindowReorderingSupportToSiblings : Control
     private readonly HashSet<Node> connectedSiblings = new();
 
     /// <summary>
-    ///   Used to save windows that ore opened at once to preserve their order.
+    ///   Used to save windows that are opened at once to preserve their order.
     /// </summary>
     private readonly List<CustomDialog> justOpenedWindows = new();
 
@@ -107,6 +107,8 @@ public class AddWindowReorderingSupportToSiblings : Control
 #pragma warning restore CA2213
 
     private bool connectionsEstablished;
+
+    private bool reorderOpenedWindowsQueued;
 
     /// <summary>
     ///   Finds window reordering nodes in ancestors, looks for manual paths if set, otherwise uses automatic search to
@@ -235,6 +237,8 @@ public class AddWindowReorderingSupportToSiblings : Control
             if (topSibling == windowSibling)
                 topSibling = null;
         }
+
+        justOpenedWindows.Remove(window);
     }
 
     protected override void Dispose(bool disposing)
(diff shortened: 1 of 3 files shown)


src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -394,6 +399,32 @@ private void OnWindowReorder(CustomDialog window)
         window.SetAsToplevel(isSetAsToplevel);
     }
 
+    private void OnWindowOpen(CustomDialog window)
+    {
+        // Tell the system that there is an opened window and wait in case more windows will be opened at once.
+        if (openedWindows.Count == 0)
+            CallDeferred(nameof(ReorderOpenedWindows));
```
hhyyrylainen, line 406: This should use instead our custom Invoke queue system.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -394,6 +399,32 @@ private void OnWindowReorder(CustomDialog window)
...
+
+        openedWindows.Add(window);
+    }
+
+    private void ReorderOpenedWindows()
+    {
+        // Sort the windows to make sure they are updated in the right order.
+        openedWindows.Sort(delegate(CustomDialog first, CustomDialog second)
```
hhyyrylainen, line 414: This makes a lambda, right? The syntax with the delegate keyword should not be used. Instead the `=>` syntax should be used.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -81,6 +81,11 @@ public class AddWindowReorderingSupportToSiblings : Control
     /// </remarks>
     private readonly HashSet<Node> connectedSiblings = new();
 
+    /// <summary>
+    ///   Used to save windows that ore opened at once to preserve their order.
+    /// </summary>
+    private readonly List<CustomDialog> openedWindows = new();
```
hhyyrylainen, line 87:
```suggestion
    private readonly List<CustomDialog> justOpenedWindows = new();
```
I think renaming this like this will make it clear what this is about.

hhyyrylainen (changes requested):
I have a few code related comments in one of the files here. Nothing major just some code clarity and uniformity things with the rest of the code base.

src/gui_common/dialogs/CustomDialog.cs
```diff
@@ -311,14 +315,19 @@ public override void _Draw()
         DrawString(titleFont, titlePosition, translatedWindowTitle, titleColor,
             (int)(RectSize.x - customPanel.GetMinimumSize().x));
 
+        var closeButtonRect = closeButton!.GetRect();
+
+        // Draw close button
+        DrawTextureRect(closeButtonTexture, closeButtonRect, false, closeButtonColor);
```
hhyyrylainen, line 321: Actually could you link the relevant issue here and briefly explain that this button is rendered in a custom way due to a bug with it otherwise disappearing sometimes?

hhyyrylainen:
Reading through the changes here on last time, I think I noticed a potential rare issue. If a window opens and then closes the next frame it's going to be problematic. I think the sorting code would throw an exception and crash the game.

Here's my suggestion on how to fix it: add a safety catch around the call to Sort (and if it fails log and just run with the list in the order it is in), and then modify the window removal logic to remove the window from "justOpenedWindows" if it is there. And finally a new bool flag needs to be used instead of justOpenedWindows being empty to queue the window opened check.

DrRadek:
> Reading through the changes here on last time, I think I noticed a potential rare issue. If a window opens and then closes the next frame it's going to be problematic. I think the sorting code would throw an exception and crash the game.

If you mean removing it from a tree then you are right. The reodering system does nothing when a window is closed though.

> Here's my suggestion on how to fix it: add a safety catch around the call to Sort (and if it fails log and just run with the list in the order it is in), and then modify the window removal logic to remove the window from "justOpenedWindows" if it is there. And finally a new bool flag needs to be used instead of justOpenedWindows being empty to queue the window opened check.

Removing the window from justOpenedWindows after disconnecting it prevents it from being in the list when it's already disposed so shouldn't this be enough? To be sure I tried to destroy the window right after opening it and it didn't raise any errors. (Without this check it raises an error)

hhyyrylainen:
> Removing the window from justOpenedWindows after disconnecting it prevents it from being in the list when it's already disposed so shouldn't this be enough?

Probably minimally, isn't that just about the core of what I suggested?
Additionally I suggested making sure the game *cannot* crash due to window opening. And to avoid triggering the justOpenedWindows reordering twice.

Because if the following happens:
- Window is opened
- The method is queued to run
- Window is closed
- Another window is opened
- The method is queued to run again

Isn't it then possible to run the code twice? That's why I suggested adding that one extra control bool here.

So I just suggested 2 instances of extra safety along with what you are saying fixes the core issue.

DrRadek:
It doesn't matter if it's queued multiple times, it will just clear the list and next time it will do nothing, but you are right, a bool is probably better.
Also I can see that the same window can be queued multiple times so I will add a check to not add it again if it's here already (it wouldn't crash, but it's better to not reorder the same window multiple times).

Okay, I will add the safety catch just to be sure. 
> and if it fails log and just run with the list in the order it is in

I think I don't get this, why not remove them and try to sort it again? I mean they will have to be skipped either way to not trigger an exception and not sorting them could get the order wrong.

hhyyrylainen:
> I think I don't get this, why not remove them and try to sort it again? I mean they will have to be skipped either way to not trigger an exception and not sorting them could get the order wrong.

Putting the exception catch inside the lambda for the sort doesn't sound like a good idea. I didn't explicitly say it but I was expecting the catch to be put around the entire sort call. Thus it would be an all-or-nothing situation, either the sort works or it is not done at a..

DrRadek:
I didn't mean it like that. I meant to put the sort into a loop and run the loop again (with removed invalid windows) if an exception is raised (and throw the exception if it is raised again)

hhyyrylainen:
You still would have to have the catch inside the sort lambda, right?

To me it seems a lot of unnecessary complication to try to make the data valid when all I want is that we don't crash the game in that case. It doesn't matter if the windows end up in slightly wrong order.

DrRadek:
> You still would have to have the catch inside the sort lambda, right?

Exception has an `InnerException` to check what was the exception inside the lambda so no.

> To me it seems a lot of unnecessary complication to try to make the data valid when all I want is that we don't crash the game in that case. It doesn't matter if the windows end up in slightly wrong order.

Sure, but I don't mind writing it and I feel more comfortable with the order not getting messed up so can I do it like that?
If you still want a simple solution then I would suggest to not reorder the windows and just clear the list and return if the exception is raised because leaving the invalid windows here means having to check their validity when reordering them which to me doesn't sound good.

hhyyrylainen:
> is raised because leaving the invalid windows here means having to check their validity when reordering them which to me doesn't sound good.

Did my guards against bad windows not fix that case? I thought I found the one place where the reorder blindly just looked into a dictionary.

> Sure, but I don't mind writing it and I feel more comfortable with the order not getting messed up so can I do it like that?

Umm... maybe it is fine? I'd like to keep the error case checking simple because it will not be possible for me to test the error conditions (as they should *never* happen). Worse than the game crashing due to the window reorder would be if there was a bug in the error condition handling code. So a really rare error (that should really never happen) could trigger a rare bug in the error handling. That seems like a worse situation to me... If you don't agree I think I can accept this kind of complicated error handling if  you can test it thoroughly and add enough comments to explain the reasoning with the approach.

DrRadek:
> Did my guards against bad windows not fix that case? I thought I found the one place where the reorder blindly just looked into a dictionary.
That's a part of connecting and disconnecting windows, this works with `OnWindowReorder` which is normally called by the window's signal so there is no need for the check.

> Umm... maybe it is fine? I'd like to keep the error case checking simple because it will not be possible for me to test the error conditions (as they should never happen). Worse than the game crashing due to the window reorder would be if there was a bug in the error condition handling code. So a really rare error (that should really never happen) could trigger a rare bug in the error handling. That seems like a worse situation to me... If you don't agree I think I can accept this kind of complicated error handling if you can test it thoroughly and add enough comments to explain the reasoning with the approach.

Oh, I get it, I think you are right.

So I will go with your approach and I will remove the invalid windows in the catch so this isn't an issue. I was thinking about the other approach with not reordering them and this could lead them to appearing behind other windows.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -394,6 +403,55 @@ private void OnWindowReorder(CustomDialog window)
...
+    {
+        try
+        {
+            // Sort the windows to make sure they are updated in the right order
+            justOpenedWindows.Sort((first, second) =>
+            {
+                return connectedWindows[first].GetIndex().CompareTo(connectedWindows[second].GetIndex());
+            });
+        }
+        catch (InvalidOperationException)
```
Kasterisk, line 434: Might be better to just catch all, that way it's more robust.

src/gui_common/AddWindowReorderingSupportToSiblings.cs
```diff
@@ -394,6 +403,55 @@ private void OnWindowReorder(CustomDialog window)
...
+        {
+            int oldCount = justOpenedWindows.Count;
+
+            // Remove invalid windows
+            justOpenedWindows.RemoveAll(window => !IsInstanceValid(window));
+
+            GD.PrintErr($"{Name} ({this}) attempted to reorder a window that was destroyed" +
+                $" ({oldCount - justOpenedWindows.Count}x)");
```
Kasterisk, line 442: Error message needs to be generalized and print out the exception?

Kasterisk (changes requested):

src/gui_common/AddWindowReorderingSupportToSiblings.cs
DrRadek: This made me think that we should also check if the connectedWindows dictionary contains the windows when an exception is thrown so I will add that too.

DrRadek:
> I didn't have a good test scene anymore for the missing X but I trust that now also works, I don't see from the code why it wouldn't work

You can confirm that it works in the WindowReorderingSupportTest scene. All the windows here are empty which resulted in a lot of missing X and now it works fine.

hhyyrylainen:
I'll take your word on it and just keep an eye out in case the issue happens to come back. I have another PR to finish now.

PR #4373, opened 2023-05-22
Fix microbe background textures without correct repeating

K0chiko:
Next textures was changed. Now it should be seamless
    Thrive_meso0 
    Thrive_ocean0 
    Thrive_tidepool0

      break existing features:
      members reviewing this PR)


hhyyrylainen:
The improved meso0 looks good. But...

It seems like the tidepool background is actually worse with these changes:

kuva

And also the ocean0 background is worse:

kuva

(image from krita).

PR #4379, opened 2023-05-25
Downgraded SharpZipLib to 1.3.3 to fix random save failures

hhyyrylainen:
this is a temporary measure until we can switch to .NET 7 (with Godot 4)

**Related Issues**

There's been way too many people randomly mentioning the save errors over the past few months. And as it still seems we can't switch to Godot 4 yet, I think we have to for the time being use an older compression library version that doesn't have the random bug (it was updated here: https://github.com/Revolutionary-Games/Thrive/pull/4152)

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I disabled the build servers that were recently updated. I'll try merging this to see if the build succeeds on master then.

Issue #3022, opened 2022-01-14
Make proper collision shapes for the game objects


hhyyrylainen:
instead of using hundreds of vertices large shapes that seem mostly to be directly converted from the visual meshes.
![Kuvakaappaus - 2022-01-14 15-37-23](https://user-images.githubusercontent.com/3796411/149526296-a3cab417-4605-4513-a2eb-71ed1bdfb3b1.png)

We can either use primitive shapes or vastly simplified meshes for them (that's why I tagged both modelling and programming).

Kiloku:
It's possible to generate "simplified" versions of the meshes at runtime and use them as collision shapes, using the [CreateConvexShape](https://docs.godotengine.org/en/stable/classes/class_mesh.html#class-mesh-method-create-convex-shape) function from Godot.   
I wish this was more controllable, so there could be varying levels of simplicity, but even then, this would immediately reduce the complexity of collision shapes. 
I did it crudely as an experiment, and it worked in the sense that it definitely used simpler shapes with less polygons, but I don't know if it had a noticeable impact on physics processing time. It also made the collision of large iron chunks a bit too simple, so you could end up colliding with the collision shape despite visually not touching it yet.  
![image](https://user-images.githubusercontent.com/4577144/182928210-97827fb7-4fd3-4998-9309-88f0f885f455.png)

Kertit:
Found this short but useful video for blender artist. 
Explaining how to manually add collision shape to any created mesh https://www.youtube.com/watch?v=rkVT252z1y8

hhyyrylainen:
That video doesn't exactly match our workflow. We just use whatever method to get a mesh instance into the game and that can be converted into a collision shape and saved with the Godot editor as a ".shape" file. Currently most organelles use their *visual* model with hundreds of triangles as their collision shape, which is very terrible.

hhyyrylainen:
Also one thing to consider would be a code approach where we could make a script that when ran would go through all the existing models and do that automatic conversion and then save the results, though that would need more programmer time rather than using something like Blender to create the low-poly collision meshes.

Kertit:
mhhh.. okay I see. 
I admit this is not a very usual way to make good 3D models/collisions. 
The usual way is to model the collision shape in same time we model the visual model. It was not asked/mentionned for all the previous model so yeah now it would require to them all. 
Never heard of a programmer way to generate good collision. that might take a huge time/effort compared to any artist making a very low poly collision shape

[...]

PR #4393, opened 2023-05-29
Testing new collision system

K0chiko:
Test for task https://github.com/Revolutionary-Games/Thrive/issues/3022

I delete old mesh, shape and scene Iron3, then I import new Iron3.gltf file with collision mesh and created scene with it.
If this object ok, other meshes will be recreated same way


hhyyrylainen:
This PR now suffers from the problem that you didn't create a branch in your fork. This now has the old branch which was squash merged as the base, so that's going to cause all kinds of problems.

hhyyrylainen:
Is the gltf file small or large in this PR? We prefer .glb files as they are vastly smaller thanks to being compressed files. But there might be some use case where they should be allowed.

K0chiko:
In the Blender export option writed that this type of file easier for editing, but I didn't notice difference. Next models will be glb

hhyyrylainen:
Okay, thanks for confirming. I'll probably add an automatic check that will reject PRs that contain gltf files and tell people to re-export as glb.

K0chiko:
> Okay, thanks for confirming. I'll probably add an automatic check that will reject PRs that contain gltf files and tell people to re-export as glb.

May be difference is when object have animation or materials. I will search information and if I find something, I'll let you know

hhyyrylainen:
The few .glb files we already had in the game for organelles are actually mostly for animated organelles as that was at the time the way it was figured out how to get those exported.

PR #4397, opened 2023-05-31
Collisions

K0chiko:
Test for task https://github.com/Revolutionary-Games/Thrive/issues/3022

I delete old mesh, shape and scene Iron3, then I import new Iron3.gltf file with collision mesh and created scene with it.
If this object ok, other meshes will be recreated same way


hhyyrylainen:

Issue #3948, opened 2022-12-17
You can click main menu options during the splash screen


Brokemia:
**Important**
Please only submit technical bug reports here. This is not a place to make feature requests
or suggestions. For that please visit our suggestions site:
https://suggestions.revolutionarygamesstudio.com/

Or our community forums for longer discussions
https://community.revolutionarygamesstudio.com/

**Describe the bug**

There's a brief moment right at the end of the splash screen but before the opening plays where you can click the buttons on the main menu, even though you can't see them. It also seems to ignore my sound settings and play the button noise even though it should be muted (although that doesn't matter as long as the first thing is fixed).

**To Reproduce**
Steps to reproduce the behaviour:
1. Open the game
2. Click repeatedly in the middle of the screen where buttons should appear while the splash screen is up. You will hear a button press noise.
3. After the opening animation plays, you should see whatever menu you clicked into already opened.

**Expected behaviour**
It shouldn't let you click the buttons until after the main menu is visible.

**System info (please complete the following information):**
- OS: Windows 10
- Graphics card: OpenGL ES 3.0 Renderer: NVIDIA GeForce GTX 1650/PCIe/SSE2
- Thrive version 0.6.0.0 (commit f7281f0be7815f1f81ed667569062b170668106b)

**Logs**

https://hastebin.com/fuxotonose.log

Kasterisk:
I'm tagging this with `engine` label as splash screen mechanism is out of our control. Though we can probably make it so that the first time the MainMenu is loaded there's a Control covering the whole screen to block mouse input and once the opening cutscene/transition is finished it could be disabled or removed.

hhyyrylainen:
Now with the keyboard focus being grabbed by the buttons by default, this problem is slightly worse. We could workaround this by hiding the entire menu container while showing the video.

hhyyrylainen:
There's now a partial fix with #4424 which as it hides things should basically solve most of this (though the social and news buttons likely can still be interacted with). So this should be now basically solved for all cases. But I'll leave this open in case anyone can still find a way to click the buttons that are technically still clickable.

PR #4424, opened 2023-07-07
Do not grab focus during intro video

Festerdam:

This PR prevents the menu from grabbing focus during the intro video.

As mentioned in https://github.com/Revolutionary-Games/Thrive/issues/3948, one can interact with the menu during the intro video, because the menu's FocusGrabber is visible (ie. the Visible property is set) and thus GUIFocuseSetter decides that the start game button should be focused.

This PR hides the menu buttons during the time the intro video plays to prevent that from happening.

**Related Issues**
Fixes only partially https://github.com/Revolutionary-Games/Thrive/issues/3948, since I was only able to reproduce menu interaction via keyboard, and I will therefore not say that this PR fixes all problems mentioned in that issue (which are engine issues, so couldn't be fixed with this PR anyways).

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/MainMenu.cs
+++ b/src/general/MainMenu.cs
@@ -538,7 +538,7 @@ public class MainMenu : NodeWithInput
         if (menuArray.Count <= 0)
             throw new InvalidOperationException("Main menu has no menus");
 
-        return menus.GetChild<Control>((int)CurrentMenuIndex);
+        return CurrentMenuIndex == uint.MaxValue ? null : menus.GetChild<Control>((int)CurrentMenuIndex);
     }
 
     private void OnMenuBackgroundTypeChanged(bool value)
--- a/src/general/MainMenu.cs
+++ b/src/general/MainMenu.cs
@@ -100,6 +100,9 @@ public class MainMenu : NodeWithInput
     [Export]
     public NodePath ThanksDialogPath = null!;
 
+    [Export]
+    public NodePath MenusPath = null!;
+
 #pragma warning disable CA2213
     private TextureRect background = null!;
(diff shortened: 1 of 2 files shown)


src/general/MainMenu.cs
```diff
@@ -521,6 +524,22 @@ private void SetBackgroundScene(string path)
...
     }
 
+    private Control? GetCurrentMenu()
+    {
+        if (menuArray == null)
+            throw new InvalidOperationException("Main menu has not been initialized");
+        if (menuArray.Count <= 0)
+            throw new InvalidOperationException("Main menu has no menus");
+
+        foreach (Control menu in menuArray)
```
DrRadek, line 534: Can't we just get a parent from first menu and then get a child on index CurrentMenuIndex to avoid this loop (and return null if CurrentMenuIndex is uint.MaxValue)? Or even better, save a reference to the parent (Menus node).

Festerdam: I inspired myself too much in the code that handled menus and forgot that GetChild was a thing. :D
I changed the code as suggested.

DrRadek: I think that you should also add the check for uint.MaxValue because GetChild would throw an exception in that case, but it should return null, I think. (that's my understanding of it, I am not a part of thrive dev team)

Kastsk: Could also have bypassed this method altogether and set the visibility of the menu's root accordingly, but better to keep this in.

Kastsk (approved):

PR #4431, opened 2023-07-23
4335 error for failed save deletion

### LINKED ISSUE #4355: Failure to delete a save should give an error popup
Right now it just gets logged in `SaveHelper.DeleteSave`

JonFazliu:

This PR adds an error message if the save fails to be deleted

**Related Issues**

close #4355

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/saving/SaveHelper.cs
+++ b/src/saving/SaveHelper.cs
@@ -591,7 +591,7 @@ public static class SaveHelper
     [Serializable]
     public class FailedToDeleteSaveException : Exception
     {
-        public FailedToDeleteSaveException() : base()
+        public FailedToDeleteSaveException()    
         {
         }
 
--- a/src/saving/SaveHelper.cs
+++ b/src/saving/SaveHelper.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.Serialization;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Godot;
@@ -311,7 +312,7 @@ public static class SaveHelper
         directory.Remove(finalPath);
 
         if (directory.FileExists(finalPath))
-            GD.PrintErr("Failed to delete: ", finalPath);
+            throw new FailedToDeleteSaveException(finalPath);
     }
 
     public static void DeleteExcessSaves(string nameStartsWith, int maximumCount)
@@ -567,5 +568,24 @@ public static class SaveHelper
     {
         TaskExecutor.Instance.AddTask(new Task(() =>
             DeleteExcessSaves("quick_save", Settings.Instance.MaxQuickSaves)));
+    }
+
+    /// <summary>
+    ///   Thrown when a save file was not deleted
+    /// </summary>
+    [Serializable]
+    public class FailedToDeleteSaveException : Exception
+    {
+        public FailedToDeleteSaveException() : base()
+        {
+        }
+
+        public FailedToDeleteSaveException(string filePath) : base(filePath)
+        {
+        }
+
+        protected FailedToDeleteSaveException(SerializationInfo info, StreamingContext context) : base(info, context)
+        {
+        }
     }
 }
--- a/src/saving/SaveHelper.cs
+++ b/src/saving/SaveHelper.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.Serialization;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Godot;
@@ -312,7 +311,7 @@ public static class SaveHelper
         directory.Remove(finalPath);
 
         if (directory.FileExists(finalPath))
-            throw new FailedToDeleteSaveException(finalPath);
+            throw new System.IO.IOException(finalPath);
     }
 
     public static void DeleteExcessSaves(string nameStartsWith, int maximumCount)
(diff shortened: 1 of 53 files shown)


src/saving/SaveList.cs
```diff
@@ -66,6 +69,7 @@ public class SaveList : ScrollContainer
     private CustomConfirmationDialog upgradeSaveDialog = null!;
     private CustomConfirmationDialog loadIncompatiblePrototypeDialog = null!;
     private ErrorDialog upgradeFailedDialog = null!;
+    private ErrorDialog errorSaveDeletionFailed = null!;
```
Kastsk, line 72: While it's not entirely wrong to use `ErrorDialog`, the actual established practice for displaying errors for this kinds of operation is just to use our plain `CustomConfirmationDialog` for friendlier user interface. For more info, see #3947.

I know this is counter-intuitive but take it as an example of [Maslow's hammer](https://en.wiktionary.org/wiki/if_all_you_have_is_a_hammer,_everything_looks_like_a_nail), I guess.

src/saving/SaveList.cs
```diff
@@ -254,7 +260,16 @@ private void OnConfirmDelete()
...
         GD.Print("Deleting save: ", saveToBeDeleted);
-        SaveHelper.DeleteSave(saveToBeDeleted);
+        try
+        {
+            SaveHelper.DeleteSave(saveToBeDeleted);
+        }
+        catch (SaveHelper.FailedToDeleteSaveException e)
+        {
+            errorSaveDeletionFailed.ShowError(TranslationServer.Translate(
+                "Failed to delete save."), e.Message, e.StackTrace);
```
Kastsk: Instead, show the exception message (containing the file's full path) in the Godot error print, which can be added below this.

Also, displaying the stack trace won't do much as we as programmers already know where and how the exception occurs and the error won't change at all in all of the game's lifetime. Let alone showing it to the player.

src/saving/SaveList.tscn
```diff
@@ -128,6 +129,11 @@ WindowTitle = "SELECTED_SAVE_IS_INCOMPATIBLE_PROTOTYPE_PROMPT"
 HideCancelButton = true
 DialogText = "LOAD_INCOMPATIBLE_PROTOTYPE_WARNING"
 
+[node name="ErrorDialog" parent="MarginContainer/VBoxContainer" instance=ExtResource( 4 )]
+margin_top = 4.0
+margin_right = 1270.0
+margin_bottom = 4.0
```
Kastsk, line 135: The consequence of doing all of the things above is that we have a static error text, hence it's more appropriate to set the error dialog's title and message from the inspector instead of setting it programmatically. With one catch: we fill out a translation key instead of plain English text, for more info see here: [Working With Translations](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/working_with_translations.md).

Kastsk (changes requested):
Welcome and thanks for opening your first PR for Thrive. I have a few comments on the changes.

src/saving/SaveList.cs
Kastsk: Forgot to mention that the exception message (which in this case is a file path), shouldn't also be shown, this is related to the same choices made in #3947.

src/saving/SaveHelper.cs
```diff
@@ -311,7 +312,7 @@ public static void DeleteSave(string saveName)
         directory.Remove(finalPath);
 
         if (directory.FileExists(finalPath))
-            GD.PrintErr("Failed to delete: ", finalPath);
+            throw new FailedToDeleteSaveException(finalPath);
```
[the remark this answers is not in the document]
JonFazliu: That makes sense, I went ahead and made the change

src/saving/SaveList.cs
hhyyrylainen: I see some problems with the translation system, so I'll just drop this link here: https://github.com/Revolutionary-Games/Thrive/blob/master/doc/working_with_translations.md

Kastsk: Already ahead of you https://github.com/Revolutionary-Games/Thrive/pull/4431#discussion_r1271449289 🙂

src/saving/SaveList.tscn
JonFazliu: I changed the ErrorMessages to CustomConfirmationDialog, and added the title and description keys in the en.po file, though I'm a bit worried I broke something when I added them there

Kastsk: It looks fine from a cursory look and looks okay now overall. I'll check back tomorrow (within my timezone).

Kastsk (approved):
Well, this works correctly and as intended in my test, code changes are also good now so I'm approving this. Thanks again for contributing!

src/saving/SaveManagerGUI.cs
```diff
@@ -40,6 +41,9 @@ public class SaveManagerGUI : Control
     [Export]
     public NodePath SaveDirectoryWarningDialogPath = null!;
 
+    [Export]
+    public NodePath ErrorSaveDeletionFailedPath = null!;
```
hhyyrylainen, line 45:
```suggestion
    public NodePath SaveDeletionFailedErrorPath = null!;
```

Ordering the words like this makes it sound natural to my ears. The current word order sounds really clunky to me. I think in the SaveList.cs there's similar kind of non-optimal word order used in the variable name.

hhyyrylainen (changes requested):
The approach seems fine to me but I'd like to see some slight tweaks to the specifics before merging.

lawsonnuland:
@JonFazliu are you planning on continuing this PR?

hhyyrylainen:
It's probably safe to assume now that this won't be continued anytime soon. It should be fine for someone else to take over this PR (note that anyone without write access to the main repo also won't be able to push to this PR branch so a new PR needs to be opened, optionally containing the commits from here as a base).

CoreyHendrey:
Work continued over at:
https://github.com/Revolutionary-Games/Thrive/pull/4485

PR #4443, opened 2023-08-20
AI will now usually choose not to eat what it can't digest

### LINKED ISSUE #4440: Allow AI to consider digestability when scanning for potential prey.
AI predators currently do not consider if they can digest their prey or not when scanning for suitable prey in the area. This leads to predators repeatedly attempting to consume prey that they cannot digest to no avail.

When scanning for prey, AI should have a chance to consider if they can digest it or not (By taking into account cell walls and presence of pre-requisite lysosome enzymes) based off of their behavioral scales. Brave and/or opportunistic AI should be more likely to attempt to consume undigestable prey items.

adQuid:

Adds a rolled chance on target acquisition to probably not engulf things it can't digest.

**Related Issues**

fixes #4440 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -741,7 +741,8 @@ public static class Constants
     public const float AI_ENGULF_INTERVAL = 300;
 
     /// <summary>
-    ///   Probability, rolled at each AI step (which happens very often), that the AI will try to engulf something it can't eat
+    ///   Probability, rolled at each AI step (which happens very often), that the AI will try to engulf something 
+    ///   it can't eat
     /// </summary>
     public const float AI_BAD_ENGULF_CHANCE = 0.15f;
 
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -741,7 +741,7 @@ public static class Constants
     public const float AI_ENGULF_INTERVAL = 300;
 
     /// <summary>
-    ///   Probability, rolled at each AI step (which happens very often), that the AI will try to engulf something 
+    ///   Probability, rolled at each AI step (which happens very often), that the AI will try to engulf something
     ///   it can't eat
     /// </summary>
     public const float AI_BAD_ENGULF_CHANCE = 0.15f;
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -741,7 +741,7 @@ public static class Constants
     public const float AI_ENGULF_INTERVAL = 300;
 
     /// <summary>
-    ///  Probability, rolled at each AI step (which very often), that the AI will try to engulf something it can't eat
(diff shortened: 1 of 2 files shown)


simulation_parameters/Constants.cs
```diff
@@ -740,6 +740,9 @@ public static class Constants
     // Cooldown for AI for toggling engulfing
     public const float AI_ENGULF_INTERVAL = 300;
 
+    // Probability, rolled at each AI step (which very often), that the AI will try to engulf something it can't eat
+    public const float AI_BAD_ENGULF_CHANCE = 0.25f;
```
hhyyrylainen, line 744: Even though other people have commented incorrectly the AI constants, could you make this into a proper XML comment?

buckly90 (approved):
This works just as well as I hoped, great stuff!

I would say that the value for AI_BAD_ENGULF_CHANCE feels a bit too big. Somewhere around 10-15% might be better.

I noticed that cells that chose to attempt engulfing something they couldn't digest would begin to use toxins as well. I assume this behavior normally wouldn't be an issue, but with digestability now part of the equation, hunting with toxins versus engulfing might need to be seperated...
While you are free to try fixing that in this PR, I'll otherwise go ahead and open an issue about this once this PR is merged.

hhyyrylainen:
I think we ought to leave out the more complex change for later, and open an issue about it to not forget about it.

@adQuid is there a reason why you haven't checked the PR workflow first item (`PR author has checked that this PR works as intended and doesn't--`)?

simulation_parameters/Constants.cs
```diff
@@ -740,6 +740,11 @@ public static class Constants
     // Cooldown for AI for toggling engulfing
     public const float AI_ENGULF_INTERVAL = 300;
 
+    /// <summary>
+    ///   Probability, rolled at each AI step (which very often), that the AI will try to engulf something it can't eat
```
hhyyrylainen, line 744: I'm still seeing the "(which very often)" part which doesn't sound grammatically correct to me here.

hhyyrylainen (changes requested):
The eating decision logic is pretty complex but as far as I can parse it in my head it is correct. I noticed one code comment typo, well at least I think it is one, that didn't get fixed. If that's fixed then with buckly's testing I think this is fine to merge.

simulation_parameters/Constants.cs
hhyyrylainen: And now the line is too long :laughing:

hhyyrylainen (approved):

hhyyrylainen:
@buckly90 do you want to open that follow up issue or should I?

hhyyrylainen:
I opened the follow up issue so that I can move on: https://github.com/Revolutionary-Games/Thrive/issues/4445 let me know if I understood incorrectly what it should have.

PR #4448, opened 2023-08-28
Improve auto-evo performance by caching

Maxonovien:

This PR adds further caching for auto-evo computations. Last checks indicated a reduction from 4m 22s to 48s to compute 20 generations.

**Related Issues**

None, but improving auto-evo performance is always good.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/simulation/SimulationCache.cs
+++ b/src/auto-evo/simulation/SimulationCache.cs
@@ -127,7 +127,7 @@
             return cached;
         }
 
-        public float GetEnergyCreationScoreForOrganelle(OrganelleTemplate organelle, BiomeConditions biomeConditions, 
+        public float GetEnergyCreationScoreForOrganelle(OrganelleTemplate organelle, BiomeConditions biomeConditions,
             Compound compound)
         {
             var key = (organelle, biomeConditions, compound);
@@ -187,7 +187,7 @@
         ///   A float to represent score. Scores are only compared against other scores from the same FoodSource,
         ///   so different implementations do not need to worry about scale.
         /// </returns>
-        public float GetEnergyGenerationScoreForSpecies(MicrobeSpecies species, BiomeConditions biomeConditions, 
+        public float GetEnergyGenerationScoreForSpecies(MicrobeSpecies species, BiomeConditions biomeConditions,
             Compound compound)
         {
             var key = (species, biomeConditions, compound);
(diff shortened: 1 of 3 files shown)


src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -23,6 +29,12 @@ public class SimulationCache
         private readonly Dictionary<(TweakedProcess, BiomeConditions), ProcessSpeedInformation> cachedProcessSpeeds =
             new();
 
+        private readonly Dictionary<MicrobeSpecies, float[]> cachedPredationToolsRawScores = new();
+        private readonly Dictionary<(OrganelleTemplate, BiomeConditions, Compound), float>
```
hhyyrylainen, line 33: I think this cache size will blow up pretty easily as each species has one OrganelleTemplate per organelle, and those instances are not equal when new attempted organelles are used. Would caching based on OrganelleDefinition work? Those wouldn't pollute the cache with single use data.

src/auto-evo/simulation/food_source/HeterotrophicFoodSource.cs
```diff
@@ -68,30 +68,12 @@ public HeterotrophicFoodSource(Patch patch, MicrobeSpecies prey, SimulationCache
...
-                    }
-
-                    if (process.Process.Outputs.TryGetValue(mucilage, out var mucilageAmount))
-                    {
-                        mucilageScore += mucilageAmount * Constants.AUTO_EVO_MUCILAGE_PREDATION_SCORE;
-                    }
-                }
-            }
+            // The array was used to have a time & memory efficient structure,
+            // but this could be turned into a class for more readable and less error-prone code.
```
hhyyrylainen, line 72: What about a tuple with named elements? Those are used a lot in Thrive to quickly package a couple of variables into one unit.

src/auto-evo/simulation/SimulationCache.cs
Maxonovien: I think it would. I'll need to have a closer look at it to be sure though.

buckly90:
With this change, most players may not ever have to wait on autoevo during normal gameplay without the help of cheats. Wonderful work!

hhyyrylainen (approved):
Code change looks good. I think the cache size is now not going to blow up. I tested this and I got just about 5% higher memory usage when just using the auto-evo exploring tool. I ran 20 generations for 10 worlds (I got around 670 MB memory usage). And for the performance I got about exactly 50% of the previous auto-evo runtime so I'd say that's a pretty huge improvement (from around 4:48 to 2:27).

PR #4452, opened 2023-08-31
Fix keyboard navigation in input options

### LINKED ISSUE #4447: Moving GUI focus down from "Pan strategic view" option doesn't work (when it is disabled)
Instead focus gets stuck there. This is probably an incorrectly set neighbour focus hint. Hopefully this isn't a bug with an underlying system so I'm marking this as extremely easy.

For reference pressing the down keyboard arrow in this state of the options doesn't select anything (or probably the disabled pan speed option) and pressing it twice will re-select the pan option instead of going to the keybindings:

0HyperCube:
- Prevent keyboard navigation to the pan speed slider when it is disabled (I'm not sure why this isn't Godot's default behaviour)
- Prevent keyboard navigation to the Y sensitivity when it is disabled (for consistency)
- Use Godot's default navigation order around the pan speed slider (which appears to work well)
- Fix changing the sliders with the arrow keys by setting the step to 1 (which is also consistent with other tabs of the menu)

**Related Issues**
Closes #4447

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/OptionsMenu.cs
+++ b/src/general/OptionsMenu.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
@@ -2033,6 +2033,7 @@ public class OptionsMenu : ControlWithInput
     private void OnMouseAxesBoundToggled(bool pressed)
     {
         mouseVerticalSensitivity.Editable = !pressed;
+        mouseVerticalSensitivity.FocusMode = pressed ? FocusModeEnum.Click : FocusModeEnum.All;
 
         if (pressed)
         {
--- a/src/general/OptionsMenu.cs
+++ b/src/general/OptionsMenu.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
@@ -2197,6 +2197,7 @@ public class OptionsMenu : ControlWithInput
     {
         Settings.Instance.PanStrategyViewWithMouse.Value = pressed;
         mouseEdgePanSensitivity.Editable = pressed;
+        mouseEdgePanSensitivity.FocusMode = pressed ? FocusModeEnum.All : FocusModeEnum.Click;
 
         UpdateResetSaveButtonState();
     }
--- a/src/general/OptionsMenu.cs
+++ b/src/general/OptionsMenu.cs
@@ -803,7 +803,8 @@ public class OptionsMenu : ControlWithInput
         mouseHorizontalSensitivity.Value = MouseInputSensitivityToBarValue(settings.HorizontalMouseLookSensitivity);
         mouseHorizontalInverted.Pressed = settings.InvertHorizontalMouseLook;
         mouseVerticalSensitivity.Editable = !mouseAxisSensitivitiesBound.Pressed;
-        mouseVerticalSensitivity.FocusMode = mouseAxisSensitivitiesBound.Pressed ? FocusModeEnum.Click : FocusModeEnum.All;
+        mouseVerticalSensitivity.FocusMode =
+            mouseAxisSensitivitiesBound.Pressed ? FocusModeEnum.Click : FocusModeEnum.All;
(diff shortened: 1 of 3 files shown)


buckly90:
The fix works! Great job.
Though now there's a weird issue with jumping across the entire list of settings when navigating past the pan speed bar. 
Not sure if this issue was present prior to the fix or not though, as you can't really experience it without the fix.

0HyperCube:
@buckly90 Thanks.

The container with all of the key bindings in is focusable (I haven't changed it), and when it is focused Godot scrolls to the bottom of it because "Follow focus" is enabled on the scroll container. Should I disable focusing the container as it doesn't really seem to be beneficial (so the next focus target is the button for the A key)?

0HyperCube:
@hhyyrylainen I think I managed to fix this without having to modify the focus setting system.

hhyyrylainen:
Testing this I now remember what issue I had before. Moving into the movement section now selects the "right-most" keybinding for "Move left" not the left-most input which is A. See this screenshot:




And going down from there correctly jumps to the left most input for "move right" *and* when going back up "A" gets selected. So this non-uniformity of navigation and unintuitive initial control selection was what I noticed when I originally tried to fix this.

I now remember that I determined this to be a bigger bug than the scroll jumping all the way to the bottom before going back to the top. I feel like maybe the Godot engine would need to have an option to more intelligently scroll too tall items into view (make sure the top edge is visible instead of the bottom edge if it can't fit entirely on screen). Does anyone know if there's an open feature suggestion to Godot.

src/general/OptionsMenu.cs
```diff
@@ -2033,6 +2033,7 @@ private void OnMaxSpawnedEntitiesSelected(int index)
     private void OnMouseAxesBoundToggled(bool pressed)
     {
         mouseVerticalSensitivity.Editable = !pressed;
+        mouseVerticalSensitivity.FocusMode = pressed ? FocusModeEnum.Click : FocusModeEnum.All;
```
hhyyrylainen, line 2036: While this probably works, could you also apply this state in the initial settings apply? Similarly to how I think I've remembered to do things for all of the settings. That's because when we eventually manage to fix the options menu causing extra control change callbacks on opening, this approach will stop working correctly.

0HyperCube: Sure

0HyperCube:
Another issue I noticed is that shift + tab from the first key goes to the first key of the next row up (whereas I would have expected focus to go to the last key of the next row up). Should I have a look at fixing this?

0HyperCube:
@hhyyrylainen Is there a situation where the default behaviour for `FocusPrevious` doesn't work? The default behaviour is to flatten the hierarchy and then iterate up from the currently focused node until another focusable node is found.

hhyyrylainen:
If that's the behaviour then it's not going to work most of the time as the hierarchy is like this:

```
- input row
| - Key input
| - Key input 2
- Another input row
| - key
```

Because going from another key should only reach "input row" and not "Key input ". The right behaviour would be to find the sibling of a parent node that has a focusable child *and* then going to the last focusable descendant of that parent's sibling node.

0HyperCube:
> ```
> - input row
> | - Key input
> | - Key input 2
> - Another input row
> | - key
> ```

If I have "key" focused and I press shift+tab then (with my last commit using the default focus order) I focus "Key input 2". Is this what you intended?

hhyyrylainen:
I did just test it and that is the behaviour, which is the optimal case I described. If Godot used the logic as you described it wouldn't work and would focus the wrong node.

I'm fine with it working, but could you put a comment explaining the reasoning in `FocusFlowDynamicChildrenHelper` why that one line was removed (and maybe leave that one line commented out there for future reference)? I couldn't get any weird behaviour in the deadzone calibration, but I don't know if I originally added that code in case there's ever so many axes that they don't fit on one line.

hhyyrylainen (approved):
I think the code is fine now, I didn't do a third re-test though. I'll give some more time for Buckly to also retest this before merging.

buckly90:
Much better. You've fixed the issue perfectly from what I experienced. It feels great.

Now the only remaining quirk is that going up from the buttons will launch you all the way to the top of the settings, but this is quite minor.

PR #4458, opened 2023-09-04
Add vacuole upgrade to store double of only one compound

### LINKED ISSUE #4109: Add vacuole upgrade to double its storage but be only able to store one type
base work towards this has been done so it shouldn't be the hardest thing to make different compound types have different storage capacities, though this might have unexpected bugs and need fixing in many places as such I'm marking this as a difficult issue.

ender-titan1:

This PR modifies CompoundBag to enable it to handle separate capacities for each compound,  as well as adding an upgrade to the vacuole to enable it to have double the capacity but only be able to handle one type of compound.

**Related Issues**

Closes #4109 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/late_multicellular_stage/MulticellularCreature.cs
+++ b/src/late_multicellular_stage/MulticellularCreature.cs
@@ -539,7 +539,9 @@ public class MulticellularCreature : RigidBody, ISpawned, IProcessable, ISaveLoa
             case InteractionType.Construct:
             {
                 if (target is IConstructable
-                        { Completed: false, HasRequiredResourcesToConstruct: true } constructable)
+                    {
+                        Completed: false, HasRequiredResourcesToConstruct: true
+                    } constructable)
                 {
                     // Start action for constructing, the action when finished will pick what it does based on the
                     // target entity
--- a/src/late_multicellular_stage/MulticellularCreature.cs
+++ b/src/late_multicellular_stage/MulticellularCreature.cs
@@ -249,7 +249,7 @@ public class MulticellularCreature : RigidBody, ISpawned, IProcessable, ISaveLoa
         Species = lateSpecies;
 
         // TODO: set from species
-        compounds.SetCapacityForAllCompounds(100);
+        compounds.SetNominalCapacity(100);
 
         // TODO: better mass calculation
         Mass = lateSpecies.BodyLayout.Sum(m => m.Size * m.CellType.TotalMass);
@@ -489,84 +489,84 @@ public class MulticellularCreature : RigidBody, ISpawned, IProcessable, ISaveLoa
             case InteractionType.Harvest:
                 return this.HarvestEntity(target);
             case InteractionType.Craft:
-            {
-                if (RequestCraftingInterfaceFor == null)
                 {
-                    // AI should directly use the crafting methods to create the crafter products
-                    GD.PrintErr(
-                        $"Only player creature can open crafting ({nameof(RequestCraftingInterfaceFor)} is unset)");
-                    return false;
-                }
-
-                // Request the crafting interface to be opened with the target pre-selected
-                RequestCraftingInterfaceFor.Invoke(this, target);
-                return true;
-            }
-
-            case InteractionType.DepositResources:
-            {
-                // TODO: instead of closing, just update the interaction popup to allow finishing construction
-                // immediately
-                if (target is IAcceptsResourceDeposit deposit)
-                {
-                    if (!deposit.AutoTakesResources)
+                    if (RequestCraftingInterfaceFor == null)
                     {
-                        // TODO: allow selecting items
-                        GD.Print("TODO: selecting items to deposit interface is not done");
+                        // AI should directly use the crafting methods to create the crafter products
+                        GD.PrintErr(
+                            $"Only player creature can open crafting ({nameof(RequestCraftingInterfaceFor)} is unset)");
+                        return false;
                     }
 
-                    var itemsToDeposit = deposit.GetWantedItems(this);
-
-                    if (itemsToDeposit != null)
-                    {
-                        var slots = itemsToDeposit.ToList();
-                        deposit.DepositItems(slots.Select(i => i.ContainedItem).WhereNotNull());
-
-                        foreach (var slot in slots)
-                        {
-                            if (!DeleteItem(slot))
-                                GD.PrintErr("Failed to delete deposited item");
-                        }
-
-                        return true;
-                    }
-                }
-
-                GD.PrintErr("Deposit action failed due to bad target or currently held items");
-                return false;
-            }
-
-            case InteractionType.Construct:
-            {
-                if (target is IConstructable { Completed: false, HasRequiredResourcesToConstruct: true } constructable)
-                {
-                    // Start action for constructing, the action when finished will pick what it does based on the
-                    // target entity
-                    StartAction(target, constructable.TimedActionDuration);
+                    // Request the crafting interface to be opened with the target pre-selected
+                    RequestCraftingInterfaceFor.Invoke(this, target);
                     return true;
                 }
 
-                return false;
-            }
-
-            default:
-            {
-                // This might be an extra interaction
-                var extraInteractions = target.GetExtraAvailableActions();
-
-                if (extraInteractions != null)
+            case InteractionType.DepositResources:
                 {
-                    foreach (var (extraInteraction, _) in extraInteractions)
+                    // TODO: instead of closing, just update the interaction popup to allow finishing construction
+                    // immediately
+                    if (target is IAcceptsResourceDeposit deposit)
                     {
-                        if (extraInteraction == interactionType)
-                            return target.PerformExtraAction(extraInteraction);
+                        if (!deposit.AutoTakesResources)
+                        {
+                            // TODO: allow selecting items
+                            GD.Print("TODO: selecting items to deposit interface is not done");
+                        }
+
+                        var itemsToDeposit = deposit.GetWantedItems(this);
+
+                        if (itemsToDeposit != null)
+                        {
+                            var slots = itemsToDeposit.ToList();
+                            deposit.DepositItems(slots.Select(i => i.ContainedItem).WhereNotNull());
+
+                            foreach (var slot in slots)
+                            {
+                                if (!DeleteItem(slot))
+                                    GD.PrintErr("Failed to delete deposited item");
+                            }
+
+                            return true;
+                        }
                     }
+
+                    GD.PrintErr("Deposit action failed due to bad target or currently held items");
+                    return false;
                 }
 
-                // Unknown action type and not an extra action provided by the target
-                GD.PrintErr($"Unimplemented action handling for {interactionType}");
-                return false;
-            }
+            case InteractionType.Construct:
+                {
+                    if (target is IConstructable { Completed: false, HasRequiredResourcesToConstruct: true } constructable)
+                    {
+                        // Start action for constructing, the action when finished will pick what it does based on the
+                        // target entity
+                        StartAction(target, constructable.TimedActionDuration);
+                        return true;
+                    }
+
+                    return false;
+                }
+
+            default:
+                {
+                    // This might be an extra interaction
+                    var extraInteractions = target.GetExtraAvailableActions();
+
+                    if (extraInteractions != null)
+                    {
+                        foreach (var (extraInteraction, _) in extraInteractions)
+                        {
+                            if (extraInteraction == interactionType)
+                                return target.PerformExtraAction(extraInteraction);
+                        }
+                    }
+
+                    // Unknown action type and not an extra action provided by the target
+                    GD.PrintErr($"Unimplemented action handling for {interactionType}");
+                    return false;
+                }
         }
     }
 
--- a/src/late_multicellular_stage/MulticellularCreature.cs
+++ b/src/late_multicellular_stage/MulticellularCreature.cs
@@ -249,7 +249,7 @@ public class MulticellularCreature : RigidBody, ISpawned, IProcessable, ISaveLoa
         Species = lateSpecies;
 
         // TODO: set from species
-        compounds.SetNominalCapacity(100);
+        compounds.NominalCapacity = 100;
 
         // TODO: better mass calculation
         Mass = lateSpecies.BodyLayout.Sum(m => m.Size * m.CellType.TotalMass);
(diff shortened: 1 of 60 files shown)


src/microbe_stage/CompoundBag.cs
```diff
@@ -13,20 +13,23 @@ public class CompoundBag : ICompoundStorage
...
+        SetCapacityForAllCompounds(capacity);
     }
 
     /// <summary>
     ///   How much of each compound this bag can store.
-    ///   Currently a CompoundBag can hold the same amount of each compound.
     /// </summary>
     [JsonProperty]
-    public float Capacity { get; set; }
+    public Dictionary<Compound, float> CompoundCapacities { get; private set; } = new();
```
hhyyrylainen, line 32: The internal details of how the capacity is stored should not be leaked. This should be made private. I created the `GetCapacityForCompound` method for a reason. And second problem is that if a mod for example adds more compounds or we add more compounds in a later update, this is so going to blow up and cause a game crash somewhere.

So here's my suggestion have a "nominal capacity" private field and a nullable dictionary of specific capacities also a private field. Then modify the GetCapacityForCompound to first check if a specific capacity exists for a compound, and if it does return it, otherwise return the nominal capacity. I think with those changes we'd have a design that I can think of being good enough.

ender-titan1: Ok, on it

ender-titan1:
I've implemented the requested changes, however, these changes created a problem where microbes would spawn without any initial compounds. I have fixed this in a somewhat hacky way and still do not know what exactly was causing the problem. I also wonder if `organellesCapacity` is still required in `Microbe.Internal.cs` since the capacity field had been made private.

hhyyrylainen:
Regarding the bug with initial compounds, I've been converting that code over to a new system on my branch this week and I remember reading comments that the initial compound add used a direct add of compounds. That way it bypasses any normal limits or usefulness check. Could you modify your code to go back to that way of doing that to not need the workaround?

ender-titan1:
So if I understand correctly, I should create a new method in `CompoundBag` to add compounds without checking capacity

hhyyrylainen:
Can your work wait until my branch is merged as I already wrote this:
```
    public void AddInitialCompounds(IReadOnlyDictionary<Compound, float> compounds)
    {
        foreach (var entry in compounds)
        {
            if (!Compounds.TryGetValue(entry.Key, out var existingAmount))
            {
                Compounds[entry.Key] = entry.Value;
                continue;
            }

            float toAdd = entry.Value - existingAmount;

            if (toAdd > 0)
                Compounds[entry.Key] = existingAmount + toAdd;
        }
    }
```

doing it differently twice will lead to nice headache merge conflicts for me...

If I remember right didn't the initial compounds just use the `Compounds.Compounds` dictionary directly to set the values? That just goes past any limits etc.

ender-titan1:
Yes, I can wait until that branch has been merged

hhyyrylainen:
If wait until then you'll need to redo Microbe.Interior.cs, MicrobeAI.cs and ProcessSystem.cs changes (or solve very difficult merge conflicts as most of those files will be gone or moved).

I think something like:
```
// Temporary code before proper initial compounds set method is added to CompoundBag
Compounds.Compounds.TryGetValue(compound, out var existing);
Compounds.Compounds[compound] = existing + stuffToAdd;
```
Would be simple enough and this PR could move forward without having to wait a month, which is my maybe a realist estimate on when my work would be completed.

ender-titan1:
I don't quite understand what you mean by "redo changes". Should I revert those files back to the original? And the code provided is meant to be inserted in the `SetInitialCompounds` method if I am to understand correctly.

hhyyrylainen:
I'm just warning that as I'm making major changes, when those changes are merged to master that will lead to a large number of merge conflicts in your PR. So completing the functionality of the PR now with that temporary change will be less work for you. And then of course cleaning up the changes a bit to get this ready for merge, as I noticed at least a few places I'm going to comment on the design approach once this is functionally ready.

But if you are fine with that, we can get back to this PR in a month and then get this ready by fixing all the new merge conflicts with master.

src/microbe_stage/organelle_components/StorageComponent.cs
```diff
@@ -29,3 +29,22 @@ public void Check(string name)
...
+[JSONDynamicTypeAllowed]
+public class StorageComponentUpgrades : IComponentSpecificUpgrades
+{
+    public StorageComponentUpgrades(bool isSpecialized, Compound specialization)
+    {
+        Specialization = specialization;
+        IsSpecialized = isSpecialized;
+    }
+
+    public Compound Specialization { get; set; }
```
hhyyrylainen, line 42: Is there a reason for using two variables? I'd just make one property `Compound? SpecializedFor` and make it so that null means no specialization and then a valid compound value is specialization for that compound.

src/late_multicellular_stage/MulticellularCreature.cs
```diff
@@ -489,84 +489,85 @@ public bool AttemptInteraction(IInteractableEntity target, InteractionType inter
...
             case InteractionType.Craft:
-            {
-                if (RequestCraftingInterfaceFor == null)
                 {
-                    // AI should directly use the crafting methods to create the crafter products
-                    GD.PrintErr(
-                        $"Only player creature can open crafting ({nameof(RequestCraftingInterfaceFor)} is unset)");
-                    return false;
-                }
+                    if (RequestCraftingInterfaceFor == null)
```
hhyyrylainen, line 493: Looks like the indentation got messed up in this file.

src/microbe_stage/CompoundBag.cs
```diff
@@ -12,21 +12,16 @@ public class CompoundBag : ICompoundStorage
...
-    /// <summary>
-    ///   Creates a new bag
-    /// </summary>
-    /// <param name="capacity">Specifies the initial capacity of the compound bag</param>
-    public CompoundBag(float capacity)
-    {
-        Capacity = capacity;
-    }
+    [JsonProperty]
+    private float nominalCapacity;
```
hhyyrylainen, line 16: I don't see why this would need to be a field. Can't this be a property? That avoid having a "useless" manually written setter for the nominal capacity.

src/microbe_stage/CompoundBag.cs
```diff
@@ -38,13 +33,30 @@ public CompoundBag(float capacity)
...
 
-        return 0;
+        if (compoundCapacities != null && compoundCapacities.TryGetValue(compound, out var capacity))
+            return capacity;
+
+        return nominalCapacity;
+    }
+
+    /// <summary>
+    ///   Sets the capacity for all compounds
```
hhyyrylainen, line 49: I think this comment needs to elaborate that this is the capacity for all compounds that don't have a specific capacity set.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using AngleSharp.Dom;
```
hhyyrylainen, line 4: I'm pretty sure this is a mistake, the microbe code should not be doing any HTML parsing...

hhyyrylainen:
I think you hit a bunch of warnings that were hiding underneath the surface on master some time ago. I think those should go away if you merge master into your branch.

src/microbe_stage/organelle_components/StorageComponent.cs
```diff
@@ -29,3 +29,19 @@ public void Check(string name)
...
+
+[JSONDynamicTypeAllowed]
+public class StorageComponentUpgrades : IComponentSpecificUpgrades
+{
+    public StorageComponentUpgrades(Compound? specializationFor)
+    {
+        SpecializationFor = specializationFor;
+    }
+
+    public Compound? SpecializationFor { get; set; }
```
hhyyrylainen, line 41: This doesn't sound grammatically correct to me, is it? To me `SpecializedFor` would sound grammatically correct when reading code aloud that uses this property. Would that be better?

src/microbe_stage/CompoundBag.cs
```diff
@@ -12,21 +12,20 @@ public class CompoundBag : ICompoundStorage
...
+    public CompoundBag(float nominalCapacity)
     {
-        Capacity = capacity;
+        NominalCapacity = nominalCapacity;
     }
 
     /// <summary>
-    ///   How much of each compound this bag can store.
-    ///   Currently a CompoundBag can hold the same amount of each compound.
+    ///   Stores the default capacity for all compounds that do
```
hhyyrylainen, line 24:
```suggestion
    ///   Specifies the default capacity for all compounds that do
```
As this variable doesn't "store" the data just specifies how much something can be stored elsewhere.

src/microbe_stage/CompoundBag.cs
```diff
@@ -38,13 +37,25 @@ public CompoundBag(float capacity)
     /// <summary>
     ///   Gets the capacity for a given compound
     /// </summary>
-    /// <returns>Returns <see cref="Capacity"/> if the compound is useful, otherwise 0</returns>
+    /// <returns>
+    ///   Returns compoundCapacities[compound] if the compound has a specific capacity set,
+    ///   or nominalCapacity if the compound is useful, otherwise 0
```
hhyyrylainen, line 42:
```suggestion
    ///   Returns the capacity this bag has for storing the compound if it is useful, otherwise 0
```
I'd omit the internal details, especially the non `<see ...` type reference to a private variable.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -1323,16 +1332,58 @@ private void OnOrganelleRemoved(PlacedOrganelle organelle)
...
-        organellesCapacity = organelles!.Sum(o => o.StorageCapacity);
-        Compounds.Capacity = organellesCapacity;
+        foreach (PlacedOrganelle organelle in organelles!)
+            UpdateCapacity(organelle, false);
+
+        UpdateCompoundBagCapacities();
+    }
+
+    /// <summary>
+    ///   Updates <see cref="organellesCapacity"/>
```
hhyyrylainen, line 1350: I'm noticing that you are wrapping the documentation text really tightly? Is there a reason? The documentation comments should be filled all the way to column 120.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -1323,16 +1332,58 @@ private void OnOrganelleRemoved(PlacedOrganelle organelle)
...
+    ///   Updates <see cref="Compounds"/> to adjust for changes in
+    ///   <see cref="organellesCapacity"/> and
+    ///   <see cref="additionalCompoundCapacities"/>
+    /// </summary>
+    private void UpdateCompoundBagCapacities()
+    {
+        Compounds.NominalCapacity = organellesCapacity;
+
+        foreach (var entry in additionalCompoundCapacities)
+            Compounds.SetCapacityForCompound(entry.Key, entry.Value + organellesCapacity);
```
hhyyrylainen, line 1386: This doesn't remove special capacities that are now gone from the cell (for example if the player switches a vacuole from glucose to ammonia, they get the extra capacity for both as the old value is never seemingly removed). There needs to be a method in the compound bag to reset the special capacities before they are re-applied to ensure no outdated values exist.

src/microbe_stage/editor/upgrades/VacuoleUpgradeGUI.cs
```diff
@@ -0,0 +1,141 @@
...
+        description = GetNode<Label>(CompoundDescriptionPath);
+        isSpecializedCheckbox = GetNode<CheckBox>(IsSpecializedCheckboxPath);
+        compoundSelection = GetNode<VBoxContainer>(CompoundSelectionPath);
+
+        compounds.Clear();
+    }
+
+    public void OnStartFor(OrganelleTemplate organelle, GameProperties currentGame)
+    {
+        shownChoices = SimulationParameters.Instance.GetAllCompounds().Values.ToList();
```
hhyyrylainen, line 40: I don't see you modifying the shown choices so you should be able to just use the returned IReadOnlyList directly here instead of taking a copy of the data.

src/microbe_stage/editor/upgrades/VacuoleUpgradeGUI.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            if (compound.IsAgent || compound.IsEnvironmental)
+                continue;
+
+            compounds.AddItem(compound.Name);
+        }
+
+        // Select glucose by default
+        var defaultCompoundIndex =
+            shownChoices.FindIndex(c => c.InternalName == "glucose");
```
hhyyrylainen, line 52: Don't the other upgrade GUIs have the default value stored in Constants? This should also be in constants and named something like `DEFAULT_VACUOLE_UPGRADE_FOR_COMPOUND = "glucose"`.

src/microbe_stage/editor/upgrades/VacuoleUpgradeGUI.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        // Apply current upgrade values or defaults
+        if (organelle.Upgrades?.CustomUpgradeData is StorageComponentUpgrades configuration)
+        {
+            Compound? specialization = shownChoices.Find(c => c == configuration.SpecializationFor);
+            isSpecializedCheckbox.Pressed = specialization != null;
+
+            compounds.Selected = specialization != null ?
+                shownChoices.IndexOf(specialization) :
+                shownChoices.FindIndex(c => c.InternalName == "glucose");
```
hhyyrylainen, line 65: Shouldn't this be `defaultCompoundIndex`? to not have to find the same index again.

hhyyrylainen:
A change was just merged that requires adding a comparison method for the upgrade data: https://github.com/Revolutionary-Games/Thrive/pull/4463 so I suspect that that'll need a small tweak to this PR to make sure this compiles (with latest master).

src/microbe_stage/Microbe.Interior.cs
ender-titan1: Regarding this, the feature still works as intended. This is because `additionalCompoundCapacities` for in your example glucose is set to zero when the capacity is begin calculated. This means that while the entry exists in the dictionary, the value is zero. I have also play-tested this and the feature works as intended.

src/microbe_stage/editor/upgrades/VacuoleUpgradeGUI.cs
ender-titan1: Forgive me but I do not understand, `GetAllCompounds` returns a dictionary and the `Values` property returns a ValueCollection that I cannot convert to a list. I am a bit confused and ask for clarification.

src/microbe_stage/Microbe.Interior.cs
hhyyrylainen: Ah, you mean with `UpdateCapacity`? I guess that is true, though with the way I'm refactoring the microbe stage the "organelle removed" callback will be entirely removed, so could you futureproof your code by doing that my suggestion? That way I don't accidentally introduce this bug when trying to fix merge conflicts with my branch.

src/microbe_stage/editor/upgrades/VacuoleUpgradeGUI.cs
hhyyrylainen: My mistake, I thought GetAllCompounds would return a `IReadOnlyList<Compound>`. Alright then this is a fine approach without any changes.

RevolutionaryGamesCommon
hhyyrylainen, line 1: Does this PR accidentally downgrade the common module version? It seems like this is moving it back to version `f1b5b669dbce77e203bc001bc662d2d549b67a40` even though it was newer version before on the master branch.

src/microbe_stage/Microbe.Interior.cs
```diff
@@ -1323,16 +1331,55 @@ private void OnOrganelleRemoved(PlacedOrganelle organelle)
...
+
+        UpdateCompoundBagCapacities();
+    }
+
+    /// <summary>
+    ///   Updates <see cref="organellesCapacity"/> and <see cref="additionalCompoundCapacities"/>
+    ///   to take into account the addition or removal of an organelle
+    /// </summary>
+    /// <param name="organelle">The organelle being placed or removed</param>
+    /// <param name="negative"> Should be true if the organelle is being removed, otherwise false </param>
```
hhyyrylainen, line 1353:
```suggestion
    /// <param name="negative">Should be true if the organelle is being removed, otherwise false</param>
```

hhyyrylainen:
Just those two minor things I still noticed (well one is possibly not minor as downgrading the common submodule is a bit of real problem). Once those are fixed I think I'll just merge this, unless I can manage to get someone to playtest these changes.

buckly90:
Seems to work perfectly well, with one unfortunate exception.

Iron storage is not being increased as it should be, and the upgrade has no appearant effect.

ender-titan1:
Acknowledged, I will see what is causing the iron capacity not to increase, and I will try to resolve the submodule problem as soon as i can.

ender-titan1:
I've fixed the iron issue, however, I do not know how to fix the submodule problem. I've never worked with submodules in the past and trying to update them using git bash creates an error in the code, so I really don't know what I am supposed to do

hhyyrylainen:
> I've never worked with submodules in the past and trying to update them using git bash creates an error in the code, so I really don't know what I am supposed to do

Well that explains it. You've pulled in new code and then committed *all changes* which incorrectly picked the old submodule version. Git pull or branch switching do not update submodules, you need to run `git submodule update` to update submodules. You can probably fix things by:
```
cd RevolutionaryGamesCommon
git checkout 3a737999e08a88cd728067a7f4be695ef1d65f10
cd ..
git add ...
git commit ...
```

to create a new commit that sets the submodule version to what it is on master.

hhyyrylainen (approved):
Code changes look good now. Also the fix to the iron probably fixed all vacuole configuration issues from the underlying problem of the Godot option button not having the same items as the internal compounds list.

hhyyrylainen:
@buckly90 do you want to re-test this to make sure the iron bug is fixed? That's the last potential thing to test before merging this. At the latest I'll merge this in two days to make sure this makes it into the BOTD.

buckly90:
Looks great, splendid job!

PR #4462, opened 2023-09-10
Add fog of war to the patch map

### LINKED ISSUE #3602: Add fog of war for the patch map
only show the patches next to where the player's species is or the player has been before. We can probably make it persistent so that once some patch has become visible it stays permanently visible. Regions with no visible patches should probably be hidden. Visible regions should probably have their connection lines to other regions visible to give a hint where there are more regions. Freebuild should ignore the fog of war.

ender-titan1:

This PR hides unexplored patches and regions from players

**Related Issues**

Closes #3602

**Progress Checklist**

      break existing features:
      members reviewing this PR)


ender-titan1:
The basic implementation of this has been completed, including a cheat to reveal the entire map. However, I'm still unsure about it aesthetically. Currently, the map looks like this:

Screenshot (108)

I don't know whether or not to include region borders with regions that only contain unknown patches. If I leave the borders, players will know how big neighbouring regions are before actually entering them, however, without the borders the patches look very disjointed from the rest of the map.

hhyyrylainen:
I'd leave out the region border, like as you said otherwise the player will see the size of the region without having explored it.

ender-titan1:
I think this is ready for review now

buckly90:
Seems to work perfectly well with no appearant errors. Excellent job!

I love the mystery that this provides, and the risk of entering an unknown patch to figure out what conditions are present makes for interesting gameplay.
I'm a little worried this might make it a bit more oppresive to new or casual players though... 
We'll need to discuss this among the team to determine how to handle it in relation to difficulty and complexity.

In the mean time, I'd say this is good to go as is, though it may be helpful to implement a toggle for this feature.

GameDungeon:
I was going to suggest a toggle in the new game settings as well. For something like this, I think that is pretty critical.

ender-titan1:
Good to hear everything works well, I will get to adding the toggle soon.

hhyyrylainen:
Does this not show the patch next to the player? I initially imagined that the fog of war would show the patch details the player is next to, that ensures the player always knows the details of the patch they are about to think about moving to before actually confirming the move.

src/microbe_stage/editor/PatchMapNode.tscn
```diff
@@ -85,5 +87,15 @@ texture = ExtResource( 1 )
 expand = true
 stretch_mode = 6
 
+[node name="UnknownPatchLabel" type="Label" parent="MarginContainer/MarginContainer"]
```
hhyyrylainen, line 90: I think this needs the "PLACEHOLDER" editor text here to make sure the translation system doesn't pick this up. Or if you want to use the translation system then change the text to "UNKNOWN_PATCH_SYMBOL".

src/general/IDifficulty.cs
```diff
@@ -44,6 +44,11 @@ public interface IDifficulty : IRegistryAssignable
     /// </summary>
     public bool PassiveReproduction { get; }
 
+    /// <summary>
+    ///   Whether the patch map should be automatically fully revealed
+    /// </summary>
+    public bool PatchMapExploration { get; }
```
hhyyrylainen, line 50: Could you clarify with the documentation which of the following is true:
- To me PatchMapExploration being true would mean that you need to explore the map
- The documentation says the opposite i.e. if this is true then the map is revealed

Changing this property to be called something like `MapIsAlwaysFullyRevealed` would remove the potential confusion.

ender-titan1:
May I suggest adding a toggle to change whether or not to show the info of neighbouring patches? That way players wanting to explore more would be able to without compromising the experience of the casual player. Just a suggestion though.

hhyyrylainen:
I guess that would be an option, but then an enum should be used to store the setting as it is a pick one of 3 type of option, and an option button for selecting them. As to what to name the options, I don't have perfect ideas, something like "Fog of war enabled", extreme fog of war, entire map always visible.

ender-titan1:
Yeah, that's how I thought I could implement that

hhyyrylainen:
One thing that came up in playtesting is that it's now a bit weird with the unexplored regions having no border but the patch shown in them being positioned away from the connecting line.

hhyyrylainen:
Another finding: it is possible to select patches that are totally unknown to the player from the report dropdown to view what is going on there. I think this should be limited to the patches that are visible on the map.

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -27,6 +27,12 @@ public class PatchMapNode : MarginContainer
     [Export]
     public NodePath AdjacentPanelPath = null!;
 
+    /// <summary>
+    ///   For patches adjacent to a discovered patch that have not been entered by the player
```
hhyyrylainen, line 31: I think this comment needs to be worded to be a bit more vague, just so that we can have the more relaxed fog of war mode work and not contradict what the code comments say elsewhere.

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -83,6 +95,12 @@ public Patch Patch
 
     public Action<PatchMapNode>? SelectCallback { get; set; }
 
+    /// <summary>
+    ///   True if the patch is adjasent to a explored patch
+    ///   but has not been entered by the player
+    /// </summary>
+    public bool IsUnknown { get; set; }
```
hhyyrylainen, line 102: Does it make sense to have something be Discovered and IsUnnknown at the same time? If not I think one variable that's an enum should be used instead to dictate the overall state of a patch.

Also regarding the state I just remembered something. The player can always move back to the patch they started at in the editor, so I think the patch map needs to store the previous exploration states to allow resetting the map if the player decides to go back. Without that the player can explore multiple patches per editor cycle by moving to one patch and then going back to the patch they were in previously and then repeating that action.

src/microbe_stage/editor/MicrobeEditorReportComponent.cs
```diff
@@ -157,6 +158,8 @@ public void UpdatePatchDetails(Patch currentOrSelectedPatch, Patch? selectedPatc
 
         UpdateReportTabPatchName(currentOrSelectedPatch);
 
+        UpdateReportTabPatchSelector();
```
hhyyrylainen, line 161: Shouldn't this only be called if the player actually moves? Right now I'm not 100% certain but I think that UpdatePatchDetails can get called even when the player is just clicking around patches to look at their details.

src/microbe_stage/editor/MicrobeEditorCheatMenu.cs
```diff
@@ -5,13 +5,20 @@ public class MicrobeEditorCheatMenu : CheatMenu
...
     private CustomCheckBox infiniteMp = null!;
+    private Button revealEntirePatchMap = null!;
 #pragma warning restore CA2213
 
     public override void _Ready()
     {
         infiniteMp = GetNode<CustomCheckBox>(InfiniteMpPath);
+        revealEntirePatchMap = GetNode<Button>(RevealEntirePatchMapPath);
+
+        revealEntirePatchMap.Connect("pressed", this, nameof(OnRevealEntirePatchMapClicked));
```
hhyyrylainen, line 21: Why is this not connected in the scene file?

src/microbe_stage/editor/MicrobeEditorCheatMenu.cs
```diff
@@ -5,13 +5,20 @@ public class MicrobeEditorCheatMenu : CheatMenu
     [Export]
     public NodePath? InfiniteMpPath;
 
+    [Export]
+    public NodePath? RevealEntirePatchMapPath;
```
hhyyrylainen, line 9: Only the first nodepath should be nullable. All other ones should be `= null!` initialized.

The dispose method should be rewritten as:
```
if(InfiniteMpPath != null){
  InfiniteMpPath.Dispose();
  RevealEntirePatchMap.Dispose();
}
```

src/microbe_stage/PatchMap.cs
```diff
@@ -12,6 +12,13 @@ public class PatchMap : ISaveLoadable
 {
     private Patch? currentPatch;
 
+    public enum FogOfWarMode
+    {
+        Disabled,
+        Light,
```
hhyyrylainen, line 18: Maybe this could be named "normal"? I'm very in favour of this being the normal mode so it would make sense to the players seeing it for the first time (and also helps if the term used in the game text is the same as in the code).

src/general/base_stage/IEditorWithPatches.cs
```diff
@@ -6,4 +6,6 @@ public interface IEditorWithPatches : IEditor
     public Patch CurrentPatch { get; }
 
     public void OnCurrentPatchUpdated(Patch patch);
+
+    public void UpdateReportTabPatchSelector();
```
hhyyrylainen, line 10: This is for updating the selector when exploration state changes, right? I think it'd be better to have a more general callback for current patch being changed, which I thought would already get communicated to the editor somehow.

src/general/WorldGenerationSettings.cs
```diff
@@ -91,6 +91,9 @@ public enum LifeOrigin
     [JsonIgnore]
     public bool FreeGlucoseCloud => Difficulty.FreeGlucoseCloud;
 
+    [JsonIgnore]
+    public PatchMap.FogOfWarMode PatchMapExplorationMode => Difficulty.PatchMapExplorationMode;
```
hhyyrylainen, line 95: I think the FogOfWarMode is general enough concept to put the enum at the global level (not inside the PatchMap class).

simulation_parameters/common/difficulty_presets.json
```diff
@@ -23,6 +24,7 @@
     "osmoregulationMultiplier": 1,
     "freeGlucoseCloud": true,
     "passiveReproduction": true,
+    "patchMapExplorationMode": "Intense",
```
hhyyrylainen, line 27: I think all options should default to the normal exploration mode because I think the going in blind to a patch is much too punishing to increase the fun in the game, even for the hard difficulty. It should be a custom difficulty thing to enable it.

src/auto-evo/RunResults.cs
```diff
@@ -599,6 +599,14 @@ public void PrintSummary(PatchMap? previousPopulations = null)
             LocalizedStringBuilder PatchString(Patch patch)
             {
                 var builder2 = new LocalizedStringBuilder(80);
+
+                if (!patch.Explored)
```
hhyyrylainen, line 603: I think this method needs to have a parameter to skip Fog of war, so that the report to be shown in the auto-evo exploration tool works well.

src/general/IDifficulty.cs
```diff
@@ -44,6 +44,12 @@ public interface IDifficulty : IRegistryAssignable
     /// </summary>
     public bool PassiveReproduction { get; }
 
+    /// <summary>
+    ///   If true, the patch map will have to be discovered by the player instead of having the
```
hhyyrylainen, line 48: This is now an enum so the value can't be true...

src/general/MainMenu.tscn
```diff
@@ -898,7 +898,6 @@ visible = false
 visible = false
 
 [node name="NewGameSettings" parent="." instance=ExtResource( 18 )]
-visible = false
```
hhyyrylainen, line 901: Unintended change to this being visible by default.

src/general/NewGameSettings.tscn
```diff
@@ -29,6 +29,7 @@
 [sub_resource type="StyleBoxEmpty" id=1]
 
 [node name="NewGameSettings" type="Control"]
+visible = false
```
hhyyrylainen, line 32: This should be reverted.

src/general/NewGameSettings.cs
```diff
@@ -787,6 +813,12 @@ private void OnPassiveReproductionToggled(bool pressed)
         UpdateSelectedDifficultyPresetControl();
     }
 
+    private void OnPatchMapExplorationModeToggled(bool pressed)
```
hhyyrylainen, line 816: Is this a leftover method from the old implementation?

hhyyrylainen (changes requested):

src/microbe_stage/editor/MicrobeEditorCheatMenu.cs
ender-titan1: That's how it's done in the other cheat menu, so I figured it would work the same here

hhyyrylainen: I must not have been paid close attention then when that was merged. Could you set this in the scene file though as an example of what's the proper way to setup static signal links?

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -678,6 +739,9 @@ private void DrawRegionLinks()
             var region1 = map.Regions[entry.Key.x];
             var region2 = map.Regions[entry.Key.y];
 
+            if ((!region1.Explored && !region2.Explored) && !IgnoreFogOfWar)
```
Oliveriver, line 742: Parentheses can be removed here.

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -232,14 +240,25 @@ public void UpdateShownPatchDetails()
...
+            unknownPatch.Visible = true;
+        }
+        else
+        {
+            details.Visible = true;
+            nothingSelected.Visible = false;
+            unknownPatch.Visible = false;
 
-        UpdatePatchDetails();
+            UpdatePatchDetails();
```
Oliveriver, line 260: I'm worried that not calling this in all circumstances could lead to information not being updated correctly. Is there a particular reason it was moved inside the else block only?

Oliveriver (commented):
Super nice work! I tested this a small amount and found myself wondering why nobody thought to add this earlier.

I think there might be a bug in the algorithm that draws region connectors to only the single connecting patch before that patch is explored. Some regions were fine, but I found a few that were disconnected until the region had been revealed. Maybe I'm misunderstanding what the intention is though.

image

Game design wise, my only complaint is that having the green tendrils coming out from the current region with no indication of where they go - or even what they are at first - will be confusing for some players. I would prefer it if the region connectors aren't drawn until the patch/region they connect to is visible.

ender-titan1:
Pretty sure the "green tendrils" were supposed to be left in, if I understood the original issue correctly, as for the region connections: regions that have only unexplored patches visible are supposed to have connections to only the said patches. However the algorithm I implemented is not perfect and leads to issues with rendering the lines. I'm working on fixing that.

src/general/base_stage/IEditorWithPatches.cs
ender-titan1: This is also called by the new cheat, that's why I decided to add a new method

Athriq:
The description in that issue is better interpreted with the context of this concept art in mind (if you haven't seen already). We don't have to show a single "unknown patch" in unexplored regions. Just add fading lines to unexplored regions and the lines don't have to be long.

image

ender-titan1:
I've tried adding fading lines however godot was very uncooperative when it came to adding a gradient with alpha to a Line2D. It just didn't work. If there is a way to add a fading line like that I don't know about it.

hhyyrylainen:
I think it would be fine to just have a short line ending in a question mark, then that short line is replaced with the actual line when the target region the line goes to becomes visible. Trying to nicely fade out the ends of the lines can be left for later for someone to try to figure out.

TheESN:
Maybe it would be possible to make a fade out line by replacing actual line with specific sprite that represents it?

ender-titan1:
> Maybe it would be possible to make a fade out line by replacing actual line with specific sprite that represents it? 

Yeah, I replaced the draw calls with physical Line2D objects, and added a gradient to them when they needed to fade. This also meant that I had to convert the region borders to Line2D to stop the lines rendering over the borders. I believe this is the best way to do this.

hhyyrylainen:
Creating extra objects just to represent some lines seems like it's not optimal to me. But I guess it's the easiest way to add this and we can worry about the resource usage if we get big enough maps for it to matter.

Athriq:
> Yeah, I replaced the draw calls with physical Line2D objects, and added a gradient to them when they needed to fade.

Makes sense, I too use Line2D nodes for the report screen graphs to overcome similar problems.
As for concerns regarding the graphs performance, I think it has been pretty nice to the FPS thus far.

src/microbe_stage/gui/PatchDetailsPanel.cs
ender-titan1: This is the only case where actually updating the info is necessary, as the other two times it's either the "select patch" screen or the "unknown patch" screen

hhyyrylainen:
Regarding the performance, I think I remember my FPS dropping to around 220 from 360 when graphs are opening, but the FPS then recovers when the graph points stop animating. So while static the line draw instances are probably more efficient than when their properties change.

src/microbe_stage/gui/PatchDetailsPanel.cs
hhyyrylainen: That was exactly my impression as well, which is why I didn't comment on this specific part of the changes. So this optimizes things a bit with leaving the old data there when the data is not visible anyway, if I understood the changes right.

ender-titan1:
Got it, I'll convert the borders back to draw calls and adjust the endpoints appropriately

ender-titan1:
Alright, I can finally say the bulk of this is done. I've noticed two issues, however:
1. Loading a save file created in the patch map makes it render without any lines or borders
2. There is one edge case where the line seems to be incorrectly drawn (it looks like for some reason it's connecting to the wrong patch). I have not found a way to reproduce this and don't have the seed where I originally encountered it

Other than that everything works as intended, at least from my testing

hhyyrylainen:
> Loading a save file created in the patch map makes it render without any lines or borders

Is this a new bug / does this happen on master? If it is a new bug then an additional callback that triggers on normal editor tab swap needs to be made sure that loading a save correctly triggers that as well to refresh the draw state.

ender-titan1:
Yeah, this is new: I tried to recreate it on master but to no avail. However, I don't understand what I am supposed to do exactly. Create a method that triggers every time a tab is swapped?

hhyyrylainen:
I assume whatever trigger you have for drawing the map doesn't get executed the same way when the editor directly opens to the map tab. I'd have to read through all the code changes to say anything more accurate.

src/microbe_stage/editor/PatchMapEditorComponent.cs
```diff
@@ -283,4 +314,15 @@ private void MoveToPatchClicked()
     {
         SetPlayerPatch(mapDrawer.SelectedPatch);
     }
+
+    private void OnRevealEntirePatchMapCheatUsed(object sender, EventArgs args)
+    {
+        mapDrawer.IgnoreFogOfWar = true;
+
+        foreach (var patch in mapDrawer.PlayerPatch!.GetAllConnectedPatches())
```
hhyyrylainen, line 322: Is it safe to assume PlayerPatch is not null here? Wouldn't it be better just to get all patches in the map and mark them as explored?

src/microbe_stage/PatchRegion.cs
```diff
@@ -47,17 +47,40 @@ public enum RegionType
     public int ID { get; }
 
     /// <summary>
-    ///   Regions this is next to
+    ///   Regions this is next to, and the connecting patches in the other regions
     /// </summary>
     [JsonIgnore]
-    public ISet<PatchRegion> Adjacent { get; } = new HashSet<PatchRegion>();
+    public Dictionary<PatchRegion, Patch?> Adjacent { get; } = new();
```
hhyyrylainen, line 53: Why is the connecting patch allowed to be null here? Can regions be connected so that the target patch it connects to doesn't exist? I had the impression that that is not allowed, i.e. that situation is invalid map data.

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -583,7 +592,7 @@ private static void ConnectPatchesBetweenRegions(PatchRegion region, Random rand
         {
             case PatchRegion.RegionType.Ocean or PatchRegion.RegionType.Sea:
             {
-                foreach (var adjacent in region.Adjacent)
+                foreach (var adjacent in region.Adjacent.CloneShallow().Keys)
```
hhyyrylainen, line 595: I assume this is here as the adjacent data is now modified here? If so the usual pattern to use is `region.Adjacent.Keys.ToList()` to take a copy of the keys.

src/microbe_stage/PatchMap.cs
```diff
@@ -4,6 +4,39 @@
 using Godot;
 using Newtonsoft.Json;
 
+/// <summary>
+///   Whether the fog of war should be disabled,
+///   mark patches adjacent to the player as explored,
+///   or mark only the patch the player has entered as explored
+/// </summary>
+public enum FogOfWarMode
```
hhyyrylainen, line 12: The first type in a file should match the filename. This means that this enum needs to be put in a separate file and MapElementVisibility should be put inside the class (if it isn't used widely outside this class, and if it is then it also needs to go int oa separate file).

src/general/base_stage/IEditorWithPatches.cs
hhyyrylainen: I still think the name of this method should be changed to something like "OnAvailablePatchesChanged" or something like that to signal why this method is called.

src/general/MainMenu.tscn
hhyyrylainen: This line of code is still here...

src/auto-evo/RunResults.cs
```diff
@@ -584,9 +584,10 @@ public void PrintSummary(PatchMap? previousPopulations = null)
         ///   Must be final effects with <see cref="ExternalEffect.Coefficient"/> set to 1 created by
         ///   <see cref="AutoEvoRun.CalculateFinalExternalEffectSizes"/>
         /// </param>
+        /// <param name="ignoreFogOfWar">If ture ignore fog of war</param>
         /// <returns>The generated summary text</returns>
         public LocalizedStringBuilder MakeSummary(PatchMap? previousPopulations = null,
-            bool playerReadable = false, List<ExternalEffect>? effects = null)
+            bool playerReadable = false, List<ExternalEffect>? effects = null, bool ignoreFogOfWar = false)
```
hhyyrylainen, line 590: I think maybe the player readable being false should also imply ignoreFogOfWar as that is meant for a technical data dump, and obfuscating any data in that doesn't make any sense to me...

src/general/NewGameSettings.cs
```diff
@@ -301,6 +311,10 @@ public override void _Ready()
             difficultyPresetAdvancedButton.AddItem(preset.UntranslatedName);
         }
 
+        patchMapExplorationMode.AddItem(TranslationServer.Translate("FOG_OF_WAR_DISABLED"));
```
hhyyrylainen, line 314: Could you add a comment to the enum for the exploration mode that the items may not be reordered, or their values may not change as otherwise the new game GUI will break?

src/microbe_stage/PatchRegion.cs
ender-titan1: The reason this is the way it is, is because the method that links regions together doesn't actually know the patches it is connecting to, but I think Ihave an idea on how to fix this

src/microbe_stage/PatchMap.cs
ender-titan1: Yeah, MapElementVisibility is an enum I use in PatchRegion, Patch and PatchNode (or whatever that class is called). It is basically a way for me to standardise how each of those elements stores their visibility state

src/auto-evo/RunResults.cs
```diff
@@ -584,13 +584,17 @@ public void PrintSummary(PatchMap? previousPopulations = null)
         ///   Must be final effects with <see cref="ExternalEffect.Coefficient"/> set to 1 created by
         ///   <see cref="AutoEvoRun.CalculateFinalExternalEffectSizes"/>
         /// </param>
+        /// <param name="ignoreFogOfWar">If ture ignore fog of war</param>
```
hhyyrylainen, line 587:
```suggestion
        /// <param name="ignoreFogOfWar">If true ignore fog of war</param>
```

Typo

src/auto-evo/RunResults.cs
```diff
@@ -584,13 +584,17 @@ public void PrintSummary(PatchMap? previousPopulations = null)
...
         /// <returns>The generated summary text</returns>
         public LocalizedStringBuilder MakeSummary(PatchMap? previousPopulations = null,
-            bool playerReadable = false, List<ExternalEffect>? effects = null)
+            bool playerReadable = false, List<ExternalEffect>? effects = null, bool ignoreFogOfWar = false)
         {
             if (previousPopulations != null && previousPopulations.CurrentPatch == null)
                 throw new ArgumentException("When previous populations is set, it must have current patch set");
 
+            if (!playerReadable)
```
hhyyrylainen, line 595: Could you add a comment here saying why this is forced to true. Future programmers will have no clue why this is done otherwise...

src/microbe_stage/Patch.cs
```diff
@@ -88,6 +88,42 @@ public Patch(LocalizedString name, int id, Biome biomeTemplate, PatchSnapshot cu
...
+                VisibilityState = MapElementVisibility.Explored;
+                Region.VisibilityState = MapElementVisibility.Explored;
+            }
+            else
+            {
+                VisibilityState = MapElementVisibility.Undiscovered;
+                Region.VisibilityState = MapElementVisibility.Undiscovered;
+            }
+
+            if (value)
```
hhyyrylainen, line 107: You already check for `if (value)` a few lines above this so I think this if's content should go there.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -668,34 +870,38 @@ private void AdjustPathEndpoints()
...
-        // Then we draw the the adjacent connections to the patch we selected
-        // Those connections have to be drawn over the normal connections so they're second
-        foreach (var points in highlightedConnections)
-        {
-            for (int i = 1; i < points.Length; i++)
-            {
-                DrawNodeLink(points[i - 1], points[i], HighlightedConnectionColor);
-            }
+            // Set the color of the line if highlighted
+            if (CheckHighlightedAdjacency(start, end))
```
hhyyrylainen, line 890: Multiline if must use braces.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -668,34 +870,38 @@ private void AdjustPathEndpoints()
...
+                entry.Value.DefaultColor = HighlightedConnectionColor;
+            else
+                entry.Value.DefaultColor = DefaultConnectionColor;
+
+            if (IgnoreFogOfWar)
+                return;
+
+            // Add a fade to the line if its ending at an unexplored region
+            if (start.VisibilityState == MapElementVisibility.Undiscovered)
```
hhyyrylainen, line 899: And also here.

hhyyrylainen:
Regarding the problem of this not updating when loading a save on the map screen, I can't immediately tell what has gone wrong. Next I'd need to use a debugger to see what's the normal map update and rendering flow (using breakpoints) and then compare that to what happens when a save is loaded. I'm pretty busy so I'm hoping someone else can do that. I'll help out there as a last resort. My guess as to what is wrong is that the editor tab swap logic is slightly flawed when loading a save which hasn't been a problem before but now with these changes is a problem. That's where I'd start investigating if the method call sequence when normally swapping tabs and when loading a save match or not.

ender-titan1:
I have managed to fix that one edge case I mentioned, where now multiple connections between the same two regions will be accounted for when creating paths. However, the way I implemented it (and what looks to be the easiest way to do so) came with a side effect. Some of the links leading to unexplored regions will be repositioned and change sides when said region is explored. I don't know if that is ok to leave in or if should it be fixed, so I'm just putting it out here.

hhyyrylainen:
I haven't tested this yet, but I think it should be a pretty important goal to ensure the links or whatever don't visually jump around a lot when the player is switching patches. Otherwise it might look weird to the player to see the world changing shape as they move around.

src/general/base_stage/IEditorWithPatches.cs
hhyyrylainen: The name here is not changed yet.

src/general/MainMenu.tscn
hhyyrylainen: Just commenting again to add this (and my previous comment) to my current review.

src/general/NewGameSettings.tscn
hhyyrylainen: The diff still shows this line existing...

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -443,6 +546,67 @@ private Vector2[] GetLeastIntersectingPath(PatchRegion start, PatchRegion end)
...
+                    {
+                        pathsToDiscard.Add(entry.Path);
+                        break;
+                    }
+                }
+            }
+        }
+
+        // TODO: There is probably a better way to do this
+        probablePaths.RemoveAll(e => pathsToDiscard.Contains(e.Path));
```
hhyyrylainen, line 608: I think this is pretty close to optimal but I usually just do something like:
```
foreach (var toDelete in patshToDiscard)
{
     probablePaths.Remove(toDelete.Path);
}
```

hhyyrylainen (changes requested):
I didn't understand the map drawer fully before this PR and I still don't fully know it, so here's basically as full review of this PR as I can do.

ender-titan1:
I'm kinda in the process of a big refactor to the drawer to solve a lot of issues I'm having, so I'll convert this to a draft until I think this new version is ready

voxelbugged:
Shouldn't this be removed from the planning board (or moved to Done) now that another pull request has implemented it?

hhyyrylainen:
You are correct, apparently this wasn't automatically moved to the done column after being closed.

PR #4463, opened 2023-09-11
Only apply an organelle upgrade if the upgrade data actually changed

### LINKED ISSUE #4091: When no organelle upgrades are done there should be no change action created
to avoid creating pointless actions in the action history

needs some interface or something for the custom data to also allow comparing itself

Kashnox:

This stops actions from being added after opening the "modify" menu if no changes were actually made.

**Related Issues**

Closes #4091 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/organelle_components/ChemoreceptorComponent.cs
```diff
@@ -107,6 +107,18 @@ public class ChemoreceptorUpgrades : IComponentSpecificUpgrades
     public float SearchAmount { get; set; }
     public Color LineColour { get; set; }
 
+    public bool Equals(IComponentSpecificUpgrades other)
+    {
+        if (other is not ChemoreceptorUpgrades otherChemoreceptor)
+            return false;
+
+        return TargetCompound?.InternalName == otherChemoreceptor.TargetCompound?.InternalName
```
hhyyrylainen, line 115:
```suggestion
        return TargetCompound == otherChemoreceptor.TargetCompound
```

Compounds should be compared by reference as is done everywhere else in the codebase.

src/microbe_stage/editor/upgrades/ChemoreceptorUpgradeGUI.cs
```diff
@@ -105,6 +105,12 @@ public void OnStartFor(OrganelleTemplate organelle, GameProperties currentGame)
                 SimulationParameters.Instance.GetCompound(Constants.CHEMORECEPTOR_DEFAULT_COMPOUND_NAME);
             var defaultConfiguration = new ChemoreceptorUpgrades(defaultCompound, null,
                 Constants.CHEMORECEPTOR_RANGE_DEFAULT, Constants.CHEMORECEPTOR_AMOUNT_DEFAULT, defaultCompound.Colour);
+
+            // Assign the default ChemoreceptorUpgrades to the OrganelleUpgrades
+            // so we can check if it's been changed at the end
+            organelle.Upgrades ??= new OrganelleUpgrades();
```
hhyyrylainen, line 111: Is this fine? This seems quite suspicious change to me as this modifies the organelle state *before* anything is accepted. Would it be possible to either compare null to be equal to the default value or just document that if the upgrade is null then applying the default values is considered a "change"?

src/microbe_stage/OrganelleUpgrades.cs
```diff
@@ -17,6 +18,20 @@ public class OrganelleUpgrades : ICloneable
     /// </summary>
     public IComponentSpecificUpgrades? CustomUpgradeData { get; set; }
 
+    public bool Equals(OrganelleUpgrades other)
+    {
+        if (!UnlockedFeatures.SequenceEqual(other.UnlockedFeatures))
+            return false;
+
+        if (CustomUpgradeData == null)
+            return other.CustomUpgradeData == null;
```
hhyyrylainen, line 27: So maybe this could have something like `other.CustomUpgradeData == null || other.CustomUpgradeData.EqualsDefault()`?

hhyyrylainen (changes requested):
Thank you for considering contributing to Thrive. I think the approach to this feature is already very good, but there's a few details which I think are problems and commented on.

src/microbe_stage/OrganelleUpgrades.cs
Kashnox: I thought about doing something like that too, but there were a few things I was concerned about.

1. An OrganelleUpgrades for a Chemoreceptor with null CustomUpgradeData would be considered equal to an OrganelleUpgrades for a Lysosome with the default value. Right now, an OrganelleUpgrades for a Chemoreceptor and an OrganelleUpgrades for a Lysosome should never be compared anyway, so this wouldn't cause any problems, it just made me a little worried.
2. LysosomeUpgrades and ChemoreceptorUpgrades don't have any way to know what their default value is, so the logic to figure that out would either have to be moved to those classes or split into a public method on their UpgradeGUIs so they can access it.

Assigning the default value to the upgrade data as soon as the GUI is opened felt a little suspicious to me too, so if you think the problems above are fine, I'll make that change.

hhyyrylainen: Those are valid concerns I didn't consider. It would be possible to specify the defaults for those missing upgrade types. But I think maybe it would be easier to just live with the problem of null not being equal to the default upgrades. I don't think it'll be that big of a problem that the user doesn't change any options but just accepts the changes and ends up with a change from null to the default values. At least I see that as the least worst of all the solutions we've come up with.

hhyyrylainen (approved):
Code looks good now and I did a quick test in game with a debugger to verify at least chemoreceptor upgrades now correctly detect when the data hasn't changed and it skips applying the update.

PR #4466, opened 2023-09-12
Toxins fired in a colony (or multicellular) can collide with the firing colony

### LINKED ISSUE #4342: Toxins fired in a colony (or multicellular) can collide with the firing colony
this should not happen, we have the collision exception with the shooter and I think also the same species check as the shooter should do something here. So this is at least 2 things that go wrong.

Kashnox:

This adds a collision exception to AgentProjectiles for each microbe in the firing microbe's colony.
Previously, all microbes of the same species as the firing microbe were prevented from taking damage from the projectile, but only the firing microbe had a collision exception.

**Related Issues**

closes #4342 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/AgentProjectile.cs
+++ b/src/microbe_stage/AgentProjectile.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using Godot;
 using Newtonsoft.Json;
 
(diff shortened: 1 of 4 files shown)


hhyyrylainen:
This probably works completely fine to solve the problem. However, I'm working on a very huge microbe stage refactoring (https://github.com/Revolutionary-Games/Thrive/tree/performance_refactoring) where the microbe colony physics will be entirely redone (colony member physics bodies will be simply made to not exist). So I'll hold off on merging this PR as that would just cause more merge conflicts and slow down my work (which is basically what will determine when we can release the next Thrive version). I hope that work will be completed within a few weeks. 
You can ask on an issue first before working on it if I think it is safe to work on. All GUI related, later stage prototype, HUD etc. work should be completely safe to work on.

Kashnox:
A rework that big sounds exciting. I'll make sure to ask before starting on issues that seem like they could interfere with that from now on.

hhyyrylainen:
With the ECS refactor making the cell colonies into a single physics body which ignores toxin collisions as it is the same species as the shooter, this should no longer be able to happen.

PR #4468, opened 2023-09-13
Editor button confirm warning icon can get stuck on sometimes

### LINKED ISSUE #4310: Editor button confirm warning icon can get stuck on sometimes
https://community.revolutionarygamesstudio.com/t/0-6-2-release-candidate-testing/6345/3?u=hhyyrylainen

Seems like some kind of tab switching results in the warning icon being stuck on.

Kashnox:

This PR started hiding the warning icon after the "hide" animation finishes playing. Previously, the badge was made visible before playing the "show" animation for the first time and its visibility was never set back to false after that. It relied on the "hide" animation to make it appear visible. If we don't like the approach I took, we could probably also skip to the end of the "hide"/"show" animation (whichever is appropriate at the time) whenever the editor tab is opened instead.

**Related Issues**

closes #4310 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/base_stage/EditorComponentWithActionsBase.cs
+++ b/src/general/base_stage/EditorComponentWithActionsBase.cs
@@ -107,6 +107,10 @@ public abstract class EditorComponentWithActionsBase<TEditor, TAction> : EditorC
     {
         SetUndoButtonStatus(canUndo && !Editor.CanCancelAction);
         SetRedoButtonStatus(canRedo && !Editor.CanCancelAction);
+
+        // Update the finish badge warning visibility aftrer updating action history
+        // because undoing and redoing actions can cause or fix warnings
+        UpdateFinishButtonWarningVisibility();
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


Kashnox:
I also noticed and fixed a similar issue while testing this where undoing or redoing to a state that had a warning didn't cause the warning icon to show up.

hhyyrylainen:
Steps to consistently reproduce the bug (from the dev discord) for testing purposes if anyone wants to test:
> causing a warning, fixing the warning, going to the Patch Map tab, then going back to the Editor tab

src/general/base_stage/EditorComponentWithActionsBase.cs
```diff
@@ -135,11 +146,17 @@ protected virtual void OnCancelActionClicked()
     protected void Undo()
     {
         Editor.Undo();
+
+        // Update the finish badge warning visibility in case undoing an action caused or fixed the warning
+        UpdateFinishButtonWarningVisibility();
```
hhyyrylainen, line 151: Is this the right place to put this? As EditorBase has `[RunOnKeyDown("e_undo")]` it can be triggered from the keyboard without this being called. I think instead you'll want to hook into the editor "on action changed". So that'd be in the component's `UpdateUndoRedoButtons` method (though the name is a bit too specific to really warrant this, but I'll accept it by thinking about the warning badge being basically kind of a thing that is related to the currently performed action history).

hhyyrylainen (approved):
The code change now looks correct to me. I'm getting the impression that no one has time to confirm the bugfix, so I'll just merge this today as-is.

PR #4469, opened 2023-09-14
Double click on organelle move crashes the game

### LINKED ISSUE #4442: Double click on organelle move crashes the game
Exception:
```
ERROR: ------------ Begin of Unhandled Exception Log ------------
ERROR: The following exception prevented the game from running:
ERROR: 
ERROR: System.InvalidOperationException: Can't begin hex move while another in progress
ERROR: 
ERROR:   at HexEditorComponentBase`4[TEditor,TCombinedAction,TAction,THexMove].StartHexMove (THexMove selectedHex) [0x0000d] in <d25c8ec50b0d44ad8b235148ea37e4bf>:0 
ERROR: 
ERROR:   at CellEditorComponent.OnMovePressed () [0x00035] in <d25c8ec50b0d44ad8b235148ea37e4bf>:0 
ERROR: 
ERROR:   at (wrapper managed-to-native) Godot.NativeCalls.godot_icall_2_695(intptr,intptr,string,object[])
ERROR: 
ERROR:   at Godot.Object.EmitSignal (System.String signal, System.Object[] args) [0x0000b] in <19daba5d5ae94c7483bcc78b7907d0ef>:0 
ERROR: 
ERROR:   at HexPopupMenu.OnMovePressed () [0x00015] in <d25c8ec50b0d44ad8b235148ea37e4bf>:0 
ERROR: 
ERROR: Please provide us with this log, thank you.
ERROR: ------------  End of Unhandled Exception Log  ------------
```

Edit: I initially thought that this was likely caused by the move hotkey, but that doesn't seem to be the case.

The hotkey should be modified so that pressing it is ignored if a move is currently in progress / maybe try to finish the current move at the current position.

So in addition to that the root cause for being able to have multiple organelle popup menus open at once needs to be fixed.

Kashnox:

This PR adds some extra checks before making changes to organelles in the editor. Before, we were only checking that the HexPopupMenu was still visible, but it takes a few frames to fully close, so multiple clicks or key presses could come through in that time. I started checking the Closing property in addition to the Visible property because Closing gets set immediately. I also added the check to a few extra places.

**Related Issues**

closes #4442 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/HexPopupMenu.cs
+++ b/src/microbe_stage/editor/HexPopupMenu.cs
@@ -108,7 +108,7 @@ public abstract class HexPopupMenu : CustomPopupMenu
         }
     }
 
-    private bool CanModifyHex => Visible && !Closing;
+    private bool IsInteractable => Visible && !Closing;
 
     public override void _Ready()
     {
@@ -132,7 +132,7 @@ public abstract class HexPopupMenu : CustomPopupMenu
     [RunOnKeyDown("e_delete", Priority = 1)]
     public bool OnDeleteKeyPressed()
     {
-        if (CanModifyHex)
+        if (IsInteractable)
         {
             EmitSignal(nameof(DeletePressed));
 
@@ -148,7 +148,7 @@ public abstract class HexPopupMenu : CustomPopupMenu
     [RunOnKeyDown("e_move", Priority = 1)]
     public bool OnMoveKeyPressed()
     {
-        if (CanModifyHex)
+        if (IsInteractable)
         {
             EmitSignal(nameof(MovePressed));
 
@@ -222,7 +222,7 @@ public abstract class HexPopupMenu : CustomPopupMenu
 
     private void OnDeletePressed()
     {
-        if (!CanModifyHex)
+        if (!IsInteractable)
             return;
 
         GUICommon.Instance.PlayButtonPressSound();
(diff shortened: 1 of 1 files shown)


src/microbe_stage/editor/HexPopupMenu.cs
```diff
@@ -108,6 +108,8 @@ public virtual bool EnableModifyOption
         }
     }
 
+    private bool CanModifyHex => Visible && !Closing;
```
hhyyrylainen, line 111: This seems overall a sensible change, but I'm kind of unsure if this is the best name for this property. Would something like `IsInteractable` or maybe `IsOpenAndNotClosing` be a better name? As this also controls if delete is allowed and not just hex modification.

Also if this fix works, could you also check the multicellular metaball popup? It probably has the exact same bug as this class.

Kashnox: The multicellular metaball popup derives from HexPopupMenu and doesn't add any new ways to accept input, so it did have this problem, but it's already fixed by the changes to HexMenuPopup.

hhyyrylainen: Ah that is good then.

hhyyrylainen (approved):
The code change looks good to me now. I'll leave this open for a bit longer to see if there's anyone available to test this.

PR #4475, opened 2023-09-16
Art gallery slideshow works properly for models

### LINKED ISSUE #4388: Art gallery slideshow feature doesn't work for models
I just noticed that the slideshow feature does nothing when viewing a model. When viewing the images the slideshow button works correctly and switches to new images.

Brabygg:

The art gallery's slideshow button, which previously did nothing at all when used in the model viewer, now correctly cycles between the viewer's models in order, complete with the correct fade transition.

**Related Issues**

Closes #4388

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/gui_common/art_gallery/CrossFadableGalleryViewport.cs
```diff
@@ -0,0 +1,62 @@
...
+    private Tween tween = null!;
+#pragma warning restore CA2213
+
+    [Signal]
+    public delegate void Faded();
+
+    /// <summary>
+    ///   Triggers the gallery model viewer's crossfade effect.
+    /// </summary>
+    public bool FadeRequested
```
Kashnox, line 21: Is there any reason to use the FadeRequested property instead of letting SlideScreen call BeginFade() directly?

src/gui_common/art_gallery/SlideScreen.cs
```diff
@@ -332,6 +332,14 @@ private void ChangeSlide(bool fade)
 
         var item = items[currentSlideIndex];
         slideTextureRect.Image = GD.Load(item.Asset.ResourcePath) as Texture;
+
+        if (slideTextureRect.Image != null)
+            return;
+
+        // If texture loading fails, the selected item is a model
+        // These are handled here
+
```
Kashnox, line 341: There's no need for an extra line here.

Kashnox (changes requested):
The CrossFadableGalleryViewport seems like a good fix for this problem. There was just one thing that looked a little overly complicated, which I left a review comment about.

src/gui_common/art_gallery/CrossFadableGalleryViewport.cs
hhyyrylainen: I had a look through the changes here and this is the only thing I also noticed that could be improved.

Brabygg: I tried this, but I don't really see a way to just call the method without running into errors about conflicting static and non-static contexts. ChangeSlide() is static, while BeginFade() needs to be nonstatic. Most direct I can do is probably a signal.

Brabygg:
This commit, I believe, is the closest possible route to calling the method directly. It's certainly shorter and more concise.

src/gui_common/art_gallery/CrossFadableGalleryViewport.cs
Kashnox: I might have worded my initial review badly. You don't have to call BeginFade() in a static way. You can just call modelViewContainer!.BeginFade()

hhyyrylainen:
I'm confused. Let me try just directly calling the method, I hope you don't mind me committing to your branch...

hhyyrylainen (approved):
I redid the signal sending with just a direct method call. It seems to work entirely fine. I noticed a very slight bug which I'm not sure is worth fixing: if the user hits the X to cancel the slideshow while it is fading between models, then the next time the slideshow is started it shows with partially faded screen until the model changes once again.

PR #4476, opened 2023-09-17
Global timeline doesn't show total population correctly

### LINKED ISSUE #3611: Global timeline doesn't show total population correctly
There's probably some logic difference between the local and global timeline that causes the global timeline to have incorrect numbers. See issue reported here: https://github.com/Revolutionary-Games/Thrive/pull/3605#pullrequestreview-1052801452

Kashnox:

This PR starts calling ApplyExternalEffects from within the CalculateFinalExternalEffectSizes method. The external effects still aren't actually applied to the world until ApplyAllResultsAndEffects is called, but they're stored in the AutoEvoRun.Results' patch information so summaries and timeline events can use them.

Previously, everything logging the results was manually adding the external effects except the global timeline.

**Related Issues**

closes #3611 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Kashnox:
I tried to thoroughly test the Auto-Evo Prediction, the Auto-Evo summary in the editor, the Local and Global Timelines, and the Auto-Evo Exploring Tool to make sure they're still working correctly, but auto-evo is complicated, so someone who has more experience working with it should take a good look at my changes too.

buckly90:
Numbers all seem to match up accordingly. No issues appearant. Well done!

src/auto-evo/AutoEvoRun.cs
```diff
@@ -242,7 +242,7 @@ public bool IsFinished(bool autoStart = true)
     /// <remarks>
     ///   <para>
     ///     This has to be called after this run is finished.
-    ///     <see cref="CalculateFinalExternalEffectSizes"/> must be called first,
+    ///     <see cref="CalculateAndApplyFinalExternalEffectSizes"/> must be called first,
     ///     that should be called even before generating the result summaries to make sure they are accurate.
     ///   </para>
     /// </remarks>
```
hhyyrylainen, line 248: Just one small comment, should `ApplyAllResultsAndEffects` now be renamed to `ApplyAllResults`? As it no longer handles applying the external effects.

hhyyrylainen: The summary on this method is `Applies things added by addExternalPopulationEffect and auto-evo results` but I think it should now be `Applies computed auto-evo results to the world` as with the name this method is no longer responsible for handling the external effects instead the computed results need to be the final numbers to apply.

Kashnox: Ah, okay, that makes sense.

hhyyrylainen (approved):
This now makes sense to me. I'll trust that Buckly's testing was enough so I won't do a second playtest on this.

PR #4479, opened 2023-09-17
Automatically prompts to disable 3D backgrounds on low performance

### LINKED ISSUE #4326: Consider adding some kind of popup suggesting disabling 3D menu backgrounds when low performance
As the 3D backgrounds seem to be really performance intensive. I'm imagining something like if over the past 10-15 seconds while in the main menu the game FPS is less than 30 and the 3D backgrounds are enabled, the game would popup asking if the player wants to disable the 3D backgrounds which are quite performance intensive (or maybe alternatively it could disable https://github.com/Revolutionary-Games/Thrive/issues/4285 options). The popup should be permanently dismissible and OK would disable the backgrounds and cancel would keep them on.

Brabygg:

When the average framerate in the main menu is lower than 30 FPS for 15 seconds and 3D menu backgrounds are enabled, a popup will appear asking if the user would like to disable the aforementioned option. This can be accepted, updating the setting immediately, declined, or permanently dismissed.

Note that the tracked framerate is an approximation and not the exact average. I believe this is acceptable, as the intent is to track general low performance, not exact FPS.

**Related Issues**

Closes #4326 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Brabygg:
Give me, like, a day to sort the localization out. Gettext and Windows do not seem to want to cooperate.

hhyyrylainen:
Last time someone had issues with that, they had some truly ancient version of gettext that they got from some random place.

src/general/MainMenu.cs
```diff
@@ -468,6 +500,12 @@ private void RunMenuSetup()
         UpdateStoreVersionStatus();
         UpdateLauncherState();
 
+        if (!Settings.Instance.IsNoticePermanentlyDismissed(DismissibleNotice.LowPerformanceWarning))
+        {
+            GD.Print("Low performance warning is not dismissed, beginning menu performance check");
```
hhyyrylainen, line 505: I think this print is unnecessary.

src/engine/Settings.cs
```diff
@@ -830,6 +830,15 @@ public void ApplyLanguageSettings()
         GD.Print("Set C# locale to: ", cultureInfo, " Godot locale is: ", TranslationServer.GetLocale());
     }
 
+    /// <summary>
+    ///   Currently only used by one dialog, called to update the specified setting without using the dedicated menu
+    /// </summary>
+    public void UpdateSettingExternally(SettingValue<bool> setting, bool value)
```
hhyyrylainen, line 836: I'm a bit unsure why this method was required to exist. Does it not work just by setting the value directly on that one option?

src/general/MainMenu.cs
```diff
@@ -752,6 +790,33 @@ private void WarnAboutNoEnabledMods()
         }
     }
 
+    private float TrackMenuPerformance(float delta)
+    {
+        var currentFrameRate = 1 / delta;
```
hhyyrylainen, line 795: I think the framerate should be read from Godot, there's an engine method to get the current framerate. Also I think the performance measurement should only begin one second in to the menu so that an initial lag spike is not counted. Also just to confirm that the measuring shouldn't happen before the intro video if it is enabled ends.

src/general/MainMenu.cs
```diff
@@ -752,6 +790,33 @@ private void WarnAboutNoEnabledMods()
...
+        {
+            GD.Print($"Average frame rate is {averageFrameRate}, prompting to disable 3D backgrounds");
+            lowPerformanceWarning.PopupIfNotDismissed();
+            canShowLowPerformanceWarning = false;
+        }
+    }
+
+    private void OnLowPerformanceDialogConfirmed()
+    {
+        Settings.Instance.UpdateSettingExternally(Settings.Instance.Menu3DBackgroundEnabled, false);
```
hhyyrylainen, line 817: The new settings should be saved. Otherwise the problem appears again after restarting the game.

Also did I notice it or how is the remembering if the notice shouldn't be shown again checkbox is handled?

hhyyrylainen (changes requested):
Thanks for giving this issue a shot, I'm sure people with lower end hardware will appreciate this and I'll appreciate not having menu performance complaints anymore.

src/general/MainMenu.cs
```diff
@@ -752,6 +790,33 @@ private void WarnAboutNoEnabledMods()
...
+        if (averageFrameRate == 0)
+            return currentFrameRate;
+
+        // Not an exact average by any means, but good enough for this purpose
+        return (averageFrameRate + currentFrameRate) / 2;
+    }
+
+    private void WarnAboutLowPerformance()
+    {
+        if (averageFrameRate < 30 && secondsInMenu >= 15)
```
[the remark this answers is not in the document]
Brabygg: > The low performance warning should only open if another menu popup is not open already, that way it's ensured to not get in the way.

Is this already checked for somewhere, or will I have to implement that from scratch?

> I think the measuring should be constant and be able to trigger later for example when the player has been in the menu for a while looking at new game settings.

Figured. Gonna sort that out.

src/general/MainMenu.cs
Brabygg: > how is the remembering if the notice shouldn't be shown again checkbox is handled?

The popup uses the already in-use PermanentlyDismissibleDialog. It seems to work on my end; just needed to be added to DismissibleNotice.cs.

src/general/MainMenu.cs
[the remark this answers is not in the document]
hhyyrylainen: You can have a look around in the main menu file, it's possible that I made one of the popups trigger if all the other ones are hidden, but otherwise just create a new method called something like `IsAnyPopupOpen` in the main menu and just `return thanksDialog.Visible || someOtherDialog.Visible ...` (I can't remember the exact variable names off the top of my head).

src/general/MainMenu.cs
hhyyrylainen: With that class you should call `PopupIfNotDismissed` to show it, I don't remember seeing you call that method but I could be wrong. Also in this situation it should act opposite, so if you dismiss the dialog then the "don't show again" state should be remembered, right now PermanentlyDismissibleDialog is setup to only remember the permanent dismiss on confirm. So you need to actually add a new export variable to that class that makes it remember the dismiss state when it is dismissed.

Brabygg:
Summary of new commit:
 - Loc files are no longer broken (although I'm not sure I updated them right, tell me if something's wrong)
 - UpdateSettingExternally() has been replaced with a more direct approach of updating the value directly from MainMenu.cs
 - Now reads framerate using Engine.GetFramesPerSecond()
 - Tracking only begins after one second has passed in the menu
 - Popup should no longer open if any other popup is displayed (The AreAnyMenuPopupsOpen() method created for this may be useful elsewhere too)
 - Measuring is now constant and the popup should be able to trigger at any time
 - "Don't show again" is now remembered on cancel rather than confirm

src/general/MainMenu.cs
```diff
@@ -752,6 +788,47 @@ private void WarnAboutNoEnabledMods()
         }
     }
 
+    private float TrackMenuPerformance()
```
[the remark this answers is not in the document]
hhyyrylainen: I think tracking menu performance is a good term to describe what this does. I already suggested swapping to the Godot engine inbuilt way of getting the current framerate, so that would make this method less technical towards framerate calculation...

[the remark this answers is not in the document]
CoreyHendrey: It is used to update averageFramerate though, and then averageFramerate is used as the performance metric. It's mostly unimportant anyway, just to stop anyone else rerolling an average FR function when it's already here.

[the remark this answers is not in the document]
hhyyrylainen: Well it's not really useful for general use at all if it is just in the main menu class...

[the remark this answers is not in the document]
CoreyHendrey: Haha yeah that's true I guess. :)

src/engine/Settings.cs
```diff
@@ -124,7 +124,7 @@ private Settings()
     ///   Enable or disable 3D background scenes in the menu
     /// </summary>
     [JsonProperty]
-    public SettingValue<bool> Menu3DBackgroundEnabled { get; private set; } = new(true);
+    public SettingValue<bool> Menu3DBackgroundEnabled { get; set; } = new(true);
```
hhyyrylainen, line 127: This isn't correct. Changing the setting value will break all observers on the value from working. To assign a new value you need to set the underlying value `Menu3DBackgroundEnabled.Value = false` that will notify all observers that the value changed.

src/general/MainMenu.tscn
```diff
@@ -1052,3 +1059,5 @@ DialogText = "STEAM_INIT_FAILED_DESCRIPTION"
 [connection signal="OnFinishedSignal" from="CreditsView/CreditsScroll" to="." method="OnReturnFromCredits"]
 [connection signal="pressed" from="CreditsView/Back" to="." method="OnReturnFromCredits"]
 [connection signal="OnClosed" from="ModManager" to="." method="OnReturnFromMods"]
+[connection signal="Cancelled" from="LowPerformanceDialog" to="LowPerformanceDialog" method="OnConfirmed"]
```
hhyyrylainen, line 1062: This seems like a hack to me. Could you instead edit the dismissible warning class to add a new property to make it remember the dismissed state on cancel? It should probably be an enum to allow flexibility later, the default should be `RememberOnConfirm` and here the value should be `RememberOnCancel`.

src/general/MainMenu.cs
```diff
@@ -273,6 +283,24 @@ public override void _Process(float delta)
...
                 }
             }
+
+            // Low menu performance will never be warned about if the popup has been dismissed,
+            // if 3D backgrounds have been disabled, if the popup has been shown but not dismissed
+            // on this menu session, or if the max framerate is set to 30
+            // In addition, tracking only begins after one second in the menu
+            if (!Settings.Instance.IsNoticePermanentlyDismissed(DismissibleNotice.LowPerformanceWarning)
+                && Settings.Instance.Menu3DBackgroundEnabled && canShowLowPerformanceWarning
+                && Settings.Instance.MaxFramesPerSecond != 30)
```
hhyyrylainen, line 293:
```suggestion
                && Settings.Instance.MaxFramesPerSecond > 30)
```

I think this would futureproof the code if we add an extremely low framerate option in the future.

src/general/MainMenu.cs
```diff
@@ -468,6 +499,11 @@ private void RunMenuSetup()
         UpdateStoreVersionStatus();
         UpdateLauncherState();
 
+        if (!Settings.Instance.IsNoticePermanentlyDismissed(DismissibleNotice.LowPerformanceWarning))
```
hhyyrylainen, line 502: I think this if is unnecessary as the actual condition check later will re-check this each frame. Also could the variable be just initialized to true or is there a reason why only after RunMenuSetup it would need to be set to true?

src/general/MainMenu.cs
```diff
@@ -752,6 +788,47 @@ private void WarnAboutNoEnabledMods()
...
+            GD.Print($"Average frame rate is {averageFrameRate}, prompting to disable 3D backgrounds");
+            lowPerformanceWarning.PopupIfNotDismissed();
+            canShowLowPerformanceWarning = false;
+        }
+    }
+
+    private void OnLowPerformanceDialogConfirmed()
+    {
+        Settings.Instance.Menu3DBackgroundEnabled = new SettingValue<bool>(false);
+        Settings.Instance.ApplyGraphicsSettings();
```
hhyyrylainen, line 816: This call should not be necessary as I can't see any settings being applied that are relevant to the 3D backgrounds by this method.

src/general/MainMenu.cs
```diff
@@ -752,6 +788,47 @@ private void WarnAboutNoEnabledMods()
...
+        Settings.Instance.Save();
+
+        // This doesn't seem to be called automatically if this setting is changed outside the designated method
```
hhyyrylainen, line 819: In my earlier comment I talked about the observables breaking on reassigning the option value, this is probably caused by that. And this should be removed once that is fixed.

src/general/MainMenu.cs
```diff
@@ -752,6 +788,47 @@ private void WarnAboutNoEnabledMods()
...
+        if (averageFrameRate == 0)
+            return currentFrameRate;
+
+        // Not an exact average by any means, but good enough for this purpose
+        return (averageFrameRate + currentFrameRate) / 2;
+    }
+
+    private void WarnAboutLowPerformance()
+    {
+        if (averageFrameRate < 30 && secondsInMenu >= 16 && !AreAnyMenuPopupsOpen())
```
hhyyrylainen, line 805: Could you add one more check here to ensure the options menu is not visible? Because saving the options due to the low performance swap when the options menu is open will probably cause issues...

hhyyrylainen (changes requested):

hhyyrylainen (approved):
The code changes look perfect to me now. All this needs to be mergeable is to fix the merge conflicts in translations. The way I recommend doing that is to just accept the master changes for all files except en.po, and after just run the localization script again to get everything up to date and commit that.
I'll try to get someone to playtest this from the team, but if I can't get anyone then I'll playtest this myself in a day or two.

Brabygg:
> Line 6875 has fuzzy (marked needing changes) translation, not allowed for en
>Line 6880 has fuzzy (marked needing changes) translation, not allowed for en

These are not my changes; how'd they even get into master, and what am I supposed to do about them? Do I just leave it like this, remove the fuzzy status, or...?

hhyyrylainen:
I think something went wrong in the merge as this PR is now deleting the cross fadable gallery view file. Also those translations marked fuzzy are because the text is in master but maybe you didn't copy it to your changes when merging? The translation tools automatically create fuzzy translations by guessing based on other translation strings *when new strings are added* those strings are not new meaning that the explanation is that the text for them was not copied from master.

Brabygg:
I think I did indeed mess up the merge somehow.

> The translation tools automatically create fuzzy translations by guessing based on other translation strings when new strings are added those strings are not new meaning that the explanation is that the text for them was not copied from master.

That does make sense. Let me try merging again, and hopefully that should sort things out.

Brabygg:
Seems like the deletion issue was due to the cross fadable gallery view file not being tracked by Git as it was pulled for some reason. This should solve it.

Brabygg:
...I, uh, _actually have no idea_ why the format check is failing. I've committed all changes to loc files, and running that script on my machine doesn't give any errors.

hhyyrylainen:
I think this PR now reverts the last few master branch commits, which I presume is to be the problem (as it probably removes the injectisome feature so the translation system wants to remove the text related to it).

You might be better off creating a new branch from master and then cherry-picking just your commits from this branch that contain the actual work here.

Brabygg:
Tried pulling the master branch, merging locally, and then pushing that. If this fails, I'll see about the other suggestion.

locale/en.po
```diff
@@ -6861,11 +6873,11 @@ msgstr "No Upgrade"
 
 #: ../simulation_parameters/microbe_stage/organelles.json:34
 msgid "UPGRADE_PILUS_INJECTISOME"
-msgstr "Injectisome Pilus"
+msgstr "When in engulf mode ([thrive:input]g_toggle_engulf[/thrive:input]) this type of cilia will cause vortices in the water that pull in nearby objects. Useful for capturing prey."
```
hhyyrylainen, line 6876: This text change is probably unintentional (and also a few lines below this). From what I can see this is the only issue still showing up in the file diff view.

Brabygg: Huh, no idea how that happened. Should be fixed now, though.

hhyyrylainen:
I didn't look through all the translation files, I assume you hopefully correctly copied and updated them from master so that no translation content was destroyed accidentally, but everything else now looks good to me with no merge problems left.

hhyyrylainen (approved):
I've now tested this in the game and this seems to work.

hhyyrylainen:
Okay so I just checked on weblate (https://translate.revolutionarygamesstudio.com/projects/thrive/thrive-game/#history) and this PR messed up a ton of random translations for various languages... So it seems like your use of the translation scripts did not go without problems for other languages either. I hoped you had properly fully reset the files to be what they were in master and then ran the translations update which should not be able to mess anything up. I now need to manually go through the changes history on weblate and undo all of the random changes.

hhyyrylainen:
Seems like the problem was weblate discarding its own changes (that were only on weblate and its own PR and not yet in master branch, so you couldn't have picked up those changes no matter what you did). I'll update weblate and hope it stops doing that. Again, sorry for first assuming that you had done something incorrectly.

PR #4480, opened 2023-09-17
Update organelle tooltip osmoregulation values based on difficulty (and selected membrane)

### LINKED ISSUE #4044: Update organelle tooltip osmoregulation values based on difficulty (and selected membrane)
so that the shown numbers match the actual situation when the organelle is placed

Kashnox:

This PR calculates and sets the osmoregulation cost tooltips for all organelles in the editor after updating the membrane instead of relying on hardcoded values.

**Related Issues**

closes #4044 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


buckly90:
Works perfectly well, with no errors to speak of. A wonderful change.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -90,6 +90,25 @@ private void SetMembraneTooltips(MembraneType referenceMembrane)
             var tooltip = GetSelectionTooltip(membraneType.InternalName, "membraneSelection");
             tooltip?.WriteMembraneModifierList(referenceMembrane, membraneType);
         }
+
+        // Osmoregulation cost is based on the membrane, so update the osmoregulation tooltips
+        // after updating the membrane
+        SetOsmoregulationTooltips();
+    }
+
+    private void SetOsmoregulationTooltips()
```
hhyyrylainen, line 99:
```suggestion
    private void UpdateOsmoregulationTooltips()
```

I think this kind of name would be more consistent with other parts of the editor.

src/microbe_stage/editor/tooltips/SelectionMenuToolTip.cs
```diff
@@ -209,6 +209,19 @@ public void WriteOrganelleProcessList(List<ProcessSpeedInformation>? processes)
...
+    /// <summary>
+    ///   Sets the value of this tooltip's osmoregulation cost modifier.
+    /// </summary>
+    /// <param name="osmoregulationCost">The value to set the osmoregulation cost to.</param>
+    public void WriteOsmoregulationCost(float osmoregulationCost)
+    {
+        foreach (var modifier in modifierInfos)
+        {
+            if (modifier.Name == "osmoregulationCost")
+                modifier.ModifierValue = $"+{osmoregulationCost}";
```
hhyyrylainen, line 221: I think it would make more sense to make this osmoregulation cost modifier like how the mutation cost is set, i.e. so create a new property with a backing field and relevant method to update the value. Then you'll need to go through all the existing tooltips in the godot editor to set the right initial values. Also for display I think the osmoregulation should be rounded to at most 2-3 decimal places.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -90,6 +90,25 @@ private void SetMembraneTooltips(MembraneType referenceMembrane)
...
+    {
+        var organelles = SimulationParameters.Instance.GetAllOrganelles();
+
+        foreach (var organelle in organelles)
+        {
+            float osmoregulationCost = organelle.HexCount * Membrane.OsmoregulationFactor
+                * Constants.ATP_COST_FOR_OSMOREGULATION
+                * Editor.CurrentGame.GameWorld.WorldSettings.OsmoregulationMultiplier;
```
hhyyrylainen, line 107: You should extract this osmoregulation multiplier as a variable here as this needs to look up a ton of object properties to follow the chain to the data on each loop iteration.

hhyyrylainen (changes requested):

src/microbe_stage/editor/tooltips/SelectionMenuToolTip.cs
Kashnox: Fortunately, the tooltips already had the right initial values in the Godot editor.

src/microbe_stage/editor/tooltips/SelectionMenuToolTip.cs
```diff
@@ -157,6 +170,11 @@ public override void _Ready()
         UpdateMpCost();
         UpdateRequiresNucleus();
         UpdateLists();
+
+        // Update osmoregulation cost last after the modifier list has been populated
+        // to make sure this tooltip even has an osmoregulation cost modifier
+        osmoregulationModifier = GetModifierInfo("osmoregulationCost");
```
hhyyrylainen, line 176: This doesn't look like the most bulletproof solution, but I guess it is fine, _Ready is called when the game loads for all tooltips so at least any future crashes this might cause should be easily noticeable.

hhyyrylainen (approved):
This seems fine to me now. @Athriq doesn't seem active enough to complain about how this slightly abuses the tooltip system (by not making a ear-marked property to always hold the osmoregulation number).

Issue #3906, opened 2022-12-04
Multicellular editor can crash when undoing due to System.InvalidOperationException: Sequence contains no matching element


Sur3:
```
Thrive has crashed at 2022-12-04T16:36:17.3887650Z (crash happened immediately before creating this report)
Exit code: 1
Description: Crash is an unhandled exception
==== START OF LAUNCHER OUTPUT ====
Thrive is starting. Log output (due to buffering this is likely not in real time):
Setting breakpad minidump AppID = 1779200
SteamInternal_SetMinidumpSteamID:  Caching Steam ID:  76561197974187562 [API loaded no]
Godot Engine v3.5.stable.mono.thrive.d2ad79a40 - https://godotengine.org
OpenGL ES 3.0 Renderer: GeForce GTX 570/PCIe/SSE2
Async. shader compilation: OFF
 
Mono: Log file is: 'C:/Users/Besitzer/AppData/Roaming/Thrive/mono/mono_logs/2022-12-04_16.56.10_11676.log'
This is Thrive version: 0.6.0.0-rc1 (see below for exact build info)
Unhandled exception logger attached
Startup C# locale is: de-DE Godot locale is: de_DE
user:// directory is: C:/Users/Besitzer/AppData/Roaming/Thrive
Game logs are written to: C:/Users/Besitzer/AppData/Roaming/Thrive\logs latest log is 'log.txt'
Doing delayed apply for some settings
Set audio output device to: Default
Set C# locale to: de-DE Godot locale is: de
Starting Steam load
Steam load finished
Game is owned by current Steam user
Our app id is: 1779200
SimulationParameters are good
This version of Thrive was built at Saturday, 03 December 2022 12:30:42 from commit 9f508dfde02abd66411319256b59aa80bbef107d on branch master
Detected store name from launch parameters: 
Launcher tells us that we are store version: steam
Launcher told us store name: steam
We were opened through the Thrive Launcher
Launcher opening us is hidden
Thrive Launcher started us, launcher hidden: True
Loading mod Nodes into the scene tree
Jukebox now playing from: Menu
Jukebox: starting track: res://assets/sounds/main-menu-theme-1.ogg position: 0
Startup successful, removing startup info file
------------ Thrive Startup Succeeded ------------
Previous patch doesn't exist, despawning all entities.
Applying patch (Dolerurun Gezeitent�mpel) settings
Number of clouds in this patch = 9
Registering new spawner: Name: ammonia density: 1,2
[...]

PR #4481, opened 2023-09-20
ArgumentException: OrganelleTemplate can't be placed at this location when undoing can crash the game

### LINKED ISSUE #3836: ArgumentException: OrganelleTemplate can't be placed at this location when undoing can crash the game
Got a report that the game crashed with the following exception:
```
------------ Begin of Unhandled Exception Log ------------
The following exception prevented the game from running: System.ArgumentException: OrganelleTemplate can't be placed at this location at HexLayout`1[T].Add (T hex) [0x00022] 
in <370f4a10657247d4be2cb49d705e1ae2>:0 at CellEditorComponent.UndoOrganelleRemoveAction (OrganelleRemoveActionData data) [0x00000] in <370f4a10657247d4be2cb49d705e1ae2>:0 at SingleEditorAction`1[T].UndoAction () [0x0000c] 
in <370f4a10657247d4be2cb49d705e1ae2>:0 at CombinedEditorAction.UndoAction () [0x00019] 
in <370f4a10657247d4be2cb49d705e1ae2>:0 at CombinedEditorAction.UndoAction () [0x00019] 
in <370f4a10657247d4be2cb49d705e1ae2>:0 at CombinedEditorAction.UndoAction () [0x00019] 
in <370f4a10657247d4be2cb49d705e1ae2>:0 at ReversibleAction.Undo () [0x00013] 
in <370f4a10657247d4be2cb49d705e1ae2>:0 at ActionHistory`1[T].Undo () [0x00026] 
in <370f4a10657247d4be2cb49d705e1ae2>:0 at EditorActionHistory`1[TAction].Undo () [0x00000] 
in <370f4a10657247d4be2cb49d705e1ae2>:0 at EditorBase`2[TAction,TStage].Undo () [0x00009] 
in <370f4a10657247d4be2cb49d705e1ae2>:0 at EditorComponentWithActionsBase`2[TEditor,TAction].Undo () [0x00007] 
in <370f4a10657247d4be2cb49d705e1ae2>:0 at (wrapper managed-to-native) Godot.NativeCalls.godot_icall_2_695(intptr,intptr,string,object[]) at Godot.Object.EmitSignal (System.String signal, System.Object[] args) [0x0000b] in <19daba5d5ae94c7483bcc78b7907d0ef>:0 at EditorComponentBottomLeftButtons.UndoPressed () [0x00015] 
in <370f4a10657247d4be2cb49d705e1ae2>:0 
Please provide us with this log, thank you. 
------------ End of Unhandled Exception Log ------------
```

Likely caused by a very specific organelle layout and editing steps, I'll try to ask the reporter for any more details as otherwise this is going to be very difficult to track down.

Edit: actually, I was given the full log so it should be possible to read it to find out the organelle placements at least: https://pastebin.com/kRseUFbQ

Kashnox:

This PR stops repositioning cells to their origin during multicellular editing until the entire multicellular editor is done, which prevents hex information in the action history from becoming outdated.

**Related Issues**

closes #3836 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Kashnox:
If anyone tests this PR, watch out for the crash described here:
https://github.com/Revolutionary-Games/Thrive/issues/3906

It has similar steps to reproduce but has a different cause. I'll be looking into that issue next.

src/early_multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -390,6 +390,13 @@ public override void OnFinishEditing()
         editedSpecies.UpdateNameIfValid(newName);
 
         behaviourEditor.OnFinishEditing();
+
+        foreach (var cellType in cellTypeSelectionButtons.Values)
```
hhyyrylainen, line 394: I think this should instead loop through the cell types inside the species to not rely on GUI state being correct (i.e. if a button is missing for some reason or not detected it would still work, and I think going from simulation data -> GUI is a very good rule to follow and not do GUI -> simulation data transfer of information unless really necessary).

src/early_multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -390,6 +390,13 @@ public override void OnFinishEditing()
...
+        {
+            // During multicellular editing, we don't reposition cells to the origin
+            // because it messes up the aciton history, so do it here once editing is finished
+            cellType.CellType.RepositionToOrigin();
```
hhyyrylainen, line 398: This is not fully correct because rotation speed calculation (and maybe some other stats are dependent on organelle positions). So this re-apply of the edit would actually need to basically do everything that CellEditorComponent does on finishing editing.

Kashnox: I double-checked that rotation speed is the only thing in CellEditorComponent.OnFinishEditing that depends on the organelle positions, so I just started updating that in CellBodyPlanComponent.OnFinishEditing too.

Kashnox:
While testing the changes I made to clean up the code, I realized a crash with the same error message could be caused by undoing a move action from another cell, following these steps:

1. Duplicate Stem
2. Move any organelle in Stem
3. Start editing Stem2
4. Undo
5. Observe that the move was not undone
6. Redo
7. Crash

There was already a fix for this applied to most other actions (including redoing a move action), so I used that same fix here. I also verified any other actions that might cause this crash had the fix.

Because PR #4483 is merged into this PR now, the multicellular editor theoretically shouldn't crash at all anymore, at least under reasonable circumstances.

src/microbe_stage/MicrobeSpecies.cs
```diff
@@ -102,6 +102,12 @@ public void CalculateRotationSpeed()
         BaseRotationSpeed = MicrobeInternalCalculations.CalculateRotationSpeed(Organelles);
     }
 
+    public void UpdatePositionalProperties()
```
hhyyrylainen, line 105: I added RepositionToOrigin to the base species type as I think it'll be useful for many stages. Related to that maybe CalculateRotationSpeed should be put in the RepositionToOrigin call (on line 98)? Is there a reason why reposition to origin would be done without also re-calculating the rotation speed?

buckly90:
Failed to reproduce the crash with this PR. Very nice.

src/microbe_stage/MicrobeSpecies.cs
Kashnox: Yeah, I don't think we'd ever want to call one without the other. It seems more likely for someone to do that on accident than on purpose.

src/microbe_stage/MicrobeSpecies.cs
```diff
@@ -95,13 +95,19 @@ public override void OnEdited()
...
         Organelles.RepositionToOrigin();
+        CalculateRotationSpeed();
     }
 
     public void CalculateRotationSpeed()
     {
         BaseRotationSpeed = MicrobeInternalCalculations.CalculateRotationSpeed(Organelles);
     }
 
+    public void RecalculatePositionAndStatistics()
```
hhyyrylainen, line 106: I commented directly on the commit here that I think this method should be removed (all use of this should go through the base Species.RepositionToOrigin method).

Kashnox: MicrobeSpecies implements ICellProperties, so it has to have RecalculatePositionAndStatistics.

I could remove RecalculatePositionAndStatistics from ICellProperties and rework it so RepositionToOrigin calls CalculateRotationSpeed in every class that implements ICellProperties, so RepositionToOrigin would essentially become what RecalculatePositionAndStatistics is now.

Or, we can keep it how it is and accept that if you're interacting with a MicrobeSpecies as a cell, you call RecalculatePositionAndStatistics and if you're interacting with it as a species, you call RepositionToOrigin.

Either way works for me.

hhyyrylainen: I originally designed things to be symmetric between the cell properties and the species. So I'd prefer to just have RepositionToOrigin be the method name both in the species and the ICellProperties.

src/early_multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -390,6 +390,13 @@ public override void OnFinishEditing()
         editedSpecies.UpdateNameIfValid(newName);
 
         behaviourEditor.OnFinishEditing();
+
+        foreach (var cellType in Editor.EditedSpecies.CellTypes)
```
hhyyrylainen, line 394: Looking at the early multicellular species OnEdited method, it seems that is already does RepositionToOrigin for all cell types. So I think this code add here is basically a duplicate operation that is not necessary.

hhyyrylainen (changes requested):
Just one tiny thing I now noticed when looking at the bit wider context. Should be very simple to fix.

src/early_multicellular_stage/editor/CellBodyPlanEditorComponent.cs
Kashnox: Yeah, everything gets repositioned just fine without that, so it wasn't needed.

hhyyrylainen (approved):
This should now be fine. I assume the testing results are still valid so that I can merge this without new testing.

PR #4485, opened 2023-09-21
4355 error for failed save deletion continuation

CoreyHendrey:

This PR continues the work done in PR: #4431

"This PR adds an error message if the save fails to be deleted"
Closes: #4355 

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/saving/SaveManagerGUI.tscn
```diff
@@ -102,6 +97,7 @@ margin_right = 1000.0
 margin_bottom = 561.0
 size_flags_vertical = 3
 SelectableItems = true
+ErrorSaveDeletionFailedPath = null
```
hhyyrylainen, line 100: This seems incorrect to me, this should be causing errors when opening the scene.

CoreyHendrey: Hmmm. It doesn't cause any crashes, I actually can't see where the heck that property even is in the editor.

image

Not sure what's going on there... Something to do with renaming the path variable and Godot freaking out?

CoreyHendrey: Removed the line, testing - everything works as expected.

image
image

hhyyrylainen: I was commenting that in the base scene this is set, but when instanced in this other scene the value is overridden as null, which to me should cause the code for the scene to print an error because it can't find the node from the path as the NodePath is null. 

In the instanced scene node there should be a revert button next to the property which resets it back to the default value that is in the original scene. This is the way you see if a scene instance overrides the default parameters as set in its own scene in Godot.

hhyyrylainen:
It seems like this updates the common submodule, which is fine but I thought it was already up to date on master. Well anyway this needs merge conflicts resolved with master before this is mergeable.

hhyyrylainen (approved):
But the code changes look fine to me now.

src/saving/SaveManagerGUI.tscn
hhyyrylainen: If you only looked at the root node of the scene, then you see the properties of that root node (and that's why you didn't see it). To see the overridden property you need to click on the instanced node name in the scene tree to look at it and not the root node of the scene.

CoreyHendrey: The screenshot posted (Showing the missing property) was inside the SaveManagerGUI scene, and I'd selected the "SaveList" that was part of the scene hierarchy there. :)

hhyyrylainen: Hmm, that should be right. Probably a more niche issue happened (like compilation failed so hot reload didn't work or something like that, or godot editor state got out of sync with the files which can very rarely happen. In that case restarting godot helps it to then properly write out the actual values that should be in the on disk files). I won't spend the time checking out an earlier version of this PR to see what's going on as this was resolved already.

hhyyrylainen:
I fixed the translation merge conflicts so that I can merge this PR.

PR #4486, opened 2023-09-21
Closing modal on "Next" pressed in editors

### LINKED ISSUE #4470: Exiting the editor while organelle popup is open is possible and crashes the game
Using the keyboard to select and click (spacebar) the confirm button while an organelle popup menu allows you to exit the editor and then crash the game when you click something on the popup.

The crashing exception:
```
ERROR: System.ObjectDisposedException: Cannot access a disposed object.
ERROR: Object name: 'CellEditorComponent'.
ERROR:   at Godot.Object.GetPtr (Godot.Object instance) [0x0001c] in <1598dd9b7b214c39a3e18009cd4aff45>:0 
ERROR:   at Godot.Node.AddChild (Godot.Node node, System.Boolean legibleUniqueName) [0x00000] in <1598dd9b7b214c39a3e18009cd4aff45>:0 
ERROR:   at NodeHelpers.ReParent (Godot.Node node, Godot.Node newParent) [0x00028] in <1608dc207f454f80a7587eceebbf12ac>:0 
ERROR:   at ModalManager.UpdateModals () [0x00027] in <1608dc207f454f80a7587eceebbf12ac>:0 
ERROR:   at ModalManager._Process (System.Single delta) [0x00008] in <1608dc207f454f80a7587eceebbf12ac>:0 
```

So this is a pretty clear case where the editor should force close the popup before it allows exiting. Multicellular editor and late multicellular editor probably need the same fix so maybe a base class should be made to have a method that ensures all open popups are closed before exiting.

CoreyHendrey:

Editors will now attempt to close all their open modals before continuing.
It will not let the editor continue if there is an exclusive modal open that refuses to close.

Closes #4470 

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/ModalManager.cs
+++ b/src/gui_common/ModalManager.cs
@@ -99,7 +99,7 @@ public class ModalManager : NodeWithInput
     }
 
     /// <summary>
-    /// Attempt to clear all open modals.
+    ///   Attempt to clear all open modals.
     /// </summary>
     /// <returns>Returns true if all modals were closed, false otherwise.</returns>
     public bool ClearModals()
@@ -133,8 +133,8 @@ public class ModalManager : NodeWithInput
     }
 
     /// <summary>
-    /// Attempts to hide the given modal.
-    /// This won't hide the modal if it's exclusive and doesn't allow closing on escape.
+    ///   Attempts to hide the given modal.
+    ///   This won't hide the modal if it's exclusive and doesn't allow closing on escape.
     /// </summary>
     /// <param name="popup">Modal to hide</param>
     /// <returns>True if the modal was hidden</returns>
--- a/src/gui_common/ModalManager.cs
+++ b/src/gui_common/ModalManager.cs
@@ -95,16 +95,23 @@ public class ModalManager : NodeWithInput
 
         var popup = modalStack.First();
 
-        if (popup.Exclusive && !popup.ExclusiveAllowCloseOnEscape)
-            return false;
+        return HideModal(popup);
+    }
 
-        // This is emitted before closing to allow window using components to differentiate between "cancel" and
-        // "any other reason for closing" in case some logic can be simplified by handling just those two situations.
-        if (popup is CustomWindow dialog)
-            dialog.EmitSignal(nameof(CustomWindow.Cancelled));
+    /// <summary>
+    /// Attempt to clear all open modals.
+    /// </summary>
+    /// <returns>Returns true if all modals were closed, false otherwise.</returns>
+    public bool ClearModals()
+    {
+        if (modalStack.Count <= 0)
+            return true;
 
-        popup.Close();
-        popup.Notification(Control.NotificationModalClose);
+        for (int i = 0; i < modalStack.Count; ++i)
+        {
+            if (!HideModal(modalStack[i]))
+                return false;
+        }
 
         return true;
     }
@@ -123,6 +130,28 @@ public class ModalManager : NodeWithInput
             return modal;
 
         return null;
+    }
+
+    /// <summary>
+    /// Attempts to hide the given modal.
+    /// This won't hide the modal if it's exclusive and doesn't allow closing on escape.
+    /// </summary>
+    /// <param name="popup">Modal to hide</param>
+    /// <returns>True if the modal was hidden</returns>
+    private static bool HideModal(TopLevelContainer popup)
+    {
+        if (popup.Exclusive && !popup.ExclusiveAllowCloseOnEscape)
+            return false;
+
+        // This is emitted before closing to allow window using components to differentiate between "cancel" and
+        // "any other reason for closing" in case some logic can be simplified by handling just those two situations.
+        if (popup is CustomWindow dialog)
+            dialog.EmitSignal(nameof(CustomWindow.Cancelled));
+
+        popup.Close();
+        popup.Notification(Control.NotificationModalClose);
+
+        return true;
     }
 
     private void UpdateModals()
--- a/src/gui_common/ModalManager.cs
+++ b/src/gui_common/ModalManager.cs
@@ -95,25 +95,23 @@ public class ModalManager : NodeWithInput
 
         var popup = modalStack.First();
 
-        return HideModal(popup);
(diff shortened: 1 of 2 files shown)


src/gui_common/ModalManager.cs
```diff
@@ -95,16 +95,23 @@ public bool HideTopMostPopup()
...
 
-        // This is emitted before closing to allow window using components to differentiate between "cancel" and
-        // "any other reason for closing" in case some logic can be simplified by handling just those two situations.
-        if (popup is CustomWindow dialog)
-            dialog.EmitSignal(nameof(CustomWindow.Cancelled));
+    /// <summary>
+    ///   Attempt to clear all open modals.
+    /// </summary>
+    /// <returns>Returns true if all modals were closed, false otherwise.</returns>
+    public bool ClearModals()
```
hhyyrylainen, line 105: Would `CloseAllOpenModals` be a better name?

src/gui_common/ModalManager.cs
```diff
@@ -95,16 +95,23 @@ public bool HideTopMostPopup()
...
+    {
+        if (modalStack.Count <= 0)
+            return true;
 
-        popup.Close();
-        popup.Notification(Control.NotificationModalClose);
+        for (int i = 0; i < modalStack.Count; ++i)
```
Athriq, line 110: You can just use the `for (var modal in modalStack)` expression instead of plain iterator.

src/gui_common/ModalManager.cs
```diff
@@ -125,6 +132,28 @@ public bool HideTopMostPopup()
...
+    /// <summary>
+    ///   Attempts to hide the given modal.
+    ///   This won't hide the modal if it's exclusive and doesn't allow closing on escape.
+    /// </summary>
+    /// <param name="popup">Modal to hide</param>
+    /// <returns>True if the modal was hidden</returns>
+    private static bool HideModal(TopLevelContainer popup)
+    {
+        if (popup.Exclusive && !popup.ExclusiveAllowCloseOnEscape)
+            return false;
```
Athriq, line 144: I don't think the method should take this into account. It makes sense in the context of `HideTopMostPopup` because that method can also be called by pressing ESC. `HideModal` however is a general method, so its hiding behavior should be absolute.

src/gui_common/ModalManager.cs
```diff
@@ -125,6 +132,28 @@ public bool HideTopMostPopup()
...
+
+        // This is emitted before closing to allow window using components to differentiate between "cancel" and
+        // "any other reason for closing" in case some logic can be simplified by handling just those two situations.
+        if (popup is CustomWindow dialog)
+            dialog.EmitSignal(nameof(CustomWindow.Cancelled));
+
+        popup.Close();
+        popup.Notification(Control.NotificationModalClose);
```
Athriq, line 152: Also, we want to make sure this method correctly handle already already-hidden popups. Otherwise, it would potentially cause unexpected behaviors.

Athriq (changes requested):
The overall logical changes makes sense to me, but it had few things that needs to be addressed.

src/gui_common/ModalManager.cs
Athriq: I think the current one is already good. This indeed clears `modalStack` by closing all.

src/gui_common/ModalManager.cs
CoreyHendrey: Moved this into the HIdeTopMostPopup, and changed return type to void (as it can only return success now)

src/gui_common/ModalManager.cs
CoreyHendrey: Updated

src/gui_common/ModalManager.cs
CoreyHendrey: Checking popup visibility now

src/gui_common/ModalManager.cs
```diff
@@ -125,6 +130,25 @@ public bool HideTopMostPopup()
         return null;
     }
 
+    /// <summary>
+    ///   Attempts to hide the given modal.
+    ///   This won't hide the modal if it's exclusive and doesn't allow closing on escape.
```
Athriq, line 135: This bit is no longer true.

Athriq (changes requested):

src/gui_common/ModalManager.cs
CoreyHendrey: Woops! Good catch :)

src/gui_common/ModalManager.cs
```diff
@@ -98,17 +98,22 @@ public bool HideTopMostPopup()
...
-        popup.Notification(Control.NotificationModalClose);
+        HideModal(popup);
 
         return true;
     }
 
+    /// <summary>
+    ///   Attempt to clear all open modals.
+    /// </summary>
+    public void ClearModals()
```
hhyyrylainen, line 109: Does the HideModal method cause a modal to be removed from the stack? If so doesn't the foreach cause a crash by continuing to iterate? Maybe instead the loop should check if the stack is empty and if not attempt to close the top most one.

If this doesn't remove the modals in the stack, then I'm still of the opinion that the name of this method is wrong and should really be something like `ForceHideModals`.

CoreyHendrey: Nah it doesn't remove from the stack, happy to rename it.

hhyyrylainen: Alright, then the foreach doesn't crash. And yeah I'd really like it to be renamed if the method doesn't even "clear" the modals from the data structure.

hhyyrylainen:
What's the status on this? As far as I can quickly see my latest review comments weren't addressed yet.

hhyyrylainen:
There was a fix for the crash, but I think the functionality of requiring to close active modals when an editor is closed would be pretty good, so I'll reimplement that one specific part in another PR so that this our oldest open PR can be finally closed.

hhyyrylainen:
My version of this which I'll merge later today to finally close this PR: https://github.com/Revolutionary-Games/Thrive/pull/5838

Issue #4193, opened 2023-03-01
Make the Thriveopedia search to also search full page text content to help find content


hhyyrylainen:
We'll at some point have so much stuff in there that a full text search feature will be necessary. A basic search for now is probably fast enough going through all of the pages to find the relevant sections for a search term, but the search should probably be done in a background operation. In the future we might need to create a separate tool for generating a page index file.

hhyyrylainen:
There's now page title search implemented but no page content search implemented.

PR #4488, opened 2023-09-21
Thriveopedia basic page search

Oliveriver:

Adds search functionality to the page tree in the Thriveopedia.

Currently only searches by page title, displaying all matching pages and their parents. Could be extended later for more extensive page content searching. The exact search algorithm also has a lot of room for improvement - for now it just checks the case-independent search string appears somewhere in order in a page name.

I decided to make search dynamically update as the user types instead of requiring a button press as it felt a lot nicer to use. I therefore removed the (disabled) search button that was above the page tree.

Empty

h

**Related Issues**

partially addresses: https://github.com/Revolutionary-Games/Thrive/issues/4193

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/thriveopedia/Thriveopedia.cs
+++ b/src/thriveopedia/Thriveopedia.cs
@@ -402,8 +402,8 @@ public class Thriveopedia : ControlWithInput
         {
             var children = pageList.Where(otherPage => otherPage.ParentPageName == page.PageName);
 
-            if (page.TranslatedPageName.ToLower().Contains(newText) ||
-                children.Any(child => child.TranslatedPageName.ToLower().Contains(newText)))
+            if (page.TranslatedPageName.ToLower().Contains(newText.ToLower()) ||
+                children.Any(child => child.TranslatedPageName.ToLower().Contains(newText.ToLower())))
             {
                 // We can assume a page is always before its children in the list of all pages
                 allPages[page] = CreateTreeItem(page, page.ParentPageName);
--- a/src/thriveopedia/Thriveopedia.cs
+++ b/src/thriveopedia/Thriveopedia.cs
@@ -118,8 +118,8 @@ public class Thriveopedia : ControlWithInput
             if (currentGame != null)
             {
                 AddPage("CurrentWorld");
-                AddPage("PatchMap", "CurrentWorld");
-                AddPage("EvolutionaryTree", "CurrentWorld");
+                AddPage("PatchMap");
+                AddPage("EvolutionaryTree");
             }
 
             // Notify all pages of the new game properties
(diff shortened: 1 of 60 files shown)


Oliveriver:
I'm going to guess that there might be a question about why I rebuild the entire page tree every time instead of e.g. setting entries in the existing tree visible/invisible. Unfortunately due to limitations with Godot's `Tree` and `TreeItem`, I think this is the only option.

hhyyrylainen:
I removed the marking as closing the search issue because that issue seems to be about full page text content search so if this only does title search, this is a good start but doesn't address the issue enough to warrant closing it (as I'd need to open a separate issue about full text search within page content).

src/thriveopedia/ThriveopediaPage.cs
```diff
@@ -29,6 +29,8 @@ public abstract class ThriveopediaPage : PanelContainer
     /// </summary>
     public abstract string TranslatedPageName { get; }
 
+    public abstract string? ParentPageName { get; }
```
hhyyrylainen, line 32: Could you add a comment on this what this stores / is? I can't immediately tell from the name of this property what this is for.

src/thriveopedia/Thriveopedia.cs
```diff
@@ -390,6 +392,25 @@ private void OnForwardPressed()
         ChangePage(pageFuture.Pop().PageName, true, false);
     }
 
+    private void OnSearchUpdated(string newText)
+    {
+        pageTree.Clear();
+        pageTree.CreateItem();
+
+        var pageList = allPages.Keys.ToList();
```
hhyyrylainen, line 400: Does this really need to create a copy of the list? Can't the keys collection be used below where ever this variable is used?

Oliveriver: The loop modifies the values of the dictionary. Even though the keys stay the same, I was getting a concurrent modification error without creating a copy.

hhyyrylainen: Ah I see. How do you then ensure that items come back when the search terms are cleared? I thought that `allPages` would be the absolute source of truth in terms of what pages exist and it shouldn't get modified.

Oliveriver: That's what I was hoping, but the issue is that you can't add an existing `TreeItem` to a tree, only create new ones with `CreateItem`. In order for the references to items in the tree to stay correct, I need to replace them with the newly created versions.

None of that would be an issue if you could set the visibility of a `TreeItem` instead of removing/replacing it completely, but you can't do that either.

src/thriveopedia/ThriveopediaPage.cs
Oliveriver: Added.

src/thriveopedia/Thriveopedia.cs
Athriq: > None of that would be an issue if you could set the visibility of a TreeItem instead of removing/replacing it completely, but you can't do that either.

Well, Godot itself uses pretty much the [same approach](https://github.com/godotengine/godot/blob/3.5/editor/filesystem_dock.cpp#L1985) as here by recreating the tree.

buckly90:
Try as I might, I could not break anything. Seems to work well, and will be a great feature for the future. Good stuff!

buckly90:
Works perfectly well in other languages as well, it seems.

Athriq (approved):

src/thriveopedia/Thriveopedia.cs
hhyyrylainen: I see so this works correctly (pages are restored when the search input is cleared)? I was just confused as I thought allPages would be where the actual pages were stored, not just the tree items.

hhyyrylainen (approved):
I still don't understand how the pages come back when the search text is cleared, but everything seems to work right.

hhyyrylainen:
This just needs fixing the translation merge conflicts before this can be merged.

src/thriveopedia/Thriveopedia.cs
Oliveriver: `allPages` is a dictionary and the keys of that dictionary are the source of truth for what pages to recreate when the search text is updated.

hhyyrylainen: I guess I just can't wrap my head around this... because on line 405 the if skips putting some items into allPages there. But then I guess partially the old items are kept there? And that works even though clear has been called on the page tree meaning the tree items should all be deleted.

PR #4490, opened 2023-09-22
Add tutorial for the become multicellular button

### LINKED ISSUE #3567: We need a tutorial for the become multicellular button
we need a tutorial pointing out the become multicellular button, or maybe the editor button when pressed should ask to move to the next stage if it is possible? So that people don't miss it when it becomes available. If the button is not visible (for example due to new game settings), the tutorial should handle that gracefully.

ender-titan1:

This PR adds a tutorial for the "Become multicellular" button

**Related Issues**

Closes #3567

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/tutorial/microbe_stage/BecomeMulticellularTutorial.cs
hhyyrylainen, line 1: Rather than hiding this tutorial after some time, I think this should be shown only after the button has been available for 30 seconds (similarly to the editor button tutorial), and then stay open forever until the player becomes multicellular.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -477,6 +477,13 @@ private void UpdateMulticellularButton(Microbe player)
 
         multicellularButton.Visible = true;
 
+        // Show the become multicellular tutorial
```
hhyyrylainen, line 480: Or is the tutorial about just collecting more colony members? My intention with the issue about the tutorial was that the players can quite easily see the numbers go up on the label on the button as they gather more cells, so that doesn't need explaining really. Instead how to actually trigger the move to multicellular is the part that needs a tutorial. Though I guess it would be worth it to also have a tutorial about how to collect colony members, *but* that tutorial should appear if the player has binding agents and not when they've already managed to form a colony.

buckly90:
Implementation seems to work as intended, nothing noticably wrong.

hhyyrylainen:
I kind of want this tutorial or a separate tutorial for the actual button clicking part when it's available so this'll need a rework and probably another testing go.

src/tutorial/microbe_stage/MicrobeTutorialGUI.tscn
```diff
@@ -218,6 +219,18 @@ margin_right = 0.0
 margin_bottom = 61.0
 Description = "MICROBE_STAGE_DAY_NIGHT_TEXT"
 
+[node name="BecomeMulticellularTutorial" parent="." instance=ExtResource( 10 )]
+visible = true
```
hhyyrylainen, line 223: Shouldn't tutorials be hidden by default like the other ones?

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -51,9 +51,13 @@ public class MicrobeHUD : CreatureStageHUDBase<MicrobeStage>
 
     private ProgressBar ingestedMatterBar = null!;
 
+    private Timer becomeMulticellularTutorialTimer = null!;
```
hhyyrylainen, line 54: This is not the way I've done the other timed tutorials. Please read through `MicrobePressEditorButton` class and implement this feature the same way.

hhyyrylainen (changes requested):

locale/en.po
```diff
@@ -3661,7 +3661,14 @@ msgstr ""
 msgid "MICROBE_STAGE"
 msgstr "Microbe Stage"
 
-#: ../src/tutorial/microbe_stage/MicrobeTutorialGUI.tscn:208
+#: ../src/tutorial/microbe_stage/MicrobeTutorialGUI.tscn:232
+msgid "MICROBE_STAGE_BECOME_MULTICELLULAR_TEXT"
+msgstr ""
+"You have unlocked the ability to progress towards multicellularity. The next step in your journey.\n"
+"\n"
+"To proceed, click the \"Become Multicellular\" button."
```
hhyyrylainen, line 3669:
```suggestion
"To proceed, press the \"Become Multicellular\" button."
```

hhyyrylainen:
The triggering of the tutorial is fine, but there should be an end trigger as well (namely going to the multicellular editor). Right now the box just stays there even after becoming multicellular.

hhyyrylainen (approved):
I implemented the missing end condition for the tutorial (the player becoming multicellular) and tested this pretty thoroughly.

Issue #4003, opened 2023-01-12
Make microbe editor fully usable with a controller


hhyyrylainen:
Placing organelles and modifying placed organelles will need a modal mode where when selecting an organelle type with a controller the focus jumps to moving the hex to be placed with a controller, this should also enable a cancel button to cancel placing and another button to open the right click menu for organelles under the hover hex

PR #4491, opened 2023-09-22
Better upgrade visibility in editor UI

Oliveriver:

* Displays upgrade icon in buttons for organelles that can be modified/upgraded
* Displays modification notice in tooltips for organelles that can be modified/upgraded
* Automatically opens the organelle modify/upgrade UI when a modifiable organelle is placed, under the following conditions:
    * Organelle was not placed with symmetry
    * Organelle placement tutorial is not active

Note this often results in the player having to undo twice if they want to remove a modifiable organelle due to #4091 which sadly still seems to be a bug.

image

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1649,6 +1645,14 @@ private bool AddOrganelle(string organelleType)
                     {
                         usedHexes.Add(hex);
                     }
+
+                    if (organelleDefinition.CanBeModified &&
+                        Symmetry == HexEditorSymmetry.None &&
+                        (TutorialState == null || !TutorialState.Enabled || TutorialState.EditorUndoTutorial.Complete))
```
Oliveriver, line 1651: This feels like a pretty ugly way to check the tutorial state is safe. Is there a better way to do it that I should be aware of? Or should this section be placed somewhere else entirely to avoid it?

hhyyrylainen:
> Automatically opens the organelle modify/upgrade UI when a modifiable organelle is placed, under the following conditions: 

I disagree with this being the default. I think fundamentally players should be fine with playing default organelle state for a while, and upgrades are mostly something different. So prompting immediately to modify a placed organelle doesn't make sense in a lot of cases. For example vacuoles, I don't think players would always want to specialize them, or for the pilus to convert it to deal toxin damage.

There could be an option in the options menu to enable this but I don't think it should be on by default.

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: There is `TutorialState.TutorialActive` to see if any tutorial is currently open.

Oliveriver:
Fair enough. Would automatically opening the right click menu be a sensible compromise? Or should I just change the text in the tooltip to read 'Right click to modify once placed' and leave it as it is otherwise?

hhyyrylainen:
I'm not sure the menu opening would be that good flow either. At least it would require me extra clicks when building specific things to test. Also explicitly referring to right click is not a long term solution (I suggest using as neutral terminology as possible to avoid having a separate text for controller input): https://github.com/Revolutionary-Games/Thrive/issues/4003

Oliveriver:
Right, but I'm just concerned that it's still obscuring information from the player if they aren't shown or told explicity how to modify the organelle. I'll remove the menu auto-opening for now so the tooltip and icon changes can at least go in.

src/microbe_stage/editor/CellEditorComponent.cs
Oliveriver: Removed anyway now so no longer relevant,

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -260,6 +260,7 @@ private void UpdateOrganelleUnlockTooltips()
             var tooltip = GetSelectionTooltip(organelle.InternalName, "organelleSelection");
             if (tooltip != null)
             {
+                tooltip.CanBeModified = organelle.CanBeModified;
```
hhyyrylainen, line 263: Is this change actually required considering the code in `SetupMicrobePartSelections`?
If this is needed, then I think it might warrant renaming this method to something like `UpdateOrganelleTooltips` (as long as there isn't already a method with basically that name).

locale/en.po
```diff
@@ -3892,6 +3892,10 @@ msgstr "Models"
 msgid "MODIFY"
 msgstr "Modify"
 
+#: ../src/microbe_stage/editor/tooltips/SelectionMenuToolTip.tscn:163
+msgid "MODIFY_ONCE_PLACED"
+msgstr "Modify once placed"
```
hhyyrylainen, line 3897: Is it the best that this text tells the player "hey, modify this" instead of being like "Can be modified after placing"? I feel like that's a bit off in terms of tone in a general characteristics info box listing.

src/microbe_stage/editor/tooltips/SelectionMenuToolTip.tscn
```diff
@@ -138,14 +140,36 @@ margin_right = 333.0
...
+expand = true
+stretch_mode = 6
+
+[node name="CanBeModifiedLabel" type="Label" parent="MarginContainer/VBoxContainer/CanBeModified"]
+margin_left = 20.0
+margin_right = 192.0
+margin_bottom = 19.0
+custom_colors/font_color = Color( 1, 0.898039, 0.337255, 1 )
+custom_fonts/font = ExtResource( 10 )
+text = "MODIFY_ONCE_PLACED"
```
hhyyrylainen, line 163: I think the text key would be better to have something like "PART_IS_MODIFIABLE"

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2015,6 +2010,7 @@ private void SetupMicrobePartSelections()
             var control = (MicrobePartSelection)organelleSelectionButtonScene.Instance();
             control.Locked = organelle.Unimplemented;
             control.PartIcon = organelle.LoadedIcon ?? throw new Exception("Organelle with no icon");
+            control.CanBeModified = organelle.CanBeModified;
```
hhyyrylainen, line 2013: Ah yeah, now I remember that this was separate from the tooltip. So this is also needed.
I guess we decided that for now showing the modifiable icon on the organelles that can be modified is fine. It'll be easy enough to make this an option / disable it once almost every single organelle can be modified.

hhyyrylainen (changes requested):
Besides the translation conflict, I'd like to see a tiny bit of technical side clean up for this before merging.

PR #4497, opened 2023-09-28
Fix editor storage statistic not taking specialized vacuole into account

### LINKED ISSUE #4496: Fix editor storage statistic not taking specialized vacuole into account
It should at least show the non-specialized storage like it did before. I'm not sure how it should really show the specialized storage amounts.

ender-titan1:

This PR updates the code the organism statistics panel uses to display storage to display the correct capacity

**Related Issues**

Closes #4496 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/MicrobeInternalCalculations.cs
+++ b/src/microbe_stage/MicrobeInternalCalculations.cs
@@ -41,31 +41,33 @@ public static class MicrobeInternalCalculations
         return organelles.Sum(o => GetNominalCapacityForOrganelle(o.Definition, o.Upgrades));
     }
 
-    public static Dictionary<Compound, float> GetTotalSpecificCapacity(IEnumerable<OrganelleTemplate> organelles)
+    public static Dictionary<Compound, float> GetTotalSpecificCapacity(ICollection<OrganelleTemplate> organelles)
     {
-        var capacities = new Dictionary<Compound, float>();
         var totalNominalCap = 0.0f;
 
         foreach (var organelle in organelles)
         {
-            var capacity = GetAdditionalCapacityForOrganelle(organelle.Definition, organelle.Upgrades);
             totalNominalCap += GetNominalCapacityForOrganelle(organelle.Definition, organelle.Upgrades);
+        }
 
-            if (capacity.Compound == null)
+        var capacities = new Dictionary<Compound, float>();
+
+        foreach (var organelle in organelles)
+        {
+            var specificCapacity = GetAdditionalCapacityForOrganelle(organelle.Definition, organelle.Upgrades);
+
+            if (specificCapacity.Compound == null)
                 continue;
 
-            if (capacities.TryGetValue(capacity.Compound, out var currentCapacity))
+            if (capacities.TryGetValue(specificCapacity.Compound, out var currentCapacity))
             {
-                capacities[capacity.Compound] = currentCapacity + capacity.Capacity;
+                capacities[specificCapacity.Compound] = currentCapacity + specificCapacity.Capacity;
             }
             else
             {
-                capacities.Add(capacity.Compound, capacity.Capacity);
+                capacities.Add(specificCapacity.Compound, specificCapacity.Capacity + totalNominalCap);
             }
         }
-
-        foreach (var compound in capacities.Keys.ToList())
-            capacities[compound] += totalNominalCap;
 
         return capacities;
     }
@@ -108,7 +110,7 @@ public static class MicrobeInternalCalculations
             o => o.Definition.Components.Storage != null).Sum(o => o.Definition.Components.Storage!.Capacity);
     }
 
-    public static float CalculateSpeed(IEnumerable<OrganelleTemplate> organelles, MembraneType membraneType,
+    public static float CalculateSpeed(ICollection<OrganelleTemplate> organelles, MembraneType membraneType,
         float membraneRigidity)
     {
         float microbeMass = Constants.MICROBE_BASE_MASS;
@@ -127,9 +129,7 @@ public static class MicrobeInternalCalculations
         float rightDirectionFactor;
         float leftDirectionFactor;
 
-        var organellesList = organelles.ToList();
-
-        foreach (var organelle in organellesList)
+        foreach (var organelle in organelles)
         {
             microbeMass += organelle.Definition.Mass;
 
@@ -155,7 +155,7 @@ public static class MicrobeInternalCalculations
             }
         }
 
-        var maximumMovementDirection = MaximumSpeedDirection(organellesList);
+        var maximumMovementDirection = MaximumSpeedDirection(organelles);
 
         // Maximum-force direction is not normalized so we need to normalize it here
         maximumMovementDirection = maximumMovementDirection.Normalized();
--- a/src/microbe_stage/MicrobeInternalCalculations.cs
+++ b/src/microbe_stage/MicrobeInternalCalculations.cs
@@ -36,7 +36,38 @@ public static class MicrobeInternalCalculations
         return (Hex.AxialToCartesian(new Hex(0, 0)) - Hex.AxialToCartesian(organelle.Position)).Normalized();
     }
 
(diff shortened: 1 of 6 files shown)


src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -36,6 +36,38 @@ public static Vector3 GetOrganelleDirection(OrganelleTemplate organelle)
         return (Hex.AxialToCartesian(new Hex(0, 0)) - Hex.AxialToCartesian(organelle.Position)).Normalized();
     }
 
+    public static float GetNominalCapacityForOrganelle(OrganelleUpgrades? upgrades, OrganelleDefinition definition)
```
hhyyrylainen, line 39:
```suggestion
    public static float GetNominalCapacityForOrganelle(OrganelleDefinition definition, OrganelleUpgrades? upgrades)
```
I'd prefer the method parameters to be this way around as the definition is the more important parameter, and also upgrades is allowed to be null.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -174,9 +175,42 @@ private void UpdateHitpoints(float hp)
         hpLabel.Value = hp;
     }
 
-    private void UpdateStorage(float storage)
+    private void UpdateStorage(float nominalStorage, Dictionary<Compound, float> storage)
     {
-        storageLabel.Value = (float)Math.Round(storage, 1);
+        storageLabel.Value = nominalStorage;
```
hhyyrylainen, line 180: I think the nominal storage value should still be rounded like it was before.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -174,9 +175,42 @@ private void UpdateHitpoints(float hp)
...
+
+        var tooltip = ToolTipManager.Instance.GetToolTip("storageDetails", "editor");
```
hhyyrylainen, line 182: This will print an error if it can't find the tooltip. If tooltipLoaded is allowed to be false, you should use GetToolTipIfExists instead.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1037,9 +1037,33 @@ public float CalculateHitpoints()
...
+    public float GetNominalCapacity()
     {
-        return MicrobeInternalCalculations.CalculateCapacity(editedMicrobeOrganelles);
+        return editedMicrobeOrganelles.Sum(o => MicrobeInternalCalculations
+            .GetNominalCapacityForOrganelle(o.Upgrades, o.Definition));
+    }
+
+    public Dictionary<Compound, float> GetAdditionalCapacities()
+    {
+        var dict = new Dictionary<Compound, float>();
```
hhyyrylainen, line 1048:
```suggestion
        var capacities = new Dictionary<Compound, float>();
```

"dict" is not a descriptive variable name.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1037,9 +1037,33 @@ public float CalculateHitpoints()
...
+
+        foreach (var organelle in editedMicrobeOrganelles)
+        {
+            var capacity = MicrobeInternalCalculations
+                .GetAdditionalCapacityForOrganelle(organelle.Upgrades, organelle.Definition);
+
+            if (capacity.Compound == null)
+                continue;
+
+            if (dict.TryGetValue(capacity.Compound, out var currentCapacity))
```
hhyyrylainen, line 1058: Multiline if conditions must use braces.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1037,9 +1037,33 @@ public float CalculateHitpoints()
...
+                dict[capacity.Compound] = currentCapacity + capacity.Capacity;
+            else
+                dict.Add(capacity.Compound, capacity.Capacity);
+        }
+
+        var nominalCap = GetNominalCapacity();
+        return dict.Select(e => new KeyValuePair<Compound, float>(e.Key, e.Value + nominalCap))
```
hhyyrylainen, line 1065: Instead of needing to redo the dictionary you could get the nominal capacity at the start of this method and then in the line 1061 do `dict.Add(capacity.Compound, capacity.Capacity + nominalCap)`

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1037,9 +1037,33 @@ public float CalculateHitpoints()
         return maxHitpoints;
     }
 
-    public float CalculateStorage()
+    public float GetNominalCapacity()
     {
-        return MicrobeInternalCalculations.CalculateCapacity(editedMicrobeOrganelles);
+        return editedMicrobeOrganelles.Sum(o => MicrobeInternalCalculations
```
hhyyrylainen, line 1042: I think the internal calculations should still have a method that takes in an enumerable to calculate the nominal capacity, instead of needing to use LINQ here in the editor code.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -174,9 +175,42 @@ private void UpdateHitpoints(float hp)
...
+            description.Append(": ");
+            description.Append(entry.Value);
+        }
+
+        storageLabel.RegisterToolTipForControl("storageDetails", "editor");
+
+        if (tooltipLoaded)
+            tooltip!.Description = description.ToString();
+        else
+            GD.PrintErr("Can't update storage tooltip");
```
hhyyrylainen, line 213: The "RegisterToolTipForControl" already fails if tooltip is not loaded, I assume so you should just early exit and print an error on line 183 (instead of storing the info in a bool).

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -36,6 +36,38 @@ public static Vector3 GetOrganelleDirection(OrganelleTemplate organelle)
...
+        }
+
+        if (definition.Components.Storage == null)
+            return 0;
+
+        return definition.Components.Storage!.Capacity;
+    }
+
+    public static (Compound? Compound, float Capacity)
+        GetAdditionalCapacityForOrganelle(OrganelleUpgrades? upgrades, OrganelleDefinition definition)
```
hhyyrylainen, line 54: I'd like to see the same argument swap here.

hhyyrylainen (changes requested):
The approach looks fine to me but I'd like to get some of the specifics reworked a bit to be more consistent with the other places in the code and some logic related to error handling improved a bit.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
hhyyrylainen: I see later you basically error out if the tooltip is not found, so no reason to swap to GetToolTipIfExists here instead. As it is an error condition to not find the tooltip.

src/microbe_stage/editor/CellEditorComponent.cs
ender-titan1: Problem with that approach is that if two organelles have specific capacity for the same compound, the nominal capacity will be added twice if I understand the suggestion correctly

hhyyrylainen: The TryGetValue returns true *only* when the value exists, so the `Add` call is made only on the first time (when TryGetValue returns false). But feel free to point out if I misunderstood.

ender-titan1: Yeah, you're right, my bad

buckly90:
No apparent errors, works fine. I feel like there could be a more elegant solution to presenting varied storage capacity, but this should suffice for now. Good job!

hhyyrylainen (approved):
I fixed this up a bit by removing two unnecessary places of memory allocation.

PR #4498, opened 2023-09-29
Multicellular Freebuild

Nabstablues:

This PR adds a basic freebuild editor for the early multicellular.

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/GameProperties.cs
```diff
@@ -110,6 +110,13 @@ public static GameProperties StartNewEarlyMulticellularGame(WorldGenerationSetti
 
         game.EnterPrototypes();
 
+        if (freebuild)
+        {
+            game.EnterFreeBuild();
+            game.GameWorld.GenerateRandomSpeciesForFreeBuild();
```
hhyyrylainen, line 116: Thanks for contributing to Thrive. Someone will do a full review once the diff for MainMenu.cs is fixed (it seems you converted all spaces to tabs), but I'll just quickly comment that could you place a todo comment here like: `// TODO: generate some other multicellular species for the player to interact with`.

Kashnox:
I disabled whitespace checking so I could take a look at MainMenu.cs and everything looks good there, apart from the formatting. The only other comment I have is that we should probably rename the old freebuild button and its related methods and fields to include the word "microbe" to match the new multicellular freebuild stuff.

This will be a really helpful feature, by the way. I've been doing a lot of work with the multicellular editor and this will make it much easier to test.

hhyyrylainen:
I'll just point out that the fastest way to test the multicellular editor is to use `res://src/stage_starters/EarlyMulticellularEditorStarter.tscn` as the startup scene. That's even faster than what you'd get by having to interact with a multicellular freebuild button.

Nabstablues:
Yes, but in freebuild you don't have to worry about mp limitations and getting recourses, and this also allows someone to get into the multicellular editor without messing with any code.

hhyyrylainen:
Of course, I wasn't rebuking the usefulness of this additional feature. Just saying that for purely debugging purposes there's those stage starter scenes I've added for easily debugging any stage / editor, which I personally use all the time.

ender-titan1:
This seems to have been abandoned, I could try to finish this if the original author won't do anything with this.

hhyyrylainen:
I think that would be fine if @Nabstablues doesn't reply in a day or so. It'd be great to finish this PR to get this pretty simple change completed and off the list of open PRs.

ender-titan1:
This should be all good now, other than the BOM, which I can't figure out for the life of me how to manually add to a file.

hhyyrylainen:
Depends on your editor but it is usually close to the file encoding settings. Rider has a separate button in the encoding menu to add BOM. Emacs can save a file with a BOM when using an utf8 encoding variant with signature in the name.

hhyyrylainen (approved):
I fixed the MainMenu.cs file by checking it out from master and only letting the changes from this PR through that were intended (I removed the various comment splitting on multiple lines).

Also I fixed keyboard navigation for the new button as it had incorrectly copied settings from the above button.

PR #4499, opened 2023-10-01
Implement a way to edit the cell division order in multicellular

### LINKED ISSUE #3206: Implement a way to edit the cell division order in multicellular
Should be put in the reproduction tab. 

My idea is:
- [ ] when that tab is enabled there would be numbers above all the cell hexes showing the growing order
- [ ] and there would be a list in that tab for the cells that will be added it would be something like "1. Stem at 0,0 2. Some other cell type at 0,1" etc. And the list would have buttons or drag and drop support for reordering
- [ ] and finally there would be a right click context menu to swap any cell and the first cell in a colony, that way the initial cell type to play as can also be swapped. This isn't really critical but would make it more obvious the colony is built around the cell in division order with number 1

Similar to: https://github.com/Revolutionary-Games/Thrive/issues/1622

Kashnox:

This PR gives users the ability to change the order their cells are created during the early multicellular stage.

**Related Issues**

closes #3206

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/HexLayout.cs
```diff
@@ -171,6 +171,17 @@ public bool Remove(T hex)
         return true;
     }
 
+    /// <summary>
+    ///   Moves the item from <see cref="index1"/> to <see cref="index2"/> and the item from <see cref="index2"/> to
+    ///   <see cref="index1"/>.
+    /// </summary>
+    /// <param name="index1">The index of one of the items that should be swapped.</param>
+    /// <param name="index2">The index of the other item that should be swapped.</param>
+    public void SwapIndexes(int index1, int index2)
```
hhyyrylainen, line 180:
```suggestion
    public void SwapIndices(int index1, int index2)
```

I think the plural goes like that for index.

src/general/HexLayout.cs
```diff
@@ -222,7 +237,7 @@ public List<Hex> GetIslandHexes()
         // These are all of the existing hexes, that if there are no islands will all be visited
         var shouldBeVisited = ComputeHexCache();
```
hhyyrylainen, line 238: Why is compute hex cache not given the cutoff parameter here?

src/early_multicellular_stage/editor/action_data/CellReproductionOrderActionData.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    public CellReproductionOrderActionData(int oldIndex, int newIndex)
+    {
+        OldIndex = oldIndex;
+        NewIndex = newIndex;
+    }
+
+    protected override ActionInterferenceMode GetInterferenceModeWithGuaranteed(CombinableActionData other)
+    {
+        return ActionInterferenceMode.NoInterference;
```
hhyyrylainen, line 20: Should this action combine with later reproduction order actions? Similarly to how multiple organelle moves in a row will combine. For example if someone moves the last cell to be the second cell and then moves the second cell to first, that could combine into one reproduction order action, right?

src/early_multicellular_stage/editor/CellBodyPlanEditorComponent.Callbacks.cs
```diff
@@ -8,12 +8,16 @@ public partial class CellBodyPlanEditorComponent
     private void OnCellAdded(HexWithData<CellTemplate> hexWithData)
     {
         cellDataDirty = true;
+
+        UpdateReproductionOrderList();
```
hhyyrylainen, line 12: This should get triggered like the other updates with the cell data dirty. I think there's already a ton of systems that get updated based on the cellDataDirty variable.

src/late_multicellular_stage/editor/MetaballEditorComponentBase.cs
```diff
@@ -565,7 +565,7 @@ protected void OnSymmetryPressed()
         */
     }
 
-    /// <inheritdoc cref="HexEditorComponentBase{TEditor,TCombinedAction,TAction,THexMove}.OnHexEditorMouseEntered"/>
+    /// <inheritdoc cref="HexEditorComponentBase{TEditor,TCombined,TAction,THexMove,TContext}.OnHexEditorMouseEntered"/>
```
hhyyrylainen, line 568: I think I fixed this on my branch where I fix various other things. I think I fixed this at least a bit better as I kept the right `TCombinedAction` name here. So maybe you can just revert this change and leave it be for a bit (I'm hopefult it won't take more than two weeks to get my changes regarding this merged)?

src/early_multicellular_stage/editor/ReproductionOrder.cs
```diff
@@ -0,0 +1,118 @@
+﻿using Godot;
+
+/// <summary>
+///   Handles showing and changing the order in which cells in an early multicellular creature will divide.
+/// </summary>
+public class ReproductionOrder : MarginContainer
```
hhyyrylainen, line 6:
```suggestion
public class ReproductionOrderEditor : MarginContainer
```

I think this kind of name would be a bit better to not confuse this with a pure data class that would just hold the order things divide in.

src/early_multicellular_stage/editor/ReproductionOrder.cs
```diff
@@ -0,0 +1,118 @@
...
+{
+#pragma warning disable CA2213
```
hhyyrylainen, line 8: This warning suppress should not be around the NodePaths. Also the first nodepath should be nullable to detect when this class was instanced in the editor. You'll need to also dispose the nodepaths. Please have a look at our other GUI control types to see how they should be structured.

src/early_multicellular_stage/editor/ReproductionOrder.cs
hhyyrylainen: Reading the changes in CellBodyPlanEditorComponent I now see that, this is just a *single* line in a reproduction order. So this should have a much different name. I think also that it would be worth it to split a new class ReproductionOrderEditor from the body plan editor to simplify that class (preferring composition of GUI elements rather than always adding more and more functionality to a single GUI class). Then this class would be named something like ReproductionOrderCellEntry or something like that to signify that this is just a single element.

src/general/HexLayout.cs
Kashnox: Yeah, looking it up, indices seems to be the preference outside of North America.

Kashnox:
This is ready for review again. After splitting the reproduction order handling into its own class, I ended up reworking it to be generic so it should be easy to reuse it for organelle division order too.

I still haven't implemented any combining for ReproductionOrderActions because the more I thought about it, the more complicated it seemed. I think, if we do end up adding an MP cost to reproduction order changes, we should worry about combining the actions then because we'll have a better idea of how exactly they should be combined. In the meantime, I just left a TODO comment in ReproductionOrderActionData related to it.

The Jetbrains Inspectcode is failing because of that bad reference to HexEditorComponentBase, but everything else is passing.

src/late_multicellular_stage/editor/MetaballEditorComponentBase.cs
```diff
@@ -575,14 +575,14 @@ protected void OnMetaballEditorMouseEntered()
         UpdateMutationPointsBar();
     }
 
-    /// <inheritdoc cref="HexEditorComponentBase{TEditor,TCombinedAction,TAction,THexMove}.OnHexEditorMouseExited"/>
+    /// <inheritdoc cref="HexEditorComponentBase{TEditor,TCombined,TAction,THexMove,TContext}.OnHexEditorMouseExited"/>
```
hhyyrylainen, line 578: There's still changes in this file, so I think you might as well put the correct stuff here (note that due to the line length, there's a need to disable the too long line check for these few lines. You can check my open refactor PR where I did this. Or you could just revert all changes to this file. I think the CI checks will stop checking this file once you make one extra commit after the one reverting the changes in this file.

src/general/base_stage/HexEditorComponentBase.cs
```diff
@@ -1071,6 +1085,7 @@ protected override void Dispose(bool disposing)
                 EditorGridPath.Dispose();
                 CameraFollowPath.Dispose();
                 IslandErrorPath.Dispose();
+                Camera?.Dispose();
```
hhyyrylainen, line 1088: We don't dispose Godot nodes. Only nodepaths are disposed.

src/general/base_stage/HexEditorComponentBase.cs
```diff
@@ -31,6 +31,11 @@ public abstract class
     [Export]
     public NodePath IslandErrorPath = null!;
 
+    /// <summary>
+    ///   The <see cref="MicrobeCamera"/> used by this editor.
+    /// </summary>
+    public MicrobeCamera? Camera;
```
hhyyrylainen, line 37: If this must be made public, could you at least make this a property without a public setter? That way this class state cannot be destroyed by external code swapping out the camera object.

src/early_multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -96,6 +99,9 @@ public partial class CellBodyPlanEditorComponent :
     private PackedScene microbeScene = null!;
 
     private CellPopupMenu cellPopupMenu = null!;
+
+    private ReproductionOrderEditor<CellTemplate, EarlyMulticellularEditor, CombinedEditorAction, EditorAction,
```
hhyyrylainen, line 103: Shouldn't this use the actual type used by the scene (`EarlyMulticellularReproductionOrderEditor`)? That'd simplify the long list of generic parameters defined here.

src/early_multicellular_stage/editor/OrderMarginContainer.tscn
hhyyrylainen, line 1: As this has an attached script of type EarlyMulticellularReproductionOrderEditor this file should be named EarlyMulticellularReproductionOrderEditor.tscn otherwise it is really hard to find what scenes are related to what script classes if they aren't named the same.

src/early_multicellular_stage/editor/ReproductionOrderEntry.cs
```diff
@@ -0,0 +1,134 @@
+﻿using Godot;
+
+/// <summary>
+///   Handles showing and changing the order in which hexes in a hex-based creature will be created.
```
hhyyrylainen, line 4: I think this summary would be work better on the reproduction order editor class. This class for a single entry should have a different summary about the role this class has in the reproduction order editing process.

src/early_multicellular_stage/editor/ReproductionOrderEntry.cs
```diff
@@ -0,0 +1,134 @@
...
+    [Signal]
+    public delegate void OnUp(int index);
+
+    [Signal]
+    public delegate void OnDown(int index);
+
+    /// <summary>
+    ///   When this hex will be created. 1 is the starting hex.
+    /// </summary>
+    public string Index
```
hhyyrylainen, line 35: Why is this a string and not an int?

src/early_multicellular_stage/editor/ReproductionOrderEditor.cs
```diff
@@ -0,0 +1,196 @@
...
+/// </summary>
+public class ReproductionOrderEditor<THex, TEditor, TCombinedAction, TAction, THexMove, TContext> : MarginContainer
+    where THex : IActionHex
+    where TEditor : class, IHexEditor, IEditorWithActions
+    where TCombinedAction : CombinedEditorAction
+    where TAction : EditorAction
+    where THexMove : HexWithData<THex>, IActionHex
+{
+    [Export]
+    public NodePath ReproductionOrderListPath = null!;
```
hhyyrylainen, line 16: First NodePath in a class needs to be nullable to allow detecting when this object is created in the Godot editor and avoiding error conditions happening there. Editor process created class instances never get their Export variables set and don't have _Ready called on them, but *do* have Dispose called on them.

hhyyrylainen:
This now is blocked by the multicellular editor not being currently able to display anything related to the cells. It might take a week or two before I get around to fixing that, but after that this PR can then be finished.

hhyyrylainen:
There's no longer any blockers regarding this, just needs to be finished and updated to work with the latest other code changes.

hhyyrylainen:
I separately implemented a GUI for growth order changing for organelles: https://github.com/Revolutionary-Games/Thrive/pull/5688

It has all the features like drag and drop and I think if this work is resumed it might be a good idea to do this again based on that code.

Issue #3910, opened 2022-12-05
Early multicellular body plan editor and cell editor can end up displaying at the same time


hhyyrylainen:
obviously they shouldn't be able to display at the same time but every now and then someone shows a screenshot of that happening, like this:

![screenshot](https://steamuserimages-a.akamaihd.net/ugc/1988932273942179975/FF6C87E81620740C0C49D68C52E992CC75EF9DC2/?imw=5000&imh=5000&ima=fit&impolicy=Letterbox&imcolor=%23000000&letterbox=false)

This should be easy enough to fix once someone comes up reliable step of instructions that cause the bug to appear.

hhyyrylainen:
Here's potential instructions on how to trigger that (from this video: https://www.youtube.com/watch?v=_G5_akrWCtg):

- Enter early multicellular
- Place another stem cell
- Duplicate the stem cell type
- Place a cell of the duplicated type
- Modify the stem cell type
- Delete an organelle
- Move an organelle
- Switch to the editor tab (by pressing the tab button, not NEXT)
- Place cell of stem type
- Place a cell of the duplicated type
- Modify the duplicated type again
- Do the exact same edits to the modified type now as the stem type
- Add an extra organelle of the removed type (different position than it was removed from)
- Switch back to the editor tab
- Place a cell of the modified type
- Place a cell of the stem type
- Switch back to the cell editor tab
- Move the organelle you last placed to the same position (but different rotation) than it was deleted from earlier
- Switch back to the editor tab
- Try to place a cell of the duplicated type (but fail due to not enough MP)
- Select the stem cell type
- Hit undo
- See the bug with both the cell and body plan editor world objects being visible

84634E1A607A:
A clear seq:

1. Add a cell
2. Move that cell
3. Go to cell editor
4. Hit undo

Kashnox:
The same root issue of the multicellular editor not switching between tabs when actions are undone/redone can lead to a deleted cell type being shown in the Cell tab following these steps:

1. Duplicate any cell type
2. Modify that cell type
3. Undo the duplication while the Cell tab is still open

This problem should also be fixed with this issue.

PR #4504, opened 2023-10-04
Infinite MP achieved by abusing Duplicate and Delete to cell types

### LINKED ISSUE #3912: Infinite MP achieved by abusing `Duplicate` and `Delete` to cell types
To reproduce:

image

1. Do something

image

2. Duplicate cell type
3. Undo all changes (MP regain, but Stem2 still preserves all changes)
4. Loop 1-3 until you are satisfied
5. Add the final cell and delete the initial one. Now you can delete all medium celltypes
6. You can do anything with 25 MP (or less)

image

(Sorry there isn't any more images because I accidentally triggered #3906 and crashed it.)

Kashnox:

This PR makes duplicating and deleting cell types undoable editor actions so trying to undo actions to regain MP will also undo the duplication, preventing the infinite MP exploit. This approach will also made the multicellular editor more manageable by preventing cases where the action to undo/redo exists on a cell type that's been deleted.

**Related Issues**

closes #3912

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
+++ b/src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
@@ -426,7 +426,7 @@ public class EarlyMulticellularEditor : EditorBase<EditorAction, MicrobeStage>, 
         reportTab.UpdatePatchDetails(patch, patch);
     }
 
-    private void OnStartEditingCellType(string name, bool switchTab)
+    private void OnStartEditingCellType(string? name, bool switchTab)
     {
         if (CanCancelAction)
         {
--- a/src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
+++ b/src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
@@ -435,14 +435,14 @@ public class EarlyMulticellularEditor : EditorBase<EditorAction, MicrobeStage>, 
             return;
         }
 
-        var newTypeToEdit = EditedSpecies.CellTypes.First(c => c.TypeName == name);
+        var newTypeToEdit = EditedSpecies.CellTypes.FirstOrDefault(c => c.TypeName == name);
 
         // Only reinitialize the editor when required
         if (selectedCellTypeToEdit == null || selectedCellTypeToEdit != newTypeToEdit)
         {
             selectedCellTypeToEdit = newTypeToEdit;
 
-            GD.Print("Start editing cell type: ", selectedCellTypeToEdit.TypeName);
+            GD.Print("Start editing cell type: ", selectedCellTypeToEdit?.TypeName);
 
             // Reinitialize the cell editor to be able to edit the new cell type
             cellEditorTab.OnEditorSpeciesSetup(EditedBaseSpecies);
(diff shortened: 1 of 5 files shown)


src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
```diff
@@ -435,14 +435,14 @@ private void OnStartEditingCellType(string name, bool switchTab)
             return;
         }
 
-        var newTypeToEdit = EditedSpecies.CellTypes.First(c => c.TypeName == name);
+        var newTypeToEdit = EditedSpecies.CellTypes.FirstOrDefault(c => c.TypeName == name);
```
hhyyrylainen, line 438: Why is this change required? If the new type to edit is not found shouldn't the code print an error, clear the cell type to be edited and then skip the rest of the code?

Kashnox: Using FirstOrDefault saves us from having to handle a null selection as a special case. If I leave that as just First, it'll throw an exception when we emit OnCellTypeToEditSelected with a null or empty name. Until now, we never actually cleared the selection once it was set, so deleted cell types could still be seen in the editor.

hhyyrylainen: See my later comment, I did eventually discover this situation. I'd prefer an explicit check against null at the start of this method which would act as a special value meaning "clear selection" instead of try to pick a cell type like that.

src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
```diff
@@ -435,14 +435,14 @@ private void OnStartEditingCellType(string name, bool switchTab)
...
 
         // Only reinitialize the editor when required
         if (selectedCellTypeToEdit == null || selectedCellTypeToEdit != newTypeToEdit)
```
hhyyrylainen, line 441: The check against null edited cell type, doesn't work as intended here anymore. I think the intention was to force initialize the editing if the previous type to be selected was null (which would mean that now this is used), but with being able to call this with a null type now, this code will always trigger even if the previous type was null and the new type was null.

src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
```diff
@@ -435,14 +435,14 @@ private void OnStartEditingCellType(string name, bool switchTab)
...
         {
             selectedCellTypeToEdit = newTypeToEdit;
 
-            GD.Print("Start editing cell type: ", selectedCellTypeToEdit.TypeName);
+            GD.Print("Start editing cell type: ", selectedCellTypeToEdit?.TypeName);
```
hhyyrylainen, line 445: Also this message should be vastly different if there's no type selected. A message saying "Start editing cell type: " is going to be really confusing to realize that literally nothing there means that previous cell type to be edited was actually cleared, rather than new edit just began.

src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
Kashnox: Yeah, that's a good point

hhyyrylainen (approved):
Code is fine enough now as long as https://github.com/Revolutionary-Games/Thrive/issues/3910 gets done before next release.

PR #4511, opened 2023-10-06
Add icons to tooltips to match new editor icons

ender-titan1:

This PR adds icons to the organelle, rigidity slider and membrane tooltips in the editor. It also changes the layout of some information to match the new editor UI.

**Related Issues**

I bought this concept up on discord and since there were no objections I have implemented it.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/editor/tooltips/ModifierInfoLabel.cs
```diff
@@ -128,6 +129,12 @@ public void AdjustValueColor(float value, bool inverted = false)
         }
     }
 
+    private void AdjustSize(float size)
+    {
+        if (valueLabel != null)
+            valueLabel!.RectMinSize = new Vector2(size, 20.0f);
```
hhyyrylainen, line 135: Shouldn't this null override be completely unnecessary thanks to the if this is inside of?

src/microbe_stage/editor/tooltips/ModifierInfoLabel.cs
```diff
@@ -128,6 +129,12 @@ public void AdjustValueColor(float value, bool inverted = false)
         }
     }
 
+    private void AdjustSize(float size)
```
hhyyrylainen, line 132:
```suggestion
    private void AdjustValueMinSize(float size)
```

hhyyrylainen:
When playtesting I noticed the following error when the game starts:
```
ERROR: (Node not found: "Name" (relative to "/root/ToolTipManager/GroupHolder/organelleSelection/signalingAgent/MarginContainer/VBoxContainer/ModifierList/storage").)
ERROR:    at: get_node (scene/main/node.cpp:1465)
ERROR: (Node not found: "HBoxContainer/Value" (relative to "/root/ToolTipManager/GroupHolder/organelleSelection/signalingAgent/MarginContainer/VBoxContainer/ModifierList/storage").)
ERROR:    at: get_node (scene/main/node.cpp:1465)
ERROR: (Node not found: "Icon" (relative to "/root/ToolTipManager/GroupHolder/organelleSelection/signalingAgent/MarginContainer/VBoxContainer/ModifierList/storage").)
ERROR:    at: get_node (scene/main/node.cpp:1465)
```

Also I think the osmoregulation cost needs a new icon to fit with the rest.

ender-titan1:
> When playtesting I noticed the following error when the game starts:

This should now be fixed. Signaling agent had an old leftover info label node without children.

src/microbe_stage/editor/BarHelper.cs
```diff
@@ -51,7 +51,7 @@ public static Color GetBarColour(SegmentedBar.Type type, string name, bool produ
                     case "baseMovement":
                         return GD.Load<Texture>("res://assets/textures/gui/bevel/baseMovementIcon.png");
                     case "osmoregulation":
-                        return GD.Load<Texture>("res://assets/textures/gui/bevel/osmoIcon.png");
+                        return GD.Load<Texture>("res://assets/textures/gui/bevel/osmoregulationIcon.png");
```
hhyyrylainen, line 54: The file was added with the incorrect case, I see so that's why it fails to load. You should rename the file to "osmoregulationIcon.png" (on windows you probably need to delete the file and then re-make it in a different folder with the right name and then move it back to the bevel folder).

hhyyrylainen (changes requested):
I couldn't read through all the changes in ToolTipManager.tscn but other than that the code changes look fine. I did just run into a new bug.

I think you may have deleted the old osmoregulation icon but didn't update the ATP balance bar. I get an error when running the game and the osmoregulation cost icon is missing from the bar:

```
ERROR: No loader found for resource: res://assets/textures/gui/bevel/osmoregulationIcon.png.
ERROR:    at: _load (core/io/resource_loader.cpp:278)
ERROR: Error loading resource: 'res://assets/textures/gui/bevel/osmoregulationIcon.png'.
ERROR:    at: load (core/bind/core_bind.cpp:76)
```

src/microbe_stage/editor/BarHelper.cs
hhyyrylainen: Also I'll just pre-emptively say that if the icon shows up with colours on the ATP balance bar, then that's inconsistent and a change needs to be done to ensure the icon is grayscale like the other ATP balance icons when it shows up in the balance bar.

ender-titan1:
I renamed the file and changed all references to it accordingly, however it still looks like it's using the old 96 dpi asset. I suspect this is something to do with LFS but I'm not sure. On another note, the ATP bar looks good on my end.

hhyyrylainen:
Does it do that locally or in the devbuild? LFS works by hashing the file, and that hash is the identifier. Unless you have managed the first ever practical collision attack on sha1 of an image file, it should not be possible for the wrong file to be used. Also you can check back a few commits and read the file as a text file to see what it was before. The latest version now is:
```
version https://git-lfs.github.com/spec/v1
oid sha256:14e336ed8dde5ef8c544a98bdb534e26ef96f255c2f674e1ccb41f118e72f431
size 9121
```
`14e336ed8dde5ef8c544a98bdb534e26ef96f255c2f674e1ccb41f118e72f431` is the file hash there that identifies the file to use, and like I said it's not possible for the system to use a wrong file unless there's a hash collision.

ender-titan1:
Ok, so I've tested it and it appears to just happen on my local copy. It looks fine in the DevBuild, and the ATP bar also looks good with the new Icon. Here is the latest screenshot from the DevBuild:

Screenshot (123)

hhyyrylainen:
You can delete your local .import folder to get Godot to re-import all the assets on your computer. Regarding the screenshot, would it be a good idea to add like 2-3 pixels of padding around the images in the ATP balance bar to avoid the icon touching the edges of the balance bar segment it is in as that doesn't look super polished?

hhyyrylainen (approved):
I don't see the padding actually affecting anything in the game... Is the node perhaps used so that the parent container automatically sets the positioning? I guess still it's fine enough so I'll approve this PR.

PR #4513, opened 2023-10-06
Prevent symmetry from creating multiple copies of unique organelle

### LINKED ISSUE #3530: Prevent symmetry from placing multiple unique organelles
it seems right now that you can exploit symmetry to place multiple unique organelles like the nucleus or binding agent

ender-titan1:

This PR makes it so that when adding a unique organelle with symmetry enabled, only one of said organelle will be placed.

**Related Issues**

Closes #3530 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -1661,9 +1661,6 @@ public partial class CellEditorComponent :
                 }
             }, overrideSymmetry);
 
-        if (GetOrganelleDefinition(organelleType).Unique)
-            componentBottomLeftButtons.SymmetryEnabled = true;
-
         if (placementActions.Count < 1)
             return false;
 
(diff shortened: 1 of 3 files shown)


src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1808,6 +1814,9 @@ private void OnOrganelleToPlaceSelected(string organelle)
         if (ActiveActionName == organelle)
             return;
 
+        var organelleDefinition = SimulationParameters.Instance.GetOrganelleType(organelle);
```
hhyyrylainen, line 1817: I found one bug regarding this, if I select a unique organelle after turning on symmetry, then if I save and load the game I end up in a state where the unique organelle is selected to be placed *but* the symmetry button is enabled. I think in the component `Init` around the other post save load actions it'd be good to re-apply the symmetry state. Probably better to write a small helper method that gets called here and there to ensure consistent applying the GUI state.

hhyyrylainen (changes requested):

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: Actually looks like "cytoplasm" selection is done in `UpdateGUIAfterLoadingSpecies` so I think this second call to that new helper method I asked you to make should go there instead as the logical place for it to be.

hhyyrylainen (approved):
That one bug I found is now gone, and the code looks good to me (though I don't fully get the one extra apply symmetry in the editor init but I'll let it be).

PR #4515, opened 2023-10-07
Early multicellular body plan editor and cell editor can end up displaying at the same time

### LINKED ISSUE #3910: Early multicellular body plan editor and cell editor can end up displaying at the same time

Kashnox:

This PR makes the early multicellular editor switch tabs when undoing and redoing actions based on that action's context.

It also copied the fix made to the CellEditorComponent in PR #4468 to fix the warning icon displaying when it shouldn't to the CellBodyPlanEditorComponent because that issue became more noticeable on the early multicellular editor with this PR.

**Related Issues**

closes #3910

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Does this also fix things for late multicellular editor? I assume this might not in which case a new follow up issue needs to be made from https://github.com/Revolutionary-Games/Thrive/issues/3910 for that.

Kashnox:
> Does this also fix things for late multicellular editor? I assume this might not in which case a new follow up issue needs to be made from #3910 for that.

It doesn't fix the late multicellular editor, no.

hhyyrylainen:
Was this ready? It seems this doesn't even have any merge conflicts with the refactoring PR being merged.

hhyyrylainen (approved):
I tested this and this seems fine. I'll make a separate PR applying the same fix to the late multicellular editor as well as the issue this is marked as closing can also happen in late multicellular.

PR #4518, opened 2023-10-10
Add tutorial for negative ATP balance

### LINKED ISSUE #2683: Add tutorial for negative ATP balance
I think a tutorial that pops up the first time the player has negative ATP would be nice.

It could say something like:
> Your ATP balance is negative. This means that when running all processes and moving your creature will run out of energy and die.
Consider adding more ATP producing organelles for the energy sources you plan to use.
Or alternatively you can add storage to make a cell that can only move in bursts.

And then a button to close at the bottom as I don't think there's a really good trigger that would close the tutorial. And I think the tutorial should be positioned on top of the middle of the microbe editor.

A bit wordier than many of the tutorials, so maybe there's a way to make this more concise without removing any advice

This should be relatively easy to implement by just adding the new tutorial class, GUI panel, and new tutorial signal triggered when ATP calculation is negative. ~~Note that this should be only started after the almost complete GUI change PR is complete: https://github.com/Revolutionary-Games/Thrive/pull/2387~~

TrialDragon:

This PR adds a tutorial to explain what negative ATP balance means and what to do about it.

**Related Issues**

Closes #2683

**Progress Checklist**

      break existing features:
      members reviewing this PR)


third_party/FastNoiseLite.cs
hhyyrylainen, line 1: Looks like unintended changes in this file.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1855,6 +1855,11 @@ private void OnOrganellesChanged()
 
         StartAutoEvoPrediction();
 
+        if (energyBalanceInfo != null && energyBalanceInfo.TotalProduction < energyBalanceInfo.TotalConsumption)
```
hhyyrylainen, line 1858: I think it would be better to have a "PlayerEnergyBalance" event that just sends the energy balance object when it is calculated to the tutorial. Then the tutorials can flexibly decide to trigger when the ATP balance warrants an action.

third_party/FastNoiseLite.cs
TrialDragon: The code checker (not sure proper name but the dotnet run --project Scripts check command) automatically made this change when I ran it to ensure my commits were good.

hhyyrylainen: It doesn't do that for me, nor does it do on the CI. It should skip modifying this file. This change should be reverted (unless the code formatting now suddenly does want to modify this file for everyone).

hhyyrylainen:
I think it would be a bit much to tack this tutorial onto such a long chain. I think this tutorial should rather just wait and the next time the player places an organelle and they have negative ATP this tutorial would trigger.

third_party/FastNoiseLite.cs
TrialDragon: Change has been restored. (Tho it still keeps changing that file every time I run it; just not going to stage that change now.)

hhyyrylainen: I just confirmed by doing a clean clone of the Thrive repo and re-running the checks from scratch, and sure enough this happens now. I'll just commit that one change to master right now so you don't have to keep doing the workaround (and everyone else in the future who will clone the code will also hit this problem).

src/tutorial/TutorialEventType.cs
```diff
@@ -142,4 +142,9 @@ public enum TutorialEventType
     ///   The "Become multicellular" button is available to be clicked
     /// </summary>
     MicrobeBecomeMulticellularAvailable,
+
+    /// <summary>
+    ///   The player has achieved a negative ATP balance
```
hhyyrylainen, line 147: This comment is no longer accurate in relation to the latest changes.

src/tutorial/TutorialEventArgs.cs
```diff
@@ -123,3 +123,13 @@ public MicrobeColonyEventArgs(MicrobeColony? colony)
 
     public bool HasColony => Colony != null;
 }
+
+public class EnergyBalanceEventArgs : TutorialEventArgs
+{
+    public EnergyBalanceEventArgs(EnergyBalanceInfo? energyBalanceInfo)
```
hhyyrylainen, line 129: I think the balance should not be nullable here. To me it doesn't make sense to send an event saying "this maybe changed but who knows what it is as it is just null now".

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1855,6 +1855,12 @@ private void OnOrganellesChanged()
 
         StartAutoEvoPrediction();
 
+        if (energyBalanceInfo != null)
+        {
+            TutorialState?.SendEvent(TutorialEventType.MicrobeEditorPlayerEnergyBalanceChanged,
```
hhyyrylainen, line 1860: This triggering of this event should go in `UpdateEnergyBalance` as that is where the updated energy balance is handled, so that is where this logically belongs.

src/tutorial/microbe_editor/NegativeAtpBalanceTutorial.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            gui.NegativeAtpBalanceTutorialVisible = ShownCurrently;
+        }
+
+        public override bool CheckEvent(TutorialState overallState, TutorialEventType eventType, EventArgs args,
+            object sender)
+        {
+            switch (eventType)
+            {
+                case TutorialEventType.MicrobeEditorPlayerEnergyBalanceChanged:
```
hhyyrylainen, line 22: Long cases should be wrapped in braces as per our styleguide.

src/tutorial/microbe_editor/NegativeAtpBalanceTutorial.cs
```diff
@@ -0,0 +1,40 @@
...
+                    var energyBalanceEventArgs = args as EnergyBalanceEventArgs;
+                    var energyBalanceInfo = energyBalanceEventArgs?.EnergyBalanceInfo;
+                    bool isNegativeAtpBalance =
+                        energyBalanceInfo?.TotalProduction < energyBalanceInfo?.TotalConsumption;
+
+                    if (!HasBeenShown && isNegativeAtpBalance && CanTrigger && !overallState.TutorialActive())
```
hhyyrylainen, line 28: This is more of a note for myself, but I'll need to playtest after the event trigger location change to make sure it doesn't take priority from the intended first set of tutorials in the editor.

hhyyrylainen (changes requested):
This now logically looks like a good approach. I'd like to see just a couple of more simple code refactoring changes to up the quality of these changes even more before merging.

src/tutorial/TutorialEventArgs.cs
TrialDragon: I see. Will make it not nullable then. I believe I had done it because multiple other event args used nullable too, but I see that it is not necessary now.

hhyyrylainen: Many of the other event types need to be able to signal data missing. I don't think that is necessary in this case. For example the find glucose tutorial can be signaled that the glucose it was told about is gone and there's no new glucose anywhere. The tutorial reacts to that by temporarily disabling itself.

src/tutorial/TutorialEventArgs.cs
```diff
@@ -123,3 +123,13 @@ public MicrobeColonyEventArgs(MicrobeColony? colony)
...
 }
+
+public class EnergyBalanceEventArgs : TutorialEventArgs
+{
+    public EnergyBalanceEventArgs(EnergyBalanceInfo energyBalanceInfo)
+    {
+        EnergyBalanceInfo = energyBalanceInfo;
+    }
+
+    public EnergyBalanceInfo? EnergyBalanceInfo { get; }
```
hhyyrylainen, line 134:
```suggestion
    public EnergyBalanceInfo EnergyBalanceInfo { get; }
```

hhyyrylainen (approved):
I did a very slight improvement to the code and the tutorial text. I think this is now perfect.

PR #4520, opened 2023-10-13
Fixed organelle place refunding regression

### LINKED ISSUE #4519: Deleting organelles added during current editor cycle no longer refunds them
I just noticed a pretty huge regression: deleting an organelle that was just placed no longer refunds the MP cost of that organelle.

Edit: this may have been caused by the recent PR adding a context for the actions, which might prevent the actions from combining like they did before (even though the context should always be null in the microbe editor).

hhyyrylainen:
and made the editor action context system more reliably set the context data in, hopefully, all cases now

**Related Issues**

closes #4519

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
+++ b/src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
@@ -112,25 +112,18 @@ public class EarlyMulticellularEditor : EditorBase<EditorAction, MicrobeStage>, 
         cellEditorTab.UpdateBackgroundImage(patch.BiomeTemplate);
     }
 
-    public override int WhatWouldActionsCost(IEnumerable<EditorCombinableActionData> actions)
-    {
-        return history.WhatWouldActionsCost(actions);
-    }
-
-    public override bool EnqueueAction(EditorAction action)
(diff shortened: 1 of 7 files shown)


Kashnox:
Setting the context in WhatWouldActionCost feels a little strange, but the alternative is expecting everyone to manually call SetContext before doing anything that would require combining actions, so I think the way it's set up now is good. I didn't see any other issues with the code and I tested the functionality which looks good in the microbe and early multicellular editors.

Kashnox (approved):

hhyyrylainen:
Yeah, that was the alternative design I was considering. When I saw that at least 50+ places create editor actions (and those are all over the place like the popup menu etc.), I decided on this design. It would have been a major pain to re-design the whole thing so that various places could ask the editor object for the current context. It would have been doable, but I'd have really wanted to make an action constructor that requires the context to ensure no one forgot the context anywhere where actions need to be created. 
This is basically just an extension of what you already did to make also the actions that are temporarily created to compare MP changes also get context added. This is important for the popup menu MP costs to actually reflect what will happen.

PR #4522, opened 2023-10-13
Thriveopedia wiki content with organelle pages

Oliveriver:

* Adds wiki updating script to generate JSON file with wiki structure and populate English translations
* Adds architecture for generating wiki pages in game using the JSON wiki
* Fills wiki with all organelle pages (which have minimal content at the moment)

image

image

image

image

image

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Oliveriver:
I should note there were a few bits of the initial plan I left out:

* I didn't extend the script to check that the info box values in the game files and on the wiki match. That can come at a later date.
* The online wiki doesn't support emoji as far as I can tell, but there's [an extension](https://www.mediawiki.org/wiki/Extension:Emoticons) to add them. I think that would be the nicest way to get custom BBCode for compounds, etc., but for now I just used bold text containing the compound name.
* No handling for images, tables, ordered lists, etc. Some of these need to wait for the Godot 4 update as Godot 3 has limited BBCode options.

Scripts/WikiUpdater.cs
```diff
@@ -0,0 +1,379 @@
...
+public static class WikiUpdater
+{
+    private const string ORGANELLE_CATEGORY = "https://wiki.revolutionarygamesstudio.com/wiki/Category:Organelles";
+    private const string WIKI_FILE = "simulation_parameters/common/wiki.json";
+    private const string ENGLISH_TRANSLATION_FILE = "locale/en.po";
+
+    /// <summary>
+    ///   Compounds to replace with custom BBCode when appearing in bold on wiki pages.
+    /// </summary>
+    private static readonly string[] CustomBbcodeCompounds =
```
hhyyrylainen, line 23: Would it be possible to automatically detect this information by reading the compounds.json / en.po to find the English text for the compound names?

Scripts/WikiUpdater.cs
```diff
@@ -0,0 +1,379 @@
...
+
+        var untranslatedWiki = new Wiki(
+            organellesRoot.UntranslatedPage,
+            organelles.Select(o => o.UntranslatedPage).ToList());
+
+        await JsonWriteHelper.WriteJsonWithBom(WIKI_FILE, untranslatedWiki, cancellationToken);
+        ColourConsole.WriteSuccessLine($"Updated wiki at {WIKI_FILE}, running translations update");
+
+        var localizationUpdater = new LocalizationUpdate(new LocalizationOptionsBase { Quiet = true });
+        if (!localizationUpdater.Run(cancellationToken).Result)
```
hhyyrylainen, line 60:
```suggestion
        if (!await localizationUpdater.Run(cancellationToken))
```
Shouldn't this be awaited like this? Or am I misremembering how the run method was defined?

Scripts/WikiUpdater.cs
```diff
@@ -0,0 +1,379 @@
...
+            cancellationToken);
+
+        ColourConsole.WriteSuccessLine("Successfully updated English translations for wiki content");
+
+        return true;
+    }
+
+    private static async Task<TranslationPair> FetchOrganellesRootPage(CancellationToken cancellationToken)
+    {
+        ColourConsole.WriteSuccessLine("Fetching organelles root page");
```
hhyyrylainen, line 76: Status text should use the info or normal text write methods, not the success style prints.

Scripts/WikiUpdater.cs
```diff
@@ -0,0 +1,379 @@
...
+            {
+                case "P":
+                    text = ConvertTextToBbcode(child.InnerHtml) + "\n\n";
+                    break;
+                case "UL":
+
+                    // Godot 3 does not support lists in BBCode, so use custom formatting
+                    text = child.Children
+                        .Where(c => c.TagName == "LI")
+                        .Select(li => $"[indent]•   {ConvertTextToBbcode(li.InnerHtml)}[/indent]")
```
hhyyrylainen, line 176: I added a bullet item translation for the news feed. I think it would be good to use that here for consistency (or to reduce the load on the translators. I've seen at least French using a separate bullet list item character than what's the default for English).

simulation_parameters/GameWiki.cs
```diff
@@ -0,0 +1,42 @@
...
+    public Page OrganellesRoot { get; set; } = null!;
+
+    public List<Page> Organelles { get; set; } = null!;
+
+    public void ApplyTranslations()
+    {
+    }
+
+    public void Check(string name)
+    {
```
hhyyrylainen, line 22: This check method could at least verify there's at least one page, and that there is no SectionBody that is missing in any of the pages.

src/general/utils/TranslationHelper.cs
```diff
@@ -81,6 +81,11 @@ public static string TranslateFeatureFlag(bool enabled)
             TranslationServer.Translate("FEATURE_DISABLED");
     }
 
+    public static string TranslateBoolean(bool value)
```
hhyyrylainen, line 84: I think this might be a bit too general of a thing, above the feature flag translation was added with that specific name to help languages where true/false need to fit the context. I kind of expect a similar thing might be needed here, but I can't say yet for sure as I didn't yet read where this is used.

src/thriveopedia/pages/wiki/ThriveopediaOrganellesRootPage.tscn
```diff
@@ -0,0 +1,20 @@
...
+
+[node name="ThriveopediaPage" instance=ExtResource( 1 )]
+script = ExtResource( 2 )
+MainArticlePath = NodePath("MarginContainer/ScrollContainer/MarginContainer/VBoxContainer/VBoxContainer")
+OrganelleListContainerPath = NodePath("MarginContainer/ScrollContainer/MarginContainer/VBoxContainer/GridContainer")
+
+[node name="VBoxContainer" type="VBoxContainer" parent="MarginContainer/ScrollContainer/MarginContainer/VBoxContainer" index="0"]
+margin_right = 864.0
+
+[node name="GridContainer" type="GridContainer" parent="MarginContainer/ScrollContainer/MarginContainer/VBoxContainer" index="1"]
```
hhyyrylainen, line 14: Would an HFlow container work here? That would allow the columns to appear flexibly, so it might be better in some situations. I haven't tested this yet so I don't know if this suggestion is better or not.

src/thriveopedia/pages/wiki/OrganelleInfoBox.cs
```diff
@@ -0,0 +1,233 @@
...
+        enzymesLabel = GetNode<Label>(EnzymesLabelPath);
+        massLabel = GetNode<Label>(MassLabelPath);
+        sizeLabel = GetNode<Label>(SizeLabelPath);
+        osmoregulationCostLabel = GetNode<Label>(OsmoregulationCostLabelPath);
+        storageLabel = GetNode<Label>(StorageLabelPath);
+        uniqueLabel = GetNode<Label>(UniqueLabelPath);
+        upgradesLabel = GetNode<Label>(UpgradesLabelPath);
+        internalNameLabel = GetNode<Label>(InternalNameLabelPath);
+
+        modelLoadingTexture = GD.Load<Texture>("res://assets/textures/gui/bevel/IconGenerating.png");
```
hhyyrylainen, line 120: Usually we have calls to UpdateModelImage() at the end of ready to ensure the properties being set before ready is called work as well.

src/thriveopedia/pages/wiki/OrganelleLinkButton.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    [Export]
+    public NodePath LabelPath = null!;
+
+#pragma warning disable CA2213
+    private Button button = null!;
+    private Label label = null!;
+#pragma warning restore CA2213
+
+    public OrganelleDefinition Organelle { get; set; } = null!;
```
hhyyrylainen, line 20: I see this quite a bit in the changes here, that there's a public property with a setter that is null overridden. I can't immediately remember that similar things would have been necessary for other GUI parts. In terms of code encapsulation this is kind of ringing alarm bells in my head, but I can't immediately remember how this kind of design is handled elsewhere (as to what could potentially be done).

Scripts/WikiUpdater.cs
Oliveriver: I considered this and agree it's a better long-term solution, but decided not to for now to avoid bloating this PR even more. If you feel it's important I can add it.

Scripts/WikiUpdater.cs
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    /// <summary>
+    ///   Inserts into en.po the English translations for all the translation keys in a list of wiki pages.
+    /// </summary>
+    private static async Task InsertTranslatedPageContent(List<TranslationPair> pages,
+        CancellationToken cancellationToken)
+    {
+        var lines = new List<string>(
+            await File.ReadAllLinesAsync(ENGLISH_TRANSLATION_FILE, Encoding.UTF8, cancellationToken));
```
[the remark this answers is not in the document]
Oliveriver: Well that would have been useful to have noticed...

src/thriveopedia/pages/wiki/OrganelleLinkButton.cs
Oliveriver: I wondered that too, but it's a definitely a pattern that's present elsewhere, e.g. Asset in GalleryCard.

Scripts/WikiUpdater.cs
Oliveriver: I'm not sure I see an obvious way to do that here, since this is the English text, not something I could put a translation key in. Maybe sections could be made a format string with the localised bullet character inserted at runtime. That would open a can of worms in regards to whether I should do the same for other characters.

simulation_parameters/GameWiki.cs
Oliveriver: Added checks for all pages.

src/thriveopedia/pages/wiki/ThriveopediaOrganellesRootPage.tscn
Oliveriver: Good point, changed and it works much better.

src/thriveopedia/pages/wiki/OrganelleInfoBox.cs
Oliveriver: Added.

Scripts/WikiUpdater.cs
Oliveriver: Fixed.

Scripts/WikiUpdater.cs
hhyyrylainen: I won't require it, but at least a TODO comment would be nice. I worry though that by the time we ever add a new compound it's going to be a case of no one remembering this and that one compound being "bugged" in the thriveopedia for a while as no one notices.

src/general/utils/TranslationHelper.cs
hhyyrylainen: I mean it could maybe be something like "PREREQUISITE_REQUIRED" and "PREREQUISITE_NOT_REQUIRED"? I think that would have enough of the context info in it to have accurate translations but still be able to use it for similar precondition required / not required display. Or maybe the icons from the crafting system for having enough of something could be used? That's accessible as `GUICommon.RequirementFulfillmentIconRichText` (though maybe in this case having a slightly different method to directly get the icon path for the fullfilled / missing icon would be fine).

src/thriveopedia/pages/wiki/OrganelleLinkButton.cs
hhyyrylainen: I guess it's fine then if neither of us has a better idea how to do this and this is an approach that's already used.

Scripts/WikiUpdater.cs
hhyyrylainen: I think you should strive to have the text held inside a localized string builder for easy update on language change. That allows easily embedding the bullet point translation followed by the content. I think for now skipping allowing translation special characters that aren't already translated is fine. We can react after the fact to people wanting to translate more text structure elements.

Scripts/WikiUpdater.cs
hhyyrylainen: Well the code like it was before would block the current thread to just wait on that single async operation, rather than yielding the current thread back to the async thread pool, which is the correct thing to do.

Scripts/WikiUpdater.cs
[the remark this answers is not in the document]
Oliveriver: Or not. I keep hitting an impasse here.

That package seems to ignore obsolete values, so while I can just about retain everything else from en.po when I parse and regenerate, I lose the commented out strings at the end of the file.

The other option I found was msgcat, which should be able to combine .po files, but keeps messing up the output by reformatting file location comments. It keeps splitting them over multiple lines so breaks the localization update next time it runs. Maybe there is a way to get it to stop doing that, but [the options are limited](https://www.gnu.org/software/gettext/manual/html_node/msgcat-Invocation.html). The width option doesn't help.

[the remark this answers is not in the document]
hhyyrylainen: I guess it is then unavoidable to need to do a bit of manual handling here. It's at least not too bad to correctly detect the place where text content needs to be inserted into / replace existing lines in the file. Unless removing the old translations is fine, which I've been thinking about cleaning at some point but I suppose it isn't the best if this script always clears those out.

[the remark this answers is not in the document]
Oliveriver: Are you happy to leave this method as it is then? It's been resilient every time I've run the script.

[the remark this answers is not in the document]
hhyyrylainen: I think the overall approach is probably fine, but I might want to comment on some of the specifics. Like it looks like the script reads all of the file into memory, rather than doing an operation like reading line by line and maybe even outputting lines as it goes to another writer (and swapping then the file to the right location after writing the full file). Though I can't exactly remember off the top of my head if there was an async way to read a file line by line. If there isn't then this is probably fine. If I found a way to do that it probably is used somewhere in the file checks with the scripts / a base class in the common repo.

Scripts/WikiUpdater.cs
Oliveriver: One extra point I've just thought of in relation to this: the reason I needed to add the bullet character is that BBCode in Godot 3 doesn't support lists. BBCode in Godot 4 does, so if and when the upgrade happens, ideally we'd replace this with `[li]` elements that Godot 4 presumably has a way to translate.

Not sure I follow your suggestion here. Where would the string builder be?

src/general/utils/TranslationHelper.cs
Oliveriver: I think required is a bit too specific still, since the other use is currently 'Unique' where that wouldn't make sense. Maybe 'CONDITION_SATISFIED' and 'CONDITION_NOT_SATISFIED'?

Scripts/WikiUpdater.cs
hhyyrylainen: Oh right, this is in the script. Not in the game yet so user language can't be known yet.

I think to be consistent with the news feed, this should just use the em-dash or whatever "—" is called. That uses this translation: https://translate.revolutionarygamesstudio.com/translate/thrive/thrive-game/en/?checksum=235288f133b71457 but as this is running before that, it cannot be easily done.

So my suggestion for now is to just replace the `•   ` part with `—` and let other language translators pick a different character if they want.

src/general/utils/TranslationHelper.cs
hhyyrylainen: That's true, it wouldn't work there. How about my suggestion about just using that icon I already put in the game? If there's ever a need to "translate" that image, using it in this context as well would get that localization for free for the Thriveopedia as well as that's needed elsewhere in the game already.

Oliveriver: It looks ok, although I'm not sure green with a white circle is enough to convey 'Yes' in isolation. Next to the red cross it's fine, but on its own the player might assume the circle would be replaced with a tick.

I'll ask on Discord and see what others think.

image

hhyyrylainen: This was brought up before. There wasn't an icon that is universal for the "ok" symbol. I think it is understandable enough given the context and how the player *will* have to understand what the icon means in regards to crafting recipes.

Scripts/WikiUpdater.cs
Oliveriver: Updated to an em dash.

Scripts/WikiUpdater.cs
Oliveriver: Added. I'll take a look at implementing this after this PR.

src/general/utils/TranslationHelper.cs
Athriq: I guess icons are fine. Would be a plus to have the "yes" icon localizable though.

Oliveriver: I've replaced yes/no with icons just to get this moving forward. I would personally still prefer text but we'll see what players think.

Scripts/WikiUpdater.cs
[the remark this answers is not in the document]
Oliveriver: I've rewritten the algorithm to asynchronously run through each line, removing or adding extra lines as needed. Tested on the versions of en.po from this branch and master, and it still produces the correct output in both cases.

buckly90:
Did not notice anything off, everyone seems to work perfectly. Outstanding job!!!

src/thriveopedia/pages/wiki/OrganelleInfoBox.cs
```diff
@@ -0,0 +1,235 @@
...
+        massLabel = GetNode<Label>(MassLabelPath);
+        sizeLabel = GetNode<Label>(SizeLabelPath);
+        osmoregulationCostLabel = GetNode<Label>(OsmoregulationCostLabelPath);
+        storageLabel = GetNode<Label>(StorageLabelPath);
+        uniqueIcon = GetNode<TextureRect>(UniqueIconPath);
+        upgradesLabel = GetNode<Label>(UpgradesLabelPath);
+        internalNameLabel = GetNode<Label>(InternalNameLabelPath);
+
+        modelLoadingTexture = GD.Load<Texture>("res://assets/textures/gui/bevel/IconGenerating.png");
+        UpdateModelImage();
```
hhyyrylainen, line 121: Sorry one more thing here. Shouldn't this also call `UpdateValues`? Because `Organelle` has that, so this still doesn't call all of the setup stuff at the end of _Ready.

src/thriveopedia/Thriveopedia.cs
```diff
@@ -148,6 +153,12 @@ public override void _Ready()
 
         // Add all pages not associated with a game in progress
         AddPage("Museum");
+        AddPage("WikiRoot");
+
+        foreach (var page in ThriveopediaWikiPage.GenerateAllWikiPages())
```
hhyyrylainen, line 158: Would it be better to only generate the wiki pages when the thriveopedia is opened? I fear we'll have hundreds of wiki pages which would get initialized here even if the player doesn't open the thriveopedia in the editor or stage. Right now the stage and editor have separate options menu instances meaning that swapping between the editor and stage re-initializes the entire wiki each time.

src/thriveopedia/Thriveopedia.cs
```diff
@@ -411,6 +452,28 @@ private void OnSearchUpdated(string newText)
         }
     }
 
+    /// <summary>
+    ///   Recursively gets all descendents of this page in the page tree.
+    /// </summary>
+    private List<ThriveopediaPage> GetAllChildren(ThriveopediaPage page)
```
hhyyrylainen, line 458: This creates a bunch of temporary lists. I think a better solution would be to return an IEnumerable and enumerate the child pages like:

```
foreach(var directChild in ...)
{
  foreach (var descendant in GetAllChildren(directChild)
  {
      yield return descendant;
  }
}
```

Scripts/WikiUpdater.cs
```diff
@@ -0,0 +1,415 @@
...
+
+        foreach (var page in pages)
+        {
+            var untranslatedPage = page.UntranslatedPage;
+            var translatedPage = page.TranslatedPage;
+
+            // Translate page names
+            translationPairs.Add(untranslatedPage.Name, translatedPage.Name);
+
+            for (var j = 0; j < untranslatedPage.Sections.Count; j++)
```
hhyyrylainen, line 245: Is there a reason why this uses `j`? I can't see a higher level loop that already used `i` so the first loop using an index variable should use `i`.

Scripts/WikiUpdater.cs
```diff
@@ -0,0 +1,415 @@
...
+        }
+
+        var reader = File.ReadLinesAsync(ENGLISH_TRANSLATION_FILE, Encoding.UTF8, cancellationToken);
+        var writer = new StreamWriter(File.Create(TEMP_TRANSLATION_FILE), new UTF8Encoding(false));
+
+        var isReplacingValue = false;
+        var isFuzzyValue = false;
+        await foreach (var line in reader)
+        {
+            if (line == string.Empty)
```
hhyyrylainen, line 268:
```suggestion
            if (string.IsNullOrEMpty(line))
```

Should be used instead of compare against empty as that's the way it is done in the Thrive codebase.

Scripts/WikiUpdater.cs
```diff
@@ -0,0 +1,415 @@
...
+                    await writer.WriteLineAsync("#, fuzzy");
+                    isFuzzyValue = false;
+                }
+
+                await writer.WriteLineAsync(line);
+            }
+        }
+
+        writer.Dispose();
+        File.Copy(TEMP_TRANSLATION_FILE, ENGLISH_TRANSLATION_FILE, true);
```
hhyyrylainen, line 339: Couldn't this use move with overwrite target file? That way a copy of the file is not needed to be made.

Scripts/WikiUpdater.cs
```diff
@@ -0,0 +1,415 @@
...
+                if (linesToInsert.Length == 1)
+                {
+                    await writer.WriteLineAsync($"msgstr \"{value}\"");
+                }
+                else
+                {
+                    await writer.WriteLineAsync("msgstr \"\"");
+
+                    // Split the content over multiple lines with correct formatting
+                    for (var j = 0; j < linesToInsert.Length; j++)
```
hhyyrylainen, line 317: I'm seeing a j here as well. Also preincrement  (`++i`) should be used instead of postincrement in loops.

Scripts/WikiUpdater.cs
```diff
@@ -0,0 +1,415 @@
...
+            }
+
+            if (!line.Contains("msgid \""))
+            {
+                // Copy existing lines that we don't want to replace
+                await writer.WriteLineAsync(line);
+                continue;
+            }
+
+            var key = line.Split("\"")[1];
```
hhyyrylainen, line 296: This isn't the approach I'd use but this probably works in all cases I can think of. Anyway reading through this code once it seems straightforward that I can probably bugfix this years down the road if necessary.

Scripts/WikiUpdater.cs
```diff
@@ -0,0 +1,415 @@
...
+    {
+        ColourConsole.WriteInfoLine("Fetching organelles root page");
+
+        var body = (await HtmlReader.RetrieveHtmlDocument(ORGANELLE_CATEGORY, cancellationToken)).Body!;
+
+        var sections = GetMainBodySections(body);
+        var untranslatedSections = sections.Select(section => UntranslateSection(section, "ORGANELLES_ROOT")).ToList();
+
+        var untranslatedPage = new Wiki.Page(
+            "WIKI_PAGE_ORGANELLES_ROOT",
```
hhyyrylainen, line 85: Usually in Thrive code we use the style where as many parameters are packed on a single line instead of this "chop" style (as called by Jetbrains). So this isn't exactly consistent coding style with the rest of the codebase, but as this is in the scripts this is a bit less serious of a style difference in my opinion. Up to you if you want to fix this.

hhyyrylainen (changes requested):
I've read through all the changes now and I think overall this is a great new feature. I did notice a few style inconsistencies with the rest of the code base and a couple of non-optimal functional details.

src/thriveopedia/Thriveopedia.cs
Oliveriver: This would lead to a change in the order of pages, with the wiki coming after the current world details in the tree when in game, but I think that would be better actually. I'll make the change.

hhyyrylainen: I guess that makes sense. But it would also be possible for the "static" pages to also only be initialized when required if the order eventually turns into a concern.

Oliveriver: Ah my bad, the only page we need to add in order here is WikiRoot, since the actual wiki pages are added as its children wherever it is in the tree. I think I'll move it anyway since as I said it might be better to show the current world information first.

src/thriveopedia/pages/wiki/OrganelleInfoBox.cs
Oliveriver: Added.

src/thriveopedia/Thriveopedia.cs
Oliveriver: Changed to this approach which (after adding `yield return directChild` as well) seems to behave the same.

Scripts/WikiUpdater.cs
Oliveriver: Sorry, that was an artifact of the nested loops in the previous algorithm that I'd modified. Fixed in both places now.

Scripts/WikiUpdater.cs
hhyyrylainen: I would have used a regex to detect and extract the msgid (as I've written this kind of .po file parsing before for the check scripts) and I would have basically structured a loop to read the file line by line and copy it to the output unless the regex matched. In that case I would have omitted the previous content and written the new data to the file before continuing to loop. It's almost the same as your approach (though I see you handled the fuzzy corner cases, so in the end my simple approach would probably have gotten a bit more complicated when it was hit with real data). What I would have ended up writing would have been pretty close to what you wrote, but I would have used a regex to ensure that the msgid extraction isn't able to go wrong and accidentally process like a commented out line with msgid text on it (which I think your script might be able to hit and get into an incorrect state).

Oliveriver: Ah true, regex would be be more robust than contains. I’ll fix that this evening. Otherwise though I don’t think there’s a real difference in these approaches.

hhyyrylainen (approved):
I no longer have anything to say about the code changes here.

PR #4523, opened 2023-10-13
Modify custom pop ups to adjust to changing heights

### LINKED ISSUE #4510: Organelle context menu doesn't adjust to longer translation text (height doesn't increase)
The organelle right click popup context menu has a height constraint that really messes with the UI when the organelle name takes multiple lines (the modify option is not visible):

problem screenshot

TrialDragon:

This PR modifies custom pop ups so that they adjust to the height of their content.

**Related Issues**
Closes #4510 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


ender-titan1 (commented):
In terms of functionality, I can confirm this works. I have tested it in both Russian and English and the modify menu sizes as expected. One thing I can mention about the code is the that there appears to be a leftover `GD.Print` in `OnOpen`

src/gui_common/CustomPopupMenu.cs
```diff
@@ -13,8 +13,8 @@ public class CustomPopupMenu : TopLevelContainer
     public NodePath ContainerPath = null!;
 
 #pragma warning disable CA2213 // Disposable fields should be disposed
+    protected Container container = null!;
```
hhyyrylainen, line 16: Why are these fields reordered? They no longer match the order of the NodePaths so I think this is a change that makes this worse...

src/gui_common/CustomPopupMenu.cs
```diff
@@ -46,9 +45,43 @@ protected virtual void ResolveNodeReferences()
     {
     }
 
+    /// <summary>
+    ///   Calculates the size the pop should be taking into account minimum
```
hhyyrylainen, line 49:
```suggestion
    ///   Calculates the size the popup should be taking into account minimum
```

Typo

src/microbe_stage/editor/HexPopupMenu.cs
```diff
@@ -119,6 +120,7 @@ public override void _Ready()
 
     public override void _EnterTree()
     {
+        normalFont = GetFont("normal", "Fonts");
```
hhyyrylainen, line 123: This should get the font from the titleLabel in CalculateSize and only use this normal font if there's no font set on the titleLabel.

src/microbe_stage/editor/HexPopupMenu.cs
```diff
@@ -175,6 +177,32 @@ protected override void ResolveNodeReferences()
...
+            // reset size of label to update it
+            // otherwise rect height doesn't update to new value soon enough
+            titleLabel.SetSize(Vector2.Zero);
+
+            size = new Vector2(size.x,
+                size.y +
+                Mathf.Abs(
+                    (titleLabel.GetVisibleLineCount() *
+                        normalFont.GetHeight()) -
+                    titleLabel.RectSize.y));
```
hhyyrylainen, line 194: Also here this code probably fits on at most 2 lines.

src/gui_common/CustomPopupMenu.cs
```diff
@@ -57,8 +90,8 @@ protected override void OnOpen()
     protected override void OnClose()
     {
         var tween = CreateTween();
-        tween.TweenProperty(this, "rect_size", Vector2.Zero, 0.15f)
-            .From(cachedMinSize)
+        tween.TweenProperty(this, "rect_scale", Vector2.Zero, 0.15f)
```
hhyyrylainen, line 93: Why is the scale now tweened instead of the size?

src/gui_common/CustomPopupMenu.cs
```diff
@@ -46,9 +45,43 @@ protected virtual void ResolveNodeReferences()
...
+    /// <summary>
+    ///   Calculates the size the pop should be taking into account minimum
+    ///   height needed by content
+    /// </summary>
+    /// <returns>Returns what should be the size of popup considering content height</returns>
+    protected virtual Vector2 CalculateSize()
+    {
+        RectSize = cachedMinSize;
+
+        var clipControl = GetNodeOrNull<Control>("Panel/Control");
```
hhyyrylainen, line 57: Why this is the "clip" control? This isn't clear at all to me what's going on here and why this uses a nodepath here to get some random child node.

src/gui_common/CustomPopupMenu.cs
```diff
@@ -46,9 +45,43 @@ protected virtual void ResolveNodeReferences()
...
+        var clipControlHeightMargin = 0.0f;
+        if (clipControl != null)
+        {
+            clipControlHeightMargin =
+                Mathf.Abs(clipControl.MarginTop) +
```
hhyyrylainen, line 62: Also could you explain how the math works out correctly if the margins are negative and this takes the absolute values?

hhyyrylainen (changes requested):
I'd like some more code comments clarifying how and *why* this works. And I added just a couple of code wrapping style comments.

hhyyrylainen:
I just forgot to add, but why is it not possible to modify the popup to allow natural sizing? What I mean by that that most Godot Controls correctly adjust to allow their content to fit, and I'm wondering if it would be possible to modify the popup so that the natural sizing would work correctly. That way none of the arcane and bizarre parts I couldn't understand in this PR wouldn't be needed.

src/gui_common/CustomPopupMenu.cs
TrialDragon: It is a node who, in spite of the un-descriptive default name, appears to have sole purpose of clipping the content of the container node and has a margin applied. Said margin is needed to properly size the height of the CustomPopupMenu. I'll fully admit that getting it with the node path was just something to get this functional at the time. I'll make it added via export ala the container and panel nodes if that is desired.

src/gui_common/CustomPopupMenu.cs
TrialDragon: The basics from what I can remember is that both negative and positive margins factor into the parent nodes size as absolute / positive values if you look in the Godot editor. I can not remember why at this time, but I'll look into it and try to provide an explanation tomorrow when I'm not sleepy.

src/gui_common/CustomPopupMenu.cs
TrialDragon: I can't entirely remember if there was a functional reason to change the way this works or if the old way broke stuff. All I can remember is I borrowed it from the CustomWindow's way of tweening for animations. I'll experiment and look into it to see if this really was a necessary change tomorrow, and if not revert it if that's what you'd prefer.

src/gui_common/CustomPopupMenu.cs
hhyyrylainen: I missed that access modifier change. Why is the container needed there? Why not add the code that handles the container sizing directly to this class? For example as a protected method if there's no other way to architect the flow of the required size calculation.

src/gui_common/CustomPopupMenu.cs
hhyyrylainen: That would be better, yes. If the clip control is entirely meant to clip the content, then maybe that is the root cause why the popup doesn't size itself correctly automatically?

src/gui_common/CustomPopupMenu.cs
hhyyrylainen: That would be better, yes. If the clip control is entirely meant to clip the content, then maybe that is the root cause why the popup doesn't size itself correctly automatically?

src/microbe_stage/editor/HexPopupMenu.cs
```diff
@@ -175,6 +177,32 @@ protected override void ResolveNodeReferences()
...
+
+            // make the container visible because for whatever
+            // reason resetting size of text makes it invisible
+            // I don't know why this works, but it just works
+            container.Visible = false;
+            container.Visible = true;
```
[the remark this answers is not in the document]
hhyyrylainen: The container probably sizes itself when it becomes visible. Or something like that. I don't really look forward to trying to port this change to Godot 4 as this seems to rely on really specific Godot under the hood stuff happening. This is why I made that comment that it'd be excellent if there was a way to let the popup naturally size itself correctly as I think that has a much lower chance of something going wrong when we can switch to Godot 4.

src/gui_common/CustomPopupMenu.cs
hhyyrylainen: I did a quick playtest, and this definitely makes the popups look much different. Before they rolled down from the top to display themselves. Now it looks like the whole popup just shrinks down to the top left corner. So the animation was completely changed and it kind of looks like a cheaper effect in my eyes.

hhyyrylainen:
Things being too large are worked around elsewhere by setting the size to 0,0 that way then the minimum size re-applies. I'd prefer having a sensible min size and using that workaround that is used elsewhere in the GUI to get the sizing right. If that approach no longer works correctly, we already have other places in the code that need the same fix as well. With the natural sizing I meant that Godot automatically makes controls fit in their parents, but in this popup case there's something that prevents that from working, which is the reason why we need to keep adding more and more delicate code workarounds. I eventually expect the house of cards to fall down and future changes or porting to Godot 4 becoming impossible without rewriting our whole popup window code.

src/microbe_stage/editor/HexPopupMenu.cs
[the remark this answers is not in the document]
TrialDragon: Thinking about it more, I think I made this change because the changing of size that was there before broke the moment the minimum size wasn't zero, which it wasn't in earlier versions where I experimented with changing minimum size. So can probably be safely reverted presuming the CustomPopupMenu insures the minimum size is always zero.

src/gui_common/CustomPopupMenu.cs
TrialDragon: Thinking about it more, I think I made this change because the changing of size that was there before broke the moment the minimum size wasn't zero, which it wasn't in earlier versions where I experimented with changing minimum size. So can probably be safely reverted presuming the CustomPopupMenu insures the minimum size is always zero.

src/gui_common/CustomPopupMenu.cs
TrialDragon: Thinking about this now that you mentioned it. It probably is the core crux of this issue and removing it probably will fix things in an ever so simpler way. Like, this clip control was probably implemented early on as a conscious design decision to have these pop-ups clip content rather than ever expand. Will experiment tomorrow. If it does fix this to remove this node then I'll probably revert prior commits and only change this then.

Athriq: If I remember right, it's probably me who added that special clipping control. I think the reason was purely to make it possible to animate the menu expanding and closing, otherwise the contents of the context menu holds the container tightly making it impossible to shrink completely. I guess this height not adjusting problem comes as an oversight.

hhyyrylainen:
I wasn't referring to the close animation as such. I was referring to the fact that the normal Godot GUI element sizing is not allowed to happen, and this PR has basically an approximation of what Godot does internally to figure out the minimum size for things, but I'm sure there's various differences which might turn out to be significant in the long run.

Athriq:
Well, I guess it is kind of suboptimal to do that, the culprit needs to be identified first which now seems to have been found and is related to control clipping.

_If_ we really do need to define our own minimum size in cases where it's sensible, Godot does provide an overridable method for that already, I think it's `Control._GetMinimumSize` or something.

TrialDragon:
I tried to use _GetMinimumSize early on, ended up not working due to the the general issue I mentioned at https://github.com/Revolutionary-Games/Thrive/pull/4523#issuecomment-1764019690 (also I now understand what you meant by natural sizing and what I mentioned there most certainly was not it.)

Basically trying to calculate the custom minimum size based on the size of content means minimum size can only get bigger due to it refusing to allow the regular sizing to shrink (ala the natural sizing you mentioned).

Tho, using _GetMinimumSize for stuff other than directly calculating based off of size of children is probably fine, or in the case where you do want to calculate size based off of children ala CustomWindow it is probably fine if you don't mind the fact that it probably can't be resized properly (tho maybe the issues I encountered regarding that were actually due to my implementation, who knows.)

src/gui_common/CustomPopupMenu.cs
TrialDragon: Apparently removing the node that clips content is not the end of the issue. Parent control nodes do not expand or grow to fit size of children. Overall dynamic sizing of controls isn't easily possible in Godot apparently outside of containers and their logic, hence the workarounds done before.

src/gui_common/CustomPopupMenu.cs
TrialDragon: So, looking into it here is the logic for why:

Putting a negative margin on a node removes so many units from the size, however to keep the size of the parent the same Godot then adds that many units to the parent node. Given in this situation the size of the child-most node of CustomPopupMenu is the one that is being considered for the base size of the content we have to manually add the positive and negative margins manually to the total size to get the parent nodes approximate size rather than relying on Godot to do this calculation for us. 

Hopefully that makes sense.

src/microbe_stage/editor/HexPopupMenu.cs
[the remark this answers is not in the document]
TrialDragon: So, still not entirely sure why this works, but better understand what it is doing. Apparently resetting the size gives the vbox container's content absurd margins that push them well out of the content clipping area. Setting visibility appears to fix the margins to where they should be. Why it does this, I don't know. My guess is it has to do with the canvas item being updated / redrawn on being made visible when it wasn't visible before.

src/gui_common/CustomPopupMenu.cs
TrialDragon: The reason is that the title text which is a child of the container only exists in the OrganellePopupMenu. Though, considering this now, I probably should've just used an exported node path in a derivative class to get that instead of switching this to protected. Will fix this.

src/microbe_stage/editor/HexPopupMenu.cs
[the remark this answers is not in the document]
TrialDragon: Hmm, only other thing I've found is others apparently experiencing quite similar issues; even in Godot 4 apparently. (Maybe it's been fixed in later versions tho?)

https://ask.godotengine.org/142597/there-make-container-panel-expand-when-adding-more-children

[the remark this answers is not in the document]
TrialDragon: Mm, wait. Maybe relevant issue comment in the Godot engine. Mentions size update relying on drawing which does fit with the problem here and inexplicable solution. So the hiding and showing of the control is probably working because it is more or less forcing a size update due to redraw? Not sure, but it is something perhaps in the vaguely right direction. https://www.github.com/godotengine/godot/issues/20623#issuecomment-1508910488

src/gui_common/CustomPopupMenu.cs
TrialDragon: Protected method works, tho I don't know what to necessarily call it due to the lack of understanding of what the code does. Best guess currently to name it is ForceSizeUpdate or ForceContainerRedraw since those are the two things names that make sense in current guesses of what this is doing. (I'm partial to the latter given it is more sure that it is at least doing that at this time compared to the former which is maybe a side effect of the redraw which may be what is used.)

src/gui_common/CustomPopupMenu.cs
hhyyrylainen: I was thinking that *why* we aren't using a container here... Would it be possible to switch this to use a container to ensure the sizing is automatically correct? That would basically solve all the problems of trying to fight with the way the Godot GUI system works here. Though I don't know what's the original reason why a container was not used here.

src/gui_common/CustomPopupMenu.cs
hhyyrylainen: I think that does kind of make sense. Could you add a code comment with this info so that someone reading the code later can also get this information?

src/microbe_stage/editor/HexPopupMenu.cs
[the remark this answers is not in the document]
hhyyrylainen: Yeah, I think that makes sense. I was thinking along the same lines that Godot when something becomes visible performs various actions to fix it for display.

src/gui_common/CustomPopupMenu.cs
hhyyrylainen: Maybe something like `WorkaroundContainerSizeProblem`? That'd be very explicit that the method exists just to get around a problem that should not exist in an ideal world.

src/gui_common/CustomPopupMenu.cs
TrialDragon: Yep, will do.

src/gui_common/CustomPopupMenu.cs
TrialDragon: My off the tip of the head guess at why containers weren't used has to do with the fact that a panel is also used. As far as I can remember and know there is no default container that works well in having a panel as a background for content more or less. Will check to see if there actually is one however. Tho, in case there isn't one that works in that situation I do wonder if we can just have the panel as a sibling to the vbox and have the root control node of CustomPopupMenu more or less act as a container changing the panel to match vbox size. Or maybe just make a generic container that does similar logic. Mm, not sure if, or how that will work. Will have to look into it probably.

src/gui_common/CustomPopupMenu.cs
```diff
@@ -9,11 +9,15 @@ public class CustomPopupMenu : TopLevelContainer
     [Export]
     public NodePath? PanelPath;
 
+    [Export]
+    public NodePath? ClipControlPath;
```
hhyyrylainen, line 13: Regarding this NodePath as this is not the first nodepath this should not be nullable (this should be `    public NodePath ClipControlPath = null!;` like ContainerPath is below this). Also if I saw correctly a dispose call is missing for this.

src/gui_common/CustomPopupMenu.cs
TrialDragon: I'm not sure that would work. Changing size or position of the panel as child of a container could maybe cause active fighting against the container logic I think. Like, if we up the size it may up the size of the container which could be bad because that maybe will affect how the logic will treat the content such as buttons and labels which may not be what we want. If position is changed, I'm not sure how that'll interact with the container either. Like will it consider the space where it thinks the panel is when calculating where the content should be and cause issues and leave empty space there?  Maybe these are just unnecessary concerns; won't know until an implementation is tried.

hhyyrylainen: If the container doesn't automatically resize the background to fit, then it's just a simple case of manually positioning the background node. I've used that approach before where I had controls that weren't sized and positioned by their parent (this was fully intentional as these places required special considerations). Godot Controls that don't get managed by a parent container can be manually moved without problems. For example this approach is used to overlap two controls on the same visual space when needed. Setting the correct mouse filters makes the cursor click the right elements. Even if the container does set the child sizing, it is possible to override that after it is set by registering the right callbacks, AFAIK this also doesn't leave to fighting as containers only reposition their children when necessary i.e. they don't check each frame that things are right.

src/gui_common/CustomPopupMenu.cs
TrialDragon: Yeah, forgot about dispose call. Getting used to how C# does things. Fixed now.

hhyyrylainen: All good. I was more surprised to see the CI check not detecting that undisposed field for some reason... I have no clue why it didn't pick up on that, the jetbrains analyzer should have caught it.

TrialDragon: Not sure either. Didn't catch it when I ran the code checker locally either.

src/gui_common/CustomPopupMenu.cs
TrialDragon: Mm, good news. There is apparently a container we could use, and it would actually remove the need to have a panel node at all. PanelContainer. It draws the panel behind the content with a stylebox. Should be functionally ideal if I read the docs correctly. The only thing that I imagine would be annoying about switching the nodes is that modifying inherited scenes in Godot is an utter trifle and not well supported in the editor if I remember correctly.

hhyyrylainen: Doesn't it just work by switching the node type in the scene? I doubt the change would need to be made in more than 2-3 files (.tscn and the C# file to specify the new base class). I've switched scene root node types before and it's not been a problem really.

TrialDragon: Not exactly. You'd also have to remove the panel node since that is no longer need with PanelContainer which may present problems with inherited scenes. Inherited scenes are... well not easy to work with. Like changing the type of the root node can sort of work; you just have to change it in every instance since that afaik doesn't update when the base scene gets changed. If you change the hierarchy of nodes, like removing the panel and clip making container child of root, stuff will not update correctly in the inherited scene if done normally from editor and you have to dig into the tscn file and manually mess with stuff instead which can be a hassle.

Overall inherited scenes are like making the children of a scene instance editable. If you seriously need it for stuff to function, then go ahead as a last resort, but if you're able to avoid it then alternatives are probably preferred 'cause otherwise stuff may be hard to maintain in long run when things change.

Tho maybe I've heard and learned wrong. Generally I've heard that composition is preferred over inheritance when dealing with Godot nodes if possible.

hhyyrylainen: Unless I misremember, we have just 2-3 inherited scenes from the popup. Everything else is just scene instances where I don't think many places would have editable child nodes enabled. So I don't foresee that this change would be that much work.

hhyyrylainen: It'll be also much easier to fix and do properly now rather than in 3 years when we have much more game done.

TrialDragon: Well, I've made progress. I'm fairly certain that changing the panel node to a PanelContainer and removing the clip control node should do intended behavior. Can't test however because somehow, some way, some change I've made in the process of this is now causing LineChart.cs lines 690 and 831 to give a null reference exception which is unexpected completely. Not sure how this is even affecting LineChart given it doesn't even use a CustomPopupMenu for it's popup or anything like that.

TrialDragon: ...and now reloading the editor it's fixed itself? I honestly have no clue, but it is lightly concerning that a random null reference appeared and disappeared without logical reason.

src/gui_common/CustomPopupMenu.cs
TrialDragon: Do you want this removed / reverted?

hhyyrylainen: Is it possible to switch back to the old size based animation? If so, I'd probably prefer that over the scale based animation.

TrialDragon: Probably possible? I don't think the way the pop-ups currently work really rely on custom minimum size, so it should work. Tho it's been a week, I'll check and try tomorrow to see if it will work.

TrialDragon: So... in terms of end result working meaning it goes to the proper size at the end of the animation; it works. But with how the new way it sizes itself with the panel container and all that the animation looks objectively horrible compared to how it was before any changes. I'm not really sure there is a way to make the size changing version of the animation look better really without potentially just over complicating the whole way of sizing popup menus again.

[Here's an example of the issue on Imgur](https://imgur.com/a/eDpSaxa)

hhyyrylainen: Okay, yeah that looks bad and was probably the reason why the size of the popups was overcomplicated. Let's just stick to the scale animation then as that allows the natural container sizing with the popup which solves a lot more headaches than it causes by having a bit worse animation.

src/gui_common/CustomPopupMenu.cs
TrialDragon: So, as far as I know this is the last of the outstanding changes requested so far. Is there any changes you want made or bugs found that need to be fixed?

hhyyrylainen: I'll read through the PR changes again today. I hopefully won't spot anything so that this can be merged (as it looks like there are no merge conflicts with master even with my huge refactor being merged).

src/gui_common/CustomPopupMenu.cs
Athriq: Can I see a gif showing how this animation looks in action? My local working git branch is a giant mess of uncommitted changes right now is so I can't easily checkout branches.

TrialDragon: @Athriq Here. 
[Scale-based animation example on Imgur](https://imgur.com/a/5bGWAON)

src/gui_common/CustomPopupMenu.cs
```diff
@@ -48,7 +43,9 @@ protected virtual void ResolveNodeReferences()
 
     protected override void OnOpen()
     {
-        CreateTween().TweenProperty(this, "rect_size", cachedMinSize, 0.2f)
+        // RectSize = CalculateSize();
```
hhyyrylainen, line 46: I guess this comment is fine to leave here as a sign that this used to work differently...

hhyyrylainen (approved):
The code now looks fine. I'll guess that this also works enough in game to merge. As 0.6.4 has the huge stage refactor all features have to be re-tested so any problems with this shold be caught then at the latest.

src/gui_common/CustomPopupMenu.cs
Athriq: Well, yeah that looks good enough to me.

src/gui_common/CustomPopupMenu.cs
TrialDragon: O shoot, I forgot that was in there. Wasn't intended to be left in, sorry.

hhyyrylainen: If it is not too much of a bother, could you make a PR removing that line? It'll be very easy to approve and merge.

PR #4526, opened 2023-10-15
Add context support for Jukebox and add new ambient tracks

### LINKED ISSUE #3586: Add a system for music tracks to specify they are context specific
so for patch specific music we need a new feature in the jukebox to only play certain tracks in a given context.

To start off I think we should make it so that there's a list of active contexts in the jukebox, those context are mandatory to give as a parameter when changing category (other contexts would be cleared). Then the tracks in a category are modified so that they have an optional field called something like "ExclusiveToContext" (type should probably be a list of strings) that way certain tracks would only be allowed to be played when one of their exclusive context tags would be set. One alternative design would be to have an enum with the possible context values, which I think might be nice as that would avoid typos in the track configurations, which would be pretty hard to test for (to notice that one track never gets played).

In the future we can expand this system to add extra features like exclusive tracks being able to specify they take priority over normal tracks when they can be played, and maybe setting a context (like in combat) could force a track change early.

### LINKED ISSUE #4218: Add the new ice and vents biome ambient tracks to the game
New ice ambiance track was created: https://dev.revolutionarygamesstudio.com/api/v1/download/39136
and a new vents ambiance: https://dev.revolutionarygamesstudio.com/api/v1/download/39134

To be able to add these, we need the system for playing context specific things with the Jukebox: https://github.com/Revolutionary-Games/Thrive/issues/3586

ender-titan1:

This PR does adds functionality to the Jukebox to play only specific tracks based on context provided when selecting categories as well as adding two new ambient tracks to the microbe stage to the vents and ice shelves using this new functionality.

**Related Issues**

Closes #3586 
Closes #4218

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/Jukebox.cs
+++ b/src/general/Jukebox.cs
@@ -38,6 +38,8 @@ public class Jukebox : Node
     /// </summary>
     private MusicCategory? previouslyPlayedCategory;
 
+    private MusicContext[] activeContexts = null!;
+
     /// <summary>
     ///   Loads the music categories and prepares to play them
     /// </summary>
@@ -108,8 +110,9 @@ public class Jukebox : Node
     ///   Starts playing tracks from the provided category
     /// </summary>
     /// <param name="category">category from music_tracks.json</param>
-    public void PlayCategory(string category)
+    public void PlayCategory(string category, MusicContext[]? contexts = null)
     {
+        activeContexts = contexts ?? new MusicContext[] { MusicContext.General };
         PlayingCategory = category;
         Resume();
     }
--- a/src/general/Jukebox.cs
+++ b/src/general/Jukebox.cs
@@ -66,6 +66,15 @@ public class Jukebox : Node
                 throw new ArgumentException("Playing category can't be set to null");
 
             GD.Print("Jukebox now playing from: ", value);
+            GD.Print("Active Music Contexts: ");
+
+            var contexts = string.Empty;
+
+            foreach (var ctx in activeContexts)
+                contexts += ctx.ToString() + ", ";
+
+            GD.Print(contexts);
+
             playingCategory = value;
             OnCategoryChanged();
         }
--- a/src/general/Jukebox.cs
+++ b/src/general/Jukebox.cs
@@ -478,7 +478,7 @@ public class Jukebox : Node
         {
             foreach (var list in target.TrackLists)
             {
-                foreach (var track in list.Tracks)
+                foreach (var track in list.GetTracksForContexts(activeContexts))
                 {
                     // Resume track (but only one per list)
                     if (track.WasPlaying)
@@ -503,7 +503,7 @@ public class Jukebox : Node
 
         foreach (var list in target.TrackLists)
         {
-            var trackResources = list.Tracks.Select(track => track.ResourcePath);
+            var trackResources = list.GetTracksForContexts(activeContexts).Select(track => track.ResourcePath);
             if (activeTracks.Any(track => trackResources.Contains(track)))
                 continue;
 
@@ -514,7 +514,7 @@ public class Jukebox : Node
 
         foreach (var list in needToStartFrom)
         {
-            if (!list.Repeat && list.Tracks.All(track => track.PlayedOnce))
+            if (!list.Repeat && list.GetTracksForContexts(activeContexts).All(track => track.PlayedOnce))
                 continue;
 
             PlayNextTrackFromList(list, index =>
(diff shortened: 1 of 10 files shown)


src/microbe_stage/MicrobeStage.cs
```diff
@@ -279,9 +279,19 @@ public override void StartNewGame()
 
     public override void StartMusic()
     {
+        var biome = CurrentGame!.GameWorld.Map.CurrentPatch!.BiomeTemplate;
+
+        var context = new[] { MusicContext.General };
+
+        if (biome.InternalName == "aavolcanic_vent")
+            context[0] = MusicContext.PatchVents;
```
Kashnox, line 287: If it's not too much work, I think it would be nice to let the biomes define for themselves which contexts they support in the json file instead of basing it off the name. That way, we don't have to come back and modify this script whenever we add a new biome or add new contexts for existing biomes.

Kashnox (changes requested):

src/microbe_stage/MicrobeStage.cs
hhyyrylainen: I think that's a great idea to reduce the hardcoding here, and I think I probably had the same idea in mind when thinking about this contexts feature years ago.

src/general/Jukebox.cs
```diff
@@ -64,6 +66,15 @@ public string PlayingCategory
                 throw new ArgumentException("Playing category can't be set to null");
 
             GD.Print("Jukebox now playing from: ", value);
+            GD.Print("Active Music Contexts: ");
+
+            var contexts = string.Empty;
+
+            foreach (var ctx in activeContexts)
+                contexts += ctx + ", ";
```
hhyyrylainen, line 74: This could just print each one normally. String concatenation without a stringbuilder is not good C# usage.

src/general/Jukebox.cs
```diff
@@ -38,6 +38,8 @@ public class Jukebox : Node
     /// </summary>
     private MusicCategory? previouslyPlayedCategory;
 
+    private MusicContext[] activeContexts = null!;
```
hhyyrylainen, line 41: I think this should be modified to allow null values, and null would be the same as general. That will improve a lot of the code here to not have to allocate dummy lists of data all over the place.

src/general/MusicCategory.cs
```diff
@@ -122,6 +124,12 @@ public void Check()
             track.Check();
     }
 
+    public Track[] GetTracksForContexts(MusicContext[] contexts)
+    {
+        return Tracks.Where(t => t.ExclusiveToContexts == null ||
+            t.ExclusiveToContexts.Any(contexts.Contains)).ToArray();
```
hhyyrylainen, line 130: This gets called quite a lot, which leads to a bunch of memory allocations. This should be refactored to make it so that tracks can be retrieved without memory allocations, so either return an IEnumerable here or/and add a variant that takes in a list as parameter where the tracks are added.

Also related to the context filtering I think it should be the case that we should have more control on the tracks:
- I think exclusive to context tracks, like you implemented, is good, only play them when a matching context is defined
- But I think non-limited tracks should play even if a context is defined, instead I think there should be a track property "DisallowInContexts" that stops a track playing when some context is defined.
- And there should also be a track property for "PlayOnlyWithoutContext" that only allows playing that track if no context is defined (you seemed to make this the default, but I'm not convinced this should be the default).

src/microbe_stage/MicrobeStage.cs
```diff
@@ -279,9 +279,11 @@ public override void StartNewGame()
 
     public override void StartMusic()
     {
+        var biome = CurrentGame!.GameWorld.Map.CurrentPatch!.BiomeTemplate;
+
         Jukebox.Instance.PlayCategory(GameWorld.PlayerSpecies is EarlyMulticellularSpecies ?
             "EarlyMulticellularStage" :
```
hhyyrylainen, line 285: Does this work in early multicellular? Does that music category also have the context specific tracks? What happens if no context matches, then I assume nothing can be played? Does that work correctly when playing multiple tracks at once? I saw a Pause call if no tracks are found but that should be skipped if another simultaneously playing track is found.

src/general/MusicCategory.cs
```diff
@@ -134,6 +155,12 @@ public class Track
 
         public string ResourcePath { get; set; } = null!;
 
+        public MusicContext[]? ExclusiveToContexts { get; set; }
```
hhyyrylainen, line 158: I think these 3 variables really need comments on them to explain this very important part of the jukebox.

hhyyrylainen (approved):
I did a bit of cleanup and added some extra data validation for the JSON. I changed the context check to first apply the exclude contexts to allow excluding on top of required contexts list. I also switched 2-3 places to use the get all tracks without context approach as those were needed to do logically the right thing (and guard against a bug where switched context means that a track is playing but its not detected as belonging to any of the music track lists).

Issue #4537, opened 2023-10-28
Add support for internal and external links for WikiUpdater


ender-titan1:
as currently, the WikiUpdater script does not support links to other Thriveiopedia pages or to external sites.

hhyyrylainen:
It definitely should support links to other thriveopedia pages, but maybe external links should be carefully considered what is allowed.

PR #4544, opened 2023-11-02
Thriveopedia link handling

Oliveriver:

* Adds support for internal and external links in Thriveopedia
* Re-orders BBCode parsing algorithm to handle links to other pages

image

**Related Issues**

Fixes https://github.com/Revolutionary-Games/Thrive/issues/4537

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Oliveriver:
To explain some of the changes here:

* I needed to generate a list of page titles before I could properly fill out the content, since creating links depended on knowing the internal name of other pages. So I swapped the order of some operations in the wiki updater. Perhaps overkill for now as I could have taken the internal names from the game's JSON instead, but this will come in useful once other page types are added.
* Since `CustomRichTextLabel` defines the action of clicking on a link, I needed a global way to open a Thriveopedia page. I created `ThriveopediaManager` which stores the current context's method of opening the Thriveopedia (via main menu or pause menu).

I realise now that the linked issue says we should maybe avoid external links for the moment. So I can remove external link handling if it's preferred.

hhyyrylainen:
> I realise now that the linked issue says we should maybe avoid external links for the moment. So I can remove external link handling if it's preferred.

I think adding a whitelist of link domains would be the better option. Probably all external links in use right now are fine, but there might be some restrictions in the future of what websites we can link directly to in the game.

Oliveriver:
Ready for review again.

I've updated the script to ignore non-whitelisted domains and handle compound icons. I've also reformatted all existing organelle pages and run the script, so on this branch at least, we can finally run the Thriveopedia update script without worry.

ender-titan1:
I just want to mention a concern I have, and that is redirect support. Does this system support redirect pages? The reason I am bringing this up is that (unless I missed a MediaWiki feature) redirects are the best way to link category pages without having the page be included in said category. Please correct me if I’m wrong.

Oliveriver:
> I just want to mention a concern I have, and that is redirect support. Does this system support redirect pages? The reason I am bringing this up is that (unless I missed a MediaWiki feature) redirects are the best way to link category pages without having the page be included in said category. Please correct me if I’m wrong.

If I'm understanding you right, no. If you created a page that redirected to another, then linked to the redirect page in some other page, the parsing script as it stands would throw an error since it believes the page doesn't exist (when really it's just not in the category).

ender-titan1:
Yeah what I mean is that when I try to link a category page in an article, the page is instead added to that category instead of just linking to it. It’s not really necessary but it would be nice to have.

Oliveriver:
Ah, I don’t know about that then. I doubt it would work.

The format script keeps failing due to connectivity problems. I have run the checks locally and they passed.

hhyyrylainen:
I actually ran into the same issue with the Onboarding page. The solution to the problem with adding to wrong category is to add an extra `:` to the front, so to just a link category you can do `[[:Category:Stuff]]` which should still make a valid link but not add the page to the category, @ender-titan1

hhyyrylainen:
The CI failure was some server problem. I rebooted 4 of the build servers and they seem to be working now. Logging into the server hoster website it notified me that one of the build servers is currently impacted by unscheduled downtime. It'll be hopefully back later today.

src/thriveopedia/ThriveopediaManager.cs
```diff
@@ -0,0 +1,32 @@
...
+public class ThriveopediaManager
+{
+    private static readonly ThriveopediaManager ManagerInstance = new();
+
+    public static ThriveopediaManager Instance => ManagerInstance;
+
+    /// <summary>
+    ///   Action to open the Thriveopedia in the current game context, e.g. from the main menu or from the pause menu.
+    /// </summary>
+    public Action<string>? OpenCallback { get; set; }
```
hhyyrylainen, line 16: This not being an event (that can have multiple listeners registered at once) is probably fine, it wouldn't make sense anyway for multiple things to react to this and try to open their own thriveopedias at once.

hhyyrylainen (changes requested):
This seems pretty much entirely fine. I'm not entirely convinced if the ThriveopediaManager is the best design, but I can't immediately come up with any really good criticism of a situation where it wouldn't work. One thing I'd like to see is fixing the WikiUpdated to not have global state, but overall this is excellent update to the wiki functionality.

src/thriveopedia/ThriveopediaManager.cs
Oliveriver: It could be an event, but functionally there should only ever be one action attached, so that would come with the added complexity of removing listeners every time and the risk of bugs associated with that.

Oliveriver:
> This seems pretty much entirely fine. I'm not entirely convinced if the ThriveopediaManager is the best design, but I can't immediately come up with any really good criticism of a situation where it wouldn't work. One thing I'd like to see is fixing the WikiUpdated to not have global state, but overall this is excellent update to the wiki functionality.

I remember going in circles a bit when I first started on this and settled on this approach when I needed to add Thriveopedia callbacks in the BBCode handler. I think having a static class you can call from anywhere makes the Thriveopedia very easy to use. Maybe there are potential bugs or maintenance issues but none are jumping out at me.

hhyyrylainen:
> I remember going in circles a bit when I first started on this and settled on this approach when I needed to add Thriveopedia callbacks in the BBCode handler. I think having a static class you can call from anywhere makes the Thriveopedia very easy to use. Maybe there are potential bugs or maintenance issues but none are jumping out at me.

This was pretty much my conclusion as well. If we convert the pause menu to a singleton (I've been thinking that we might need to switch to this if generating the thriveopedia pages causes long lag spikes), then this could be done differently, but as things currently stand this is pretty much on par with the best design I could come up with on the spot.

hhyyrylainen:
@Oliveriver do you mind if I do those two changes we talked about? Unless you were already working on them.

Oliveriver:
> @Oliveriver do you mind if I do those two changes we talked about? Unless you were already working on them.

I was planning to work on them this evening but if you have the time, go ahead.

hhyyrylainen:
If you can do it this evening, that'd be perfect. I was thinking that if it were to take longer I could take care of this so that this is done at least this week to get this out of the way.

src/thriveopedia/ThriveopediaManager.cs
Oliveriver: Updated.

Scripts/WikiUpdater.cs
```diff
@@ -19,27 +21,18 @@ public static class WikiUpdater
...
+    private static readonly Regex[] WhitelistedDomains =
+    [
+        new Regex(@".*\.wikipedia\.org\/.*"),
+        new Regex(@".*\.revolutionarygamesstudio\.com\/.*"),
+    ];
+
+    /// <summary>
+    ///   Mapping from English page names to internal page names, required for inter-page linking in game.
+    /// </summary>
+    private static Dictionary<string, string> pageNames = new();
```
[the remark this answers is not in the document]
Oliveriver: Some of those tools are static while others aren't, although in fairness, none seem to have global state. I ended up making all methods in the wiki updater non-static as many needed to access the page list and there seemed no reason to keep any others as static.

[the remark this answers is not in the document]
hhyyrylainen: Yeah, that's exactly the reason I've used for making static stuff: anything that doesn't need any state can be static.

Scripts/WikiUpdater.cs
```diff
@@ -1,61 +1,58 @@
...
-        "Nitrogen",
-        "Oxygen",
-
-        // ReSharper disable once StringLiteralTypo
-        "OxyToxy",
-        "Phosphates",
-        "Sunlight",
-        "Temperature",
-    };
+    private readonly Regex[] whitelistedDomains =
```
hhyyrylainen, line 26: This could have stayed static, but as this class is basically instantiated only once ever when running a specific script, this is not going to matter.

hhyyrylainen (approved):
Changes look fine. I didn't playtest. One last concern regarding that would be if the button press sound now correctly plays when pressing the statistics button.

PR #4549, opened 2023-11-08
Implement the RuBisCo Organelle

### LINKED ISSUE #4222: Implement the new organelle for converting ATP to glucose (RuBisCo)
- [ ] We need to add a new protein type organelle (called RuBisCo)
- [ ] The organelle needs a process that converts ATP to glucose but only if the ATP balance is currently good (maybe full on ATP for at least one second?), and glucose storage is not full (or maybe even like lower than 50%?)
- [ ] The organelle needs to scale its speed to not cause ATP to drop which would immediately disengange it
- [ ] Probably a good idea to do some tweaks to the ATP balance and compound balance displays to not make this organelle really confusing there.

https://forum.revolutionarygamesstudio.com/t/new-organelle-for-storing-atp-as-glucose/1001

Requires: https://github.com/Revolutionary-Games/Thrive/issues/4221

GameDungeon:

This PR implements the RuBisCo Organelle and adds the Microbe Process Manager 

**Related Issues**

Closes #4222

**Progress Checklist**

      break existing features:
      members reviewing this PR)


assets/models/organelles/SlimeJet.tscn
hhyyrylainen, line 1: I'm pretty sure this file got broken (the animation is no longer referenced).

assets/models/organelles/RuBisCo.png
hhyyrylainen, line 1: Is this file unused?

assets/textures/gui/RuBisCoIcon.png.import
```diff
@@ -0,0 +1,35 @@
...
+[params]
+
+compress/mode=0
+compress/lossy_quality=0.7
+compress/hdr_mode=0
+compress/bptc_ldr=0
+compress/normal_map=0
+flags/repeat=0
+flags/filter=true
+flags/mipmaps=false
```
hhyyrylainen, line 24: Icons must have mipmaps on.

assets/models/organelles/SlimeJet.tscn
GameDungeon: I'll revert this file, but isn't that the slimejet animation in the rusticyanin file?

hhyyrylainen: What do you mean? This file I commented on is `assets/models/organelles/SlimeJet.tscn`. The file above this which was also changed is `assets/models/organelles/Rusticyanin.tscn` but I cannot see `SlimeJet.anim` in it.

assets/models/organelles/RuBisCo.png
GameDungeon: Yes, it was auto imported and I forgot to delete it

assets/models/organelles/SlimeJet.tscn
GameDungeon: This is my mistake, I was misreading the file. I've reverted both. Godot making random changes is quite annoying.

hhyyrylainen: I think the rustcyanin changes would have been fine, as I think those were purely Godot updating things to reflect new default values (as scene parameters are simply omitted if they match the default value).

GameDungeon:
There are a couple of small things left, but otherwise this is ready for Codereview, Testing, and Balancing.

simulation_parameters/microbe_stage/organelles.json
```diff
@@ -269,6 +269,37 @@
...
+    "prokaryoteChance": 0.5,
+    "chanceToCreate": 0.5,
+    "Density": 900,
+    "displayScene": "res://assets/models/organelles/RuBisCo.tscn",
+    "name": "RUBISCO",
+    "productionColour": "#26e0ff",
+    "consumptionColour": "#ff5649",
+    "iconPath": "res://assets/textures/gui/bevel/parts/RuBisCoIcon.png",
+    "editorButtonGroup": "Protein",
+    "editorButtonOrder": 1
```
hhyyrylainen, line 301: This should use a unique value for the button order to ensure consistent sorting.

src/gui_common/thrive_theme.tres
hhyyrylainen, line 1: Is it intentional that a part of the global thrive style was removed?

src/microbe_stage/CompoundBag.cs
```diff
@@ -244,4 +244,13 @@ IEnumerator IEnumerable.GetEnumerator()
     {
         return GetEnumerator();
     }
+
+    public CompoundBag Clone()
```
hhyyrylainen, line 248: Shouldn't this clone specific capacities as well?

src/microbe_stage/TweakedProcess.cs
```diff
@@ -5,12 +5,14 @@
 /// </summary>
 public class TweakedProcess : ICloneable
 {
+    public float Count;
```
hhyyrylainen, line 8: If the count is kept here this *must be* documented as it's very confusing to have both a count and the actual rate. There needs to be comments explaining what's the difference here.

src/microbe_stage/MicrobeWorldSimulation.cs
```diff
@@ -203,6 +206,8 @@ public void Init(Node visualDisplayRoot, CompoundCloudSystem cloudSystem)
         // Systems stored in properties
         CameraFollowSystem = new CameraFollowSystem(EntitySystem);
 
+        MicrobeProcessManagerSystem = new MicrobeProcessManagerSystem(EntitySystem, couldParallelize);
```
hhyyrylainen, line 209: I'd say this should parallelize itself as it looks to be doing as much computation as the process system.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -352,7 +352,7 @@ void MakeSureResultExists(Compound compound)
 
             result.Efficiency = efficiency;
 
-            speedFactor *= process.Rate;
+            speedFactor *= process.Rate * process.Count;
```
hhyyrylainen, line 355: I don't understand why this needs to be made this confusing. Why not just have an "EffectiveRate" property in the tweaked process class? That then can take into account the Rate and a maybe a variable called AdjustedRate or ControlledRate which is set by the process control (or even the player through the GUI in the future).

src/microbe_stage/systems/MicrobeProcessManagerSystem.cs
```diff
@@ -0,0 +1,180 @@
...
+            if (processor.ActiveProcesses == null)
+                return;
+
+            if (biome == null)
+                throw new NullReferenceException("Biome needs to be set");
+
+            var glucoseConsuming = false;
+
+            // Create dummy compound bag
+            var bag = storage.Compounds.Clone();
```
hhyyrylainen, line 106: This is a no-go, allocating duplicate compounds bag for each entity on each game update. This is going to cause so much GC garbage.

src/microbe_stage/systems/MicrobeProcessManagerSystem.cs
```diff
@@ -0,0 +1,180 @@
...
+                .OrderBy(p => GetPriority(p.Process)).ToList();
+
+            processor.ActiveProcesses = processPriority;
+
+            foreach (var process in processor.ActiveProcesses)
+            {
+                process.Rate = 1;
+            }
+
+            foreach (var process in processor.ActiveProcesses)
```
hhyyrylainen, line 120: I really think there should be a way for the process system to store the information as it is calculating it for this system to use. Now this basically doubles the amount of calculation needed to run the process system and makes the overall thing *much* harder to change in the future.

GameDungeon: Could you elaborate exactly what you mean by this? I now see a way to remove the storageConstraintModifier from this process which would be the main culprit here. I think ProcessSystem.CalculateEnvironmentModifier would heavily benefit from some sort of caching, but I'm not sure the best way to go about it.

src/microbe_stage/systems/MicrobeProcessManagerSystem.cs
GameDungeon: The exact method of how I implemented the RuBisCo logic here changed several times. I know how to remove this extra compound bag, but it's at the cost of future ease of editing. I overall agree that it would be a problem as is.

src/microbe_stage/systems/MicrobeProcessManagerSystem.cs
hhyyrylainen: I'd start by modifying the `BioProcesses` component. To start off I'd put in fields like "bool UsedGlucose" and `float AtpBalance`. Those fields would be written by the process system as it does its thing as it basically already needs to know that information it would be just a tiny bit of extra overhead to store that information. 
Based on those I'd write a control algorithm where the speed for rubisco was gradually increased as long as UsedGlucose was false and AtpBalance was over 0.5f or something. Then conversely if I just didn't icrease the rubisco speed, I'd maybe half it? Or at least reduce it a lot more than I increment it. (halfing would be very similar to how TCP flow control adjusts the speed).

src/microbe_stage/systems/MicrobeProcessManagerSystem.cs
hhyyrylainen: It needs to be done, unless you are fine with the process speed control running only like once per second for each microbe... With the ECS refactor I've been way more careful than before to ensure that no operation that runs each frame allocates memory (and the few places where that couldn't be avoided use an array pool).

src/gui_common/thrive_theme.tres
GameDungeon: Nope. Godot auto changed something and I missed it. Reverted

src/microbe_stage/CompoundBag.cs
GameDungeon: Due to a change in how stuff is done, I just undid this change.

src/microbe_stage/systems/ProcessSystem.cs
GameDungeon: Previously the rate was basically a count anyway. I basically did precisely what you just suggested with count as "rate" and rate (now set by the process control) as "EffectiveRate"

GameDungeon:
Most of the code should stay the same now unless a review asks for a change, but this can't be merged until I track down whatever is messing up the process speeds.

GameDungeon:
The large bugs are fixed now, but I'm sure there are still things that need to be resolved. But it's in a state that I'm comfortable with it going through testing and review again.

buckly90:
Seems to be in a good state as-is, though some later features aught to provide a more robust use for this part.

I recommend roughly doubling the process rate for more effectiveness. Should be ready to go otherwise on the gameplay side.

GameDungeon:
When I get time pretty soon, I will update this and try and get it merged.

GameDungeon:
After the major changes that have happened since this PR was first created it would be very impractical to try and get it up to date. I will close this, but I will leave the branch for now so I have code to look back on for implementing RuBisCo

PR #4575, opened 2023-11-23
simplify and improve rotation speed calculation

adQuid:

Rebalances rotation rates post ECS refactor, including the effect of cillia. This simplifies the calculation somewhat, but still accounts for leverage from cillia placed further from the center.

**Related Issues**

https://github.com/orgs/Revolutionary-Games/projects/5/views/1?pane=issue&itemId=44943890

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/MicrobeInternalCalculations.cs
+++ b/src/microbe_stage/MicrobeInternalCalculations.cs
@@ -265,7 +265,8 @@ public static class MicrobeInternalCalculations
             if (organelle.Definition.HasCiliaComponent)
             {
                 cilliaFactor += Constants.CILIA_ROTATION_FACTOR +
-                    Hex.AxialToCartesian(organelle.Position).LengthSquared() * Constants.CILIA_RADIUS_FACTOR_MULTIPLIER;
+                    Hex.AxialToCartesian(organelle.Position).LengthSquared() *
+                    Constants.CILIA_RADIUS_FACTOR_MULTIPLIER;
             }
         }
 
--- a/src/microbe_stage/MicrobeInternalCalculations.cs
+++ b/src/microbe_stage/MicrobeInternalCalculations.cs
@@ -274,8 +274,7 @@ public static class MicrobeInternalCalculations
 
             if (organelle.Definition.HasCiliaComponent)
             {
-                cilliaFactor += Constants.CILIA_ROTATION_FACTOR +
-                     distance * Constants.CILIA_RADIUS_FACTOR_MULTIPLIER;
+                cilliaFactor += Constants.CILIA_ROTATION_FACTOR + distance * Constants.CILIA_RADIUS_FACTOR_MULTIPLIER;
             }
         }
 
--- a/src/microbe_stage/MicrobeInternalCalculations.cs
+++ b/src/microbe_stage/MicrobeInternalCalculations.cs
@@ -282,9 +282,12 @@ public static class MicrobeInternalCalculations
     {
         float cilliaFactor = 0;
         float maxRadius = 0;
+        float totalMass = 0;
 
         foreach (var organelle in organelles)
         {
+            totalMass += organelle.Definition.HexCount;
+
             var distance = Hex.AxialToCartesian(organelle.Position).LengthSquared();
 
             if (distance > maxRadius)
@@ -298,8 +301,7 @@ public static class MicrobeInternalCalculations
             }
         }
 
-        var effectiveMass = organelles.Sum(x => x.Definition.HexCount)
-            + (maxRadius * Constants.CELL_ROTATION_RADIUS_FACTOR);
+        var effectiveMass = totalMass + (maxRadius * Constants.CELL_ROTATION_RADIUS_FACTOR);
 
         return effectiveMass / (Constants.CELL_ROTATION_INFLECTION_MASS + cilliaFactor + effectiveMass)
             * Constants.CELL_MAX_ROTATION + Constants.CELL_MIN_ROTATION;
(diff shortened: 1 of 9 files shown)


hhyyrylainen:
What about cell colonies? It uses basically the same math approach to the rotation speed calculation (`MicrobeColonyHelpers.CalculateRotationMultiplier`).

adQuid:
> What about cell colonies? It uses basically the same math approach to the rotation speed calculation (`MicrobeColonyHelpers.CalculateRotationMultiplier`).

crud, forgot about colonies...

adQuid:
Let me know if behavior is different for you, but for me Thrive still crashes when entering the editor tab on the multicellular freebuild editor, so it might be too early to start testing the effects of colonies on turning.

hhyyrylainen:
That's still not reimplemented, but colonies can still be accessed, and their rotation speed (and movement speed) are really not tuned in well. I'll merge these changes today to get them in for today's build to have a more playable microbe stage.

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -253,56 +253,22 @@ public static float CalculateCapacity(IEnumerable<OrganelleTemplate> organelles)
...
+            }
         }
 
-        // Add the extra speed from cilia
-        if (ciliaCount > 0)
-        {
-            speed -= ciliaCount * Mathf.Sqrt(radiusSquared) * Constants.CILIA_RADIUS_FACTOR_MULTIPLIER *
-                Constants.CILIA_ROTATION_FACTOR;
-        }
+        var effectiveMass = shape.GetMass();
```
hhyyrylainen, line 268: So this now ignores the shape of the microbe? Well that's probably fine as microbes are always centered, but taking actual angular velocity inertia into account would be a much more accurate calculation. I think right now this will calculate the same rotation speed for a thin and long stick as well as a small round blob, which I think should get vastly different speeds due to how far away the mass is from the center in the long thin stick.

hhyyrylainen:
I'm still not super keen on the approach as I can't really see how it could be adapted for the microbe colony rotation code, which would be pretty strange if rotation switched to a completely different approach when you bind with another cell.

adQuid:
Given your announcement to re-implement early multicellular next week, I think it makes sense to leave this PR on ice until then then try to do both stages in one go.

adQuid:
I've made a new algorithm for calculating colony rotation that gets rid of the strange privileged position of the root cell (particularly since it's not always the center anyway). There are still frequent crashes due to some kind of issue with making the physics shape, but before I took time to try to dive into that I wanted to know if you even liked the approach of this PR.

hhyyrylainen:
I guess it is a doable approach. One note though is that you cannot get the mass of the individual cells in a colony as the physics shapes for colony members are skipped to save on performance (the overall colony shape is available, which is why I was hoping that inertia based calculation could be made to work as that would use the real physical rotational inertia of the shape). Oh and also as-written the colony rotation does an unnecessary variable assignment *and* disables the rotation speed increase cheat from working.

But overall I guess having the rotation speed not be physically accurate for now is the lesser evil of releasing a new Thrive version with the rotation speeds all over the place.

src/microbe_stage/components/MicrobeColony.cs
```diff
@@ -891,26 +885,17 @@ public static void CalculateRotationMultiplier(this ref MicrobeColony colony, Ph
 
                 var distanceSquared = memberPosition.RelativePosition.LengthSquared();
 
-                if (distanceSquared < MathUtils.EPSILON)
-                    continue;
+                // Multiply both the propulsion and mass by the distance from center to simulate leverage
+                // This relies on the bounding of the cell rotation, as a colony can never be faster than the
+                // fastest cell inside it
+                var memberRotation = MicrobeInternalCalculations
+                        .CalculateRotationSpeed(colonyMember.Get<OrganelleContainer>().Organelles!)
```
hhyyrylainen, line 892: I was about to comment the following but then I read how you basically gutted the rotation speed calculation to no longer have any relation to the actual physics shape.
> This calculate rotation speed is going to absolutely kill, as this will compute the full physics shape for each colony member.

adQuid: That is correct. Any attempts to go half-and-half proved very cumbersome, so if we want to go with this PR's approach at all I think it's best to just stay away from the physics shapes as a whole. It's less accurate from a Newtonian standpoint, but it is easier to control for gameplay balance.

src/microbe_stage/components/MicrobeColony.cs
```diff
@@ -873,16 +873,10 @@ public static float CalculateColonyAdditionalEntityWeight(float singleCellWeight
...
-
-            // TODO: a better function (should also update MicrobeInternalCalculations.CalculateRotationSpeed)
-            var rotationHindering = 1 + Mathf.Clamp(Mathf.Pow(speedFraction, 1 / 4.0f), 0.0001f, 2.0f);
-
-            float colonyRotationHelp = 0;
+            float colonyRotation = 0;
 
             foreach (var colonyMember in colony.ColonyMembers)
             {
-                // Leader uses its own rotation value as the base on top which this rotation multiplier is applied so
```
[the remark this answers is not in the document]
hhyyrylainen: I saw you also removed the distance check which would naturally skip the root cell, which I suspect could cause some divide by zero or other such issue.

src/microbe_stage/components/MicrobeColony.cs
hhyyrylainen: I was kind of hoping that in the future we could go for full physics based rotating, which I guess now that I also think about it doesn't require the physics shape when calculating the allowed speed.

Still to clean up things the variant of the speed calculation taking in the physics body should be entirely removed and the code comments adjusted on where the speed values are calculated to say that they are only calculated there for historical reasons and don't currently require the physics state.

hhyyrylainen (approved):
I spotted one more issue: a temporary array was created just to get its length. I fixed that so now this should be fine for merge.

PR #4583, opened 2023-11-25
Engulf progress indicator can now go over 100% 

### LINKED ISSUE #4565: Engulf progress indicator can now go over 100%
seems to end at 200%, likely caused by hh fixing the cells being killed at 50% digested, and no longer says “devoured” on the health.

CI09:

I am not sure if that's the way I am supposed to do it, but I divided engulf progress indicator value by 2 so it stops at 100% instead of 200%.

**Related Issues**

Closes #4565

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/MicrobeHUD.cs
+++ b/src/microbe_stage/MicrobeHUD.cs
@@ -283,7 +283,7 @@ public class MicrobeHUD : CreatureStageHUDBase<MicrobeStage>
             var percentageValue = TranslationServer.Translate("PERCENTAGE_VALUE");
 
             // Show the digestion progress to the player
-            hp = 1 - (stage.Player.Get<Engulfable>().DigestedAmount / Constants.PARTIALLY_DIGESTED_THRESHOLD / 2);
+            hp = 1 - stage.Player.Get<Engulfable>().DigestedAmount;
             maxHP = Constants.FULLY_DIGESTED_LIMIT;
             hpText = percentageValue.FormatSafe(Mathf.Round((1 - hp) * 100));
             playerWasDigested = true;
--- a/src/microbe_stage/MicrobeHUD.cs
+++ b/src/microbe_stage/MicrobeHUD.cs
@@ -283,7 +283,7 @@ public class MicrobeHUD : CreatureStageHUDBase<MicrobeStage>
             var percentageValue = TranslationServer.Translate("PERCENTAGE_VALUE");
 
             // Show the digestion progress to the player
-            hp = 1 - (stage.Player.Get<Engulfable>().DigestedAmount / Constants.PARTIALLY_DIGESTED_THRESHOLD);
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
I think the divide by PARTIALLY_DIGESTED_THRESHOLD should just be removed because the player is digested when the digestion value reaches 1 and not half of that like the engulf system previously always stopped engulfing stuff at 50% progress.

CI09:
Works for me.

hhyyrylainen:
Does the health bar now say "DEVOURED" when the player dies to engulfing? If not that would be also excellent to fix at the same time. I have a hunch that maybe updating the if:
```
        // Update to the player's current digested progress, unless the player does not exist or is dead already
        if (stage.HasPlayer && !stage.Player.Get<Health>().Dead)
```
to that would solve that issue.

CI09:
> Does the health bar now say "DEVOURED" when the player dies to engulfing? If not that would be also excellent to fix at the same time. I have a hunch that maybe updating the if:
> 
> ```
>         // Update to the player's current digested progress, unless the player does not exist or is dead already
>         if (stage.HasPlayer && !stage.Player.Get<Health>().Dead)
> ```
> 
> to that would solve that issue.

No, after it reaches 100% it goes back to normal health indicator for a while and then to 0.

hhyyrylainen:
> No, after it reaches 100% it goes back to normal health indicator for a while and then to 0.

That's still a new bug with 0.6.4, then.

CI09:
> I have a hunch that maybe updating the if:
> 
> ```
>         // Update to the player's current digested progress, unless the player does not exist or is dead already
>         if (stage.HasPlayer && !stage.Player.Get<Health>().Dead)
> ```
> 
> to that would solve that issue.

It doesn't fix the DEVOURED text not displaying, but I can push new commit with code as it is above.

hhyyrylainen:
If it doesn't fix the bug I don't see much point in complicating the logic here as that might make it more difficult to find the real root cause of the "devoured" text not displaying.

CI09:
I managed to bring "Devoured" back (health indicator still goes back to normal health after player gets engulfed and it switches to "Devoured" a moment later) by removing `playerWasDigested = false;` in  normal health display update for non-engulfed player (it is in the same function I made commits in), but I don't know if this does break any features so I am not going to commit it yet.

hhyyrylainen:
I would expect that there's a reason for that. The player can escape engulfment if the engulfing cell dies before the player is digested, so there is likely a need to reset that variable. I suspect that swapping the check for ingested status to be `PhagocytosisStep is not (None or Ingestion)` (adjust syntax to get it to compile).

@Athriq probably knows how this feature should be fixed.

hhyyrylainen (approved):
This is a good improvement to the digestion indicator so I'm merging this.

hhyyrylainen:
Testing that change, I was able to verify that this fixes the devoured text:
kuva

PR #4587, opened 2023-11-26
Fix PilusDamageSystem not triggering

ender-titan1:

This PR makes one very small change that stopped `PilusDamageSystem` from triggering.
I'm guessing this was because `Species` is not actually a component so when used in the `With` attribute, it blocked the system from running properly.

**Related Issues**

- https://github.com/orgs/Revolutionary-Games/projects/5/views/1?pane=issue&itemId=44922458
- This appears to also work as intended now: https://github.com/orgs/Revolutionary-Games/projects/5/views/1?pane=issue&itemId=44922694

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/PilusDamageSystem.cs
+++ b/src/microbe_stage/systems/PilusDamageSystem.cs
@@ -11,7 +11,6 @@
     [With(typeof(OrganelleContainer))]
     [With(typeof(CollisionManagement))]
     [With(typeof(MicrobePhysicsExtraData))]
-    [With(typeof(Species))]
     public sealed class PilusDamageSystem : AEntitySetSystem<float>
     {
         public PilusDamageSystem(World world, IParallelRunner parallelRunner) : base(world, parallelRunner)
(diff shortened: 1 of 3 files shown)


src/microbe_stage/systems/PilusDamageSystem.cs
```diff
@@ -11,7 +11,6 @@
     [With(typeof(OrganelleContainer))]
     [With(typeof(CollisionManagement))]
     [With(typeof(MicrobePhysicsExtraData))]
-    [With(typeof(Species))]
```
hhyyrylainen, line 14: Ah, I've made a serious typo :facepalm: 
This line should be `    [With(typeof(SpeciesMember))]` as this system reads the species member component of the entity.

hhyyrylainen:
I put the damage not dealing through a pilus t othe needs verification column as it is about two things:
- Engulfing shouldn't be able to start when a pilus is touching
- Toxin projectiles shouldn't be able to apply damage when hitting a pilus.

hhyyrylainen (approved):
Changed the system run attribute to what I originally intended. I also rescaled the pilus damage to get it to about what it was before while still allowing harder physics collisions to deal more damage.

Issue #4227, opened 2023-03-16
Make late multicellular editor use the panoramic backgrounds


hhyyrylainen:
instead of displaying without a background

PR #4606, opened 2023-12-01
Make late multicellular stage use patch specific backgrounds

### LINKED ISSUE #4228: Make late multicellular stage use patch specific backgrounds
Right now only one background is used regardless of patch: https://github.com/Revolutionary-Games/Thrive/pull/4225

In addition to changing the image, the ambient colour needs to change so the background image path and colour both probably should be added to the JSON data for biomes.

CI09:

This PR adds function to multicellular stage script which updates panorama background, and calls it where it is needed (doesn't work in editor, might need some work once #4227 is completed). Also adds panorama texture reference in biomes.json and references in simulation parameters.

Some biomes currently don't have their panorama backgrounds so I used other as replacement I think it is coastal (I used epipelagic) and sea floor (abyssopelagic).

(This is probably my most complex PR yet.)

**Related Issues**

Closes #4228

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/Biome.cs
```diff
@@ -45,7 +50,7 @@ public class Biome : IRegistryType
 
     public void Check(string name)
     {
-        if (string.IsNullOrEmpty(Name) || string.IsNullOrEmpty(Background))
+        if (string.IsNullOrEmpty(Name) || string.IsNullOrEmpty(Background) || string.IsNullOrEmpty(Panorama))
         {
             throw new InvalidRegistryDataException(name, GetType().Name,
                 "Empty normal or damaged texture");
```
hhyyrylainen, line 56: Could you fix the error message here? Looks like this is copy-pasted from membrane but someone (probably me, honestly) forgot to update the message.

src/late_multicellular_stage/MulticellularStage.cs
```diff
@@ -635,6 +637,23 @@ public void OnSocietyFounded(PlacedStructure societyCenter)
         animationCamera.ReParent(rootOfDynamicallySpawned);
     }
 
+    public void UpdateBackgroundPanorama()
+    {
+        if (CurrentGame != null)
+        {
+            if (CurrentGame.GameWorld.Map.CurrentPatch != null)
```
hhyyrylainen, line 644: You can combine this and the previous if with null propagation (this is what the CI is complaining about).

hhyyrylainen (approved):
I fixed the problems myself that I found. I was especially able to cut down vastly on the number of lines used in the panorama applying method, and I also put that panorama apply in the intended place (where there was even a TODO about it).

hhyyrylainen:
I opened a follow up issue as this doesn't change the sky ambient colour: https://github.com/Revolutionary-Games/Thrive/issues/4608

PR #4637, opened 2023-12-10
Remove Protoplasm from the Autoevo Mutation Pool

GameDungeon:

This removes protoplasm from the autoevo mutation pool. Protoplasm in an old feature to help microbes in autoevo, but now it is no longer necessary and and often causes more problems than it solves. This PR just removes it from the pool and not the game, but I think it should be removed entirely next time we decide to break save compatibility. 

Basic testing in the autoevo explorer tells me that this doesn't affect anything too badly. (Slightly fewer eukaryotes and they no longer spam cytoplasm as much, but that's the extent)
**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Did you double the cytoplasm chance to compensate for the AI now having just one type of "cytoplasm" to pick from?

GameDungeon:
Yes. Not having protoplasm did cut down on total "plasma" in the cell quite a bit (which is in some cases a good thing compared to the past) I tried getting it at least closer to normal by increasing cytoplasm.

buckly90:
Seems to work just fine. Doesn't break saves or anything, and autoevo seems to like ye olde cytoplasm balloons a little less now.

hhyyrylainen (approved):
I guess now that 0.6.5 is confirmed to be the next release (and not a patch in between) this can be merged.

hhyyrylainen (changes requested):
Actually hang on, could you redo the PascalCase naming convention? This PR now reverts a previous PR partially.

hhyyrylainen (approved):
I fixed the case change.

PR #4457, opened 2023-09-04
Implement part unlocks

0HyperCube:
**Brief Description of What This PR Does**
- [x] Unlock system
- [x] Some basic unlock conditions
- [x] Show progress towards unlocking
- [x] Unlock all in freebuild mode
- [x] Show undiscovered icon
- [ ] Use translation system

**Related Issues**
Closes #3935

**Progress Checklist**
Note: before starting this checklist the PR should be marked as non-draft.

- [ ] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [ ] Initial code review passed (this and further items should not be checked by the PR author)
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author. Merging must follow our
[styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md#git).

PR #4730, opened 2023-12-20
Finish implementation of organelle unlocks

### LINKED ISSUE #910: Add unlock conditions for organelles
It's pretty overwhelming to have a ton of organelles unlocked from the start, we should implement conditions for unlocking organelles, like needing to be in a light biome to get thykaloids, getting engulfed or engulfing another cell to get a pilus, that sort of thing.

### LINKED ISSUE #3935: Implement system for tracking progress of part unlocks
and add a few conditions to cut down on the initial number of organelles available to the player. As it was decided only the player should need to unlock the parts (AI species can evolve just like before) we can just put an unlock progress object into the GameWorld.

When unlocking an organelle there should probably be a popup in the cell editor that displays for a few seconds that a new part is unlocked.

https://forum.revolutionarygamesstudio.com/t/organelle-unlocks/696

ender-titan1:

This PR finishes the implementation of organelle unlocks by 0HyperCube.

First, as this PR literally continues from where HyperCube left off, the previous PR will need to be fixed and updated: 

- [x] Update translations
- [x] Fix font size
- [x] Make undiscovered organelles look nicer
- [x] Implement the new tracking system that held up the previous PR
- [x] Reimplement all unlock conditions using the new system
- [x] Implement WorldBasedUnlockCondition
- [x] Loading from registry
- [x] Saving & Loading progress
- [x] Coloured requirements 
- [x] Fix typos in translations

Secondly, additional features have to be implemented for this to be viable:

- [x] Cheat to unlock all organelles
- [x] Setting to turn off organelle unlocks
- [x] Handle multicellular
- [x] Update tooltips after organelles/membrane changed
- [x] ~~Create a  JsonConverter for ConditionSet~~

**Related Issues**

Closes #3935
Closes #910 (This is an old issue that is basically the same thing)

Continues from #4457 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


ender-titan1:
I have now reached the point with this PR equivalent to that of the previous PR. It is not ready for review yet, as there are plenty of bugs and adjustments I'd like to make, but it's getting close

ender-titan1:
Finally! After fighting with CI and style checks, I can say this PR is ready for review!

Thrive.csproj
```diff
@@ -634,7 +642,11 @@
     <Compile Include="src\microbe_stage\PatchNameOverlay.cs" />
     <Compile Include="src\microbe_stage\PhagocytosisPhase.cs" />
     <Compile Include="src\microbe_stage\PlacedOrganelle.cs" />
+    <Compile Include="src\statistics_tracking\ReproductionStatistic.cs" />
```
hhyyrylainen, line 645: I think this doesn't really deserve a top level folder. I'd put this folder in the general folder so there would be `src/general/statistics_tracking` as the path for these files.

src/engine/CheatManager.cs
```diff
@@ -32,6 +32,11 @@ static CheatManager()
     /// </summary>
     public static event EventHandler<EventArgs>? OnRevealAllPatches;
 
+    /// <summary>
+    ///   Fired whenever the pleyer uses the "Unlock All Organelles" cheat
```
hhyyrylainen, line 36: Typo

src/general/IDifficulty.cs
```diff
@@ -53,6 +53,11 @@ public interface IDifficulty : IRegistryAssignable
     ///   How intense should the fog-of-war be
     /// </summary>
     public FogOfWarMode FogOfWarMode { get; }
+
+    /// <summary>
+    ///   Whether organelle unlocks are enabled or not
```
hhyyrylainen, line 58: This should clarify that if unlocks are disabled if that means then that everything is unlocked from the start.

src/general/NewGameSettings.cs
```diff
@@ -384,10 +389,13 @@ public void OpenFromDescendScreen(GameProperties currentGame)
         playerDeathPopulationPenalty.Value = difficulty.PlayerDeathPopulationPenalty;
         glucoseDecayRate.Value = difficulty.GlucoseDecay * 100;
         osmoregulationMultiplier.Value = difficulty.OsmoregulationMultiplier;
+        fogOfWarModeDropdown.Selected = (int)difficulty.FogOfWarMode;
```
hhyyrylainen, line 392: I assume this fixes a mistake that was in a previous PR.

src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -558,6 +558,8 @@ public void EditorButtonPressed()
 
         EnsureGameIsUnpausedForEditor();
 
+        stage.RecordPlayerReproduction();
```
hhyyrylainen, line 561: This shouldn't happen here as the stage can still prevent the player from switching to the editor if preconditions are no longer met.

src/gui_common/CollapsibleList.cs
```diff
@@ -120,6 +120,18 @@ public void RemoveItem(string name)
...
     }
 
+    public void RemoveAllOfType<T>()
+        where T : Control
+    {
+        var found = items.FindAll(item => item is T);
+
+        foreach (var item in found)
+        {
+            item.DetachAndQueueFree();
```
hhyyrylainen, line 130: Does this really need this? Or would QueueFree or Free be enough?

src/microbe_stage/MicrobeStage.cs
```diff
@@ -368,6 +368,15 @@ public void PauseKeyPressed()
         }
     }
 
+    public override void RecordPlayerReproduction()
+    {
+        if (!HasPlayer || Player.Get<Health>().Dead || PlayerIsEngulfed(Player))
```
hhyyrylainen, line 373: If the stage was in control when the player reproduction was recorded, this duplicate logic would not be needed.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -515,6 +524,9 @@ public void MoveToMulticellular()
             throw new Exception("failed to keep the current scene root");
         }
 
+        // Temporary
```
hhyyrylainen, line 527: "Temporary" this needs to go before merging or this needs a massive TODO slapped on this and an issue opened to make multicellular not "break" the unlock system.

src/saving/SaveList.tscn
hhyyrylainen, line 1: Could you revert the changes in this file, they don't seem related to the other changes?

src/modding/NewModGUI.tscn
hhyyrylainen, line 1: Could you revert the changes in this file, they don't seem related to the other changes?

src/statistics_tracking/IStatistic.cs
```diff
@@ -0,0 +1,7 @@
+﻿/// <summary>
+///   A statitsic about the player or world tracked by <see cref="WorldStatsTracker"/>.
```
hhyyrylainen, line 2: Typo

src/microbe_stage/systems/EngulfedDigestionSystem.cs
```diff
@@ -240,6 +247,9 @@ protected override void Update(float delta, in Entity entity)
                 if (totalAmountLeft <= 0 || engulfable.DigestedAmount >= Constants.FULLY_DIGESTED_LIMIT)
                 {
                     engulfable.PhagocytosisStep = PhagocytosisPhase.Digested;
+
+                    if (entity.Has<PlayerMarker>() && engulfedObject.Has<MicrobeControl>())
+                        gameWorld?.StatisticsTracker.TotalEngulfedByPlayer.Increment(1);
```
hhyyrylainen, line 252: There should be a preupdate overridden method that ensures game world is not null and this uses then null suppression to be consistent with other systems.

src/microbe_stage/systems/EngulfedDigestionSystem.cs
```diff
@@ -240,6 +247,9 @@ protected override void Update(float delta, in Entity entity)
                 if (totalAmountLeft <= 0 || engulfable.DigestedAmount >= Constants.FULLY_DIGESTED_LIMIT)
                 {
                     engulfable.PhagocytosisStep = PhagocytosisPhase.Digested;
+
+                    if (entity.Has<PlayerMarker>() && engulfedObject.Has<MicrobeControl>())
```
hhyyrylainen, line 251: Is it necessary to check that the object has microbe control here? Only entities that have a bunch of the microbe components can be processed by this system, so it is 100% known that entity is a valid microbe when code execution reaches here. Also I'd put the player check outside the loop all the way up on line 79 to just check the component existing once at the start of this method instead for each engulfed object.

src/microbe_stage/systems/MicrobeDeathSystem.cs
```diff
@@ -340,6 +340,12 @@ private bool HandleMicrobeDeath(ref CellProperties cellProperties, in Entity ent
                     Constants.CREATURE_DEATH_POPULATION_LOSS, TranslationServer.Translate("DEATH"));
             }
 
+            // Record player death in statistics
+            if (entity.Has<PlayerMarker>())
+            {
+                gameWorld?.StatisticsTracker.TotalPlayerDeaths.Increment(1);
```
hhyyrylainen, line 346: This should use null suppression like on line 339

src/statistics_tracking/GenericStatistic.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    public T Value { get; protected set; }
+
+    public abstract void Increment(T value);
+}
+
+public class SimpleStatistic : GenericStatistic<int>
+{
+    public SimpleStatistic(StatsTrackerEvent @event)
+        : base(0, @event)
```
hhyyrylainen, line 21: I prefer having `:` on the previous line, I think that looks better to me and its the style I've been enforcing everywhere in THrive so this is now inconsistent style to fit with all of the rest of the code.

src/statistics_tracking/ReproductionStatistic.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Components;
+using DefaultEcs;
+using Newtonsoft.Json;
+
+public class ReproductionStatistic : IStatistic
+{
+    public StatsTrackerEvent LinkedEvent { get; set; } = StatsTrackerEvent.PlayerReproduced;
```
hhyyrylainen, line 9: Shouldn't this be json ignored?

src/statistics_tracking/ReproductionStatistic.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    [JsonProperty]
+    public int TimesReproduced { get; set; }
+
+    [JsonProperty]
+    public Dictionary<string, int> ReproducedInBiomes { get; set; } = new();
```
hhyyrylainen, line 15: We don't use the approach of storing names of things, the real instances are used instead. I suspect that there's some saving trouble? If so could you prep this to use the instances as keys and I can dive into the save format problems?

src/statistics_tracking/WorldStatsTracker.cs
```diff
@@ -0,0 +1,27 @@
...
+    [JsonProperty]
+    public SimpleStatistic TotalEngulfedByPlayer { get; private set; } = new(StatsTrackerEvent.PlayerEngulfedOther);
+
+    [JsonProperty]
+    public SimpleStatistic TotalPlayerDeaths { get; private set; } = new(StatsTrackerEvent.PlayerDied);
+
+    [JsonProperty]
+    public ReproductionStatistic PlayerReproductionStatistic { get; private set; } = new();
+
+    public IStatistic[] CollectStatistics()
```
hhyyrylainen, line 18: To avoid extra list allocations, could this have an internal Lazy variable that holds this list so that it is only constructed once per tracker instance?

src/microbe_stage/organelle_unlocks/WorldBasedUnlockCondition.cs
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+
+    /// <summary>
+    ///   The player's current patch has a certain amount of certain compound.
+    /// </summary>
+    public class PatchCompound : WorldBasedUnlockCondition
+    {
+        [JsonProperty("Compound")]
+        public string RawCompound = string.Empty;
```
hhyyrylainen, line 129: It should also be possible here to just store the Compound, the saving system should be capable of it.

src/microbe_stage/organelle_unlocks/WorldBasedUnlockCondition.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            if (Min.HasValue && Max.HasValue)
+                builder.Append(new LocalizedString("COMPOUND_IS_BETWEEN", compoundName, Min, Max));
+            else if (Min.HasValue)
+                builder.Append(new LocalizedString("COMPOUND_IS_ABOVE", compoundName, Min));
+            else if (Max.HasValue)
+                builder.Append(new LocalizedString("COMPOUND_IS_BELOW", compoundName, Max));
+        }
+
+        public override void Check(string name)
```
hhyyrylainen, line 178: Shouldn't this check that either min or max is defined?

src/microbe_stage/organelle_unlocks/WorldBasedUnlockCondition.cs
```diff
@@ -0,0 +1,187 @@
...
+        public override void GenerateTooltip(LocalizedStringBuilder builder)
+        {
+            if (compound == null)
+                return;
+
+            var compoundName = compound.InternalName;
+
+            if (Min.HasValue && Max.HasValue)
+                builder.Append(new LocalizedString("COMPOUND_IS_BETWEEN", compoundName, Min, Max));
+            else if (Min.HasValue)
```
hhyyrylainen, line 172: Missing braces in multiline if chain.

src/microbe_stage/organelle_unlocks/UnlockProgress.cs
```diff
@@ -0,0 +1,79 @@
...
+using Newtonsoft.Json;
+
+/// <summary>
+///   Stores which organelles have been unlocked by the player.
+/// </summary>
+[UseThriveSerializer]
+public class UnlockProgress
+{
+    [JsonProperty]
+    public bool UnlockAll;
```
hhyyrylainen, line 12: Why is this not a property? By default public fields should be avoided.

src/microbe_stage/organelle_unlocks/UnlockProgress.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    /// <summary>
+    ///   The organelles that the player can use.
+    /// </summary>
+    [JsonProperty]
+    private readonly HashSet<OrganelleDefinition> unlockedOrganelles =
+        SimulationParameters
+            .Instance
```
hhyyrylainen, line 20: The chop style shouldn't be preferred over the style of packing arguments. This is especially short in the grand scheme of things so I'd rewrite this as :
```
     SimulationParameters.Instance.GetAllOrganelles().Where(organelle => organelle.UnlockConditions == null).
     ToHashSet();
```

With the ToHashSet also on the same line if it fits.

src/microbe_stage/organelle_unlocks/UnlockProgress.cs
```diff
@@ -0,0 +1,79 @@
...
+///   Stores which organelles have been unlocked by the player.
+/// </summary>
+[UseThriveSerializer]
+public class UnlockProgress
+{
+    [JsonProperty]
+    public bool UnlockAll;
+
+    /// <summary>
+    ///   The organelles that the player can use.
```
hhyyrylainen, line 15: I'd add "The default value is the by default unlocked organelles (ones without a condition)" here as it took me a bit to understand that key part here.

src/microbe_stage/organelle_unlocks/StatisticBasedUnlockCondition.cs
```diff
@@ -0,0 +1,188 @@
...
+        {
+        }
+    }
+
+    /// <summary>
+    ///   The number of microbes engulfed by the player is above the specified amount.
+    /// </summary>
+    public class EngulfedMicrobesAbove : StatisticBasedUnlockCondition
+    {
+        public override StatsTrackerEvent RelevantEvent => StatsTrackerEvent.PlayerEngulfedOther;
```
hhyyrylainen, line 34: All of these properties that have a fixed value need to have json ignore on them, otherwise these are unnecessarily written out into the save data.

src/microbe_stage/organelle_unlocks/StatisticBasedUnlockCondition.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        [JsonProperty]
+        public int Required { get; set; }
+
+        [JsonIgnore]
+        public SimpleStatistic? Engulfed => (SimpleStatistic?)RelevantStatistic;
```
hhyyrylainen, line 40: I think making the base class generic would be better than needing to rely on casting in the derived class here. So this would inherit `StatisticBasedUnlockCondition<SimpleStatistic>`.

src/microbe_stage/organelle_unlocks/StatisticBasedUnlockCondition.cs
```diff
@@ -0,0 +1,188 @@
...
+
+    public class ReproduceInBiome : StatisticBasedUnlockCondition
+    {
+        [JsonIgnore]
+        private Biome biome = null!;
+
+        public override StatsTrackerEvent RelevantEvent => StatsTrackerEvent.PlayerReproduced;
+
+        [JsonProperty("Biome")]
+        public string RawBiome { get; set; } = string.Empty;
```
hhyyrylainen, line 97: Biome type as a registry type should be JSON loadable directly without a string here.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2073,6 +2102,7 @@ private void OnModifyPressed()
     /// </remarks>
     private void SetupMicrobePartSelections()
     {
+        GD.Print("Setup selection");
```
hhyyrylainen, line 2105: Debug print here

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -496,6 +516,31 @@ public void Resolve(SimulationParameters parameters)
...
+    {
+        LocalizedStringBuilder unlockRequirements = new();
+        if (UnlockConditions != null)
+        {
+            bool first = true;
+            foreach (var unlockCondition in UnlockConditions)
+            {
+                if (!first)
+                {
+                    unlockRequirements.Append(new LocalizedString("UNLOCK_OR"));
```
hhyyrylainen, line 532:
```suggestion
                    unlockRequirements.Append(new LocalizedString("UNLOCK_WITH_ANY_OF_FOLLOWING"));
```

I think having a really clear translation key here would be beneficial.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -496,6 +516,31 @@ public void Resolve(SimulationParameters parameters)
         }
     }
 
+    /// <summary>
+    ///   A bbcode string containing all the unlock conditions for this organelle.
+    /// </summary>
+    public LocalizedStringBuilder GenerateUnlockRequirementsText(GameProperties game)
+    {
+        LocalizedStringBuilder unlockRequirements = new();
```
hhyyrylainen, line 524: Would it be possible for this to take in a string builder as a parameter to allow reusing string builder objects?

hhyyrylainen (changes requested):
I tried to do a full review but just barely didn't manage it (there's a few files I didn't manage to take a look at yet that were changed here). I think the overall approach is doable and good. I have some comments regarding the specifics how to lay out the data references and some thrive codebase style uniformity things.

src/general/NewGameSettings.cs
ender-titan1: Indeed

src/gui_common/CollapsibleList.cs
ender-titan1: DetachAndQueueFree is used in RemoveItem so I assume it's preferred

hhyyrylainen: It's not really preferred, the order goes: (most preferred) QueueFree, Free, DetachAnd_ (group of methods, these are the least preferred and there's even an issue about removing these entirely).

ender-titan1: Useful to know, thanks!

src/microbe_stage/systems/EngulfedDigestionSystem.cs
ender-titan1: From my testing, engulfing those floating organelle leftover bits also triggered this, which is why I added this check here

hhyyrylainen: Ah yes, of course I was partly mixing up the current entity and the engulf target in my head. In case of the target, you should check for CellProperties, that's my goto property to check if I need to know if something is a microbe or not.

ender-titan1: Got it

src/microbe_stage/organelle_unlocks/StatisticBasedUnlockCondition.cs
ender-titan1: From what I can tell, this is never actually loaded into saves, just loaded from the registry.

But for futureproofing and clarity I'll add the attribute anyway

hhyyrylainen: So there's the statistics data and this unlock condition as separate classes? And one of those is saved while this is not. I guess that is probably good design as many unlocks do need similar kind of information so adding that one layer of decoupling might be necessary.

src/microbe_stage/organelle_unlocks/StatisticBasedUnlockCondition.cs
ender-titan1: You mean `BiomeType`? Well, the `Biome` class doesn't actually contain a `BiomeType` field.

Instead, I won't actually get the biome, just use the biome type.

ender-titan1: I'm a bit confused. Correct me if I'm wrong, but I don't think we can use the actual `Biome` type here. I tried and the json converter complained that it can't convert a string to a type of `Biome`. We're not trying to define new biomes here after all, just reference the already defined ones in `biomes.json` right?

In that case I believe the best option is to either use a string or use `BiomeType` as that contains all the data this unlock condition needs.

hhyyrylainen: That's one of those save tricky parts I mentioned. It should just work by default, like how organelles can be defined. Though there is a potential order of loading problem if the unlock conditions are retrieved from JSON, so the init order might need swapping in simulation parameters. Like I offered before, if you do the conversion to the "right" code I can take a look at getting the JSON loading wrangled into behaving nicely.

For dictionaries things get much more fun but a basic object reference should just work. As I already wrote this I'll share this bit of info with you: in case of dictionaries there's about 2 possible solutions, slapping `[TypeConverter(typeof(ThriveTypeConverter))]` on `Biome` or slapping `[JsonConverter(typeof(DictionaryWithJSONKeysConverter<Biome, float>))]` on the dictionary property. But again both of those should be unnecessary, it's much more likely to be an initialization order problem where the biome data is not yet available. That should be fixable as I don't foresee biome data referring back to organelle types which would cause a circular dependency.

ender-titan1: Good to know and I'm sure it might prove useful, though it is not what I was originally asking about. I wasn't referencing the dictionary in `ReproductionStatistic`, but that `Biome` cannot be loaded from the registry (not save files, this data is not even included in saves) as just a string. Replacing the field `RawBiome` with a field of type `Biome` throws an error that a sting cannot be converted into an object of type biome on registry load.

Deus-Codes:
Works great from playtesting and is an important addition. Great job. 

Question though: the unlock conditions for the eukaryotic parts are a good amount more lenient than the conditions described here, which generally required more commitment in morphology to unlock an organelle - https://forum.revolutionarygamesstudio.com/t/upgrades-unlocks-endosymbiosis-master-thread/915

Is there a particular reason for that? Since endosymbiosis isn't implemented yet, I think it's okay to have unlock conditions be a bit easier. But it is eventually important to make these conditions a bit more committal to ensure endosymbiosis is an important aspect of gameplay, and to reinforce specialization.

src/microbe_stage/organelle_unlocks/StatisticBasedUnlockCondition.cs
hhyyrylainen: I addressed that first in my comment. I think it is a data dependency ordering problem. If you can convert the code to what it should be here, I can look into the JSON deserialize problem.

src/microbe_stage/OrganelleDefinition.cs
ender-titan1: `UNLOCK_WITH_ANY_OF_FOLLOWING` does not match what the translation is. The translation string is literally just `or`.

However this is a good suggestion for the current `UNLOCK_WITH` key.

hhyyrylainen: "or" is very likely much too context dependent. So the translation key needs to do the heavy lifting here. I do agree though that my suggestion doesn't really match this context then (I didn't read the translation text yet so I didn't know it was "or").

ender-titan1: Guess I'll rename it to something like `OR_UNLOCK_CONDITION` then

ender-titan1:
Its mostly because I copied the unlock conditions from HyperCube's implementation for the most part, so if you have any suggestions regarding the difficulty of unlock conditions please say it.

For now I will make some of the unlock conditions a bit harder like you said and improve on Captain's feedback regarding the lysosome.

src/microbe_stage/organelle_unlocks/UnlockProgress.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// <summary>
+    ///   Is the organelle unlocked?
+    /// </summary>
+    public bool IsUnlocked(OrganelleDefinition organelle, GameProperties game, bool autoUnlock)
+    {
+        if (organelle.UnlockConditions == null || game.FreeBuild || UnlockAll)
+            return true;
+
+        if (organelle.UnlockConditions.Any(unlock => unlock.Satisfied()) && autoUnlock)
```
hhyyrylainen, line 56:
```suggestion
        if (autoUnlock && organelle.UnlockConditions.Any(unlock => unlock.Satisfied()))
```
I'd swap the condition here to be this way around, that way the more expensive satisfied condition loop is not executed if auto unlock happens to be false.

src/microbe_stage/organelle_unlocks/UndiscoveredOrganellesTooltip.tscn
```diff
@@ -0,0 +1,72 @@
...
+[node name="UndiscoveredOrganellesTooltip" type="PanelContainer"]
+margin_right = 350.0
+margin_bottom = 155.0
+rect_min_size = Vector2( 350, 0 )
+mouse_filter = 2
+theme = ExtResource( 9 )
+custom_styles/panel = SubResource( 1 )
+script = ExtResource( 7 )
+__meta__ = {
+"_editor_description_": "NOTE: processes description string should only be set here on the ProcessesDescription property and not directly on the rich text label node as it will be overidden otherwise."
```
hhyyrylainen, line 29: This seems like a node duplication unnecessarily copied property.

src/microbe_stage/organelle_unlocks/UndiscoveredOrganellesButton.tscn
```diff
@@ -0,0 +1,119 @@
...
+margin_bottom = 76.0
+rect_min_size = Vector2( 85, 0 )
+size_flags_horizontal = 4
+custom_constants/line_spacing = -3
+custom_fonts/font = ExtResource( 1 )
+text = "UNDISCOVERED_ORGANELLES"
+align = 1
+__meta__ = {
+"_edit_use_anchors_": false,
+"_editor_description_": "PLACEHOLDER"
```
hhyyrylainen, line 118: The text seems to be a translation key so why is this label marked as untranslated?

src/microbe_stage/organelle_unlocks/UndiscoveredOrganellesTooltip.cs
```diff
@@ -0,0 +1,97 @@
+﻿using Godot;
+
+/// <summary>
+///   The tooltip for undiscovered organelles
+/// </summary>
+public class UndiscoveredOrganellesTooltip : Control, ICustomToolTip
+{
+#pragma warning disable CA2213
```
hhyyrylainen, line 8: The dispose pragma disable *should not* be around the NodePaths as the style we use is to dispose those.

src/microbe_stage/organelle_unlocks/StatisticBasedUnlockCondition.cs
```diff
@@ -0,0 +1,181 @@
...
+    public class ReproduceWithOrganelle : StatisticBasedUnlockCondition<ReproductionStatistic>
+    {
+        [JsonIgnore]
+        private OrganelleDefinition organelle = null!;
+
+        [JsonIgnore]
+        public override StatsTrackerEvent RelevantEvent => StatsTrackerEvent.PlayerReproduced;
+
+        [JsonProperty("Organelle")]
+        public string RawOrganelle { get; set; } = string.Empty;
```
hhyyrylainen, line 140: There's still this raw value business here.

src/microbe_stage/organelle_unlocks/OrganelleUnlockHelpers.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using UnlockConstraints;
+
+public static class OrganelleUnlockHelpers
+{
+    private static IEnumerable<IUnlockCondition>? allUnlockConditions;
```
hhyyrylainen, line 7: This should not be global state. This should be instead stored in the GameWorld. And I guess that would basically remove the need for this class entirely so I'm in favour of rolling all of the code in this file into the GameWorld class directly (or another place if you can think of some place more sensible).

src/microbe_stage/editor/PatchMapDrawer.tscn
hhyyrylainen, line 1: I think changes in this file should probably be reverted for this PR?

src/microbe_stage/MicrobeStage.cs
```diff
@@ -515,6 +523,9 @@ public void MoveToMulticellular()
             throw new Exception("failed to keep the current scene root");
         }
 
+        // TODO: The multicellular stage needs to be able to track statistics and not break organelle unlocks
+        GameWorld.UnlockProgress.UnlockAll = true;
```
hhyyrylainen, line 527: Is there code directly in multicellular to unlock all? I think at least freebuild mode for multicellular will need that. And directly debugging starting later stages will be much worse if organelles aren't available and can't be unlocked normally.

CI09:
CI hasn't completed checks in like a day or so, I wonder why...

hhyyrylainen:
I think the commit was made just exactly when I was doing server updates so the event about new commits was never processed. The CI should run again just fine when a new commit is added for this PR.

src/general/statistics_tracking/ReproductionStatistic.cs
```diff
@@ -0,0 +1,56 @@
...
+    public StatsTrackerEvent LinkedEvent { get; set; } = StatsTrackerEvent.PlayerReproduced;
+
+    [JsonProperty]
+    public int TimesReproduced { get; set; }
+
+    [JsonProperty]
+    public Dictionary<Biome, int> ReproducedInBiomes { get; set; } = new();
+
+    [JsonProperty]
+    public Dictionary<string, int> ReproducedWithOrganelle { get; set; } = new();
```
hhyyrylainen, line 19:
```suggestion
    public Dictionary<OrganelleDefinition, int> ReproducedWithOrganelle { get; set; } = new();
```

ender-titan1:
Honestly I’m very sorry there are so many of these issues. While a lot of them affect HyperCube’s code I should have spotted them beforehand. I’ll try to put in more effort to try to spot these proactively for future PRs.

src/general/statistics_tracking/ReproductionStatistic.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    public void RecordPlayerReporduction(in Entity player, Biome? biome)
+    {
+        TimesReproduced++;
```
hhyyrylainen, line 23:
```suggestion
        ++TimesReproduced;
```

I vastly prefer the preincrement style whenever it is semantically valid usage.

src/general/statistics_tracking/ReproductionStatistic.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        var organelles = player.Get<OrganelleContainer>().Organelles?.Organelles.ToHashSet();
```
hhyyrylainen, line 25: This HashSet call here seems pretty useless. The organelles are looped just once and hashing the placed organelles to unique shouldn't even do anything.

src/general/statistics_tracking/ReproductionStatistic.cs
```diff
@@ -0,0 +1,56 @@
...
+    [JsonProperty]
+    public int TimesReproduced { get; set; }
+
+    [JsonProperty]
+    public Dictionary<Biome, int> ReproducedInBiomes { get; set; } = new();
+
+    [JsonProperty]
+    public Dictionary<string, int> ReproducedWithOrganelle { get; set; } = new();
+
+    public void RecordPlayerReporduction(in Entity player, Biome? biome)
```
hhyyrylainen, line 21:
```suggestion
    public void RecordPlayerReproduction(in Entity player, Biome? biome)
```

Typo in the name. Also shouldn't the biome be always required here? Is there a situation where the player can reproduce without being in any biome?

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -302,14 +312,101 @@ private void UpdateOrganelleUnlockTooltips()
...
+            button.RegisterToolTipForControl(tooltip, true);
+        }
+    }
+
+    private void RemoveUndisoveredOrganelleButtons()
+    {
+        foreach (var child in partsSelectionContainer.GetChildren())
+        {
+            if (child is CollapsibleList list)
+                list.RemoveAllOfType<UndiscoveredOrganellesButton>();
```
hhyyrylainen, line 409: Does this need to unregister the tooltips? I can't immediately remember if the AddToolTip call above works so that it automatically unregisters the tooltips.

I think as the specific group is used, it would be pretty simple to clear that group here.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -302,14 +312,101 @@ private void UpdateOrganelleUnlockTooltips()
...
+        }
+
+        // Remove any buttons that might've been created before
+        if (refresh)
+            RemoveUndisoveredOrganelleButtons();
+
+        // Generate undiscovered organelle buttons
+        foreach (var groupWithUndiscovered in groupsWithUndiscoveredOrganelles)
+        {
+            var group = partsSelectionContainer.GetNode<CollapsibleList>(groupWithUndiscovered.Key.ToString());
```
hhyyrylainen, line 389: This seems a bit brittle as this requires a specific name node. Though, I guess this might be exactly what's done elsewhere in the editor (I can't immediately remember), if this is done elsewhere in the editor then this is fine and not any more problematic than other editor code.

src/microbe_stage/organelle_unlocks/ConditionSet.cs
```diff
@@ -0,0 +1,81 @@
+﻿namespace UnlockConstraints
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+
+    public class ConditionSet : IUnlockCondition
```
hhyyrylainen, line 9: I think I can do this when fixing other JSON stuff but making this a generic collection would allow the JSON in organelles to avoid one level of extra wrapping for simpler more understandable way to specify the conditions.

src/microbe_stage/organelle_unlocks/OrganelleUnlockHelpers.cs
ender-titan1: I'd probably say UnlockProgress would be more sensible

src/microbe_stage/organelle_unlocks/StatisticBasedUnlockCondition.cs
ender-titan1: Wouldn't this cause some type of circular dependency? Unlock conditions are defined in organelle definitions, and in this case the unlock conditions would be referencing those organelle definitions. Wouldn't that mean that this could reference not yet defined organelles?

hhyyrylainen: Yeah that would cause a circular dependency. So this will likely have to stay. Would be nice to add a documentation comment here explaining that reasoning.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
ender-titan1: Well it does have `autoUnregister` set to true so it should work

src/microbe_stage/editor/CellEditorComponent.GUI.cs
ender-titan1: Yup it appears that's how it's used in the other parts of the editor code

src/microbe_stage/editor/CellEditorComponent.GUI.cs
hhyyrylainen: That's just the tooltip control data, right? I've never been 100% familiar with the tooltip system, but there's two different things: the core tooltip data, and then the control <-> tooltip mapping (this second type of data is the one with the auto unregister property).

src/microbe_stage/organelle_unlocks/ConditionSet.cs
ender-titan1: This is not needed anymore as I've implemented a custom Json Converter

src/microbe_stage/systems/EngulfingSystem.cs
```diff
@@ -1039,6 +1046,11 @@ private void EjectEverythingFromDeadEngulfer(ref Engulfer engulfer, in Entity en
                 }
             }
 
+            if (engulferEntity.Has<PlayerMarker>() && targetEntity.Has<CellProperties>())
+            {
+                gameWorld!.StatisticsTracker.TotalEngulfedByPlayer.Increment(1);
```
hhyyrylainen, line 1051: This should have the check for gameworld being set in PreUpdate.

src/microbe_stage/organelle_unlocks/UnlockProgress.cs
```diff
@@ -0,0 +1,126 @@
...
+            if (constraint is WorldBasedUnlockCondition condition)
+                condition.UpdateWorldData(World, data, energyBalance);
+        }
+    }
+
+    private void GetAllUnlockConditions()
+    {
+        allOrganelleUnlockConditions ??= SimulationParameters.Instance.GetAllOrganelles()
+            .Where(o => o.UnlockConditions != null)
+            .Select(o => o.UnlockConditions)
```
hhyyrylainen, line 121: Can't this already be the SelectMany line? This seems there's extra line here that doesn't need to be here. Also the same thing applies to `c.Requirements`

src/microbe_stage/organelle_unlocks/UndiscoveredOrganellesButton.cs
```diff
@@ -0,0 +1,54 @@
...
+        UpdateCount();
+    }
+
+    private void UpdateCount()
+    {
+        if (countLabel == null)
+            return;
+
+        countLabel.Visible = count > 1;
+        countLabel.Text = "x" + count;
```
hhyyrylainen, line 44: I think there is a multiple of something translation key already, that should be used here for localization support.

src/microbe_stage/organelle_unlocks/UndiscoveredOrganellesTooltip.cs
```diff
@@ -0,0 +1,111 @@
...
+    private Label? nameLabel;
+    private CustomRichTextLabel? unlockTextLabel;
+#pragma warning restore CA2213
+
+    private string? displayName;
+    private LocalizedStringBuilder? unlockText;
+
+    [Export]
+    public float DisplayDelay { get; set; }
```
hhyyrylainen, line 23: This should be a field for consistency in Thrive. (so that this can be grouped up with the other export variables).

src/general/statistics_tracking/WorldStatsTracker.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    [JsonProperty]
+    public SimpleStatistic TotalPlayerDeaths { get; private set; } = new(StatsTrackerEvent.PlayerDied);
+
+    [JsonProperty]
+    public ReproductionStatistic PlayerReproductionStatistic { get; private set; } = new();
+
+    public IStatistic[] CollectStatistics()
+    {
+        return allStatistics ??= new IStatistic[]
```
hhyyrylainen, line 25: `Lazy` is still not used here.

src/general/GameWorld.cs
```diff
@@ -113,6 +126,8 @@ public GameWorld()
 
         // Register glucose reduction
         TimedEffects.RegisterEffect("reduce_glucose", new GlucoseReductionEffect(this));
+
+        UnlockProgress.World = this;
```
hhyyrylainen, line 130: This should not be required, if this is then UnlockProgress object is not correctly being loaded from saves...

src/microbe_stage/organelle_unlocks/UnlockProgress.cs
```diff
@@ -0,0 +1,126 @@
...
+            typeof(MicrobeStage),
+        };
+    }
+
+    public void InitUnlockConditionTracking()
+    {
+        GetAllUnlockConditions();
+
+        foreach (var constraint in allOrganelleUnlockConditions!)
+            constraint.OnInit(World.StatisticsTracker);
```
hhyyrylainen, line 53: Am I right in assuming that it seems like the unlock condition objects from the registry are directly used here and their state changes? If I'm correct then this is a pretty bad architecture problem. Any registry loaded type should be immutable. Instead the design should be "inverted" so that whenever the constraints are checked the statistics tracker needs t obe passed in as a parameter there.

src/microbe_stage/organelle_unlocks/UndiscoveredOrganellesTooltip.cs
ender-titan1: It is required to be a property by the interface this class implements, besides this is done the same way in other custom tooltip classes.

hhyyrylainen: Ah, I did not realize. That's the one exception when ExportAttribute is good to use on properties: when it is required by some external factors.

src/microbe_stage/organelle_unlocks/IUnlockCondition.cs
```diff
@@ -0,0 +1,39 @@
+﻿namespace UnlockConstraints
+{
+    using System;
+
+    public interface IUnlockCondition
+    {
+        public bool Satisfied(EventArgs args);
+
+        public void GenerateTooltip(LocalizedStringBuilder builder, EventArgs args);
```
hhyyrylainen, line 9: This isn't an event so why is `EventArgs` type used here? It'd be better to use an empty interface that both statistics and world and player data classes implement.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
hhyyrylainen: I'm unresolving this as I don't see this code changed (sorry if you addressed this in one of your commits, but it's been way too much for me to read through each one).

src/microbe_stage/organelle_unlocks/ConditionSet.cs
```diff
@@ -0,0 +1,72 @@
...
+            var first = true;
+            foreach (var entry in Requirements)
+            {
+                if (!first)
+                {
+                    builder.Append(new LocalizedString("AND_UNLOCK_CONDITION"));
+                    builder.Append(" ");
+                }
+
+                var color = IsSatisfied(entry, worldAndPlayerArgs, statisticArgs) ? "green" : "red";
```
hhyyrylainen, line 37: These should use colour codes and be defined in Constants.cs. Also the pure red and green are likely way too strong colours and should be diluted a bit to match the other reds and greens used in the game.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
ender-titan1: Pretty sure the call to `ToolTipManager.Instance.ClearToolTips` under this loop deletes all the related tooltips in the group.

hhyyrylainen: Yep, I see it now. I had once again accidentally been reading the outdated diff on github (as it is not very visible the one button you need to press to view all changes).

locale/en.po
```diff
@@ -5233,14 +5258,18 @@ msgstr "Protect just created species from species cap"
...
+#: ../simulation_parameters/microbe_stage/organelles.json:154
 msgid "PROTOPLASM"
 msgstr "Protoplasm"
 
 #: ../src/gui_common/CreditsScroll.cs:530
 msgid "PULL_REQUESTS_PROGRAMMING"
 msgstr "Pull Requests / Programming"
 
+#: ../src/microbe_stage/organelle_unlocks/UndiscoveredOrganellesButton.tscn:93
+msgid "QUESTION"
```
hhyyrylainen, line 5270: Note to self: this is too unclear translation key which I should fix before merging.

hhyyrylainen (approved):
I've now made some final cleanups I still wanted to see on this branch. I also did manual squash as there's now 3 authors. Hopefully I didn't mess anything up with that so that merging this doesn't cause any issues.

PR #4744, opened 2023-12-23
Add filtering buttons to save manager (menu)

CI09:

Noone requested it, but I already made it, so if anyone approves it, it may be merged.
Adds filtering buttons to save manager in menu. Clicking on them hides save items outside of filter and shows the filtered ones.

image

---

my TODO:
- [x] Add sounds for clicking buttons

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09:
What's about these conflicting files? I am kinda new to this codebase...

ender-titan1:
Those really are not a problem, conflicting translation files happen all the time. The solution to this is to run `git merge` with the `-X theirs` flag which should auto-resolve these conflicts.

CI09:
But do I have to run it or can I just leave it?

CI09:
By the way, it only filters save items that have loaded already which sometimes takes some time

ender-titan1:
> But do I have to run it or can I just leave it?

Yeah merge conflicts have to be resolved before PRs can be merged

> By the way, it only filters save items that have loaded already which sometimes takes some time

We’ll do a review once this marked as non draft

CI09:
Okay, I just reverted the translations update commit and it still works, so I guess it's ready for review.

ender-titan1 (changes requested):
A dropdown (in godot an `OptionButton`) would probably be better suited for this situation

CI09:
image

ender-titan1 (changes requested):
I think what would solve the issue would be to:
A - Set the filter to All whenever the menu is entered, and
B - Disable the dropdown when the menu is being refreshed

CI09:
So... I don't know if 

> B

would be easily possible...
If I am correct loading the saves is queued in task executor so I don't know if there's a viable or easy option to reenable filters once it finished loading.

ender-titan1:
Pretty sure the field `refreshing` of `SaveList` can be used

CI09:
Okay, so the thing is `refreshing` is set to false when saves are loaded, but I need to get some signal for when information of these saves are loaded
image

CI09:
I added counter for counting all saves and counter for counting all saves with data loaded and once they are equivalent it enables filters.

src/saving/SaveList.cs
```diff
@@ -101,6 +104,8 @@ public class SaveList : ScrollContainer
     [Signal]
     public delegate void OnSaveLoaded(string saveName);
 
+    [Signal]
+    public delegate void OnItemsDataLoadedCompleted();
     public override void _Ready()
```
ender-titan1, line 109: There should be a newline between these two elements, that's why Jetbrains is failing

src/saving/SaveList.cs
```diff
@@ -75,6 +75,9 @@ public class SaveList : ScrollContainer
     private PackedScene listItemScene = null!;
 #pragma warning restore CA2213
 
+    private int allItemsCount = 0;
+    private int loadedItems = 0;
```
ender-titan1, line 79: These items do not need default values, another reason why checks are complaining

src/saving/SaveList.cs
```diff
@@ -207,12 +215,38 @@ public void Refresh()
...
+        allItemsCount = 0;
+        loadedItems = 0;
+
         loadingItem.Visible = true;
         readSavesList = new Task<List<string>>(() => SaveHelper.CreateListOfSaves());
         TaskExecutor.Instance.AddTask(readSavesList);
         EmitSignal(nameof(OnItemsChanged));
     }
 
+    public void Filter(SaveInformation.SaveType saveTypeFilter = SaveInformation.SaveType.Manual, bool showAll = false)
```
ender-titan1, line 227: I don't think that first argument requires a default

src/saving/SaveList.cs
```diff
@@ -207,12 +215,38 @@ public void Refresh()
...
+    {
+        foreach (var item in savesList.GetChildren())
+        {
+            SaveListItem? saveItem = item as SaveListItem;
+
+            if (saveItem != null)
+            {
+                if (showAll)
+                {
+                    ((CanvasItem)item).Show();
```
ender-titan1, line 237: If I'm not mistaken `SaveListItem` inherits from `CanvasItem` so these casts are unnecessary

src/saving/SaveList.cs
```diff
@@ -452,4 +486,14 @@ private void LoadSave()
         saveToBeLoaded = null;
         isLoadingSave = false;
     }
+
+    private void CountLoadedItems()
+    {
+        loadedItems += 1;
```
ender-titan1, line 492: You can use `loadedItems++` here instead

src/saving/SaveManagerGUI.cs
```diff
@@ -103,8 +107,17 @@ public override void _Ready()
...
         saveDirectoryWarningDialog = GetNode<CustomConfirmationDialog>(SaveDirectoryWarningDialogPath);
         errorSaveDeletionFailed = GetNode<CustomConfirmationDialog>(SaveDeletionFailedErrorPath);
+        filterOptions = GetNode<OptionButton>(FilterOptionsPath);
 
         saveList.Connect(nameof(SaveList.OnItemsChanged), this, nameof(RefreshSaveCounts));
+
+        filterOptions.AddItem(TranslationServer.Translate("ALL"), 0);
+        filterOptions.AddItem(TranslationServer.Translate("SAVE_MANUAL"), 1);
+        filterOptions.AddItem(TranslationServer.Translate("SAVE_AUTOSAVE"), 2);
+        filterOptions.AddItem(TranslationServer.Translate("SAVE_QUICKSAVE"), 3);
```
ender-titan1, line 117: Instead of using hard coded numbers here, convert the corresponding `SaveType` to an int

src/saving/SaveManagerGUI.cs
```diff
@@ -317,4 +334,30 @@ private void OnSaveListItemConfirmed(SaveListItem item)
...
+                break;
+            case 1:
+                saveList.Filter(SaveInformation.SaveType.Manual);
+                break;
+            case 2:
+                saveList.Filter(SaveInformation.SaveType.AutoSave);
+                break;
+            case 3:
+                saveList.Filter(SaveInformation.SaveType.QuickSave);
+                break;
```
ender-titan1, line 355: Instead of a switch statement, again, just convert the index to `SaveInformation.SaveType`

CI09:
I might need help with the check failure, but only if someone _wants_ to help me, I'm not forcing anybody.

src/saving/SaveManagerGUI.cs
CI09: What about All saves?

ender-titan1: You can leave the All saves as is - just leave a comment explaining why it is equal to 0. I think that should suffice.

This would also mean you'd have to manually set `SaveType.Manual` to 1, `SaveType.Autosave` to 2 and so on... in the actual enum.

CI09:
I just made option `3` (All) the standard option.

src/saving/SaveList.cs
```diff
@@ -207,12 +216,38 @@ public void Refresh()
...
+                if (showAll)
+                {
+                    saveItem.Show();
+                }
+                else
+                {
+                    if (saveItem.SaveType != saveTypeFilter)
+                        saveItem.Hide();
+                    else
+                        saveItem.Show();
```
ender-titan1, line 245: If/Else statements require braces I believe

src/saving/SaveManagerGUI.cs
```diff
@@ -317,4 +336,30 @@ private void OnSaveListItemConfirmed(SaveListItem item)
...
+            case (int)SaveInformation.SaveType.AutoSave:
+                saveList.Filter(SaveInformation.SaveType.AutoSave);
+                break;
+            case (int)SaveInformation.SaveType.QuickSave:
+                saveList.Filter(SaveInformation.SaveType.QuickSave);
+                break;
+            case 3:
+                saveList.Filter(SaveInformation.SaveType.Manual, true);
+                break;
+        }
```
ender-titan1, line 358: You could do something like this instead.

(You can use `null!` here as that argument is never read if `showAll` is set to true)

```C#
if (index == 3)
{
    saveList.Filter(null!, true);
    return;
}

saveList.Filter((SaveInformation.SaveType)index);
```

CI09:
You could probably do it way faster yourself but anyways.

ender-titan1 (approved):
> You could probably do it way faster yourself but anyways.

I've commited onto your branch to fix up some issues and add a tooltip so this could be done faster. Everything looks good to me from the code side of things.

CI09:
Also, I'd recommend discussing this feature with the other team members considering it is quite slow thus may be not even useful. Also GUI placement might be tweaked, but I don't know...

src/saving/SaveManagerGUI.cs
```diff
@@ -103,8 +107,19 @@ public override void _Ready()
         deleteOldConfirmDialog = GetNode<CustomConfirmationDialog>(DeleteOldConfirmDialogPath);
         saveDirectoryWarningDialog = GetNode<CustomConfirmationDialog>(SaveDirectoryWarningDialogPath);
         errorSaveDeletionFailed = GetNode<CustomConfirmationDialog>(SaveDeletionFailedErrorPath);
+        filterOptions = GetNode<OptionButton>(FilterOptionsPath);
 
         saveList.Connect(nameof(SaveList.OnItemsChanged), this, nameof(RefreshSaveCounts));
+
+        filterOptions.AddItem(TranslationServer.Translate("SAVE_MANUAL"), (int)SaveInformation.SaveType.Manual);
```
hhyyrylainen, line 114: Using enum values and getting the thing to translate like (example from different context):
```c#
        foreach (var mode in new[] { FogOfWarMode.Ignored, FogOfWarMode.Regular, FogOfWarMode.Intense })
        {
            fogOfWarModeDropdown.AddItem(
                TranslationServer.Translate(mode.GetAttribute<DescriptionAttribute>().Description), (int)mode);
        }
```

this also means that the `3` value below may not be hardcoded as adding new save types would conflict.

src/saving/SaveList.cs
```diff
@@ -207,12 +216,40 @@ public void Refresh()
...
+
         loadingItem.Visible = true;
         readSavesList = new Task<List<string>>(() => SaveHelper.CreateListOfSaves());
         TaskExecutor.Instance.AddTask(readSavesList);
         EmitSignal(nameof(OnItemsChanged));
     }
 
+    public void Filter(SaveInformation.SaveType? saveTypeFilter)
+    {
+        foreach (var item in savesList.GetChildren())
```
hhyyrylainen, line 230: GetChildren should be avoided, the style in Thrive is to store child nodes in a list. A new field like `List<SaveListItem> savesListItems` needs to be added, I didn't originally add such a thing as it wasn't necessary as the list didn't need to care about its child items in the past.

hhyyrylainen (changes requested):
I think this is fine to add as a feature. This isn't really important feature to me but with the dropdown now cluttering the GUI, I can accept this as long as the code style / minor code approach details are improved.

PR #4747, opened 2023-12-25
Add middle mouse button panning for strategy camera

### LINKED ISSUE #4271: Make StrategicCamera (strategy stage) pan when holding down middle mouse and moving the cursor
so that, that natural thing for strategy game players (and consistency with out editors) works.

CI09:

This PR adds middle mouse button panning for strategy camera and input for gameplay middle mouse button panning in input  map.

**Related Issues**

Closes #4271 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/society_stage/StrategicCamera.cs
```diff
@@ -137,6 +138,35 @@ public bool PanCamera(float delta, float forwardMovement, float leftRightMovemen
...
+            ApplyPan(movement * CameraPanSpeed * ZoomLevel * ZoomLevelMovementSpeedModifier * delta / -10);
+        }
+
+        return false;
+    }
+
+    [RunOnKeyUp("g_pan_mouse")]
+    public bool ReleasePanCameraWithMouse()
+    {
+        if (!Visible)
```
hhyyrylainen, line 163: Shouldn't the mouse pan still be reset even if this is not visible?
Also eating the mouse up event is rarely a good idea.

src/society_stage/StrategicCamera.cs
```diff
@@ -137,6 +138,35 @@ public bool PanCamera(float delta, float forwardMovement, float leftRightMovemen
         return true;
     }
 
+    [RunOnKey("g_pan_mouse", CallbackRequiresElapsedTime = false)]
+    public bool PanCameraWithMouse(float delta)
```
hhyyrylainen, line 142:
```suggestion
    public bool PanCameraWithMouse(float dummy)
```

I'd name the parameter like that to ensure no one later accidentally tries to use it and gets surprised when the variable doesn't actually contain the delta.

hhyyrylainen (changes requested):
This looks just about how I'd do the core implementation. Besides the code comments I have two thoughts:
- How does the microbe editor do this? Wouldn't it be nice to have the same mouse input trigger both actions?
- The new input should be added to the input options JSON so that it can be rebound to a different key if someone wishes to do so.

CI09:
> * How does the microbe editor do this? Wouldn't it be nice to have the same mouse input trigger both actions?
I think microbe editor utilizes e_pan_mouse. I am pretty sure they work the same way, but some inputs had their e_ and g_ counterparts, so I tried to make it consistent. It can be removed if unnecessary.

hhyyrylainen (approved):
This change looks fine. I'll separately add the missing entry in the input options for `e_pan_mouse`.

hhyyrylainen:
Actually I'll need to do some other tweaks as I noticed that the mouse up handling is inconsistent *and* multiple of the mouse pan callbacks actually use the delta value *after* specifying they don't want to use it.

hhyyrylainen:
And I found a bunch of bugs that had snuck in at some point (I don't suspect this PR was the cause) about the mouse panning. I fixed all of that and made the strategy mode mouse pan snappy to exactly mirror mouse movements: https://github.com/Revolutionary-Games/Thrive/pull/4759

PR #4751, opened 2024-01-01
Late multicellular musculature dependant speed

CI09:

It isn't really necessary right now, but it makes late multicellular organisms' speed dependant on amount of muscular tissue (cell of that tissue has at least one myofibril). It might be fun addition until development focus switches to the stage itself.

I made a poll on the forums: https://community.revolutionarygamesstudio.com/t/thrive-development-polls/2971/34

Thing to note: there is no speed limit at the moment, so the organisms can swim/walk pretty fast.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/late_multicellular_stage/LateMulticellularSpecies.cs
```diff
@@ -190,6 +194,23 @@ private static float CalculateBrainPowerFromLayout(MetaballLayout<MulticellularM
...
 
+    private static float CalculateMuscularPowerFromLayout(MetaballLayout<MulticellularMetaball> layout, float scale)
+    {
+        float result = 0;
+
+        foreach (var metaball in layout)
+        {
+            if (metaball.CellType.IsMuscularTissueType())
+            {
+                // TODO: check that volume scaling in physically sensible way (using GetVolume) is what we want here
```
hhyyrylainen, line 205: This comment is copy-pasted without adjustment from the brain power calculation...

hhyyrylainen (changes requested):
I have just one small code comment. Does the speed get way too much with a bunch of muscle cells? If there is a chance that someone adds too much muscle and can't effectively pick up the items needed to advance, then a speed cap is kind of necessary.

CI09:
Shall I use `Mathf.Clamp()`?

hhyyrylainen:
Having a max speed would be excellent for making sure the game cannot become unplayable.
Just don't make the default speed too much slower to be annoying.

CI09:
Also, I realized that organism's `Mass` could be utilized to slow down the movement, so muscles would have bigger role. Not necessary, but still a nice addition (I think).

hhyyrylainen:
I wouldn't overcomplicate things with that. See for example what I needed to do to convert the mass based microbe approach to density based (to work with the Jolt physics engine), as that's what will need to happen to the prototypes as well when they are no longer prototypes.

PR #4774, opened 2024-01-05
Fix 3D environment for cell editor

CI09:

This PR makes game disable 3D environment when entering cell editor and brings it back when going back to body plan editor.

**Related Issues**

https://community.revolutionarygamesstudio.com/t/ambient-color-applies-in-late-multicellular-editor/7150

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/late_multicellular_stage/editor/LateMulticellularEditor.cs
+++ b/src/late_multicellular_stage/editor/LateMulticellularEditor.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Godot;
 using Newtonsoft.Json;
+using Environment = Godot.Environment;
 
 [JsonObject(IsReference = true)]
 [SceneLoadedClass("res://src/late_multicellular_stage/editor/LateMulticellularEditor.tscn")]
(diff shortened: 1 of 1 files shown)


src/late_multicellular_stage/editor/LateMulticellularEditor.cs
```diff
@@ -333,6 +335,9 @@ protected override void ApplyEditorTab()
         cellEditorTab.Hide();
         noCellTypeSelected.Hide();
 
+        // Remember environment
```
hhyyrylainen, line 338: This comment seems very redundant.

src/late_multicellular_stage/editor/LateMulticellularEditor.cs
```diff
@@ -63,6 +64,7 @@ public class LateMulticellularEditor : EditorBase<EditorAction, MulticellularSta
     private Light bodyEditorLight = null!;
 
     private WorldEnvironment worldEnvironmentNode = null!;
+    private Environment? environment;
```
hhyyrylainen, line 67: I think this variable might need to be disposed, as if the environment is currently detached when this node is removed from the scene tree the environment will cause a resource leak as it won't be automatically disposed due to being unattached.

hhyyrylainen (approved):
I guess I already looked at this enough, and I'll trust you at this point to have playtested this enough so I'll just merge this (as I just noticed this on my open list of stuff to review).

PR #4788, opened 2024-01-10
Auto evo prediction now shows energy first and population second

### LINKED ISSUE #3845: Modify auto-evo prediction to show total energy first (and population second)
this way players can see their energy gain of their species, this should hopefully reduce people incorrectly interpreting becoming bigger having a negative impact on their population.

https://forum.revolutionarygamesstudio.com/t/redefining-species-success/964

I'd say the display should be "energy (population)", and the tutorial about the prediction should be also updated a bit to make this clear. If there's an icon to be shown it should be for the total energy and not the less important, secondary value of population.

CI09:

This PR makes auto evo prediction in editor show energy first and population second, in brackets.
image
Also renames population label variable to energy label for consistency.

***Tutorial is not yet updated and there might be tooltip necessary perhaps.*** 
Edit: not anymore.

**Related Issues**

Closes #3845 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -92,7 +92,7 @@ public partial class CellEditorComponent :
     public NodePath AutoEvoPredictionPanelPath = null!;
 
     [Export]
-    public NodePath TotalPopulationLabelPath = null!;
+    public NodePath TotalEnergyLabelPath = null!;
 
     [Export]
     public NodePath AutoEvoPredictionFailedLabelPath = null!;
@@ -202,7 +202,7 @@ public partial class CellEditorComponent :
 
     private Label generationLabel = null!;
 
-    private CellStatsIndicator totalPopulationLabel = null!;
+    private CellStatsIndicator totalEnergyLabel = null!;
     private Label autoEvoPredictionFailedLabel = null!;
     private Label bestPatchLabel = null!;
     private Label worstPatchLabel = null!;
@@ -593,7 +593,7 @@ public partial class CellEditorComponent :
         digestionSpeedLabel = GetNode<CellStatsIndicator>(DigestionSpeedLabelPath);
         digestionEfficiencyLabel = GetNode<CellStatsIndicator>(DigestionEfficiencyLabelPath);
         generationLabel = GetNode<Label>(GenerationLabelPath);
-        totalPopulationLabel = GetNode<CellStatsIndicator>(TotalPopulationLabelPath);
+        totalEnergyLabel = GetNode<CellStatsIndicator>(TotalEnergyLabelPath);
         autoEvoPredictionFailedLabel = GetNode<Label>(AutoEvoPredictionFailedLabelPath);
         worstPatchLabel = GetNode<Label>(WorstPatchLabelPath);
         bestPatchLabel = GetNode<Label>(BestPatchLabelPath);
@@ -1334,7 +1334,7 @@ public partial class CellEditorComponent :
                 DigestionEfficiencyLabelPath.Dispose();
                 GenerationLabelPath.Dispose();
                 AutoEvoPredictionPanelPath.Dispose();
-                TotalPopulationLabelPath.Dispose();
+                TotalEnergyLabelPath.Dispose();
                 AutoEvoPredictionFailedLabelPath.Dispose();
                 WorstPatchLabelPath.Dispose();
                 BestPatchLabelPath.Dispose();
@@ -2359,7 +2359,7 @@ public partial class CellEditorComponent :
     {
         if (autoEvoPredictionRunSuccessful is false)
         {
-            totalPopulationLabel.Value = float.NaN;
+            totalEnergyLabel.Value = float.NaN;
             autoEvoPredictionFailedLabel.Show();
         }
         else
@@ -2426,10 +2426,23 @@ public partial class CellEditorComponent :
         autoEvoPredictionRunSuccessful = true;
 
         // Set the initial value
-        totalPopulationLabel.ResetInitialValue();
-        totalPopulationLabel.Value = run.PlayerSpeciesOriginal.Population;
+        totalEnergyLabel.ResetInitialValue();
+        totalEnergyLabel.Value = run.PlayerSpeciesOriginal.Population;
 
-        totalPopulationLabel.Value = newPopulation;
+        results.GetPatchEnergyResults(run.PlayerSpeciesNew).TryGetValue(
+            Editor.CurrentPatch, out var newTotalPatchEnergyGathered);
+        var totalEnergyLabelText = totalEnergyLabel.GetNode<Label>(totalEnergyLabel.ValuePath);
+
+        if (newTotalPatchEnergyGathered != null)
+        {
+            totalEnergyLabel.Value = newTotalPatchEnergyGathered.TotalEnergyGathered;
+            if (totalEnergyLabelText != null)
+            {
+                totalEnergyLabelText.Text =
+                    @$"{StringUtils.ThreeDigitFormat(
+                        newTotalPatchEnergyGathered.TotalEnergyGathered)} ({newPopulation})";
+            }
+        }
 
         var sorted = results.GetPopulationInPatches(run.PlayerSpeciesNew).OrderByDescending(p => p.Value).ToList();
 
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -471,6 +471,13 @@ public partial class CellEditorComponent :
     [JsonIgnore]
     public TutorialState? TutorialState { get; set; }
 
+    /// <summary>
+    ///   Needed for auto-evo prediction to be able to compare the new energy to the old energy
+    /// </summary>
+    [JsonProperty]
+    public float? PreviousPlayerGatheredEnergy { get; set; }
+
+    [JsonIgnore]
     public IEnumerable<OrganelleDefinition> PlacedUniqueOrganelles => editedMicrobeOrganelles
         .Where(p => p.Definition.Unique)
         .Select(p => p.Definition);
@@ -2362,7 +2369,7 @@ public partial class CellEditorComponent :
 
     private void UpdateAutoEvoPredictionTranslations()
     {
-        if (autoEvoPredictionRunSuccessful is false)
+        if (autoEvoPredictionRunSuccessful == false)
         {
             totalEnergyLabel.Value = float.NaN;
             autoEvoPredictionFailedLabel.Show();
@@ -2372,14 +2379,14 @@ public partial class CellEditorComponent :
             autoEvoPredictionFailedLabel.Hide();
         }
 
-        var populationFormat = TranslationServer.Translate("POPULATION_IN_PATCH_SHORT");
+        var energyFormat = TranslationServer.Translate("ENERGY_IN_PATCH_SHORT");
 
         if (!string.IsNullOrEmpty(bestPatchName))
         {
             var formatted = StringUtils.ThreeDigitFormat(bestPatchEnergyGathered);
 
             bestPatchLabel.Text =
-                populationFormat.FormatSafe(TranslationServer.Translate(bestPatchName), formatted);
+                energyFormat.FormatSafe(TranslationServer.Translate(bestPatchName), formatted);
         }
         else
         {
@@ -2391,12 +2398,18 @@ public partial class CellEditorComponent :
             var formatted = StringUtils.ThreeDigitFormat(worstPatchEnergyGathered);
 
             worstPatchLabel.Text =
-                populationFormat.FormatSafe(TranslationServer.Translate(worstPatchName), formatted);
+                energyFormat.FormatSafe(TranslationServer.Translate(worstPatchName), formatted);
         }
         else
         {
             worstPatchLabel.Text = TranslationServer.Translate("N_A");
         }
+    }
+
+    private void DummyKeepTranslation()
+    {
+        // This keeps this translation string existing if we ever still want to use worst and best population numbers
+        TranslationServer.Translate("POPULATION_IN_PATCH_SHORT");
     }
 
     private void OpenAutoEvoPredictionDetails()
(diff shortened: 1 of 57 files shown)


hhyyrylainen:
> Tutorial is not yet update and there might be tooltip necessary perhaps.

In my opinion the tutorial is an absolute requirement before this can be merged. A tooltip would be really good for existing players to find out what the new change is about, but I don't consider that as absolutely required.

CI09:
Like this `This panel shows a prediction of your future total energy gathered and population. These numbers are from the simulation of auto-evo.`?

hhyyrylainen:
That seems pretty good, tough maybe I'd recommend `This panel shows a prediction of your future total energy gathered (and population in parentheses). These numbers are from the simulation of auto-evo.`

That way it explains the number format as well to avoid any confusion.

CI09:
Ok now just the conflicts left to go... I just don't know how to resolve them. @ender-titan1  already explained me to merge (thanks), but I think I just lack the skill to even do it correctly, so uhh...

hhyyrylainen:
You need an up to date copy of the Thrive main repo master branch locally on your computer. Then you need to either rebase this branch on that or merge master into this branch. How to get that depends on how you've setup your git remotes.

CI09:
I might have messed up my local files. (format_script check)...

ender-titan1:
Try running the localization update script again

CI09:
I merged some fog-of-war translations too, that might be causing the problem.

ender-titan1:
Idk what's happening here, but the best course of action would probably be to delete the locale folder on your local project and grab the newest folder from master, then run the translation update script again.

CI09:
Ok I did some changes, tutorial might need some testing though because I don't know how to trigger auto-evo explaination tutorial. Also tooltip is yet (if) to be implemented.

hhyyrylainen:
That tutorial should trigger automatically the second time you go into the microbe editor (and as long as tutorials are enabled).

CI09:
Okay, it is updated.

CI09:
image

By the way why git shows me changes to margin in .tscn files? Should I avoid them?

hhyyrylainen:
> By the way why git shows me changes to margin in .tscn files? Should I avoid them?

They happen all the time, it's fine to leave them in files you have otherwise modified.

I think the tooltip is a bit weird to have "Energy" capitalized. We don't us the style of capitalizing random words in the game. Also it should clarify that the population number is in parentheses as I think that's a pretty key term to use to guide the player to read the numbers correctly.

CI09:
> I think the tooltip is a bit weird to have "Energy" capitalized.

Yeah, I didn't notice it.

CI09:
image
Other than energy word parentheses I tweaked the sentence a bit.

hhyyrylainen:
Seems pretty fine to me. Do the other numbers also have a tooltip? I think they should also have it as the player might not hover over the first number (only first looking at the other numbers) and then be confused maybe something like "Energy gathered by your species and resulting population (in parentheses)", something like that could work as the tooltip for all the other prediction panel numbers.

CI09:
For some reason the others (best path, etc.) tooltips don't work (they just don't display while hovering above them). I also noticed auto-evo panel tooltip would likely need small update, like this maybe?
`This panel shows the expected total energy gathered and population numbers from auto-evo for the edited species.
Auto-evo runs the population simulation that is the other part (besides your own performance) that affects your population.`

hhyyrylainen:
I think it is a good idea to update that as well. That tooltip could have the info that population is in parentheses. I suspect the other numbers have mouse ignore set to on so that the prediction panel tooltip is easier to see, which I think is fine if that tooltip also properly explains the format of the numbers.

CI09:
`This panel shows the expected total energy gathered and population numbers in parentheses from auto-evo for the edited species.
Auto-evo runs the population simulation that is the other part (besides your own performance) that affects your population.
`

> if that tooltip also properly explains the format of the numbers.

Explaining best/worst patch indicators?

hhyyrylainen:
I guess that could be added but if you have trouble with getting tooltips for them to show up, it's fine to not have them. The players have managed so far without tooltips on those.

CI09:
It would be ready then I think.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2410,10 +2410,23 @@ private void OnAutoEvoPredictionComplete(PendingAutoEvoPrediction run)
         autoEvoPredictionRunSuccessful = true;
 
         // Set the initial value
-        totalPopulationLabel.ResetInitialValue();
-        totalPopulationLabel.Value = run.PlayerSpeciesOriginal.Population;
+        totalEnergyLabel.ResetInitialValue();
+        totalEnergyLabel.Value = run.PlayerSpeciesOriginal.Population;
```
hhyyrylainen, line 2414: Why is the total energy label assigned the value of population here? It seems this is overridden anyway a few lines below this, but it seems pretty silly to set this at all to the wrong value.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2410,10 +2410,23 @@ private void OnAutoEvoPredictionComplete(PendingAutoEvoPrediction run)
...
+            totalEnergyLabel.Value = newTotalPatchEnergyGathered.TotalEnergyGathered;
+            if (totalEnergyLabelText != null)
+            {
+                totalEnergyLabelText.Text =
+                    @$"{StringUtils.ThreeDigitFormat(
+                        newTotalPatchEnergyGathered.TotalEnergyGathered)} ({newPopulation})";
+            }
+        }
 
         var sorted = results.GetPopulationInPatches(run.PlayerSpeciesNew).OrderByDescending(p => p.Value).ToList();
```
hhyyrylainen, line 2431: It seems you didn't modify the best patch and worst patch numbers. They should also be updated to the new format (energy first and then in parentheses the population).

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2410,10 +2410,23 @@ private void OnAutoEvoPredictionComplete(PendingAutoEvoPrediction run)
...
-        totalPopulationLabel.Value = run.PlayerSpeciesOriginal.Population;
+        totalEnergyLabel.ResetInitialValue();
+        totalEnergyLabel.Value = run.PlayerSpeciesOriginal.Population;
 
-        totalPopulationLabel.Value = newPopulation;
+        results.GetPatchEnergyResults(run.PlayerSpeciesNew).TryGetValue(
+            Editor.CurrentPatch, out var newTotalPatchEnergyGathered);
+        var totalEnergyLabelText = totalEnergyLabel.GetNode<Label>(totalEnergyLabel.ValuePath);
+
+        if (newTotalPatchEnergyGathered != null)
```
hhyyrylainen, line 2420: The usual style I do is `if (thing.TryGetValue()) ...` and then use null suppression inside the if-statement if necessary.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2410,10 +2410,23 @@ private void OnAutoEvoPredictionComplete(PendingAutoEvoPrediction run)
...
+        {
+            totalEnergyLabel.Value = newTotalPatchEnergyGathered.TotalEnergyGathered;
+            if (totalEnergyLabelText != null)
```
hhyyrylainen, line 2423: Isn't this a label that should be always set so null override should be used here?

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2410,10 +2410,23 @@ private void OnAutoEvoPredictionComplete(PendingAutoEvoPrediction run)
...
+            {
+                totalEnergyLabelText.Text =
+                    @$"{StringUtils.ThreeDigitFormat(
```
hhyyrylainen, line 2426: Why is the `@` symbol here? Is it required.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2410,10 +2410,23 @@ private void OnAutoEvoPredictionComplete(PendingAutoEvoPrediction run)
...
+                        newTotalPatchEnergyGathered.TotalEnergyGathered)} ({newPopulation})";
+            }
+        }
```
hhyyrylainen, line 2429: If this if doesn't succeed, then I think an error message should be printed as otherwise the wrong number is displayed in the GUI without further information as to what went wrong.

hhyyrylainen (changes requested):

src/microbe_stage/editor/CellEditorComponent.cs
CI09: I had to somehow split the line, @ is used for strings in such case if I am familiar with it enough (Correct me if I'm wrong).

hhyyrylainen: I kind of assumed that after writing the comment. I think it would be cleaner to have a temporary variable to hold the result of ThreeDigitFormat call, that way the string interpolation should then fit on a single line.

CI09: `string formatted = StringUtils.ThreeDigitFormat(newTotalPatchEnergyGathered.TotalEnergyGathered);`
Can I leave the name `formatted` as is? it barely fits and I can fit just up to 6 characters more there.

src/microbe_stage/editor/CellEditorComponent.cs
CI09: Does the red text that pops up saying something like "prediction failed" has to display?

Also I noticed the icon is somewhat small when prediction fails:
image

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: I'd use `var formatted` here. `formatted` is not the most descriptive variable name but as it is used in such a small scope I think it is fine to compromise on it.

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: > Also I noticed the icon is somewhat small when prediction fails:

I think it's been like that since it was added. It's visible for such a small amount of time / rarely in case of failure, that I haven't bothered fixing it as it might be a bit tricky to fix it.

CI09: I meant should I make it display when game can't get the gathered energy. I don't know if it displays when auto-evo fails itself, but I don't think I messed up anything to make it not do so.

hhyyrylainen: That would be pretty nice. Otherwise we might get a really confusing bug report in the future about the numbers not making sense (and only a log file being able to clarify the situation).

src/microbe_stage/editor/CellEditorComponent.cs
CI09: Theoretically I made it work, but I don't know if it looks pretty enough:
image

hhyyrylainen: >  Edit: also predicted global population is lower than predicted local population, is that normal?

It *might* be a case of population number adjustment being kinder on the player, so the population result has that lower strength applied, but the per patch result might not.

It does look a bit busy so maybe just showing the energy numbers in the best and worst patches would be better? If I remember right that was my original idea I had about this, but I didn't recall that until just now.

hhyyrylainen: I guess they could be kept for future reference if they are easy to comment out (you should also comment that they used to be displayed but are now kept just for the potential future use they might have).

CI09:
I was playtesting and when moved away from my origin patch, entered the editor and went into editor tab game crashed with error:
https://pastebin.com/VzA4bDnY](https://pastebin.com/VzA4bDnY)
(sorry if it is poorly visible, lines got wrapped)

hhyyrylainen:
Like the error says, there's is a call to LINQ OrderBy that is told to sort things it doesn't have a clue in `CellEditorComponent.OnAutoEvoPredictionComplete`. The solution is to 1) verify that the list to be sorted contains expected data and not anything weird 2) writing a custom comparer that correctly compares the object types and 3) passing that as the sorter object to the OrderBy (or Sort call if done directly on the data structure). That should switch to using the custom sorter and the exception `Failed to compare two elements in the array. ---> System.ArgumentException: At least one object must implement IComparable.`. Or I guess the other solution is to make the custom types in the collection implement the IComparable interface, this is also done in Thrive in a few places.

CI09:
I didn't fix it yet, do I am gonna mark it as draft (and if anyone is willing, to, they can try to fix it instead; I might do it later).

CI09:
Someone might have to do it for me because I'm just too newbie to do it myself🤔 But take as much time as you want.

CI09:
This is now fixed, there is still a problem which causes indicator to show NaN when moved to a new patch though, but it theoretically could be merged.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2426,31 +2435,65 @@ private void OnAutoEvoPredictionComplete(PendingAutoEvoPrediction run)
...
+            totalEnergyLabel.Value = newTotalPatchEnergyGathered.TotalEnergyGathered;
+
+            var formatted = StringUtils.ThreeDigitFormat(newTotalPatchEnergyGathered.TotalEnergyGathered);
+            totalEnergyLabelText.Text =
+                $"{formatted} ({newPopulation})";
+        }
+        else
+        {
+            // TODO: Make indicator not show NaN when moved to a new patch
+            GD.PrintErr("Failed to get total patch energy gathered");
```
hhyyrylainen, line 2456: Reading this logic, I now realize that there is a problem here: the total energy needs to be the total energy gathered, not any per-patch energy. I'll modify this to calculate the total energy gathered and show it properly.

hhyyrylainen (approved):
I fixed the small issues with the text left I saw, and the bigger problem of the comparing to previous data not working and showing only the current patch energy info. This should be now all good to go but I'll give @CI09 a chance to check that I didn't accidentally break anything with the rebase/squash or my intended changes.

CI09:
I changed patch and it didn't show me NaN, that's good; game also didn't crash during second editor session, but weird thing was that Energy Gathered tooltip again capitalized Energy, like before:
> image

even though in changed files on github it is not (line 2452 in en.po file I think). May be something on my side - I'm still kinda tripping with git.

hhyyrylainen:
> even though in changed files on github it is not (line 2452 in en.po file I think). May be something on my side - I'm still kinda tripping with git.

I probably broke it somehow in the rebase. I fixed it at least like 3 times, but apparently that wasn't enough.

hhyyrylainen:
Actually, I'm not seeing that. As I did a rebase, you need to force pull to get the latest changes.

CI09:
Okay, I checked out remote branch instead of local branch and everything looks fine to me.
(Maybe I had to checkout remote branch due to me already fetching (without force) and trying to force pull later, I don't know.)

PR #4800, opened 2024-01-14
Vary Oxygen Throughout Patches

Deus-Codes:

Simple tweak of biomes.json file changing oxygen concentration amounts across patches. Opening this up to playtesting and feedback. Note that numbers aren't necessarily final, I'm just trying things out, and would like to show an accessible history of various alternatives.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


simulation_parameters/microbe_stage/biomes.json
```diff
@@ -1899,7 +1899,7 @@
         "oxygen": {
           "Amount": 0,
           "Density": 0,
-          "Ambient": 0.21
+          "Ambient": 0.22
```
hhyyrylainen, line 1902: Shouldn't the max be kept as 21 as that's what's on Earth right now?

CI09:
I playtested it a bit and here's my conclusion
Note that I had LAWK off, but I tried to avoid using non-LAWK organelles
I encourage others to playtest this, especially LAWK-only and patches furhter away from origin patches.
## Hydrothermal Vents - Generation 0
_O2 - 2%_
_CO2 - 9%_
In the first editor session besides cytoplasm and hydrogen sulfide proteins (and thermosynthase) there weren't really much other options; metabolosomes didn't find purpose here.
Relying on glucose wouldn't probably be good for long so I was maxing hydrogen sulfide proteins. Seems like if your organism consists of hydrogen sulfide proteins only, its ATP balance is 1:1 which doesn't allow for adding new organelles pretty much. Although with additional cytoplasm, spare ATP production allows for more variety of organelles, but still not much.
**I'd say it gives more motivation to leave the cozy vents and move towards other patches.**
## Sea Floor - Generation 5
_O2 - 16%_
_CO2 - 9%_
Glucose amount is already reduced due to other species inhabiting this patch. I reseted my cell to 2 hex cytoplasm to be less adapted to hydrogen sulfide (which isn't present here now). Finding glucose wasn't really tough, but I found my place at the bottom of food chain.
Nitrogen fixing proteins have been unlock and thus I added them along with metabolosomes which now are more useful.
I got devoured 2 times and by that, got patch extinction. I wanted to load save but It gave me error `Error Loading - An exception happened while loading the save data exception: Newtonsoft.Json.JsonSerializationException: Error getting value from 'Value' on 'System.Lazy'` (I don't think it is related to this change though.)

## Tidepool - Generation 0 (new save)
_O2 - 5%_
_CO2 - 9%_
By adding metabolosomes, my ATP consumption was getting bigger than production, cytoplasm kept ATP production/consumption ratio the same. I could add some organelles in this case (also when combining metabolosomes and cytoplasm), but spamming metabolosomes definetly is not the best way.
**Feels limiting, in a good way. moving out to another patch feels desireable.**

I again crashed, via Windows bug I keep forgetting to report. (when clicking the bottom right button on the end of the task bar, which is used to quickly minimize all apps to show desktop, game just crashes.)
image

# Freebuild
## Abyssopelagic
_O2 - 16%_
_CO2 - 9%_
Metabolosomes work quite good here, I need at least 5 to sustain nucleus, and probably more to have flagella so the cell is not too slow (actually 2-3 per flagellum).
## Bathypelagic
_O2 - 14%_
_CO2 - 9%_
(It has less oxygen than abyssopelagic)
7 metabolosomes for nucleus,
Now 2-3 metabolosomes per flagellum.
## Mesopelagic
_O2 - 18%_
_CO2 - 9%_
5 metabolosomes for nucleus,
1-2 metabolosomes per flagellum (3 metabolosomes per 2 flagella)
## Ice shelf
_O2 - 22%_
_CO2 - 9%_
4 metabolosomes to sustain nucleus.
1-2 metabolosomes per flegellum but more likely just 1 (depends on spare ATP)
## Underwater cave
_O2 - 15%_
_CO2 - 9%_
6 metabolosomes to sustain nucleus.
2-3 metabolosomes per flegellum.

Sorry if the post is too long, I might got carried away a bit... :)

hhyyrylainen:
> I got devoured 2 times and by that, got patch extinction. I wanted to load save but It gave me error Error Loading - An exception happened while loading the save data exception: Newtonsoft.Json.JsonSerializationException: Error getting value from 'Value' on 'System.Lazy'

Please give me the save (or JSON error debug log). Seems like a case of unintended variable getting caught up in the save system (maybe after the organelle unlocks feature) that shouldn't be saved.

CI09:
https://pastebin.com/5L197DJz

hhyyrylainen:
That's the exception, I need `json_debug.txt` from the Thrive logs folder (it should be created on error as long as you have JSON debug mode set to "on error"). The exception unhelpfully doesn't list the path to the problematic property, that's only in the JSON debug log.

CI09:
https://pastebin.com/bierKbSG

hhyyrylainen:
`'SavedProperties.GameWorld.StatisticsTracker.AllStatistics.IsValueCreated'` is the problematic property, and sure enough it is something that should not be saved under any circumstances. I'll make a PR with a fix for this tomorrow. Seems like it is entirely unused so it can be deleted entirely.

hhyyrylainen:
There's now a PR that should fix that: https://github.com/Revolutionary-Games/Thrive/pull/4802

simulation_parameters/microbe_stage/biomes.json
Deus-Codes: From what I've read, the 21% represents more of an "average" throughout the entire atmosphere and not necessarily a max capacity. So local concentrations of atmosphere within patches can have more oxygen than 21% in them.

Deus-Codes:
Wanted to note that I am still working on this, I'm just doing some research on more accurate numbers and have been busy.

hhyyrylainen:
I think unless someone objects this should be merged so that people stop getting ahead of themselves talking about other compound variations before we even have this.

hhyyrylainen (approved):

PR #4801, opened 2024-01-14
Implement sprinting mechanics

ender-titan1:

This PR implements sprinting and cell strain to the game.

To do:
- [ ] Balance
- [x] Red flash when strain is high
- [x] Strain Bar (for accessibility)
- [x] Strain Bar visibility options

Maybe:
- [ ] Tutorial
- [x] Damage cell when ATP is close to zero 

**Related Issues**

https://forum.revolutionarygamesstudio.com/t/organism-stamina-mechanics/757/7

**Progress Checklist**

      break existing features:
      members reviewing this PR)


ARecurringProblem (approved):
If we're settled on the basic implementation idea then I think it's good to get this into a BOTD soon so that we can get player feedback on how it feels.

src/general/OptionsMenu.cs
```diff
@@ -501,6 +506,12 @@ public override void _Ready()
     {
         ResolveNodeReferences(true);
 
+        foreach (var mode in typeof(Settings.StrainBarVisibility).GetEnumValues().Cast<Settings.StrainBarVisibility>())
```
hhyyrylainen, line 509: For the options menu the usual way to setup these dropdowns is to set them in the Godot editor and not load them manually like this.

src/engine/Settings.cs
```diff
@@ -489,6 +505,11 @@ private Settings()
         Constants.CONTROLLER_DEFAULT_DEADZONE,
     });
 
+    // Gameplay and Accessibility settings
```
hhyyrylainen, line 508: Isn't this in the same section in the menu as the display abilities bar?

simulation_parameters/Constants.cs
```diff
@@ -183,12 +183,38 @@ public static class Constants
     /// </remarks>
     public const float BASE_MOVEMENT_ATP_COST = 1.0f;
 
+    public const float BASE_MOVEMENT_FORCE = 900.0f;
+
+    public const float MAX_STRAIN_PER_CELL = 100.0f;
+
+    public const float PASSIVE_STRAIN_DECREASE_PER_UPDATE = 0.5f;
```
hhyyrylainen, line 190: Rather than per update rate, any change rates should be scaled to be per second.

src/microbe_stage/MicrobeHUD.tscn
```diff
@@ -1943,10 +2102,29 @@ __meta__ = {
...
 }
 
+[node name="Sprint" parent="ScrollContainer/HotBar" instance=ExtResource( 65 )]
+visible = false
+anchor_right = 0.0
+anchor_bottom = 0.0
+margin_left = 252.0
+margin_right = 298.0
+margin_bottom = 51.0
+hint_tooltip = "TOGGLE_ENGULF_TOOLTIP"
```
hhyyrylainen, line 2112: Wrong tooltip text is used here.

src/microbe_stage/components/StrainAffected.cs
```diff
@@ -0,0 +1,39 @@
...
+    [JSONDynamicTypeAllowed]
+    public struct StrainAffected
+    {
+        /// <summary>
+        ///   True when sprinting or when strain is supposed to be otherwise generated
+        /// </summary>
+        public bool IsUnderStrain;
+
+        /// <summary>
+        ///   The current amount of strain, the maximum is calculated based on the hex size
```
hhyyrylainen, line 17: Where is the maximum calculated as per this comment?

src/microbe_stage/systems/MicrobeMovementSystem.cs
```diff
@@ -194,6 +206,16 @@ private static float CalculateRotationSpeed(in Entity entity, ref OrganelleConta
...
                 (cellProperties.MembraneRigidity * Constants.MEMBRANE_RIGIDITY_BASE_MOBILITY_MODIFIER);
 
+            if (control.Sprinting && canSprint)
+            {
+                force *= Constants.SPRINTING_FORCE_MULTIPLIER;
+                strain.IsUnderStrain = true;
+            }
+            else
+            {
+                strain.IsUnderStrain = false;
```
hhyyrylainen, line 216: Shouldn't strain be unset when no movement keys are set? In the if checking `control.MovementDirection == Vector3.Zero`.

src/microbe_stage/systems/OsmoregulationAndHealingSystem.cs
```diff
@@ -141,7 +141,7 @@ protected override void Update(float delta, in Entity entity)
         /// </summary>
         private void ApplyATPDamage(CompoundBag compounds, ref Health health, ref CellProperties cellProperties)
         {
-            if (compounds.GetCompoundAmount(atp) > 0)
+            if (compounds.GetCompoundAmount(atp) > 0.1f)
```
hhyyrylainen, line 144: Why is this needed? I think this may have massive implications on taking ATP damage when trying to move.

src/microbe_stage/systems/StrainSystem.cs
```diff
@@ -0,0 +1,85 @@
...
+    using DefaultEcs.Threading;
+    using Godot;
+    using World = DefaultEcs.World;
+
+    [With(typeof(CompoundStorage))]
+    [With(typeof(OrganelleContainer))]
+    [With(typeof(MicrobeControl))]
+    [With(typeof(StrainAffected))]
+    [RunsAfter(typeof(PhysicsBodyDisablingSystem))]
+    public sealed class StrainSystem : AEntitySetSystem<float>
```
hhyyrylainen, line 15: Why doesn't this system rely on `IsUnderStrain`? It seems the purpose of that variable is so that the movement system sets that and then that greatly simplifies this system by not needing to act on so much microbe-specific information.

src/microbe_stage/systems/StrainSystem.cs
```diff
@@ -0,0 +1,85 @@
...
+    using DefaultEcs.System;
+    using DefaultEcs.Threading;
+    using Godot;
+    using World = DefaultEcs.World;
+
+    [With(typeof(CompoundStorage))]
+    [With(typeof(OrganelleContainer))]
+    [With(typeof(MicrobeControl))]
+    [With(typeof(StrainAffected))]
+    [RunsAfter(typeof(PhysicsBodyDisablingSystem))]
```
hhyyrylainen, line 14: Why does this run after the body disabling system?

src/microbe_stage/systems/StrainSystem.cs
```diff
@@ -0,0 +1,85 @@
...
+    using Components;
+    using DefaultEcs;
+    using DefaultEcs.System;
+    using DefaultEcs.Threading;
+    using Godot;
+    using World = DefaultEcs.World;
+
+    [With(typeof(CompoundStorage))]
+    [With(typeof(OrganelleContainer))]
+    [With(typeof(MicrobeControl))]
```
hhyyrylainen, line 12: As far as I can tell these two components are just read from and as such should have the reads from component attributes here to mark that to make multithreading more effective related to this system.

src/microbe_stage/systems/StrainSystem.cs
```diff
@@ -0,0 +1,85 @@
...
+                {
+                    ReduceStrain(ref strain);
+                }
+                else
+                {
+                    strain.StrainDecreaseCooldown -= delta;
+                    ReduceStrain(ref strain, Constants.PASSIVE_STRAIN_DECREASE_PRE_COOLDOWN_DIVISOR);
+                }
+
+                if (strain.CurrentStrain < 0)
```
hhyyrylainen, line 59: Why not put this check and clamp inside the ReduceStrain method?

src/microbe_stage/systems/StrainSystem.cs
ender-titan1: The logic behind this was that since this is supposed to run before `MicrobeMovementSystem`, and that system ran after `PhysicsBodyDisablingSystem`, I moved the `RunsAfterAttribute` to `StrainSystem`, but looking at it now it might not be correct.

src/microbe_stage/systems/OsmoregulationAndHealingSystem.cs
ender-titan1: This was changed as ATP never reached a low enough level for it to do damage. The bar would look empty but no damage would be done. Should've probably set it to something lower though.

src/microbe_stage/systems/StrainSystem.cs
hhyyrylainen: Then set this directly to run after the movement system. The threaded run generator can handle recursive dependencies so A -> B -> C runs after attributes will ensure that A will only run after C.

src/microbe_stage/systems/OsmoregulationAndHealingSystem.cs
hhyyrylainen: Adjusting when the ATP is taken would be a better choice. That's what's been done in the past to ensure cells can be damaged (for example making this run before whatever system applies ATP damage, but after the process system).

ender-titan1: I ended up changing this to `0.05f`, as I couldn't get this to work otherwise. I think it's sensible to set it to that, as the difference at that point is barely noticeable and I don't think it would have any major implications.

hhyyrylainen: Could you add a comment before this if to explain the number? Or actually I think it'd be pretty nice to make it a constant in Constants and add a summary comment explaining the reason for it there.

ender-titan1:
I think mechanically, this feature is mostly ready. But since it still needs some work maybe it should be classed as an "experimental feature"?

To recap, what is needed is balance, AI implementation, a tutorial and maybe some graphics improvements.

hhyyrylainen:
I suppose without those still undone parts classifying this as an experimental feature would be an option to merge this soon.

locale/fi.po
```diff
@@ -5205,6 +5227,10 @@ msgstr ""
 msgid "TOGGLE_PAUSE"
 msgstr "Pysäytä"
 
+#, fuzzy
```
hhyyrylainen, line 5230: Could you redo the translations from scratch for this PR? I see this and the en.po file have a bunch of old translations added that will never be useful as the translation keys have never existed on the master branch.

src/microbe_stage/systems/StrainSystem.cs
```diff
@@ -0,0 +1,61 @@
+﻿namespace Systems
+{
+    using Components;
+    using DefaultEcs;
+    using DefaultEcs.System;
+    using DefaultEcs.Threading;
+    using Godot;
+    using World = DefaultEcs.World;
+
+    [With(typeof(CompoundStorage))]
```
hhyyrylainen, line 10: This system seems to only use StrainAffected and OrganelleContainer components now, so the attributes here need to be fixed.

src/microbe_stage/components/StrainAffected.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// <summary>
+    ///   Contains variables related to strain
+    /// </summary>
+    [JSONDynamicTypeAllowed]
+    public struct StrainAffected
+    {
+        /// <summary>
+        ///   True when sprinting or when strain is supposed to be otherwise generated
+        /// </summary>
+        public bool IsUnderStrain;
```
hhyyrylainen, line 14: Could you reorder these variables for memory efficiency? i.e. biggest variables first and this bool last as this takes just one byte so this being first causes 3 extra padding bytes to be added (in this case this doesn't matter but if there was a second bool in this struct then this would consume extra memory unnecessarily so it is good practice).

src/microbe_stage/components/StrainAffected.cs
```diff
@@ -0,0 +1,34 @@
...
+        ///   The amount of time the player has to wait before <see cref="CurrentStrain"/> sarts to fall
+        /// </summary>
+        public float StrainDecreaseCooldown;
+    }
+
+    public static class StrainAffectedHelpers
+    {
+        public static float CalculateStrainFraction(this ref StrainAffected affected)
+        {
+            return Math.Max(0, affected.CurrentStrain - Constants.CANCELED_STRAIN) / Constants.MAX_STRAIN_PER_CELL;
```
hhyyrylainen, line 31: Is the max strain per cell variable well named? I ask because for cell colonies they contain many cells. Should it perhaps be named `MAX_STRAIN_PER_ENTITY`?

src/microbe_stage/PlayerMicrobeInput.cs
```diff
@@ -297,6 +297,40 @@ public void CloseSignalingCommandsMenu()
             stage.HUD.ApplySignalCommand(command, stage.Player);
     }
 
+    [RunOnKeyDown("g_sprint")]
+    public bool StartSprint()
+    {
+        if (!stage.HasPlayer || !stage.GameWorld.WorldSettings.ExperimentalFeatures)
+            return true;
```
hhyyrylainen, line 304: Shouldn't this *not* consume input, this doesn't do anything except cause other things to not be able to react to this keypress?

src/microbe_stage/PlayerMicrobeInput.cs
```diff
@@ -297,6 +297,40 @@ public void CloseSignalingCommandsMenu()
...
+        control.Sprinting = true;
+
+        // We need to not consume the input, otherwise the key up for this will not run
+        return false;
+    }
+
+    [RunOnKeyUp("g_sprint")]
+    public void EndSprint()
+    {
+        if (!stage.HasPlayer || !stage.GameWorld.WorldSettings.ExperimentalFeatures)
```
hhyyrylainen, line 316: I don't think this needs to check experimental features, sprinting should always be able to end even if experimental features are not on.

src/microbe_stage/PlayerMicrobeInput.cs
```diff
@@ -297,6 +297,40 @@ public void CloseSignalingCommandsMenu()
...
+            return;
+
+        ref var control = ref stage.Player.Get<MicrobeControl>();
+
+        control.Sprinting = false;
+    }
+
+    public void ToggleSprint()
+    {
+        if (!stage.HasPlayer)
```
hhyyrylainen, line 326: Instead checking here for experimental features would be more important I think.

src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -125,6 +125,9 @@ public abstract class CreatureStageHUDBase<TStage> : HUDWithPausing, ICreatureSt
     [Export]
     public NodePath HotBarPath = null!;
 
+    [Export]
+    public NodePath SprintHotkeyPath = null!;
```
hhyyrylainen, line 129: It seems the multicellular HUD scene is not setup for strain, it should be setup to ensure the game doesn't crash when getting to the prototypes.

src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -1052,14 +1057,17 @@ protected void UpdatePanelSizing(float delta)
     protected abstract void UpdateAbilitiesHotBar();
 
     protected void UpdateBaseAbilitiesBar(bool showEngulf, bool showToxin, bool showSlime,
-        bool showingSignaling, bool engulfOn, bool showEject)
+        bool showingSignaling, bool engulfOn, bool sprintOn, bool showEject)
     {
+        sprintHotkey.Visible = stage?.GameWorld.WorldSettings.ExperimentalFeatures == true;
```
hhyyrylainen, line 1062: Instead of allowing stage to be null here, I think a better thing to do would be to just flat out:
```
if (stage == null)
    throw new InvalidOperationException("Stage must be set before abilities update");
```

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -30,10 +30,22 @@ public class MicrobeHUD : CreatureStageHUDBase<MicrobeStage>
...
     public NodePath UnbindAllHotkeyPath = null!;
 
+    [Export]
+    public NodePath StrainBarPanelPath = null!;
+
+    [Export]
+    public NodePath StrainBarPath = null!;
+
+    [Export]
+    public NodePath StrainBarFadeAnimationPlayerPath = null!;
```
hhyyrylainen, line 40: I think this should maybe be moved to the base HUD as shouldn't big creatures be eventually be able to sprint as well? So it makes sense to put there rather than having to copy-paste all this stuff to the multicellular HUD.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -542,12 +566,58 @@ protected override void Dispose(bool disposing)
...
 
         base.Dispose(disposing);
     }
 
+    private void UpdateStrain(Entity player)
+    {
+        if (!stage!.GameWorld.WorldSettings.ExperimentalFeatures)
+        {
+            strainBar.Visible = false;
+        }
```
hhyyrylainen, line 583: Shouldn't there be an early return here?

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -542,12 +566,58 @@ protected override void Dispose(bool disposing)
...
+
+        var strainFraction = player.Get<StrainAffected>().CalculateStrainFraction();
```
hhyyrylainen, line 585: This should check if the component exists and if not print an error *once* and then do nothing. This will ensure compatibility with older saves and rare bugs if the component is somehow missing. This isn't a critical feature to display the info so it missing shouldn't crash the game.

project.godot
```diff
@@ -476,6 +476,11 @@ g_eject_engulfed={
 "events": [ Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":0,"alt":false,"shift":false,"control":false,"meta":false,"command":false,"pressed":false,"scancode":72,"physical_scancode":0,"unicode":0,"echo":false,"script":null)
  ]
 }
+g_sprint={
+"deadzone": 0.5,
+"events": [ Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":0,"alt":false,"shift":false,"control":false,"meta":false,"command":false,"pressed":false,"scancode":16777237,"physical_scancode":0,"unicode":0,"echo":false,"script":null)
```
hhyyrylainen, line 481: What key is this? Would it be possible to make sprint be shift, like it is in most games?

src/microbe_stage/systems/StrainSystem.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            ref var strain = ref entity.Get<StrainAffected>();
+            ref var organelles = ref entity.Get<OrganelleContainer>();
+
+            if (strain.IsUnderStrain)
+            {
+                var strainIncrease = Constants.SPRINTING_STRAIN_INCREASE_PER_SECOND * delta;
+                strainIncrease += organelles.HexCount * Constants.SPRINTING_STRAIN_INCREASE_PER_HEX;
+
+                if (strain.CurrentStrain > Constants.MAX_STRAIN_PER_CELL)
```
hhyyrylainen, line 32: Shouldn't this check be below line 35 so that the new value is correctly clamped?

src/microbe_stage/systems/MicrobeMovementSystem.cs
```diff
@@ -247,6 +273,12 @@ private static float CalculateRotationSpeed(in Entity entity, ref OrganelleConta
             return position.Rotation.Xform(movementVector);
         }
 
+        private float GetStrainMultiplier(ref StrainAffected strain)
```
hhyyrylainen, line 276:
```suggestion
        private float GetStrainAtpMultiplier(ref StrainAffected strain)
```

src/microbe_stage/systems/MicrobeMovementSystem.cs
```diff
@@ -140,8 +144,19 @@ private static float CalculateRotationSpeed(in Entity entity, ref OrganelleConta
...
+            ref var strain = ref entity.Get<StrainAffected>();
+
             if (control.MovementDirection == Vector3.Zero)
             {
+                // Make sure the microbe is not under strain when not moving
+                strain.IsUnderStrain = false;
+
+                // Remove ATP due to strain even if not moving
+                // This is calculated similarily to the regular movement cost for consistency
+                var strainMultiplier = GetStrainMultiplier(ref strain);
```
hhyyrylainen, line 156: Couldn't this line be straight below the `ref var strain` line as this value seems to be always needed? That way this doesn't need to have the code to calculate this in two places.

hhyyrylainen (changes requested):
Here's a full code review. I don't see any major issues anymore (unless some pop up when I playtest this), but there's quite many small details that can be tweaked to make this feature more maintainable and improve code clarity, and maybe fix one small bug.

locale/fi.po
hhyyrylainen: Also I see this has translation file conflicts so anyway redoing the translations shouldn't be harder than fixing those conflicts manually.

project.godot
ender-titan1: This is shift. From my testing it seems to work.

hhyyrylainen: Alright, that's great then :+1:

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -311,6 +317,18 @@ protected override void UpdateHealth(float delta)
         hpLabel.HintTooltip = hpText;
     }
 
+    protected override float? ReadPlayerStrainFraction()
+    {
+        if (stage!.Player.Has<StrainAffected>() && !playerMissingStrainAffected)
```
CI09, line 322: I think first condition should be with ! at the beggining

src/microbe_stage/components/MicrobeControl.cs
```diff
@@ -53,6 +53,11 @@ public struct MicrobeControl
         /// </summary>
         public bool SlowedBySlime;
 
+        /// <summary>
+        ///   Whether this microbe is currenyly sprinting
```
CI09, line 57: Typo

CI09 (changes requested):

Deus-Codes:
Limited testing due to battery issues with my laptop. I am not sure how it scales with greater mass, but sprinting felt pretty good in the few generations I played. Great job implementing a feature that makes movement atleast a bit more engaging than pointing your mouse in a direction and pressing W.

One thing though: in the original concept, sprinting was attached to the flagella as an ability rather than a default ability, which synergizes with the flagella providing upgrades changing the behavior of sprinting. Was it an intentional decision to have it be available from the get go?

ender-titan1:
Oh I wasn't expecting this to be reviewed. Slight heads up that this will probably have to be changed into a new Godot 4 branch. I don't recall there being a particular reason for why the sprinting is enabled by default, iirc it was just an arbitrary decision that's pretty easy to change if needed. 

However I **did** intentionally leave out flagella upgrades for the simple reason that I found them too hard to implement and decided to do them down the line when the basic system was done.

hhyyrylainen:
Will this be continued soon? It's not going to be too much longer now that the 0.6.x roadmap items are all planned to be completed and this is one of them.

PR #4814, opened 2024-01-18
Unlimited growth speed cheat

CI09:

This PR adds cheat that disables growth speed limit. It doesn't take now that long to reproduce organism with infinite compounds.

image

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09:
The commit is theoretically on this branch, at least on my local:
image

hhyyrylainen:
That's not the latest commit, this is:
```
commit 010bfe717788b9a65f5f2039027a6c793c458d05 (HEAD -> master, origin/master)
Author: Henri Hyyryläinen
Date:   Fri Jan 19 15:24:07 2024 +0200

    Fixed engulfing target not taking colonies into account (#4813)

```

CI09:
It's ready now.

ender-titan1:
This is a bit of a nitpick, but I think the cheat should be called more like "Disable growth speed limit" instead.

CI09:
It would be kinda longer than other GUI elements for cheat menu, but doable.

hhyyrylainen:
What's the current text? What about my suggestion of "Unlimit Growth Speed" it has still quite few words, but fully describes what the cheat does.

ender-titan1:
Again, this is me being nitpicky, but unlimit isn't a real word. Go ahead and use it though, it gets the meaning across. It just doesn't sound right to me.

hhyyrylainen:
Well, we can make it a real word. Also someone already in 2018 came up with the word: https://en.wiktionary.org/wiki/unlimit

ARecurringProblem (approved):
Working for me, really glad this is implemented!

src/general/DifficultyPreset.cs
```diff
@@ -52,7 +52,7 @@ public class DifficultyPreset : IDifficulty, IRegistryType
     public bool PassiveReproduction { get; private set; }
 
     [JsonProperty]
-    public bool LimitGrowthRate { get; private set; }
+    public bool LimitGrowthRate { get; set; }
```
hhyyrylainen, line 55: Presets are registry data *they are not* allowed to be modified during runtime, otherwise starting new games will use incorrect data.

hhyyrylainen (approved):
I refactored the way the setting is applied to make sure it isn't persistent in the public data of the difficulty settings objects that are loaded as a registry type.

PR #4816, opened 2024-01-21
Add eject all engulfed objects action

### LINKED ISSUE #3583: Add a button to eject all engulfed things
so that the player is not stuck getting no glucose. Also would give attentive players a chance to eject a toxic cell they engulfed.

Needs an icon for the hot bar

CI09:

This PR add eject engulfed action to eject all objects currently engulfed by player's cell.
Also fixes crash where couldn't get best/worst patches from list.

**Related Issues**

closes #3583

**Progress Checklist**

      break existing features:
      members reviewing this PR)


project.godot
```diff
@@ -193,6 +193,11 @@ g_toggle_engulf={
 "events": [ Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":0,"alt":false,"shift":false,"control":false,"meta":false,"command":false,"pressed":false,"scancode":71,"physical_scancode":0,"unicode":0,"echo":false,"script":null)
  ]
 }
+g_eject_engulfed={
```
hhyyrylainen, line 196: The new key needs to be added to `input_options.json` otherwise it is not reboundable and doesn't show up in the game settings.

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen, line 1: I made a separate PR about these changes: https://github.com/Revolutionary-Games/Thrive/pull/4819

ARecurringProblem (commented):
Looks like this is working fine as far as gameplay goes (also makes for a handy workaround to the issue of some things being undigestible since they can be manually ejected now), but I do occasionally get "Force ejection on engulfer that doesn't have engulfed object list setup is skipping normal eject logic" in the log. Since there's still other potential changes being discussed that should be tested if they're implemented, I'm going to hold off on approving this one until that's all figured out.

hhyyrylainen:
The player can still press the key directly, so guards against doing it when inappropriate would be best avoided. It's probable that the player alive check I suggested adding won't be completely enough.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -432,10 +435,17 @@ protected override void UpdateAbilitiesHotBar()
             engulfing = control.State == MicrobeState.Engulf;
         }
 
+        bool isDigesting = false;
+
+        ref var engulfer = ref stage.Player.Get<Engulfer>();
+
+        if (engulfer.EngulfedObjects != null && engulfer.EngulfedObjects?.Count != 0)
```
hhyyrylainen, line 442:
```suggestion
        if (engulfer.EngulfedObjects is {Count: > 0})
```

I think a single check like this is equivalent.

src/microbe_stage/MicrobeHUD.tscn
```diff
@@ -1955,6 +1957,17 @@ toggle_mode = true
 ActionIcon = ExtResource( 9 )
 ActionName = "g_toggle_engulf"
 
+[node name="EjectEngulfed" parent="ScrollContainer/HotBar" instance=ExtResource( 65 )]
```
hhyyrylainen, line 1960: Did you fix the focus style here already? As engulfment is an actual toggle that is different from this. You should instead look at how the FireToxin button is made to copy a single press action button for this feature.

src/microbe_stage/PlayerMicrobeInput.cs
```diff
@@ -163,6 +163,26 @@ public void ToggleEngulf()
...
 
+    [RunOnKeyDown("g_eject_engulfed")]
+    public void EjectAllEngulfed()
+    {
+        ref var engulfer = ref stage.Player.Get<Engulfer>();
+
+        if (engulfer.EngulfedObjects == null || engulfer.EngulfedObjects?.Count == 0)
+            return;
+
+        if (!stage.HasAlivePlayer)
```
hhyyrylainen, line 174: This check should be *before* the engulfer component is retrieved, otherwise the game might crash if the player object has been destroyed and no longer has an engulfer component.

src/microbe_stage/PlayerMicrobeInput.cs
```diff
@@ -163,6 +163,26 @@ public void ToggleEngulf()
         }
     }
 
+    [RunOnKeyDown("g_eject_engulfed")]
+    public void EjectAllEngulfed()
+    {
+        ref var engulfer = ref stage.Player.Get<Engulfer>();
+
+        if (engulfer.EngulfedObjects == null || engulfer.EngulfedObjects?.Count == 0)
```
hhyyrylainen, line 171: Same shorter check should be usable here as well than what I commented on before.

src/microbe_stage/PlayerMicrobeInput.cs
```diff
@@ -163,6 +163,26 @@ public void ToggleEngulf()
...
+            return;
+
+        if (!stage.HasAlivePlayer)
+            return;
+
+        if (engulfer.EngulfedObjects != null)
```
hhyyrylainen, line 177: There's now a check above against this being null so this check should be unnecessary.

src/microbe_stage/MicrobeHUD.tscn
hhyyrylainen: Doesn't there also need to be this button but as a dummy button in the multicell hud as well to avoid crashing?

CI09: I think it works as non-toggle button would. And about multicellular HUD, yes, I forgot.

hhyyrylainen: Looking at the overridden themes etc. properties in this text format, this looks much more like the toggle button than the toxin fire button. So I was wondering if something was wrong. Maybe you could check those property differences and make doubly sure that this is all like it should be.

CI09: Well, I possibly copy-pasted toggle engulf button and then set it to work more like normal button, but I can inspect it and redo if needed

hhyyrylainen: I think redoing the button by copy pasting the toxin fire button would ensure there's nothing wrong with the new button.

CI09:
Okay, *now* it is ready

hhyyrylainen (approved):
I fixed a few minor details and tested this briefly. I didn't see any error prints now with the code only being allowed to trigger when there is an engulfed object in the player.

Issue #4579, opened 2023-11-24
Create a memory cache for photostudio microbe images


hhyyrylainen:
so that cell billboards at least can share a lot of the generated images without a bunch of extra rendering work happening
A simplified version of: https://github.com/Revolutionary-Games/Thrive/issues/3156

Issue #3156, opened 2022-04-07
Create a cache that is stored on disk for PhotoStudio


hhyyrylainen:
to reduce how often images need to be regenerated.

This is a pretty complex issue:
- Hashes need to be made very good for various photographable objects so that the hashes can be cache keys
- There needs to be a maximum cache size and a periodic cleanup
- Cache size needs to be configurable from options
- Current used cache size needs to be shown and allow manual clear by the player
- Saving images to disk should happen in the background so that it doesn't cause (much) extra lag compared to just generating the images

PR #4817, opened 2024-01-21
Species preview popup when hovering over formatted species name

### LINKED ISSUE #3152: Add some way to view preview images of species in the report tab
now that I got this done: https://github.com/Revolutionary-Games/Thrive/issues/2882 the hard part of this is done, images of species can be generated (well at least cells, might need a tiny bit of extra plumbing to hook things up to make images of Species objects directly). Should be implemented something like hovering over species names or clicking them opening a small hovering panel to open that shows the image (and maybe some other stats of the species).

Oliveriver:

Adds a custom tooltip showing the hex and membrane layouts of cell species when hovering over them, e.g. in the editor or evolutionary tree. In the images below, my cursor is hovering over a species name.

This does have the side effect of adding an underline to species names. It's possible to fix that, but a bit annoying: I'd have to disable MetaUnderlined for CustomRichTextLabel then manually add underline tags to every link we still want underlined in the translation files. The cursor would stop changing state when hovering over links as well.

**Related Issues**

Closes #3152 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/Species.cs
```diff
@@ -119,7 +119,9 @@ public Color GUIColour
     ///   with bolding.
     /// </summary>
     [JsonIgnore]
-    public string FormattedNameBbCode => PlayerSpecies ? $"[b][i]{FormattedName}[/i][/b]" : $"[i]{FormattedName}[/i]";
+    public string FormattedNameBbCode => PlayerSpecies ?
+        $"[url=species:{GUICommon.ConvertToSafeMetaTag(StringCode)}][b][i]{FormattedName}[/i][/b][/url]" :
```
Oliveriver, line 123: Storing the entire serialised class in the meta tag feels really wrong but I don't think there's a much simpler way to do it, since (like the Thriveopedia links) the CustomRichTextLabel could be used anywhere, even outside of games in progress where species are accessible. Please let me know if there's a better way.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -632,4 +644,58 @@ private void OnMetaClicked(object meta)
...
+            if (species is not MicrobeSpecies)
+                return;
+
+            var tooltip = ToolTipManager.Instance
+                .GetToolTipIfExists<SpeciesPreviewTooltip>(species.FormattedName, "speciesPreview");
+            if (tooltip == null)
+            {
+                tooltip = speciesPreviewTooltipScene.Instance<SpeciesPreviewTooltip>();
+                tooltip.Name = species.FormattedName;
+                ToolTipManager.Instance.AddToolTip(tooltip, "speciesPreview");
```
Oliveriver, line 667: I had to fudge the tooltip system due to the way Godot's meta tags work, since I needed to connect from specific meta tags instead of the whole control, and connect unusual signals.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -632,4 +644,58 @@ private void OnMetaClicked(object meta)
...
+
+            if (species is not MicrobeSpecies)
+                return;
+
+            var tooltip = ToolTipManager.Instance
+                .GetToolTipIfExists<SpeciesPreviewTooltip>(species.FormattedName, "speciesPreview");
+            if (ToolTipManager.Instance.MainToolTip == tooltip)
+            {
+                ToolTipManager.Instance.MainToolTip = null;
+                ToolTipManager.Instance.Display = false;
```
CI09, line 697: Very possible I am getting something wrong, but It's the same as `DisplayName` here.

CI09 (commented):
Looks good for me, but I'm kinda newbie so I'd recommend not relying on my "good" too much.

src/general/Species.cs
hhyyrylainen: This can be a ton of data that the bbcode parser needs to get through each time it is updated. I think a better solution would be to just have the species internal id here and maybe have the thriveopedia mediate the request for the species data? Each species should have a Thriveopedia page about it in the future so that kind of design would make sense to make it a requirement for the thriveopedia to have info on all species data.

src/microbe_stage/editor/tooltips/SpeciesPreviewTooltip.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        if (speciesPreview == null || hexesPreview == null)
+            return;
+
+        speciesPreview.PreviewSpecies = PreviewSpecies;
+
+        if (PreviewSpecies is MicrobeSpecies microbeSpecies)
+        {
+            hexesPreview.PreviewSpecies = microbeSpecies;
+            DisplayName = PreviewSpecies.FormattedName;
```
hhyyrylainen, line 81: `FormattedName` exists in the base `Species` class. Why is this just assigned here inside the if condition?

src/gui_common/GUICommon.cs
```diff
@@ -111,6 +111,32 @@ public static void MarkInputAsValid(LineEdit control)
         control.Set("custom_colors/font_color", new Color(1, 1, 1));
     }
 
+    /// <summary>
+    ///   Makes a string safe to store as a meta tag in rich text.
+    /// </summary>
+    public static string ConvertToSafeMetaTag(string tagContent)
```
hhyyrylainen, line 117: I think this and the next method would make more sense to be in `StringUtils` class than here.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -73,12 +77,20 @@ public enum ThriveBbCode
...
 
+    [Export]
+    public bool EnableTooltipsForMetaTags { get; set; } = true;
+
     public override void _Ready()
     {
         // Make sure bbcode is enabled
         BbcodeEnabled = true;
 
+        speciesPreviewTooltipScene = GD.Load<PackedScene>(
```
hhyyrylainen, line 88: There can be a ton of rich text labels. I think it'd be better to have this be a lazy static property or a property on tooltip manager to keep a reference to this scene and not need to attempt to load it for each created text control.

src/gui_common/CustomRichTextLabel.cs
hhyyrylainen: There should be a remove tooltip call to go with the add tooltip call because otherwise the tooltip internal data is never released.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -632,4 +644,58 @@ private void OnMetaClicked(object meta)
...
     }
+
+    private void OnMetaHoverStarted(object meta)
+    {
+        if (!EnableTooltipsForMetaTags || speciesPreviewTooltipScene == null)
+            return;
+
+        if (meta is string metaString && metaString.StartsWith("species", StringComparison.Ordinal))
+        {
+            var speciesCode = GUICommon.ConvertFromSafeMetaTag(metaString.Split("species:")[1]);
```
hhyyrylainen, line 655: This split could be avoided here by just getting a substring that skips the "species:" amount of characters. I think that would be a slightly less garbage generating / faster operation.

hhyyrylainen:
My only real major concern how all of the species data is just dumped into the meta tag. That can end up being an absolute massive amount of data for example if a player makes a 3000 organelle behemoth. So some out of band way to save the species data is needed.

Also the way the tooltips hang around and presumably keep the full images in memory of the species they are for seem like it will make the game memory usage just go up and up. Maybe using a single tooltip instance would be better to just switch out the species it shows? Though that might be a bit slower until one quite long time ago specified background caching system is done: https://github.com/Revolutionary-Games/Thrive/issues/4579 (and also: https://github.com/Revolutionary-Games/Thrive/issues/3156)

hhyyrylainen:
I can probably give that some experimenting with how it would make sense to try to make accessing current game species data this week. In case you don't have good ideas or could do with some more eyes on a potential design here.

Oliveriver:
> I can probably give that some experimenting with how it would make sense to try to make accessing current game species data this week. In case you don't have good ideas or could do with some more eyes on a potential design here.

Sorry, I didn't have time to work on anything the rest of the week. If you want to have a go then please do, but otherwise I might just abandon this PR as it sounds like an optimal solution is dependent on substantial Thriveopedia changes.

hhyyrylainen:
I can pick this up next week (probably as I plan on finishing my multithreading improvements). My idea is basically just that each Thriveopedia instance will report itself to the thriveopedia manager when it becomes active, and offer to provide current game species data that way. Should be only about 20-ish lines of code, and allows the future direction of having one singleton Thriveopedia instance in the future easier (or at least not harder).

hhyyrylainen (approved):
With the new design I think this works well and doesn't have scalability problems. Doesn't work in the auto-evo exploring tool though due to missing Thriveopedia instance, I've opened a follow up issue: https://github.com/Revolutionary-Games/Thrive/issues/4842

PR #4830, opened 2024-01-25
Fix Thriveopedia patch map not working when opened before editor

ender-titan1:

This PR fixes a problem with opening the Thriveopedia patch map before the editor, resulting in the patch map not loading.

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/thriveopedia/pages/ThriveopediaPatchMapPage.cs
+++ b/src/thriveopedia/pages/ThriveopediaPatchMapPage.cs
@@ -120,9 +120,14 @@ public class ThriveopediaPatchMapPage : ThriveopediaPage
         playerPatchOnEntry = mapDrawer.Map?.CurrentPatch ??
             throw new InvalidOperationException("Map current patch needs to be set / SetMap needs to be called");
 
-        // Make sure the map setting of fog of war always matches the world,
-        // this needs to be called here in case this page is opened before the editor
+        // Make sure the map setting of fog of war always matches the world
+        // These need to be called here in case this page is opened before the editor
         mapDrawer.Map.FogOfWar = CurrentGame!.GameWorld.WorldSettings.FogOfWarMode;
+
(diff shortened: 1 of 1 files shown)


src/thriveopedia/pages/ThriveopediaPatchMapPage.cs
```diff
@@ -104,6 +104,10 @@ protected override void Dispose(bool disposing)
     private void UpdatePlayerPatch(Patch? patch)
     {
         mapDrawer.PlayerPatch = patch ?? playerPatchOnEntry;
+
+        if (mapDrawer.Map!.UpdatePatchVisibility(mapDrawer.PlayerPatch))
```
hhyyrylainen, line 108: Please add an exception throw here if the map is null something like `throw new InvalidOperationException("This can be only called after map is set from current game");`

src/thriveopedia/pages/ThriveopediaPatchMapPage.cs
```diff
@@ -115,6 +119,16 @@ private void SetGameForMap()
         mapDrawer.Map = CurrentGame!.GameWorld.Map;
         playerPatchOnEntry = mapDrawer.Map?.CurrentPatch ??
             throw new InvalidOperationException("Map current patch needs to be set / SetMap needs to be called");
+
+        // Make sure the map setting of fog of war always matches the world
+        // These need to be called here in case this page is opened before the editor
+        mapDrawer.Map.FogOfWar = CurrentGame!.GameWorld.WorldSettings.FogOfWarMode;
```
hhyyrylainen, line 125: With the null suppression above on line 119, isn't this null suppression redundant and could be removed?

hhyyrylainen:
Code changes look fine to me now.

hhyyrylainen (approved):
I did a basic test in game that starting a new game and immediately viewing the thriveopedia map isn't blank.

PR #4832, opened 2024-01-26
Fixed engulfing not taking being pulled in object sizes into account

### LINKED ISSUE #4823: Going over engulf storage capacity incorrect ejects everything
Instead only enough things to get back under the capacity should be ejected. This is maybe a bug with the eject logic not updating the used engulf storage before continuing ejecting things (so in effect any item that causes the capacity to be crossed causes everything else to be also expelled). This is like a regression from the ECS refactor.

### LINKED ISSUE #4824: Engulfing shouldn't start on objects that will exceed the available engulfable storage
We got reports that attempting to engulf something that doesn't fit in the engulfing storage is started (and causes a related bug: https://github.com/Revolutionary-Games/Thrive/issues/4823). If this happens then this is a regression from the ECS refactor. Engulfing shouldn't start at all on something that would overflow the engulfed object storage. Also the bug might be in not taking account currently pulled in objects when starting new engulfing operations, which could explain why such a core feature of engulfing has a bug.

hhyyrylainen:
and also fixed ejecting not immediately seeing the new size and potentially ejecting everything

I have not tested this locally, so testing before merging is needed

**Related Issues**

closes #4824
closes #4823

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/components/Engulfer.cs
+++ b/src/microbe_stage/components/Engulfer.cs
@@ -45,7 +45,8 @@
         public float UsedIngestionCapacity;
 
         /// <summary>
-        ///   Total size that all engulfed objects need to fit in
+        ///   Total size that all engulfed objects need to fit in (the max value <see cref="UsedIngestionCapacity"/>
+        ///   should have)
         /// </summary>
         public float EngulfStorageSize;
     }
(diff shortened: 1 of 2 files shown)


CI09:
Looks good but I'm not especially good in making code look good myself🙃, so I'll just check these 2 items for now (from playtest it seems to work well)

ARecurringProblem (approved):

PR #4851, opened 2024-02-05
Renamed ICellProperties to ICellDefinition to make things clearer

hhyyrylainen:
where the different classes need to be used

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
+++ b/src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
@@ -79,7 +79,7 @@ public class EarlyMulticellularEditor : EditorBase<EditorAction, MicrobeStage>, 
     public Patch? SelectedPatch => patchMapTab.SelectedPatch;
 
     [JsonIgnore]
-    public ICellProperties? EditedCellProperties => selectedCellTypeToEdit;
+    public ICellDefinition? EditedCellProperties => selectedCellTypeToEdit;
 
     protected override string MusicCategory => "EarlyMulticellularEditor";
 
--- a/src/microbe_stage/editor/ICellEditorData.cs
+++ b/src/microbe_stage/editor/ICellEditorData.cs
@@ -3,5 +3,5 @@
 /// </summary>
 public interface ICellEditorData : IHexEditor, IEditorWithPatches, IEditorWithActions
 {
-    public ICellProperties? EditedCellProperties { get; }
+    public ICellDefinition? EditedCellProperties { get; }
 }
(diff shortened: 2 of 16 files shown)


src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
```diff
@@ -79,7 +79,7 @@ public override bool CanCancelAction
     public Patch? SelectedPatch => patchMapTab.SelectedPatch;
 
     [JsonIgnore]
-    public ICellProperties? EditedCellProperties => selectedCellTypeToEdit;
+    public ICellDefinition? EditedCellProperties => selectedCellTypeToEdit;
```
CI09, line 82: EditedCellProperties could also be renamed, not necessarily though

src/microbe_stage/editor/ICellEditorData.cs
```diff
@@ -3,5 +3,5 @@
 /// </summary>
 public interface ICellEditorData : IHexEditor, IEditorWithPatches, IEditorWithActions
 {
-    public ICellProperties? EditedCellProperties { get; }
+    public ICellDefinition? EditedCellProperties { get; }
```
CI09, line 6: Same here

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -49,7 +49,7 @@ public class MicrobeEditor : EditorBase<EditorAction, MicrobeStage>, IEditorRepo
         editedSpecies ?? throw new InvalidOperationException("species not initialized");
 
     [JsonIgnore]
-    public ICellProperties EditedCellProperties =>
+    public ICellDefinition EditedCellProperties =>
```
CI09, line 52: Here too

src/early_multicellular_stage/editor/EarlyMulticellularEditor.cs
hhyyrylainen: I was thinking about that, but as goes so high up the chain that at that level there doesn't really logically exist a cell definition that is being modified, but just the general properties of a thing that is being modified. I know that's not the best explanation but when I did initially rename this property as well, it didn't really fit in my head as to what the editor does, so I left it unused. This shouldn't really be able to cause confusion as there doesn't exist a gameplay simulation in the editor that would have `Components.CellProperties` that could be confused with this.

CI09 (approved):
Seems good but I was playing early multicellular prototype and my main cell membrane froze at red color from damage.
Also in awakening stage when founding society center, during the animationg where camera goes upwards game showed be patch extinction menu but it is likely not related

PR #4893, opened 2024-02-18
Prevent engulfing/binding sound effect continuing after death

Oliveriver:

Stops all sounds for a microbe on its death.

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/common_components/SoundEffectPlayer.cs
+++ b/src/engine/common_components/SoundEffectPlayer.cs
@@ -226,9 +226,9 @@
 
                     slot.InternalAppliedState = false;
                 }
-            }
 
-            soundEffectPlayer.SoundsApplied = false;
+                soundEffectPlayer.SoundsApplied = false;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/engine/common_components/SoundEffectPlayer.cs
```diff
@@ -189,6 +189,48 @@ public static bool PlaySoundEffect(this ref SoundEffectPlayer soundEffectPlayer,
...
+                    else
+                    {
+                        slot.Play = false;
+                    }
+
+                    slot.InternalAppliedState = false;
+                }
+            }
+
+            soundEffectPlayer.SoundsApplied = false;
```
hhyyrylainen, line 231: It would be better to have this inside the lock to be consistent with other methods and ensure everyone who uses lock on the slots will see this updated variable at the same time.

hhyyrylainen (changes requested):
I guess this is the most flexible way to ensure no weird sounds keep playing after microbe death. I have one code related concern though.

hhyyrylainen (approved):
Code looks good. There probably isn't too bad of a sound cutoff mid playing problem, but I didn't playtest this.

PR #4897, opened 2024-02-19
Increased minimum entity count options a bit

hhyyrylainen:
and made the options menu work with non-exact entity amounts more sensibly than before

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/utils/ListUtils.cs
+++ b/src/general/utils/ListUtils.cs
@@ -92,4 +92,32 @@ public static class ListUtils
 
         list.RemoveAt(itemCount - 1);
     }
+
+    /// <summary>
+    ///   Finds the closest value in the list of values and returns that
+    /// </summary>
+    /// <param name="valueToSearchFor">The value to look for</param>
+    /// <param name="values">Where to search for the value</param>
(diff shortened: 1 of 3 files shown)


src/general/utils/ListUtils.cs
```diff
@@ -92,4 +92,32 @@ public static void RemoveWithoutPreservingOrder<T>(this IList<T> list, int index
...
+    /// <returns>The closest element in <see cref="values"/> when compared to <see cref="valueToSearchFor"/></returns>
+    public static int FindClosestValue(int valueToSearchFor, params int[] values)
+    {
+        if (values.Length < 1)
+            throw new ArgumentException("Must be given at least one value to match against");
+
+        var closest = values[0];
+        var closestDistance = Math.Abs(valueToSearchFor - closest);
+
+        for (int i = 1; i < values.Length; ++i)
```
CI09, line 110: Should it be `i < values.Length;` or `i <= values.Length;`?

hhyyrylainen: `values.Length` is an invalid index meaning the code would crash with `<=`. Though, if I'm missing something feel free to clarify.

PR #4913, opened 2024-02-25
Auto Evo Miche Inital Implementation 

### LINKED ISSUE #1650: Make initial freebuild species more viable
fix up freebuild species generation for new auto-evo so that they don't immediately die.
https://github.com/Revolutionary-Games/Thrive/pull/1335#issuecomment-677851465

Might need to run auto-evo on the new species something like 2-3 times to get viable mutations for them.

### LINKED ISSUE #2918: Factor movement into Compound Food Source
Realistically, a microbe that survives off of compounds needs to be able to move, and can't be slow, sessile, or metabolically unable to support movement. Currently, https://github.com/Revolutionary-Games/Thrive/blob/master/src/auto-evo/simulation/food_source/CompoundFoodSource.cs doesn't support that.

### LINKED ISSUE #2942: Rework the crude extinction system into something proper
for 0.5.6.1 we quickly added an overall patch species cap that causes extinction to all the lowest population species (with some protections) to reduce the amount of species.

### LINKED ISSUE #3012: Auto-evo needs to take maximum movement direction into account
When calculating how well cells can move to find resources as otherwise they can just place random free flagella that increases their fitness but don't cost energy.

This is now needed after: https://github.com/Revolutionary-Games/Thrive/commit/0178e0cbcedcfa89b9bdc6c340e6071df352b432 which modifies the energy balance calculation to only take flagella that push in the maximum movement force direction of the cell.

### LINKED ISSUE #3044: Auto-evo predation math needs to take glucose use into account
auto-evo doesn't take glucose usability into account when calculating predation, meaning that fully iron eating microbe can appear to do very well from predation but in the game the species would die if they tried only eating glucose by consuming other cells

### LINKED ISSUE #4232: Don't apply auto-evo storage score to environmental food that doesn't vary during a day
As it seems like thermosynthesis has the storage score calculated for it due to it being environmental. The storage score should only be calculated for things that vary throughout a day/night cycle.

### LINKED ISSUE #5218: Evolutionary tree sometimes has extinct species circles detached from everything
kuva

Seems like there's some kind of species relationship that is not detected correctly causing those species that aren't linked to anything.

Reported at: https://community.revolutionarygamesstudio.com/t/0-6-7-release-candidate-testing/7497/19?u=hhyyrylainen

GameDungeon:

Implements Thim's Miche design for auto evo. While this is still far from being ready to merge it currently serves it's purpose as a prototype and is ready for preliminary testing and review. 

Closes #2918
Closes #2942
Closes #5218
Closes #1650
Closes #3012
Closes #4232
Closes #3044

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


GameDungeon:
Currently the cache is not thread-safe and it is possible for a race condition to cause a double add to a dictionary in the cache class. I don't know enough about our threading system to fix this.

hhyyrylainen:
Sorry for replying while on vacation but I saw this when skimming my emails and got the following thought which I want to write down:

The auto-evo simulation cache object is designed to not be shared across threads. That's why I haven't added thread safety there as each instance of the cache is meant to be used by a single thread. So the problem here is that cache instances are shared between threads and at that point things are going already very wrong (in terms of how the existing auto-evo architecture I designed is meant to work).

GameDungeon:
Thanks for the response despite being out. I'm not sure where it is crossing threads (as that is not intentional), but I guess I need to track it down. Part of me wonders if a global thread-safe cache would be more performant than our many local ones.

GameDungeon:
While there is a lot of small stuff left to do, I would be grateful for a high level review. Given how large this PR is, I don't want to start messing with small stuff until I'm sure the fundamentals are good.

simulation_parameters/microbe_stage/organelles.json
```diff
@@ -137,39 +137,6 @@
     "EditorButtonOrder": 5,
     "EndosymbiosisUnlocks": "nitrogenfixingplastid"
   },
-  "protoplasm": {
```
hhyyrylainen, line 140: Removing an organelle type necessitates a save breakage point so I'd prefer to avoid it if not absolutely required.

third_party/JoltPhysics
hhyyrylainen, line 1: This PR probably has a commit that reverts Jolt to a really old version...

src/auto-evo/AutoEvoRun.cs
```diff
@@ -6,9 +6,10 @@
 using System.Threading.Tasks;
 using AutoEvo;
 using Godot;
-using Xoshiro.PRNG64;
 using Thread = System.Threading.Thread;
 
+// using Xoshiro.PRNG64;
```
hhyyrylainen, line 11: Commented out stuff should be cleaned up if this whole miche feature is completed and it seems like this is no longer required.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -161,14 +162,16 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
     /// <param name="isPlayerSpecies">Whether this microbe is a member of the player's species</param>
     /// <param name="worldSettings">The world generation settings for this game</param>
     /// <param name="amountType">Specifies how changes during an in-game day are taken into account</param>
+    /// <param name="cache">AutoEvo Cache for speeding up the function</param>
```
hhyyrylainen, line 165: Auto-evo is written with a hyphen so I'd consider this a typo (also in anywhere else in comments where AutoEvo is written like that without a hyphen).

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -141,14 +141,20 @@ public static float CalculateCapacity(IEnumerable<OrganelleTemplate> organelles)
 
     // TODO: maybe this should return a ValueTask as this is getting pretty computation intensive
     public static float CalculateSpeed(IReadOnlyList<OrganelleTemplate> organelles, MembraneType membraneType,
-        float membraneRigidity, bool isBacteria)
+        float membraneRigidity, bool isBacteria, bool useEstimate = false)
     {
-        // This is pretty expensive as we need to generate the membrane shape and *then* the collision shape to figure
```
hhyyrylainen, line 146: This comment should be kept on the code that was moved.

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -191,8 +199,14 @@ public static float CalculateCapacity(IEnumerable<OrganelleTemplate> organelles)
                 rightwardDirectionMovementForce += MovementForce(movementConstant, rightDirectionFactor);
                 leftwardDirectionMovementForce += MovementForce(movementConstant, leftDirectionFactor);
             }
+
+            massEstimate += averageDensity * organelle.Definition.HexCount;
         }
 
+        // If this estimate could be made more accurate without additional computation that would be great
```
hhyyrylainen, line 206: Why doesn't the estimate use just the density of each organelle as it is processed above? Organelle mass estimates could be precalculated inside the OrganelleDefinition load to reduce runtime calculations.

src/gui_common/MicheDetailsPanel.cs
```diff
@@ -0,0 +1,99 @@
+﻿namespace AutoEvo;
+
+using System;
+using System.Linq;
+using Godot;
+
+/// <summary>
+///   Shows various details a bout a species for the player
```
hhyyrylainen, line 8:
```suggestion
///   Shows various details about a species to the player
```

src/gui_common/MicheDetailsPanel.cs
```diff
@@ -0,0 +1,99 @@
...
+/// </summary>
+public partial class MicheDetailsPanel : MarginContainer
+{
+    [Export]
+    public NodePath? MicheDetailsLabelPath;
+
+    public WorldGenerationSettings WorldSettings = null!;
```
hhyyrylainen, line 15: This should be a property and if this is set externally I would make this nullable to make the compiler force adding right checks that this class instance is setup correctly.

doc/auto_evo.md
hhyyrylainen, line 1: Would be pretty nice for me that this file was written before I try to understand the core Miche class etc.

src/auto-evo/Miche.cs
```diff
@@ -0,0 +1,212 @@
+﻿namespace AutoEvo;
+
+using System.Collections.Generic;
+using System.Linq;
+
+public class Miche
```
hhyyrylainen, line 6: This seems like a core class of the miche feature, right? So I think this class really ought to have a summary comment. Also I see quite a lot of memory allocations here. For example `GetLeafNodes` should probably be written to take in a list to put the results in to start guiding the code here towards a way of writing it that minimizes memory allocations where it is easy to do (I'm not going to ask to remove all repeated memory allocations but where I think they are easy to avoid it would be worth it in order to ensure auto-evo doesn't add a lot of garbage collector pressure while running).

src/auto-evo/AutoEvoRun.cs
GameDungeon: I'm not sure how / when Xoshiro is supposed to bu handled compared to normal random. I just commented it out as something to come back too when I figured out how it should be used.

hhyyrylainen: It should basically always be used instead of normal random. The reason the random is at the top level is so that it can be passed down to anything needing randomness in the auto-evo, and we'd hopefully in the future have fully deterministic auto-evo if given the same random seed and starting conditions. That would be really helpful in debugging any auto-evo problems (for players the results would still be random as the game uses the current time as the random seed when starting a new auto-evo run).

GameDungeon: Should I go through and replace any new randoms with that or should I leave that for a future issue? Do you have any suggestions on determinism with threads?

src/auto-evo/Miche.cs
GameDungeon: I'm not sure if "A node for the Miche Tree" is enough, but it seems to follow the general format of the others summaries.

src/auto-evo/AutoEvoRun.cs
hhyyrylainen: In .NET 8 the default system random actually defaults to using Xoshiro, but the seeds obviously need to be ferried through the method call sequences otherwise the randoms will not provide a consistent sequence. So I recommend that every place in auto-evo that creates a new random takes in a Random parameter that is used to generate a seed for the new random number generator. That will take care of basically everything that matters. Xoshiro randoms absolutely have to be stored in variables of Xoshiro type *for saving to work* but as auto-evo state is not saved in save files, that is not absolutely necessary.

src/auto-evo/Miche.cs
hhyyrylainen: That would work as a summary, but in addition I'd like to see a `remarks` section either explaining the miche tree shortly or linking to the miche tree class and then that class having an overall explanation for what the miche tree does.

GameDungeon:
At this point I would say this branch is ready to enter final review. Anything else I can think of to do would probably better be reserved for another PR to stop this one from scope creeping until it is unreviewable.

RevolutionaryGamesCommon
hhyyrylainen, line 1: I get the feeling this is accidentally downgrading the module.

doc/auto_evo.md
```diff
@@ -0,0 +1,75 @@
...
+species. Strategies are often very specific, eg add an organelle that produces
+ATP from iron or remove an organelle that consumes glucose. These allow Auto-Evo
+to only attempt to generate and try a specific subset of the possible mutations
+that are each much more likely to be an improvement on the parent. Each Mutation
+Strategy also has an MP cost to limit the amount of mutations per generation.
+
+### Steps
+Each time Auto-Evo is run it goes through a series of steps.
+
+### Generate Miche Step
```
hhyyrylainen, line 45: I think these are sub-sections of "steps" so these should have `####` to be a smaller heading.

third_party/CMakeLists.txt
```diff
@@ -13,11 +13,9 @@ if(THRIVE_AVX)
 
   # Not directly AVX instructions, but are about as new so are only on when AVX is on
   set(USE_F16C ON)
+  set(USE_FMADD ON)
   set(USE_LZCNT ON)
   set(USE_TZCNT ON)
-
-  # this is off anyway as this causes indeterminism in the simulation
-  set(USE_FMADD OFF)
```
hhyyrylainen, line 20: Why was this toggled? I'd prefer to keep the physics deterministic so that if we ever add multiplayer we don't have a ton of game state synchronization issues with clientside prediction.

third_party/JoltPhysics
hhyyrylainen, line 1: This also seems to be a module downgrade...

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -260,6 +289,7 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
         }
 
         result.FinalBalance = result.TotalProduction - result.TotalConsumption;
+        result.ConservativeFinalBalance = result.TotalProduction - result.TotalConsumption;
```
hhyyrylainen, line 292: Is there a typo on this line? This uses the same variables to calculate as the non-conservative version.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -192,6 +203,23 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
                     processATPProduction += amount;
 
                     result.AddProduction(organelle.Definition.InternalName, amount);
+
+                    var isInEnvironment = true;
+                    foreach (var input in processData.WritableInputs)
```
hhyyrylainen, line 208: The performance of this slightly concerns me. I think an improvement would be to first check / precalculate if the process has environmental inputs, and only then run this piece of code.

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -141,14 +141,23 @@ public static float CalculateCapacity(IEnumerable<OrganelleTemplate> organelles)
 
     // TODO: maybe this should return a ValueTask as this is getting pretty computation intensive
     public static float CalculateSpeed(IReadOnlyList<OrganelleTemplate> organelles, MembraneType membraneType,
-        float membraneRigidity, bool isBacteria)
+        float membraneRigidity, bool isBacteria, bool useEstimate = false)
     {
+        float shapeMass = 0;
+
         // This is pretty expensive as we need to generate the membrane shape and *then* the collision shape to figure
-        // out the mass. We rely on the caches working extra hard here to ensure reasonable performance.
```
hhyyrylainen, line 147: Second part of this comment is cut out, I think it should be put back on the 
`            var membraneShape = MembraneComputationHelpers.GetOrComputeMembraneShape(organelles, membraneType);` line.

src/auto-evo/EditorAutoEvoRun.cs
```diff
@@ -21,6 +21,11 @@ protected override void GatherInfo(Queue<IRunStep> steps)
         var map = Parameters.World.Map;
         var worldSettings = Parameters.World.WorldSettings;
 
+        foreach (var entry in map.Patches)
+        {
+            steps.Enqueue(new GenerateMiche(entry.Value, new SimulationCache(worldSettings), worldSettings));
```
hhyyrylainen, line 26: Does the parallelism help here a lot? If not it might be overall better to force the steps to not run in parallel and share a single simulation cache between all of them.

src/microbe_stage/systems/ProcessSystem.cs
GameDungeon: Yep that is a typo. Good catch.

src/auto-evo/EditorAutoEvoRun.cs
GameDungeon: Well I was referring to GenerateMiche specifically. That should be a super quick step that just preps the miche for the future steps. I think the additional time is added by the mutation code checking far more mutations. It's a time / result tradeoff to some extent. I think I have a few ideas to get it to prune out some more sub-optimal mutations though. That should speed stuff up a bit.

hhyyrylainen: In case if the step doesn't benefit from parallelism, it would be then better to share the cache instances to avoid doing a bunch of recalculations on a single thread that had already done the work.

src/auto-evo/steps/GenerateMiche.cs
```diff
@@ -0,0 +1,136 @@
+﻿namespace AutoEvo;
+
+public class GenerateMiche : IRunStep
+{
+    public Patch Patch;
+    public SimulationCache Cache;
```
hhyyrylainen, line 6: These seem public for no reason?

src/auto-evo/steps/GenerateMiche.cs
```diff
@@ -0,0 +1,136 @@
...
+
+    private static readonly Compound Glucose = SimulationParameters.Instance.GetCompound("glucose");
```
hhyyrylainen, line 8: These being static is a bit iffy to me as that keeps state for the whole program. I've much preferred using non-static references to the registry types as that still reduces the lookups a lot but doesn't force keep the references around for the whole program runtime, which could be a problem in some modding contexts (that's basically all I got, but I'd like to keep the future flexibility and avoid static references to registry stuff).

hhyyrylainen: Also doing the references in the constructor would allow eliminating duplicate property reads (`SimulationParameters.Instance` is a property that I assume the compiler has to re-read for each of these field initializers).

src/auto-evo/steps/MigrateSpecies.cs
```diff
@@ -0,0 +1,64 @@
...
+    public bool RunStep(RunResults results)
+    {
+        var miche = results.MicheByPatch[patch];
+
+        var occupants = miche.GetOccupants().Distinct().ToList();
+
+        if (occupants.Count == 0)
+            return true;
+
+        for (int i = 0; i < Constants.AUTO_EVO_MOVE_ATTEMPTS; i++)
```
hhyyrylainen, line 32:
```suggestion
        for (int i = 0; i < Constants.AUTO_EVO_MOVE_ATTEMPTS; ++i)
```

src/auto-evo/steps/MigrateSpecies.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            var species = occupants.ToList().Random(random);
```
hhyyrylainen, line 34: Instead of creating a list and immediately discarding it, it would be better to update our random utils to directly be able to use the `occupants` here.

src/auto-evo/steps/MigrateSpecies.cs
```diff
@@ -0,0 +1,64 @@
+﻿namespace AutoEvo;
+
+using System;
+using System.Linq;
+
+public class MigrateSpecies : IRunStep
```
hhyyrylainen, line 6: A summary comment on this would be great. If I understood this correctly, then migrations now happen a fixed number of attempts per patch. Is that better than each species be given the same number of migration attempts?

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -20,28 +20,28 @@
 /// </remarks>
 public class SimulationCache
 {
+    public readonly Dictionary<(MicrobeSpecies, SelectionPressure), float> CachedPressureScores = new();
```
hhyyrylainen, line 23: This should be private.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -61,22 +61,26 @@ public EnergyBalanceInfo GetEnergyBalanceForSpecies(MicrobeSpecies species, Biom
...
         cached = ProcessSystem.ComputeEnergyBalance(species.Organelles, biomeConditions, species.MembraneType,
-            true, species.PlayerSpecies, worldSettings, CompoundAmountType.Average);
+            maximumMovementDirection, true, species.PlayerSpecies, worldSettings, CompoundAmountType.Average, this);
 
         cachedEnergyBalances.Add(key, cached);
         return cached;
     }
 
+    // TODO: Both of these seem like something that could easily be stored on the species with OnEdited
     public float GetBaseSpeedForSpecies(MicrobeSpecies species)
```
hhyyrylainen, line 75: Name should be updated of this method as this isn't the base speed anymore.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -95,148 +99,214 @@ public float GetBaseHexSizeForSpecies(MicrobeSpecies species)
...
-    /// </summary>
-    /// <returns>
-    ///   A float to represent score. Scores are only compared against other scores from the same FoodSource,
-    ///   so different implementations do not need to worry about scale.
-    /// </returns>
-    public float GetEnergyGenerationScoreForSpecies(MicrobeSpecies species, BiomeConditions biomeConditions,
-        Compound compound)
-    {
-        var key = (species, biomeConditions, compound);
+        // don't use mucus for now
```
hhyyrylainen, line 268: Maybe a TODO comment here? Also I noticed there's a few comments in around this line that don't start with a capital letter.

src/auto-evo/selection_pressure/SelectionPressure.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        Strength = strength;
+        Mutations = mutations;
+    }
+
+    public abstract float Score(Species species, SimulationCache cache);
+    public abstract float GetEnergy();
+
+    // TODO: Make sure this does not make things worse
+    public float CacheScore(MicrobeSpecies species, SimulationCache cache)
```
hhyyrylainen, line 21: Why is there a separate method like this? Wouldn't it be better to bake this into the Score method (and make it non-abstract) and then add a new protected abstract method called something like `CalculateScore`.

src/auto-evo/selection_pressure/SelectionPressure.cs
```diff
@@ -0,0 +1,76 @@
...
+    public void ApplyTranslations()
+    {
+        TranslationHelper.ApplyTranslations(this);
+    }
+
+    /// <summary>
+    ///   A description of this miche. Needs to support translations changing and be player readable
+    /// </summary>
+    /// <returns>A formattable that has the description in it</returns>
+    public abstract IFormattable GetDescription();
```
hhyyrylainen, line 73: Why doesn't this return `LocalizedString`? Which is our Thrive type for handling user translatable content. (or the localized string builder if that is more suitable).

src/auto-evo/selection_pressure/SelectionPressure.cs
```diff
@@ -0,0 +1,76 @@
...
+
+    public abstract override string ToString();
```
hhyyrylainen, line 75: Could you add a comment why this forces all child classes to implement `ToString`?

src/auto-evo/selection_pressure/RootPressure.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace AutoEvo;
+
+using System;
+
+public class RootPressure : SelectionPressure
+{
+    // Needed for translation extraction
+    // ReSharper disable ArrangeObjectCreationWhenTypeEvident
+    public static readonly LocalizedString Name = new LocalizedString("ROOT_PRESSURE");
```
hhyyrylainen, line 9: I think some of the new auto-evo translation keys are too simple for translators to understand. Maybe prefixing most if not all of them with `MICHE_` or here with `PRESSURE_` ? Prefixes also help the localization tool from grouping the related text to be translated at once.

src/auto-evo/selection_pressure/RootPressure.cs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public override float GetEnergy()
+    {
+        return 0;
+    }
+
+    public override IFormattable GetDescription()
+    {
+        // This shouldn't be called on 0 energy pressures
```
hhyyrylainen, line 33: Why is this comment here? This sounds more like what the base class should have.

src/auto-evo/selection_pressure/PredatorRoot.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    public override IFormattable GetDescription()
+    {
+        // This shouldn't be called on 0 energy pressures
+        return Name;
+    }
+
+    public override string ToString()
+    {
+        return Name.ToString();
```
hhyyrylainen, line 61: Couldn't the base class have this as the default implementation as it seems to be the same in most of the pressure classes?

src/auto-evo/selection_pressure/NoOpPressure.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace AutoEvo;
+
+using System;
+
+public class NoOpPressure : SelectionPressure
```
hhyyrylainen, line 5: `op` is not clear what it means here so a summary comment would be very nice.

src/auto-evo/simulation/SimulationCache.cs
GameDungeon: It is referenced from the cachedScore. Not sure if it is possible to make private while cachedScore still exists. Well unless I moved that to this class, which in hindsight probably makes more sense.

src/auto-evo/mutation_strategy/IMutationStrategy.cs
```diff
@@ -0,0 +1,12 @@
...
+
+using System;
+using System.Collections.Generic;
+
+public interface IMutationStrategy<T>
+    where T : Species
+{
+    public bool Repeatable { get; }
+
+    public List<Tuple<T, float>> MutationsOf(T baseSpecies, float mp);
```
hhyyrylainen, line 11: To avoid memory allocations I think it would be maybe preferable to change the signature of this... Maybe something like `public T? GenerateMutation(T baseSpecies, float mp, Func<T, bool> evaluator)` that could evaluate things one by one (of course the delegate would need some way to cache it to not create too many of them, but this could allow much more efficient use and re-using species objects as well (I'll still need to dig into some of the core classes here to make more concrete suggestions).

src/auto-evo/simulation/SimulationCache.cs
hhyyrylainen: I did try in my IDE making this private and it didn't give an error, but it may not have seen that reference immediate. Better to move all the caching stuff here instead of exposing something that no one outside this class should be allowed to modify for code encapsulation reasons.

src/auto-evo/mutation_strategy/CommonMutationFunctions.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+
+    private static int[] SideTraversalOrder(Hex hex, Direction direction, Random random)
+    {
+        if (hex.Q < 0)
+        {
+            if (direction == Direction.Front)
+            {
+                return [1];
```
hhyyrylainen, line 149: Sadly this syntax allocates memory each time, so all of these lists need to be cached in static const variables. Also this should return `Hex.HexSide` for type safety reasons as we are working with a type safe language and not JavaScript here.

src/auto-evo/mutation_strategy/CommonMutationFunctions.cs
```diff
@@ -0,0 +1,192 @@
...
+                                return result;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        // We didn't find an open spot, this doesn't make much sense
+        throw new Exception("Mutation code could not find a good position " +
```
hhyyrylainen, line 87: Exceptions for kind of foreseeable things is not good, I tracked down one problem with the game completely freezing in the .NET runtime to throwing background thread exceptions. I modified the copy in master of finding a good organelle position to return a nullable value I think. Maybe this copy of the function hasn't gotten the latest fixes?

src/auto-evo/RunResults.cs
```diff
@@ -18,6 +18,8 @@
 /// </remarks>
 public class RunResults : IEnumerable<KeyValuePair<Species, RunResults.SpeciesResult>>
 {
+    public Dictionary<Patch, Miche> MicheByPatch = new();
```
hhyyrylainen, line 21: This should be private to avoid other code messing with this in a major way, and to match how this class handles the other results to not have a 50/50 implementation of a POD and a functioning class type.

src/auto-evo/MutationLogicFunctions.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace AutoEvo;
+
+using System.Linq;
+
+public class MutationLogicFunctions
```
hhyyrylainen, line 5: I think this + the mutation common functions would be best served to find a new home in a new folder like `auto-evo/mutations` and also bringing in the main Mutations.cs file there would be good for code grouping.

src/auto-evo/Miche.cs
```diff
@@ -0,0 +1,210 @@
...
+///     it is considered a leaf node and can have a species Occupant instead. This class handles insertion into the
+///     tree through scores from the selection pressure it contains. For a fuller explanation see docs/auto_evo.md
+///   </para>
+/// </remarks>
+public class Miche
+{
+    public Miche? Parent;
+    public List<Miche> Children = new();
+
+    // Occupant should always be null if this Miche has children
```
hhyyrylainen, line 21: This should be a summary.

src/auto-evo/Miche.cs
```diff
@@ -0,0 +1,210 @@
...
+///   <para>
+///     The Miche class forms a tree by storing a list of child instances of Miche Nodes. If a Miche has no children
+///     it is considered a leaf node and can have a species Occupant instead. This class handles insertion into the
+///     tree through scores from the selection pressure it contains. For a fuller explanation see docs/auto_evo.md
+///   </para>
+/// </remarks>
+public class Miche
+{
+    public Miche? Parent;
+    public List<Miche> Children = new();
```
hhyyrylainen, line 19: I think it would be good to have this and `Pressure` be read only or is there a valid use case of these being changed after the miche is created?

src/auto-evo/Miche.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        if (IsLeafNode())
+        {
+            nodes.Add(this);
+            return nodes;
+        }
+
+        foreach (var child in Children)
+        {
+            nodes.AddRange(child.GetLeafNodes());
```
hhyyrylainen, line 60: Shouldn't this be `child.GetLeafNodes(nodes)`? That would avoid many memory allocations.

src/auto-evo/Miche.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        if (Occupant != null)
+        {
+            occupants.Add(Occupant);
+            return occupants;
+        }
+
+        foreach (var child in Children)
+        {
+            occupants.AddRange(child.GetOccupants());
```
hhyyrylainen, line 81: Same here.

src/auto-evo/Miche.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        if (Parent == null)
+        {
+            return this;
+        }
+
+        return Parent.Root();
+    }
+
+    public IEnumerable<Miche> GetLeafNodes()
```
hhyyrylainen, line 45: I think it would be pretty good to remove these memory allocating variants of these various methods. That would force all code to be updated to the approach that allocates fewer temporary list objects.

src/auto-evo/Miche.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        foreach (var child in newChildren)
+        {
+            AddChild(child);
+        }
+    }
+
+    public bool InsertSpecies(Species species, SimulationCache cache, bool dry = false)
+    {
+        var scoresDictionary = GetOccupants().Distinct().ToDictionary(x => x, _ => 0.0f);
```
hhyyrylainen, line 118: I think avoiding LINQ here would be good as this probably allocates memory.

Also I saw that Distinct is used quite many times so it would probably be pretty good to create a separate GetDistinctOccupants method (that takes in a HashSet<Species> as the parameter to fill as the result).

src/auto-evo/Miche.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        if (IsLeafNode() && Occupant == null)
+        {
+            if (!dry)
+                Occupant = species;
+
+            return true;
+        }
+
+        var newScores = new Dictionary<Species, float>();
```
hhyyrylainen, line 147: There's probably some way to write this in a way that reduces memory allocations...

src/auto-evo/Miche.cs
```diff
@@ -0,0 +1,210 @@
...
+
+                if (dry)
+                    return true;
+            }
+        }
+
+        return inserted;
+    }
+
+    public Miche DeepCopy()
```
hhyyrylainen, line 180: I don't yet have any suggestion here yet but I dread a bit how much memory allocation this might do. I'll do another go at memory allocation profiling once master is next merged into this.

src/auto-evo/simulation/MichePopulation.cs
```diff
@@ -44,12 +37,41 @@ public static class PopulationSimulation
...
 
+    /// <summary>
+    ///   Estimates the initial population numbers
+    /// </summary>
+    public static int CalculateMicrobePopulationInPatch(MicrobeSpecies species, Miche miche,
+        BiomeConditions biomeConditions, SimulationCache cache)
+    {
+        var energyBalanceInfo = cache.GetEnergyBalanceForSpecies(species, biomeConditions);
+        var individualCost = energyBalanceInfo.TotalConsumptionStationary + energyBalanceInfo.TotalMovement
+            * species.Behaviour.Activity / Constants.MAX_SPECIES_ACTIVITY;
```
hhyyrylainen, line 53: This math seems pretty complicated so extracting this as a method would be excellent to ensure there's just one source of truth for how much individual cost is in a species.

src/auto-evo/simulation/MichePopulation.cs
```diff
@@ -44,12 +37,41 @@ public static class PopulationSimulation
...
+
+        var leafNodes = miche.GetLeafNodes();
+
+        var miches = leafNodes.Where(x => x.Occupant == species).Select(x => x.BackTraversal())
```
hhyyrylainen, line 57: This LINQ chain seems like something that should be written out in code. Maybe the miche class could have this energy collection method?

src/auto-evo/simulation/MichePopulation.cs
```diff
@@ -141,157 +163,112 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
...
     private static void SimulatePatchStep(SimulationConfiguration simulationConfiguration, Patch patch,
-        IEnumerable<Species> genericSpecies, Random random, SimulationCache cache,
-        IAutoEvoConfiguration autoEvoConfiguration, WorldGenerationSettings worldSettings)
+        IEnumerable<Species> genericSpecies, Random random, SimulationCache cache)
     {
         _ = random;
 
         var populations = simulationConfiguration.Results;
         bool trackEnergy = simulationConfiguration.CollectEnergyInformation;
-        bool dayNightCycle = worldSettings.DayNightCycleEnabled;
```
hhyyrylainen, line 167: Does the relevant selection pressures adjust correctly when day/night is disabled? Or is that automatic thanks to using the average light value now?

src/auto-evo/simulation/MichePopulation.cs
```diff
@@ -141,157 +163,112 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
...
 
-        // This algorithm version is for microbe species
-        // TODO: add simulation for multicellular
-        var species = genericSpecies.Select(s => s as MicrobeSpecies).Where(s => s != null).Select(s => s!)
-            .ToList();
+        populations.MicheByPatch.TryGetValue(patch, out var miche);
+
+        foreach (var s in simulationConfiguration.ExtraSpecies)
```
hhyyrylainen, line 189:
```suggestion
        foreach (var extraSpecies in simulationConfiguration.ExtraSpecies)
```

Foreach loops can't use single variable names according to our styleguide. Or at least I hope it kind of says that there (with just a few single variable letter names being allowed in some contexts).

src/auto-evo/simulation/MichePopulation.cs
```diff
@@ -141,157 +163,112 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
...
+            miche!.InsertSpecies(s, cache);
+
+        var species = genericSpecies.Where(x => !simulationConfiguration.ExcludedSpecies.Contains(x))
+            .Union(simulationConfiguration.ExtraSpecies).Distinct().ToList();
```
hhyyrylainen, line 193: While this LINQ looks nice, using it in the core of the population simulation is probably not the best design choice.

src/auto-evo/simulation/MichePopulation.cs
```diff
@@ -141,157 +163,112 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
...
-            new ChunkFoodSource(patch, "marineSnow"),
-            new ChunkFoodSource(patch, "ironSmallChunk"),
-            new ChunkFoodSource(patch, "ironBigChunk"),
-        };
-
-        foreach (var currentSpecies in species)
-        {
-            niches.Add(new HeterotrophicFoodSource(patch, currentSpecies, cache));
-        }
+        var leafNodes = miche!.GetLeafNodes().Where(x => x.Occupant != null).ToList();
```
hhyyrylainen, line 199: If getting the miche fails above, it should probably throw an exception as something is very wrong.

src/auto-evo/simulation/MichePopulation.cs
```diff
@@ -141,157 +163,112 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
...
+        var energyDictionary = species.ToDictionary(x => x, _ => 0.0);
 
-        foreach (var niche in niches)
+        foreach (var node in leafNodes)
```
hhyyrylainen, line 202: I see a bunch of enumerator and other memory allocations here, I'll need to get back to this later.

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -0,0 +1,327 @@
...
+                outputSpecies.Clear();
+
+                foreach (var speciesTuple in inputSpecies)
+                {
+                    var mutated = mutationStrategy.MutationsOf(speciesTuple.Item1, speciesTuple.Item2);
+
+                    outputSpecies.AddRange(PruneMutations(speciesTuple.Item1, mutated, cache, selectionPressures));
+                }
+
+                // TODO: Make these a performance setting?
```
hhyyrylainen, line 58: Should be at least a Constant for easy modifying.

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -0,0 +1,327 @@
...
+    public bool RunStep(RunResults results)
+    {
+        // TODO: Have this be passed in
+        var random = new Random();
+
+        var oldMiche = results.MicheByPatch[Patch];
+        var oldOccupants = oldMiche.GetOccupants().Distinct().ToList();
+
+        // TODO: Possibly make this a performance setting?
+        const int possibleMutationsToTryPerSpecies = 3;
```
hhyyrylainen, line 156: Isn't there an auto-evo config variable for this already? That should be hooked up to here.

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -0,0 +1,327 @@
...
+    public Patch Patch;
+    public SimulationCache Cache;
+
+    public ModifyExistingSpecies(Patch patch, SimulationCache cache)
+    {
+        Patch = patch;
+        Cache = cache;
+    }
+
+    public int TotalSteps => 1;
```
hhyyrylainen, line 20: This seems like a step that should *really* be able to be split into multiple ones. It is an important part of auto-evo that each step would take at most like 50-100ms but in my testing with miches some steps took dozens of seconds to complete. One immediate suggestion I have is to preserve the state of looking for good changes and just doing at most like 50 at once before returning and letting the auto-evo system call the step method again.

src/microbe_stage/systems/ProcessSystem.cs
GameDungeon: This is probably a case of bad variable naming on my part, but this check is instead checking that the inputs for the process are present in that Patch at all, not if they are environmental. isInPatch would likely be a better name.

src/auto-evo/Miche.cs
GameDungeon: No, I agree that making this readonly would be good.

src/auto-evo/steps/MigrateSpecies.cs
GameDungeon: I'm not sure better is really objective here, but with the miche data structure it is certainly easier.

src/auto-evo/selection_pressure/PredatorRoot.cs
GameDungeon: Afaik, no. The way Name is defined prevents that.

src/auto-evo/Miche.cs
GameDungeon: I added a GetDistinctOccupants, but I quickly realized that there was not a single place in the code that I wanted to use the original GetOccupants. As any usecase is pretty wall covered by GetLeafNodes, I just made GetDistinctOccupants into GetOccupants and replaced it entirely.

src/auto-evo/Miche.cs
GameDungeon: After rereading, I will note that Parent and children can't be readonly due to the way the Miche Tree is constructed. It is a lot easier to do outside of a constructor function to avoid like 6 layers of nesting. I was agreeing that Pressure should be readonly.

src/auto-evo/Miche.cs
GameDungeon: That is my bad, I thought I updated that properly.

src/auto-evo/Miche.cs
GameDungeon: The only comfort I can provide here is that in the latest version of the code that function is only called by the auto-evo exploring tool

src/auto-evo/simulation/MichePopulation.cs
GameDungeon: Due to Process Speed and the average light, I believe it should be automatic in all of the places I have checked.

src/auto-evo/mutation_strategy/CommonMutationFunctions.cs
GameDungeon: I can't seem to get this happy with making a Hex.HexSide[] static

src/auto-evo/steps/GenerateMiche.cs
GameDungeon: So should it just be const?

src/microbe_stage/systems/ProcessSystem.cs
hhyyrylainen: But wouldn't it then make more sense to always check that the process can run even theoretically in the patch? Because shouldn't energy balance contribution be 0 from processes that have no chance of running at all?

src/auto-evo/Miche.cs
hhyyrylainen: Note that making the list `readonly` only prevents reassigning the list, modifying the contents will still be possible even if the variable is readonly.

src/auto-evo/steps/MigrateSpecies.cs
hhyyrylainen: I guess that is fine then as long as the performance or gameplay implications aren't bad. Still I'd like to see the summary comment added for this class.

src/auto-evo/selection_pressure/PredatorRoot.cs
hhyyrylainen: Can't name be an abstract property then to force it to exist and allow access to it in the base class?

src/auto-evo/Miche.cs
hhyyrylainen: That's good enough for me, as long as this is not in the main algorithms, I'm fine with having a quick and dirty thing that's just used for GUI stuff outside the main game loop.

src/auto-evo/simulation/MichePopulation.cs
hhyyrylainen: Yeah alright that makes sense to me. Let's keep this in mind while checking stuff just in case there's some bug that gets noticed related to this at some point.

src/auto-evo/mutation_strategy/CommonMutationFunctions.cs
hhyyrylainen: To me it seems like this should work:
```
private static readonly Hex.HexSide[] traversalOrder1 = [Hex.HexSide.Top];
```
(put in the right place in the class). And then `return traversalOrder1` here. At least my IDE doesn't immediately complain that shouldn't compile.

src/auto-evo/steps/GenerateMiche.cs
hhyyrylainen: It cannot be const as this needs to be looked up during runtime (in hindsight it probably would have been better if Compound was an enum but we don't live in that timeline sadly). So just remove `static` and leave it be readonly and fetch all of these compounds in the constructor.

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -0,0 +1,337 @@
...
+
+    /// <summary>
+    ///   Returns a new list of all possible species that might emerge in response to the provided pressures,
+    ///   as well as a copy of the original species.
+    /// </summary>
+    /// <returns>List of viable variants, and the provided species</returns>
+    public static List<MicrobeSpecies> GenerateMutations(MicrobeSpecies baseSpecies, int amount,
+        SimulationCache cache, List<SelectionPressure> selectionPressures, Random random)
+    {
+        // TODO: Possibly make that 100 a difficulty setting
```
hhyyrylainen, line 35: Doesn't the MP numbers here need to be hooked up to the AI mutation rate variable that is in new game settings? Otherwise that option will stop working.

src/auto-evo/selection_pressure/PredatorRoot.cs
GameDungeon: It could be made into a property, but it would be no longer static. I'm not sure if that is an issue.

hhyyrylainen: I guess one kind of hacky way would be to have the abstract property but then implement it as returning a private static field.

```c#
private static readonly LocalizedString name = ...;

public override LocalizedString Name => name;
```

Something along those lines. In some situations I haven't cared and just assigned the property individually for each instance but I guess in this case there would be a benefit to the above pattern to avoid a bunch of duplicate localized string objects in memory.

src/auto-evo/Miche.cs
GameDungeon: Sadly I had to bring back the call to this in MutateSpeciesStep. Will need to find a way around that.

hhyyrylainen: Once we are down to last few files needing review, I'll try to fully understand the core of the miche alrogithm, and at that point I should hopefully be able to offer design help related to this.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -189,9 +199,26 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
...
                 {
-                    processATPProduction += amount;
-
                     result.AddProduction(organelle.Definition.InternalName, amount);
+
+                    var isInPatch = true;
+
+                    foreach (var input in processData.WritableInputs)
+                    {
+                        if (biome.Compounds.TryGetValue(input.Key, out var inputCompoundData))
```
hhyyrylainen, line 208: Actually doesn't this not see any chunk type compounds? If that is the case then I guess it would make sense to re-introduce that separate variable tracking compound balance when relying purely on clouds. Sorry for not realizing this sooner.

src/auto-evo/selection_pressure/RootPressure.cs
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public override float GetEnergy()
+    {
+        return 0;
+    }
+
+    public override LocalizedString GetDescription()
+    {
+        return Name;
```
hhyyrylainen, line 33: A default implementation of this can be in the base class, with only child classes that want to override `return Name;` needing to then define this method explicitly.

src/auto-evo/mutations/CommonMutationFunctions.cs
```diff
@@ -0,0 +1,237 @@
...
+        newSpecies.Organelles.Add(position);
+
+        // If the new species is a eukaryote, mark this as such.
+        if (organelle == Nucleus)
+        {
+            newSpecies.IsBacteria = false;
+        }
+    }
+
+    public static OrganelleTemplate? GetRealisticPosition(OrganelleDefinition organelle,
```
hhyyrylainen, line 74: This method still doesn't have all the upgrades that GetRealisticPosition has on master, including taking in the working memory as parameters to avoid things like line 107.

src/auto-evo/mutations/CommonMutationFunctions.cs
```diff
@@ -0,0 +1,237 @@
...
+        var workMemory2 = new List<Hex>();
+        var workMemory3 = new Queue<Hex>();
+        var islandHexes = new List<Hex>();
+
+        organelles.GetIslandHexes(islandHexes, workMemory1, workMemory2, workMemory3);
+
+        // Attach islands
+        while (islandHexes.Count > 0)
+        {
+            var mainHexes = organelles.ComputeHexCache().Except(islandHexes);
```
hhyyrylainen, line 143: This should use the hex cache variant taking in an existing collection to put the result in.

src/auto-evo/mutation_strategy/AddOrganelleAnywhere.cs
```diff
@@ -0,0 +1,78 @@
...
+        this.direction = direction;
+    }
+
+    public bool Repeatable => true;
+
+    public static AddOrganelleAnywhere ThatUseCompound(Compound compound, CommonMutationFunctions.Direction direction
+        = CommonMutationFunctions.Direction.Neutral)
+    {
+        return new AddOrganelleAnywhere(organelle => organelle.RunnableProcesses
+            .Where(proc => proc.Process.Inputs.ContainsKey(compound)).Any(), direction);
```
hhyyrylainen, line 26: I think these can be written in a shorter form like: `RunnableProcesses.Any(proc => proc.Process.Inputs.ContainsKey(compound))`.

I won't suggest this yet but if it turns out that constructing the mutation strategies causes many allocations, pre-caching on game startup these differently configured mutation strategies might be the optimal way to do these.

src/auto-evo/steps/GenerateMiche.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        // Iron
+        if (patch.Biome.Chunks.TryGetValue("ironSmallChunk", out var smallChunk) && smallChunk.Density > 0)
+        {
+            var ironMiche = new Miche(new CompoundConversionEfficiencyPressure(patch, iron, atp, 1.0f));
+
+            ironMiche.AddChild(new Miche(new ChunkCompoundPressure(patch, 1, "ironSmallChunk", iron)));
+
+            if (patch.Biome.Chunks.TryGetValue("ironBigChunk", out var bigChunk) && bigChunk.Density > 0)
```
hhyyrylainen, line 68: What if there's a patch that has big iron but no small iron?

src/auto-evo/steps/MigrateSpecies.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        var occupantsSet = new HashSet<Species>();
+        miche.GetOccupants(occupantsSet);
+
+        var occupants = occupantsSet.ToList();
+
+        if (occupants.Count == 0)
+            return true;
+
+        for (int i = 0; i < Constants.AUTO_EVO_MOVE_ATTEMPTS; ++i)
```
hhyyrylainen, line 39: I think as there's this loop, it would be pretty easy to split this step up into sub-steps to better have the auto-evo progress flow smoothly (which it doesn't do when there's only like 80 steps even at generation 15).

src/microbe_stage/systems/ProcessSystem.cs
GameDungeon: I think it might be worth actually checking for the compound in chunks if the first check fails. That will make it work as intended.

hhyyrylainen: It could but at that point I think the compound amount calculation method should be split out to a helper method as the editor also needs to display this information (it would make sense if process numbers and GUI display numbers were the same, just with some scaling probably in the editor display).

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -924,6 +1002,85 @@ private void EvolutionaryTreeNodeSelected(int generation, uint id)
...
+
+            generateMiche.PopulateMiche(newMiche);
+
+            micheTree.SetMiche(newMiche);
+        }
+    }
+
+    private void MicheTreeNodeSelected(int micheHash)
+    {
+        if (!micheTree.MicheByHash.TryGetValue(micheHash, out var micheData))
```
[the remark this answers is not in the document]
GameDungeon: Even if they overlap it would just display the incorrect miche, not error here. Still not great, but I don't think it is an issue in the grand scheme of things.

src/auto-evo/Miche.cs
hhyyrylainen: Ah, so each so to speak "branch" of the execution path through the tree requires its own scores?

GameDungeon: Yep

hhyyrylainen: If this turns out to be one of the top sources of memory allocations, that's not going to be fun, but for now I'll mark this as resolved.

src/auto-evo/mutation_strategy/AddOrganelleAnywhere.cs
GameDungeon: If there was a simple way to cache all of the mutation stratiges that are used by GenerateMiche in between Auto-Evo runs I agree that would be ideal. Once created they are readonly so sharing them between threads or even generations should pose no issues.

src/auto-evo/mutations/CommonMutationFunctions.cs
```diff
@@ -0,0 +1,237 @@
...
+            // Calculate the path to move island organelles.
+            // If statement is there because otherwise the path could be (0, 0).
+            if (minSubHex.Q != minSubHex.R)
+                minSubHex.Q = (int)(minSubHex.Q * (minDistance - 1.0) / minDistance);
+
+            minSubHex.R = (int)(minSubHex.R * (minDistance - 1.0) / minDistance);
+
+            // Move all island organelles by minSubHex
+            foreach (var organelle in organelles.Where(o => islandHexes.Any(h =>
+                         o.Definition.GetRotatedHexes(o.Orientation).Contains(h - o.Position))))
```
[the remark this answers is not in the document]
GameDungeon: Turning that where into an if statement is pretty easy, but I'm not sure of a good way to handle that Any. With just the if it would still have to capture organelle for the any.

src/auto-evo/mutation_strategy/AddOrganelleAnywhere.cs
hhyyrylainen: Yeah, maybe they could even be stored in the auto-evo settings object to have all auto-evo runs forever be able to share them.

src/auto-evo/mutations/CommonMutationFunctions.cs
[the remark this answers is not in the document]
hhyyrylainen: Maybe the only option is to just unpack everything as manual loops that build the result of stuff to move?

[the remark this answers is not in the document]
GameDungeon: It just took me a second to figure out the best way to go about that. I figured it out now.

src/auto-evo/mutation_strategy/AddOrganelleAnywhere.cs
GameDungeon: Thinking about it more, what I said above also applies to all of the Selection Pressures except predation (which obviously depends on the spieces in the patch), so it would probably better to store those.

src/auto-evo/selection_pressure/PredationEffectivenessPressure.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    public readonly Species Prey;
+
+    // Needed for translation extraction
+    // ReSharper disable ArrangeObjectCreationWhenTypeEvident
+    private static readonly LocalizedString NameString = new LocalizedString("MICHE_PREDATION_EFFECTIVENESS_PRESSURE");
+
+    // ReSharper restore ArrangeObjectCreationWhenTypeEvident
+
+    private static readonly Compound Oxytoxy = SimulationParameters.Instance.GetCompound("oxytoxy");
```
hhyyrylainen, line 13: Registry loaded stuff shouldn't be in static variables.

src/auto-evo/selection_pressure/PredationEffectivenessPressure.cs
```diff
@@ -0,0 +1,78 @@
+﻿namespace AutoEvo;
+
+public class PredationEffectivenessPressure : SelectionPressure
+{
+    public readonly Species Prey;
```
hhyyrylainen, line 5: Does this have to be public?

src/auto-evo/selection_pressure/MetabolicStabilityPressure.cs
```diff
@@ -0,0 +1,53 @@
+﻿namespace AutoEvo;
+
+public class MetabolicStabilityPressure : SelectionPressure
+{
+    // Needed for translation extraction
+    // ReSharper disable ArrangeObjectCreationWhenTypeEvident
+    private static readonly LocalizedString NameString = new LocalizedString("MICHE_METABOLIC_STABILITY_PRESSURE");
+
+    // ReSharper restore ArrangeObjectCreationWhenTypeEvident
+    private static readonly Compound ATP = SimulationParameters.Instance.GetCompound("atp");
```
hhyyrylainen, line 10: I'll stop checking these pressures for now as many seem to have the same static variable storing registry data problem.

src/auto-evo/selection_pressure/PredationEffectivenessPressure.cs
GameDungeon: Most of them have static variables because otherwise the MutationStrategies constructors error. Anything that does not need to be marked static is not.

src/auto-evo/selection_pressure/PredationEffectivenessPressure.cs
GameDungeon: Yes, it is checked aginst in ModifyExistingSpeices as a way to get the preadators of a creature.

src/auto-evo/selection_pressure/PredationEffectivenessPressure.cs
hhyyrylainen: ~~You can put the field initializer in the constructor.~~ Update: I'm like 99% sure that this works:

```c#
private readonly Compound atp = SimulationParameters.Instance.GetCompound("atp");
```

As used here: https://github.com/Revolutionary-Games/Thrive/blob/3f77b72693693028db228a430462fad2237700f0/src/microbe_stage/organelle_components/MovementComponent.cs#L11

GameDungeon: I remove the static I just get this: `An object reference is required for the non-static field, method, or property 'AvoidPredationSelectionPressure.ChitinMembrane'`

hhyyrylainen: Like a default value or? Or base constructor? For those cases using null as a placeholder value meaning "no value, will be fetched in the method" has been what I do.

Still some other way needs to be used here than allowing a static field to hold a Compound reference.

GameDungeon: The selection pressures all create mutation strats (with these compounds) and pass them into there parent though the base constructor. I don't see a good way to avoid this being an issue.

src/microbe_stage/systems/ProcessSystem.cs
GameDungeon: Well the separated function exists now, but I'm not sure what to do with it.

src/auto-evo/selection_pressure/PredationEffectivenessPressure.cs
hhyyrylainen: Some other alternative needs to be made here. Could maybe even make the super class constructors take in internal names of compounds and they then handle calling `GetCompound` if nothing else comes to mind for anyone.

GameDungeon: That might actually simplify some of the code and fixes this so I will just go ahead and implement that.

src/auto-evo/selection_pressure/SelectionPressure.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// <summary>
+    ///   A description of this miche.
+    /// </summary>
+    public virtual LocalizedString GetDescription()
+    {
+        return Name;
+    }
+
+    // ToString is used to display the Selection Pressure in the Miche Tree
+    public override string ToString()
```
hhyyrylainen, line 61: Why is ToString used for displaying to the user? This seems a bit bad to me as elsewhere in Thrive I use ToString purely for easier to see values in the debugger.

src/auto-evo/selection_pressure/EnvironmentalCompoundPressure.cs
```diff
@@ -0,0 +1,76 @@
...
+    public override float GetEnergy(Patch patch)
+    {
+        return patch.Biome.AverageCompounds[compound].Ambient * energyMultiplier;
+    }
+
+    public override LocalizedString GetDescription()
+    {
+        // TODO: somehow allow the compound name to translate properly. We now have custom BBCode to refer to
+        // compounds so this should be doable
+        return new LocalizedString("DISSOLVED_COMPOUND_FOOD_SOURCE", compound.Name);
```
hhyyrylainen, line 69: I'll fix this but for future reference I think the correct code is:
```c#
        return new LocalizedString("DISSOLVED_COMPOUND_FOOD_SOURCE",
            new LocalizedString(compound.GetUntranslatedName()));
```

src/auto-evo/MicheTree.cs
```diff
@@ -0,0 +1,441 @@
...
+
+    public override void _Ready()
+    {
+        base._Ready();
+
+        tree = GetNode<Control>(TreePath);
+
+        treeNodeScene = GD.Load<PackedScene>("res://src/auto-evo/MicheTreeNode.tscn");
+
+        // Font size is adjusted dynamically so this needs to be a copy.
```
hhyyrylainen, line 129: Is this comment or the code wrong? As far as I can see this is not modified, and `bodyFont` is totally unused. I'll make these changes myself to remove the unused things.

src/auto-evo/simulation/MichePopulation.cs
```diff
@@ -141,157 +161,124 @@ private static List<Species> CopyInitialPopulationsToResults(SimulationConfigura
...
 
-        foreach (var niche in niches)
+        // Skip if there aren't any species in this patch
+        if (species.Count < 1)
+            return;
+
+        var leafNodes = new List<Miche>();
+        miche.GetLeafNodes(leafNodes, x => x.Occupant != null);
+
+        var energyDictionary = species.ToDictionary(x => x, _ => 0.0);
```
hhyyrylainen, line 213: Was this intentionally or accidentally using doubles? I'll swap this to floats.

hhyyrylainen: Or actually I'll just for safety put these back to double as if the numbers get so high that it is important for accuracy.

src/auto-evo/selection_pressure/EnvironmentalCompoundPressure.cs
GameDungeon: This was copy pasted from somewhere else in the code base that did the same thing.

hhyyrylainen: Searching for "compound name to translate" gives no results. So could it be that this was copy-pasted before I fixed this elsewhere in the codebase? Which I guess would be a really good example why copy-pasting is harmful as fixing it in one place still leaves the other places with the problem.

src/auto-evo/simulation/MichePopulation.cs
GameDungeon: Accidental, it should have been a float

hhyyrylainen: Alright, I'll put it as floats then but add a comment about checking it.

hhyyrylainen (approved):
The code is now clean but the auto-evo doesn't run when trying it due to a missing score in a dictionary. Probably one of my memory allocation improvement changes caused that. I'm totally out of time for today but  I *will* pick this back up tomorrow morning if someone else cannot fix it my then.

PR #4954, opened 2024-03-19
Remove 3 variables trying to access another scene's nodes

CI09:
and added a replacement function. Also disconnected unnecessary signals.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/base_stage/CreatureStageHUDBase.cs
+++ b/src/general/base_stage/CreatureStageHUDBase.cs
@@ -808,7 +808,8 @@ public partial class CreatureStageHUDBase<TStage> : HUDWithPausing, ICreatureSta
             GD.PrintErr("can't get reproduction phosphates progress: ", e);
         }
 
-        EmitSignal(SignalName.OnUpdateReproductionProgressBars, fractionOfAmmonia, fractionOfPhosphates, AmmoniaBW, PhosphatesBW);
+        EmitSignal(SignalName.OnUpdateReproductionProgressBars, fractionOfAmmonia, fractionOfPhosphates,
+            AmmoniaBW, PhosphatesBW);
     }
 
     protected virtual void CalculatePlayerReproductionProgress(Dictionary<Compound, float> gatheredCompounds,
--- a/src/general/base_stage/CreatureStageHUDBase.cs
+++ b/src/general/base_stage/CreatureStageHUDBase.cs
@@ -39,15 +39,6 @@ public partial class CreatureStageHUDBase<TStage> : HUDWithPausing, ICreatureSta
     public NodePath HealthBarPath = null!;
 
     [Export]
-    public NodePath AmmoniaReproductionBarPath = null!;
-
-    [Export]
-    public NodePath PhosphateReproductionBarPath = null!;
-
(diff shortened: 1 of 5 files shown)


src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -229,6 +217,12 @@ protected CreatureStageHUDBase()
     [Signal]
     public delegate void OnOpenMenuToHelpEventHandler();
 
+    [Signal]
+    public delegate void OnSetEditorButtonFlashEffectEventHandler();
```
hhyyrylainen, line 221: Instead of a signal here, this should just directly call methods on the editor button. Note that I see some variables like `protected TextureButton editorButton = null!;` still existing. That should be updated to be:

```c#
[Export]
protected EditorEntryButton editorButton = null!;
```

To be able to call the C# methods on that variable.

hhyyrylainen (changes requested):

src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -817,11 +800,7 @@ protected void UpdateReproductionProgress()
             GD.PrintErr("can't get reproduction phosphates progress: ", e);
         }
 
-        ammoniaReproductionBar.Value = fractionOfAmmonia * ammoniaReproductionBar.MaxValue;
-        phosphateReproductionBar.Value = fractionOfPhosphates * phosphateReproductionBar.MaxValue;
-
-        CheckAmmoniaProgressHighlight(fractionOfAmmonia);
-        CheckPhosphateProgressHighlight(fractionOfPhosphates);
+        editorButton.UpdateReproductionProgressBars(fractionOfAmmonia, fractionOfPhosphates, AmmoniaBW, PhosphatesBW);
```
hhyyrylainen, line 803: I think these black and white textures should be managed by the editor button scene, as that's pretty much a functionality entirely managed by it. I don't think the complexity of needing to pass in specific texture variants to this methods is worth it (as I can't see a reason why the editor button would need to act differently in some special condition).

src/microbe_stage/gui/EditorEntryButton.cs
```diff
@@ -11,7 +11,51 @@ public partial class EditorEntryButton : TextureButton
...
+    [Export]
+    private TextureRect phosphateIcon = null!;
+
+    [Export]
+    private TextureProgressBar phosphateReproductionBar = null!;
+
+    [Export]
+    private PointLight2D editorButtonFlash = null!;
 #pragma warning restore CA2213
+    public void SetEditorButtonFlashEffect(bool enabled)
```
hhyyrylainen, line 30: There should be a blank line after a pragma directive.

hhyyrylainen (changes requested):
This is starting to remove a good number of variables but I think there's still some more stuff that can be moved to the editor entry button scene to avoid needing to specify those in the HUDs.

hhyyrylainen (approved):
This is probably fine now, I assume your testing hasn't brought up any more problems with this.

PR #4991, opened 2024-04-01
Colour picker fix

CI09:

Removed most of customizations for built-in colour picker. Might need some post-fixes, for example make preset colors work across editor sessions, but it should work.
Also on new OKHSL mode most pallete types (The color rect thing you pick color from, now there are variants of it) don't work and just default is used; maybe changing pallete types could be disabled or something but it's not gamebreaking issue.

**Related Issues**

 - https://github.com/orgs/Revolutionary-Games/projects/6/views/1?pane=issue&itemId=56460930
 - https://github.com/orgs/Revolutionary-Games/projects/6/views/1?pane=issue&itemId=57742832
 - Maybe https://github.com/orgs/Revolutionary-Games/projects/6/views/1?pane=issue&itemId=56559611

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Is this now ready? I put various of the colour picker related things to the "in progress" column on the Godot 4 planning board.

CI09:
It can be reviewed.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -109,182 +62,55 @@ public new ColorModeType ColorMode
...
-            presetsContainer.Visible = value;
-            addPresetButton.Visible = value;
-        }
-    }
-
     // TODO: fix this with Godot 4
     /// <summary>
     ///   Hide Godot property PresetsEnabled to avoid unexpected changes
     ///   which may cause the hidden native buttons reappear.
     /// </summary>
```
hhyyrylainen, line 73: This seems like a comment leftover for a removed property.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -109,182 +62,55 @@ public new ColorModeType ColorMode
...
-    private bool PresetsEnabled { get => PresetsEditable; set => PresetsEditable = value; }
-
     public override void _Ready()
     {
         base._Ready();
 
         // Hide replaced native controls. Can't delete them because it will crash Godot.
         // TODO: fix this: this no longer has the same child layout
```
hhyyrylainen, line 79: This TODO comment was about Godot 4, so if this PR works this TODO comment needs to be removed.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -35,66 +34,20 @@ public partial class TweakedColourPicker : ColorPicker
...
-    private double pickerTimeElapsed;
-
-    private PresetGroupStorage? groupStorage;
-
-    private delegate void AddPresetDelegate(Color colour);
-
-    private delegate void DeletePresetDelegate(Color colour);
-
     [Export]
     public string PresetGroup { get; private set; } = "default";
```
hhyyrylainen, line 49: This should have a big warning that this property no longer works. I assume this doesn't work as I saw you remove all of the group handling code.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -35,66 +34,20 @@ public partial class TweakedColourPicker : ColorPicker
...
-    private TextureButton? addPresetButton;
 #pragma warning restore CA2213
 
-    private bool hsvButtonEnabled = true;
-    private bool rawButtonEnabled = true;
-    private bool presetsEditable = true;
-    private bool presetsVisible = true;
-
     // TODO:
     private ColorModeType mode = ColorModeType.Rgb;
```
hhyyrylainen, line 46: It looks like to me that this variable should be removed. This is not updated anywhere so the tooltips which depend on this end up incorrect when the user switches to HSV mode for example.

src/gui_common/TweakedColourPicker.cs
```diff
@@ -559,34 +203,6 @@ public override void _Notification(int what)
             base._Notification(what);
         }
 
-        public override void _GuiInput(InputEvent inputEvent)
```
hhyyrylainen, line 562: It seems you removed the functionality of this child class, but didn't add back anything? To me it looks like this class is entirely non-functional now so this should be entirely deleted to avoid confusion for future programmers who try to reimplement the preset sharing function.

hhyyrylainen (changes requested):

src/gui_common/TweakedColourPicker.tscn
```diff
@@ -15,83 +11,4 @@ offset_bottom = 28.0
...
-size_flags_horizontal = 0
-size_flags_vertical = 4
-texture_normal = ExtResource("3")
-texture_pressed = ExtResource("4")
-texture_hover = ExtResource("4")
-texture_disabled = ExtResource("5")
-ignore_texture_size = true
-stretch_mode = 5
-
 [connection signal="color_changed" from="." to="." method="OnColourChanged"]
```
hhyyrylainen, line 14: It seems like this callback method was deleted from the code.

hhyyrylainen (changes requested):
I didn't test in game yet, but I noticed some things that weren't cleaned up already and some code style related problems. But nothing major looks to be wrong with this so this should be easy enough to finish now.

locale/en.po
```diff
@@ -658,9 +658,6 @@ msgstr "Colour"
 msgid "COLOURBLIND_CORRECTION"
 msgstr "Colourblind correction:"
 
-msgid "COLOUR_PICKER_ADD_PRESET"
```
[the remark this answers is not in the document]
CI09: In full diff view it is re-added a few lines below (although with hashtags, I don't know what that means)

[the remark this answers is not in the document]
hhyyrylainen: The system is keeping old text around when removed. If you add the keys back, you should just revert all the translation file changes to have a clean PR.

src/gui_common/TweakedColourPicker.cs
CI09: Regular comment or summary? The latter one would be more noticable I think.

hhyyrylainen: XML comment as this is a field. And also maybe remove the "export" attribute so that no one can set this from the Godot editor and wonder why it doesn't do anything?

CI09:
I hope I didn't mess up too much.

hhyyrylainen (changes requested):
I fixed most of the problems I saw but there's still 2 bugs: the red slider tooltip doesn't get set and instead the green slider gets that tooltip:


Also it seems like the tooltips are no longer updated when switching the mode (being in HSV mode still shows the RGB tooltips). I was sure that one of your earlier commits should have had that working but it seems like you removed the signal connection that made that possible?

hhyyrylainen:
Also one more point, I think the tabs for selecting the type of colour mode should be kept as those are part of the standard Godot control and I think they lead pretty well to the player being able to discover that HSV is supported.

CI09:
Should I replace dropdown with the buttons then? That makes issue of OKHSL not having (I think) its button (it's only in dropdown menu)

CI09:
Seems this event doesn't work (or my code doesn't work):
image

hhyyrylainen:
The signal will fail due to parameters if you didn't put the dummy parameter values back.

CI09:
I thought it was complaining about dummyIndex'es presence earlier but seems it doesn't now. It should be now fixed.

hhyyrylainen (approved):
The behaviour of the picker now seems fully correct to me. I did a few slight code tweaks to get the last few things solved I saw in there.

PR #4995, opened 2024-04-03
Replace CustomExpandingWordWrappedLabel with regular Label in credits scroll scene

CI09:

Also removes it.

**Related Issues**

https://github.com/orgs/Revolutionary-Games/projects/6?pane=issue&itemId=56456688

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/gui_common/CreditsScroll.cs
```diff
@@ -605,7 +605,12 @@ private string GetTranslatedHeading(string team)
 
         foreach (var columnText in splitTexts)
         {
-            var label = new CustomExpandingWordWrappedLabel(columnText.ToString());
+            var label = new Label
+            {
+                Text = columnText.ToString(),
+                HorizontalAlignment = HorizontalAlignment.Center,
+                SizeFlagsHorizontal = SizeFlags.ExpandFill,
```
hhyyrylainen, line 612: Shouldn't this actually specify smart word wrap? (and minimum width of 50 pixels)

hhyyrylainen: I ask because I recall that Labels don't have word wrap enabled by default... so this *shouldn't* work correctly without word wrap enabled.

CI09: Seems smart word wrap throws the text much lower than where it should be. Regular word (not smart) doesn't do that though.

hhyyrylainen: Smart is needed because if someone has a name like "thisisareallylongwordwithnospacesinitthatpreventsnormalwordwrapfromworkingatallonthisname" things will break.

Having the labels be contained in a hbox is probably the best way to ensure they can expand and take up all of the horizontal space to make things work well. Though ensuring uniform column widths might be tiny bit complicated.

CI09: Well, that `label` is already added as child to the hbox so I don't know why that's happening.

hhyyrylainen: Make sure the hbox is also expand and the minsize is set to equal to screen width.

src/gui_common/CreditsScroll.cs
```diff
@@ -646,7 +647,7 @@ private int CreateTeamNameList(int offset, string team, IEnumerable<string> peop
         // If past team leads need to be marked might need to use rich text label or some other approach there.
         // And this method might need to be split into separate implementations as not all places would need that.
         var memberLabel = CreateDynamicPart(offset, people, columns);
-        offset += (int)memberLabel.Height + ExtraOffsetAfterTeam;
+        offset += (int)Math.Round((memberLabel.Height + ExtraOffsetAfterTeam) / columns);
```
hhyyrylainen, line 650: Can you explain *how* this works? The number of columns should have *no* impact on height value. The height value is the height of the Godot control node here, the control node shouldn't multiply its height by the number of columns.

CI09:
I think it should be good now.

hhyyrylainen (approved):
The code change is now nice and simple and I verified in game that the credits show up correctly.

PR #5006, opened 2024-04-08
Fix panoramic backgrounds

CI09:
**Brief Description of What This PR Does**

Hopefully fixes panoramic backgrounds. Now creates a whole new Sky instance instead of modifying any that could already exist but I didn't make them because I think Godot was messing the scene a bit and I didn't want to commit it together with the .cs files fix.

**Related Issues**

https://github.com/orgs/Revolutionary-Games/projects/6/views/1?pane=issue&itemId=59027737

**Progress Checklist**

Note: before starting this checklist the PR should be marked as non-draft.

- [x] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [ ] Initial code review passed (this and further items should not be checked by the PR author)
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author. Merging must follow our
[styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md#git).

PR #5008, opened 2024-04-09
Fixed basic playing through of all the prototypes

hhyyrylainen:

Makes the prototypes mostly the same experience as before with only a few slight visual and movement speed related things that could be tweaked further

**Related Issues**

closes #5006

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09 (approved):
I took a peek at the changes and they look fine to me. From playtesting, I noticed that the ground is missing in industrial stage and rock model is kinda breaking when moving. Initial part of credits scroll is kinda off center and also this thing (bottom left)
image
but it doesn't really affect the gameplay. Also creature is kinda glitching in the ascension cutscene but that's okay.

CI09:
I also happened to drag some window around in space stage and upon clicking it disappeared (as designed) but bringing it back would make it still follow cursor and I wasn't able to drop it back down. Minor thing though.

hhyyrylainen:
> but it doesn't really affect the gameplay

Food going negative is effect of growing until there's no more food to be had, and the science -1 is a literal placeholder.

Both of those have been so since I made the prototypes.

> I noticed that the ground is missing in industrial stage and rock model is kinda breaking when moving

Interesting that I did not see that, for me society and industrial stages worked fine.

> I also happened to drag some window around in space stage and upon clicking it disappeared (as designed) but bringing it back would make it still follow cursor and I wasn't able to drop it back down. Minor thing though.

I'm not yet fully sure what we should do with our custom window system as it causes tons of those uneven anchors warnings. Might just need to replace it entirely with the standard Godot window class and hope it was improved enough for 4 that we don't need any hacks to make it work for our use cases.

PR #5046, opened 2024-04-17
Make licenses loading faster by increasing initial text size

CI09:

This PR does some stuff...

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/dialogs/LicensesDisplay.cs
+++ b/src/gui_common/dialogs/LicensesDisplay.cs
@@ -141,7 +141,7 @@ public partial class LicensesDisplay : CustomWindow
                 AutowrapMode = TextServer.AutowrapMode.WordSmart,
 
                 // Increase the initial size to make loading faster
-                Size = new Vector2(1000, 1000),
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
Does this specifically need the extra space on both axes (x, and y)? Also as the resulting height will be huge I'd be fine with setting the initial height to be even much larger (like 5000 probably still is less than the final height) if that helps performance even more.

CI09:
Seems that Y axis size increase didn't help much, instead X axis does that.

hhyyrylainen:
Well X axis is going to matter for how many line breaks there needs to be calculated to happen for line wrapping. 

Would be interesting to also try pre-processing the license text to ensure like there aren't any lines longer than 70 characters, and if there are splitting them before giving it to Godot. That + disabling word wrapping might be the ultimate solution but I can merge this PR as it is now just fine tomorrow.

hhyyrylainen (approved):
This makes loading the license menu so much faster.

PR #5057, opened 2024-04-22
Disabled more instructions for the non-AVX Thrive native library build

hhyyrylainen:

As I noticed some of the other instruction flags were about as new instructions as AVX, so I disabled those as well in the non-AVX build.

**Related Issues**

https://community.revolutionarygamesstudio.com/t/becoming-a-thrive-developer/7415/9?u=hhyyrylainen

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Can't confirm that this works for the original user reporting, but I at least confirmed this works fine with a really bad Windows 7 laptop (though it did work even before, but I think it's better to not use instructions that are probably unsupported if AVX isn't supported).

PR #5087, opened 2024-04-30
Improve Chemoreceptor Visuals

### LINKED ISSUE #3073: Implement thicker line drawing and use it for chemoreceptor and tutorial guidance line
to make it easier to see. There's a TODO comment in the code about this (we need to switch to a different drawing algorithm).

### LINKED ISSUE #4371: Improve visuals for the chemoreceptor line
Although functional, honestly it's bad. 

Currently the chemoreceptor uses a simple line to show the way to compounds. Ideally, it would be either a better-looking line or some kind of directed particle effect.

Needs somebody proficient in graphics.

Links:
- https://forum.revolutionarygamesstudio.com/t/chemoreceptor-organelle/765
- https://discord.com/channels/464846402732687391/464846875187478548/880892194024722553
- https://discord.com/channels/464846402732687391/464846906971914250/979535726398021724

@Oliveriver has done some things on this but it's no longer continued (for reference):

https://github.com/Revolutionary-Games/Thrive/assets/54026083/bc93a4af-64ea-4c10-94df-6f8672fc6ae4

wabuilderman:

Changed GuidenceLine mesh from line to triangle
Added lineWidth variable to GuidenceLine

**Related Issues**
closes #4371
closes #3073

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/engine/GuidanceLine.cs
```diff
@@ -85,11 +102,21 @@ public override void _Process(double delta)
...
+        mesh.SurfaceBegin(Mesh.PrimitiveType.Triangles);
 
         mesh.SurfaceSetColor(colour);
-        mesh.SurfaceAddVertex(LineStart);
+        
+        // To form quad, we want it in 'orgin + vector' form, not 'start + end' form
+        // Be sure to flatten the Y-axis of the vector, so it's all on a 2D plane
+        Vector3 lineVector = lineEnd - lineStart;
+        lineVector[1] = 0.0f;
+        float lineVectorLength = (float)Math.Sqrt(lineVector[0]*lineVector[0] + lineVector[2]*lineVector[2]);
```
hhyyrylainen, line 113: Why not use the length (or was it Distance) property available in the vector class?
Also same thing a few lines below, there's a `.Normalized()` method that returns a normalized variant.

hhyyrylainen: Oh and does this handle the case where the line ends up being exactly 0 length? Some player is probably going to run into that at some point in the future and then make a bug report if the line behaves strange in that situation.

wabuilderman: I tried, but then it was throwing me a bunch of errors as if the Vector3 didn't have those methods. I suppose it would be more proper for me to figure out *why* it wasn't working, so the code would be cleaner. I can work to revise that.

hhyyrylainen: Yeah, that sounds very strange. As-is this code doesn't fit in with the rest of the codebase because there's 0 of this manual vector math anywhere else, it's all done using the vector class methods to make it clear what is being done.

wabuilderman: Okay; I'll work on a new version that both uses the built-in vector math and tests if the length is zero first. Thank you.

hhyyrylainen:
Also one more thing I thought of to check: how does this look if the player has for example 5 different chemoreceptors (set to detect different compounds) and the lines are pointing almost in the same direction? If there's z-fighting or some other glitching that would be good to fix as well. I may not have time to review this again until next week but some other programmer might be able to do a review for you in the meantime.

hhyyrylainen:
Will this be ready for a final review on Monday? (that's when I'll definitely have time to check this fully).

src/engine/GuidanceLine.cs
```diff
@@ -76,6 +97,7 @@ public override void _Ready()
         var material = new StandardMaterial3D();
         material.VertexColorUseAsAlbedo = true;
         MaterialOverride = material;
+        yOffset = random.NextFloat() - 2.0f;
```
hhyyrylainen, line 100: Is this visible? What I mean that would it be safer to multiply the offset by 0.5f here to make the range half as big to maybe avoid some visual artifacts.

I tested in game and I didn't really notice anything, so I'll just comment this for future reference.

PR #5095, opened 2024-05-03
Buff corpse chunks

CI09:

Buffs compounds from corpse chunks.

For testers:
 - you can check if marine snow (randomly encountered cell chunks) are worth searching for
 - you can check if killing cells (other than engulfing) doesn't grant too much compounds than it should
 - you can check if sticking to and absorbing compounds that are passively vented out of either of above don't grant too much or too least compounds

---

 - you can check if they do not despawn too quickly

**Related Issues**

Scavenging and combat balancing (it doesn't fully complete them but helps) item on roadmap

**Progress Checklist**

      break existing features:
      members reviewing this PR)


HexapodPhilosopher:
The marine snow doesn't provide much. You get about the same from swallowing it and absorbing the cloud it already left as from waiting for it to fully disintegrate - about the same as 3 Ingested Matter. It's definitely worth swallowing them, but not worth seeking them out if there are other clouds nearby.

I seem to always be able to get the same from killing a cell and then engulfing the chunks as from engulfing the whole cell. It seems like the chunks barely vent anything, though.

HexapodPhilosopher:
The marine snow is now definitely worth chasing down to engulf. You get a lot less from the cloud it creates, though.

I seem to always be able to get the same from killing a cell and then engulfing the chunks as from engulfing the whole cell. The chunks still seem to vent very little, though.

The small iron chunk is much more worth while to engulf, now. I feel it should probably be made to count as more Ingested matter, but that's probably a different issue.

Maybe the VentAmount should be 0.3f ?

hhyyrylainen:
I think one important parameter that is not tweaked is how much of a compound is actually preserved when spawned as a cloud. I think buffing that is going to be important to get this balanced so that cell wall gameplay balance is not destroyed.

CI09:
Isn't this managed by VentAmount?

hhyyrylainen:
No, I'm talking about `Constants.CHUNK_VENT_COMPOUND_MULTIPLIER`

CI09:
It might be too strong now (especially marine snow) but I'm open for opinion.

buckly90:
This is a great start!

Marine snow chunks are wildly satiating now. I recommend lowering their compound amount by a significant degree. Probably by 75%. (This would make the initial compounds 12-13 for nutrients, and 25 for glucose).
Other than that, I think this is a good step in the right direction.

I think that we could make due with a few changes to how corpse chunks retain their compounds, such as ensuring they always contain some minimum amount of glucose before they dissolve as often times they can be severely lacking in energy value despite their abundance of nutrients (phosphate, ammonia). But I would like to discuss that with the team before we commit to it so I would consider that an optional change for this PR.

CI09:
Thanks for review. I might try to rebalance marine snow, though it is quite bonded with other chunk emission (including corpse chunks)

hhyyrylainen:
Does this make marine snow too overpowered, then? If so then I think some further tweaking would be warranted before merging.

CI09:
AI Microbes seem to ignore marine snow which makes competition over them almost non-existant, so...
Anyways, reduced some value so it might change something.

buckly90:
Works great! Thank you for your work.

hhyyrylainen:
This buffs the compound amounts in corpse chunks by 10x, but if it is better game balance I guess I can accept that and I'll merge this on Monday.

PR #5100, opened 2024-05-06
Mucocyst Lite

CI09:

Adds upgrade to slime jet with purpose of letting the player encapsulate themselves in a layer of mucilage effectively protecting their cell from any harm at cost of greatly reduced linear and agular speed and blocked abilities and compound absorption (passive and from compound clouds).

Can be unbalanced at the moment.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09:
I wrote in description that it adds new organelle but I corrected it to "upgrade".

CI09:
ToDo:
 - Visual effect around cell
 - Text for mucocyst name, description and hotbar tooltip
 - Something for AI cells maybe I guess

hhyyrylainen:
Should this be marked as a draft at least until a visual effect can be made? And when the text is also done.

Deus-Codes:
I will try to text this tonight

Deus-Codes:
Took it for a test run and I can see why you mention it as being overpowered in Discord.

I think two things that will help balance it...

1.) Instead of having the mucocyst ability slowly draining stored slime, we should probably have it completely deplete slime storage to 0. This, with the combination of it only being activated with a full bar of slime, should make it more balanced and require more judgement from the player as to when it is deployed. 

2.) We should probably make it so that the upgrade drains a bit more glucose. This makes it a more demanding upgrade which the player will have to build around rather than just an additional ability they can tack on. Maybe we can make it so that the mucocyst is slightly slower at producing slime as well, though that might be something which requires more playtesting. As the player grows, so does their storage, so that adds another layer of balance.

Great job though in terms of functionality. I think it's an exciting ability.

CI09:
I tweaked it to drain mucilage when active. But I think it shouldn't be too slow at regenerating mucilage but instead should drain quite some mucilage. That way players will have short window of mucilage usage they'll have to time correctly. But if you think your idea is better I can implement that just as well.

HexapodPhilosopher:
I've been over the effect code again. My C# knowledge is not good enough to make the changes needed. I'll just upload what I had before. Basically, you'd need to add a function to scale the vertices of the effect mesh if the effect that's needed is different than what was last used. 
Something like this to scale it down for switching to the mucocyst effect:
                vertices[index + 1] = new Vector3(vertices[index].X*1.1, height / 2, vertices[index].Y*1.1);
                vertices[index] = new Vector3(0.5f, height / 2, 0.5f);

hhyyrylainen:
Modifying vertices during runtime is extremely inefficient and should be avoided. Wouldn't a transform scale work on the model?

CI09:
I was thinking of using the same thing engulf effect used.

hhyyrylainen:
Engulf effect doesn't scale anything as it is generated at the same time as the membrane mesh to be a matching size for the membrane.

HexapodPhilosopher:
As the engulf effect mesh is a ring, scaling it would leave a gap in the middle. It would be more efficient to have the mesh as a circle (but matching the cell shape). It would mean adjusting the engulf shader to exclude the middle. We could then have the mucocyst shader set to exclude the outside, and wouldn't need to scale the mesh at all.

hhyyrylainen:
I assume this is probably also ready for another round of testing?

CI09:
Almost - all mechanics work, I just need help with visual effect not being accessable from MucocystSystem somehow.

shaders/EngulfEffect.gdshader
hhyyrylainen, line 1: Is it intentional that this PR changes the engulfing effect?

simulation_parameters/Constants.cs
```diff
@@ -435,6 +435,18 @@ public static class Constants
     /// </summary>
     public const float MUCILAGE_COOLDOWN_TIMER = 1.5f;
 
+    /// <summary>
+    ///   How many time cell gets slowed down with mucocyst (slime jet upgrade) on
```
hhyyrylainen, line 439:
```suggestion
    ///   How many times cell gets slowed down with mucocyst (slime jet upgrade) on
```

simulation_parameters/Constants.cs
```diff
@@ -435,6 +435,18 @@ public static class Constants
...
+    /// </summary>
+    public const float MUCOCYST_SPEED_DIVISOR = 60;
+
+    public const float MUCOCYST_SPEED_MINIMUM_MUCILAGE = 0.2f;
```
hhyyrylainen, line 443:
```suggestion
    public const float MUCOCYST_MINIMUM_MUCILAGE = 0.2f;
```
Should this variable be named like this (so removing the speed part of the name)?

shaders/EngulfEffect.gdshader
CI09: Hexapod tweaked it and I accepted it assuming it's better. Probably is.

hhyyrylainen: I see a lot of changes to the membrane etc. changes related to the engulfing effect. If that is in this PR, I need to insist on getting visual feedback on that change from other team members (visual comparisons should be posted on the dev discord for feedback).

src/microbe_stage/MicrobeState.cs
```diff
@@ -19,4 +19,9 @@
     ///   The microbe is currently in engulf mode
     /// </summary>
     Engulf,
+
+    /// <summary>
+    ///   The microbe is currently shielded my mucocyst shield
```
hhyyrylainen, line 24:
```suggestion
    ///   The microbe is currently shielded by mucocyst shield
```

src/microbe_stage/components/MicrobeControl.cs
```diff
@@ -181,6 +182,10 @@ public static void SetMoveSpeed(this ref MicrobeControl control, float speed)
         if (organelleInfo.SlimeJets == null || organelleInfo.SlimeJets.Count < 1)
             return;
 
+        // Not valid if all slime jets are mucocysts
+        if (organelleInfo.SlimeJets.All(c => c.IsMucocyst))
```
hhyyrylainen, line 186: The slimejet detection system should instead be made to not consider mucocysts slime jets, as they are not slime jets at all.

src/microbe_stage/systems/CompoundAbsorptionSystem.cs
```diff
@@ -55,6 +55,13 @@ protected override void Update(float delta, in Entity entity)
             throw new NotImplementedException();
         }
 
+        if (entity.Has<MicrobeControl>())
+        {
+            // Microbe can't passively absorb compounds with mucocyst active
+            if (entity.Get<MicrobeControl>().State == MicrobeState.MucocystShield)
```
hhyyrylainen, line 61: It would be a ton cleaner if enabling mucocyst shield instead set the absorb speed to 0 and restored it later, that way there's less intertangling of the systems when absorption system won't need to know that mucocyst feature exists.

src/microbe_stage/systems/MicrobeReproductionSystem.cs
```diff
@@ -368,6 +368,13 @@ protected override void PostUpdate(float state)
             CalculateFreeCompoundsAndLimits(gameWorld!.WorldSettings, organelles.HexCount, false,
                 reproductionDelta);
 
+        if (entity.Has<MicrobeControl>())
+        {
+            // Microbe can't passively absorb compounds with mucocyst active
+            if (entity.Get<MicrobeControl>().State == MicrobeState.MucocystShield)
+                remainingFreeCompounds = 0;
+        }
+
```
hhyyrylainen, line 377: I think it is better to disallow all reproduction system running when in mucocyst mode than specifically here (also need to do the same change for multicellular growth system). Also if this system didn't use MicrobeControl before the attributes need t obe adjusted.

src/microbe_stage/organelle_components/SlimeJetComponent.cs
```diff
@@ -7,7 +7,10 @@
 /// </summary>
 public class SlimeJetComponent : IOrganelleComponent
 {
+    public bool IsMucocyst;
```
hhyyrylainen, line 10: This should be a property with private setter.

src/microbe_stage/MicrobeWorldSimulation.cs
```diff
@@ -216,6 +217,7 @@ public void Init(Node visualDisplayRoot, CompoundCloudSystem cloudSystem, IMicro
         osmoregulationAndHealingSystem = new OsmoregulationAndHealingSystem(EntitySystem, couldParallelize);
         pilusDamageSystem = new PilusDamageSystem(EntitySystem, couldParallelize);
         slimeSlowdownSystem = new SlimeSlowdownSystem(cloudSystem, EntitySystem, couldParallelize);
+        mucocystSystem = new MucocystSystem(EntitySystem, couldParallelize);
```
hhyyrylainen, line 220: Did you test the performance of parallelizing the mucocyst system? I ask because the mucocyst system will run on *each* cell on each game update so it could be more performant to use multiple threads to run through all the entities. Though I guess as the AI won't know how to turn on the mucocyst the system will be very branch preditor friendly so it might run as fast without threads.

src/microbe_stage/components/MicrobeControl.cs
CI09: Isn't it doing similiar (if all slime jets are mucocyst then return, 'cause all slime jets are not normal slime jets)?

src/microbe_stage/systems/MucocystSystem.cs
```diff
@@ -0,0 +1,59 @@
...
+///   Handles gaining immortality from mucocyst ability
+/// </summary>
+[With(typeof(MicrobeControl))]
+[With(typeof(Health))]
+[With(typeof(CompoundStorage))]
+[Without(typeof(AttachedToEntity))]
+[WritesToComponent(typeof(CellProperties))]
+[RunsAfter(typeof(OrganelleComponentFetchSystem))]
+[RunsBefore(typeof(MicrobeMovementSystem))]
+[RuntimeCost(7)]
```
hhyyrylainen, line 19: This seems a bit suspicious to me, did you profile the system and confirm it takes up 7x more time than just simple systems?

CI09: I'm still not proffesional at systems, how much would you give here?

src/microbe_stage/systems/MucocystSystem.cs
```diff
@@ -0,0 +1,59 @@
...
+                control.State = MicrobeState.Normal;
+
+            var membrane = entity.Get<CellProperties>().CreatedMembrane;
+
+            if (membrane != null)
+                membrane.SetMucocystEffectVisible(true);
+        }
+        else
+        {
+            entity.Get<Health>().Invulnerable = CheatManager.GodMode;
```
hhyyrylainen, line 52: Doesn't this set AI cells to be invulnerable as well?

Also it is pretty non-optimal to constantly have to call methods on the membrane. Most systems are designed so that they need to at most like check a single bool flag to know if they need to process an entity or not.

There isn't a clear way to do that here, but it might be more performant to keep an internal list of active mucocyst effects in this system and then turning off the effects for cells that don't mark themselves as being still active in the update method.

src/microbe_stage/systems/MucocystSystem.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        if (control.State == MicrobeState.MucocystShield)
+        {
+            entity.Get<Health>().Invulnerable = true;
+            entity.Get<CompoundStorage>().Compounds.
+                Compounds[mucilageCompound] -= Math.Min(
+                Constants.MUCOCYST_MUCILAGE_DRAIN * delta,
+                entity.Get<CompoundStorage>().Compounds.Compounds[mucilageCompound]);
+
+            if (entity.Get<CompoundStorage>().Compounds.Compounds[mucilageCompound] <= 0)
```
hhyyrylainen, line 42: Calling `Get` multiple times is bad for performance as component look ups are expensive. Any components used multiple times should be stored in a local ref variable.

src/microbe_stage/components/MicrobeControl.cs
hhyyrylainen: Why are mucocysts placed in the list that is *only for slimejets*? That doesn't make sense to me and makes the whole code base design much more complex when the list may or may not contain something that doesn't match the variable name.

src/microbe_stage/systems/MucocystSystem.cs
hhyyrylainen: I would need to profile but I would *guess* that the performance impact of this system would be at most a 2 but most likely just 1 (so just removing this attribute) is the closest.

src/microbe_stage/systems/MicrobeReproductionSystem.cs
CI09: For muticellular, one cell might have mucocyst while other don't; is it necessary?

hhyyrylainen: The main cell does reproduction only in multicellular, and it will call this base reproduction process method.

shaders/MucocystEffect.gdshader
hhyyrylainen, line 1: This file has a mixture of tabs and spaces, also the spacing between operators is not consistent.

shaders/MucocystEffect.gdshader
```diff
@@ -0,0 +1,36 @@
...
+	
+	//VERTEX.y = 5.f;
+}
+
+void fragment() {
+    float outer_edge = UV.y - 0.45f;
+	float inner_edge = (0.5f - UV.y) - 0.1f;
+	
+    float mask = max(0.f, outer_edge*10.f);
+    float mask2 = max(0.f, inner_edge*(1.f+fade));
```
hhyyrylainen, line 32:
```suggestion
    float mask2 = max(0.f, inner_edge * (1.f + fade));
```
For example here

shaders/MucocystEffect.gdshader
```diff
@@ -0,0 +1,36 @@
...
+uniform float movementWigglyNess = 1.f;
+
+uniform float fade = 0.25f;
+uniform vec3 tint : source_color = vec3(0.40, 0.80, 0.60);
+
+void vertex(){
+    vec3 worldVertex = (MODEL_MATRIX * vec4(VERTEX, 1.0)).xyz;
+    float size = length(VERTEX);
+
+	//worldVertex /= 2.f;
```
hhyyrylainen, line 16: Unless Membrane.gdshader has these same commented out lines, these should be removed. Also the comment near the top says the positioning math must match exactly the one in Membrane otherwise this effect won't match the shape of the membrane, so double checking that this math matches would be good.

src/microbe_stage/MembraneShapeGenerator.cs
```diff
@@ -303,7 +303,7 @@ private static ArrayMesh BuildEngulfMesh(Vector2[] vertices2D, int vertexCount,
             indices[index] = index;
             indices[index + 1] = index + 1;
 
-            vertices[index] = new Vector3(sourceVertex.X, height / 2, sourceVertex.Y);
+            vertices[index] = new Vector3(0.5f, height / 2, 0.5f);
```
hhyyrylainen, line 306: This still impacts the engulf mesh effect, in that this should be totally incorrect and should break the mesh shape, I don't see why this change should be a fine change. SOmeone will need to explain this to me for me to accept this.

src/microbe_stage/MicrobeWorldSimulation.cs
hhyyrylainen: This didn't get a reply yet.

src/microbe_stage/systems/MicrobeEmissionSystem.cs
```diff
@@ -242,6 +242,9 @@ private static Vector3 FacingDirection(in Entity entity, ref WorldPosition posit
             // Activate all jets, which will constantly secrete slime until we turn them off
             foreach (var jet in organelles.SlimeJets)
             {
+                if (jet.IsMucocyst)
+                    continue;
+
```
hhyyrylainen, line 247: This should not be a required change now.

src/microbe_stage/systems/MucocystSystem.cs
hhyyrylainen: This was also not addressed yet.

src/microbe_stage/systems/MicrobeReproductionSystem.cs
```diff
@@ -368,6 +368,13 @@ protected override void PostUpdate(float state)
             CalculateFreeCompoundsAndLimits(gameWorld!.WorldSettings, organelles.HexCount, false,
                 reproductionDelta);
 
+        if (entity.Has<MicrobeControl>())
+        {
+            // Microbe can't passively absorb compounds with mucocyst active
```
hhyyrylainen, line 373: This comment is inaccurate, should say that reproduction progress is halted.

src/microbe_stage/systems/MicrobeReproductionSystem.cs
```diff
@@ -368,6 +368,13 @@ protected override void PostUpdate(float state)
             CalculateFreeCompoundsAndLimits(gameWorld!.WorldSettings, organelles.HexCount, false,
                 reproductionDelta);
 
+        if (entity.Has<MicrobeControl>())
```
hhyyrylainen, line 371: This system needs `[ReadsComponent(typeof(MicrobeControl))]` attribute.  This skip should also be earlier in the system, on line 227, I'd say.

Also unless I missed it, the multicell system also needs this change.

src/microbe_stage/systems/MucocystSystem.cs
```diff
@@ -0,0 +1,60 @@
...
+                Compounds[mucilageCompound] -= Math.Min(
+                Constants.MUCOCYST_MUCILAGE_DRAIN * delta,
+                storage.Compounds.Compounds[mucilageCompound]);
+
+            if (storage.Compounds.Compounds[mucilageCompound] <= 0)
+                control.State = MicrobeState.Normal;
+
+            entity.Get<Health>().Invulnerable = true;
+
+            var membrane = entity.Get<CellProperties>().CreatedMembrane;
```
hhyyrylainen, line 47: As this writes to membrane this must have the runs on main thread attribute. And also as this just affects the mucocyst parameters it would be enough to mark this as running on entities with CellProperties but just in read mode:

```
[With(typeof(CellProperties))]
[ReadsComponent(typeof(CellProperties)]
```

src/microbe_stage/components/MicrobeControl.cs
```diff
@@ -195,4 +195,55 @@ public static void SetMoveSpeed(this ref MicrobeControl control, float speed)
...
+            if (compoundStorage.Compounds.Compounds[mucilage] < Constants.MUCOCYST_MINIMUM_MUCILAGE)
+                return;
+        }
+
+        if (state)
+        {
+            control.State = MicrobeState.MucocystShield;
+            var membrane = entity.Get<CellProperties>().CreatedMembrane;
+
+            membrane?.SetMucocystEffectVisible(true);
```
hhyyrylainen, line 234: This method should not directly modify the graphics like this, this goes against the ECS architecture. Instead this should just let the mucocyst system handle the graphics.

src/microbe_stage/components/MicrobeControl.cs
```diff
@@ -195,4 +195,55 @@ public static void SetMoveSpeed(this ref MicrobeControl control, float speed)
...
+                m.Get<MicrobeControl>()
+                    .SetMucocystState(ref m.Get<OrganelleContainer>(), m, state));
+        }
+
+        // It might be tricky as there is now mucocyst upgrade that doesn't emit slime normally
+        if (organelleInfo.Mucocysts == null || organelleInfo.Mucocysts.Count < 1)
+            return;
+
+        // Must have at least one mucocyst
+        if (organelleInfo.Mucocysts?.Count < 1)
```
hhyyrylainen, line 217: As mucocyst state is not really needed, for performance reasons the mucosyst shouldn't be a list, but just an int similarly to how `AgentVacuoleCount` is handled.

src/microbe_stage/components/MicrobeControl.cs
```diff
@@ -195,4 +195,55 @@ public static void SetMoveSpeed(this ref MicrobeControl control, float speed)
...
+
+        if (state)
+        {
+            control.State = MicrobeState.MucocystShield;
+            var membrane = entity.Get<CellProperties>().CreatedMembrane;
+
+            membrane?.SetMucocystEffectVisible(true);
+
+            entity.Get<CompoundAbsorber>().AbsorbSpeed = -1;
+            entity.Get<SoundEffectPlayer>().PlaySoundEffect("res://assets/sounds/soundeffects/microbe-slime-jet.ogg");
```
hhyyrylainen, line 237: These also should be handled by the mucocyst system. Also would be pretty nice to have a dedicated sound for mucocyst. Though if the sound is purely non-looping then it is fine to play it here, as long as any system calling (if there is any) SetMucocystState needs to have the right attributes to mark which components it accesses.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -395,8 +398,10 @@ protected override void UpdateAbilitiesHotBar()
 
         bool showToxin;
         bool showSlime;
+        bool showMucocyst = false;
```
hhyyrylainen, line 401:
```suggestion
        bool showMucocyst;
```
I think automatic checks should have complained about this...

simulation_parameters/Constants.cs
```diff
@@ -435,6 +435,18 @@ public static class Constants
     /// </summary>
     public const float MUCILAGE_COOLDOWN_TIMER = 1.5f;
 
+    /// <summary>
+    ///   How many times cell gets slowed down with mucocyst (slime jet upgrade) on
+    /// </summary>
+    public const float MUCOCYST_SPEED_DIVISOR = 60;
```
hhyyrylainen, line 441: Usually speed modifiers are multipliers instead of divisors as dividing is a much more expensive floating point operation than multiplying.

src/microbe_stage/Membrane.cs
```diff
@@ -170,6 +176,8 @@ public override void _Ready()
 
     public override void _Process(double delta)
     {
+        GD.Print(mucocystAnimationMeshInstance.Visible);
```
hhyyrylainen, line 179: Debug print

src/microbe_stage/Membrane.cs
```diff
@@ -294,6 +307,9 @@ private void SetMesh()
 
         engulfAnimationMeshInstance.Mesh = membraneData.GeneratedEngulfMesh;
         engulfAnimationMeshInstance.MaterialOverride = EngulfShaderMaterial;
+
+        mucocystAnimationMeshInstance.Mesh = membraneData.GeneratedEngulfMesh;
```
hhyyrylainen, line 311: If the mucocyst due to reverting mesh generation changes cannot use the same mesh as engulfing, then it can generate a separate mesh for it, but I'd obviously prefer if that wasn't required.

src/microbe_stage/MicrobeHUD.cs
hhyyrylainen, line 1: Doesn't the multicellular HUD also need these changes? As the base HUD now includes new stuff the multicellular HUD also need to implement those.

hhyyrylainen (changes requested):
This is now pretty much a full review except I didn't review the mucocyst system fully as I think it ought to be refactored quite thoroughly. I can help with that if all the other minor style and architecture nitpicks are solved first.

src/microbe_stage/components/MicrobeControl.cs
CI09: I would need help with this

hhyyrylainen: Just remove the code from here, as far as I remember reading the code the mucocyst system should already be setup to apply the visual effect to the membrane.

src/microbe_stage/Membrane.cs
CI09: I think it was easier to program as you can just turn on visibility or off.

src/microbe_stage/MicrobeHUD.cs
hhyyrylainen: Then the warning caught a bug, you aren't updating the mucocyst state for player who is in a colony.

src/microbe_stage/Membrane.cs
hhyyrylainen: I don't doubt that, but unless it gets wide support any graphics change that impacts the engulfing animation I won't be accepting. So either the mucocyst effect needs to work with the exact same mesh (like it is done here) or it needs to generate a fully different mesh that would then be assigned here.

src/late_multicellular_stage/MulticellularHUD.tscn
```diff
@@ -456,6 +481,17 @@ tooltip_text = "EJECT_ENGULFED_TOOLTIP"
 ActionIcon = ExtResource("33")
 ActionName = "g_eject_engulfed"
 
+[node name="Shield" parent="ScrollContainer/HotBar" instance=ExtResource("78")]
+visible = false
+layout_mode = 2
+tooltip_text = "MUCOCYST_ACTION_TOOLTIP"
+theme_override_styles/normal = SubResource("StyleBoxFlat_xmnai")
```
hhyyrylainen, line 488: Is there a reason why this button is setup differently from the other ones?

src/microbe_stage/components/OrganelleContainer.cs
```diff
@@ -43,6 +43,12 @@ public struct OrganelleContainer
     [JsonIgnore]
     public List<SlimeJetComponent>? SlimeJets;
 
+    /// <summary>
+    ///   The slime jets with mucocyst
+    /// </summary>
+    [JsonIgnore]
```
hhyyrylainen, line 49: This should not be json ignore and also this needs to be set among the int variables, otherwise struct padding will waste 4 bytes of space for each struct instance.

src/microbe_stage/components/OrganelleContainer.cs
```diff
@@ -588,6 +601,7 @@ public static void CalculateOrganelleLayoutStatistics(this ref OrganelleContaine
     /// </summary>
     public static void FetchLayoutOrganelleComponents(this ref OrganelleContainer container)
     {
+        container.MucocystCount = 0;
```
hhyyrylainen, line 604: Setup for this variable should be like for the agent vacuole count rather than the slime jets list. (so this is the wrong method to initialize this).

hhyyrylainen (approved):
Took quite a while but I basically managed to refactor the entire mucocyst system. I ended up needing a state applied variable to be added to an existing component, which is not optimal in terms of component byte packing, but a new way of saving state turned out to be way unfitting with the current saving system.

I'll open a bunch of follow up issues to polish up the mucocyst system as I noticed quite many deficiencies and one bug.

PR #5110, opened 2024-05-09
Add logic for upgrade-relevant models

### LINKED ISSUE #4106: Add feature for general upgrades to replace organelle graphics scene
so that upgraded organelles can be visually distinguished from other ones.
Only slight complication with this is that in the editor this should apply correctly, other than that this is should be pretty easy to just override the graphics scene that is to be loaded based on the specified upgrades.

Edit: in the future this will likely be quite heavily used by organelle upgrades.

CI09:

Now upgrades can have `"ScenePath": ` in organelles.json to have upgrade-specific models.

**Related Issues**

Closes #4106

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09:
Appearance tab has problems with updating though...

CI09:
It's ready for review.

HexapodPhilosopher:
I tried to run it from Godot, but it failed, saying: "Semicolon after accessor block is not valid".
Deleting the semicolon at the end of the line "public string? ModelPath { get; private set; };" Allowed it to run normally.
Adding the line " "ModelPath": "res://assets/models/organelles/Flagellum.tscn" " to organelles.json set it so that upgrading the pilus to an injectisome made it look like a fully-animated flagellum.

src/microbe_stage/editor/CellEditorComponent.Callbacks.cs
```diff
@@ -294,8 +294,9 @@ private void DoOrganelleUpgradeAction(OrganelleUpgradeActionData data)
     {
         data.UpgradedOrganelle.Upgrades = data.NewUpgrades;
 
-        // Uncomment when upgrades can visually affect the cell
-        // UpdateAlreadyPlacedVisuals();
+        microbeVisualizationOrganellePositionsAreDirty = true;
```
GameDungeon, line 297: It might be good to set this to true only when the upgraded organelle has a new model path.

CI09: That might be a bit more complicated; it would need to check out of all available upgrades for the organelle any that are being currently applied and have model path.

hhyyrylainen: I think this should just call the `OnOrganellesChanged` method as upgrades will eventually be able to impact most things. I think especially directly setting `microbeVisualizationOrganellePositionsAreDirty` to true should be avoided here as that feels hacky.

CI09: Without setting `microbeVisualizationOrganellePositionsAreDirty` to true, game doesn't always update visuals even if upgrade changed models.

hhyyrylainen: Okay, I checked the code and it looks like the organelle add actions also set that to true, and so yeah turns out that is the intended usage, and using that'll be fine here after all. Still I'd like to see call to OnOrganellesChanged to prepare for the future where organelle upgrades might affect things like species survivability in auto-evo.

CI09: I think it does 2 lines below

hhyyrylainen (changes requested):

src/microbe_stage/editor/CellEditorComponent.Callbacks.cs
```diff
@@ -294,8 +294,9 @@ private void DoOrganelleUpgradeAction(OrganelleUpgradeActionData data)
...
+
+        OnOrganellesChanged();
```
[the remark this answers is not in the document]
hhyyrylainen: There's a method called `UndoOrganelleUpgradeAction` a few lines below this, which also has:
```
        // Uncomment when upgrades can visually affect the cell
        // UpdateAlreadyPlacedVisuals();
```
so very much yes.

src/microbe_stage/editor/CellEditorComponent.Callbacks.cs
```diff
@@ -294,8 +294,9 @@ private void DoOrganelleUpgradeAction(OrganelleUpgradeActionData data)
...
 
         UpdateStats();
```
hhyyrylainen, line 301: Now that `OnOrganellesChanged` is called from here, I think the update stats and StartAutoEvoPrediction calls need to be removed as they are basically duplicates as organelles changing should already trigger those actions.

CI09:
I renamed ModelPath to ScenePath because model path now is equivalent to model **inside** the scene

hhyyrylainen:
I think that is a good change to make this more uniformly named compared to how the organelle definition is setup.

If I had had futuresight when designing the organelle visuals system, I would have made a struct that holds the scene path, model path, and animation player path all in one as they are very closely related.

CI09:
I changed it now that upgrades' model/animation paths naming are the same as organelles' so it can be related easier (I hope it won't conflict)

Now, for example, to make slime jet visuals for pilus you have to add
`"ScenePath": "res://assets/models/organelles/SlimeJet.tscn",`
`"DisplaySceneModelPath": "Armature/Skeleton3D/Cube",`
`"DisplaySceneAnimation": "AnimationPlayer"`
to the upgrade in organelles.json.

What I am concerned about though, is the second commit (https://github.com/Revolutionary-Games/Thrive/pull/5110/commits/4e8c7d7ea0700db1de04706e7d56e535db1840d4), I don't know if it's good to do it the way like I did, but it worked for me and I am open for criticism.

hhyyrylainen:
> What I am concerned about though, is the second commit (https://github.com/Revolutionary-Games/Thrive/commit/4e8c7d7ea0700db1de04706e7d56e535db1840d4), I don't know if it's good to do it the way like I did, but it worked for me and I am open for criticism.

I don't know why you added that second part of that commit. It seems overtly complicated and I don't even understand why that code is necessary (this is a case where comments are really necessary to explain the reasoning behind the code).

hhyyrylainen:
Actually on second read that code now always creates a new node visuals for an organelle, even when it already had visuals. This creates a bunch of nodes that are never attached to the scene tree causing a resource leak, not to mention the performance impact of re-creating a ton of unnecessary organelle visuals.

CI09:
Could someone help with CI check failure though?

src/microbe_stage/AvailableUpgrade.cs
```diff
@@ -7,7 +7,15 @@
 /// </summary>
 public class AvailableUpgrade : IRegistryType
 {
+    private LoadedSceneWithModelInfo loadedSceneData;
+
 #pragma warning disable 169,649 // Used through reflection
+    /// <summary>
+    ///   A path to a scene to display this organelle with. If empty won't have a display model.
```
hhyyrylainen, line 14: I don't think this comment matches what the variable actually does in an upgrade.

src/microbe_stage/AvailableUpgrade.cs
```diff
@@ -7,7 +7,15 @@
...
+    /// </summary>
+    [JsonProperty]
+    private SceneWithModelInfo graphics;
```
hhyyrylainen, line 17:
```suggestion
    private SceneWithModelInfo overrideGraphics;
```

To be clear what the JSON variable does when it is in an upgrade.

src/microbe_stage/AvailableUpgrade.cs
```diff
@@ -75,6 +83,19 @@ public void Resolve()
...
+        // Preload the scene for instantiating in microbes
+        // TODO: switch this to only load when loading the microbe stage to not load this in the future when we have
+        // playable stages that don't need these graphics
+        if (!string.IsNullOrEmpty(graphics.ScenePath))
+        {
+            loadedSceneData.LoadFrom(graphics);
+        }
+    }
+
+    public LoadedSceneWithModelInfo? TryGetGraphicsScene()
```
hhyyrylainen, line 96: This cannot return a nullable value as that return type is a struct, which will cause boxing in this case. That means this has to be written like the `TryGet` method I wrote for the OrganelleDefinition.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1861,7 +1861,7 @@ private void RenderHighlightedOrganelle(int q, int r, int rotation, OrganelleDef
         bool showModel = !hadDuplicate;
 
         // Model
-        if (showModel && shownOrganelle.TryGetGraphicsScene(out var modelInfo))
+        if (showModel && shownOrganelle.TryGetGraphicsScene(null, out var modelInfo))
```
hhyyrylainen, line 1864: When moving an organelle, shouldn't this method take in a parameter specifying the organelle upgrades so that the visuals are consistent?

src/microbe_stage/editor/CellEditorComponent.Callbacks.cs
```diff
@@ -294,8 +294,9 @@ private void DoOrganelleUpgradeAction(OrganelleUpgradeActionData data)
...
```
hhyyrylainen, line 302: The line for `StartAutoEvoPrediction` should be deleted as `OnOrganellesChanged` already calls that. This same thing is in the undo method.

src/microbe_stage/editor/CellEditorComponent.Callbacks.cs
```diff
@@ -308,8 +309,9 @@ private void UndoOrganelleUpgradeAction(OrganelleUpgradeActionData data)
     {
         data.UpgradedOrganelle.Upgrades = data.OldUpgrades;
 
-        // Uncomment when upgrades can visually affect the cell
-        // UpdateAlreadyPlacedVisuals();
+        microbeVisualizationOrganellePositionsAreDirty = true;
+
+        OnOrganellesChanged();
 
         UpdateStats();
```
hhyyrylainen, line 316: Also the `UpdateStats` call should be removed from here and the do method as organelles changed already does that.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -276,31 +276,31 @@ public enum OrganelleGroup
     ///   boxing)
     /// </para>
     /// <returns>True when this has a scene</returns>
-    public bool TryGetGraphicsScene(out LoadedSceneWithModelInfo modelInfo)
+    public bool TryGetGraphicsScene(OrganelleUpgrades? upgrades, out LoadedSceneWithModelInfo modelInfo)
```
hhyyrylainen, line 279: XML doc is missing documentation for the upgrades parameter.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -657,6 +657,30 @@ private Vector3 CalculateCenterOffset()
         return offset;
     }
 
+    private LoadedSceneWithModelInfo? TryGetGraphicsForUpgrade(OrganelleUpgrades? upgrades)
```
hhyyrylainen, line 660: This has the boxing problem as well.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -657,6 +657,30 @@ private Vector3 CalculateCenterOffset()
...
+    {
+        if (upgrades != null)
```
hhyyrylainen, line 662: I'd swap this the otherway around to early return if upgrades is null.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -657,6 +657,30 @@ private Vector3 CalculateCenterOffset()
...
+        {
+            foreach (var availableUpgrade in AvailableUpgrades)
+            {
+                var upgradeGraphics = availableUpgrade.Value.TryGetGraphicsScene();
+
+                if (upgradeGraphics != null)
+                {
+                    if (upgrades.UnlockedFeatures.Contains(availableUpgrade.Key))
```
hhyyrylainen, line 670: I'd write this the other way around: check if an upgrade is present before doing the logic for fetching its scene.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -657,6 +657,30 @@ private Vector3 CalculateCenterOffset()
...
+                    {
+                        if (availableUpgrade.Value.TryGetGraphicsScene().HasValue)
+                        {
+                            return availableUpgrade.Value.TryGetGraphicsScene();
```
hhyyrylainen, line 674: Here, refactoring the TryGet method in the upgrade to directly return the data like TryGetGraphicsScene would allow simplifying this and removing the duplicate method call (or actually considering line 666 there's actually 3 calls to the same method here).

hhyyrylainen (changes requested):

src/microbe_stage/editor/CellEditorComponent.cs
CI09: Upgrades are available from PlacedOrganelle and there is not much of it available for this line (`shownOrganelle` is OrganelleDefinition)

src/microbe_stage/OrganelleDefinition.cs
CI09: Now only the thing I commented on somewhere above and this (if it's necesary) yet to go.

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: When moving an organelle, the variable `MovingPlacedHex` is of type `OrganelleTemplate` which contains upgrade data. That's why I made that comment. So updating this method to allow passing in upgrade data would hopefully make it relatively simple to pass those upgrades in when moving a hex.

src/microbe_stage/OrganelleDefinition.cs
hhyyrylainen: The XML documentation should be something like `Applied upgrades for this organelle. This is required as some upgrades alter organelle visuals`. Also XML elements that fit on a single line should be placed on a single line.

src/microbe_stage/AvailableUpgrade.cs
```diff
@@ -7,7 +7,15 @@
 /// </summary>
 public class AvailableUpgrade : IRegistryType
 {
+    private LoadedSceneWithModelInfo loadedSceneData;
+
 #pragma warning disable 169,649 // Used through reflection
+    /// <summary>
+    ///   A path to a scene to override organelle's display scene. If empty won't have a display model.
```
hhyyrylainen, line 14: > If empty won't have a display model.

Isn't this wrong or very annoying? Because upgrades that don't specify visual changes should just use the normal organelle graphics.

src/microbe_stage/AvailableUpgrade.cs
```diff
@@ -75,6 +83,22 @@ public void Resolve()
...
+        // Preload the scene for instantiating in microbes
+        // TODO: switch this to only load when loading the microbe stage to not load this in the future when we have
+        // playable stages that don't need these graphics
+        if (!string.IsNullOrEmpty(overrideGraphics.ScenePath))
+        {
+            loadedSceneData.LoadFrom(overrideGraphics);
+        }
+    }
+
+    public LoadedSceneWithModelInfo TryGetGraphicsScene()
```
hhyyrylainen, line 96:
```suggestion
    public bool TryGetGraphicsScene(out LoadedSceneWithModelInfo overrideGraphics)
```

The signature should be like this to match the other TryGet methods I added, and also to follow the convention set by the C# standard library with its TryGet methods.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -275,10 +275,19 @@ public enum OrganelleGroup
     ///   The model info returned like this (as it may be a struct type this can't return a nullable reference without
     ///   boxing)
     /// </para>
+    /// <para cref="upgrades">
```
hhyyrylainen, line 278:
```suggestion
    /// <param cref="upgrades">
```

Typo in the XML tag.

CI09: When I do as you suggested, my IDE complains that it is lacking a name. I tried replacing cref= with name=, it stopped complaining but nothing really changed when I hovered my mouse above the function.

hhyyrylainen: The right syntax is:
```c#
/// <param name="useDefaults">If true default checks are added automatically</param>
```
(just a random example I grabbed from Thrive).

CI09:
Might be good now.

src/microbe_stage/AvailableUpgrade.cs
```diff
@@ -7,7 +7,15 @@
...
 public class AvailableUpgrade : IRegistryType
 {
+    private LoadedSceneWithModelInfo loadedSceneData;
+
 #pragma warning disable 169,649 // Used through reflection
+    /// <summary>
+    ///   A path to a scene to override organelle's display scene. If empty will use organelle's default model.
+    /// </summary>
+    [JsonProperty]
+    private SceneWithModelInfo overrideGraphics;
```
hhyyrylainen, line 17: Does this have to be inside the pragma warning disable? If not absolutely required, then this should be outside the pragma block.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -271,14 +271,23 @@ public enum OrganelleGroup
     /// <summary>
     ///   Gets the visual scene that should be used to represent this organelle (if there is one)
     /// </summary>
-    /// <para cref="modelInfo">
+    /// <param name="upgrades">
+    ///   Some upgrades alter organelle visuals
+    /// </param>
```
hhyyrylainen, line 276:
```suggestion
    /// <param name="upgrades">Some upgrades alter organelle visuals</param>
```
This fits on a single line.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -271,14 +271,23 @@ public enum OrganelleGroup
...
     ///   The model info returned like this (as it may be a struct type this can't return a nullable reference without
     ///   boxing)
-    /// </para>
+    /// </param>
     /// <returns>True when this has a scene</returns>
-    public bool TryGetGraphicsScene(out LoadedSceneWithModelInfo modelInfo)
+    public bool TryGetGraphicsScene(OrganelleUpgrades? upgrades, out LoadedSceneWithModelInfo modelInfo)
     {
-        // TODO: allow this to be affected by upgrades (add as a first parameter to this method)
+        var hasUpgradeGraphics = TryGetGraphicsForUpgrade(upgrades, out var upgradeGraphics);
```
hhyyrylainen, line 284:
```suggestion
        var hasUpgradeGraphics = TryGetGraphicsForUpgrade(upgrades, out modelInfo);
```

This can directly write to the out variable of this method to save one data copy.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -271,14 +271,23 @@ public enum OrganelleGroup
...
+
+        if (hasUpgradeGraphics)
```
hhyyrylainen, line 286: This doesn't need a separate bool so I'd recommend something like:

```c#
if (TryGetGraphicsForUpgrade(upgrades, out modelInfo)
{
     return true;
}
```

then also the line `modelInfo = default(LoadedSceneWithModelInfo);` should be removable to avoid one more data copy.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -290,9 +299,15 @@ public bool TryGetGraphicsScene(out LoadedSceneWithModelInfo modelInfo)
         return true;
     }
 
-    public bool TryGetCorpseChunkGraphics(out LoadedSceneWithModelInfo modelInfo)
+    public bool TryGetCorpseChunkGraphics(OrganelleUpgrades? upgrades, out LoadedSceneWithModelInfo modelInfo)
     {
-        // TODO: allow this to be affected by upgrades (add as a first parameter to this method)
+        var hasUpgradeGraphics = TryGetGraphicsForUpgrade(upgrades, out var upgradeGraphics);
```
hhyyrylainen, line 304: This part of this method should have the changes applied I mentioned above for the very similar block of code.

hhyyrylainen (changes requested):
I think the architecture approach and other code is good now, I just want some improvements still to the struct data copy handling and removing some unnecessary intermediate bool values.

CI09:
If it's still to be tested, here's example object you can insert into upgrade in `organelles.json`:

`"OverrideGraphics": {`
`  "ScenePath": "res://assets/models/organelles/Rusticyanin.tscn",`
`  "ModelPath": "Armature/Skeleton3D/Cube",`
`  "AnimationPlayerPath": "AnimationPlayer"`
`}`

src/microbe_stage/OrganelleDefinition.cs
hhyyrylainen: No, just remove the line entirely, it should compile without it.

hhyyrylainen (approved):
I did ever so slight changes and verified my last comments require no further actions. I didn't test the actual override graphics feature, but that can wait until we have a completed model and are adding it in a PR.

PR #5122, opened 2024-05-14
Fixed fluid current system force not applying

hhyyrylainen:
the force might still be too weak to be noticeable currently, but it should at least be applied right now

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


wabuilderman:
I've tested it, and it seems to work. It is interesting that I have to scale the force by about a factor of 1000 for any noticeable movement, however.

hhyyrylainen:
I probably missed this system not working back when converting stuff to ECS. That number actually makes sense because before that the masses of things were around 1, but with Jolt the default density of things is 1000 so most forces needed to be adjusted by about a factor of about 1000 to get them back to working.

PR #5123, opened 2024-05-14
Endosymbiosis lite

### LINKED ISSUE #3936: Implement endosymbiosis lite
endosymbiosis lite is a system whereby engulfing prokaryotes the player can unlock organelles based on what the eaten bacteria had. On top of the base functionality there should be at least some tutorial somewhere letting the player know that this is possible. Maybe triggered when the player engulfs a cell for the first time?

Related to unlocking organelles: https://github.com/Revolutionary-Games/Thrive/issues/3935

hhyyrylainen:

I'm finally doing the endosymbiosis feature: https://forum.revolutionarygamesstudio.com/t/upgrades-unlocks-endosymbiosis-master-thread/915/90

**Related Issues**

closes #3936

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/EndosymbiosisData.cs
+++ b/src/general/EndosymbiosisData.cs
@@ -101,7 +101,7 @@ public class EndosymbiosisData
         return true;
     }
 
-    public InProgressEndosymbiosis MarkEndosymbiosisDone()
+    public InProgressEndosymbiosis MarkEndosymbiosisDone(InProgressEndosymbiosis endosymbiosisToComplete)
     {
         if (StartedEndosymbiosis == null)
             throw new InvalidOperationException("No in-progress endosymbiosis");
@@ -110,6 +110,9 @@ public class EndosymbiosisData
 
         if (!endosymbiosis.IsComplete)
             GD.PrintErr("Marking endosymbiosis that is not complete as done");
+
+        if (endosymbiosis != endosymbiosisToComplete)
+            GD.PrintErr("Wrong endosymbiosis object passed to marking it as complete");
 
         StartedEndosymbiosis = null;
 
--- a/src/general/EndosymbiosisData.cs
+++ b/src/general/EndosymbiosisData.cs
@@ -68,6 +68,39 @@ public class EndosymbiosisData
         return true;
     }
 
+    public bool StartEndosymbiosis(int targetSpecies, OrganelleDefinition organelle, int cost)
+    {
+        if (StartedEndosymbiosis != null)
+        {
+            GD.PrintErr("Endosymbiosis is already in progress");
+            return false;
+        }
+
+        if (cost < 1)
+            throw new ArgumentException("Cost should be at least one for endosymbiosis");
+
+        // Need to resolve the species before can start
+        foreach (var candidate in EngulfedSpecies)
+        {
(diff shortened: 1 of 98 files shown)


CI09:
Even though I used cheats to speed up testing, it seemed to be a really fun feature to play with.
When endosymbiosis is done, the candidate species wasn't present in the patch anymore (maybe intentional).
Also it takes 5 editor enters to fully complete the endosymbiosis process (at least for me); it can be consulted with some theorist how long did endosymbiosis take (if it is known) on Earth and just do some number tweak.
Very cool feature either way 👏

src/general/EndosymbiosisData.cs
```diff
@@ -0,0 +1,274 @@
...
+                {
+                    if (endosymbiont.OriginallyFromSpecies.ID == targetSpecies)
+                    {
+                        GD.PrintErr("Trying to form symbiosis with something that is already an endosymbiont");
+                        return false;
+                    }
+                }
+            }
+
+            StartedEndosymbiosis = new InProgressEndosymbiosis(candidate.Key, cost, organelle);
```
CI09, line 100: I'm probably wrong, but this is never discarded, so...

hhyyrylainen:
> the candidate species wasn't present in the patch anymore (maybe intentional).

This would be a pretty tricky GUI part, so I decided that this will be opened as an issue to be done later (or at least highlight species not present in current patch).

> Also it takes 5 editor enters to fully complete the endosymbiosis process (at least for me);

The time it takes depends on how many organelles that can be turned into the endosymbiote there are in the species, so a species with a bunch of metabolosomes can be turned into a mitochondrion in just 2 cycles (so the potential range is 2-5 editor cycles).

src/general/EndosymbiosisData.cs
hhyyrylainen: What do you mean by discarded? This value is stored in the property and this method next returns true so the value will stay there until cancelled or completed.

CI09:
This would be a pretty tricky GUI part, so I decided that this will be opened as an issue to be done later (or at least highlight species not present in current patch).

I mean, after I finished endosymbiosis, endosymbiont wasn't in patch anymore (like if it has gone extinct).

hhyyrylainen:
> I mean, after I finished endosymbiosis, endosymbiont wasn't in patch anymore (like if it has gone extinct).

That's probably just a random chance it happened. There's no logic to delete a species that becomes an endosymbiote. It doesn't affect the target species at all that someone made an endosymbiote from it.

src/general/EndosymbiosisData.cs
CI09: > What do you mean by discarded?
Disposed. You know, it was created with `new` but isn't destroyed anywhere (I saw that `StartedEndosymbiosis` is being set to null somewhere though). But as I said, I am probably wrong here...

hhyyrylainen: `InProgressEndosymbiosis` is not an `IDisposable` so it is not possible to even dispose it. As it is a plain C# class that just uses memory and references other similar objects, the garbage collector takes care of deallocating it when it is no longer used (which is kind of the point with garbage collected programming languages like C# versus C++).

PR #5125, opened 2024-05-16
Add the ferroplast organelle, and remodel rusticyanin

HexapodPhilosopher:

Adds in models, icon and settings for the ferroplast - a non-LAWK, eukaryotic organelle that is a more efficient equivalent of rusticyanin. Also remodels the rusticyanin to resemble the real-life protein.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


locale/en.po
hhyyrylainen, line 1: It looks like you have updated just one locale file. All the locale files must be updated to have all of the translation keys (there's the localization script which does this automatically).

assets/models/organelles/Ferroplast.tscn
hhyyrylainen, line 1: If I'm reading this correctly, the ferroplast is made up of multiple parts. This is not recommended as that requires the use of `OrganelleMeshWithChildren` to work correctly. So I'd consider if it is possible to merge this into a single model (if not then this needs the same approach setup as other models that use `OrganelleMeshWithChildren` at least the nucleus does this).

HexapodPhilosopher: I used a separate texture for the membrane, so I needed a separate model. I did make the iron chunks a separate model, so they could move separately. I'll add OrganelleMeshWithChildren and set them to use the same shader.

locale/en.po
HexapodPhilosopher: So, what do i need to put in there? Just add the full msgid line, and leave the msgstr blank? e.g.
msgid "FERROPLAST"
msgstr

hhyyrylainen: Run the localization script `dotnet run --project Scripts -- localization`. https://github.com/Revolutionary-Games/Thrive/blob/master/doc/working_with_translations.md#updating-the-localizations

assets/models/organelles/Ferroplast.tscn
hhyyrylainen: I guess that's fine, I just don't want there to be too many graphically intense organelles (as with 100+ cells on screen where each cell might have up to like 10 copies, it quickly adds up a bunch of extra drawcalls if there are organelles that require multiple drawcalls). The nucleus is a bit of a special case as you can have only one of them per cell. But I guess this is fine as there won't be that many new organelles added anymore.

simulation_parameters/microbe_stage/organelles.json
```diff
@@ -1098,5 +1098,57 @@
...
+      }
+    },
+    "Enzymes": {
+      "rusticyanin": 1
+    },
+    "ProkaryoteChance": 0,
+    "ChanceToCreate": 0.5,
+    "Density": 1100,
+    "DisplayScene": "res://assets/models/organelles/Ferroplast.tscn",
+    "Name": "Ferroplast",
```
hhyyrylainen, line 1133: I think this line is where the CI failure comes from, the name here needs to be a translation key.

simulation_parameters/microbe_stage/organelles.json
```diff
@@ -1098,5 +1098,57 @@
...
+    "ProductionColour": "#3293f7",
+    "ConsumptionColour": "#f04681",
+    "IconPath": "res://assets/textures/gui/bevel/parts/FerroplastIcon.png",
+    "UnlockConditions": [
+      {
+        "AtpProductionAbove": {
+          "Atp": 25
+        },
```
hhyyrylainen, line 1141: Why is there this kind of condition? What's the reasoning why ferroplast is unlocked when you have a big cell producing a bunch of ATP?

assets/textures/gui/bevel/parts/RusticyaninIcon.png.import
```diff
@@ -21,7 +21,7 @@ compress/lossy_quality=0.7
 compress/hdr_compression=1
 compress/normal_map=0
 compress/channel_pack=0
-mipmaps/generate=true
+mipmaps/generate=false
```
hhyyrylainen, line 24: This needs to be kept on for the icon to look good.

simulation_parameters/microbe_stage/organelles.json
HexapodPhilosopher: I just had that in there because the mitochondrion also used it. I didn't know if it was worth having there or not.

hhyyrylainen: I'd remove this unlock condition. Also with endosymbiosis merged, you should mark the ferroplast as the upgraded version of rusticyanin so that it becomes available through endosymbiosis.

simulation_parameters/microbe_stage/organelles.json
HexapodPhilosopher: Yes, it copied that value into all the translation files, and then caused errors due to the lower case.

hhyyrylainen:
Was this closed on accident or why was this closed?

HexapodPhilosopher:
I've been having a nightmare trying to update the branch. I'm going to redo it on a different branch.

hhyyrylainen:
You don't need to close the PR to do that. Even if force pushing fails (which already can replace any and all commits in this PR), modifying the PR to use a different source branch is possible.

PR #5153, opened 2024-05-23
Fix leaving editor transition

### LINKED ISSUE #5145: Entering the game from freebuild editor for the first time briefly flashes the screen
I just noticed that when exiting the freebuild editor for the first time it looks like there's one frame where the fade in is cancelled and then begins again causing a visible bright flash when going from the freebuild editor to the microbe stage for the first time. There's been various bugs like this before so I think this is some kind of initialization order problem.

CI09:

There was call to add transition when returning from editor and in `StartGUIStageTransition`. I removed the first one and I think longer transition feels a little bit better. I didn't really dive much but I hope it doesn't break anything (I playtested and it works).

**Related Issues**

Closes #5145

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I feel like this is a non-optimal fix as the second parameter to the sequence should be a bool indicating if the game is returning from the editor or not. So this removes the only place where that bool parameter is true.

I think a better fix would be to make it so that `SetupStage` already would know if the game is returning from the editor , and then the call to start transition that is deleted here would be good. Maybe with a new property that indicates the stage will be returned to soon, if there isn't such a thing already?

hhyyrylainen:
This now has the problem that there's no transitions on subsequent returns from the editor. I think that maybe the correct approach would be to refactor the GUI transition start to only start a transition if there isn't one already in progress. This would make it easier to handle situations where some code wants to override the default transition.

hhyyrylainen (approved):
I think I found the underlying bug in TransitionManager that prevented the sequence cancellation from working. I've fixed that now. Any extra testing is welcome to make sure I didn't break anything else.

hhyyrylainen:
I'll assume everything is fine as no one said anything.

Issue #4632, opened 2023-12-08
Pilus shape sometimes doesn't match the visual position


hhyyrylainen:
I noticed when testing tha sometimes the pilus physics shape is slightly away from the visual position of the organelle's visuals.
![2023-12-08_11 47 17 2531](https://github.com/Revolutionary-Games/Thrive/assets/3796411/4b2a0d10-4cb9-4552-86b2-dcb630223a2b)

hhyyrylainen:
This might get fixed by: https://github.com/Revolutionary-Games/Thrive/pull/6993

hhyyrylainen:
This is a very old bug so I'll assume it's safe to close this with some stuff being done related to this. If this can still happen we can reopen the issue.

PR #5154, opened 2024-05-23
Add Injectisome Model

HexapodPhilosopher:

Adds a model for the injectisome pilus upgrade, removes unused detail on the perforator model, and replaces the pilus texture to use for both.

Pili

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09:
Seems you commited some of my commits from another branch (probably when merging from master). You can can create new **fresh** branch, cherry-pick *your* commit, delete this branch and you can rename the newly created branch to the name of this one (so if you force-push it, it will overwrite this branch here).

CI09:
Sorry for bothering but models are a bit too small and too centered. Also inject pilus is kinda broken:
image
Left - New, Right - Old
My assumption is that someone has to fix it in .tscn files by scaling up and moving towards front (?)

hhyyrylainen:
Slight problem with new pilus would be that it might not match the physics shape anymore (the physics shape is a pretty narrow cylinder, and that cannot be easily customized, and won't sway like the organelles do visually). So I'd recommend changing the new model to be closer to the dimensions of the old one (and positioned the same way as before as that was the best compromise position I could find to get it to pretty neatly match up with the physics position both for small and big cells, https://github.com/Revolutionary-Games/Thrive/issues/4632).

HexapodPhilosopher:
For some reason, the texture isn't being applied, which is the cause of the appearance issue. They are both using the old texture. I didn't make the pilus thicker. They should both be entirely within the bounds of the physics shape, and at the right position in the cell. 
Here's how they appear for me (old on the right:
Pilus-Positions

HexapodPhilosopher:
They are both using the old texture. I just tried replacing the new texture in my folder with the old one, and it showed exactly as it does in [CI09](https://github.com/CI09)'s screenshot. Only the back-right portion of the perforator and back portion of the injectisome are visible. The new texture has the perforator part higher up, which is why it is only visible on part of the model - it's mapped to the new texture. The only thing wrong in that image is the texture. The rest is a combination of perspective and the part waving from side to side.

I'll try to see what's going on. There may have been a problem with the PR. @CI09 what branch have you tested it on?

hhyyrylainen:
Oh right so the problem is in [CI09](https://github.com/CI09)'s screenshot? I see. If the new texture is with the same name as the old, this might be a case of running the game without reopening Godot editor to get it to re-import assets. Or if the build is from CI then this is a situation where CI cache numbers need t obe bumped to stop old cache usage.

hhyyrylainen:
The code looks fine to me. This is now waiting for playtesting confirming things work. At the latest I'll test this myself and merge next week if no one else has time before that.

HexapodPhilosopher:
I just added the icon, and the texture I forgot to update when I updated the model. Just to remove noticeable repetition on the base of the injector.

hhyyrylainen:
The icon is the one you showed on discord, right? I think it's good to go as long as it is 800x800.

Actually it looks like there is no .import file for that png. Godot should always create one when importing a file, and it should be committed otherwise the import settings are *not consistent*.

HexapodPhilosopher:
Okay. Import added now. The icon is the one I just showed on discord and it is 800 x 800.

hhyyrylainen:
Am I missing something but doesn't this PR need a change to the organelles.json file to specify the injectisome model to be used?

CI09:
I was thinking about doing it in another PR.

hhyyrylainen:
I guess that would be kind of fine but then the title should be changed to something like "Added model for injectisome but it isn't used at all yet". Also it is just a few lines in the JSON so there's like no reason why it couldn't be included in this PR.

hhyyrylainen:
That commit changing the JSON file looks fine to me on the surface.

hhyyrylainen (approved):
I did a quick test in game and it seems the model positioning is the same as before and I didn't see any graphical glitches and the override graphics worked fine. So I'll merge this now and if there happens to be any bugs that turn up later those can be fixed then.

PR #5160, opened 2024-05-27
Hardcore mode

CI09:

This PR adds hardcore mode - removes standard saving from game together with quicksaves and in place of auto-saves a save is performed that writes **into one save name selected upon save creation**.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


HexapodPhilosopher (commented):
No bugs found. It is lacking a save on returning to the main menu, though. I was able to revert a bit by using that.

hhyyrylainen:
I didn't check but this mode should also suppress the warning for exiting without saving as this should force the save when exiting so that warning is not incorrectly shown.

hhyyrylainen (changes requested):
This needs more bulletproofing against players exploiting the fact that the game doesn't save when exiting in this mode before I think this is a polished enough feature to include in the game.

HexapodPhilosopher:
Also, wouldn't it make sense to include the checkbox for this in the Difficulty settings, rather than Misc? It's where most people would look for it.

hhyyrylainen:
What's the current progress on this? Last I checked this was very brittle in terms of how well it saves and keeps the player from very easily cheating in the difficulty mode.

CI09:
I got stuck at saving upon leaving so help would be useful. It's not mandatory feature though so you know...

hhyyrylainen:
That's fine, I'll just take this off the list of PRs I need to review.

PR #5166, opened 2024-05-28
Replace ice shard and remove unused ice textures

### LINKED ISSUE #1270: Damaging ice chunks and the particles look the same
which might be a bit confusing

HexapodPhilosopher:

Replaces ice shard model with a snowflake-like ice crystal with a new texture and removes unused ice textures.

**Related Issues**
Closes #1270

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09:
You can put `Closes #1270` to make this PR close that issue 👍

simulation_parameters/microbe_stage/biomes.json
```diff
@@ -1861,7 +1861,7 @@
           "Radius": 1,
           "ChunkScale": 1,
           "PhysicsDensity": 4000,
-          "Size": 1000,
+          "Size": 3,
```
hhyyrylainen, line 1864: Why did you change the size? I think the size is 1000 here to prevent this being accidentally engulfed (if the engulfed property is accidentally enabled on this).

assets/textures/Iceshards.png.import
hhyyrylainen, line 1: Why was this renamed to iceshard**s**? Isn't this still a single object so this should follow the convention fo the file name not being plural?

hhyyrylainen (changes requested):
Tested in-game and everything seemed fine, I just have 2 technical side comments.

simulation_parameters/microbe_stage/biomes.json
hhyyrylainen: And I'll note that this "size" is really the engulfing size for engulfing calculations and impacts nothing else.

HexapodPhilosopher: Ah, okay. I thought the engulfing was intentional. It damages you, but then quickly dissolves. I can change it.

hhyyrylainen: The size was 1000 before as it was for gameplay purposes to prevent eating the ice. It could be changed but I'd want someone from the game design team to weigh-in on such a change (I personally think that preventing the player from eating the ice for no other purpose than to receive damage is better than allowing engulfing the ice).

assets/textures/Iceshards.png.import
HexapodPhilosopher: I was considering adding a second, noticeably different one. Again, I can change the name.

hhyyrylainen: Even in that case I'd expect the texture name to be singular. Only if the actual texture image contained multiple things, like for example like a particle texture might do, then I'd think it would be reason enough to break from the file naming convention and have an "unexpected" name for the file.

hhyyrylainen (approved):
I'll approve this without re-testing as I'll assume based on reading the diff that there likely wasn't any small mistake that slipped in the file rename.

PR #4607, opened 2023-12-01
Convert help menu to thriveopedia pages

ender-titan1:
**Brief Description of What This PR Does**

This PR adds new pages to the Thriveopedia to replace the old help menu.

**Related Issues**

Closes #4192

**Progress Checklist**

Note: before starting this checklist the PR should be marked as non-draft.

- [ ] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [ ] Initial code review passed (this and further items should not be checked by the PR author)
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author. Merging must follow our
[styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md#git).

PR #5176, opened 2024-06-03
Expand Thriveopedia

### LINKED ISSUE #3771: Make a common module for the scripts and game to both use to share data objects
The credits and translation progress objects (though might need to solve which json lib to use), also some other stuff like the current build info, could be shared between the game and the scripts to not have two copies of the same class.

This is a bit difficult before Godot 4 as I think the common module will need to use a really old runtime version to be compatible with mono.

### LINKED ISSUE #4192: Convert the help menu to Thriveopedia pages
and make the help buttons take the player directly to the relevant help page for the current part of the game.

https://forum.revolutionarygamesstudio.com/t/the-thriveopedia/930/30

> I think creating a help pages main category with one help page for each stage and each editor would be the best. Then the help button in the pause menu and bottom left could open directly the relevant Thriveopedia page.

The help pages for the stage and its relevant editor should cross reference each other so that hte player is more likely to discover both pages at once if they see at least one of them.

### LINKED ISSUE #5211: Thriveopedia page tree view control has wrong focused colour
When selecting a tree item and it is focused, it has a white outline around it:

This should be fixed to match the focused state of other elements (a specific blue colour outline)

ender-titan1:

This PR adds three new categories to the Thriveopedia:
- Stages 
- Mechanics
- and Development

It also ports over any notices on the wiki pages into the game (outdated, not yet implemented)

Finally, it adds more customizability when converting from HTML to BBCode and adds a utility to scrape infobox data from the wiki (this can be later used to make sure that the wiki is in synch with the game).

**Related Issues**

Closes #4192
Now also closes #3771

closes #5211

Supersedes #4607

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Should this be marked as a draft still? This seems quite unmergeable and the description notes some parts that are not done yet.

ender-titan1:
Yeah, looks like I forgot to set it to draft.

Though I would like to note that I originally intended to implement the infobox synching feature separately and noted a future use, but if it's more convenient i can get it done with this PR.

hhyyrylainen:
I guess that's fine to omit but I was worried that as you marked that help pages to thriveopedia as being superceded by this but didn't mention anywhere that in this PR's description that that part wasn't done and it would be pretty important considering if this is meant to replace that previous PR.

ender-titan1:
Oh, sorry for the confusion. This PR still adds pages which replace the role of the help menu (the mechanics category), just that since starting that task the scope of it has expanded to include a more general expansion of the thriveopedia.

I could add another top-level page called "Help" which tells the user where in the thriveopedia to look to find help and link to those pages.

P.S.
This PR is just a redo of #4607, as I couldn't get my branch to synch with the Godot 4 update on master.

ender-titan1:
Unless I'm missing something, this feature _should_ be ready for review.

Not that this PR doesn't have all the pages ready, and a separate PR will be needed later after the rest of the mechanics pages are created, but all the code is ready

src/thriveopedia/pages/wiki/organelle/OrganelleInfoBox.cs
hhyyrylainen, line 1: I think the path should be plural, so like `.../organelles/...`

simulation_parameters/common/wiki.json
```diff
@@ -48,7 +50,46 @@
           "SectionHeading": "WIKI_HEADING_SCIENTIFIC_BACKGROUND",
           "SectionBody": "WIKI_AXON_SCIENTIFIC_BACKGROUND"
         }
-      ]
+      ],
+      "InfoboxData": [
+        {
+          "InfoboxKey": "INFO_BOX_COST",
+          "InfoboxValue": "100"
```
hhyyrylainen, line 57: I think it would be better to reuse existing JSON keys that are translated as they fit well enough and that wouldn't require changing the translation extract. Something like `Name` should be fine and understandable. Perhaps `Description` could work for the value or maybe this could have a new property name that is added to the translation system, maybe something like `DisplayedValue`?

Scripts/WikiUpdater.cs
```diff
@@ -16,43 +17,112 @@
...
+        // Get root pages
+        var organellesRootTask = FetchRootPage(ORGANELLE_CATEGORY, "Organelles", cancellationToken);
+        var stagesRootTask = FetchRootPage(STAGES_CATEGORY, "Stages", cancellationToken);
+        var mechanicsRootTask = FetchRootPage(MECHANICS_CATEGORY, "Mechanics", cancellationToken);
+        var developmentRootTask = FetchRootPage(DEVELOPMENT_CATEGORY, "Development", cancellationToken);
+
+        // Get pages in categories
+        var organellesTask = FetchPagesFromCategory(ORGANELLE_CATEGORY, "Organelle", cancellationToken);
+        var stagesTask = FetchPagesFromCategory(STAGES_CATEGORY, "Stage", cancellationToken);
+        var mechanicsTask = FetchPagesFromCategory(MECHANICS_CATEGORY, "Mechainc", cancellationToken);
```
Oliveriver, line 86: Typo.

Scripts/WikiUpdater.cs
```diff
@@ -62,93 +132,225 @@ public async Task<bool> Run(CancellationToken cancellationToken)
...
-            var name = organelle.TextContent.Trim();
+            var name = page.TextContent.Trim();
             var url = $"https://wiki.revolutionarygamesstudio.com/wiki/{name.Replace(" ", "_")}";
 
-            ColourConsole.WriteInfoLine($"Found organelle {name}");
+            ColourConsole.WriteInfoLine($"Found {pageType} {name}");
 
             var body = (await HtmlReader.RetrieveHtmlDocument(url, cancellationToken)).Body!;
-            var internalName = body.QuerySelector("#info-box-internal-name")!.TextContent.Trim();
+            var internalName = textInfo.ToTitleCase(name).Replace(" ", string.Empty);
```
Oliveriver, line 190: Can this be moved down and integrated into a ternary with lines 200 to 203? I think it needs to be clearer for future maintenance that `internalName` isn't always the page name.

Scripts/WikiUpdater.cs
```diff
@@ -62,93 +132,225 @@ public async Task<bool> Run(CancellationToken cancellationToken)
...
+            var value = row.Children.Where(e => e.LocalName == "td").First();
+            var translatedKey = row.Children.Where(e => e.LocalName == "th").First().TextContent.Trim();
+            var id = value.Id;
+
+            if (id == null)
+                continue;
+
+            var textContent = value.TextContent.Trim();
+
+            if (textContent == internalName)
```
Oliveriver, line 333: Why is this skipped?

Scripts/WikiUpdater.cs
```diff
@@ -62,93 +132,225 @@ public async Task<bool> Run(CancellationToken cancellationToken)
...
+                continue;
+
+            // Format the found content for use in translation files
+            var untranlsatedKey = id.Replace("#", string.Empty).ToUpperInvariant().Replace("-", "_");
+            var untranslatedValue = textContent.ToUpperInvariant()
+                .Replace(' ', '_')
+                .Replace(",", "_COMMA")
```
Oliveriver, line 340: Missing an underscore in `_COMMA_`.

Scripts/WikiUpdater.cs
```diff
@@ -184,6 +386,13 @@ private List<Wiki.Page.Section> GetMainBodySections(IHtmlElement body)
                         .Select(l => $"[indent]—   {ConvertParagraphToBbcode(l)}[/indent]")
                         .Aggregate((a, b) => a + "\n" + b) + "\n\n";
                     break;
+                case "H3":
```
Oliveriver, line 389: If new tags are being parsed, you'll need to update the page styleguide. I suppose it might need a more general update for everything.

https://wiki.revolutionarygamesstudio.com/wiki/Thriveopedia

Scripts/WikiUpdater.cs
```diff
@@ -326,15 +547,29 @@ private string ConvertTextToBbcode(string paragraph)
             var translatedPage = page.TranslatedPage;
 
             // Translate page names
-            translationPairs.Add(untranslatedPage.Name, translatedPage.Name);
+            translationPairs.TryAdd(untranslatedPage.Name, translatedPage.Name);
```
Oliveriver, line 550: Why has this been changed to `TryAdd` now? Surely it's an error for the script to want to add two pages with the same name.

src/general/HelpScreen.cs
```diff
@@ -1,159 +0,0 @@
-﻿using System;
```
Oliveriver, line 1: Has all the content from the help screen been moved to the Thriveopedia? Just double checking as we definitely don't want to lose any.

src/thriveopedia/pages/SimpleWikiPage.cs
```diff
@@ -0,0 +1,9 @@
+﻿/// <summary>
+///   The bare minimum wiki page. Requires the parent to be set manually when created.
+/// </summary>
+public partial class SimpleWikiPage : ThriveopediaWikiPage
```
Oliveriver, line 4: Following the naming convention, this should be `ThriveopediaSimpleWikiPage`.

src/thriveopedia/pages/wiki/mechanics/MechanicLinkButton.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Godot;
+
+/// <summary>
+///   Simple way to link mechanic pages from the mechanics root page
+/// </summary>
+public partial class MechanicLinkButton : Button
```
Oliveriver, line 6: Is there a reason this is a separate class to `PageLinkButton`?

Oliveriver:
Very nicely done! A lot of sensible abstraction. I'm excited to see the Thriveopedia put to more use.

Scripts/WikiUpdater.cs
ender-titan1: I did this intentionally, as you always have a space after a comma, you'd always end up with `_COMMA__`, and technically could end up with a triple underscore if you did something weird. But I'll probably change the first argument to `", "` to avoid that instead.

src/general/HelpScreen.cs
ender-titan1: The `Mechanics` category is supposed to replace the help menu, and should even be more detailed in the end.

However, I didn't get around to creating most of the pages in that category yet, and the ones I _did_ create most likely have some style inconsistencies and missing links.

I would really appreciate if you looked over what I added in this category. https://wiki.revolutionarygamesstudio.com/wiki/Category:Mechanics

hhyyrylainen: I didn't check how the pages look in the game yet, but it would be extremely good to have a short summary equivalent to the current help menu with links to other wiki pages that then explain things more in-depth.

ender-titan1: That could work, the MechanicsRoot page functions as the stand-in for the help menu, and links to other pages. I could add a short general summary to the root page (though I don't know if it won't end up being too long).

Each mechanic page features a short description in the intro paragraphs, and then goes into each part of the mechanic in-depth in separate sections, or at least they should.

hhyyrylainen: I think there really needs to be a stage and each editor specific "help" page. Which would basically be most important information covered briefly. Otherwise the help button in the game doesn't have any sensible place to send the player to read stuff.

ender-titan1: In that case there could be a help category with each editor and stage, as well as linking to the mechanics category for specific mechanic explanations

hhyyrylainen: Pretty soon that'll require pages to be able to be in multiple places because it makes much more logical sense to put the stage specific help pages under their stage category so that it is next to the in-depth information and other information the player needs. When playing the game normally you won't want to read the industrial stage help page while on the microbe help page.

My suggestion for now is to put help pages under their stages as that is where they belong in terms of when the player is going to read them.

ender-titan1: What I could suggest, to avoid both extreme page nesting and providing too much information to the player in the current stage, is to have one big dropdown somewhere in the thriveopedia (in a place that is easily accessible) which will select certain stages and then display only the relevant pages. For example, when Microbe Stage is selected, only show the categories for Organelles, Mechanics, Editor Help and Stage Info. And similarly, when Industrial Stage is selected, only show pages for buildings, tech and hide the organelle category as that information is no longer relevant.

When opened, the thriveopedia would automatically select the current stage, links to pages outside the current stage would still open the page, but that page would not appear in the tree. 

I think this approach would minimise the tree clutter (as it is already becoming a problem in my experience). In this case the tree would look like:

- Wiki
  - Mechanics
    - Compounds, Patch Map, etc...
  - Stages
    - Microbe Stage
    - Microbe Stage Help 
  - Organelles
    - Mitochondria, Chloroplast, etc...

For the time being I can implement your suggestion.

hhyyrylainen: Ultimately I think the information should primarily be grouped by stages, this means that there needs to be like a mechanics category under each stage separately. A layout like that only really works so far as we only have one stage.

Other stages in the tree can start off collapsed to reduce the clutter. This is how a tree view usually works. It would be against tree view GUI conventions to have a separate control that might be hard to understand for hiding tree elements.

So to me the nesting should be like:

- Wiki
  - Microbe Stage
    - Quick Help (microbe)
    - Mechanics
      - Compounds
    - Organelles
      - Axon
  - Multicellular Stage
    - Quick Help (multicellular)  
    - Mechanics
       - Some stuff
    - Organs
    - Organelles (this would duplicate info in the microbe stage tree)

That way the in-game wiki would be organized like a real page tree that is organized around information the player needs at the same.

ender-titan1: Godot has a feature for tree elements that are also dropdowns. That would be the best compromise in my opinion. The dropdown would contain all the stages, and then would show only the pages in the respective stage, layed out in the same way you described. In this way, you wouldn't have duplicated pages, you'd only see what you need to see, all the while being integrated naturally into the tree

Deus-Codes:
Format looks very good.

Just wanted to make sure though, is this text displayed as intended? 
image

src/general/HelpScreen.cs
hhyyrylainen: I suppose that could work but there are also non-stage specific pages that would need to be always visible in the tree.

ender-titan1:
I tried to make it display code/formulas properly but it didn't work when I tried to do it. I should probably remove that HTML tag when converting to BBCode

hhyyrylainen:
I think I remember doing some work related to having a monospace font in Thrive, so those tags could at least use the font bbcode tag (or maybe there is a monospace / preformatted tag already as rich text labels have a theme property for monospace font) to display the text slightly differently.

ender-titan1:
That's what I tried to do, but I couldn't get it to work for me. Might try again.

Scripts/WikiUpdater.cs
ender-titan1: `internalName` is its own variable in `Page` and shouldn't be translated either.

src/thriveopedia/pages/wiki/mechanics/MechanicLinkButton.cs
hhyyrylainen: It'd make things clear if these two classes were renamed based on what they do rather than what they are used for.

hhyyrylainen:
I misremembered already having setup a shared project for Thrive and the scripts. I can do that next week if you'd have a problem with getting that setup (as it is needed to take care of one open item on the godot 4 board still).

ender-titan1:
I'll try adding the shared project myself, how should it be named?

hhyyrylainen:
I'm thinking something like `ThriveScriptsShared`? That's about the best I've come up with after knowing quite a long time about the need to name such a thing.

hhyyrylainen:
Is this ready for a new round of reviews?

hhyyrylainen:
I guess I'll need to take over this PR next week (unless someone else can finish this) so that we can resume updating the content from the wiki into the game.

hhyyrylainen:
Where did the stages root page go? It is partially in but doesn't really work at all... and there's a button that tries to open it.

hhyyrylainen:
Follow up issues I opened to improve things further:
- https://github.com/Revolutionary-Games/Thrive/issues/5507
- https://github.com/Revolutionary-Games/Thrive/issues/5508
- https://github.com/Revolutionary-Games/Thrive/issues/5509
- https://github.com/Revolutionary-Games/Thrive/issues/5510

Issue #4962, opened 2024-03-24
Properly fix main menu backgrounds (.glb files)


CI09:
Main menu models are broken due to conversion to Godot 4 and their .glb should be fixed. There is PR that temporarily fixes it but it can be reverted or something else if someone takes this task up: https://github.com/Revolutionary-Games/Thrive/pull/4961

PR #5203, opened 2024-06-11
Redo main menu scenes 2 and 3, fixing rock scaling and backgrounds

### LINKED ISSUE #5201: Main menu scene 2 placed rocks are invisible (regression, again)
Seems like the placed rocks in the MainMenuScene2.tscn are broken, again when I've just fixed them for the Godot 4 update to make them visible.

These are hopefully easily fixable just using the Godot Editor.

HexapodPhilosopher:

Updates 3D main menu scenes 2 and 3 (both of the underwater ones). Improves visibility of backgrounds, resizes rocks,
reduces bulk of the chimney models, and improves emitters and lighting.

**Related Issues**

closes #5201
closes [#4962](https://github.com/Revolutionary-Games/Thrive/issues/4962)

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I assume this is now ready for looking at?

assets/models/menu_backgrounds/MainMenuScene3.tscn
```diff
@@ -118,70 +123,507 @@ glow_hdr_threshold = 0.5
...
+normal_scale = 6.2
+normal_texture = ExtResource("4_nq5x3")
+disable_receive_shadows = true
+billboard_mode = 3
+particles_anim_h_frames = 1
+particles_anim_v_frames = 1
+particles_anim_loop = false
+
+[sub_resource type="Curve" id="Curve_qwc45"]
+resource_local_to_scene = true
```
hhyyrylainen, line 261: It seems a bit like you are overusing the local to scene checkbox. As far as I can tell these resources are only in this scene so checking the checkbox doesn't do anything, and it should only be used for resources that are *modified* during runtime. Otherwise it basically has no effect so it is an unnecessary non-default value that is set.

For unsharing resources set in the editor on multiple objects, you need to right click the resource and click "make unique"
As you haven't written any C# code to interact with these resources there's actually nothing that would modify them during runtime so `resource_local_to_scene = true` does nothing useful, and I think it might cause some de-optimization in the engine so I'd prefer these unnecessary usages of these to be removed.

hhyyrylainen (approved):
This is a good improvement, thanks for making it.

I noticed one minor thing that maybe no one else has seen but for me when rendering on my 4K screen the MainMenuBackground 2 particle emitters seem to be running at a lower framerate than the game itself and they look slightly distracting. Though this is pretty minor and not present in the other menu scene or when at a lower than 4K resolution.

PR #5242, opened 2024-06-20
Convert turn algorithm to not be so dependent on cursor position

### LINKED ISSUE #4764: Investigate potential math for cells to always turn at max speed even if cursor isn't behind
Now when the cursor is exactly behind the cell, that causes the highest rate of rotation. It was brought up that maybe it'd be nicer if the cell always turned at max speed (if the cell isn't already nearly pointing in the right direction to avoid overshoot and oscillation). I don't immediately know how the quaternion math could be updated to have a linear + step speed to accomplish this goal.

adQuid:

Changes calculations to pretend the cursor is 90 degrees to the left or right as appropriate, unless the cursor is facing very close to the front. Some turning constants were changed as well to adjust for this.

**Related Issues**

fixes #4764 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/systems/MicrobeMovementSystem.cs
```diff
@@ -83,12 +83,10 @@ protected override void Update(float delta, in Entity entity)
...
         lookVector.Y = 0;
 
-        var length = lookVector.Length();
-
-        if (length > MathUtils.EPSILON)
+        if (lookVector.Length() > MathUtils.EPSILON)
         {
             // Normalize vector when it has a length
-            lookVector /= length;
+            lookVector = lookVector.Normalized();
```
hhyyrylainen, line 89: Why is the length now calculated twice? (Normalization of a vector must calculate the length).

src/microbe_stage/systems/MicrobeMovementSystem.cs
```diff
@@ -97,6 +95,18 @@ protected override void Update(float delta, in Entity entity)
             lookVector = Vector3.Forward;
         }
 
+        // Simplify turns to 90 degrees to keep consistant turning speed
```
hhyyrylainen, line 98:
```suggestion
        // Simplify turns to 90 degrees to keep consistent turning speed
```
typo

src/microbe_stage/systems/MicrobeMovementSystem.cs
```diff
@@ -97,6 +95,18 @@ protected override void Update(float delta, in Entity entity)
...
+        if ((position.Rotation * Vector3.Forward).SignedAngleTo(lookVector, Vector3.Up)
```
hhyyrylainen, line 99: `(position.Rotation * Vector3.Forward).SignedAngleTo(lookVector, Vector3.Up)`

could be extracted as a variable to not have to calculate that number twice (I'm pointing this out as the movement system runs for each microbe on each game simulation update).

hhyyrylainen (changes requested):
I tried playtesting this a bit and in my experience small mouse movements are still fine but larger ones now cause a very noticeable "snap" once the target vector is far away enough. I feel like that snap doesn't feel good. I think instead this algorithm should just exaggerate the look vector with a (linear) scaling based on how far away the mouse is. That would still speed up the rotation without having to put the mouse really far away past the target rotation you want and without having a very noticeable snapping point in the rotation.

hhyyrylainen (changes requested):
I'm still getting pretty weird rotation results:

[2024-07-23 10-56-50.webm](https://github.com/user-attachments/assets/f1c4af4b-1323-44fe-bf64-c5efdb8c15e6)


That should be with the latest commit on this branch `5fb9a41fa31ab9d48a55e10c720c40443b40fe37`.

adQuid:
> I'm still getting pretty weird rotation results:
> 
>  2024-07-23.10-56-50.webm 
> That should be with the latest commit on this branch `5fb9a41fa31ab9d48a55e10c720c40443b40fe37`.

you were very correct; I had some issues with my trigonometry and covered it up with my smoothing threshold. I've now reworked it.

src/microbe_stage/systems/MicrobeMovementSystem.cs
```diff
@@ -82,13 +82,30 @@ protected override void Update(float delta, in Entity entity)
...
         var lookVector = control.LookAtPoint - position.Position;
         lookVector.Y = 0;
+        var lookVectorLength = lookVector.Length();
 
-        var length = lookVector.Length();
+        var turnAngle = (position.Rotation * Vector3.Forward).SignedAngleTo(lookVector, Vector3.Up);
+        var unsignedTurnAngle = Math.Abs(turnAngle);
 
-        if (length > MathUtils.EPSILON)
+        // linear function reaching 1 at CELL_TURN_INFLECTION_RADIANS
```
hhyyrylainen, line 90:
```suggestion
        // Linear function reaching 1 at CELL_TURN_INFLECTION_RADIANS
```

hhyyrylainen (approved):
I can't say I pretend to understand the math, but in game testing results no longer show any weirdness and it feel good (I'm sure after getting used to this for a bit I'll wonder why rotation was ever slower).

PR #5244, opened 2024-06-20
Made KeyPrompt queue a refresh instead of doing one immediately

### LINKED ISSUE #1635: Make key prompts queue refresh
instead of immediately loading the icon. This should save some resources when there are a lot of key prompts.
So when they detect a change they mark themselves as dirty, and in the _Process they would check if they are not visible and return to skip processing. If they are visible and dirty they should load the key icon.

hhyyrylainen:
Now KeyPrompt queues a refresh for when it is next visible before it loads / updates the icon it is showing

**Related Issues**

closes #1635

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/ActionButton.cs
+++ b/src/gui_common/ActionButton.cs
@@ -103,10 +103,9 @@ public partial class ActionButton : Button
 
     private void UpdateKeyPrompt()
     {
-        if (keyPrompt == null || keyPrompt.ActionName == actionName)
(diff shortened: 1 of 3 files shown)


hhyyrylainen:
I'm merging this even if it might have been a bit better to wait for review as otherwise this would have been just staying around until I got back.

PR #5248, opened 2024-06-22
Add red damage screen effect

CI09:

This PR does some stuff...

**Related Issues**

https://wiki.revolutionarygamesstudio.com/wiki/Release_Roadmap#0.6.x_(the_combat_and_part_upgrades_updates)

**Progress Checklist**

      break existing features:
      members reviewing this PR)


shaders/ScreenDamage.gdshader
```diff
@@ -0,0 +1,17 @@
+shader_type canvas_item;
+
+uniform vec4 damageColor : source_color = vec4(0.85f, 0, 0, 0.65f);
+
+// Use to fade effect in or out
```
hhyyrylainen, line 5:
```suggestion
// Used to fade effect in and out
```

src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -665,6 +671,19 @@ protected virtual void UpdateHealth(float delta)
         var hpText = StringUtils.FormatNumber(Mathf.Round(hp)) + " / " + StringUtils.FormatNumber(maxHP);
         hpLabel.Text = hpText;
         hpLabel.TooltipText = hpText;
+
+        var shaderMaterial = (ShaderMaterial)damageScreenEffect.Material;
```
hhyyrylainen, line 675: This should be cached in _Ready and not read basically each frame.

src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -665,6 +671,19 @@ protected virtual void UpdateHealth(float delta)
...
+
+        if ((float)shaderMaterial.GetShaderParameter(fadeParameterName) > 0)
```
hhyyrylainen, line 677: Shader parameters shouldn't be read back as they are much more performance intensive than just having a field in this class that keeps track of the last value applied to the shader.

src/late_multicellular_stage/MulticellularHUD.tscn
```diff
@@ -37,6 +38,12 @@
 [ext_resource type="PackedScene" uid="uid://qwg2dluu8ow0" path="res://src/microbe_stage/PausePrompt.tscn" id="77"]
 [ext_resource type="PackedScene" path="res://src/gui_common/ActionButton.tscn" id="78"]
 
+[sub_resource type="ShaderMaterial" id="ShaderMaterial_32qwl"]
```
hhyyrylainen, line 41: I think it would be better to setup the damage effect as a separate scene so that all the stage HUDs can use it as an instance instead of needing all HUD scenes to have the shader etc. setup correctly.

src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -174,6 +174,8 @@ public partial class CreatureStageHUDBase<TStage> : HUDWithPausing, ICreatureSta
 
     // This block of controls is split from the reset as some controls are protected and these are private
 #pragma warning disable CA2213
+    [Export]
+    private Panel damageScreenEffect = null!;
```
hhyyrylainen, line 178: Can't this be just a base `Control` to give HUDs more flexibility (in case it is needed in the future) for some HUDs.

src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -665,6 +671,19 @@ protected virtual void UpdateHealth(float delta)
         var hpText = StringUtils.FormatNumber(Mathf.Round(hp)) + " / " + StringUtils.FormatNumber(maxHP);
         hpLabel.Text = hpText;
         hpLabel.TooltipText = hpText;
+
+        var shaderMaterial = (ShaderMaterial)damageScreenEffect.Material;
+
+        if ((float)shaderMaterial.GetShaderParameter(fadeParameterName) > 0)
+        {
+            shaderMaterial.SetShaderParameter(fadeParameterName, (float)shaderMaterial
+                .GetShaderParameter(fadeParameterName) - delta);
```
hhyyrylainen, line 680: I think it would be excellent to hide the damage effect overlay when not necessary as it otherwise wastes quite a bit of GPU power (as fullscreen draw passes are pretty expensive) without any visible effect.

src/late_multicellular_stage/MulticellularHUD.tscn
```diff
@@ -152,6 +170,7 @@ offset_right = 150.0
 offset_bottom = 25.0
 fit_content = true
 ExtendedBbcode = "HOLD_TO_SHOW_CURSOR_ADVICE_TEXT"
+EnableTooltipsForMetaTags = null
```
hhyyrylainen, line 173: I think you have some godot broken accidental commits of stuff later in this file which all should be removed...

src/late_multicellular_stage/MulticellularHUD.tscn
```diff
@@ -367,7 +395,6 @@ grow_horizontal = 1
 grow_vertical = 1
 
 [node name="ProcessPanel" parent="." instance=ExtResource("13")]
-anchors_preset = 0
```
hhyyrylainen, line 370: Anchor changes also seem suspicious as these popups are pretty picky now in what their anchors are...

src/late_multicellular_stage/MulticellularHUD.tscn
```diff
@@ -522,21 +554,11 @@ custom_minimum_size = Vector2(450, 0)
 DialogText = "MOVING_TO_AWAKENING_PROTOTYPE"
 WindowTitle = "MOVING_TO_AWAKENING_PROTOTYPE_TITLE"
 
-[connection signal="OnCompoundsToggled" from="BottomLeft/HUDBottomBar" to="." method="CompoundButtonPressed"]
```
hhyyrylainen, line 525: And here's a lot of devastation by godot automatically breaking stuff

src/microbe_stage/MicrobeHUD.tscn
```diff
@@ -689,14 +722,6 @@ custom_minimum_size = Vector2(450, 0)
 offset_right = 300.0
 offset_bottom = 97.0
 
-[connection signal="OnCompoundsToggled" from="BottomLeft/HUDBottomBar" to="." method="CompoundButtonPressed"]
```
hhyyrylainen, line 692: This file has also been hit with the automatic breaking.

hhyyrylainen (changes requested):
This new feature seems fine, but the technical implementation needs polishing up. Also it'd be super good to have an option to turn off in case someone is sensitive to screen flashes or just prefers not having red thrown in their face.

src/late_multicellular_stage/MulticellularHUD.tscn
CI09: Is it really necessary for `anchors_preset = 0` to be here? Appearently Godot really wants it to be gone as every time I save some change it is gone. I can revert it with notepad (IDK what preset it actually is) but it would be tedious to do when saving each time. It might be related that I have Godot opened with Compatibility mode because I need to use OpenGL otherwise I'd need to have quite many attempts at opening Godot only

CI09:
This now has settings and hopefully fixed scenes.

hhyyrylainen (changes requested):
I fixed everything except the splitting the extra panel to a separate scene. Also it looks like the damage effect control doesn't have mouse ignore and scroll ignore set on it, which should be done as it is a fullscreen overlay.

CI09:
It's hard for me to tell if warnings I get are infact uneven anchor warnings but I'll just push it...

hhyyrylainen (approved):
I noticed wrong mouse mode on the overlay but I fixed that. As far as I can tell there's no warnings (I did go through the anchor settings and tweaked them). But this should be fine to merge.

PR #5254, opened 2024-07-02
Begin implementing siderophore

CI09:

Adds siderophore mechanic; rusticyanin and ferroplast now give let player shoot a projectile that breaks smaller, more easily engulfable chunks off of big iron chunks. These are more likely to decay now but emit alot less iron.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Is this nearing feature completion or? If an initial implementation is wanted to be added this should add that planned experimental features option when starting a new game.

CI09:
It is pretty much done except for AI, but I already have idea for how it could work, just locating chunks for AI must be fixed first (I think it's broken). Should experimental features have it's standalone PR?

hhyyrylainen:
> Should experimental features have it's standalone PR?

I guess it could, but I don't want to accidentally end up in a situation where we release a thrive version with the experimental features checkbox but no actual experimental features...

hhyyrylainen:
The experimental features checkbox now exists so this PR can be updated to use that.

CI09:
> The experimental features checkbox now exists so this PR can be updated to use that.

It does now.
Note that it makes big iron chunks not emit compounds almost at all (for both experimental and not, I said something similiar on Discord but I prefer if message was more visible) so it needs some opinion

hhyyrylainen:
I think it is fine for an experimental feature but the overall design of the feature should get another pass before becoming enabled by default feature.

CI09:
It's ready for review.

simulation_parameters/VisualResourceIdentifier.cs
```diff
@@ -16,6 +16,11 @@ public enum VisualResourceIdentifier
 
     CellBurstEffect,
 
+    /// <summary>
+    ///   Iron projectile effect
+    /// </summary>
+    SidenophoreProjectile,
```
hhyyrylainen, line 22: It's not correct to add values here in the middle, the summary comment on this enum says:
```
Exact values are used in saves so new values must be appended
///   at the end.
```

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -755,11 +779,6 @@ ref otherMicrobeInfo.Entity.Get<SpeciesMember>(), speciesOpportunism, speciesFoc
                 // If the predator is right on top of us there's a chance to try and swing with a pilus
                 ai.MoveWithRandomTurn(2.5f, 3.0f, position.Position, ref control, speciesActivity, random);
             }
-
-            // Sprint until full strain
```
hhyyrylainen, line 759: This seems to revert the AI running PR changes?

src/microbe_stage/components/SiderophoreProjectile.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Components;
+
+using DefaultEcs;
+using Newtonsoft.Json;
+
+/// <summary>
+///   Defines toxin damage dealt by an entity
```
hhyyrylainen, line 7: This comment seems incorrect.

src/microbe_stage/components/SiderophoreProjectile.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    /// <summary>
+    ///   Scales the efficiency
+    /// </summary>
+    public bool IsUsed;
+
+    /// <summary>
+    ///   Sender
+    /// </summary>
+    public Entity Sender;
```
hhyyrylainen, line 25: Due to struct padding, largest fields should be first. So `Sender` should be first, then the float fields and finally all the booleans.

src/microbe_stage/components/OrganelleContainer.cs
```diff
@@ -532,6 +538,12 @@ public static void CalculateOrganelleLayoutStatistics(this ref OrganelleContaine
             if (organelleDefinition.HasBindingFeature)
                 container.HasBindingAgent = true;
 
+            if (organelleDefinition.InternalName == "rusticyanin")
```
hhyyrylainen, line 541: These shouldn't check internal names like this, instead components or another property should be used. Perhaps even an "iron score" field should be added to organelle definition that is dynamically calculated from the specified processes at registry load time.

src/microbe_stage/Spawners.cs
```diff
@@ -228,6 +312,8 @@ public static void SpawnCellBurstEffect(IWorldSimulation worldSimulation, Vector
             ApplyVisualScale = Math.Abs(chunkType.ChunkScale - 1) > MathUtils.EPSILON,
         });
 
+        entity.Set(chunkType);
```
hhyyrylainen, line 315: ChunkConfiguration is not meant to be used as a component.

src/microbe_stage/components/MicrobeControl.cs
```diff
@@ -179,6 +185,36 @@ public static class MicrobeControlHelpers
...
+            colony.PerformForOtherColonyMembersThanLeader(m =>
+                m.Get<MicrobeControl>()
+                    .EmitSiderophore(ref m.Get<OrganelleContainer>(),
+                        m));
+        }
+
+        if (control.AgentEmissionCooldown > 0)
+            return false;
+
+        // Only shoot if you have an agent vacuole.
```
hhyyrylainen, line 209: Seems like a copy-pasted comment

src/microbe_stage/systems/MicrobeEmissionSystem.cs
```diff
@@ -209,6 +216,48 @@ private static Vector3 FacingDirection(in Entity entity, ref WorldPosition posit
...
+        float ejectionDistance = cellProperties.CreatedMembrane!.EncompassingCircleRadius +
+            Constants.AGENT_EMISSION_DISTANCE_OFFSET;
+
+        if (cellProperties.IsBacteria)
+            ejectionDistance *= 0.5f;
+
+        // Find the direction the microbe is facing
+        // (actual rotation, not LookAtPoint, also takes colony membership into account and uses the
+        // parent rotation)
+        var direction = FacingDirection(entity, ref position);
```
hhyyrylainen, line 249: This position etc. logic is duplicated from EmitToxin so this should be split into its own method so that there is less code copy-paste.

src/microbe_stage/Spawners.cs
```diff
@@ -84,6 +84,90 @@ public static void SpawnCellBurstEffect(IWorldSimulation worldSimulation, Vector
...
+            location, direction, scale, emitter);
+
+        return recorder;
+    }
+
+    public static EntityRecord SpawnIronProjectileWithoutFinalizing(IWorldSimulation worldSimulation,
+        EntityCommandRecorder commandRecorder, float amount, float lifetime,
+        Vector3 location, Vector3 direction, float scale, Entity emitter)
+    {
+        var normalizedDirection = direction.Normalized();
```
hhyyrylainen, line 114: This seems to have a lot in common with oxytoxy emission. So again here I'd suggest making a common helper method that sets up everyhing except the SiderophoreProjectile versus oxytoxy properties, that'll reduce code duplication and any projectile fixes are automatically picked up by both places.

src/microbe_stage/systems/SiderophoreSystem.cs
```diff
@@ -0,0 +1,137 @@
...
+[With(typeof(CollisionManagement))]
+[With(typeof(Physics))]
+[With(typeof(TimedLife))]
+[With(typeof(SiderophoreProjectile))]
+[RunsAfter(typeof(PhysicsCollisionManagementSystem))]
+[RuntimeCost(0.5f, false)]
+public sealed class SiderophoreSystem : AEntitySetSystem<float>
+{
+    /// <summary>
+    ///   Holds a persistent instance of the collision filter callback to not need to create multiple delegates, and
```
hhyyrylainen, line 22: This summary seems really suspicious...

src/microbe_stage/systems/SiderophoreSystem.cs
```diff
@@ -0,0 +1,137 @@
...
+
+    public SiderophoreSystem(World world, IParallelRunner runner, IWorldSimulation worldSimulation) :
+        base(world, runner)
+    {
+        this.worldSimulation = worldSimulation;
+    }
+
+    protected override void Update(float delta, in Entity entity)
+    {
+        if (!entity.Has<SiderophoreProjectile>())
```
hhyyrylainen, line 36: This is not required as the With attribute requires this system can only run on entities that have this component.

src/microbe_stage/systems/SiderophoreSystem.cs
```diff
@@ -0,0 +1,137 @@
...
+            // Applied a damaging hit, destroy this toxin
+            // TODO: We should probably get some *POP* effect here.
+
+            // Expire right now
+            ref var timedLife = ref entity.Get<TimedLife>();
+            timedLife.TimeToLiveRemaining = -1;
+
+            ref var physics = ref entity.Get<Physics>();
+
+            // TODO: should this instead of disabling the further collisions be removed from the world immediately
```
hhyyrylainen, line 74: These TODOs that are duplicated from the toxin system, I think should just be removed. Unless you want to keep the pop effect TODO if you think that the siderophore on collision should have such an effect.

src/microbe_stage/systems/SiderophoreSystem.cs
```diff
@@ -0,0 +1,137 @@
...
+            // to cause less of a physics impact?
+            // physics.BodyDisabled = true;
+            physics.DisableCollisionState = Physics.CollisionState.DisableCollisions;
+
+            // And make sure the flag we check for is set immediately to not process this projectile again
+            // (this is just extra safety against the time over callback configuration not working correctly)
+            projectile.IsUsed = true;
+
+            // Only deal damage at most to a single thing
```
hhyyrylainen, line 83: This seems like an incorrectly worded copy-pasted comment.

src/microbe_stage/systems/SiderophoreSystem.cs
```diff
@@ -0,0 +1,137 @@
...
+    {
+        var target = collision.SecondEntity;
+
+        // Skip if hit something that's not a chunk
+        if (!target.Has<ChunkConfiguration>())
+            return false;
+
+        ref var configuration = ref target.Get<ChunkConfiguration>();
+
+        var iron = SimulationParameters.Instance.GetCompound("iron");
```
hhyyrylainen, line 99: Iron should be stored as a field in this class (also means that this method cannot be static anymore).

src/microbe_stage/systems/SiderophoreSystem.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        if (configuration.Compounds == null)
+            return false;
+
+        // Check if it is the big iron chunk
+        if (configuration.Name == "BIG_IRON_CHUNK")
```
hhyyrylainen, line 105: This seems pretty dirty coding to me but for now this is probably sufficient.

src/microbe_stage/systems/SiderophoreSystem.cs
```diff
@@ -0,0 +1,137 @@
...
+        {
+            if (configuration.Compounds[iron].Amount > 0)
+            {
+                var smallIronChunk = SimulationParameters.Instance.GetBiome("default")
+                    .Conditions.Chunks["ironSmallChunk"];
```
hhyyrylainen, line 110: This should be fetched in this class constructor.

src/microbe_stage/systems/SiderophoreSystem.cs
```diff
@@ -0,0 +1,137 @@
...
+
+                var efficiency = sender.Get<OrganelleContainer>().IronBreakdownEfficiency;
```
hhyyrylainen, line 112: This needs a has component check as the sender entity can already be destroyed meaning it has no components on it. I think maybe on ECS design it would also make more sense to copy the efficiency to the projectile to avoid this kind of entity referencing back to a data source from which the data could have easily been copied to the projectile entity.

src/microbe_stage/systems/SiderophoreSystem.cs
```diff
@@ -0,0 +1,137 @@
...
+
+                // Not to do operation twice
```
hhyyrylainen, line 114: Is this comment correct?

src/microbe_stage/systems/SiderophoreSystem.cs
```diff
@@ -0,0 +1,137 @@
...
+                    Amount = smallIronChunk.Size,
+                };
+
+                // TODO: biting into iron should deplete it faster
+
+                SpawnHelpers.SpawnChunk(worldSimulation, smallIronChunk, collision.FirstEntity.Get<WorldPosition>()
+                    .Position, new Random(), false);
+
+                // Spawn effect
+                SpawnHelpers.SpawnCellBurstEffect(worldSimulation, collision.FirstEntity.Get<WorldPosition>()
```
hhyyrylainen, line 130: This is obviously a placeholder so a TODO comment should be here about replacing this effect with something more suitable ASAP.

hhyyrylainen (changes requested):
Overall architecture I think is pretty good already but there are some code duplication issues and some small code cleanliness problems (especially related to copied comments which seem to not be suitable to where they were copied to).

src/microbe_stage/systems/MicrobeAISystem.cs
CI09: Is it fine to re-add in new commit? I probably accidentally did it while merging and thus has other stuff in its commit.

hhyyrylainen: It should be totally fine. This PR is likely going to be squash merged so this hiccup won't even show up on the master branch history.

src/microbe_stage/systems/SiderophoreSystem.cs
CI09: Like this?
image
I get https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0120?f1url=%3FappId%3Droslyn%26k%3Dk(CS0120) when trying to access it later.

src/microbe_stage/Spawners.cs
CI09: How do I change chunk configuration?

src/microbe_stage/systems/SiderophoreSystem.cs
hhyyrylainen: Did you make the method non-static? I think for another thing I mentioned you'll need to make the method not static to have access to the class fields.

src/microbe_stage/Spawners.cs
hhyyrylainen: Just don't set it as a component? If you need some specific data to the passed to the chunk then either an existing component or a new component is needed. AFAIK ECS systems handle rarely accessed components pretty fine, so this could set like a component `SiderophoneHitTarget` or something that purely marks stuff siderophore can hit and has the needed info to process the hit then.

src/microbe_stage/systems/MicrobeEmissionSystem.cs
CI09: I don't think that's necessary (for me)

src/microbe_stage/Spawners.cs
CI09: Sorry, I don't get it. How do I change chunk configuration without .Set()ting it? Also I'm kinda lazy to do the SiderophoreHitTarget thing as chunk configuration worked pretty fine for me

src/microbe_stage/systems/MicrobeEmissionSystem.cs
hhyyrylainen: ? If we still had the duplicate code check it would complain about this. This has quite a lot of complex math that is duplicated. If someone writes a fix for example for cell colonies no longer hitting their own toxins, there's a chance the bug is fixed only in one place, which is why code duplication is a serious problem in terms of codebase maintainability.

src/microbe_stage/Spawners.cs
hhyyrylainen: ChunkConfiguration is not meant to be used as a component type, so it should not be set as a component. It would cause all kinds of havoc with the code architecture if ChunkConfiguration was made into a component. So that's why I gave two choices: a new component that stores the needed info, or extending an existing component with the required information.

src/microbe_stage/Spawners.cs
CI09: If I want to make SpawnAgentProjectile with siderophore flag, what do I do with AgentProperties?

hhyyrylainen: Make a helper function that siderohone and agent projectiles calls, and leave out the AgentProperties etc. components that aren't used by both. That way the common part is neatly contained in a single method and just the differences are then in the separate siderohpone and agent spawn methods.

> So again here I'd suggest making a common helper method that sets up everyhing except the SiderophoreProjectile versus oxytoxy properties, that'll reduce code duplication and any projectile fixes are automatically picked up by both places.

CI09:
Hopefully it's better

CI09:
Should I revert changes done to biomes.json before it is merged?

hhyyrylainen:
Oh yes, good catch. Those need to be reverted as it cannot be controlled with the experimental feature flag.

src/microbe_stage/systems/SiderophoreSystem.cs
hhyyrylainen: This comment hasn't been addressed.

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -1250,9 +1275,6 @@ private void BuildChunksCache()
                 // Ignore chunks that wouldn't yield any useful compounds when absorbing
                 ref var compounds = ref chunk.Get<CompoundStorage>();
 
-                if (!compounds.Compounds.HasAnyCompounds())
```
hhyyrylainen, line 1253: Is it intentional that AI considers chunks that don't have compounds for eating? I foresee this causing issues like in the ice biome cells trying to eat the damaging chunks.

src/microbe_stage/components/SiderophoreProjectile.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    public float Amount;
+
+    /// <summary>
+    ///   Is already used and to be disposed
+    /// </summary>
+    public bool IsUsed;
+
+    /// <summary>
+    ///   Used by systems internally to know when they have processed the initial adding of a toxin. Should not be
```
hhyyrylainen, line 28: This is not a toxin so this comment cannot be up to date as it mentions "toxin"

src/microbe_stage/components/OrganelleContainer.cs
hhyyrylainen: This comment is not addressed.

src/microbe_stage/systems/MicrobeAISystem.cs
CI09: I must've consider big iron chunks not compound filled or something.

hhyyrylainen: Well at least with the reverted biomes.json, this change shouldn't be necessary.

CI09:
Sorry for not adressed issues, I added new component for big iron chunks as I noticed cells with iron are considered siderophore target.

src/microbe_stage/components/OrganelleContainer.cs
```diff
@@ -75,6 +75,11 @@ public struct OrganelleContainer
     /// </summary>
     public int AgentVacuoleCount;
 
+    /// <summary>
+    ///   How good organelle is at breaking down iron (rusticyanin - 1, ferroplast - 2)
```
hhyyrylainen, line 79:
```suggestion
    ///   How good this microbe is at breaking down iron (score is sum of scores from all organelles
```

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -431,17 +431,18 @@ private void AIThink(Random random, in Entity entity, ref MicrobeAI ai, ref Heal
             }
         }
 
+        var isIronEater = organelles.IronBreakdownEfficiency > 0;
```
hhyyrylainen, line 434: I think there should be an added if here, that if experimental features are off then it forces `isIronEater` to false to disable all the new behaviours when siderohpore cannot be used.

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -1251,7 +1276,7 @@ private void BuildChunksCache()
                 ref var compounds = ref chunk.Get<CompoundStorage>();
 
                 if (!compounds.Compounds.HasAnyCompounds())
-                    continue;
+                    return;
```
hhyyrylainen, line 1279: I think this return is a mistake here as it stops processing entities on the first chunk that has no compounds. There'd be pretty wild AI bug reports if randomly the AI was unable to see most chunks.

src/microbe_stage/systems/SiderophoreSystem.cs
```diff
@@ -0,0 +1,117 @@
...
+[With(typeof(TimedLife))]
+[With(typeof(SiderophoreProjectile))]
+[RunsAfter(typeof(PhysicsCollisionManagementSystem))]
+[RuntimeCost(0.5f, false)]
+public sealed class SiderophoreSystem : AEntitySetSystem<float>
+{
+    private static ChunkConfiguration smallIronChunk = SimulationParameters.Instance.GetBiome("default")
+        .Conditions.Chunks["ironSmallChunk"];
+
+    private static Compound iron = SimulationParameters.Instance.GetCompound("iron");
```
hhyyrylainen, line 24: Registry data store fields should not be static, these should be `private readonly` and not static.

src/microbe_stage/systems/MicrobeAISystem.cs
CI09: Note that for some reason microbes don't work with chunks and thus siderophore doesn't too. I added it "blindly" (without testing) so it's done but chunks for microbe must be fixed first to see if it works.

src/microbe_stage/systems/SiderophoreSystem.cs
CI09: I get CS0120 with that

src/microbe_stage/systems/MicrobeAISystem.cs
hhyyrylainen: I don't get what you are referring to here... Could you explain?

src/microbe_stage/systems/SiderophoreSystem.cs
hhyyrylainen: Is the processing method still static for the collisions? If it is then make it non-static.

src/microbe_stage/systems/MicrobeAISystem.cs
CI09: Microbes don't see chunks, in short. Either something with cache or further in code.

hhyyrylainen: This being in the cache code definitely breaks the cache as the first chunk without compounds causes all further chunks to never be put in the cache (due to the return).

CI09: Microbes don't care for regular microbial chunks either on master

hhyyrylainen: What do you mean by don't care? Did you check with a debugger to see where the microbes reject trying to pursue a chunk? I'd assume at least that the cache should now work without the `return`.

CI09: I don't have debugger set up, at least yet (if you mean the Thrive debugger, I don't even know how to use my IDE debugger). I'll maybe try to find the cause but no promises.

hhyyrylainen: The other option than a debugger is putting a ton of print statements all over the place to try to figure out where the AI chunk checks fail. As it is a pretty complex situation it's going to be much harder to try to debug without a proper debugger, I recommend taking a bit of time to set a C# debugger up for use with Thrive.

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -526,6 +539,14 @@ private void AIThink(Random random, in Entity entity, ref MicrobeAI ai, ref Heal
 
                     break;
```
hhyyrylainen, line 540: This is probably the line that prevents microbes from wanting to use siderophore. I opened a follow up issue: https://github.com/Revolutionary-Games/Thrive/issues/5328

hhyyrylainen (approved):
I did some code cleaning up and fixing bugs with shared data modification. This should be now fine enough to merge as an experimental feature. I'll open a separate issue about siderophore being on by default.

PR #5256, opened 2024-07-05
Fix radial menu center text being offcenter

### LINKED ISSUE #5255: Signaling agent command GUI text is no longer centered
Just saw in a Thrive let's play that the text for "signal to emit" is not centered anymore (likely a regression with Godot 4):

kuva

I'm just quickly opening this issue. This should be extremely simple to fix by just setting the text alignment property back to centered for that one label.

Edit:
the selection pointer is also not handling correctly, but it looks like it is centered around the central text so hopefully that will automatically get fixed as well when fixing the text.
kuva

CI09:

This PR does some stuff...

**Related Issues**
 
 - Closes #5255 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/RadialMenu.cs
+++ b/src/gui_common/RadialMenu.cs
@@ -335,6 +335,8 @@ public partial class RadialMenu : CenterContainer
         var center = Size / 2;
 
         // Used to move indicator to its own left to avoid bug where it is offset
+        // This is required as Godot 4 no longer allows customizing control rotation pivots, so we need to ourselves
+        // apply this extra offset to get rotation and positioning correct for the indicator
         var pivotOffset = indicator.GetTransform().X * -IndicatorSize / 2;
 
         indicator.Visible = true;
@@ -362,6 +364,8 @@ public partial class RadialMenu : CenterContainer
         // In the indicator rotation coordinates the mouse is a quarter circle off
         indicator.Rotation = mouseAngle + Mathf.Pi * 0.5f;
 
+        // Set indicator position based on the mouse angle. Thanks to this re-calculating the value from scratch the
+        // use of GetTransform above doesn't cause an error that would otherwise start to accumulate.
         indicator.Position = new Vector2(Mathf.Cos(mouseAngle), Mathf.Sin(mouseAngle)) *
             RadialCircleStart + pivotOffset + center;
 
--- a/src/gui_common/RadialMenu.cs
+++ b/src/gui_common/RadialMenu.cs
@@ -333,6 +333,8 @@ public partial class RadialMenu : CenterContainer
         }
 
         var center = Size / 2;
+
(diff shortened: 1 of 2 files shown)


CI09:
I also found a bug where indicator briefly flashes near center when cursor leaves center deadzone, if anyone knows a fix it can be pushed to this branch.

ender-titan1 (approved):
Tested and it appears to work.

src/gui_common/RadialMenu.cs
```diff
@@ -96,6 +96,8 @@ public string CenterText
     public override void _Ready()
     {
         centerLabel = GetNode<Label>(CenterLabelPath);
+        centerLabel.HorizontalAlignment = HorizontalAlignment.Center;
```
hhyyrylainen, line 99: Why is this required if the scene already sets this alignment?

src/gui_common/RadialMenu.cs
```diff
@@ -332,7 +334,6 @@ private void UpdateIndicator()
         }
 
         var center = Size / 2;
-        var indicatorOffset = center - new Vector2(IndicatorSize / 2.0f, IndicatorSize);
```
hhyyrylainen, line 335: What's the reasoning behind modifying this math?

hhyyrylainen (changes requested):

src/gui_common/RadialMenu.cs
CI09: Honestly, I don't know. It just worked. Probably Godot update caused indicator center to be already centered without need to set offset or something like that.

hhyyrylainen: If having this math doesn't hurt, I'd prefer to keep it to make sure it works in the future as well (and just doesn't luckily end up with the correct position).

CI09: Well, it did hurt so...

hhyyrylainen: Alright then, I'll check this without that code being in.

hhyyrylainen: It doesn't seem like the indicator lines up correctly...

hhyyrylainen (changes requested):
The duplicate centered set is now gone but the indicator isn't lined up correctly, here my cursor is pointing directly down but the indicator is not pointing exactly along the line towards the bottom of the screen:

CI09:
Might be hacky but should do.

hhyyrylainen (approved):

hhyyrylainen:
Not sure if this was a thing before but I noticed the indicator starts off at the center of the screen before jumping to the right position, so I opened a follow up issue: https://github.com/Revolutionary-Games/Thrive/issues/5271

PR #5261, opened 2024-07-10
Add option to migrate without moving player themselves

CI09:

If player presses button to spread, they won't be able to move this session anymore (to prevent an exploit where you can just move to patch adjacent to already inhibited one and inhibit during same session) and undo button.

**Related Issues**

https://wiki.revolutionarygamesstudio.com/wiki/Release_Roadmap#0.7.x_(the_graph_updates)

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/locale/af.po
+++ b/locale/af.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2024-07-23 09:54+0300\n"
+"POT-Creation-Date: 2024-07-24 11:43+0300\n"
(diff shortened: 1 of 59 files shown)


CI09:
Besides locale, it is done excluding cosmetics

CI09:
Either red screen setting text wasn't translated or it got removed by accident in this PR but now I changed it in addition to other text

hhyyrylainen (approved):
I did a redo of the workflow for setting up the migration for the player. My changes also include some patch map drawing fixes I think are necessary to cleanly make the feature for only allowing the player to select patches that make sense when setting up the migration. A second look at my changes would be appreciated.

hhyyrylainen:
Added support for saving which I forgot before, and disabled trying to migrate into undiscovered patches.

PR #5265, opened 2024-07-17
Implement chunk scaling for prokaryotes

CI09:

This PR does some stuff...

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/components/Engulfable.cs
+++ b/src/microbe_stage/components/Engulfable.cs
@@ -293,10 +293,14 @@ public static class EngulfableHelpers
 
                     var recorder = worldSimulation.StartRecordingEntityCommands();
 
+                    var isBacteria = true;
+
+                    if (entity.Has<CellProperties>())
+                        isBacteria = entity.Get<CellProperties>().IsBacteria;
+
                     MicrobeDeathSystem.SpawnCorpseChunks(ref organelleContainer,
                         entity.Get<CompoundStorage>().Compounds, spawnSystem, worldSimulation, recorder,
-                        position.Position, new XoShiRo128starstar(), customizeCallback, null, entity
-                            .Get<MicrobeSpeciesMember>().Species.IsBacteria);
+                        position.Position, new XoShiRo128starstar(), customizeCallback, null, isBacteria);
 
                     SpawnHelpers.FinalizeEntitySpawn(recorder, worldSimulation);
 
--- a/src/microbe_stage/components/Engulfable.cs
+++ b/src/microbe_stage/components/Engulfable.cs
@@ -295,7 +295,8 @@ public static class EngulfableHelpers
 
                     MicrobeDeathSystem.SpawnCorpseChunks(ref organelleContainer,
                         entity.Get<CompoundStorage>().Compounds, spawnSystem, worldSimulation, recorder,
-                        position.Position, new XoShiRo128starstar(), customizeCallback, null);
+                        position.Position, new XoShiRo128starstar(), customizeCallback, null, entity
+                            .Get<MicrobeSpeciesMember>().Species.IsBacteria);
 
                     SpawnHelpers.FinalizeEntitySpawn(recorder, worldSimulation);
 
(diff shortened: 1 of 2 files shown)


src/microbe_stage/systems/MicrobeDeathSystem.cs
```diff
@@ -140,13 +140,18 @@ public sealed class MicrobeDeathSystem : AEntitySetSystem<float>
...
             var chunkType = new ChunkConfiguration
             {
-                ChunkScale = 1.0f,
+                ChunkScale = chunkScale,
                 Dissolves = true,
                 PhysicsDensity = 1200.0f,
-                Radius = 1.0f,
-                Size = 3.0f,
+                Radius = chunkScale,
+                Size = chunkScale,
```
adQuid, line 154: Do we want to factor in the hex size of the chunk to the engulfment size?

CI09: I wanted to make it look so if chunk is smaller than microbe visuals, it is engulfable

src/microbe_stage/components/Engulfable.cs
```diff
@@ -295,7 +295,8 @@ public static void OnReportBecomeIngestedIfCallbackRegistered(this ref Engulfabl
 
                     MicrobeDeathSystem.SpawnCorpseChunks(ref organelleContainer,
                         entity.Get<CompoundStorage>().Compounds, spawnSystem, worldSimulation, recorder,
-                        position.Position, new XoShiRo128starstar(), customizeCallback, null);
+                        position.Position, new XoShiRo128starstar(), customizeCallback, null, entity
+                            .Get<MicrobeSpeciesMember>().Species.IsBacteria);
```
hhyyrylainen, line 299: This should access `CellProperties` to check if the cell is bacteria. Accessing the species data is incorrect here as it might not match the spawned cell.

src/microbe_stage/systems/MicrobeDeathSystem.cs
```diff
@@ -412,7 +418,7 @@ private bool HandleMicrobeDeath(ref CellProperties cellProperties, in Entity ent
 
         // Eject compounds and build costs as corpse chunks of the cell
         SpawnCorpseChunks(ref organelleContainer, compounds, spawnSystem, worldSimulation, commandRecorder,
-            position.Position, random, null, glucose);
+            position.Position, random, null, glucose, entity.Get<MicrobeSpeciesMember>().Species.IsBacteria);
```
hhyyrylainen, line 421: Same here.

src/microbe_stage/systems/MicrobeDeathSystem.cs
hhyyrylainen: Shouldn't this still be `3 * chunkScale` otherwise this will massively buff chunk eating as you can fit 3x more chunks in your cell than before? (this line exactly Radius or ChunkScale lines shouldn't be changed).

src/microbe_stage/components/Engulfable.cs
hhyyrylainen: This method already accesses CellProperties below, but due to ordering of the logic you probably need to add a new block that checks if the component exists (and caches that as a bool) and then stores the fact that if the cell is bacteria or not in a variable that is then read on this line.

src/microbe_stage/systems/MicrobeDeathSystem.cs
hhyyrylainen: This method already takes CellProperties as a parameter so this method will be super easy to correct.

hhyyrylainen (changes requested):
I think this is pretty good other than the wrong source of bacteria-ness info usage and the unintended massive balance change for all chunks (not just prokaryotic).

src/microbe_stage/systems/MicrobeDeathSystem.cs
hhyyrylainen: I'll fix this up properly (there's already an accessible reference to CellProperties here).

hhyyrylainen (approved):
I didn't test engulfing the smaller cell chunks in game but I checked through everything else so this should be fine to merge now.

Issue #1991, opened 2021-01-02
Add button in the process panel to pause a process


hhyyrylainen:
So the player can for example pause oxytoxy synthesis to save ATP when not needed, or just burn one ATP source at a time.

We might want a nice pause / play icon for the buttons.

Another idea that has come up is to have a slider to allow tweaking the process speed. So we either might want both or just a slider where the other extreme end is 0 speed, which is the same as pausing.

hhyyrylainen:
Related to @buckly90's idea about controlling processes from the process panel.

hhyyrylainen:
This is kind of a duplicate of: https://github.com/Revolutionary-Games/Thrive/issues/192 but I'll leave both open for now.

hhyyrylainen:
Already done: https://github.com/Revolutionary-Games/Thrive/pull/5332

PR #5273, opened 2024-07-23
Add bioprocess speed control

CI09:

It can be theoretically merged, but only thing it does is add boolean that is always true to make bio process either active or not. Base for https://github.com/Revolutionary-Games/Thrive/issues/1991, might develop it later.

This PR does some stuff...

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/TweakedProcess.cs
+++ b/src/microbe_stage/TweakedProcess.cs
@@ -16,6 +16,8 @@ public struct TweakedProcess
 
     public float Rate;
 
+    public float SpeedMultiplier = 1;
+
     [JsonConstructor]
     public TweakedProcess(BioProcess process, float rate = 1.0f)
     {
--- a/src/microbe_stage/TweakedProcess.cs
+++ b/src/microbe_stage/TweakedProcess.cs
@@ -5,7 +5,7 @@
 /// </summary>
 /// <remarks>
 ///   <para>
-///     This is a struct as this just packs one float and a single object reference in here. This allows much tighter
+///     This is a struct as this just packs a few floats and a single object reference in here. This allows much tighter
 ///     data packing when this is used in lists.
 ///   </para>
 /// </remarks>
--- a/src/microbe_stage/systems/ProcessSystem.cs
+++ b/src/microbe_stage/systems/ProcessSystem.cs
@@ -662,6 +662,9 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
         if (environmentModifier <= MathUtils.EPSILON)
             canDoProcess = false;
 
+        if (process.SpeedMultiplier <= 0)
+            canDoProcess = false;
+
         // Compute spaceConstraintModifier before updating the final use and input amounts
         foreach (var entry in processData.Inputs)
         {
@@ -713,7 +716,7 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
             // For now lets assume compounds we produce are also useful
             bag.SetUseful(entry.Key);
 
-            var outputAdded = entry.Value * process.Rate * environmentModifier;
+            var outputAdded = entry.Value * process.Rate * environmentModifier * process.SpeedMultiplier;
 
             // currentProcessStatistics?.AddOutputAmount(entry.Key, 0);
             currentProcessStatistics?.AddOutputAmount(entry.Key, outputAdded);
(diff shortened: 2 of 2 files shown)


hhyyrylainen:
Actually it would be better to have a float between 0-1 to control process speed (instead of a bool), because there's foreseeable need to have more fine grained process control than just on / off.

src/microbe_stage/BioProcess.cs
```diff
@@ -29,6 +29,8 @@ public class BioProcess : IRegistryType
 
     public string InternalName { get; set; } = null!;
 
+    public bool Active { get; set; } = true;
```
hhyyrylainen, line 32: This cannot be added here, BioProcess is similar to OrganelleDefinition as it is a registry data type. It may not be modified once the game has started up. The speed control needs to be put into TweakedProcess. And if I remember right the tweaked process list is not saved, which will need to be changed with this change for it to persist. Also the method to rebuild the tweaked process list should copy any adjusted speeds when it makes the new list, otherwise the speed control state would get reset each time the player grows a new organelle.

hhyyrylainen (changes requested):

CI09:
> Actually it would be better to have a float between 0-1 to control process speed (instead of a bool), because there's foreseeable need to have more fine grained process control than just on / off.

There is already `Rate` but I assume speed modifier would be better in this case?

CI09:
Okay this should be good now

src/microbe_stage/TweakedProcess.cs
```diff
@@ -16,6 +16,8 @@ public struct TweakedProcess
 
     public float Rate;
 
+    public float SpeedMultiplier = 1;
```
hhyyrylainen, line 19: Looks like the comment on line 8 is now outdated and should say "packs a few floats ..."

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -713,7 +716,7 @@ private void ProcessNode(ref BioProcesses processor, ref CompoundStorage storage
             // For now lets assume compounds we produce are also useful
             bag.SetUseful(entry.Key);
 
-            var outputAdded = entry.Value * process.Rate * environmentModifier;
+            var outputAdded = entry.Value * process.Rate * environmentModifier * process.SpeedMultiplier;
```
hhyyrylainen, line 719: This is not fully correct as this just multiplies the outputs, and not the inputs.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -662,6 +662,9 @@ private void ProcessNode(ref BioProcesses processor, ref CompoundStorage storage
         if (environmentModifier <= MathUtils.EPSILON)
             canDoProcess = false;
 
+        if (process.SpeedMultiplier <= 0)
```
hhyyrylainen, line 665: Laws of physics will break if the speed multiplier is over 1 so I'd add an else here that makes sure speed multiplier can never be over 1.

CI09: Would regular cap do (maximize at 1)?

hhyyrylainen: What do you mean? My suggestion is have this code:
```c#
if (process.SpeedMultiplier <= 0)
{
    canDoProcess = false;
}
else if (process.SpeedMultiplier > 1)
{
    process.SpeedMultiplier = 1;
}
```

src/microbe_stage/systems/ProcessSystem.cs
hhyyrylainen: This is still the case, you only added the multiplier check to the input validation, the input consuming loop (which is on line 797) is not yet updated.

hhyyrylainen (changes requested):

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -790,7 +799,7 @@ private void ProcessNode(ref BioProcesses processor, ref CompoundStorage storage
             if (entry.Key.IsEnvironmental)
                 continue;
 
-            var inputRemoved = entry.Value * totalModifier;
+            var inputRemoved = entry.Value * totalModifier * process.SpeedMultiplier;
```
hhyyrylainen, line 802: I think applying the speed multiplier to `totalModifier` on line 775 would be the most error-proof way as I still see that this doesn't apply to the outputs, so now inputs are consumed less than outputs are added.

hhyyrylainen (approved):
I fixed a few other issues I saw with this (mostly with ProcessStatistics that would have reported incorrect numbers). It's untested whether this works when someone actually wants to use the speed modifier but this doesn't impact the gameplay currently.

CI09:
Thanks :)

PR #5276, opened 2024-07-24
Improved cloud visuals with a new shader

AverageNerdz:
Hello, i choose a really tedious and difficult task as my first contribution to this amazing project because i really wanted to express my gratitude for giving me a chance to participate in it, but without further ado here's the main changes you will experience in the 5 files i edited:
1) Perlin noise --> literally tried to sharpened the pic on photoshop
2) TestCloudConcentrations --> same thing as above as i felt it would impact the edge logic i put in the shader
3) Edited the compoundcloudplane and compoundcloudsystem classes to make the way the cell is surrounded by the compound more circular and less blocky
4) Edited the shader to match the previous look

The only problem i've experienced with the changes i made so far is the random clouds rendering glitch, which apparently happens randomly and seem to only impact brightness so far, some artifacts could be linked to operations among float values too maybe a look into the logic could be a good shot. Anyway thank you for the amazing opportunity i had an amazing time writing and tweaking the code to achieve something unique, i hope to hear you back from you soon.


src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -224,66 +186,44 @@ public void GetAllAvailableAt(Vector3 worldPosition, Dictionary<Compound, float>
...
-                    {
-                        // Not in it
+                    float distance = Mathf.Sqrt(Mathf.Pow(x - cloudRelativeX, 2) + Mathf.Pow(y - cloudRelativeY, 2));
+                    if (distance > localGrabRadius)
                         continue;
-                    }
 
-                    // Then just need to check that it is within the cloud simulation array
+                    float factor = 1.0f - (distance / localGrabRadius);
+
```
hhyyrylainen, line 223: What's the logic behind changes in this file? To me this seems like it first of all removes a ton of comments and seems to change the microbe compound absorption logic in a major way which is a game balance and performance concern as one of the most performance intensive parts of the game is all microbes absorbing compounds each game update (there's an added square root call per loop iteration which is a very heavy math operation).

src/microbe_stage/CompoundCloudPlane.cs
hhyyrylainen, line 1: This file seems to have the same issue with a bunch of code comments and blank lines removed that help readability.

shaders/CompoundCloudPlane.gdshader
```diff
@@ -5,58 +5,94 @@ render_mode unshaded;
...
+        vec2(1.0, 1.0),
+        vec2(-1.0, -1.0),
+        vec2(1.0, -1.0),
+        vec2(-1.0, 1.0)
+    );
+    
+    float sum = noiseValue;
+    for (int i = 0; i < 8; i++) 
+    {
+        sum += texture(noise, uv + offsets[i] * scale).r;
```
hhyyrylainen, line 61: Is this where most of the performance impact comes from for these shader changes? I don't know that much about graphics programming but if this was possible to avoid by generating a noise texture ahead of time in some specific way, that would be much more performance friendly, right?

src/microbe_stage/CompoundCloudPlane.cs
AverageNerdz: i'm importing my project to be compliant to the Devcenter checks and putting the comments back into place, with explainatons regarding the lines you outlined in the comments

hhyyrylainen:
@AverageNerdz it looks like you are making a ton of small commits. That bogs down the CI system a lot by filling the job queue with a bunch of code check jobs. It would be much more preferable to fix everything one CI failure detected instead of making a new commit for basically each line.

If you are having trouble getting the code right in one go, you can run the CI checks locally to know if your commit is going to pass the checks before pushing the commit. https://github.com/Revolutionary-Games/Thrive/blob/master/doc/setup_instructions.md#running-the-format-checks

AverageNerdz:
> @AverageNerdz it looks like you are making a ton of small commits. That bogs down the CI system a lot by filling the job queue with a bunch of code check jobs. It would be much more preferable to fix everything one CI failure detected instead of making a new commit for basically each line.
> 
> If you are having trouble getting the code right in one go, you can run the CI checks locally to know if your commit is going to pass the checks before pushing the commit. https://github.com/Revolutionary-Games/Thrive/blob/master/doc/setup_instructions.md#running-the-format-checks

I apologize, i didn't know this i will try to limit the commits, thank you for letting me know, i will soon look into your reviews as soon as i finish to reimplement the comments and introduce the ones that explains my changes

AverageNerdz:
> @AverageNerdz it looks like you are making a ton of small commits. That bogs down the CI system a lot by filling the job queue with a bunch of code check jobs. It would be much more preferable to fix everything one CI failure detected instead of making a new commit for basically each line.
> 
> If you are having trouble getting the code right in one go, you can run the CI checks locally to know if your commit is going to pass the checks before pushing the commit. https://github.com/Revolutionary-Games/Thrive/blob/master/doc/setup_instructions.md#running-the-format-checks

All the issues that have been pointed out have been fixed thanks to an intuition i had and Game Dungeon's diligent efforts. After thorough testing the build seems to be stable and appearts to be functioning well.

src/microbe_stage/CompoundCloudSystem.cs
hhyyrylainen, line 1: I'm still unsure why this gameplay (absorbing) logic needs to be changed...

src/microbe_stage/CompoundCloudPlane.tscn
```diff
@@ -1,26 +1,37 @@
...
+shader_parameter/colour2 = Color(0, 0, 1, 1)
+shader_parameter/colour3 = Color(0, 1, 0, 1)
+shader_parameter/colour4 = Color(1, 0, 0, 1)
 shader_parameter/UVOffset = Vector2(0, 0)
-shader_parameter/BrightnessMultiplier = 1.0
-shader_parameter/CloudAlpha = 0.7
-shader_parameter/densities = ExtResource("4")
-shader_parameter/noise = ExtResource("2")
+shader_parameter/NoiseScale = 0.05
+shader_parameter/BrightnessMultiplier = 2.0
```
hhyyrylainen, line 32: If brightness multipliers are intended to be changed the biomes.json file also needs to change. Also the default colours above, I think they should stay black transparent colour as unused channels should default to that value to not accidentally display.

shaders/CompoundCloudPlane.gdshader
```diff
@@ -5,58 +5,62 @@ render_mode unshaded;
...
 uniform sampler2D noise;
 
-uniform vec4 colour1 : source_color = vec4(0, 0, 0, 0);
-uniform vec4 colour2 : source_color = vec4(0, 0, 0, 0);
-uniform vec4 colour3 : source_color = vec4(0, 0, 0, 0);
-uniform vec4 colour4 : source_color = vec4(0, 0, 0, 0);
+uniform vec4 colour1 : source_color = vec4(1.0, 1.0, 0.0, 1.0);
+uniform vec4 colour2 : source_color = vec4(0.0, 0.0, 1.0, 1.0);
+uniform vec4 colour3 : source_color = vec4(0.0, 1.0, 0.0, 1.0);
+uniform vec4 colour4 : source_color = vec4(1.0, 0.0, 0.0, 1.0);
```
hhyyrylainen, line 11: Same with the default colours here.

src/microbe_stage/CompoundCloudSystem.cs
GameDungeon: That logic appears to have been changed to make that absorb paths that microbes take though compound clouds smoother and more circular.

hhyyrylainen: I got that now on like the third readthrough. But to me it still seems like it would affect the overall absorption part of gameplay balance, and due to the use of square root and adding multiple new divisions, in terms of CPU performance impact I would guess that the absorption code is now much slower (which is kind of a problem as the cloud performance is basically the game bottleneck for really low end CPUs). 

So I'd really prefer if that was split into a separate PR and maybe even adding a performance option for "Smooth compound absorbing" that can be turned off (though with a setting like this the balance cannot be different between toggling it on or off).

AverageNerdz: > I got that now on like the third readthrough. But to me it still seems like it would affect the overall absorption part of gameplay balance, and due to the use of square root and adding multiple new divisions, in terms of CPU performance impact I would guess that the absorption code is now much slower (which is kind of a problem as the cloud performance is basically the game bottleneck for really low end CPUs).
> 
> So I'd really prefer if that was split into a separate PR and maybe even adding a performance option for "Smooth compound absorbing" that can be turned off (though with a setting like this the balance cannot be different between toggling it on or off).

i apologize for not being able to explain it well, anyway i agree that splitting the solution into a separate PR would be a good approach

AverageNerdz: PS: I also tested the code again with the old compound system and tbh it's not even that noticeable anymore so i think that the usefulness of such system should be discussed in the future, it's cool to have the clouds wrap around the cell like that but if it's going to impact the performance that much we might need to weigh the benefits of such changes.

src/microbe_stage/CompoundCloudPlane.tscn
AverageNerdz: oh if you're referring to this part of code i think it was a typo, i restored the original brightness and it should remain as it was initially

GameDungeon: The brightness was actually intentionally set to 2 there, as at 1 with the new shader they looked desaturated to me. The brightness values would have to be tested the patches, but they are multipliers, so if it looks fine it should be fine.

AverageNerdz:
Also i've added the results of the code to the PR for the sake of completeness.

src/microbe_stage/CompoundCloudSystem.cs
hhyyrylainen: I guess I'll do the PR split as no one did it while I was sleeping. Once done I'm going to use the microbe benchmark to see the the performance impact of the different absorbing (as I think that should show it as there's a bunch of cells in there running the absorbing logic).

src/microbe_stage/CompoundCloudPlane.tscn
hhyyrylainen: I guess we can have 2 follow up PRs then (the first one will be the changed absorbing logic).

hhyyrylainen (approved):
I guess this can now be merged to get in the new cloud look.

I do have one suggestion: if possible it would be nice if the different cloud layers had different noise offsets so they would blend together better when there are multiple clouds at a single location in the world.

hhyyrylainen:
I made the PR with the absorption changes: https://github.com/Revolutionary-Games/Thrive/pull/5301

PR #5285, opened 2024-07-25
Disable sprinting when taking ATP damage

CI09:

I squashed it into healing system, so each ATP damage tick it makes you able to sprint or not and thus adds some delay before you can sprint again. (This PR may be scrapped if it's done the wrong way around)
**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (changes requested):
It would be better to force sprint mode off when out of ATP, similarly to how engulf mode gets forced off (though it needs to be confirmed here that even if the player holds shift they can't move faster, i.e. the sprint disabling effect needs to happen before the movement system to make sure the sprint is not in effect if there is not enough ATP to do so).

src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -781,6 +793,11 @@ protected void UpdateStrain()
         throw new GodotAbstractMethodNotOverriddenException();
     }
 
+    protected virtual bool CanSprint()
```
hhyyrylainen, line 796: I think that additionally it would be pretty nice if the hotbar button state was set to disabled when not being able to sprint to add a second way to communicate to the player what it means when the strain bar turns red.

src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -1088,6 +1105,8 @@ protected override void Dispose(bool disposing)
             }
 
             fadeParameterName.Dispose();
+
+            redFill?.Dispose();
```
hhyyrylainen, line 1109: This line should be within the `if (disposing)` statement.

Also I'm a tiny bit suspicious if this is required, but I suppose it is because when the stylebox is not attached as the theme override, this is an unatached resource.

src/general/base_stage/CreatureStageHUDBase.cs
```diff
@@ -195,6 +195,9 @@ public partial class CreatureStageHUDBase<TStage> : HUDWithPausing, ICreatureSta
 
     private StringName fadeParameterName = new("fade");
 
+    [Export]
+    private StyleBoxFlat? redFill;
```
hhyyrylainen, line 199: It'd be nice to have a bit more descriptive name like `strainBarRedFill`

src/microbe_stage/PlayerMicrobeInput.cs
```diff
@@ -313,7 +313,9 @@ public bool StartSprint()
             return false;
 
         ref var control = ref stage.Player.Get<MicrobeControl>();
-        control.Sprinting = true;
+
+        if (control.CanSprint)
```
hhyyrylainen, line 317: I would prefer an approach where this just sets Sprinting to true and relies on other systems to remove invalid sprint states. That's how the engulfing state works and makes logic much easier to write and less error-prone when calling code doesn't need to worry about if sprinting or engulfing is actually possible.

src/microbe_stage/components/MicrobeControl.cs
```diff
@@ -69,6 +69,11 @@ public struct MicrobeControl
     /// </summary>
     public bool Sprinting;
 
+    /// <summary>
+    ///   Whether this microbe is can sprint
+    /// </summary>
+    public bool CanSprint = true;
```
hhyyrylainen, line 75: Components should be in a fine state when default struct constructed, which sets everything to its type specific default value. So this having true as the default value is not good. Instead this bool should be the reverse (see for example the various `StateApplied` flags that exist instead of there being `Dirty` flags). So this should be something like `IsSprintingExhausted` which would be false by default and true when too exhausted to be able to sprint.

src/general/base_stage/CreatureStageHUDBase.cs
CI09: > This line should be within the if (disposing) statement.

I think it already is

hhyyrylainen: Ah right it is, now that I'm looking again I did indeed misread the number of close braces.

CI09:
Is it possible to fit balancing into this PR or should it be left for another PR?

hhyyrylainen:
You can add balance changes here as well, they shouldn't be very complicated from a gameplay standpoint. And as the feature isn't yet balanced really I won't scrutinize any balance changes much at all.

hhyyrylainen (approved):
I fixed the system attributes not being correct and reverted the red colour to match the other reds used in the game.

PR #5301, opened 2024-07-30
New much more round compound cloud absorption logic

hhyyrylainen:

Separated out part of: https://github.com/Revolutionary-Games/Thrive/pull/5276

Makes for a smoother grab radius for microbe absorption

WIP: I'll test the performance impact of this

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I made this PR based on the other one which is why there's a bunch of commits here but not many changes compared to master.

hhyyrylainen:
Trying to run the microbe benchmark freezes for me each time with these changes.

AverageNerdz:
I had to scramble around a little bit to find the working version, accidentaly loading the wrong backup a few times too and i struggled with incorrect identifiers a little bit, this version is stable and shouldn't freeze during benchmark . I appreciate that you made a separate PR for this cause it needs to be discussed further. We should delve into the utility of this changes more especially since the new shader effectively masks masks the square look during absorbition.

hhyyrylainen:
@AverageNerdz I tried to improve the performance in my version of these changes by not using square root at all. Would it be possible for you to fix things back to that approach?

AverageNerdz:
sure

AverageNerdz:
> @AverageNerdz I tried to improve the performance in my version of these changes by not using square root at all. Would it be possible for you to fix things back to that approach?

i tried to implement the changes you did to the logic and the game crashes whenever i start a game or i run the benchmark tool

hhyyrylainen:
So it *was* my changes. I thought I did the equivalent. Maybe you can figure out the right math to calculate in squared distance mode so that square root operation can be avoided (which is much more expensive math operation than even division)?

AverageNerdz:
> So it _was_ my changes. I thought I did the equivalent. Maybe you can figure out the right math to calculate in squared distance mode so that square root operation can be avoided (which is much more expensive math operation than even division)?




Hello! I've found a way to do this, literally pulled it out from the link below and converted it into a function, i've delved into it in the comments too: https://www.chegg.com/homework-help/questions-and-answers/using-delta-x-delta-y-2-find-overestimate-underestimate-int-r-g-x-y-d--q119913178
It seems to produce far better results during benchmarks and doesn't seem to lag like the old one but i would discuss the usefulness of such changes, i know that having a more rounded absorbition logic is cool but it is just a small graphical improvement and it will impact the game's performance, just look at the two pics i've posted as comparision, during gameplay i think that it's pretty difficult to notice the square trail the cloud leaves.

src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -271,19 +268,25 @@ public void GetAllAvailableAt(Vector3 worldPosition, Dictionary<Compound, float>
...
-                        // Not in it
+                    // Delta calculation using simple math to calculate the square distance
+                    float deltaX = x - cloudRelativeX;
+                    float deltaY = y - cloudRelativeY;
+                    float distanceSquared = deltaX * deltaX + deltaY * deltaY;
+
+                    // This ensures that the distance will be non zero and scales it down
+                    // Despite its simplicity it helps not getting negative values and
+                    // keeps the same results as the old one
+                    float distance = (distanceSquared + 1.0f) / 2.0f;
```
hhyyrylainen, line 279: I don't really understand how this math goes from squared to non-squared distance coordinates.

hhyyrylainen:
I guess if this doesn't really help with anything and just costs performance, then maybe it'd be fine to abandon this idea.

src/microbe_stage/CompoundCloudSystem.cs
AverageNerdz: The naming of the variable i've used could be a bit misleading it should have probably been named in a differtent manner like "distanceaprox" or similar which is clearer in my opinion, however most of the logic that ensures the cloud's round shape precedes these lines. I appreciate and  understand your decision and i'm eager to share some benchmark results in here or dms to help with the final decision.

hhyyrylainen:
Sharing benchmark numbers would be helpful.

If we consider that this doesn't really improve the visuals much, the impact to performance has to be very minimal. Even from the code clarity standpoint I'm leaning more towards that this change isn't really worth it trying to get this done.

AverageNerdz:
**Old logic**
Stationary microbes score: 267
AI microbes score: 201.667
Spawns until no 60 FPS: 557
Microbe stress average FPS: 125.276
Microbe stress min FPS: 41
Alive microbes: 415
Waiting for microbes to die: 130.558
Microbe deaths minimum FPS: 35
Remaining microbes: 57
Total test duration: 288.8
CPU: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz (used tasks: 4, native: 2, sim threads: True)
GPU: NVIDIA TITAN Xp
--------------------------------------------
**Circular logic**
Stationary microbes score: 282.569
AI microbes score: 281.02
Spawns until no 60 FPS: 615
Microbe stress average FPS: 145.109
Microbe stress min FPS: 40
Alive microbes: 467
Waiting for microbes to die: 157.669
Microbe deaths minimum FPS: 20
Remaining microbes: 62
Total test duration: 298.6
CPU: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz (used tasks: 4, native: 2, sim threads: True)
GPU: NVIDIA TITAN Xp
----------------------
Here you go

PR #5315, opened 2024-07-31
Changed ferroplast so that it is non-LAWK

fralegend015:

Makes it so that the ferroplast organelle is only aviable with LAWK turned off.

**Related Issues**

Ferroplasts are non LAWK.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/microbe_stage/organelles.json
+++ b/simulation_parameters/microbe_stage/organelles.json
@@ -1224,7 +1224,8 @@
     ],
     "RequiresNucleus": true,
     "EditorButtonGroup": "Organelle",
-    "EditorButtonOrder": 5
+    "EditorButtonOrder": 5,
+    "LAWK": false
   },
   "axon": {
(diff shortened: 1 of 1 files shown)


fralegend015:
I forgot about the endosymbiosis system...

I just tested it and yes, it allows you to place a ferroplast if you get it as an option.

hhyyrylainen:
I don't think this should be accepted as-is as this will just make it harder to get the ferroplast, but not actually fully lock it. The endosymbiosis systems needs quite a few checks added to not accidentally show LAWK-only things as acceptable.

And of course in general I think this locking would only really be accepted if this leaves gameplay choices to be still good. i.e. we aren't going so hard with LAWK that ferroplast *must* be disabled from the game by default. If Ferroplast being available by default makes the game better I think most Thrive people would agree that it is fine to leave it enabled, even if it is a tiny bit scientifically inaccurate.

PR #5318, opened 2024-08-01
Added pilus damage cooldown on ejecting something

MalyshevArtem:
The value of `Constants.PILUS_MIN_COOLDOWN` was not enough.

### DIFF
--- a/src/microbe_stage/systems/EngulfingSystem.cs
+++ b/src/microbe_stage/systems/EngulfingSystem.cs
@@ -1460,6 +1460,8 @@ public sealed class EngulfingSystem : AEntitySetSystem<float>
 
         RemoveEngulfedObject(ref engulfer, engulfableObject, ref engulfable, false);
 
+        // When ejecting something with a pilus it can immediately deal max pilus damage to the engulfer, to prevent
+        // that damage cooldown is added when ejecting
         if (entity.Has<DamageCooldown>())
         {
             ref var damageCooldown = ref entity.Get<DamageCooldown>();
--- a/src/microbe_stage/systems/EngulfingSystem.cs
+++ b/src/microbe_stage/systems/EngulfingSystem.cs
@@ -50,7 +50,6 @@ using World = DefaultEcs.World;
 [ReadsComponent(typeof(MicrobeEventCallbacks))]
 [ReadsComponent(typeof(WorldPosition))]
 [ReadsComponent(typeof(EntityRadiusInfo))]
-[ReadsComponent(typeof(DamageCooldown))]
 [RunsAfter(typeof(ColonyCompoundDistributionSystem))]
 [RunsAfter(typeof(PilusDamageSystem))]
 [RunsAfter(typeof(MicrobeVisualsSystem))]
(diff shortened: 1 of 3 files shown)


src/microbe_stage/systems/EngulfingSystem.cs
```diff
@@ -43,12 +43,14 @@
...
 [WritesToComponent(typeof(MicrobeAI))]
 [WritesToComponent(typeof(TemporaryEndosymbiontInfo))]
+[WritesToComponent(typeof(DamageCooldown))]
 [ReadsComponent(typeof(CollisionManagement))]
 [ReadsComponent(typeof(MicrobePhysicsExtraData))]
 [ReadsComponent(typeof(OrganelleContainer))]
 [ReadsComponent(typeof(MicrobeEventCallbacks))]
 [ReadsComponent(typeof(WorldPosition))]
 [ReadsComponent(typeof(EntityRadiusInfo))]
+[ReadsComponent(typeof(DamageCooldown))]
```
hhyyrylainen, line 53: This is a good start but I have a few comments.

This attribute shouldn't be here as this writes to the damage cooldown component (write implies also reading).

src/microbe_stage/systems/EngulfingSystem.cs
```diff
@@ -1459,6 +1461,9 @@ private void EjectEverythingFromDeadEngulfer(ref Engulfer engulfer, in Entity en
 
         RemoveEngulfedObject(ref engulfer, engulfableObject, ref engulfable, false);
 
+        ref var damageCooldown = ref entity.Get<DamageCooldown>();
+        damageCooldown.StartCooldown(Constants.PILUS_MAX_COOLDOWN);
```
hhyyrylainen, line 1465: For safety I would add:

```
if (entity.HasComponent<DamageCooldown>())
{
    ...
}
```

here.

Also did you run the threaded run generation script? Having this system now access DamageCooldown may impact the valid execution orders.

hhyyrylainen:
I just checked `DealPilusDamage` method, and isn't your implementation backwards? I mean that it looks like you apply the cooldown to the engulfer and not the engulfed object that is ejected.

src/microbe_stage/systems/EngulfingSystem.cs
MalyshevArtem: No, I didn't. Do I need to run this tool every time I've added some metadata to a system?

hhyyrylainen: Yes, that's correct because otherwise the old system run order will be used, and it may be incorrect due to the new system run requirements.

MalyshevArtem:
@hhyyrylainen At first I added the cooldown to the ejected cell but it didn't work. Then I added the cooldown to my cell.

In the `DealPilusDamage `there are lines:
```
ref var cooldown = ref collision.SecondEntity.Get<DamageCooldown>();

if (cooldown.IsInCooldown())
    return;
```

As I can remember while debugging the `SecondEntity` was my entity ejecting the cell. So the `if` was true and the following lines responsible for damage dealing were not executed. And this worked.

I'll check it again.

hhyyrylainen:
Ah yes, you are probably correct. I was the one who was thinking about this backwards. It certainly is the engulfer who receives the pilus damage and as such needs the damage cooldown applied to it.

src/microbe_stage/systems/EngulfingSystem.cs
MalyshevArtem: Do you mean `[WritesToComponent(typeof(DamageCooldown))]`? I'm just learning the `GenerateThreadedSystems` class. It can take some time.

I added these two attributes because I get the damage cooldown information for the entity and then I modify it. So I thought I need both: reading and writing.

hhyyrylainen: Yes, if there is a WritesTo attribute that means there should not be ReadsFrom attribute for the same component type. Writing already implies reading in the system.

hhyyrylainen (approved):
I added some comments explaining things and re-run the generate threaded script, which ended up finding the same optimal order as before.

All that's left is that you need to sign our CLA to give us permission to use your code. You can do that here: https://dev.revolutionarygamesstudio.com/cla 
Let me know if you have any questions as I really want to get this merged after all the effort that was spent here.

For reference on my systems that script outputs:
```
Found new best thread simulation: 182 thread variance: 0.25
Found new best thread simulation: 181.75 thread variance: 0
Simulated 151835 thread orderings to find the best one
```

PR #5319, opened 2024-08-01
Add support for LAWK organelles in endosymbiosis

CI09:

This PR does some stuff...

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/MicrobeInternalCalculations.cs
+++ b/src/microbe_stage/MicrobeInternalCalculations.cs
@@ -620,7 +620,7 @@ public static class MicrobeInternalCalculations
     }
 
     public static void CalculatePossibleEndosymbiontsFromSpecies(MicrobeSpecies species,
-        List<(OrganelleDefinition Organelle, int Cost)> result)
+        List<(OrganelleDefinition Organelle, int Cost)> result, bool lawk)
     {
         var organelles = species.Organelles.Organelles;
 
@@ -629,9 +629,12 @@ public static class MicrobeInternalCalculations
         {
             var organelle = organelles[i];
 
-            // Skip things that don't give anything from endosymbiosis
+            // Skip things that don't give anything from endosymbiosis or don't suit LAWK
             var resultType = organelle.Definition.EndosymbiosisUnlocks;
             if (resultType == null)
+                continue;
+
+            if (!resultType.LAWK && lawk)
                 continue;
 
             // Handle each resulting organelle type just once
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -768,6 +768,9 @@ public partial class CellEditorComponent :
             behaviourEditor.Visible = false;
         }
 
+        if (endosymbiosisPopup != null)
+            endosymbiosisPopup.Lawk = Editor.CurrentGame.GameWorld.WorldSettings.LAWK;
+
         UpdateMicrobePartSelections();
 
         // After the "if multicellular check" so the tooltip cost factors are correct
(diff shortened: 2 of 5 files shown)


src/microbe_stage/editor/EndosymbiosisPopup.cs
```diff
@@ -141,6 +143,8 @@ private void ShowDataToStartNew(IEnumerable<KeyValuePair<Species, int>> candidat
 
             var choice = candidateGUIScene.Instantiate<EndosymbiosisCandidateOption>();
 
+            choice.Lawk = Lawk;
+
             MicrobeInternalCalculations.CalculatePossibleEndosymbiontsFromSpecies(microbeSpecies, tempSymbionts);
```
hhyyrylainen, line 148: I think a better implementation would be to not pass the LAWK flag to [EndosymbiosisCandidateOption.cs](https://github.com/Revolutionary-Games/Thrive/pull/5319/files#diff-1f85d8238700eb6c64d6bc917a83b6cce349be676eaa3a4ab18e8deae5bef464) and instead filter inside the CalculatePossibleENdosymbiontsFromSpecies method.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -768,6 +768,8 @@ public override void Init(ICellEditorData owningEditor, bool fresh)
             behaviourEditor.Visible = false;
         }
 
+        endosymbiosisPopup.Lawk = Editor.CurrentGame.GameWorld.WorldSettings.LAWK;
```
hhyyrylainen, line 771: I think it would be better to put this configuration just before the endosymbiosis popup is opened, that way the value is always up to date when the popup is opened.

src/microbe_stage/editor/EndosymbiosisPopup.cs
```diff
@@ -8,6 +8,8 @@
 /// </summary>
 public partial class EndosymbiosisPopup : CustomWindow
 {
+    public bool Lawk;
```
hhyyrylainen, line 11: I think this should be marked as [Export] because otherwise this is just a random public field (which is not optimal for OOP style).

hhyyrylainen (changes requested):

CI09:
I got it to working state with addressed issues of yours now.

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -629,9 +629,9 @@ public static float CalculateDigestionEfficiency(int enzymeCount)
         {
             var organelle = organelles[i];
 
-            // Skip things that don't give anything from endosymbiosis
+            // Skip things that don't give anything from endosymbiosis or don't suit LAWK
             var resultType = organelle.Definition.EndosymbiosisUnlocks;
-            if (resultType == null)
+            if (resultType == null || resultType.LAWK != lawk)
```
hhyyrylainen, line 634: How does this LAWK check work? Because if I turn off lawk (so lawk = false) and then doesn't this *only* show non-lawk?

I think the right code would be:

```c#
if (!resultType.LAWK && lawk)
   continue;
```

So if organelle is non-LAWK but the game is in LAWK mode, then it should be hidden. But if the organelle is LAWK it doesn't matter what the game's LAWK setting is.

hhyyrylainen (changes requested):
One more logic thing I have suspicion of.

hhyyrylainen (approved):
The code change looks correct. I don't really have time to playtest and verify things this week so I hope someone else can do it, or I can at the latest verify and merge this next week.

hhyyrylainen:
I suppose no one was available to test, so let's hope this breaks nothing with the merge.

PR #5324, opened 2024-08-05
Make mouse hover over take chromatic abberation into account

### LINKED ISSUE #1394: Make cursor hover over take the barrel distortion into account
Added in this PR: https://github.com/Revolutionary-Games/Thrive/pull/1385

dligr:

Makes mouse hover over menu take screen distortion (from chromatic abberation) into account, so that the menu shows what visually is under cursor.

**Related Issues**

Closes #1394

**Progress Checklist**

      break existing features:
      members reviewing this PR)


AverageNerdz:
This one seems to work much better i tested it for a while and it didn't impact the performances as much as the last one did

AverageNerdz (approved):
This one works much better, i tested it for a while and i can state that it doesn't affect performances as much as the previous one did.

src/general/utils/ScreenUtils.cs
```diff
@@ -0,0 +1,58 @@
+﻿using Godot;
+
+/// <summary>
+///   Utility for screen and view-related effects
+/// </summary>
+public static class ScreenUtils
+{
+    /// <summary>
+    ///   Applies barrel distortion to the given position
```
hhyyrylainen, line 9: I'd like to see a remarks section added here saying that this must match the math in `Chromatic.gdshader`.

Also I'd like to see a comment in `Chromatic.gdshader` saying that if the math is updated then the math in this file needs to be updated.

hhyyrylainen (changes requested):
Thanks for contributing to Thrive. This already works very well, but I noticed when testing that if I crank the chromatic aberration slider to the max, that affects the distortion and seems to cause this math to fail. So maybe this needs to additionally take the chromatic effect strength from settings into account? Also just a tiny nitpick on the code side that placing warning comments on the related pieces of code that must match would be nice.

src/general/utils/ScreenUtils.cs
dligr: Added the comments now

dligr:
I investigated the problem with maths in BarrelDistortion() and found out that I didn't apply one of the operations that were done in the shader script. Now (according to my testing) the distortion to cursor works almost perfect on any setting of chromatic abberation.

third_party/JoltPhysics
hhyyrylainen, line 1: I think you accidentally downgraded this git submodule.

hhyyrylainen:
How does this work with all different distortion values when the CPU-side distortion calculation isn't given `Settings.Instance.ChromaticAmount`? Because as far as I understand that also impacts the distortion amount calculated in the shader, so a function that doesn't take that into account cannot be fully accurate if comparing the chromatic amount slider being at minimum vs when at maximum.

dligr:
`BarrelDistortion()` is given `Settings.Instance.ChromaticAmount` through `ApplyScreenEffects()` (as `distortion` argument) in both `MicrobeCamera` and `MicrobeInspectInfo`, so this setting is taken into account.

hhyyrylainen:
Ah you are correct. I must have already forgotten that code existed as in testing I noticed the result not being correct at max distortion setting. I shall re-test this. If the issue is solved the only problem is the submodule downgrade.

hhyyrylainen (approved):
This seems to now fully work. The actual math is way too complex for me to analyze but the code is neatly split into methods. 

Thanks for the long-awaited fix for this problem.

PR #5332, opened 2024-08-07
Implement process pause panel

CI09:

Implements panel opened together with process panel used to pause processes. It was hard for me to fit it in process panel so I just made another panel. That way it would also make simpler implementing process panel in the editor.

Also I know I could make process buttons appear only when such process exists but I'm kinda tired for today so I'll leave it like that for now.

**Related Issues**

https://wiki.revolutionarygamesstudio.com/wiki/Release_Roadmap#0.7.x_(the_graph_updates)

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I really think there needs to be some way to fit this in because otherwise it'd be pretty impossible to play with the process panel open if it takes up any more screen real estate.

CI09:
Now it is changed to be part of process panel - checkbox next to processes. It is ready for review

src/microbe_stage/ProcessList.cs
```diff
@@ -78,6 +81,8 @@ private ChemicalEquation CreateEquation(StrictProcessDisplayInfoEquality process
         equation.ShowSpinner = ShowSpinners;
         equation.MarkRedOnLimitingCompounds = MarkRedOnLimitingCompounds;
 
+        equation.ToggleProcessPressed += HandleToggleProcess;
```
hhyyrylainen, line 84: This syntax should not be used with Godot signals as this leaks resources.

The old style `Connect` syntax must be used.

src/gui_common/ChemicalEquation.cs
```diff
@@ -77,6 +83,17 @@ public bool ShowSpinner
         }
     }
 
+    public bool ProcessEnabled
+    {
+        get => toggleProcess.ButtonPressed;
+        set
+        {
+            EmitSignal(SignalName.ToggleProcessPressed, this);
```
hhyyrylainen, line 91: This should check if the new value is the same as the old one and not trigger the signal if that is the case.

src/gui_common/ChemicalEquation.cs
hhyyrylainen, line 1: One general thing I noticed is that there doesn't seem to be info flowing from the process list to here. So for example if a process is disabled and then the chemical equation is recreated it seems like to me that this will lose the state and cause the GUI to desync from the actual data, which is not good.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -838,4 +838,38 @@ private void OnTranslationsChanged()
         UpdateColonySizeForMulticellular();
         UpdateColonySizeForMacroscopic();
     }
+
+    private void ToggleProcessPressed(ChemicalEquation equation)
+    {
+        if (!stage!.Player.Has<BioProcesses>())
```
hhyyrylainen, line 844:
```suggestion
        if (!stage!.HasAlivePlayer || !stage.Player.Has<BioProcesses>())
```

Suggestion to more closely match other player related checks to use the general has alive player guard check.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -838,4 +838,38 @@ private void OnTranslationsChanged()
...
+            return;
+
+        ref var processes = ref stage.Player.Get<BioProcesses>();
+
+        if (processes.ActiveProcesses == null)
+            return;
+
+        var newProcesses = new List<TweakedProcess>();
+
+        foreach (var process in processes.ActiveProcesses)
```
hhyyrylainen, line 854: This logic seems a bit weird to me. Why doesn't this just loop the active processes and modify the speed multiplier of the process that matches the equation based on the `BioProcess` it does?

hhyyrylainen (changes requested):

src/gui_common/ChemicalEquation.cs
CI09: I can make it force process to re-enable when equation is recreated (with checkbutton set to true by default), is that good?

src/microbe_stage/MicrobeHUD.cs
CI09: Foreach elements cannot be modified while running the loop.

CI09: I had some trouble with this one

src/gui_common/ChemicalEquation.cs
hhyyrylainen: I'd prefer an approach where the state is preserved. For example if you load a save it would be good for this to not break the expectation that loading a save at any point is fine and doesn't impact the game.

src/microbe_stage/MicrobeHUD.cs
hhyyrylainen: Use a `for (int i = 0; i < ...) ` loop

CI09: "Cannot modify the return value of 'expression' because it is not a variable" `processes.ActiveProcesses[i].SpeedMultiplier = equation.ProcessEnabled ? 1 : 0;`

hhyyrylainen: I think you need to read the struct to a local variable, modify it and then write the full thing back to the list.

src/gui_common/ChemicalEquation.cs
CI09: `equation.ProcessEnabled = process.DisplayInfo.CurrentSpeed > 0;` Kinda hacky but I don't have better ideas for now

hhyyrylainen: That's a fine way to check it, but I think the data needs to be designed to flow nicely into the chemical equation object through the scene hierarchy properly.

src/microbe_stage/MicrobeHUD.cs
CI09: image
This one doesn't work either

hhyyrylainen: You need:
```c#
var list = ....;
var count = list.Count;
for (int i = 0; i < count; ++i)
{
     if (list[i].Stuff == condition)
     {
            var process = list[i];
            process.Thing = updated thing;
            list[i] = process;
     }
}
```

CI09:
Now should be

src/gui_common/ChemicalEquation.cs
```diff
@@ -77,6 +85,22 @@ public bool ShowSpinner
...
 
+    public bool ProcessEnabled
+    {
+        get => toggleProcess.ButtonPressed;
+        set
+        {
+            if (value == lastToggle)
+                return;
+
+            EmitSignal(SignalName.ToggleProcessPressed, this);
```
hhyyrylainen, line 96: Shouldn't this trigger after the value is changed? Because otherwise code reading the event cannot inspect the latest state of ProcessEnabled?

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -838,4 +838,25 @@ private void OnTranslationsChanged()
...
+    {
+        if (!stage!.HasAlivePlayer || !stage!.Player.Has<BioProcesses>())
+            return;
+
+        ref var processes = ref stage.Player.Get<BioProcesses>();
+
+        if (processes.ActiveProcesses == null)
+            return;
+
+        for (int i = 0; i < processes.ActiveProcesses.Count; ++i)
```
hhyyrylainen, line 852: For more efficient code processes.ActiveProcesses and the Count should be stored in local variables to remove the need to load the variables multiple times.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -838,4 +838,25 @@ private void OnTranslationsChanged()
...
+        {
+            if (processes.ActiveProcesses[i].Process.Name == equation.EquationFromProcess!.Name)
```
hhyyrylainen, line 854: For safety this method should check if equation has no process set and error out with a GD.PrintErr around line 846. Also comparing processes by name is a really bad way to compare them so something else should be done. Perhaps the EquationFromProcess needs to be extended to have a method to check if the underlying process matches.

src/microbe_stage/MicrobeHUD.cs
CI09: Count is being get only one time right?

hhyyrylainen: `for (... ; this part here; ...)`

the "this part here" executes once for each loop *iteration*. So Count property causes a method call each iteration of the loop. Whereas fetching it once before the loop and storing in a local variable causes C# just to load the local back for comparison which is much simpler operation.

hhyyrylainen (approved):
I did a few fixes like process comparisons and added theming for the CheckButton control type.

PR #5336, opened 2024-08-09
AI Microbes can move laterally

adQuid:

Grants AI the ability to use non-forward movement vectors. AI will use these side movements when pursuing microbes and chunks, while otherwise just moving forward in order to maximize speed when fleeing or moving to more compound clouds.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09:
AI seems to not be able to move towards player when they are behind them.

adQuid:
> AI seems to not be able to move towards player when they are behind them.

Ok, after testing this more this logic really doesn't work in a lot of cases. I'm taking off the review label.

hhyyrylainen:
This should probably get another look now that: https://github.com/Revolutionary-Games/Thrive/pull/5341 is merged and microbes can actually see chunks again (bit of a whoops if that had been broken for 6 months)

adQuid:
After testing more carefully, I feel more confident this is ready for review now.

CI09 (approved):
Significantly better

src/microbe_stage/components/MicrobeControl.cs
```diff
@@ -215,11 +215,31 @@ public static class MicrobeControlHelpers
...
 
+    /// <summary>
+    ///   Moves microbe towards target position, even if that position is not forward.
+    ///   This does NOT handle any turning.
+    /// </summary>
+    /// <param name="control">Control to hold commands.</param>
+    /// <param name="selfPosition">Position of microbe moving.</param>
+    /// <param name="targetPosition">Vector3 that microbe will move towards.</param>
+    /// <param name="speed">Speed at which to move.</param>
+    public static void SetMoveSpeedPrecise(this ref MicrobeControl control, WorldPosition selfPosition,
```
hhyyrylainen, line 236:
```suggestion
    public static void SetMoveSpeedPrecise(this ref MicrobeControl control, ref WorldPosition selfPosition,
```

Component should be taken in as a ref parameter to avoid an extra struct data copy.

src/microbe_stage/components/MicrobeControl.cs
```diff
@@ -215,11 +215,31 @@ public static class MicrobeControlHelpers
...
+        Vector3 targetPosition, float speed)
+    {
+        control.MovementDirection = selfPosition.Rotation.Inverse()
+            * (targetPosition - selfPosition.Position) * speed;
```
hhyyrylainen, line 240: What happens if the vector is 0 length here (same position currently as target)? I worry that it will cause NaNs or other problems. For safety I would do something like this:
```c#
var vectorToTarget = targetPosition - selfPosition.Position;
if (vectorToTarget.LenqthSquared() < MathUtils.EPSILON)
{
    control.MovementDirection = Vector3.Zero;
    return;
}
```

hhyyrylainen:
I guess I can fix those slight code problems to get this ready for merge and off my list of open tabs.

hhyyrylainen (approved):
I fixed the slight code issues I noticed myself. I'll give a bit of time before merging for other people to comment on how the new AI cell movement looks. I'm not yet 100% sold on it, but at least it is quite distinct change.

adQuid:
Thanks for the support here; I wasn't around to make the changes myself.

PR #5339, opened 2024-08-09
Fix Mucocyst and Engulf effects being disconnected from membrane.

### LINKED ISSUE #5292: Mucocyst shader doesn't always line up exactly with the membrane
My quick analysis on Discord:

> Looks like the mucocyst shader doesn't exactly match the membrane shader when at specific world coordinates (visible when something pushes you or you force your speed up with cheats) 

Probably need to weak and double check that the mucocyst effect shader vertex part is the same as the membrane so that it matches the membrane as the cell moves through the world.

dligr:

Fixes Mucocyst and Engulf effects not being aligned with membrane. Apparently, the shader materials for the two effects weren't marked as `resource_local_to_scene`, which meant that any changes to their values applied to all of their instances.

Additionally, fixes a mistake in membrane code that probably contributed to the problem as well.

**Related Issues**

Fixes #5292

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Membrane.cs
+++ b/src/microbe_stage/Membrane.cs
@@ -337,7 +337,7 @@ public partial class Membrane : MeshInstance3D
 
         MembraneShaderMaterial.SetShaderParameter(wigglynessParameterName, finalWiggly);
         EngulfShaderMaterial.SetShaderParameter(wigglynessParameterName, finalWiggly);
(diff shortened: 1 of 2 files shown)


hhyyrylainen (approved):
Thanks for the fix. I can confirm with my test I did before where I noticed this problem that it is now gone (in that case at least). A bit of a facepalm moment for us, but I suppose you can't always catch simple mistakes like these in reviews.

PR #5343, opened 2024-08-11
Update Extinction Messages

adQuid:

Changes the text for extinction when you have the option to switch to another species (the "it won't be you" followed by "it's not over for you" didn't sit well with me), and added behavior for the case where the player has driven every species extinct. You know someone is going to try it.

Also fixed a small bug where species didn't seem to record as extinct in time resulting in the code trying to shift a player to the species that just died.

(This PR is going to take a while since I'm terrible with translations.)

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
It's not necessary to drive all species to extinction actually to no longer be able to continue, as each time you go extinct the game looks to the closest relative of the *current* species you are playing as. So if the player swaps to a species that hasn't split they'll get a game over the next time they go extinct. So the game doesn't know if there are no more any existing species (only if there are no extant related species to the player).

src/general/GameWorld.cs
```diff
@@ -290,10 +290,6 @@ public void ReplacePlayerSpeciesWith(Species species, bool deleteOldPlayer = tru
         if (deleteOldPlayer)
         {
             oldPlayer.Obsolete = true;
-
-            // Remove all populations of the old species to make sure it is extinct
-            oldPlayer.Population = 0;
-            Map.RemoveAllPopulationsOf(oldPlayer);
```
hhyyrylainen, line 296: What's the reasoning behind this change? The species is marked obsolete so all systems will yell really loudly if there is a situation where the species has somehow stuck around and tries to spawn for example.

src/microbe_stage/gui/ExtinctionBox.cs
```diff
@@ -11,9 +12,13 @@ public partial class ExtinctionBox : CustomWindow
     [Export]
     public NodePath LoadMenuPath = null!;
 
+    [Export]
+    public NodePath? ExtinctionMessagePath;
```
hhyyrylainen, line 16:
```suggestion
    public NodePath ExtinctionMessagePath = null!;
```

Looks like you aren't taking advantage of this being nullable so that's the correct way to define this.

adQuid:
Upon further review I don't think this change is nessicary. I may open a similar PR with something better in the future.

PR #5346, opened 2024-08-12
Make species preview tooltip work in Auto-Evo explorer

### LINKED ISSUE #4842: Make species preview hover tooltip work in auto-evo exploring tool
Follow up to: https://github.com/Revolutionary-Games/Thrive/pull/4817
Currently it doesn't work as the auto-evo exploring tool doesn't have a Thriveopedia instance set so the species data is not available. Could maybe do: https://github.com/Revolutionary-Games/Thrive/issues/4805 or could maybe make an ISpeciesDataProvide interface that is used instead to get the data and the exploring tool would register itself as that data.

dligr:

Makes species preview tooltip work in Auto-Evo explorer.

**Related Issues**

Closes #4842

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
This is a good start on fixing issue https://github.com/Revolutionary-Games/Thrive/issues/4842.
However, I think it would be better to make this a more extensible system. Right now it is probably fine to pass in the data for an override for the species data provider, but in the future when there might be way, way more places that display species data it would become a massive hassle and a very likely source of bugs with some place being forgotten to be updated.

Feel free to try something else, but my initial idea when thinking of fixing this bug was to extend the `ThriveopediaManager` class to allow registering extra species data providers. The interface you created seems pretty suitable for that use so reworking this PR to that approach shouldn't require too many changes. The only hard part is to ensure that the extra species data provider instances are cleaned up from the ThriveopediaManager (to not run into issues with disposed Nodes being accessed).

hhyyrylainen:
I don't have time to do a full review right now but the architecture change commit looked pretty good to me.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -418,6 +422,29 @@ public void AskExit()
         exitConfirmationDialog.PopupCenteredShrink();
     }
 
+    public Species? GetActiveSpeciesData(uint speciesId)
+    {
+        var gameworld = world.GameProperties.GameWorld;
```
hhyyrylainen, line 427: I was originally going to comment that this should always look from the latest generation and in all worlds, but now that I've thought about this a bit I think your solution might be more correct.

So just to confirm this doesn't look starting from the latest generation to make sure this doesn't show too new images of species if going back in time, right?

If that is the case then I think starting from the displayed generation is the correct approach, and then there isn't a need to do anything else to this method except fixing "gameworld" to use camelCase.


For reference I was thinking the code could be something like:

```c#
private Species? LookForSpeciesInWorld(GameWorld specificWorld, uint species)
{
    for (int i = max generation from somewhere; i >= 0; --i)
    {
        return species if found
    }
     
      return null;
}

public Species? GetActiveSpeciesData(uint speciesId)
{
     var species = LookForSpeciesInWorld(world, speciesId);
     
     if(species)
           return species;
     
     foreach (var currentWorld in worldList)
     {
         return species if found
     }
     return null;
}
```

hhyyrylainen (changes requested):
Overall this is already in good shape, but I have a few small comments related to a couple of places in the code.

src/gui_common/CustomRichTextLabel.cs
```diff
@@ -156,7 +156,7 @@ public override void _Draw()
 #pragma warning restore CA2245
     }
 
-    private static bool GetSpeciesFromMeta(string metaString, out Species? species)
+    private bool GetSpeciesFromMeta(string metaString, out Species? species)
```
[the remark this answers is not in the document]
dligr: Right, reverted this

src/auto-evo/AutoEvoExploringTool.cs
dligr: Yes, it starts looking from the displayed generation in the current world to display the right 'version' of the species.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -418,6 +422,29 @@ public void AskExit()
...
+    {
+        var gameWorld = world.GameProperties.GameWorld;
+
+        for (int i = generationDisplayed; i >= 0; --i)
+        {
+            gameWorld.GenerationHistory[i].AllSpeciesData
+                .TryGetValue(speciesId, out var speciesRecord);
+
+            if (speciesRecord == null)
+                return null;
```
hhyyrylainen, line 435: Now that I'm re-reading this I have one more question. What's this code for? Doesn't this cause an early return that would potentially miss data? Shouldn't this just be a case of continuing the search if there is no species record in the current generation for the wanted species?

dligr: I did some thorough investigation on this, and found out that this indeed causes missing data in some edge-cases. More exactly, if a species that is extinct in current generation is requested, it is null. Technically, the former is impossible to do in the current Auto-Evo explorer, but still may cause problems in future. Anyway, the fix is as simple as switching `return` to `continue`.

hhyyrylainen (approved):
With the latest change I no longer see any problems with the code. Great work.
I'm not going to playtest the last change as I think it should be correct and instead go ahead with the merge.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -418,6 +422,29 @@ public void AskExit()
...
+
+        for (int i = generationDisplayed; i >= 0; --i)
+        {
+            gameWorld.GenerationHistory[i].AllSpeciesData
+                .TryGetValue(speciesId, out var speciesRecord);
+
+            if (speciesRecord == null)
+                continue;
+
+            var species = speciesRecord.Species;
```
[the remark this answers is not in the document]
hhyyrylainen: (So no merge just yet as the automatic checks failed)

hhyyrylainen:
@dligr you've done a few excellent PRs. If you want to officially join the Thrive team we'd be glad to have you. Instructions on how to apply are here: https://revolutionarygamesstudio.com/application/ and you can just add a link to any of your PRs as earlier work reference when applying. If you don't want to join the team that's also totally fine, you can keep contributing PRs whenever you feel like it and if you'd like I can add you to the game credits list.

dligr:
Thank you. I'm indeed looking to join the team in the near future, and have now sent an application for it.

PR #5347, opened 2024-08-12
Initial editor process list

CI09:

This PR does some stuff...

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Is this feature actually complete and ready for review?

CI09:
I mean it theoretically works, though it doesn't take environmental conditions into account so it could use some polishing.

hhyyrylainen:
If it doesn't take environmental conditions into account, it's useless as it would be giving the players misinformation. And we'd need to constantly have someone replying on the community discord to bug reports with inconsistent numbers (assuming players would use the feature in the first place, but I expect at least a few players using this).

CI09:
It should be ready for review now.

src/gui_common/ChemicalEquation.cs
```diff
@@ -85,6 +85,12 @@ public bool ShowSpinner
         }
     }
 
+    public bool ShowToggle
+    {
+        get => toggleProcess!.Visible;
+        set => toggleProcess!.Visible = value;
```
hhyyrylainen, line 91: This needs safeguards added. Check how the other GUI things support properties that may be set before the property is set (in short: toggleProcess needs to be nullable and a method for applying its state is needed that is called here and in _Ready).

src/gui_common/ChemicalEquation.cs
```diff
@@ -85,6 +85,12 @@ public bool ShowSpinner
...
+    }
```
hhyyrylainen, line 92: Also would it be more efficient if when this is false this node doesn't need to _Process to update the animation?

src/microbe_stage/ProcessSpeedInformation.cs
```diff
@@ -15,8 +15,8 @@ public ProcessSpeedInformation(BioProcess process)
 
     public string Name => Process.Name;
 
-    public Dictionary<Compound, float> WritableInputs { get; } = new();
-    public Dictionary<Compound, float> WritableOutputs { get; } = new();
+    public Dictionary<Compound, float> WritableInputs { get; set; } = new();
+    public Dictionary<Compound, float> WritableOutputs { get; set; } = new();
```
hhyyrylainen, line 19: I'm against this change as this is meant to be a simple data container, it makes this much more complicated to use code if the properties can be reassigned by any piece of code in the entire game.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -166,6 +166,7 @@ public partial class CellEditorComponent :
 
     private readonly List<EditorUserOverride> ignoredEditorWarnings = new();
 
+    private readonly Compound atp = SimulationParameters.Instance.GetCompound("atp");
```
hhyyrylainen, line 169: This should be outside the pragma warning disable block.

src/microbe_stage/editor/CellEditorComponent.tscn
```diff
@@ -1024,6 +1026,10 @@ layout_mode = 2
 [node name="CompoundStorageStatistics" parent="RightPanel/VBoxContainer/Statistics/VBoxContainer/Body/ScrollContainer/MarginContainer/VBoxContainer" instance=ExtResource("28_vnd4w")]
 layout_mode = 2
 
+[node name="ProcessList" parent="RightPanel/VBoxContainer/Statistics/VBoxContainer/Body/ScrollContainer/MarginContainer/VBoxContainer" instance=ExtResource("41_n2bpg")]
```
hhyyrylainen, line 1029: If I'm looking at this correctly this is just a node you put somewhere? The original idea was that the process panel would be a panel like in the game, so you need to wrap this inside a custom popup and add a button for opening / closing it.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -644,7 +644,7 @@ private void ProcessNode(ref BioProcesses processor, ref CompoundStorage storage
             if (!entry.Key.IsEnvironmental)
                 continue;
 
-            // Processing runs on the current game time following values
+            // Processing runs on the current game time foll//owing values
```
hhyyrylainen, line 647: There's a typo added here.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -788,6 +792,9 @@ public override void Init(ICellEditorData owningEditor, bool fresh)
             behaviourEditor.Visible = false;
         }
 
+        processList.ShowToggles = false;
+        processList.ShowSpinners = false;
```
hhyyrylainen, line 796: These should be set with the Godot Editor.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1902,6 +1909,39 @@ private void StartAutoEvoPrediction()
 
         UpdateCompoundLastingTimes(compoundBalanceData, nightBalanceData, nominalStorage,
             specificStorages ?? throw new Exception("Special storages should have been calculated"));
+
+        // Handle process list
```
hhyyrylainen, line 1913: This should be split as its own method. For simplicity it can be called from here as it needs the energy balance info.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1902,6 +1909,39 @@ private void StartAutoEvoPrediction()
...
+        var processes = new List<TweakedProcess>();
+        var processStatistics = new List<ProcessSpeedInformation>();
+
+        ProcessSystem.ComputeActiveProcessList(editedMicrobeOrganelles, ref processes);
+
+        float consumptionProductionRatio = energyBalance.TotalConsumption / energyBalance.TotalProduction;
+
+        foreach (var process in processes!)
+        {
+            var singleProcess = ProcessSystem.CalculateProcessMaximumSpeed(process, biome, CompoundAmountType.Current);
```
hhyyrylainen, line 1923: The compound amounts here may or may not be needed to be hooked up to the time of day selector button in the editor. Though I'm not fully sure so this needs more investigation (I'd appreciate anyone else trying to look into this before I have to).

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1902,6 +1909,39 @@ private void StartAutoEvoPrediction()
...
+
+            if (consumptionProductionRatio < 1.0)
```
hhyyrylainen, line 1925: What does this line do (could you add a comment)? Also the number should be `1.0f`.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1902,6 +1909,39 @@ private void StartAutoEvoPrediction()
...
+            {
+                foreach (var input in singleProcess.Inputs)
+                {
+                    if (input.Key == atp)
```
hhyyrylainen, line 1929: What's the reason for skipping ATP here?

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1902,6 +1909,39 @@ private void StartAutoEvoPrediction()
...
+                        continue;
+
+                    singleProcess.WritableInputs[input.Key] = input.Value * consumptionProductionRatio;
```
hhyyrylainen, line 1932: Does this actually work? Doesn't this write to the same dictionary this is looping over?

hhyyrylainen (changes requested):

src/gui_common/ChemicalEquation.cs
CI09: How about:
```suggestion
        public bool? ShowToggle
{
    get
    {
        if (toggleProcess == null)
            return false;

        return toggleProcess.Visible;
    }
    set
    {
        if (toggleProcess == null || value == null)
            return;

        toggleProcess.Visible = (bool)value;
    }
}
```

src/gui_common/ChemicalEquation.cs
CI09: Spinner still spins slightly (and slows down) after disabling the process until it stops completely which looks quite nice. If I do as you say we might lose it.

src/gui_common/ChemicalEquation.cs
hhyyrylainen: The get should have an underlying field that can be returned. The set should set the underlying field which is then applied later. Right now the property *will ignore* a value if set too early. That's a super unexpected effect.

src/gui_common/ChemicalEquation.cs
hhyyrylainen: My comment was about the ShowSpinner property (I think I may have commented a line early or a bit late here). So basically my idea: if spinner is not shown at all, it would save on performance if the spinner wasn't attempted to be updated when it isn't visible at all.

hhyyrylainen: I see. I don't know what I saw earlier but now that I checked again, this isn't a property you added and there's already code to skip the animation. This comment is resolved.

src/microbe_stage/editor/CellEditorComponent.cs
CI09: It is

src/microbe_stage/editor/CellEditorComponent.tscn
CI09: It is contained within rightmost editor panel. For me it looks okay

src/microbe_stage/editor/CellEditorComponent.cs
CI09: These aren't export fields. Make them such?

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: But line 170 shows in the diff as "#pragma warning disable CA2213" to me which means that this *is* inside the pragma block.

src/microbe_stage/editor/CellEditorComponent.tscn
hhyyrylainen: There's already a ton of stuff there. I think it is starting to be too much (and especially after my compound storage etc. times are already put there). Especially if you have like at least 8 different processes. So I really think it'd be better to at least initially make this a proper "process panel" like the original issue calls for. This can be further discussed in an UX design context (on dev discord maybe).

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: Yeah, make them exports.

src/microbe_stage/editor/CellEditorComponent.cs
CI09: I thought it's like:

> disable

disabled content

> restore

hhyyrylainen: No wait, you are correct. I'm just misreading everything this morning apparently. Could you add a blank line before that starting pragma? That'll make it visually clear to me that this line is not inside the pragma block and it is before it.

src/microbe_stage/editor/CellEditorComponent.cs
CI09: While WritableInputs write later in the code into Inputs, this somehow works.

hhyyrylainen: Well I'd like to find out why this works because if this suddenly stops working when someone else makes a completely unrelated change, that's going to be super unfun to that person making the other change.

CI09: https://github.com/Revolutionary-Games/Thrive/blob/master/src/microbe_stage/ProcessSpeedInformation.cs#L29

hhyyrylainen: That creates an enumerator that is not immediately read. So that should run into exactly the same bug if a foreach loop was used as it also fundamentally also uses an enumerator on the dictionary.

src/microbe_stage/ProcessSpeedInformation.cs
CI09: What would you suggest? I don't want to make any complicated changes really...

hhyyrylainen: Remove these setters. Then deal with any code that doesn't work anymore in a different way at the code site that breaks.

src/microbe_stage/editor/CellEditorComponent.tscn
```diff
@@ -1024,6 +1028,10 @@ layout_mode = 2
 [node name="CompoundStorageStatistics" parent="RightPanel/VBoxContainer/Statistics/VBoxContainer/Body/ScrollContainer/MarginContainer/VBoxContainer" instance=ExtResource("28_vnd4w")]
 layout_mode = 2
 
+[node name="ProcessListButton" type="Button" parent="RightPanel/VBoxContainer/Statistics/VBoxContainer/Body/ScrollContainer/MarginContainer/VBoxContainer"]
+layout_mode = 2
+text = "PROCESS_PANEL_TITLE"
```
hhyyrylainen, line 1033: A button shouldn't have a text key on it that ends in "TITLE" as that makes no sense and can make translating this impossible. I recommend changing the translation key to "VIEW_CELL_PROCESSES" or something like that.

src/gui_common/ChemicalEquation.cs
```diff
@@ -85,6 +86,17 @@ public bool ShowSpinner
         }
     }
 
+    public bool ShowToggle
+    {
+        get => showToggle;
+
+        set
```
hhyyrylainen, line 93:
```suggestion
        set
```

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: Was this checked?

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1930,6 +1944,42 @@ private void StartAutoEvoPrediction()
...
+
+        foreach (var process in processes!)
+        {
+            var singleProcess = ProcessSystem.CalculateProcessMaximumSpeed(process, biome, CompoundAmountType.Current);
+
+            // If produces more ATP than consumes, lower down production for inputs and for outputs,
+            // otherwise use maximum production values
+            if (consumptionProductionRatio < 1.0f)
+            {
+                foreach (var input in singleProcess.Inputs)
```
hhyyrylainen, line 1966: If I remember right the Inputs property filters out environmental inputs. Does this not adjusting those match the behaviour that is in the during gameplay process panel?

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: This was not addressed yet.

hhyyrylainen (changes requested):

src/microbe_stage/editor/CellEditorComponent.tscn
CI09: I wanted to use already existing translation.

src/microbe_stage/editor/CellEditorComponent.cs
CI09: Gameplay process panel shows environment inputs at most as @ multiplier at the end of equation.

src/microbe_stage/editor/CellEditorComponent.tscn
hhyyrylainen: It's possible in some cases, but not in this as a button is an action so a translation key that is a static title should not be reused here.

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: Well yes, but are those affected by the actual process speed (I cannot remember off the top of my head)? If not then this code is correct.

CI09: At least not by Rate

hhyyrylainen: Looking at the stuff in the process system it needs the environmental compound usage that is reported is just the raw values and not adjusted.

src/microbe_stage/editor/CellEditorComponent.cs
hhyyrylainen: I think this works only because all processes have at most 1 non-environmental input and output. I'll redo the code with a temp dictionary to avoid the problem.

src/microbe_stage/editor/CellEditorComponent.cs
CI09: What do you mean by that?

hhyyrylainen: If you modify a dictionary that you are iterating with foreach and exit the loop, it won't cause an exception.

Collection modified exception is only triggered on the next loop iteration. The same applies here, as there's always just a single item to iterate: the code exits the loop after modifying the dictionary so there is no error. But if anyone ever adds a bio process to the game that takes in 2 non-environmental inputs, then they are going to have a really bad time trying to debug this.

hhyyrylainen (approved):
I fixed the problem with the process speed data generation and scaling. This is now good enough for an initial implementation even if the button to open the panel and the panel initial position aren't the most well adjusted.

Issue #5284, opened 2024-07-25
Enable sprinting by default (no longer an experimental feature)


hhyyrylainen:
This is a meta issue to track when all of these are done:
- [x] https://github.com/Revolutionary-Games/Thrive/issues/5278
- [ ] https://github.com/Revolutionary-Games/Thrive/issues/5279
- [x] https://github.com/Revolutionary-Games/Thrive/issues/5280
- [ ] https://github.com/Revolutionary-Games/Thrive/issues/5281
- [x] https://github.com/Revolutionary-Games/Thrive/issues/5283
- [x] https://github.com/Revolutionary-Games/Thrive/issues/5282

TheESN:
Shouldn't be some issues closed as completed here now?

hhyyrylainen:
Which ones? I think #5280 is done, and perhaps #5283 but I'm pretty sure none of the others are.

TheESN:
What about #5279? Didn't #5293 adressed it?

hhyyrylainen:
I think it maybe only partly did, @CI09 ? I thought it only made the AI sprint when in danger, but the AI is not able to for example use it to attack a target that is slightly faster than it.

CI09:
It attached piece of code to EngagePrey so theoretically it should also attack with sprint. Sorry for not putting it in PR name.

hhyyrylainen:
Alright then https://github.com/Revolutionary-Games/Thrive/issues/5279 is depending on a bit of testing / verification of the AI behaviour.

hhyyrylainen:
I'll close this as the sprinting is enabled by default, but there's still the AI improving and sprinting balancing issues open.

Issue #5352, opened 2024-08-13
Rebinding a key with conflicts seems to no longer work


hhyyrylainen:
It seems like there is a bug with the key rebinding system where now stealing a keybinding that is already in use doesn't work correctly anymore.

Might be related to: https://github.com/Revolutionary-Games/Thrive/issues/4638

PR #5353, opened 2024-08-13
Enabled sprinting by default (fixed keybinding, running with no ATP, GUI tweaks)

### LINKED ISSUE #3887: Allow binding game input actions to modifier keys (Shift, alt etc.)
though maybe some kind of warning should be shown that if you have another keybinding using that modifier, the results may be unpredictable (and at least some minimal testing should be done that using for example shift for something works as intended)

### LINKED ISSUE #5283: Sprinting allows speeding up when dying from lack of ATP
related to: https://github.com/Revolutionary-Games/Thrive/issues/5281

kind of related it seems that sprinting never deals damage even after like a minute of max speed sprinting

hhyyrylainen:

Enables sprinting by default. Even though not everything in https://github.com/Revolutionary-Games/Thrive/issues/5284 is done I think we'll have enough time to fix the most important things.

**Related Issues**

closes #5283
closes #3887

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/input/key_mapping/InputEventItem.cs
+++ b/src/engine/input/key_mapping/InputEventItem.cs
@@ -30,6 +30,21 @@ public partial class InputEventItem : MarginContainer
 #pragma warning restore CA2213
 
     /// <summary>
+    ///   Keeps track of binding modifier keys directly as inputs and not as part of a combination
+    /// </summary>
+    private ModifierKeyMode modifierKeyStatus;
+
+    private double timeSinceModifier;
+
+    private enum ModifierKeyMode
+    {
+        None,
+        ShiftPressed,
+        AltPressed,
+        CtrlPressed,
+    }
+
+    /// <summary>
     ///   If this is currently awaiting the user to press a button (for rebinding purposes)
     /// </summary>
     public bool WaitingForInput { get; private set; }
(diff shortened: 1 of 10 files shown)


src/engine/input/key_mapping/InputEventItem.cs
```diff
@@ -132,6 +147,83 @@ public override void _ExitTree()
         KeyPromptHelper.ControllerTypeChanged -= OnControllerChanged;
     }
 
+    public override void _Process(double delta)
+    {
+        // Delayed key binding (to allow shift to be used in modifier combinations, for example)
+        if (modifierKeyStatus == ModifierKeyMode.None)
+            return;
+
+        timeSinceModifier += delta;
```
CI09, line 156: I noticed that with standalone Shift keybind there is delay between when button is pressed and when it is accepted, it is minor thing though and I guess sooner or later player would figure it out

CI09 (approved):
I think everything is good, maybe except a bug where (at least) sprinting doesn't work when rebound, at least for me.

src/engine/input/key_mapping/InputEventItem.cs
hhyyrylainen: It's kind of necessary because otherwise you cannot bind stuff to like "Shift+W" as it would immediately just bind to shift and the W input would be just ignored. So there has to be some delay, though if lower value still works that could be fine (like maybe 300 ms or so).

hhyyrylainen:
Yeah, the conflict stuff is kind of bugged right now. I opened one more issue about it today as it was *really* not working for me: https://github.com/Revolutionary-Games/Thrive/issues/5352

hhyyrylainen:
I forgot to push my last commit for the day, so there's now another commit.

PR #5357, opened 2024-08-14
Improve AI to sprint to some edibles

CI09:

Before it gets merged, others can also push changes to this branch if they have ideas and are willing to.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (commented):
The AI system now needs `[ReadsComponent(typeof(StrainAffected))]` on it. 
But to keep save compatibility it needs to first check if the StrainAffected component exists before reading it. Otherwise this change will cause loading old saves to immediately crash. When adding that conditional check please also add a TODO comment that the extra check can be replaced with a `[With(typeof(StrainAffected))]` attribute once save compatibility is next broken.

CI09:
Can I in this case use Get<>() on StrainAffected several times (when it is needed?). I can't do something like this:
`if (entity.Has<StrainAffected>())`
`    ref var strain = ref entity.Get<StrainAffected>();`

hhyyrylainen:
> on StrainAffected several times (when it is needed?). 

If it is just like in two places that aren't constantly accessed I think that's fine then. If it is more or used often then I'd recommend having a local variable that holds the strain (and defaults to 0). And you only initialize that variable if the strain affected component exists.That approach is used here: https://github.com/Revolutionary-Games/Thrive/blob/6dd4f604ff76dd9706d3970cf8c5cbe5dcdbf3d7/src/microbe_stage/systems/MicrobeMovementSystem.cs#L186

hhyyrylainen:
I think no one is going to comment on https://github.com/Revolutionary-Games/Thrive/pull/5336 so I'll merge that now so that you can make AI sprint tweaks with that other AI movement affecting change already merged (remember to pull master to get the latest changes).

CI09:
Sorry for force push but I started tweaking things with last commit ammended for easy locating stuff in diff.

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -867,6 +879,10 @@ private void SeekCompounds(in Entity entity, ref MicrobeAI ai, ref WorldPosition
             {
                 ai.LastSmelledCompoundPosition = null;
                 RunAndTumble(ref ai, ref control, ref position, ref absorber, compounds, speciesActivity, random);
+
+                if (ai.Strain < Constants.MAX_STRAIN_PER_ENTITY * 0.7f)
+                    control.Sprinting = true;
```
adQuid, line 884: I could think of two reasons to not sprint for a compound cloud:

1) It's a less efficient use of resources if you're trying to survive compound cloud to compound cloud

2) it might leave you vulnerable if a predator happens by right when you finish your sprint

Maybe this behavior should be behind a check for activity or focus above a certain threshold?

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -787,11 +799,11 @@ private void FleeFromPredators(ref WorldPosition position, ref MicrobeAI ai, ref
                 // If the predator is right on top of us there's a chance to try and swing with a pilus
                 ai.MoveWithRandomTurn(2.5f, 3.0f, position.Position, ref control, speciesActivity, random);
             }
-
-            // Sprint until full strain
-            control.Sprinting = true;
         }
 
+        // Sprint until full strain
+        control.Sprinting = true;
```
adQuid, line 805: This would probably make prey run from very far away. I could see this helping against some predators, and wearing the prey out or even running into another predator in other cases. Maybe keep the other command while putting this one behind a fear threshold?

src/microbe_stage/systems/MicrobeAISystem.cs
hhyyrylainen: If I understood the context properly, then I agree that the AI should not try to run to find compound clouds as running to gather basic resources should be a waste of resources (as long as sprinting is properly balanced in terms of ATP cost).

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -787,10 +799,11 @@ private void FleeFromPredators(ref WorldPosition position, ref MicrobeAI ai, ref
                 // If the predator is right on top of us there's a chance to try and swing with a pilus
                 ai.MoveWithRandomTurn(2.5f, 3.0f, position.Position, ref control, speciesActivity, random);
             }
+        }
 
-            // Sprint until full strain
+        // Sprint until full strain if fears enough
+        if (RollCheck(speciesFear, Constants.MAX_SPECIES_FEAR, random))
```
adQuid, line 805:
```suggestion
        if (speciesFear > Constants.MAX_SPECIES_FEAR / 2.0f)
```

It's confusing, but I would suggest a raw check on the value rather than a roll check. The RollCheck function re-rolls 3 times a second every time the AI runs, which means any microbe would eventually sprint, whereas a threshold check is constant, meaning some species will choose a sprinting strategy and other will select a "slow and steady" route.

adQuid: Actually, I had another idea, and per your description of the PR I went ahead and pushed it.

src/microbe_stage/components/MicrobeAI.cs
```diff
@@ -26,6 +26,8 @@ public struct MicrobeAI
 
     public float PursuitThreshold;
 
+    public float Strain;
```
hhyyrylainen, line 29: At least I think this is like wrong, the data should be read from the component it is in and not duplicated.

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -38,6 +38,7 @@
 [ReadsComponent(typeof(Engulfable))]
 [ReadsComponent(typeof(MicrobeColony))]
 [ReadsComponent(typeof(WorldPosition))]
+[ReadsComponent(typeof(StrainAffected))]
```
hhyyrylainen, line 41: This addition makes it necessary to check with the generate threaded run script that this doesn't invalidate the current order.

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -809,7 +827,7 @@ private void FleeFromPredators(ref WorldPosition position, ref MicrobeAI ai, ref
     private void EngagePrey(ref MicrobeAI ai, ref MicrobeControl control, ref OrganelleContainer organelles,
         ref WorldPosition position, CompoundBag ourCompounds, in Entity entity, Vector3 target, bool engulf,
         float speciesAggression,
-        float speciesFocus, float speciesActivity, Random random)
+        float speciesFocus, float speciesActivity, Random random, float strain)
```
hhyyrylainen, line 830: Shouldn't strain be named currentStrain here as well to be consistent with the other method taking in a strain parameter?

src/microbe_stage/systems/MicrobeAISystem.cs
CI09: It usually crashes if RunsAfter/Before has some conflicts but this time it also crashes and I don't know why.

hhyyrylainen: What happens? If the generate threaded fails to run, then that means this is logically incorrect change. And it only works in the game as it is not enforced but systems may be corrupting the data they are working accidentally simultaneously.

CI09: /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/ExceptionUtils.cs:112 - System.ArgumentException: An item with the same key has already been added. Key: EditorComponentBase`1[ICellEditorData]

I tried updating libraries before that but it didn't do anything

hhyyrylainen: That's just a godot engine bug, that in my understanding causes nothing. You *can* restart the godot editor to get the error to go away for a little bit.

CI09:
I revamped the branch (Also removed Thim's changes so he's welcome to do them again). I still don't understand how simple ReadsComponent can crash generate threaded systems though...

hhyyrylainen:
I ran the generate threaded script and there doesn't seem to be anything wrong, so the problem is purely on your computer:

```
Found new best thread simulation: 184.75 thread variance: 1.25
Found new best thread simulation: 183.75 thread variance: 0.25
Found new best thread simulation: 182.5 thread variance: 0.5
Simulated 76564 thread orderings to find the best one
Removing extra barrier at barrier index 4
Removing extra barrier at barrier index 3
Removing extra barrier at barrier index 2
Updating simulation class partial in src/microbe_stage/MicrobeWorldSimulation.generated.cs
Successfully handled. src/microbe_stage/MicrobeWorldSimulation.generated.cs has been updated
Shutting down native library
Shutdown actions complete
```

hhyyrylainen (approved):
More careful sprinting is probably fine now.

hhyyrylainen:
I created a quick follow up issue about what I talked about yesterday on discord: https://github.com/Revolutionary-Games/Thrive/issues/5429

PR #5360, opened 2024-08-15
Added toxin potency versus fire rate customization

### LINKED ISSUE #5192: Implement toxin potency customization, to allow picking between fire rate and how hard the toxins hit
the slider was already added in the GUI but I disabled it for now as I didn't implement it. Should be relatively simple to adjust the max fire rate calculations based on the slider value when placed into an upgrade, and then when firing multiplying the toxin amount with the toxicity to scale the damage / effect.

hhyyrylainen:

Should technically work but is probably a bit unbalanced right now

**Related Issues**

closes #5192

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/components/OrganelleContainer.cs
```diff
@@ -507,6 +518,11 @@ public static void CalculateOrganelleLayoutStatistics(this ref OrganelleContaine
 
                     container.AvailableToxinTypes.TryGetValue(toxinType, out var existing);
                     container.AvailableToxinTypes[toxinType] = existing + 1;
+
+                    if (organelle.Upgrades?.CustomUpgradeData is ToxinUpgrades toxinUpgrades)
+                    {
+                        rawToxicityValue += toxinUpgrades.Toxicity;
```
CI09, line 524: Does this affect overall toxicity of shots? I know it's hard to make stuff to do otherwise, I just don't want to have another empty review...

src/microbe_stage/systems/MicrobeEmissionSystem.cs
```diff
@@ -231,12 +256,31 @@ private void EmitProjectile(in Entity entity, ref MicrobeControl control, ref Or
...
+            // (so could share a bit of code for checking if ready to shoot yet)
 
             // The cooldown time is inversely proportional to the amount of agent vacuoles.
-            control.AgentEmissionCooldown = Constants.AGENT_EMISSION_COOLDOWN / organelles.AgentVacuoleCount;
+            control.AgentEmissionCooldown =
+                ToxinCooldownWithToxicity(organelles.AgentVacuoleCount, organelles.AverageToxinToxicity);
+        }
+    }
+
+    private float ToxinCooldownWithToxicity(int vacuoleCount, float toxicity)
```
CI09, line 268: For me difference between high firerate and low firerate is insignificant (from playtest). Is there a way to change it?

src/microbe_stage/components/OrganelleContainer.cs
hhyyrylainen: It does. I clarified in other comments and in the tooltip that toxicity is averaged over the entire microbe. As the alternative would be very difficult to implement (especially when keeping save compatibility).

src/microbe_stage/systems/MicrobeEmissionSystem.cs
hhyyrylainen: I calculated and with 2 toxin vacuoles, the maximum firerate is 2 per second. And with minimum firerate you can only fire once per 2 seconds, so there is quite a bit of variation. Though I guess with larger amounts of toxin vacuoles the difference will be smaller.

I think I probably made the damage difference too much, so I'll probably make a follow up that halves the toxicity modifier for damage.

CI09: Oh appearently if you don't change toxin type upgrade changes don't save (toxicity/firerate, that is).

hhyyrylainen: I'll have to check that tomorrow. I may have forgotten update some code that checks if no upgrades are necessary (or there is a more difficult bug at play here).

hhyyrylainen: I couldn't resist so I checked things and found at least one bug where toxin upgrades were considered the same if only the toxicity was changed. Pushed a fix for that.

PR #5363, opened 2024-08-16
Reduced toxin toxicity effect on damage to make it more balanced

hhyyrylainen:

As the effect was way too strong in the initial implementation

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -430,6 +430,8 @@ public static class Constants
     public const float MACROLIDE_BASE_MOVEMENT_DEBUFF = 0.8f;
     public const float MACROLIDE_DEBUFF_DURATION = 5;
 
+    public const float TOXIN_TOXICITY_DAMAGE_MODIFIER_STRENGTH = 0.5f;
+
     /// <summary>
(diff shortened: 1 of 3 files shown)


CI09:
For some reason even at maximum firerate each shot still uses up over 1.0 toxin which is alot compared to what one toxin can produce (for small microbes)

CI09:
Also it take a little longer to defeat a cell with maximum toxicity cytotoxin than maximum firerate cytotoxin, might be enemy cell related (they were different).

hhyyrylainen:
> For some reason even at maximum firerate each shot still uses up over 1.0 toxin which is alot compared to what one toxin can produce (for small microbes)

Well I didn't want to change how much compounds it takes to shoot, just how "toxic" per unit the toxin is.
I'm not sure if I should do that or not as it would be a lot more complicated to balance as you'd need to make sure the effective damage potential is balanced (so there'd be 3 variables to balance rather than 2).

hhyyrylainen:
As there's no major problem with this PR, I need to merge this now to get it into the BOTD. There's still time next week to do tweaks. Or if you can come up with better math for the toxicity changes, you can make a PR suggesting them.

PR #5367, opened 2024-08-17
No ATP engulf damage

CI09:

Removes disabling engulf when no ATP and replace with additional damage.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/systems/OsmoregulationAndHealingSystem.cs
```diff
@@ -86,7 +86,13 @@ protected override void Update(float delta, in Entity entity)
             if (compounds.TakeCompound(atp, cost) < cost)
             {
                 // Ran out of ATP, disable engulf
-                control.SetStateColonyAware(entity, MicrobeState.Normal);
+                // control.SetStateColonyAware(entity, MicrobeState.Normal);
+
+                health.CurrentHealth -= Constants.ENGULF_NO_ATP_DAMAGE;
```
hhyyrylainen, line 91: Doesn't this cause damage really fast as this runs 60 times per second?

I'd prefer an approach where the damage was applied when applying the effect. And there was a new property like `control.ForcedStateDuration` that protects getting kicked out of engulf mode while it lasts.

That way the AI doesn't need changes to avoid killing itself accidentally, and the damage can be applied all at once and the message is sent just once when the player enables the engulf mode.

If this gradual damage approach is kept the damage *must be* multiplied by delta here as otherwise the damage amount is inconsistent if the player doesn't get a steady 60 FPS.

src/microbe_stage/systems/EngulfingSystem.cs
```diff
@@ -206,15 +206,17 @@ protected override void Update(float delta, in Entity entity)
...
-            // Stop engulfing if out of ATP or if this is an engulfable that has been engulfed
+            // Stop engulfing if this is an engulfable that has been engulfed
             bool engulfed = false;
 
             if (entity.Has<Engulfable>())
             {
                 engulfed = entity.Get<Engulfable>().PhagocytosisStep != PhagocytosisPhase.None;
             }
 
-            if (compounds.TakeCompound(atp, cost) < cost - 0.001f || engulfed)
```
hhyyrylainen, line 217: This is actually a good catch that there's a duplicate engulf cost taken here as well as in osmoregulation and healing system. Probably should remove that other place that causes double the intended ATP drain rate.

locale/en.po
```diff
@@ -1690,6 +1690,9 @@ msgstr "Energy Sources:"
 msgid "ENERGY_SUMMARY_LINE"
 msgstr "Total energy gathered is {0} with individual cost of {1} resulting in unadjusted population of {2}"
 
+msgid "ENGULF_NO_ATP_DAMAGE_MESSAGE"
+msgstr "Engulf with no ATP damage"
```
hhyyrylainen, line 1694:
```suggestion
msgstr "Damaged by entering engulf mode with no ATP"
```

hhyyrylainen (changes requested):

src/microbe_stage/systems/OsmoregulationAndHealingSystem.cs
hhyyrylainen: I think I lean towards having it in the engulfing system, that way it is logically in the right place (engulfing energy cost in engulfing system rather than in *osmoregulation and healing* system).

CI09:
Looks good to me (I also playtested).

hhyyrylainen (approved):
I fixed one more thing (being able to die due to the engulf mode enter damage) and I've also now finished basic testing on this. Probably could do with more optimized values for the constants but the initial implementation should be fine.

PR #5368, opened 2024-08-18
Add maximum corpse chunks dropped

### LINKED ISSUE #5296: Add a maximum count for dropped cell corpse chunks
as if I remember right the organelle count is just divided by 3 and that's how many chunks drop. There should be a soft cap that kicks in at like 10 and maybe up to 15-20 chunks is the absolute max.

CI09:

This PR does some stuff...

**Related Issues**

Closes #5296

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -917,6 +917,8 @@ public static class Constants
     public const int CORPSE_CHUNK_DIVISOR = 3;
     public const float CORPSE_CHUNK_AMOUNT_MULTIPLIER = 1.0f;
     public const int CORPSE_CHUNK_AMOUNT_CAP = 20;
+    public const int CORPSE_CHUNK_AMOUNT_DIMINISH_AFTER = 8;
+    public const int CORPSE_CHUNK_AMOUNT_DIMINISH_MORE_AFTER = 16;
     public const float CHUNK_ENGULF_COMPOUND_DIVISOR = 30.0f;
     public const string DEFAULT_CHUNK_MODEL_NAME = "cytoplasm";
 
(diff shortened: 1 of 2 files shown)


hhyyrylainen (changes requested):
This only adds the hard cap to the chunk count. I think it would be better than having a hard cap is that once there's more than 8 chunks to spawn the chunks are doubly rare to spawn, and then at 16 even more hexes would be needed for more chunks and then *finally* at 20 there would be no more chunks spawned. This is what I wanted when I wrote the description of #5296

CI09:
So you want me to utilize randomization? Xoshiro?

hhyyrylainen:
Not randomization, just implementing multiple thresholds.

Something like:
```
chunkCount = ...;

if (chunkCount > 8)
{
    chunkCount = 8 + (chunkCount - 8) / 2;
}

if (chunkCount > 16)
{
    chunkCount = 16 + (chunkCount - 16) / 3;
}

if (chunkCount > 20)
   chunkCount = 20;
``` 

I didn't test that but I imagine that kind of math can be tweaked to be right here.

CI09:
Very well. Though when testing I noticed that cell with 12 metabolosomes only drops like 4, probably due to unrelated to what you sent, divisor. Maybe it is time to change it, or is it good as is?

hhyyrylainen:
I wouldn't change the base divisor, just make the divisor progressively higher to implement a soft cap before hitting the hard cap of maximum dropped chunks.

hhyyrylainen (approved):
I tested this in game and cleaned up the code a bit (and added more tweak constants to let game balancing be done more easily for this feature).

PR #5371, opened 2024-08-19
Add cooldown on healing

### LINKED ISSUE #5370: Add a cooldown before resuming health generation after receiving damage
There should be a cooldown of like 3-5 seconds before cell health starts to regenerate after taking damage. There should be a threshold that like 0.2-0.4 units of damage doesn't trigger the cooldown as pilus can cause quite many small damage hits, which might seem like a bug if health doesn't regenerate but the player doesn't seem to be taking damage.

Implementation is probably quite simple to add detection in the healing system if a cell has taken damage (needs to make sure the system is set as running before the system that clears recent damage events), and then adding a new variable to some component already used by the healing system to track when it should resume healing.

https://forum.revolutionarygamesstudio.com/t/comprehensive-combat-revamp/933/21

CI09:

I'd prefer auto-evo PR to get merged first by the way.

**Related Issues**

 Closes #5370

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -645,6 +645,11 @@ public static class Constants
     /// </summary>
     public const float HEALTH_REGENERATION_RATE = 1.5f;
 
+    /// <summary>
+    ///   Time until cell can regenerate after taking damage;
+    /// </summary>
+    public const float HEALTH_REGENERATION_COOLDOWN = 5.0f;
+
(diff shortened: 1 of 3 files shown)


hhyyrylainen (approved):
The code change seems clean and looks like the implementation ended up being pretty simple. As I'm pressed for time, I'll assume you tested this enough to confirm this works.

PR #5375, opened 2024-08-20
Add creature meshes

dligr:

Adds creature mesh generator based on convolution surfaces.
Along with that, some changes to late multicellular editor to accommodate for the new meshes.

**Related Issues**

https://forum.revolutionarygamesstudio.com/t/convolution-surfaces-a-better-alternative-to-metaballs

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09 (approved):
If it makes you happier, I don't understand a thing in convolution math because I've never been in touch with anything like that. Maybe some day I'll try to analyze that more. One thing that would be nice is make convolution point take into consideration metaball radius, I think it has variable for something like that.
And in addition to that, make a convolution point has bigger radius in default:
image

Also one problem is that with singular metaball, it displays nothing.

hhyyrylainen:
> One thing that would be nice is make convolution point take into consideration metaball radius, I think it has variable for something like that.

There is support in the metaball structure for different metaball sizes, but once again I haven't gotten around to adding a slider in the editor for choosing the metaball size that is to be placed. I assume dligr didn't either add that in this PR.

src/late_multicellular_stage/MulticellularMetaballDisplayer.cs
```diff
@@ -34,7 +34,7 @@ public override void _Ready()
         base._Ready();
 
         // This is here in case we need custom shader effects at some point
-        // Material = new ShaderMaterial()
+        // material = new ShaderMaterial()
         // {
         //     Shader = GD.Load<Shader>("res://shaders/Metaball.shader"),
```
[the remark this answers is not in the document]
hhyyrylainen: I think this needs to reverted as if the code is kept as-is when uncommented it shouldn't compile as Godot properties are capitalized.

src/late_multicellular_stage/SCALIS.cs
```diff
@@ -0,0 +1,272 @@
+﻿using System;
+using System.Collections.Generic;
+using Godot;
+
+public class Scalis : IMeshGeneratingFunction
```
hhyyrylainen, line 5: Filename should match exactly the name of the class.

src/late_multicellular_stage/editor/MetaballEditorComponentBase.cs
```diff
@@ -241,7 +242,7 @@ public override void _Process(double delta)
             throw new InvalidOperationException($"{GetType().Name} not initialized");
 
         // TODO: should we display the hover metaballs setup on the previous frame here?
-        hoverMetaballsChanged = true;
+        // hoverMetaballsChanged = true;
```
hhyyrylainen, line 245: Can you explain the reasoning behind this change?

src/late_multicellular_stage/editor/MetaballEditorComponentBase.cs
```diff
@@ -69,7 +69,8 @@ public partial class MetaballEditorComponentBase<TEditor, TCombinedAction, TActi
 
     protected List<TMetaball> hoverMetaballData = new();
 
-    protected IMetaballDisplayer<TMetaball>? alreadyPlacedVisuals;
+    protected IMetaballDisplayer<TMetaball>? structuralMetaballDisplayer;
+    protected IMetaballDisplayer<TMetaball>? visualMetaballDisplayer;
```
hhyyrylainen, line 73: Could you add comments here clarifying the different purposes of these? I couldn't immediately from this code see what the reason was for splitting this into two.

src/late_multicellular_stage/SCALIS.cs
hhyyrylainen: Also would be pretty nice to have a summary comment on this. I won't ask for detailed comments for everything in this file as this is probably some advanced math stuff but a summary would be nice of what this does.

src/late_multicellular_stage/MulticellularConvolutionDispayer.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            VertexColorUseAsAlbedo = true,
+        };
+
+        ApplyAlpha();
+
+        ExtraCullMargin = AABBMargin;
+    }
+
+    public void DisplayFromList(IReadOnlyCollection<MulticellularMetaball> layout)
```
hhyyrylainen, line 49:
```suggestion
    public void DisplayFromLayout(IReadOnlyCollection<MulticellularMetaball> layout)
```

I think that kind of name would be more informative.

src/late_multicellular_stage/MulticellularConvolutionDispayer.cs
```diff
@@ -0,0 +1,105 @@
...
+            minExtends.X = Mathf.Min(minExtends.X, metaball.Position.X - metaball.Radius - 0.5f);
+            minExtends.Y = Mathf.Min(minExtends.Y, metaball.Position.Y - metaball.Radius - 0.5f);
+            minExtends.Z = Mathf.Min(minExtends.Z, metaball.Position.Z - metaball.Radius - 0.5f);
+
+            maxExtends.X = Mathf.Max(maxExtends.X, metaball.Position.X + metaball.Radius + 0.5f);
+            maxExtends.Y = Mathf.Max(maxExtends.Y, metaball.Position.Y + metaball.Radius + 0.5f);
+            maxExtends.Z = Mathf.Max(maxExtends.Z, metaball.Position.Z + metaball.Radius + 0.5f);
+        }
+
+        var meshGen = new DualContourer();
```
hhyyrylainen, line 65: Could you add a TODO comment here saying that these expensive calculations really need a cache in the future.

For now this is fine as long as it doesn't freeze the game for too long, but in the long run we'll need a proper generation time for when to do the expensive calculations to not freeze during gameplay when a new creature spawns.

src/late_multicellular_stage/DualContourer.cs
```diff
@@ -0,0 +1,423 @@
...
+    public IMeshGeneratingFunction? MathFunction;
+
+    private static System.Collections.Generic.Dictionary<int, Vector3I[]>? lookupTableInt;
+
+    public DualContourer()
+    {
+        CalculateLookupTableIfNeeded();
+    }
+
+    public void CalculateLookupTableIfNeeded()
```
hhyyrylainen, line 32: This seems like something that should be a private method.

src/late_multicellular_stage/DualContourer.cs
```diff
@@ -0,0 +1,423 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Godot;
+using Godot.Collections;
+
+public class DualContourer
```
hhyyrylainen, line 7: A summary comment on this would be helpful.

src/late_multicellular_stage/DualContourer.cs
```diff
@@ -0,0 +1,423 @@
...
+    public Vector3 UnitsFrom = new(-4.0f, -4.0f, -4.0f);
+
+    /// <summary>
+    ///   Specifies end of generated box. Inclusive.
+    /// </summary>
+    public Vector3 UnitsTo = new(4.0f, 4.0f, 4.0f);
+
+    public bool Smoothen = true;
+
+    public IMeshGeneratingFunction? MathFunction;
```
hhyyrylainen, line 23: This seems like it should be a constructor parameter (and made non-nullable) as it doesn't seem like this class can do much if this is null.

src/late_multicellular_stage/DualContourer.cs
```diff
@@ -0,0 +1,423 @@
...
+            return true;
+
+        return false;
+    }
+
+    public Mesh DualContour()
+    {
+        // GD.Print("Starting mesh generation");
+
+        var sw = new Stopwatch();
```
hhyyrylainen, line 139: This stopwatch seems like leftover debug code, unless you want to keep it here printing the times, which I think would be fine for now to kind of track how fast this code is.

src/late_multicellular_stage/DualContourer.cs
```diff
@@ -0,0 +1,423 @@
...
+        // 4. Apply change, using instanteous speed's vector as direction
+        // 5. Make a second pass of 1-4. This algorithm is only an approximation, so sometimes one-pass approach
+        // results in spiky meshes. Optionally clamp the value more than on the first pass.
+
+        float maxToleratedChange = changeClamp / PointsPerUnit;
+
+        float d = 0.25f / PointsPerUnit;
+
+        int count = points.Count;
+        Parallel.For(0, count, delegate(int i)
```
hhyyrylainen, line 343: This should use our TaskExecutor as we don't use the default threading system. (also applies to anywhere else in case you added threading elsewhere as well in this PR).

src/late_multicellular_stage/editor/MetaballBodyEditorComponent.cs
```diff
@@ -135,6 +139,28 @@ public enum SelectionMenuTab
         Appearance,
     }
 
+    public bool PreviewMode
```
hhyyrylainen, line 142: Please add a summary comment on this.

src/late_multicellular_stage/editor/MetaballBodyEditorComponent.cs
```diff
@@ -393,6 +422,22 @@ public bool ShowMetaballOptions()
         return true;
     }
 
+    public override void SetEditorWorldTabSpecificObjectVisibility(bool shown)
+    {
+        SetEditorWorldGuideObjectVisibility(shown);
```
hhyyrylainen, line 427: Shouldn't this also call the base method or is there a reason why this doesn't?

src/late_multicellular_stage/editor/MetaballBodyEditorComponent.cs
```diff
@@ -450,9 +497,16 @@ protected override MetaballLayout<MulticellularMetaball> CreateLayout()
         return new MetaballLayout<MulticellularMetaball>(OnMetaballAdded, OnMetaballRemoved);
     }
 
-    protected override IMetaballDisplayer<MulticellularMetaball> CreateMetaballDisplayer()
+    protected override IMetaballDisplayer<MulticellularMetaball> CreateVisualMetaballDisplayer()
     {
-        var displayer = metaballDisplayerScene.Instantiate<MulticellularMetaballDisplayer>();
+        var displayer = visualMetaballDisplayerScene.Instantiate<MulticellularConvolutionDispayer>();
```
hhyyrylainen, line 502: Ah, this seems to be where I realize that visual metaball displayer is the convolution displayer.
I guess the variable naming is fine but I'll still ask for those extra comments to explain things to anyone new stumbling upon the variables.

src/late_multicellular_stage/SCALIS.cs
```diff
@@ -0,0 +1,272 @@
...
+
+        return linePos;
+    }
+
+    private float SquareCutoffValue(Vector3 pos, Vector3 a, Vector3 b)
+    {
+        return (ClosestPoint(pos, a, b) - pos).LengthSquared();
+    }
+
+    private Vector3 VectorFromTo(Vector3 from, Vector3 to)
```
hhyyrylainen, line 200: This seems a bit weird to have as a separate method so I think this should be just inlined whereever this is needed.

src/late_multicellular_stage/SCALIS.cs
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        if (i == 3)
+        {
+            return sigma * sigma * sigma * 2;
+        }
+
+        if (i < 0)
+        {
+            throw new Exception("i is less that 0");
```
hhyyrylainen, line 159:
```suggestion
            throw new ArgumentException("i is less that 0");
```

src/late_multicellular_stage/SCALIS.cs
```diff
@@ -0,0 +1,272 @@
...
+    public float CutoffPointMultiplier = 1.5f;
+
+    public float InnerCutoffPointMultiplier = 2.0f;
+
+    public bool Cutoff = true;
+
+    /// <summary>
+    ///   Every two consecutive points form a bone
+    /// </summary>
+    public Metaball[]? Points;
```
hhyyrylainen, line 16: Does this need to be public?

src/late_multicellular_stage/SCALIS.cs
```diff
@@ -0,0 +1,272 @@
...
+
+    private float surfaceValue = 1.0f;
+
+    public float SurfaceValue
+    {
+        get => surfaceValue;
+        set => surfaceValue = value;
+    }
+
+    public void FindBones(IReadOnlyCollection<MulticellularMetaball> layout)
```
hhyyrylainen, line 28: What if the constructor was already made to require to be given the layout and it would calculate the Points? That way it can be read-only.

Also I think the array could be created directly without an intermediate list as it should be the case that there's exactly one root metaball, so the list length is already known based on the layout item count. This will save on allocating an extra list.

src/late_multicellular_stage/DualContourer.cs
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    private void CalculatePoints(bool[,,] shapePoints, Vector3I gridFrom, Vector3I gridTo)
+    {
+        var sw = new Stopwatch();
+        sw.Start();
+
+        Vector3I gridOffset = -gridFrom;
+
+        Parallel.For(gridFrom.X, gridTo.X + 1, delegate(int x)
```
hhyyrylainen, line 248: This seems also pretty questionable to be, and should really be unrolled into a fixed number of tasks to use with our TaskExecutor.

hhyyrylainen:
That's my initial review. I didn't pay super close attention to the math itself, but I think I hopefully caught all the major problems. Only really summary comments and the Parallel to Task stuff is something I couldn't do myself in an emergency to get this in time for next release.

src/late_multicellular_stage/DualContourer.cs
```diff
@@ -0,0 +1,423 @@
...
+        var arrays = new Array();
+        arrays.Resize((int)Mesh.ArrayType.Max);
+
+        arrays[(int)Mesh.ArrayType.Vertex] = points.ToArray();
+        arrays[(int)Mesh.ArrayType.Index] = triIndices.ToArray();
+        arrays[(int)Mesh.ArrayType.Normal] = normals;
+        arrays[(int)Mesh.ArrayType.Color] = colors;
+
+        // arrays[(int)Mesh.ArrayType.TexUV] = newUV;
+        // arrays[(int)Mesh.ArrayType.TexUV2] = newUV1;
```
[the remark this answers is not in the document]
dligr: Right, I mostly kept it as a template for when UV mapping gets done.

src/late_multicellular_stage/editor/MetaballEditorComponentBase.cs
dligr: I mostly added it not to update hover metaball each frame (back when meshes were updated in real-time), but it seems unneeded now.

src/late_multicellular_stage/DualContourer.cs
dligr: Added a single debug print for overall time spent.

src/late_multicellular_stage/editor/MetaballBodyEditorComponent.cs
dligr: Well, the base method does almost the same operations as this one, except for applying preview mode. So I don't think it really is needed to call it.

src/late_multicellular_stage/editor/MetaballBodyEditorComponent.cs
dligr: I added the comments (specifically clarified that the visual displayer displays convolution surfaces)

src/late_multicellular_stage/DualContourer.cs
dligr: Done

src/late_multicellular_stage/DualContourer.cs
dligr: Should be done now too

src/late_multicellular_stage/MulticellularMetaballDisplayer.cs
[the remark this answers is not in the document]
dligr: Right, I just thought that material here referred to the 'material' field of this class

src/late_multicellular_stage/editor/MetaballEditorComponentBase.cs
hhyyrylainen: Alright, good to then revert this change so that it is easier to playtest this PR (to make the chance of there being some corner case bug here less likely).

src/late_multicellular_stage/MulticellularMetaballDisplayer.cs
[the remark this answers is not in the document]
hhyyrylainen: I could be mistaken, but if this was previously exactly like `Material = new ...` then I think this is likely a placeholder in case the `Material` property of the current Node needs to use a more advanced shader in the future.

hhyyrylainen:
Regarding the threading changes, have you tried doing more per task? I ask because it is relatively expensive to fire off work to a background thread meaning you need to have enough work for it to make sense.

I didn't fully read the changes yet, but to me it looked like the thread task size was pretty small and probably costs quite a lot of performance. Also if it is easy to control the task count, checking how many threads the task executor is using and only making that many tasks could be another optimization (at least in terms of how many Task objects need to be created as each one causes garbage thus increasing garbage collector pressure).

dligr:
Controlling task amount in `AdjustVertices()` would be as simple as controlling its step variable, so the change here should be trivial. It's somewhat more complex in `CalculatePoints()`, since steps are used for all three coordinates here. To properly control task amount with thread count here, we would kind of need to do cube root of the latter, so perhaps an approximation of the proper task count would do (in `CalculatePoints()`)?

hhyyrylainen:
Sounds pretty good. I think even a very rough heuristic would be pretty good. Anyway the task count needs to only really support a value range like 4-16, that should be fine for basically any CPU running Thrive (even if not absolutely optimal for the low and extremely high end).

src/late_multicellular_stage/editor/MetaballBodyEditorComponent.cs
dligr: Done

dligr:
I have implemented the thread-based task amount. After some testing, however, it seems like the algorithm actually works somewhat faster when task amount outnumbers threads. For a ~40 metaball creature, time goes from ~0.34 s to ~0.26 s when task amount is around 10x the thread amount. It probably has to do with the fact that task execution time varies in different parts of the grid. So, maybe it would actually be better to have more tasks than threads?

hhyyrylainen:
I think it is fine to have more tasks than threads. You should just confirm that if the background threads are at a low value (like 2-4) that the more tasks doesn't negatively impact that setup as the CPUs with fewer threads are the ones that most need the additional little bit of performance. You can easily test that with the thread count slider in the thrive performance options.

src/late_multicellular_stage/MulticellularConvolutionDispayer.cs
```diff
@@ -0,0 +1,104 @@
...
+            ApplyAlpha();
+        }
+    }
+
+    public override void _Ready()
+    {
+        base._Ready();
+
+        // This is here in case we need custom shader effects at some point
+        // Material = new ShaderMaterial()
```
hhyyrylainen, line 35: On reading the full code I now see that this probably was a case where originally I used the current node's material here, then left the code commented out and finally moved to a different approach. This is probably the reason why Material is capitalized here, but it shouldn't really matter so this is just fine to leave like it is currently.

src/late_multicellular_stage/Scalis.cs
```diff
@@ -0,0 +1,309 @@
...
+        set => surfaceValue = value;
+    }
+
+    public float GetValue(Vector3 pos)
+    {
+        const float sigma = 1.0f;
+
+        float value = 0.0f;
+
+        pos /= sigma;
```
hhyyrylainen, line 53: I'll assume that changing this to `pos *= 1 / sigma` is the equivalent (not counting floating point imprecision). As division is a much more expensive operation, I'll try swapping all the divisions that are easy to convert to multiplications here.

src/late_multicellular_stage/Scalis.cs
```diff
@@ -0,0 +1,309 @@
...
+
+            contributionSum += contribution;
+
+            colourSum += point.Colour.Lerp(point.Parent.Colour, point.Position.DistanceSquaredTo(linePos)
+                / abDistanceSquared) * contribution;
+        }
+
+        colourSum /= contributionSum;
+
+        colourSum.A = 0.0f;
```
hhyyrylainen, line 182: I assume that this is a typo as I assume surface colour shouldn't be fully transparent...

hhyyrylainen (approved):
I did a few small tweaks here and there, and put in some TODO comments about future work to optimize this feature when the time for that comes. I don't think I broke anything as my creature still looked fine.

I'll merge this now so that we can have this feature in 0.7.0 (feature freeze is on Saturday but we have one more week to fix any more bugs). So if you see anything wrong I did, please make a new PR of the fixes. Also finally a huge thanks for working on the convolution surfaces, which seemed like a feature no one would be able to tackle for years. I'm sure many players will be very pleasantly surprised about this sudden progress.

PR #5377, opened 2024-08-20
Fix AutoEvo Error

GameDungeon:

This currently fixes the dictionary part of the auto evo error. I have not yet been able to track down the duplicate species error yet which is the other one remaining

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/steps/ModifyExistingSpecies.cs
+++ b/src/auto-evo/steps/ModifyExistingSpecies.cs
@@ -157,7 +157,8 @@ public class ModifyExistingSpecies : IRunStep
         {
             mutation.Item2.OnEdited();
 
-            newMiche.InsertSpecies(mutation.Item2, patch, scores, cache, false, workMemory);
+            if (newMiche.InsertSpecies(mutation.Item2, patch, scores, cache, false, workMemory))
+                scores[mutation.Item2] = 0;
         }
 
         newOccupantsWorkMemory.Clear();
@@ -205,7 +206,7 @@ public class ModifyExistingSpecies : IRunStep
                         var oldScore = cache.GetPressureScore(pressure, patch, species);
 
                         // Break if mutation fails a pressure
-                        if (newScore <= 0)
+                        if (newScore < 0)
                         {
                             combinedScores = -1;
                             break;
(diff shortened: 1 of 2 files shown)


src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -205,7 +206,7 @@ public bool RunStep(RunResults results)
                         var oldScore = cache.GetPressureScore(pressure, patch, species);
 
                         // Break if mutation fails a pressure
-                        if (newScore <= 0)
+                        if (newScore < 0)
```
adQuid, line 209: Are you sure this should be less than zero? I recall the logic being strictly positive before...

GameDungeon: I think I will revert that. I think I misremembered two species that score the same being 0 instead of 1.

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -157,7 +157,8 @@ public bool RunStep(RunResults results)
         {
             mutation.Item2.OnEdited();
 
-            newMiche.InsertSpecies(mutation.Item2, patch, scores, cache, false, workMemory);
+            if (newMiche.InsertSpecies(mutation.Item2, patch, scores, cache, false, workMemory))
+                scores[mutation.Item2] = 0;
```
hhyyrylainen, line 161: I'll accept this change as a base, but when I was testing stuff locally I added the score setting *before* calling this method. That's because I thought that the insert species would need to be able to see the species score *before* being able to insert it?

hhyyrylainen (approved):
The RunResults change doesn't seem to logically change anything. But if this improves auto-evo working better I'll merge this right away to get master into a more working state.

src/auto-evo/steps/ModifyExistingSpecies.cs
hhyyrylainen: I removed this change and did a different fix that in my testing now can run a bunch of generations in the auto-evo exploring tool without errors: https://github.com/Revolutionary-Games/Thrive/pull/5378

PR #5388, opened 2024-08-22
Fix Miche Bugs

### LINKED ISSUE #5382: Find a way to avoid `Miche.DeepCopy` in new auto-evo
As it recursively allocates quite a bit of memory

### LINKED ISSUE #5385: Double check all auto-evo exploring tool settings and see if they are still used by the new auto-evo
and remove / update the new auto-evo if they aren't used (to either remove them or make them be used again).

GameDungeon:

Closes #5385
Closes #5382 

Fixes the large miche bugs

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I think I'll merge this for the important fixes but I found one pretty big problem: the auto-evo now takes over 5 minutes at generation 20. Perhaps the way species are allowed to migrate way more than before (multiple attempts per patch for *each* species) is the cause?

src/auto-evo/AutoEvoRun.cs
```diff
@@ -407,17 +407,19 @@ protected virtual void GatherInfo(Queue<IRunStep> steps)
         {
             steps.Enqueue(new ModifyExistingSpecies(entry.Value, new SimulationCache(worldSettings), worldSettings,
                 random));
+        }
 
-            steps.Enqueue(new MigrateSpecies(entry.Value, new SimulationCache(worldSettings), random));
+        foreach (var species in allSpecies)
+        {
+            steps.Enqueue(new MigrateSpecies(species, map, worldSettings, new SimulationCache(worldSettings), random));
```
hhyyrylainen, line 414: Is it absolutely necessary that no migration steps are interspersed between the modify steps? They should not conflict, right?

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -199,45 +189,13 @@ public bool RunStep(RunResults results)
                 // Add these mutant species into a new miche to test them
                 foreach (var mutation in mutationsToTry)
                 {
-                    micheForMutationTests!.InsertSpecies(mutation.Item2, patch, null, cache, false, workMemory);
+                    miche.InsertSpecies(mutation.Item2, patch, null, cache, false, workMemory);
```
hhyyrylainen, line 192: Doesn't this modify the miche state now? Isn't this like super bad? I did test in game and things didn't crash with an exception but I wouldn't expect this to work at all...

hhyyrylainen:
I took all the good parts I think don't need further consideration and merged them separately. I've now rebased this onto master after those changes so this is much smaller and only really has the mutation step change, which I'm still doubtful about, because it seems like removing that one miche deep copy causes a lot of complications to the architecture. Like this I think it is just better to keep the miche deep copy for the time being until a better approach to removing that can be come up with.

src/auto-evo/steps/ModifyExistingSpecies.cs
GameDungeon: It does, but nothing outside population actually calls miche after that so it does not lead to any issues.

GameDungeon:
The reason the register species exists now isn't actually related to the miche deep copy. Weirdly that was the simplest change. Register Species was created so the step would look at all species globally to decide the best mutation from there, otherwise a species could generate multiple mutations and they would simply overwrite each-other.

src/auto-evo/steps/ModifyExistingSpecies.cs
hhyyrylainen: But the mutations then *do* affect the population number simulation, which is different from before and makes the auto-evo prediction much less accurate.

I think I'd prefer to just have the deepcopy back and later consider something like implementing a separate miche type that has cached child nodes and it could be reinitialized to restore back to the default state. I think that would be easier than implementing like a overlay feature that allowed temporary species to be applied on top of the miche tree without affecting the underlying tree.

hhyyrylainen:
Wouldn't a simpler solution be that the species result could store a score and then when a new mutation tries to be added as the result, that is only added if its score is better than before?

GameDungeon:
Given handling threading, having to turn the old result into a spilt, dealing with the population gained from the mutation properly, and generally handling results raw (that gave me same pain in other methods I tried first) I'm not sure it ends up that much simpler. Also with register you get a global post processing step which in theory would be used for other things.

hhyyrylainen:
I guess it does then make more sense in context of the current algorithm, but I have to say I don't like the fact that this will result in even more temporary required memory allocations. 

Would it make more sense to rewrite the mutation step to be per-species rather than per-patch (so the way it was previously in auto-evo)? That way all the species data can be processed at once and duplicate mutations can be pruned a lot earlier (and for speed we could also reduce how many times each mutation strategy is allowed to run on a species).

src/auto-evo/RunResults.cs
```diff
@@ -7,6 +7,9 @@
 using System.Linq;
 using Godot;
 using JetBrains.Annotations;
+using PossibleSpecies =
+    System.Tuple<Species, System.Collections.Generic.IEnumerable<System.Collections.Generic.KeyValuePair<Patch, long>>,
```
hhyyrylainen, line 11: Couldn't this be a named tuple that would make the code in [RegisterNewSpecies.cs](https://github.com/Revolutionary-Games/Thrive/pull/5388/files#diff-773e52f10083a2dbed39bb1156158836e8e8e94b219d9e74dc54fa238fc9407d) much easier to follow?

GameDungeon: C# won't let me type alias that. I could turn it into a struct and do it that way.

src/auto-evo/steps/ModifyExistingSpecies.cs
hhyyrylainen: Ah, also I realized one other thing is that the reason why this doesn't blow up is that each of these steps is handling just one patch meaning they aren't messing with any miche that another step would be using at the same time.

src/auto-evo/RunResults.cs
```diff
@@ -182,21 +211,6 @@ public void AddNewSpecies(Species species, IEnumerable<KeyValuePair<Patch, long>
...
 
-    public void AddSplitResultForSpecies(Species species, Species splitSpecies, List<Patch> patchesToConvert)
-    {
-        if (patchesToConvert == null || patchesToConvert.Count < 1)
-            throw new ArgumentException("split patches is missing", nameof(patchesToConvert));
-
-        MakeSureResultExistsForSpecies(species);
-        MakeSureResultExistsForSpecies(splitSpecies);
-
-        results[species].SplitOff = splitSpecies;
```
hhyyrylainen, line 193: As this is deleted, how are species now marked as being split?

src/auto-evo/RunResults.cs
hhyyrylainen: I think that would be better. In the first place needing to have this using to make things readable kind of indicates to me that this is pretty crazy amount of parameters to pack together like this in the places this is used.

src/auto-evo/RunResults.cs
GameDungeon: It is done through AddNewSpecies. The population is init through the population sim. .SplitOff probably needs a full rework as it cannot handle two species splitting off of one anyway. AddNewSpecies is how the miche system has done it the whole time, with split result being unused. As far as the gui goes NewlyCreated is set properly so everything displays how it should.

hhyyrylainen: I didn't realize that the GUI read the species split from the other side. I guess it would be now then better for clarity to remove the split off property entirely from the species result if it is no longer working / used.

GameDungeon:
Species overwrite each other in the miche tree constantly, so a per species algorithm would need to be able to handle pruning their new species as they get overwritten, as well as the complexity of handling all of the patches miche trees at the same time. I believe that doing it per-species would end up skyrocketing complexity (and possibly memory)

hhyyrylainen:
I guess it is true that in that version each patch would need to have the miche tree cloned by each per-species step. I guess in that context the situation with needing to generate a lot of extra mutations that are kept around in lists and even more allocated lists of populations in patches is the better way to go (for now at least in case an even better architecture can be realized in the future).

src/auto-evo/RunResults.cs
```diff
@@ -182,21 +208,6 @@ public void AddNewSpecies(Species species, IEnumerable<KeyValuePair<Patch, long>
...
-        results[species].SplitOffPatches = patchesToConvert;
```
hhyyrylainen, line 194: Looking at this more, there's a ton of code that depends on SplitOffPatches. So I'm highly super questioning that things are very broken related to that due to the new auto-evo and this situation needs to be fixed. I don't think I have time to fix the 23 errors that come from commeting out the split off and split off patches variables.

hhyyrylainen (approved):
I still have some concerns but I think this needs to be merged now so that there's at least one more chance for you to double check what I've changed and maybe if you have time to make some more fixes by tomorrow to include in the RC build.

PR #5394, opened 2024-08-24
Add resizing metaballs

### LINKED ISSUE #5392: Implement a metaball size slider for the macroscopic editor
With the convolution surfaces in it'd be pretty nice to add a slider for the placed metaball size to give players some more tools to at least make pretty interesting looking creatures even if you cannot really play with them yet.

CI09:

This PR does some stuff...

**Related Issues**

Closes #5392

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09:
Hold shift and scroll to resize

hhyyrylainen:
There should be a bar in the GUI with hover tooltip and a short label to set the metaball size.

CI09:
Isn't scroll more popular amongst similiar games. It is also easier to handle in my opinion

hhyyrylainen:
Well the scroll can be kept, but the average player would have 0 clue that was even a feature they could use.

CI09:
How can I make a label somewhere on screen that doesn't feel out of place?

hhyyrylainen:
I would personally try putting it in the bottom middle or on the right side above the confirm button.

Also one more thing to consider that just popped into my head is that holding shift and dragging with the middle mouse pans the camera. I guess that doesn't really conflict if doing that camera movement it is possible then to accidentally change the metaball size. Though I won't really know if that happens accidentally or if it feels bad before testing.

CI09:
image

hhyyrylainen:
Is that text like on screen all the time? To me that doesn't seem that professional looking. And it would take the same amount of space to just put in a slider to select the metaball size.

One more thing I think is true that all other editor features are usable purely with the mouse, so this feature would be one that makes it impossible to use the editor with purely a mouse.

CI09:
image

hhyyrylainen:
I think the tooltip should explain the slider and only on a second line explain the alternative way to control the size with the mouse keybinding (which BTW would be pretty nice to allow rebinding in the inputs settings).

CI09:
Shift is bound to e_pan_mode and scroll I don't know if anyone would want to rebind

hhyyrylainen:
I meant that it'd be better to create a custom action for the increase metaball size and decrease the size (it could be named like e_increase_size and e_decrease_size) that allows rebinding this specific input.

CI09:
Reject scroll?

hhyyrylainen:
What do you mean? I meant that you shouldn't force the size inputs to be related to the zoom input by adding a new input action type for increasing and decreasing the size.

CI09:
Should I keep disabling zooming when holding shift?

hhyyrylainen:
That sounds like something that should properly be fixed in the input system. Would be probably easier for now to make the size increase / decrease bound to W / S keys as I don't think those are currently used in the editor.

PR #5397, opened 2024-08-24
Optimize SCALIS

dligr:

Optimizes SCALIS by switching to a better cutoff algorithm. More specifically, it is now based on square distance to line's center, which is cheaper to calculate than distance to closest point on line (which it was previosly based on).

In my testing, there was an x2 increase in performance.
I.e. around 0.13s -> 0.06s on the default multicellular creature.

**Related Issues**

SCALIS was too slow

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/late_multicellular_stage/Scalis.cs
+++ b/src/late_multicellular_stage/Scalis.cs
@@ -133,7 +133,7 @@ public class Scalis : IMeshGeneratingFunction
             }
         }
 
-        return value / 2.0f;
+        return value / NormalizationFactor(i, sigma);
     }
 
     public Color GetColour(Vector3 pos)
--- a/src/late_multicellular_stage/Scalis.cs
+++ b/src/late_multicellular_stage/Scalis.cs
@@ -28,7 +28,7 @@ public class Scalis : IMeshGeneratingFunction
     /// <remarks>
     ///   <para>
     ///     i-th element here refers to the i-th point's bone in points[]. If that point has no parent, then it has no
-    ///     bone, so the distance in this array is meningless for the point.
+    ///     bone, so the distance in this array is meaningless for the point.
     ///   </para>
     /// </remarks>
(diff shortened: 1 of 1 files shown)


src/late_multicellular_stage/Scalis.cs
```diff
@@ -36,6 +46,26 @@ public Scalis(IReadOnlyCollection<MulticellularMetaball> metaballs)
             points[i] = point;
             ++i;
         }
+
+        boneMaxSquareDistanceCache = new float[points.Length];
+
+        for (int j = 0; j < points.Length; j++)
```
hhyyrylainen, line 52: Can't this loop reuse the i variable so that this then looks like a simple "for i" loop?

`        for (i = 0; i < points.Length; ++i)`

src/late_multicellular_stage/Scalis.cs
```diff
@@ -36,6 +46,26 @@ public Scalis(IReadOnlyCollection<MulticellularMetaball> metaballs)
...
+        {
+            if (points[j].Parent == null)
+                continue;
+
+            float maxDistanceFromCenter = (points[j].Position - points[j].Parent!.Position).Length() * 0.5f
```
hhyyrylainen, line 57: It might be slightly faster to read the points[j] into a local variable once and then reusing it here as this potentially looks up the value 4 times from the list.

src/late_multicellular_stage/Scalis.cs
dligr: Yes, this is probably a better way

hhyyrylainen (approved):
I don't get as good improvement in speed, but I still do get the improvement. I fixed a few minor code stuff and a comment word typo to polish this up for merging.

With this I get about 0.075s as the generation time whereas before it was around 0.098s.

Issue #5391, opened 2024-08-23
Fix whatever causes the auto-evo prediction to show the player to predate on themselves


hhyyrylainen:
Leads to pretty funny energy breakdowns:

![2024-08-23_15 04 27 6693](https://github.com/user-attachments/assets/e6a1cea9-e814-4124-b383-a221d504729f)

PR #5398, opened 2024-08-24
Miche Population Calculation Fixes

GameDungeon:

Fixes:
- [ ] Auto-Evo prediction incorrect
- [x] Species gain energy from failed Miches Closes:  #5391
- [x] Fix Bug that makes auto-evo think there is iron in every patch

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/Miche.cs
+++ b/src/auto-evo/Miche.cs
@@ -254,14 +254,14 @@ public class Miche
     {
         public HashSet<Species> WorkingHashSet = new();
 
-        private List<Dictionary<Species, float>> ScoreDictionaries = new();
+        private readonly List<Dictionary<Species, float>> scoreDictionaries = new();
 
         public Dictionary<Species, float> GetScoresAtDepth(int depth)
         {
-            while (ScoreDictionaries.Count <= depth)
-                ScoreDictionaries.Add(new Dictionary<Species, float>());
+            while (scoreDictionaries.Count <= depth)
+                scoreDictionaries.Add(new Dictionary<Species, float>());
 
-            return ScoreDictionaries[depth];
+            return scoreDictionaries[depth];
         }
     }
 }
--- a/src/auto-evo/Miche.cs
+++ b/src/auto-evo/Miche.cs
@@ -125,6 +125,26 @@ public class Miche
         newChild.Parent = this;
     }
 
+    public int GetDeepestDepth()
+    {
+        return GetDeepestDepth(1);
+    }
+
+    public int GetDeepestDepth(int depth)
+    {
+        if (IsLeafNode())
+            return depth;
+
+        var maxDepth = depth;
+
+        foreach (var child in Children)
+        {
+            maxDepth = Math.Max(maxDepth, child.GetDeepestDepth(depth + 1));
+        }
+
+        return maxDepth;
+    }
+
     /// <summary>
     ///   Inserts a species into any spots on the tree where the species is a better fit than any current occupants
     /// </summary>
@@ -136,11 +156,12 @@ public class Miche
     /// <param name="cache">Data cache for faster calculation</param>
     /// <param name="dry">If true the species is not inserted but only checked if it could be inserted</param>
     /// <param name="workingMemory">Temporary working memory to use by this method, automatically cleared</param>
+    /// <param name="depth">Depth in the miche tree</param>
     /// <returns>
     ///   Returns a bool based on if the species was inserted into a leaf node
     /// </returns>
     public bool InsertSpecies(Species species, Patch patch, Dictionary<Species, float>? scoresSoFar,
-        SimulationCache cache, bool dry, HashSet<Species> workingMemory)
+        SimulationCache cache, bool dry, InsertWorkingMemory workingMemory, int depth = 0)
     {
         var myScore = Pressure.Score(species, patch, cache);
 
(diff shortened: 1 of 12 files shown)


src/auto-evo/simulation/MichePopulation.cs
```diff
@@ -213,15 +213,12 @@ private static void SimulatePatchStep(SimulationConfiguration simulationConfigur
...
         var leafNodes = new List<Miche>();
-        miche.GetLeafNodes(leafNodes, x => x.Occupant != null);
+        miche.GetLeafNodes(leafNodes, x => x.Occupant is MicrobeSpecies);
 
         // TODO: check if energy should be calculated as doubles because the summed numbers can get pretty high that
         // might benefit from extra precision
         var energyDictionary = species.ToDictionary(x => x, _ => 0.0f);
 
-        // This doesn't use miche.SetupScores as this wants scores for all species, even ones that aren't in the
-        // tree yet
```
hhyyrylainen, line 223: Why was this comment removed?

src/auto-evo/simulation/MichePopulation.cs
```diff
@@ -213,15 +213,12 @@ private static void SimulatePatchStep(SimulationConfiguration simulationConfigur
             return;
 
         var leafNodes = new List<Miche>();
-        miche.GetLeafNodes(leafNodes, x => x.Occupant != null);
+        miche.GetLeafNodes(leafNodes, x => x.Occupant is MicrobeSpecies);
```
hhyyrylainen, line 216: I'm pretty worried about this change because once we add other species types to auto-evo I reckon this might be forgotten and wreak havoc. What's the reason behind this change?

src/auto-evo/simulation/MichePopulation.cs
GameDungeon: You removed miche.SetupScores as a function, so I figured we no longer needed to explain why it was not used.

hhyyrylainen: Okay that makes sense.

src/auto-evo/simulation/MichePopulation.cs
hhyyrylainen: Reading the later code I now think I understand this but this should get a pretty big TODO and warning slapped on that this will be majorly wrong once other than microbe species can be fed into auto-evo.

GameDungeon: We need to make sure that a later call to the return from this is not null or not a microbe species. I && the microbe species check to this, but the lint wanted me to simplify it. Autoevo is currently littered with checks like this as Multicell species are partially supported. I doubt that this specific check would be missed as there are so many other like it I would probably just crtl + f "is MicrobeSpecies".

hhyyrylainen: Could you still add the TODO comment for my piece of mind here?
In most cases in my head it is that Species is usually used but just before microbe specific stuff is done it is then checked, but in this situation the other species types are pruned from the data much earlier.

GameDungeon: I can add that TODO

src/microbe_stage/systems/ProcessSystem.cs
GameDungeon, line 1: That is a cursed way for git to merge that

src/auto-evo/steps/MigrateSpecies.cs
```diff
@@ -99,7 +97,9 @@ public bool RunStep(RunResults results)
                 var moveAmount = (long)random.Next(population * Constants.AUTO_EVO_MINIMUM_MOVE_POPULATION_FRACTION,
                     population * Constants.AUTO_EVO_MAXIMUM_MOVE_POPULATION_FRACTION);
 
-                if (moveAmount > 0 && targetMiche.InsertSpecies(species, patch, null, cache, true, speciesWorkMemory))
+                var insertWorkMemory = new Miche.InsertWorkingMemory(targetMiche);
```
hhyyrylainen, line 100: Why was this made a local variable? I'd prefer the field approach as it ensures the list isn't recreated each time in this loop which burns through a bunch of unnecessary memory allocations.

src/auto-evo/Miche.cs
```diff
@@ -246,4 +266,21 @@ public override int GetHashCode()
         return Pressure.GetHashCode() * 131 ^ parentHash * 587 ^
             (Occupant == null ? 17 : Occupant.GetHashCode()) * 5171;
     }
+
+    public class InsertWorkingMemory
```
hhyyrylainen, line 270: Documenting this class and making this reusable for new miches would help a lot in this class' usage.

src/auto-evo/Miche.cs
```diff
@@ -156,26 +177,25 @@ public bool InsertSpecies(Species species, Patch patch, Dictionary<Species, floa
             return true;
         }
 
-        // TODO: somehow avoid this recursive method call memory allocations
-        var newScores = new Dictionary<Species, float>();
+        var newScores = workingMemory.ScoreDictionaries[depth];
```
hhyyrylainen, line 180: Couldn't the cache be made so that this automatically increased the dictionary size if depth is higher? That way the dictionary wouldn't need to be pre-populated making it easier to use? Or could maybe even a stack (or maybe a queue) type be used instead of a dictionary? Or maybe just a plain list which is indexed by the depth? Dictionary lookups are more expensive than list index lookups.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -427,73 +427,76 @@ public static (float Production, float Consumption) CalculateOrganelleATPBalance
...
+        return (processATPProduction, processATPConsumption);
+    }
 
-                foreach (var input in processData.WritableInputs)
+    /// <summary>
+    ///   Calculates the maximum speed a process can run at in a biome based on the environmental compounds.
+    ///   Can be switched between the average, maximum etc. conditions that occur in the span of an in-game day.
+    /// </summary>
+    public static ProcessSpeedInformation CalculateProcessMaximumSpeed(TweakedProcess process,
+        BiomeConditions biome, CompoundAmountType pointInTimeType, bool checkIfInPatch)
```
hhyyrylainen, line 456: I think some other name for the parameter is needed. Something like `bool requireInputCompoundsInBiome`?

hhyyrylainen (approved):
I added some slight code cleanups and verified the self eating is now gone. I did see a random bug with the jumping into the game from the auto-evo tool where my species got 0 energy no matter what I did until I exited the editor and got back to it, but that's probably a totally unrelated bug to these changes.

PR #5409, opened 2024-08-26
Iron chunks now use visual identificator

### LINKED ISSUE #5402: Make chunk visuals load through the VisualResourceIdentifier instead of exact paths
as this will make renaming files possible without breaking save compatibility

https://github.com/Revolutionary-Games/Thrive/blob/4174a44e6ca7e02e71bf47b59db5771af3d3f808/src/microbe_stage/Spawners.cs#L225

CI09:

This PR does some stuff...

**Related Issues**

Closes #5402

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/microbe_stage/biomes.json
+++ b/simulation_parameters/microbe_stage/biomes.json
@@ -46,19 +46,19 @@
           "Name": "SMALL_IRON_CHUNK",
           "Meshes": [
             {
-              "ScenePath": "res://assets/models/IronRock1.tscn",
+              "VisualResource": 9,
               "ConvexShapePath": "res://assets/models/IronRock1.shape"
             },
             {
-              "ScenePath": "res://assets/models/IronRock2.tscn",
+              "VisualResource": 10,
               "ConvexShapePath": "res://assets/models/IronRock2.shape"
             },
             {
-              "ScenePath": "res://assets/models/Iron3.tscn",
+              "VisualResource": 11,
               "ConvexShapePath": "res://assets/models/Iron3.shape"
             },
             {
-              "ScenePath": "res://assets/models/Iron4.tscn",
+              "VisualResource": 12,
               "ConvexShapePath": "res://assets/models/Iron4.shape"
             }
           ],
@@ -82,11 +82,11 @@
           "Name": "BIG_IRON_CHUNK",
           "Meshes": [
             {
-              "ScenePath": "res://assets/models/Iron5.tscn",
+              "VisualResource": 13,
               "ConvexShapePath": "res://assets/models/Iron5.shape"
             },
             {
-              "ScenePath": "res://assets/models/IronRockLarge2.tscn",
+              "VisualResource": 14,
               "ConvexShapePath": "res://assets/models/IronRockLarge2.shape"
             }
           ],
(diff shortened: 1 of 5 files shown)


simulation_parameters/microbe_stage/biomes.json
```diff
@@ -46,19 +46,19 @@
           "Name": "SMALL_IRON_CHUNK",
           "Meshes": [
             {
-              "ScenePath": "res://assets/models/IronRock1.tscn",
+              "VisualResource": 9,
```
hhyyrylainen, line 49: Yeah this, is not very nice because people need to manually write this file so using string keys here would be vastly superior.

When referring to the integer values I meant how this will be saved in the game saves, which are not edited by humans so the compactness of just pure numbers helps there.

src/microbe_stage/ChunkConfiguration.cs
```diff
@@ -146,6 +146,8 @@ public class ChunkScene
         /// </summary>
         public string ScenePath;
```
hhyyrylainen, line 147: This and the loaded scene have to be removed in order to not cause confusion by having 2 separate mechanisms at once specifying how the visual is defined.

src/microbe_stage/Spawners.cs
```diff
@@ -222,14 +222,11 @@ public static EntityRecord SpawnChunkWithoutFinalizing(IWorldSimulation worldSim
...
-        // {
-        //     VisualIdentifier = VisualResourceIdentifier.AgentProjectile,
-        // });
-        entity.Set(new PathLoadedSceneVisuals
+        var meshes = chunkType.Meshes;
+
+        entity.Set(new PredefinedVisuals
         {
-            ScenePath = selectedMesh.ScenePath,
+            VisualIdentifier = meshes[random.Next(meshes.Count)].VisualResource,
```
hhyyrylainen, line 229: This should use the selected mesh because otherwise the visual mesh may not match what was selected as the physics mesh.

hhyyrylainen (changes requested):
Here's some comments. And also when reading this I realized that the `CollisionShapeLoader` component has a hardcoded path to a physics data object. Meaning that just these changes aren't sufficient to protect the chunks from save breakage when moving resource files around.

src/microbe_stage/ChunkConfiguration.cs
CI09: This will cause to at least chunks on microbe death not function. Last time I tried to do it with fixing said issue, it froze my game.

simulation_parameters/microbe_stage/biomes.json
CI09: So what do I put here instead if visual id is `BigIronChunk1`?

src/microbe_stage/ChunkConfiguration.cs
hhyyrylainen: Those chunks must also be updated to the new system. I won't have the confusion of having 2 systems for cell chunk visuals.

simulation_parameters/microbe_stage/biomes.json
hhyyrylainen: `"BigIronChunk1"`
which either works immediately or a custom JSON converter is needed that supports both string and integer values.

CI09: JSON converter is needed then. Unfortunately.

hhyyrylainen: I just looked and so far it seems like no enum values have been wanted to read from strings in the game. The RevolutionaryWebApp repo has that setup already, but sadly it uses a different JSON lib, so a separate approach is needed here.

PR #5410, opened 2024-08-26
Remove Old Mutation Code

GameDungeon:

This removes the old mutation code. I believe that this breaks save compat, unless removing a field is handled somehow. 

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


simulation_parameters/SimulationParameters.cs
hhyyrylainen, line 1: Shouldn't the random chance lists be also removed as if I remember right these were the only uses for those?

src/auto-evo/mutations/CommonMutationFunctions.cs
```diff
@@ -53,6 +55,46 @@ public enum Direction
...
+    public static MicrobeSpecies GenerateRandomSpecies(MicrobeSpecies mutated, MutationWorkMemory workMemory,
+        Random random, float mp = 300)
+    {
+        var mutationStrategy = new AddOrganelleAnywhere(_ => true);
+
+        GameWorld.SetInitialSpeciesProperties(mutated, workMemory.WorkingMemory1, workMemory.WorkingMemory2);
+
+        while (mp > 0)
+        {
+            var mutationList = mutationStrategy.MutationsOf(mutated, mp, true)?.OrderBy(_ => random.Next()).ToList();
```
hhyyrylainen, line 67: Wouldn't FirstOrDefault be better here? That avoids potentially building a list of 100 items but then discarding all but the first one.

hhyyrylainen:
This is a good change to do to remove outdated code.

I don't see why this would break save compatibility as this just removes a property which when loaded from a save will be ignored in the future. Or does the removal of a class the JSON loader would want to instantiate cause issues?

GameDungeon:
I didn't know how that was handled so I just put a warning to make sure. I guess I should probably just test it.

GameDungeon:
Ok, I was wrong, everything loads fine

simulation_parameters/SimulationParameters.cs
GameDungeon: I tried removing those and everything breaks, so I decided to ignore those for now.

hhyyrylainen: ??? how does that happen? Wouldn't it be better to figure out why and remove all the outdated code that is no longer needed at once?

GameDungeon: When I get home later I can explore it more, but I think the entire organelles simulation instance was null or something like that.

GameDungeon: I can no longer reproduce the issues A had before, so I must have changed something else by mistake.

hhyyrylainen (changes requested):
There's a pretty major issue now: the species generated are random now instead of always the same (as I think was the case previously in the benchmark). This will make benchmark results less consistent than before (not that they were super consistent before).

hhyyrylainen:
For the record I checked that the `random` state is the same before the `GenerateRandomSpecies` method is called. So there has to be something in there that doesn't consistently use random state.

GameDungeon:
AddOrganelleAnywhere's random is not passed in, which is the issued.

hhyyrylainen:
That sounds very plausible. Would you be able to make that change? That would be great for this to then be merged as keeping the benchmark at least a bit more consistent would be excellent.

GameDungeon:
I can in a couple hours from now.

hhyyrylainen:
I'll try to find a good new seed for the benchmark as it now causes many big cells that start dying and playing the damage sound really loudly.

hhyyrylainen (approved):
I couldn't find an absolutely perfect seed so I've opened up a follow up issue but I'll merge this now as it is much better to have a consistent benchmark species than not have that: https://github.com/Revolutionary-Games/Thrive/issues/5424

PR #5411, opened 2024-08-26
Fixed various problems with too wide translations

hhyyrylainen:
the added cell stats indicator wrap may cause problems for English in the future but hopefully the minimum size is fine to avoid most issues

This PR does some stuff...

**Related Issues**

locale issues with too wide text reported as is usual after an RC

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/CellStatsIndicator.tscn
+++ b/src/microbe_stage/editor/CellStatsIndicator.tscn
@@ -27,9 +27,11 @@ stretch_mode = 5
 
 [node name="Description" type="Label" parent="."]
 editor_description = "PLACEHOLDER"
+custom_minimum_size = Vector2(130, 0)
 layout_mode = 2
 text = "CELL_STAT_PLACEHOLDER"
 label_settings = ExtResource("2_756k6")
+autowrap_mode = 2
 
 [node name="HBoxContainer" type="HBoxContainer" parent="."]
 layout_mode = 2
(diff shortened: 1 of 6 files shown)


src/microbe_stage/editor/CellStatsIndicator.tscn
```diff
@@ -27,9 +27,11 @@ stretch_mode = 5
 
 [node name="Description" type="Label" parent="."]
 editor_description = "PLACEHOLDER"
+custom_minimum_size = Vector2(130, 0)
 layout_mode = 2
 text = "CELL_STAT_PLACEHOLDER"
 label_settings = ExtResource("2_756k6")
+autowrap_mode = 2
```
CI09, line 34: Is there any particular reason it is not mode 3?

hhyyrylainen: The label is so short that I was thinking it would be better to force horizontal scrolling rather than cause a word like "digestion" be split at a random character.

CI09: Looks good then

CI09 (approved):

CI09:
Actually I should've asked first how did you get issue where stuff doesn't fit

hhyyrylainen:
If you set the min x width to just 80 pixels (instead of the 130, I think, which I set it to) and keep the word wrap on, then it starts to word wrap even some of the longer English text on the cell stats.

CI09:
If you want you can merge this but when I resized container it only expanded horizontal scroll

hhyyrylainen:
Where specifically? Could you post a screenshot?

CI09:
image

hhyyrylainen:
But that can only happen if the container is set to be narrower than it is currently. So that's just a general thing whoever works on the GUI next and thinks they can squeeze the panel to be narrower, right? That's not something a player could trigger no matter what they do.

PR #5419, opened 2024-08-27
Fix timeline cleaning bug

### LINKED ISSUE #3135: Timeline history cleaning logic has a bug (in global timeline view)
It's been reported that the timeline always removes events at 2100 Myr to 2700 instead of actually removing the oldest stuff like it should. Probably the way it detects the oldest things to delete is wrong.

dligr:

Fixes the timeline cleaning bug, i.e. the latest timeline entries being removed instead of the oldest. Apparently, the cleaning itself had no bugs (the oldest entries were correctly removed), but the newest entries were displayed in place of the removed old entries. Which was caused by foreach not ordering key-value pairs.

2200 MYR save for easier testing:
[2200.zip](https://github.com/user-attachments/files/16761785/2200.zip)

**Related Issues**

Closes #3135

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/TimelineTab.cs
+++ b/src/microbe_stage/editor/TimelineTab.cs
@@ -110,7 +110,7 @@ public partial class TimelineTab : PanelContainer
             globalEventsContainer.FreeChildren();
             cachedGlobalTimelineElements = new List<TimelineSection>();
 
-            foreach (var entry in editor.CurrentGame.GameWorld.EventsLog)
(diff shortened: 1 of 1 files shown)


hhyyrylainen (approved):
This does look like it solves the problem.
I guess it is not avoidable that this now needs to allocate more temporary memory, but as this is in the code that creates a bunch of GUI objects anyway that doesn't seem too bad (and I can't see an easy way to avoid it either).

PR #5422, opened 2024-08-27
Add input modifier text to graphical input keys

### LINKED ISSUE #5396: Input actions with a modifier and an icon representing them in the inputs menu doesn't display the modifier at all
image

Edit by hhyyrylainen: it looks like to me like the modifier isn't added when an icon can be generated for the action, so the bug is likely in the code generating that image then not adding text before it.

dligr:

Adds input modifier text (for shifts, ctrls, etc.) to input keys that use graphical representation (mouse buttons and such).

**Related Issues**

Fixes #5396

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/engine/input/key_mapping/SpecifiedInputKey.cs
```diff
@@ -194,10 +194,41 @@ public static (JoyButton Code, int Device) UnpackCodeAndDevice(ulong packed)
         return ((JoyButton)code, device);
     }
 
-    public Control GenerateGraphicalRepresentation()
+    public Control GenerateGraphicalRepresentation(LabelSettings labelSettings)
     {
         var container = new HBoxContainer();
 
+        if (Shift || Control || Alt)
```
hhyyrylainen, line 201: I think this should only be done if the event type is not a controller button, as an input like "shifit + controller button" doesn't make sense (and hopefully is prevented by the binding system from being generated).

src/engine/input/key_mapping/SpecifiedInputKey.cs
```diff
@@ -10,6 +10,9 @@ public class SpecifiedInputKey : ICloneable
 {
     private StringBuilder? toStringBuilder;
 
+    private StringName? marginLeftName;
+    private StringName? marginRightName;
```
dligr, line 14: I think that this should be marginTopName

hhyyrylainen (approved):
I added a tiny bit of sanity checking and string name caching. I think this now sufficiently solves the issue for merging. Though like I noted in my earlier comment the spacing related to the icon isn't the best.

PR #5427, opened 2024-08-29
Some auto-evo technical improvements

hhyyrylainen:

Slightly improves performance and some code stuff (I won't say that this changes the architecture much).

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


GameDungeon:
All of these changes look good to me. We really need a way to properly benchmark auto Evo performance. (and actually support seeds)

hhyyrylainen:
Yeah, that'd be pretty nice. Would be nice to have like an overall hash of the final species to be able to quickly verify that changes don't happen when making changes that shouldn't affect auto-evo results. 

So far I've used the auto-evo exploring tool to run to generation 10, then 15 and finally 20 and checking the elapsed time of the latest run (and also memory usage). Though that is not super accurate but I think it does show like the overall impact of the changes if the impact is big enough.

PR #5438, opened 2024-09-01
Patch dynamics

CI09:

I felt confident enough to make this PR now. It's not perfect but stuff like oxidation happens and glucose drops (less than before but still) or rises if overall population decreases.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


dligr:
After some testing, I can confirm that overall the dynamics work. I have noticed that in patches with photosynthesizers, the amount of oxygen increases, while in the patches with no photosynthesizers, the amount of CO2 increases. Compound amount fluctuate too.

One issue I've noticed is that in patches with sunlight, after oxygen reaches 100%, most of the compounds go NaN, like on the picture:
image

I'd also like to add that ideally, at least in my opinion, it would probably be better to separate microbes' environmental effects from their gameplay-based processes. So, how I imagine it is that microbes' processes will have "environment-affecting" inputs and outputs, separate from the gameplay-affecting ones. It would allow us to fine-tune environmental effects without affecting gameplay, and vice versa.

CI09:
It's probably caused by environmental compounds not adding up to perfectly 100%. We might need to move away from percentages before it gets merged or do some math here

hhyyrylainen:
Does this PR not have nearby patches diffusing into each other (the previous attempts at this feature had that)? I think that would be pretty important to have to ensure there cannot be truly massive differences between patches.

src/microbe_stage/BioProcess.cs
```diff
@@ -64,6 +64,25 @@ public void Check(string name)
...
+        var rate = 1.0f;
+
+        foreach (var input in Inputs)
+        {
+            if (input.Key.IsEnvironmental)
+            {
+                foreach (var compound in conditions.AverageCompounds)
+                {
+                    if (compound.Key.Equals(input.Key))
+                        rate = compound.Value.Ambient / rate;
```
dligr, line 78: I think that the problem with NaN compounds comes from here. To get NaN, you usually have to divide by zero, which seems possible here. If `compound.Value.Ambient` = 0, then on the next iteration, a number will be divided by rate, which is 0. Maybe there was supposed to be multiplication here, instead of division?

Deus-Codes:
Haven't playtested, but skimmed through some notes here. Good on you for taking on an ambitious task, and I do think it is best to take it compound by compound rather than having every single thing by dynamic at once. 

Here are some background notes on what we kind of are aiming for in terms of the progression of compounds: https://forum.revolutionarygamesstudio.com/t/evolution-of-early-planets/445/21

And here is some writing on the topic by Buckly which might be useful, though not sure how applicable it is: https://forum.revolutionarygamesstudio.com/t/the-role-of-life-in-shifting-environmental-conditions/1076/1

CI09:
So now that other section in compound composition PR is merged, what needs to be done to this PR?

hhyyrylainen:
I'd really like to just see an initial oxygen production and diffusion between patches. Once we have that I think we'd be in a much better place to discuss the full dynamic compound balances (based on species locked resources) in patches that this PR has.

src/microbe_stage/BioProcess.cs
```diff
@@ -64,6 +64,25 @@ public void Check(string name)
...
+    {
+        var rate = 1.0f;
+
+        foreach (var input in Inputs)
+        {
+            if (input.Key.IsEnvironmental)
+            {
+                foreach (var compound in conditions.AverageCompounds)
+                {
+                    if (compound.Key.Equals(input.Key) && rate > 0)
```
dligr, line 77: I think that this can be simplified by getting compound with `conditions.AverageCompounds.TryGetValue()` instead of `foreach`, as the result should be, as I understand, the same, but with less calculations

src/microbe_stage/BioProcess.cs
```diff
@@ -64,6 +64,25 @@ public void Check(string name)
...
+                        rate = compound.Value.Ambient / rate;
```
dligr, line 78: I'm also not entirely sure about this formula. If its purpose is to scale the process rate with compound concentrations, then there should probably be multiplication instead of division.
For example, a compound concentration of 0.01 would increase the rate 100x on the next iteration, while it is supposed to actually decrease it 100x, if I understand it correctly.

dligr (commented):
I want to add that sunlight and temperature are able to diffuse throught patches:
image

I also still think that there should be a separation between gameplay-based processes environment-affecting processes, for the reasons I stated in one of my previous comments. It would also help with only simulating dynamics of some compounds at first, as this approach would be more easily tweakable. The implementation can probably be based off of the existing process code.

src/microbe_stage/BioProcess.cs
CI09: I'm not from the first rows when it comes to making good algorithms, what you or someone else could propose?

dligr: Something like this:
````
if (compound.Key.Equals(input.Key))
     rate *= compound.Value.Ambient;
````

Comment (CI09) [reply src/microbe_stage/BioProcess.cs:78]: Oh right

File: src/microbe_stage/BioProcess.cs [hunk 4]
```diff
@@ -75,6 +75,23 @@ public void Check(string name)
         TranslationHelper.CopyTranslateTemplatesToTranslateSource(this);
     }
 
+    public float GetRateWithConditions(BiomeConditions conditions)
```
hhyyrylainen, line 78: Isn't this very similar to some of the methods in the ProcessSystem? For example `CalculateProcessMaximumSpeed` which has summary comment starting with: `Calculates the maximum speed a process can run at in a biome based on the environmental compounds.`

If that method is too much a simpler variant should be written in the process system class and notes added so that both variants are remembered to be updated at once if future changes is needed for this logic.

hhyyrylainen: Was this comment of mine addressed by some commit?

src/general/IWorldEffect.cs
```diff
@@ -34,71 +36,177 @@ public void OnRegisterToWorld()
...
-
-        var totalAmount = 0.0f;
-        var totalChunkAmount = 0.0f;
+        ApplyCompoundsAddition();
+        HandlePatchCompoundDiffusion();
+    }
 
-        var initialTotalDensity = 0.0f;
-        var finalTotalDensity = 0.0f;
+    private void ApplyCompoundsAddition()
```
hhyyrylainen, line 43: This effect no longer just reduces the glucose, so calling this `GlucoseReductionEffect` is completely incorrect in my mind.

Also this is so long that I think the new effect (called maybe something like `CompoundProductionEffect`) should be put in its separate file.

src/general/IWorldEffect.cs
```diff
@@ -34,71 +36,177 @@ public void OnRegisterToWorld()
...
+
+                targetWorld.Map.Patches[patchKeyValue.Key].Biome.ModifyLongTermCondition(compound.Key,
+                    tweakedBiomeConditions);
             }
-
-            finalTotalDensity += patch.Biome.ChangeableCompounds[Compound.Glucose].Density;
         }
+    }
+
+    private void HandlePatchCompoundDiffusion()
```
hhyyrylainen, line 184: Also wouldn't it be possible to split the compound diffusion effect to its own class? That way the separate "steps" of this change can be viewed in isolation and can be reasoned and read much more easily by programmers.

src/microbe_stage/BioProcess.cs
CI09: I forgot to delete this section, but rate uses the function you suggested now.

hhyyrylainen: Should be easy to then make a small commit to revert this part of the PR then, right?

hhyyrylainen:
If it is fine, I'd like to create a new branch from this and restore the old world effects and make these new ones activate only in experimental features mode. That way this code would get into master but wouldn't be enabled in general as I'm not yet comfortable with the way this modifies a lot of the compound amounts.

hhyyrylainen:
Merged as part of https://github.com/Revolutionary-Games/Thrive/pull/5556

PR #5440, opened 2024-09-02
Make waiting behavior handle more diverse metabolisms

### LINKED ISSUE #5423: AI likes to stay stationary (and try to engulf) too much now
Not sure what's caused this but now there are a bunch of cells that don't want to move, and just turn on engulf mode:

I noticed this when trying to find a new good seed for the microbe benchmark.
This has also been noticed by other people and talked about.

adQuid:

The oddly stationary microbes seem to be species with metabolisms that require multiple compounds, that are out of one but not the other. This logic now makes the AI only wait if ALL inputs for making ATP are in compound storage, under the assumption that ATP might never go back up on any one process alone.

**Related Issues**

Fixes #5423 (perhaps partially)

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -339,18 +339,24 @@ private void ChooseActions(in Entity entity, ref MicrobeAI ai, ref CompoundAbsor
             ai.ATPThreshold = 0.95f * speciesFocus / Constants.MAX_SPECIES_FOCUS;
         }
 
-        if (ai.ATPThreshold > 0.0f)
+        if (ai.ATPThreshold > Mathf.Epsilon)
```
hhyyrylainen, line 342: We have our own epsilon value in MathUtils that is used rather than the engine provided value.

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -339,18 +339,24 @@ private void ChooseActions(in Entity entity, ref MicrobeAI ai, ref CompoundAbsor
...
             {
+                bool outOfSomething = false;
                 foreach (var compound in compounds.Compounds)
                 {
-                    if (IsVitalCompound(compound.Key, compounds) && compound.Value > 0.0f)
+                    if (IsVitalCompound(compound.Key, compounds) && compound.Value <= Mathf.Epsilon)
                     {
-                        control.SetMoveSpeed(0.0f);
-                        return;
+                        outOfSomething = true;
```
hhyyrylainen, line 351: One idea I have is that you are missing a break here in the loop once the condition is satisfied, which causes a bunch of extra checking once the value is already true and no further iterations of the loop will change it at all.

src/microbe_stage/systems/MicrobeAISystem.cs
adQuid: right, forgot about that!

src/microbe_stage/systems/MicrobeAISystem.cs
adQuid: added

hhyyrylainen (approved):
AI cells now move more while taking ATP damage, which I guess is an improvement but I'm not totally sure.

PR #5445, opened 2024-09-02
Add "other" section in gas composition

CI09:

To fill gaps from future dynamic compounds

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (changes requested):
This is the wrong approach in my opinion. There should not be an other compound type. Instead the display GUI should tally up the other stuff and if it doesn't add up to 100% then it should dynamically generate a compound listing for "other" that is the remaining percentage so that all of the stuff together adds up to 100%.

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -383,6 +388,10 @@ private void UpdatePatchDetails()
             percentageFormat.FormatSafe(Math.Round(GetCompoundAmount(SelectedPatch, carbondioxideCompound),
                 Constants.ATMOSPHERIC_COMPOUND_DISPLAY_DECIMALS));
 
+        otherCompoundLabel.Text = percentageFormat.FormatSafe(100.0f - Math.Round(GetCompoundAmount(SelectedPatch, carbondioxideCompound)
+            + GetCompoundAmount(SelectedPatch, nitrogenCompound) + GetCompoundAmount(SelectedPatch, oxygenCompound)),
+            Constants.ATMOSPHERIC_COMPOUND_DISPLAY_DECIMALS);
```
hhyyrylainen, line 393: I'm not fully sure myself, but can the rounding mode end up in a situation where the rounded numbers add up to slightly more than 100%? Like 100.1%. I wonder if there is a way to round down this to the display precision to ensure this never overshoots.

src/microbe_stage/gui/PatchDetailsPanel.tscn
```diff
@@ -302,6 +303,25 @@ layout_mode = 2
...
+layout_mode = 2
+size_flags_vertical = 4
+tooltip_text = "CARBON_DIOXIDE"
+texture = ExtResource("16_baud4")
+expand_mode = 1
+
+[node name="Label" type="Label" parent="NormalContent/MarginContainer/ScrollContainer/VBoxContainer/PatchDetailsContainer/AtmosphericGasses/Other"]
+editor_description = "PLACEHOLDER"
+layout_mode = 2
+text = "Other"
```
hhyyrylainen, line 322: Shouldn't the text be translatable as this will be displayed like it is.?

src/microbe_stage/gui/PatchDetailsPanel.tscn
```diff
@@ -302,6 +303,25 @@ layout_mode = 2
...
+[node name="Other" type="HBoxContainer" parent="NormalContent/MarginContainer/ScrollContainer/VBoxContainer/PatchDetailsContainer/AtmosphericGasses"]
+layout_mode = 2
+size_flags_horizontal = 3
+mouse_filter = 2
+
+[node name="TextureRect" type="TextureRect" parent="NormalContent/MarginContainer/ScrollContainer/VBoxContainer/PatchDetailsContainer/AtmosphericGasses/Other"]
+custom_minimum_size = Vector2(30, 25)
+layout_mode = 2
+size_flags_vertical = 4
+tooltip_text = "CARBON_DIOXIDE"
```
hhyyrylainen, line 315: Outdated tooltip data?

hhyyrylainen (changes requested):

hhyyrylainen:
The approach is now good but I had some comments about technical implementation details.

src/microbe_stage/gui/PatchDetailsPanel.tscn
CI09: I was wondering too, but other compounds are written lowercase too...

CI09: Probably doesn't matter as it gets replaced with numbers

src/microbe_stage/gui/PatchDetailsPanel.cs
hhyyrylainen: No, like a floor operation but just at the display decimal count - 1 and not towards the lower integer like `Floor` by default does. Here's a discussion related to this, but it is for python: https://stackoverflow.com/questions/58065055/floor-and-ceil-with-number-of-decimals though porting the code over to C# doesn't seem impossible.

src/microbe_stage/gui/PatchDetailsPanel.tscn
hhyyrylainen: Well that's maybe a slight problem because if these don't have any text, then the player is going to wonder what the mystery compound is that takes up a ton of space.

src/microbe_stage/gui/PatchDetailsPanel.cs
hhyyrylainen: Also the other label should not show up at all if there's nothing in there. It makes no sense to say "btw these above numbers add up to 100% so there's no remainder left".

locale/en.po
```diff
@@ -3899,6 +3899,9 @@ msgstr "Osmoregulation cost multiplier"
 msgid "OSMOREGULATION_COST_MULTIPLIER_EXPLANATION"
 msgstr "(cost of player species osmoregulation)"
 
+msgid "OTHER"
```
hhyyrylainen, line 3902: The translation key should be more clear, like "OTHER_ITEMS_CATEGORY" or something equally long and clear as I bet many languages will need the additional context for translation".

simulation_parameters/microbe_stage/biomes.json
```diff
@@ -164,7 +164,7 @@
         "nitrogen": {
           "Amount": 0,
           "Density": 0,
-          "Ambient": 0.7
+          "Ambient": 0.6
```
hhyyrylainen, line 167: Is this an intentional change, or just for debugging this? If for just debugging I'd prefer this to be removed when this feature is done.

src/microbe_stage/BiomeConditions.cs
```diff
@@ -261,8 +261,6 @@ public void Check(string name)
                 "Chunks missing");
         }
 
-        float sumOfGasses = 0;
```
hhyyrylainen, line 264: I think this should be kept, because gasses may not add up to more than 100%.

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -383,31 +390,47 @@ private void UpdatePatchDetails()
             percentageFormat.FormatSafe(Math.Round(GetCompoundAmount(SelectedPatch, carbondioxideCompound),
                 Constants.ATMOSPHERIC_COMPOUND_DISPLAY_DECIMALS));
 
+        var otherAmount = CalculateOtherCompoundAmount();
+
+        otherCompoundLabel.Text = percentageFormat.FormatSafe(otherAmount,
+            Constants.ATMOSPHERIC_COMPOUND_DISPLAY_DECIMALS);
+
+        if (SelectedPatch != null)
```
hhyyrylainen, line 398: This null check is unnecessary as the calculate other compound method returns 0 if the patch is null.

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -383,31 +390,47 @@ private void UpdatePatchDetails()
...
+        {
+            otherCompoundBase.Visible = false;
+        }
+
         // Compounds
         hydrogenSulfideLabel.Text =
-            Math.Round(GetCompoundAmount(SelectedPatch, hydrogensulfideCompound),
-                    Constants.PATCH_CONDITIONS_COMPOUND_DISPLAY_DECIMALS)
-                .ToString(CultureInfo.CurrentCulture);
+            Math.Round(GetCompoundAmount(SelectedPatch!, hydrogensulfideCompound),
```
hhyyrylainen, line 416: These seem like unnecessary changes that you should revert. If you remove the above null check the automatic checks should stop complaining about null problems here.

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -820,6 +843,16 @@ private void MigrationAmountSliderChanged(double value)
         UpdateMigrationStatusText();
     }
 
+    private float CalculateOtherCompoundAmount()
```
hhyyrylainen, line 846:
```suggestion
    private float CalculateUnaccountedForCompoundPercentage()
```

I think this kind of name would be more clear.

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -820,6 +843,16 @@ private void MigrationAmountSliderChanged(double value)
...
+    {
+        if (SelectedPatch == null)
+            return 0;
+
+        return (float)Math.Floor(10 * (100.0f - (GetCompoundAmount(SelectedPatch, carbondioxideCompound)
```
hhyyrylainen, line 851: Again, here I'd really like you to write out this in like 10 lines so that it is very easy to add new compounds to be taken into account here.

hhyyrylainen (changes requested):

hhyyrylainen (approved):
I fixed the weird sizing of the other icon, and fixed some of the remaining technical problems I saw.

PR #5459, opened 2024-09-04
Fixed engulfed storage size flickering when ejecting too big objects

hhyyrylainen:
after cell loses size, also made the other eject conditions only print their relevant message once when the ejection starts

**Related Issues**

https://discord.com/channels/228300288023461893/958598553389903913/1280833370682560533

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/components/Engulfer.cs
+++ b/src/microbe_stage/components/Engulfer.cs
@@ -33,7 +33,7 @@ public struct Engulfer
     public float EngulfingSize;
 
     /// <summary>
-    ///   The amount of space all of the currently engulfed objects occupy in the cytoplasm. This is used to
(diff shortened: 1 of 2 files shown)


CI09:
I still get this bug.

hhyyrylainen:
That's an unrelated issue. And I think it is caused as the used storage size is temporarily zero on the frame when ejecting indigestible is started. I made one more commit to this branch that also tries to solve that case as well by counting the size of the indigestible chunk for the frame when it is decided to be ejected. As it seemed pretty simple to fix, though I don't have any save to easily check that.

The actual bug I was trying to solve with this was: https://discord.com/channels/228300288023461893/958598553389903913/1280827908507697183

Both of these problems should now be fixed by this PR.

hhyyrylainen:
I assume there were no further problems and this can be merged.

PR #5461, opened 2024-09-04
Add 3D membranes

### LINKED ISSUE #1748: Make membrane meshes properly 3D
and not just flat disks.

Also needs some edge shader to make them look good from the side.

Will fix the problem of organelles sticking outside the membrane. Could even remove the tint from organelles and have the membrane cause the colour change instead.

dligr:

Not complete yet, but probably ready for an initial code review.

To do:
- [x] Make a better membrane normal map (was actually a problem with UV mapping)
- [x] Make bigger microbes' membrane higher
- [x] Remake membrane textures for 3D?

Also, I encourage everyone to give feedback on the new membrane visuals. They are a quite significant change, and may require fine-tuning before being added into the game.

New_membranes

**Related Issues**

Closes #1748

**Progress Checklist**

      break existing features:
      members reviewing this PR)


AverageNerdz:
The feature works pretty well, the entire texturing aspect has to be reworked but i think it should be saved for later. I would love to be able to make custom shapes with it too but so far it's coming along pretty great

src/microbe_stage/systems/MicrobeVisualsSystem.cs
```diff
@@ -69,7 +67,18 @@ public bool HasPendingOperations()
 
     public override void Dispose()
     {
-        Dispose(true);
```
hhyyrylainen, line 72: Why was the proper dispose pattern removed from this class?

src/microbe_stage/systems/MicrobeRenderPrioritySystem.cs
```diff
@@ -59,12 +59,13 @@ protected override void Update(float state, in Entity entity)
...
         {
-            // First material is always the membrane, which gets the base value. Organelles are rendered above it.
+            // First material is always the membrane, which gets the base value with an offset,
+            // so that organelles are rendered below it.
             var materialList = materials.Materials;
 
             if (materialList.Length > 0)
             {
-                materialList[0].RenderPriority = renderOrder.RenderPriority;
+                materialList[0].RenderPriority = renderOrder.RenderPriority + 20;
```
hhyyrylainen, line 68: I think this 20 should be in constants so that it is clear there's this kind of hardcoded value exists and might conflict with something if there are future code changes.

shaders/Membrane.gdshader
```diff
@@ -29,28 +27,41 @@ void vertex(){
         * wigglyNess * size;
 }
 
+float fresnel(float power, vec3 normal, vec3 view)
+{
+	return clamp(pow(((1.0 - clamp(dot(normal, view), 0.0, 1.0)) * 0.875 + 0.125) * 3.0, power), 0.0, 1.0);
```
hhyyrylainen, line 32: There's a tab character instead of spaces here.

shaders/Membrane.gdshader
```diff
@@ -29,28 +27,41 @@ void vertex(){
...
-    
-    vec4 normalmap = texture(normalTexture, UV);
-    vec3 normals = vec3(normalmap.r * 2.0 - 1.0, normalmap.g * 2.0 - 1.0, 
-        normalmap.b * 2.0 - 1.0);
 
-    float y = normals.y;
-    normals.y = -normals.z;
-    normals.z = y;
+    vec4 normalmap = texture(normalTexture, UV);
+	NORMAL_MAP = normalmap.xyz;
```
hhyyrylainen, line 39: Also seems like there's more tabs in this file, but I'll just comment this second time.

src/microbe_stage/systems/MicrobeVisualsSystem.cs
```diff
@@ -347,29 +354,6 @@ private void CreateOrganelleVisuals(Node3D parentNode, ref OrganelleContainer or
 
                 organelleContainer.CreatedOrganelleVisuals.Add(placedOrganelle, visualsInstance);
             }
-
-            // Visuals already exist
-            var graphics = placedOrganelle.OrganelleGraphics;
```
hhyyrylainen, line 352: Is there a reason why you removed this code? If I'm seeing things correctly then you haven't replaced this with anything meaning the organelle visuals are now gone.

dligr: The `graphics` variable was only needed to assign shader variables to it; the graphics themselves are created before, so this change doesn't really remove the visuals.

Anyway, after some thought, I decided to revert removing this code, so that tint might still be added to organelles if needed.

src/microbe_stage/systems/MicrobeRenderPrioritySystem.cs
dligr: Done

src/microbe_stage/systems/MicrobeVisualsSystem.cs
hhyyrylainen: But if tint was removed, doesn't that make all species look very similar? As the colour was kind of the primary way (well besides size and some really distinct organelles) to quickly identify species.

dligr: The tint in this function is only applied to organelles (after all, it's called `CreateOrganelleVisuals`), while mebranes' tint is left as is

shaders/Membrane.gdshader
```diff
@@ -29,28 +29,42 @@ void vertex(){
...
+    vec3 adjusted_view = VIEW;
+    // Move fresnel's center closer to membrane's top
+    // Not an absolutely necessary step, but it makes fresnel look better
+    adjusted_view = normalize(adjusted_view + vec3(0.0, 0.0, 2.0));
+
+    normalmap = normalmap * 2.0 - 1.0;
+
+    vec3 adjusted_normals = normalmap.z * NORMAL
+    + normalmap.x * TANGENT
+    + normalmap.y * BINORMAL;
```
hhyyrylainen, line 62: These lines should have extra indentation as these are a continued statement.

src/microbe_stage/MembranePointData.cs
```diff
@@ -105,6 +108,20 @@ public float Radius
         }
     }
 
+    /// <summary>
+    ///   Membrane's height. Should be used as a multiplier for a constant value
```
hhyyrylainen, line 112: Could you expand the comment a bit here, I'm not fully understanding what the constant value here means?

src/microbe_stage/systems/MicrobeVisualsSystem.cs
```diff
@@ -300,7 +300,9 @@ private void CreateOrganelleVisuals(Node3D parentNode, ref OrganelleContainer or
     {
         organelleContainer.CreatedOrganelleVisuals ??= new Dictionary<PlacedOrganelle, Node3D>();
 
-        var organelleColour = PlacedOrganelle.CalculateHSVForOrganelle(cellProperties.Colour);
+        // To make organelles' color dependent on membrane's color:
```
hhyyrylainen, line 303:
```suggestion
        // To make organelles' colour dependent on membrane's colour:
```

src/microbe_stage/systems/MicrobeVisualsSystem.cs
```diff
@@ -300,7 +300,9 @@ private void CreateOrganelleVisuals(Node3D parentNode, ref OrganelleContainer or
...
+        // var organelleColour = PlacedOrganelle.CalculateHSVForOrganelle(cellProperties.Colour);
+        var organelleColour = Colors.White;
```
hhyyrylainen, line 305: I guess we can see if people prefer organelles to be white tinted always now...

Though I have one question: would something like using like 0.1 multiplier on the HSV colour to blend from white look good? That way we'd still have a bit of species customization on the organelles. I think that would be a less drastic change and maybe would make both groups happy?

src/microbe_stage/MembraneShapeGenerator.cs
```diff
@@ -144,42 +149,42 @@ public MembranePointData GenerateShape(IMembraneDataSource parameters)
...
-        // about chitin later)
-        if (membraneType.CellWall)
+        for (int layer = 0; layer < layerCount; ++layer)
         {
-            height = 0.05f;
-        }
+            for (int i = 0; i < vertexCount; ++i)
+            {
+                int id = i + layer * vertexCount;
+                float turnFraction = 2.0f * MathF.PI * i / vertexCount;
```
hhyyrylainen, line 163: Would pre-calculating `2 * PI / vertexCount` be faster? (as that would avoid quite many divisions and multiplies per loop iteration)

src/microbe_stage/MembraneShapeGenerator.cs
```diff
@@ -205,44 +210,162 @@ private static Vector2 FindClosestOrganelleHex(Vector2[] hexPositions, int hexCo
...
+
+                vertices[i + layer * vertexCount] = point;
+            }
         }
 
-        // Write mesh data //
-        var vertices = new Vector3[bufferSize];
-        var uvs = new Vector2[bufferSize];
+        vertices[vertices.Length - 1] = new Vector3(center.X,
+            vertices[(layerCount - 1) * vertexCount].Y, center.Z); // Top vertex
```
hhyyrylainen, line 360: The usual style in thrive is to not use trailing comments on lines, so these comments could be moved to be before the lines they comment on for style uniformity.

assets/textures/MembraneNormals.png
hhyyrylainen, line 1: This is the same resolution as the membrane textures, right? Just to confirm (I haven't opened this PR locally yet).

hhyyrylainen (changes requested):
I don't have much to say on the core algorithm of the mesh generation but I have a few small technical nitpicks before moving to final testing for this.

assets/textures/MembraneNormals.png
dligr: Yes, it's 1024x1024

src/microbe_stage/MembranePointData.cs
dligr: I decided to remove the part about this being used as a multiplier as it makes more sense to immediately calculate true membrane height (and matched it in the code, of course)

hhyyrylainen (approved):
The code is good now and I did some checking in game and everything seemed good (except maybe the chitin texture is way too similar to normal membrane now).

PR #5470, opened 2024-09-08
Reduce delay on patch selection in editor

dligr:

Reduces lag when selecting a patch in editor's patch map. Instead of updating auto-evo report when clicking on a patch, it now does it when opening the report tab.

Additionally, does a slight refactor of patch-related editor code. `SelectedPatch` is now the patch that the player selected, while `TargetPatch` (formerly `SelectedPatch`) is the patch that the player wants to move to

Also, fixes an issue when auto-evo report would switch to the player's current patch (instead of selected patch) on language change.

**Related Issues**

No particular issue, but I think that this change leads to better UX. Clicking around the patch map will now be much more smooth.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/IEditorReportData.cs
+++ b/src/microbe_stage/editor/IEditorReportData.cs
@@ -3,10 +3,20 @@
 /// </summary>
 public interface IEditorReportData : IEditor
 {
+    /// <summary>
+    ///   Returns the patch the player is currently in
+    /// </summary>
     public Patch CurrentPatch { get; }
 
+    /// <summary>
+    ///   Returns the patch the player wants to move after editing. If the player doesn't want to move, returns
+    ///   the same as <see cref="CurrentPatch"/>
+    /// </summary>
     public Patch? TargetPatch { get; }
 
+    /// <summary>
+    ///   Returns the patch the player selected in the patch map
+    /// </summary>
     public Patch? SelectedPatch { get; }
 
     public void SendAutoEvoResultsToReportComponent();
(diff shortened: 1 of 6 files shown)


CI09:
Code side looks good to my eye.

HexapodPhilosopher:
I still get an empty tab in the patch map, until clicking on a patch.

dligr:
> I still get an empty tab in the patch map, until clicking on a patch.

This PR's purpose is to reduce the lag that happens when clicking on a patch, rather than to automatically select a patch when entering the editor.

hhyyrylainen:
I was about to say that wasn't that the existing behaviour that no info on a patch is shown before you select one...
Though isn't there supposed to be a text saying to select a patch in that right panel? Has there been a regression or is that text still there?

src/microbe_stage/editor/IEditorReportData.cs
```diff
@@ -3,8 +3,20 @@
...
+    /// </summary>
     public Patch CurrentPatch { get; }
 
+    /// <summary>
+    ///   Returns the patch the player wants to move after editing
+    /// </summary>
+    public Patch? TargetPatch { get; }
+
+    /// <summary>
+    ///   Returns the patch the player selected in the patch map
```
hhyyrylainen, line 18:
```suggestion
    ///   Returns the patch the player selected in the patch map (for displaying in the GUI)
```
I think this clarification would be good.

hhyyrylainen (approved):
I suppose this is fine, even though I hadn't really noticed the lag. I was trying to think if there was a reason to make this another way but I think this approach pretty much covers everything that editors might do (I was trying to think if there was a case where the selected patch change would need to immediately apply but I couldn't think of one).

PR #5491, opened 2024-09-12
Flagellum scaling

### LINKED ISSUE #4107: Add flagellum upgrade to increase thrust force (speed) along with ATP usage
can now be done with the general upgrades system, though it is missing one part: allowing specifying different organelle graphics scene when an upgrade is in effect: https://github.com/Revolutionary-Games/Thrive/issues/4106

CI09:

Lets you scale flagellum (shorter/ less ATP usage, less force, longer - vice versa)

**Related Issues**

closes #4107

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
This needs text for this feature (removing the placeholders so that the translation system will see the new strings).

CI09:
There is still a serious ATP balance display bug / balance which I'm not sure right now how to fix

hhyyrylainen:
ATP balance calculation code where flagellum movement cost is taken into account needs to be updated, if you haven't done that already. The code is in the ProcessSystem file probably.

src/microbe_stage/editor/upgrades/FlagellumUpgradeGUI.tscn
```diff
@@ -0,0 +1,31 @@
...
+anchor_right = 1.0
+anchor_bottom = 1.0
+offset_right = -880.0
+offset_bottom = -650.0
+theme = ExtResource("1_5l116")
+script = ExtResource("2_f2xv7")
+lengthSlider = NodePath("HSlider")
+
+[node name="Description" type="Label" parent="."]
+editor_description = "PLACEHOLDER"
```
hhyyrylainen, line 19: This needs to be removed to make this translatable.

src/microbe_stage/editor/upgrades/FlagellumUpgradeGUI.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Godot;
+
+/// <summary>
+///   Upgrade GUI for the vacuole that allows specializing it
```
hhyyrylainen, line 4: This comment is incorrect.

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -30,6 +32,17 @@ public MovementComponent(float momentum)
 
     public void OnAttachToCell(PlacedOrganelle organelle)
     {
+        var configuration = organelle.Upgrades?.CustomUpgradeData;
+
+        if (configuration == null)
+        {
+            flagellumLength = 0;
```
hhyyrylainen, line 39: Does this result in flagellum being the length it is currently? Because for all upgrades when the data is missing I've made sure it results in the behaviour that happens if the upgrade didn't exist at all.

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -126,7 +139,9 @@ private float CalculateMovementForce(CompoundBag compounds, Vector3 wantedMoveme
         var newAnimationSpeed = 2.3f;
         lastUsed = true;
 
-        var requiredEnergy = Constants.FLAGELLA_ENERGY_COST * elapsed;
+        var requiredEnergy = (Constants.FLAGELLA_ENERGY_COST + Constants.FLAGELLA_MAX_UPGRADE_ATP_USAGE
+            * flagellumLength)
+            * elapsed;
```
hhyyrylainen, line 144: This being on its own line is bad formatting in my eyes.

src/microbe_stage/systems/MicrobeVisualsSystem.cs
```diff
@@ -372,6 +372,15 @@ private void CreateOrganelleVisuals(Node3D parentNode, ref OrganelleContainer or
...
             }
+
+            if (placedOrganelle.Upgrades?.CustomUpgradeData is FlagellumUpgrades flagellumUpgrades)
+            {
+                var flagellumLength = flagellumUpgrades.LengthFraction;
+
+                graphics.Position = new Vector3(0, 0, Constants.FLAGELLA_UPGRADE_VISUAL_OFFSET * flagellumLength);
+                graphics.Scale = new Vector3(1, 1, Constants.FLAGELLA_MAX_UPGRADE_VISUAL_LENGTH * flagellumLength
+                    + Constants.FLAGELLA_MIN_VISUAL_LENGTH);
+            }
```
hhyyrylainen, line 383: This is the wrong place. The `transform` is already calculated above in this method. You should make it so that the flagellum adjustment is applied there. (if there's a bug where the transform doesn't fully apply when graphics already exist, then that bug should be fixed separately).

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -192,7 +192,17 @@ public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemp
             // Take special cell components that take energy into account
             if (includeMovementCost && organelle.Definition.HasMovementComponent)
             {
-                var amount = Constants.FLAGELLA_ENERGY_COST;
+                var amount = 0.0f;
```
hhyyrylainen, line 195:
```suggestion
                float amount;
```

src/microbe_stage/organelle_components/MovementComponent.cs
CI09: Length for all related stuff is BASE + flagellumLength * modifier so if it's 0 stuff should be their base form

hhyyrylainen: That sounds good.

CI09:
Seems for some reason when very short flagellum doesn't have enough mass to push, editor shows ATP consumption as negative and I don't know why.

hhyyrylainen:
Could it be that the math actually works out for it to be negative force / energy cost at certain length?

CI09:
Ah I used wrong constant, my bad

src/microbe_stage/editor/upgrades/FlagellumUpgradeGUI.tscn
```diff
@@ -0,0 +1,30 @@
...
+lengthSlider = NodePath("HSlider")
+
+[node name="Description" type="Label" parent="."]
+layout_mode = 2
+text = "FLAGELLUM_LENGTH_DESCRIPTION"
+label_settings = ExtResource("3_pr3nu")
+
+[node name="HSlider" type="HSlider" parent="."]
+layout_mode = 2
+tooltip_text = "TOXIN_TOXICITY_CUSTOMIZATION_TOOLTIP"
```
hhyyrylainen, line 25: This tooltip is wrong.

src/microbe_stage/editor/upgrades/FlagellumUpgradeGUI.cs
```diff
@@ -0,0 +1,36 @@
...
+    private Slider lengthSlider = null!;
+#pragma warning restore CA2213
+
+    public override void _Ready()
+    {
+    }
+
+    public void OnStartFor(OrganelleTemplate organelle, GameProperties currentGame, float costMultiplier)
+    {
+        if (organelle.Upgrades?.CustomUpgradeData is FlagellumUpgrades flagellumUpgrades)
```
hhyyrylainen, line 19: Upgrade GUIs aren't reused (right now) but you should still put in the logic that the slider is reset to the default value if there are no flagellum upgrades specified.

src/microbe_stage/PlacedOrganelle.cs
```diff
@@ -328,14 +328,30 @@ public Transform3D CalculateVisualsTransform()
...
+
         var scale = CalculateTransformScale();
 
         cachedExternalOrientation = orientation;
         cachedExternalPosition = externalPosition;
 
         // TODO: check that the rotation of ModelOffset works correctly here (also in
         // CalculateVisualsTransformExternalCached)
-        return new Transform3D(new Basis(orientation).Scaled(new Vector3(scale, scale, scale)),
+        if (scaleZ == 0.0f)
```
hhyyrylainen, line 348: Couldn't scaleZ start off by being equal to scale? That way this extra if and duplicated complicated math line is avoided.

src/microbe_stage/editor/upgrades/FlagellumUpgradeGUI.tscn
```diff
@@ -0,0 +1,30 @@
+[gd_scene load_steps=4 format=3 uid="uid://begbjuynar75j"]
+
+[ext_resource type="Theme" uid="uid://b4cx0o110g4b6" path="res://src/gui_common/thrive_theme.tres" id="1_5l116"]
+[ext_resource type="Script" path="res://src/microbe_stage/editor/upgrades/FlagellumUpgradeGUI.cs" id="2_f2xv7"]
+[ext_resource type="LabelSettings" uid="uid://dvqx73nhtr0y2" path="res://src/gui_common/fonts/Body-Regular-Small.tres" id="3_pr3nu"]
+
+[node name="VacuoleUpgradeGUI" type="VBoxContainer" node_paths=PackedStringArray("lengthSlider")]
```
hhyyrylainen, line 7: This node has a name that doesn't match the filename.

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -28,6 +30,17 @@ public MovementComponent(float momentum)
...
     {
+        var configuration = organelle.Upgrades?.CustomUpgradeData;
+
+        if (configuration == null)
+        {
+            flagellumLength = 0;
+        }
+        else
+        {
+            flagellumLength = ((FlagellumUpgrades)configuration).LengthFraction;
```
hhyyrylainen, line 41: For safety the if here should be rewritten as `if (organelle.Upgrades?.CustomUpgradeData is FlagellumUpgrades flagellumUpgrade)`

src/microbe_stage/organelle_components/MovementComponent.cs
```diff
@@ -141,9 +155,13 @@ private float CalculateMovementForce(CompoundBag compounds, Vector3 wantedMoveme
         SetSpeedFactor(newAnimationSpeed);
 
         if (isBacteria)
-            return Constants.FLAGELLA_BASE_FORCE * forceMagnitude;
+        {
+            return (Constants.FLAGELLA_BASE_FORCE + Constants.FLAGELLA_MAX_UPGRADE_FORCE * flagellumLength)
```
hhyyrylainen, line 159: I think it would be better to refactor out a new variable called `var baseForce = Constants.FLAGELLA_BASE_FORCE + Constants.FLAGELLA_MAX_UPGRADE_FORCE * flagellumLength` than have a really long calculation on line 163 and 164.

hhyyrylainen (changes requested):
I noticed a few technical issues. Besides those I guess this doesn't have any major roadblocks if no one wants to comment about the balance of this feature.

src/microbe_stage/organelle_components/MovementComponent.cs
CI09: Also I noticed a flaw code for non-bacterias, good you mentioned it here.

CI09:
There is weird bug where when upgrade popup shows up stuff in it don't fit and you need to scroll but it's generally in usable state

hhyyrylainen:
Is the minimum size set big enough for it? Also if the explanation text doesn't have x-axis minimum width (and word wrap enabled), that would be good to do. And might help with godot messing up the size calculation and showing the scroll bar.

CI09:
I tried setting minimum size but it kinda looked like didn't do anything

hhyyrylainen:
The minimum size should be bigger than what the controls naturally want, then it starts affecting things and might fix the scrollbar issue if the natural size has a slight bug where it sometimes calculates the space as slightly too small and adds a scrollbar.

src/microbe_stage/editor/upgrades/FlagellumUpgradeGUI.tscn
hhyyrylainen: This is still here.

src/microbe_stage/editor/upgrades/FlagellumUpgradeGUI.cs
hhyyrylainen: And this still looks like it is missing the `else` branch.

hhyyrylainen (changes requested):
Was the new bug that was discovered fixed? I just reviewed the source code again and only 2 of the files I had previous unsolved comments on seem ones with any problems left.

CI09:
I must worry you that I don't really know how to fix that problem. I tried changing custom minimum size, enable expand mode but neither worked.

CI09:
I can make it at least fit within for now (scroll remains still though)
image

hhyyrylainen:
Okay, so you need help with making the GUI work correctly?

CI09:
Actually it looks acceptable now. Not perfect (slider doesn't fill out entire horizontal space) but good.

hhyyrylainen (changes requested):
I didn't see the problem with the scrollbar or wrong horizontal spacing.

What I did immediately notice that *default flagellum length* is not the same as before:




I need to ask that the numbers are rebalanced so that the default flagellum length is the same as before this PR.

hhyyrylainen:
Was the default length changed? This was force pushed again so I can't easily check the changes, and GitHub only shows that en.po file was changed since I last marked the file as viewed.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2301,6 +2301,14 @@ private void UpdateAlreadyPlacedVisuals()
                 organelleModel.Visible = !MicrobePreviewMode;
 
                 UpdateOrganellePlaceHolderScene(organelleModel, modelInfo, Hex.GetRenderPriority(organelle.Position));
+
+                if (organelle.Upgrades?.CustomUpgradeData is FlagellumUpgrades flagellumUpgrades)
```
hhyyrylainen, line 2305: I just noticed this is in the wrong place, I'll fix this.

hhyyrylainen (approved):
I reduced the extremes of the scaling as I though that that really made it obvious that it was just a model being stretched on a single axis for the graphics. I also fixed that one wrong place where the scale calculation was for the editor. So now I think this is good to go. I opened a follow up issue, though: https://github.com/Revolutionary-Games/Thrive/issues/5536

PR #5497, opened 2024-09-13
Implement Catmull-Clark subdivision

dligr:

Implements [Catmull-Clark subdivision](https://en.wikipedia.org/wiki/Catmull%E2%80%93Clark_subdivision_surface) for late multicellular creatures. Simply put, this algorithm subdivides a low-poly mesh into a smooth higher-poly one.

**Related Issues**

3D creatures were quite low-poly and this subdivision algorithm counteracts this with almost no performace loss (unlike, for example, increasing Dual Contouring resolution).

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09 (approved):
Smooth (playtest). I'm too dumb to understand math behind this but I guess it's fine

hhyyrylainen:
I also can't really comment on the math and algorithm, but to me it looks like this requires quite a bit of temporary memory (which is basically my only observation). That's fine as this is a prototype, but if there is some easy way to reuse a few lists or use pooled arrays that could decrease the garbage collector pressure that would make this code easier to polish up in the future once this is no longer a prototype part of the game.

dligr:
I have made a number of memory optimizations. Namely, `triIndices` are now reused, and `newNormals` are created as an array right away.

I don't think it's possible to remove the `edgeTriangles` Dictionary, so the only other memory optimization that can be made is to remove arrays from the aforementioned Dictionary's values.
Would it be right to replace those arrays with structs?

hhyyrylainen:
If they are always the same size, it is probably a lot clearer to make them structs anyway (in addition to the memory allocation and cache locality benefits).

hhyyrylainen (approved):
Code is fine. Seems like the performance hit isn't much from this. I can't say much about the visual change but I'll aprove this.

PR #5499, opened 2024-09-14
Balancing tweaks to engulfing

CI09:

Increases time to digest something and yield of it

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -844,7 +844,7 @@ public static class Constants
 
     public const float OPTIMAL_THERMOPLAST_TEMPERATURE = 100.0f;
 
-    public const float ADDITIONAL_DIGESTIBLE_GLUCOSE_AMOUNT_MULTIPLIER = 2f;
+    public const float ADDITIONAL_DIGESTIBLE_GLUCOSE_AMOUNT_MULTIPLIER = 1.25f;
 
     public const string LYSOSOME_DEFAULT_ENZYME_NAME = "lipase";
 
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -793,12 +793,12 @@ public static class Constants
     /// <summary>
     ///   The speed of which a cell can absorb compounds from digestible engulfed objects.
     /// </summary>
-    public const float ENGULF_COMPOUND_ABSORBING_PER_SECOND = 0.5f;
+    public const float ENGULF_COMPOUND_ABSORBING_PER_SECOND = 0.3f;
 
     /// <summary>
     ///   How much compounds a cell can absorb per second from digestible engulfed objects.
     /// </summary>
-    public const float ENGULF_BASE_COMPOUND_ABSORPTION_YIELD = 0.3f;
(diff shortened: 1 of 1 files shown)


dligr (approved):
The code changes are fine.
I like the gameplay change too. Playing as a heterotroph now feels more satisfying and more viable. I think that this would require a game design review, however.

simulation_parameters/Constants.cs
```diff
@@ -793,12 +793,12 @@ public static class Constants
     /// <summary>
     ///   The speed of which a cell can absorb compounds from digestible engulfed objects.
     /// </summary>
-    public const float ENGULF_COMPOUND_ABSORBING_PER_SECOND = 0.5f;
+    public const float ENGULF_COMPOUND_ABSORBING_PER_SECOND = 0.3f;
 
     /// <summary>
     ///   How much compounds a cell can absorb per second from digestible engulfed objects.
```
hhyyrylainen, line 799: Looks like this comment is copy pasted from above. Maybe you could fix this comment in this PR?

CI09: It's not really copy-pasted in my eyes

hhyyrylainen: Okay, I misworded my comment. I should have said that the comment here didn't seem to make sense in relation to what the constant is. So the comment talks about "compound a cell can absorb per second" but the constant name contains "COMPOUND_ABSORPTION_YIELD". Yield is not the same as speed. So either the comment or the variable name has to change because they don't match.

simulation_parameters/Constants.cs
```diff
@@ -793,12 +793,12 @@ public static class Constants
...
     /// </summary>
-    public const float ENGULF_COMPOUND_ABSORBING_PER_SECOND = 0.5f;
+    public const float ENGULF_COMPOUND_ABSORBING_PER_SECOND = 0.3f;
 
     /// <summary>
-    ///   How much compounds a cell can absorb per second from digestible engulfed objects.
+    ///   How much compounds in relation to real compound amount can be absorbed from digestible engulfed objects.
     /// </summary>
-    public const float ENGULF_BASE_COMPOUND_ABSORPTION_YIELD = 0.3f;
+    public const float ENGULF_BASE_COMPOUND_ABSORPTION_YIELD = 1.8f;
```
hhyyrylainen, line 801: Would the balance work if this was at 1? Lysosome is meant to further increase the absorption efficiency. I ask because we try to pretend things are physically realistic in Thrive, but here if we multiply the compound amount by 1.8 we are basically creating new matter out of nothing.

CI09: True, but while testing prokaryotic, I got like up to 1 glucose from engulfing cell half my size which isn't much to suffice. One option is to give the stuff cell is built of more glucose, that would solve that problem.

hhyyrylainen (approved):
From code standpoint this looks fine to me. I can't say anything related to the balancing so I guess I'll have to accept this within a few days if no one has anything to say.

hhyyrylainen:
Looks like no one wants to comment on the balance so I'll merge this now.

PR #5503, opened 2024-09-15
Update setup_instructions.md with more linux info

HexapodPhilosopher:
Added extra info for Linux setup

This PR does some stuff...

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/doc/setup_instructions.md
+++ b/doc/setup_instructions.md
@@ -50,11 +50,19 @@ LFS. Note that the GitHub option to download as .zip will not work
 
 You need at least Git LFS version 2.8.0, old versions do not work.
 
-On Linux use your package manager to install Git. On Mac install the
-package manager [homebrew](https://brew.sh/) if you don't already have
-it, and use it to install Git. On Mac and Linux Git LFS is likely available
-as a package named `git-lfs`. If it is not install it manually. After
-installing remember to run `git lfs install` in terminal.
+On Linux use your package manager to install Git. This can be done in the
+terminal - for instance, on Ubuntu, type: `sudo apt-get install git` . On
+Mac, install the package manager [homebrew](https://brew.sh/) if you don't
+already have it, and use it to install Git. On Mac and Linux Git LFS is
+likely available as a package named `git-lfs`. If it is not, install it
+manually. You can download Git LFS from https://git-lfs.com. After
+downloading the file, extract it wherever you want. In the terminal, change
+directory to the newly extracted folder (e.g. " cd ~/Games/git-lfs-3.4.1 ").
+You can then install it via the terminal - for instance on Ubuntu, type:
+`sudo bash ./install.sh`.
+OR, install Git LFS entirely by typing in the terminal - for instance on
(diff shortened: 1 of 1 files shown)


doc/setup_instructions.md
```diff
@@ -551,6 +560,15 @@ is available through Homebrew.
 `msgmerge` is part of the gettext tools. So any errors with that
 missing are about these gettext tools.
 
+## Other tools
```
hhyyrylainen, line 563: Aren't these purely for native libraries, so this should be in the native libraries section? And llvm unless I very misremember how I setup the cmake files, it doesn't default to clang so g++ being installed is just fine with the gold linker.

doc/setup_instructions.md
```diff
@@ -50,11 +50,19 @@ LFS. Note that the GitHub option to download as .zip will not work
 
 You need at least Git LFS version 2.8.0, old versions do not work.
 
-On Linux use your package manager to install Git. On Mac install the
-package manager [homebrew](https://brew.sh/) if you don't already have
-it, and use it to install Git. On Mac and Linux Git LFS is likely available
-as a package named `git-lfs`. If it is not install it manually. After
-installing remember to run `git lfs install` in terminal.
+On Linux use your package manager to install Git. This can be done in the
+terminal - for instance, on Ubuntu, type: `sudo apt-get install git` . On
```
hhyyrylainen, line 54: This section is going to get really long if we start to list all the package manager commands. I'd be fine with keeping text saying something like: "This can be done in the terminal with your distro's package manager (search online for what your distro's package manager is if you don't know)"

doc/setup_instructions.md
```diff
@@ -50,11 +50,19 @@ LFS. Note that the GitHub option to download as .zip will not work
...
+already have it, and use it to install Git. On Mac and Linux Git LFS is
+likely available as a package named `git-lfs`. If it is not, install it
+manually. You can download Git LFS from https://git-lfs.com. After
+downloading the file, extract it wherever you want. In the terminal, change
+directory to the newly extracted folder (e.g. " cd ~/Games/git-lfs-3.4.1 ").
+You can then install it via the terminal - for instance on Ubuntu, type:
+`sudo bash ./install.sh`.
+OR, install Git LFS entirely by typing in the terminal - for instance on
+Ubuntu, type: `git lfs install`. 
+It should then say "Git LFS initialized."
```
hhyyrylainen, line 65: This last part no longer makes sense to me, because even if you use a package manager to install git LFS, you *have to* run the install command. So if you want to put in the manual install command (which BTW is not distro specific, and should probably be in the form on `sudo ./install.sh` to not force bash to evaluate the file if the file shebang line asks for example just sh) I'd make it much more clear that that part was end of that section. Probably would be clearer to make the git lfs install part a separate paragraph.

hhyyrylainen (changes requested):

hhyyrylainen (approved):
I updated the text more to my liking to make it less distro specific and improved the flow of the steps a bit in my opinion.

PR #5521, opened 2024-09-20
Fix Thriveopedia search

dligr:

Fixes Thriveopedia search

**Related Issues**

Thriveopedia's search seems to be broken currently. Sometimes items aren't shown when they have to be, and after removing search prompt, page tree breaks. It seems to be caused by pages being added in wrong order (error messages confirm this), which is probably an issue with Dictionary's element order.

This PR avoids those problems by hiding tree items instead of completely rebuilding the tree (which wasn't possible before Godot 4.0, including when the search was first implemented: #4488)

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/thriveopedia/Thriveopedia.cs
```diff
@@ -618,24 +618,50 @@ private void OnForwardPressed()
...
-        pageTree.CreateItem();
+        stageDropdown.Visible = false;
 
-        // Need to copy to new list as loop modifies the list
-        var pageList = allPages.Keys.ToList();
-        foreach (var page in pageList)
+        foreach (var page in allPages)
         {
-            var children = GetAllChildren(page);
+            var visible = page.Key.TranslatedPageName.ToLower().Contains(newText.ToLower());
```
hhyyrylainen, line 625: This should specify the culture explicitly. Maybe invariant culture would be preferred or maybe as this is a purely user facing thing this would be a case where current UI culture should be used.

dligr: Would this be the right way to do this?
`.ToLower(CultureInfo.CurrentCulture)`

I don't think using invariant culture makes sense here as the search should be culture-based

src/thriveopedia/Thriveopedia.cs
```diff
@@ -618,24 +619,47 @@ private void OnForwardPressed()
...
+        stageDropdown.Visible = false;
 
-        // Need to copy to new list as loop modifies the list
-        var pageList = allPages.Keys.ToList();
-        foreach (var page in pageList)
+        foreach (var page in allPages)
         {
-            var children = GetAllChildren(page);
+            var visible = page.Key.TranslatedPageName.ToLower(CultureInfo.CurrentCulture)
+                .Contains(newText.ToLower(CultureInfo.CurrentCulture));
```
hhyyrylainen, line 627: One more comment on this line. Why isn't the `newText.ToLower...` call outside the loop? Surely that cannot change each loop iteration and this way we save on creating new strings from the new text ToLower call once per loop iteration.

hhyyrylainen (approved):
Thanks for the fix. This looks like it makes search work pretty well again.

PR #5535, opened 2024-09-24
Change space and industrial visuals

### LINKED ISSUE #5225: Space stage graphics need tweaking (stars, planet graphics, light amount)
I mostly fixed the main menu space 3D background for Godot 4 so similar fixes need to be duplicated to the space prototype.

Light can probably be fixed by turning up the sunlight and copying the background environment from the main menu 3D scene 1.

### LINKED ISSUE #5226: Rocket particles need tweaking (it pulsates in space, presumably when the light is visible and when not as the rocket moves)

HexapodPhilosopher:

Changes planet visuals and rotation.
Changes comet visuals.
Adds rocket model, adds detail to spaceport in industrial stage, and changes rocket scaling, movement and emitters.
Adds spaceship model to space stage and sets it to turn towards target position before moving.

Still to do
* Set space ship to complete turn (it currently stops before reaching the exact right direction)
* Set thrusters on ship while moving
* Set up proper unit selection when switching to space
* Make separate function for ship movement

**Related Issues**

close #5225
temporarily fixes #5226  (engine thrusters are currently inactive in space)

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/MainMenu.tscn
```diff
@@ -116,7 +116,7 @@ theme = ExtResource("1")
 script = ExtResource("6")
 ThriveLogoPath = NodePath("MenuContainers/LogoCenterContainer/Logo")
 MenuBackgrounds = [ExtResource("8"), ExtResource("7"), ExtResource("9")]
-Menu3DBackgroundScenes = ["res://assets/models/menu_backgrounds/MainMenuScene1.tscn", "res://assets/models/menu_backgrounds/MainMenuScene2.tscn", "res://assets/models/menu_backgrounds/MainMenuScene3.tscn"]
+Menu3DBackgroundScenes = ["res://assets/models/menu_backgrounds/MainMenuScene1.tscn"]
```
hhyyrylainen, line 119: This looks like an unintentional change...

assets/models/menu_backgrounds/Planetclouds_Material.tres
```diff
@@ -1,12 +1,12 @@
 [gd_resource type="StandardMaterial3D" load_steps=3 format=3 uid="uid://dcy33qrmimba4"]
 
-[ext_resource type="Texture2D" uid="uid://dgrmvarreasks" path="res://assets/textures/menu_backgrounds/PlanetClouds_Mat_Opacity.png" id="1"]
-[ext_resource type="Texture2D" uid="uid://bqrqlrqq1vlwa" path="res://assets/textures/menu_backgrounds/PlanetClouds_Mat_Normal_DirectX.png" id="2"]
+[ext_resource type="Texture2D" uid="uid://dgrmvarreasks" path="res://assets/textures/menu_backgrounds/PlanetClouds_Albedo_Opacity..png" id="1"]
+[ext_resource type="Texture2D" uid="uid://bdm8ocmn6s0w7" path="res://assets/textures/menu_backgrounds/PlanetClouds_Normal..png" id="2_0wv30"]
```
hhyyrylainen, line 4: This filename here looks like it has an extra dot `..png`

src/space_stage/SpaceStage.cs
```diff
@@ -280,14 +280,15 @@ public void SetupForExistingGameFromAnotherStage(bool spawnPlanet, UnitType spac
...
             AddPlanet(Transform3D.Identity, true);
 
-        var fleet = AddFleet(new Transform3D(Basis.Identity, new Vector3(6, 0, 0)),
+        var fleet = AddFleet(new Transform3D(Basis.Identity, new Vector3(20, 0, 0)),
             spacecraft, true);
 
         // Focus the camera initially on the ship to make the stage transition smoother
         ZoomOutFromFleet(fleet);
 
+        // TODO: Set launched craft to move from the planet
```
hhyyrylainen, line 289: Was this intended to not be completed for this?

src/general/MainMenu.tscn
HexapodPhilosopher: Yes, that was for testing.

src/space_stage/FleetMovementOrder.cs
```diff
@@ -18,22 +18,46 @@ public FleetMovementOrder(SpaceFleet fleet, Vector3 targetPosition) : base(fleet
...
 
-        if (distanceToTarget < adjustedSpeed)
+        if (direction.Dot(Unit.Transform.Basis.Z) < 0.99f)
         {
-            Unit.GlobalPosition = TargetPosition;
-            return true;
+            var smoothRotation = currentRotation.Slerp(targetRotation, 0.8f * delta);
+            Unit.Transform = new Transform3D(new Basis(smoothRotation), unitPosition);
+
+            Unit.Transform = Unit.Transform.Orthonormalized();
```
hhyyrylainen, line 39: Why does this use `Orthonormalized` here? I've never used that in Godot.

src/industrial_stage/IndustrialStage.cs
```diff
@@ -418,7 +425,7 @@ private void SwitchToSpaceScene()
         {
             GD.PrintErr("Spacecraft to put in initial fleet not found, using fallback unit");
 
-            launchedSpacecraftType = SimulationParameters.Instance.GetUnitType("simpleSpaceRocket");
+            launchedSpacecraftType = SimulationParameters.Instance.GetUnitType("advancedSpaceship");
```
hhyyrylainen, line 428: In normal situations this code should never trigger. So I think it makes sense to keep the basic stuff here, because otherwise players might be able to find an exploit to skip directly to the more advanced unit type.

src/space_stage/SpaceStage.cs
HexapodPhilosopher: As it's now switching to a separate ship, I felt it was unnecessary. I could revert that if you think it'd look better.

hhyyrylainen: I think it looks much better with the unit moving initially. And this code would need to be fixed up and re-enabled when the transition from industrial to space is being worked on in the far future.

src/space_stage/FleetMovementOrder.cs
HexapodPhilosopher: That was based on a guide that recommended to do that when rotating - but actually the new transform would achieve the same thing anyway. I'll remove it.

src/space_stage/SpaceFleet.cs
```diff
@@ -5,13 +5,15 @@
...
 /// <summary>
-///   A fleet (or just one) ship out in space.
+///   A fleet of ships (or just one ship) out in space.
 /// </summary>
 public partial class SpaceFleet : Node3D, IEntityWithNameLabel, IStrategicUnit
 {
     [Export]
     public NodePath? VisualsParentPath;
 
+    // public Node _engineEmittor;
```
hhyyrylainen, line 15: I think this commented out code in this class (also below in this file) should be removed if it is not going to be implemented.

simulation_parameters/industrial_stage/units.json
```diff
@@ -3,7 +3,18 @@
     "Name": "UNIT_SIMPLE_ROCKET",
     "UnitIcon": "res://assets/textures/gui/bevel/technologies/UnknownTechnology.png",
     "VisualScene": "res://assets/models/space/PlaceholderRocket.tscn",
-    "SpaceVisuals": "res://assets/models/space/PlaceholderRocketSpace.tscn",
+    "SpaceVisuals": "res://assets/models/space/PlaceholderAdvancedSpaceship.tscn",
```
hhyyrylainen, line 6: Is this space visuals override here specifically needed still? I though the code change would make this work without having to hack the unit to look way different, which is a hidden bug now that will trigger at some point in the future.

src/space_stage/FleetMovementOrder.cs
```diff
@@ -18,22 +18,42 @@ public FleetMovementOrder(SpaceFleet fleet, Vector3 targetPosition) : base(fleet
...
+            Unit.Transform = new Transform3D(new Basis(smoothRotation), unitPosition);
 
-        var direction = toTarget / distanceToTarget;
-        Unit.GlobalPosition += direction * adjustedSpeed;
-        return false;
+            if (distanceToTarget < adjustedSpeed)
+            {
+                Unit.GlobalPosition = TargetPosition;
+
+                // Unit._engineEmittor.Emitting = false;
```
hhyyrylainen, line 49: I think it'd be better to make a property called `Moving` in a fleet and use that here so that the fleet class can then in the future easily be extended to handle visual changes when moving. I can do that if you need help with the code specifically.

assets/models/menu_backgrounds/Comet.tscn
```diff
@@ -1,94 +1,98 @@
...
+
+[sub_resource type="QuadMesh" id="QuadMesh_80136"]
+material = ExtResource("6_h1j6i")
+size = Vector2(0.5, 0.45)
 
-[sub_resource type="Gradient" id="12"]
-offsets = PackedFloat32Array(0, 0.438776, 0.99095)
-colors = PackedColorArray(1, 1, 1, 0.294118, 1, 1, 1, 0, 1, 1, 1, 0)
+[sub_resource type="Curve" id="Curve_122gh"]
+resource_local_to_scene = true
```
hhyyrylainen, line 37: Do these curves have to be local to the scene? Unless these are modified during runtime this should be unnecessary and is a waste of memory and processing time. (and I didn't see any code attached to this node that could do that).

assets/models/menu_backgrounds/MainMenuScene1.tscn
```diff
@@ -374,3 +415,5 @@ lifetime = 1.42
 preprocess = 2.08
 mesh = SubResource("38")
 gravity = Vector3(0, 0, 0)
+
+[editable path="MenuPlanet"]
```
hhyyrylainen, line 419: Would it be possible to fix the planet scene so that at least the space or the main menu scenes don't need to modify it? Right now it looks like both of the places that use the planet need to modify it to make it work, so the planet scene is hard to use right now if the default settings of it are unusable.

assets/models/menu_backgrounds/CometTailEnd.tres
```diff
@@ -0,0 +1,34 @@
+[gd_resource type="StandardMaterial3D" load_steps=4 format=3 uid="uid://d1e2we22pcrbu"]
+
+[ext_resource type="Texture2D" uid="uid://bh2oogng6fs2u" path="res://assets/textures/menu_backgrounds/RoundParticle.png" id="1_6bfm1"]
+
+[sub_resource type="Gradient" id="Gradient_vy355"]
+colors = PackedColorArray(0.2639, 0.58695, 0.91, 0.756863, 0, 0, 0, 0.603922)
+
+[sub_resource type="GradientTexture2D" id="GradientTexture2D_4gidx"]
+resource_local_to_scene = true
```
hhyyrylainen, line 9: This also has local to scene stuff.

assets/models/menu_backgrounds/MenuPlanetShadow.tscn
```diff
@@ -0,0 +1,50 @@
...
+render_priority = 1
+transparency = 4
+blend_mode = 3
+cull_mode = 2
+shading_mode = 0
+vertex_color_use_as_albedo = true
+albedo_texture = SubResource("36")
+
+[sub_resource type="StandardMaterial3D" id="StandardMaterial3D_kno1u"]
+resource_local_to_scene = true
```
hhyyrylainen, line 34: local to scene here as well.

assets/models/space/PlaceholderAdvancedSpaceship.tscn
```diff
@@ -0,0 +1,87 @@
...
+roughness = 0.7
+roughness_texture = ExtResource("1_7u0ip")
+roughness_texture_channel = 3
+emission_enabled = true
+emission = Color(0.0588235, 0.0588235, 0.0588235, 1)
+emission_energy_multiplier = 2.0
+emission_texture = ExtResource("2_sojrh")
+
+[sub_resource type="Curve" id="Curve_tq58m"]
+resource_local_to_scene = true
```
hhyyrylainen, line 23: This file also has local to scene stuff, I'll stop commenting on this now as it seems a ton of files have these. So could you please go through your own changes and look for the rest?

assets/textures/menu_backgrounds/SphereSkyMaterial_Base_color.png.import
```diff
@@ -2,21 +2,20 @@
...
 importer="texture"
 type="CompressedTexture2D"
-uid="uid://blntn0acli076"
-path.s3tc="res://.godot/imported/SphereSkyMaterial_Base_color.png-5187083fa936831c9a77057c5ae3c37b.s3tc.ctex"
+uid="uid://cxfwmrejyt75q"
+path="res://.godot/imported/SphereSkyMaterial_Base_color.png-5187083fa936831c9a77057c5ae3c37b.ctex"
 metadata={
-"imported_formats": ["s3tc_bptc"],
-"vram_texture": true
+"vram_texture": false
```
hhyyrylainen, line 8: Why is this texture no longer compressed? This will take a ton of more disk space and memory when loaded now.
If for visual quality it is absolutely necessary that's fine, but I'd first try tweaking the compression options to see if some of them would be good enough without using as much memory as an uncompressed texture.

hhyyrylainen:
I've now commented on everything I noticed on a full read through (except rest of the local to scene options I stopped commenting on after a few).

src/space_stage/SpaceFleet.cs
HexapodPhilosopher: Yes. That part could be done later.

hhyyrylainen: Then I'd say that this commented out code definitely needs to go, especially as the property here is named against our code naming conventions so uncommenting it would cause an immediate warning.

simulation_parameters/industrial_stage/units.json
HexapodPhilosopher: I didn't manage to get it working any other way, so that line is solely determining what appears in the space stage.

hhyyrylainen: This should be reverted anyway, I can help with the code implementation once the graphics files have been cleaned up.

src/space_stage/FleetMovementOrder.cs
HexapodPhilosopher: I would need help with that. My understanding of c# is still very limiting, and leads to many warning messages.

assets/models/menu_backgrounds/Comet.tscn
HexapodPhilosopher: No, that's unneeded. I just realised there's a whole load of those I missed when doing cleanup. I'll use that checkbox responsibly in future.

assets/models/menu_backgrounds/MainMenuScene1.tscn
HexapodPhilosopher: I was just working on that. I think part of the issue in how it was setup was due to the way Godot uses .glb scenes, which adds complexity. I'm switching to using two .mesh files for the planet layers - one basic sphere, and one with the normals flipped to use for the atmosphere (so it doesn't show in front of the planet). I've also made the planet in the menu scene local to there, to remove overlap. I wasn't sure what that specific _editable path_ line does.

assets/models/space/PlaceholderAdvancedSpaceship.tscn
HexapodPhilosopher: I think I started making these changes before I was clear about when to use that. I've realized that it's much more efficient for removing them to just search for them all, too.

assets/models/menu_backgrounds/MainMenuScene1.tscn
hhyyrylainen: It makes instanced scene child nodes editable. Which is fine for one-off customizations, but if the customizations are needed in multiple places. Then it is better to make one derived scene from the base scene, apply the customizations there, and then instantiate that derived scene where it is needed. This way the manual scene customizations only need to be in a single place.

assets/textures/menu_backgrounds/SphereSkyMaterial_Base_color.png.import
HexapodPhilosopher: I think that change was made randomly by Godot. I don't recall doing anything that should have affected that file, but it started being flagged as changed, and I didn't realize that was why. I'll change it back.

HexapodPhilosopher: Godot keeps reverting this whenever I change it. Even without it loaded in a scene.

hhyyrylainen: What do you mean by reverting it? You need to use the Godot import options panel to tweak the options back to what works for one used in a 3D scene. If you are just trying to revert this single line then it is likely incorrect as it depends on other settings that also must be changed at the same time.

HexapodPhilosopher: I've re-imported it to revert line 9, but not 8. I can import it as s3tc_bptc with the path endings changed from .ctex to .s3tc.ctex. I'm not sure which is better.

hhyyrylainen: That path ending if I remember correctly is just what Godot 4 now does, so the import paths changing to whatever godot wants is fine as long as the compression / other options are correct. The import path is just some godot internal stuff we don't need to worry about at all.

src/space_stage/SpaceStage.tscn
hhyyrylainen, line 1: There's a bug where structure placeholders are not drawn above the background so I'll be required to enable the old stars for now to make the prototype stage playable.

hhyyrylainen (approved):
I fixed up all problems I still found with this. Two things I couldn't fix:
- keeping the new stars background in the space stage (makes structure ghosts not render on top of it)
- the ascension gate now doesn't have shadows on it so it looks mostly like a formless mess so it needs a bit more visual tweaking

PR #5541, opened 2024-09-26
Fix mucocyst for colonies

### LINKED ISSUE #5525: Fix mucocyst activation for colony members
The state setting is done, but the MucocystSystem is marked as not running on attached entities. So that attribute needs to be removed and then made sure that that doesn't break anything. Probably should add a check that the mucocyst-using cell is not engulfed currently as that might be what the attached entity check is for originally.

Also the sound could be made to play only for the colony lead cell if it is too loud.

dligr:

Fixes mucocyst only activating on the lead cell of the colony

**Related Issues**

Closes #5525

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/MucocystSystem.cs
+++ b/src/microbe_stage/systems/MucocystSystem.cs
@@ -34,13 +34,6 @@ public sealed class MucocystSystem : AEntitySetSystem<float>
         if (control.MucocystEffectsApplied == wantsMucocyst && !control.MucocystEffectsApplied)
             return;
 
-        // Disable mucocyst if engulfed
-        if (entity.Get<Engulfable>().PhagocytosisStep != PhagocytosisPhase.None)
-        {
-            control.State = MicrobeState.Normal;
-            wantsMucocyst = false;
-        }
-
         if (wantsMucocyst)
         {
             ref var storage = ref entity.Get<CompoundStorage>();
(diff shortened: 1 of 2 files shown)


src/microbe_stage/systems/MucocystSystem.cs
```diff
@@ -37,6 +34,13 @@ protected override void Update(float delta, in Entity entity)
         if (control.MucocystEffectsApplied == wantsMucocyst && !control.MucocystEffectsApplied)
             return;
 
+        // Disable mucocyst if engulfed
+        if (entity.Get<Engulfable>().PhagocytosisStep != PhagocytosisPhase.None)
```
hhyyrylainen, line 38: I think it would be better to put this code where other special state resetting when engulfed handling is done, because as written this system has to be modified to say that this only processes entities with Engulfable component (and just reads it).

dligr: I have fixed it now. I'm not sure if this is the proper solution, as the special state resetting is technically already done in 'EngulfedHandlingSystem'. However, it runs after the mucocyst system, so mucocyst is deactivated with a slight delay, causing a flicker.
That could probably be fixed with a `[RunsAfter()]` attribute, but it didn't seem to help.

hhyyrylainen (approved):
I think this change is now good. I checked in the engulfed system and the way it disables engulf mode on the engulfed thing should work here as well. I trust that you playtested this enough so that I don't have to context switch from my current branch I'm working on to test this.

src/microbe_stage/systems/MucocystSystem.cs
hhyyrylainen: I assume the check to not allow entering mucocyst state when engulfed in the input handling should take care of this.

PR #5546, opened 2024-09-28
Create a photostudio priority system

### LINKED ISSUE #5526: Create a priority system for PhotoStudio to prioritise cell sprites and cell type selection buttons
So that image generation being busy doesn't leave more critical elements in the game blank.

dligr:

Changes photostudio's task queue to be a priority queue and adds priorities to image photographing tasks.

Currently, the priorities are in that order:
Tooltips > Early multicellular billboards > Cell preview buttons in the auto-evo report

**Related Issues**

Closes #5526

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/PhotoStudio.cs
+++ b/src/engine/PhotoStudio.cs
@@ -469,7 +469,13 @@ public partial class PhotoStudio : SubViewport
     {
         public int Compare((int, int) x, (int, int) y)
         {
-            return x.CompareTo(y);
+            if (x.Item1 < y.Item1)
+                return -1;
+
+            if (x.Item1 > y.Item1)
+                return 1;
+
+            return x.Item2.CompareTo(y.Item2);
         }
     }
 }
--- a/src/engine/PhotoStudio.cs
+++ b/src/engine/PhotoStudio.cs
@@ -39,7 +39,7 @@ public partial class PhotoStudio : SubViewport
     private readonly Dictionary<ISimulationPhotographable.SimulationType, IWorldSimulation> worldSimulations = new();
     private readonly Dictionary<IWorldSimulation, Node3D> simulationWorldRoots = new();
 
-    private readonly PriorityQueue<ImageTask, (int Priority, int Index)> tasks = new();
+    private readonly PriorityQueue<ImageTask, (int Priority, int Index)> tasks = new(new TaskComparer());
     private ImageTask? currentTask;
     private Step currentTaskStep = Step.NoTask;
 
@@ -463,5 +463,13 @@ public partial class PhotoStudio : SubViewport
         simulationWorldRoots[worldSimulation] = node;
 
         return node;
+    }
+
+    private class TaskComparer : IComparer<(int, int)>
+    {
+        public int Compare((int, int) x, (int, int) y)
+        {
+            return x.CompareTo(y);
+        }
     }
 }
--- a/src/engine/PhotoStudio.cs
+++ b/src/engine/PhotoStudio.cs
@@ -45,7 +45,7 @@ public partial class PhotoStudio : SubViewport
 
     /// <summary>
     ///   <see cref="PriorityQueue{TElement, TPriority}"/> doesn't guarantee the first-in-first-out order.
-    ///   This offset makes up for that, being increased for each consecutive image task.
+    ///   This index offset makes up for that, being increased for each consecutive image task.
     /// </summary>
     private int lastTaskIndex;
 
--- a/src/engine/PhotoStudio.cs
+++ b/src/engine/PhotoStudio.cs
@@ -47,7 +47,7 @@ public partial class PhotoStudio : SubViewport
     ///   <see cref="PriorityQueue{TElement, TPriority}"/> doesn't guarantee the first-in-first-out order.
     ///   This offset makes up for that, being increased for each consecutive image task.
     /// </summary>
-    private int taskPriorityOffset = 0;
+    private int taskPriorityOffset;
 
     private bool waitingForBackgroundOperation;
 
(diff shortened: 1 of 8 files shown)


hhyyrylainen:
The code looks fine to me (other than that one unnecessary zero initializer that the CI caught). I'll test this in-game later if no one else can do that.

CI09:
When I switched from auto-evo tab to timeline tab, the square in which cell preview is shown loaded didn't finish loading before timeline tab preview, is this expected? Sorry if I'm not clear but I'm not sure how to explain it.

src/engine/PhotoStudio.cs
```diff
@@ -39,10 +39,16 @@ public partial class PhotoStudio : SubViewport
     private readonly Dictionary<ISimulationPhotographable.SimulationType, IWorldSimulation> worldSimulations = new();
     private readonly Dictionary<IWorldSimulation, Node3D> simulationWorldRoots = new();
 
-    private readonly Queue<ImageTask> tasks = new();
+    private readonly PriorityQueue<ImageTask, int> tasks = new();
```
noname0443, line 42: Using `PriorityQueue<ImageTask, int>` with `taskPriorityOffset` looks unfancy to me. I think it's not a good idea to mix an offset priority with an index number of a task. I think the code would be better if you used `pair<int, int>` or a small structure with a Compare function. The first value should be the priority; the second value should be the index number.

dligr: I actually agree with this proposal. This is a cleaner approach, especially considering that without it, tasks of different priorities might have an incorrect order (in some edge cases).
According to the [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.tuple-2.system-icomparable-compareto?view=net-8.0), (int, int) tuples seem to be compared in the such way, that the second ints are used for comparison only when the first ones are equal, which is pretty much what we need here.

dligr:
> When I switched from auto-evo tab to timeline tab, the square in which cell preview is shown loaded didn't finish loading before timeline tab preview, is this expected? Sorry if I'm not clear but I'm not sure how to explain it.

If you mean that the species buttons in auto-evo report don't load before the timeline tab loads, then yes, this is the expected behavior. This PR is intended to make some species previews load faster (with more priority) than some other species previews.
For example, species tooltips should now load faster than species preview buttons in the auto-evo report.

HexapodPhilosopher:
This seems to completely eliminate the issues I was having. The multicell cell type preview images mostly load quickly, and the ones that don't load quickly are all loaded in less that 30 seconds. No permanent lockups to the images loading.
I am still getting continual error messages, though. It's this one over and over:
_void CellHexesPhotoBuilder.BuildHexStruct(): System.InvalidOperationException: Species is not initialized_

noname0443 (approved):

src/engine/PhotoStudio.cs
hhyyrylainen: I don't *fully* agree because in a case where a higher priority level is absolutely filled with tasks, it'd be extremely nice to still rarely allow lower priority items to get through (though in this case it would be beneficial to make newer tasks more important so that for example if the player a minute ago looked at some high priority images and no goes elsewhere in the game, then that other place might take multiple minutes to load rather than get to be in front of the queue).

I still agree that this current PR is an improvement and is probably fine for now.

hhyyrylainen:
I'm not 100% sure, but I'm pretty sure this causes boxing:

kuva

So I need to veto the current implementation because it can cause up to 100 extra object allocations per generated image increasing garbage collector pressure, which means that the new implementation is worse than before.

Is it possible to pass a custom comparer that wouldn't take in a nullable reference to the tuple? That way the comparison operation wouldn't need to box the arguments.

dligr:
> Is it possible to pass a custom comparer that wouldn't take in a nullable reference to the tuple? That way the comparison operation wouldn't need to box the arguments.

Seems to be possible ([documentation](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.priorityqueue-2.comparer?view=net-8.0#system-collections-generic-priorityqueue-2-comparer)). And since the priority is non-nullable, there wouldn't be boxing here in case of a non-generic implementation.

HexapodPhilosopher:
This seems to completely eliminate the issues I was having. The multicell cell type preview images mostly load quickly, and the ones that don't load quickly are all loaded in less that 30 seconds. No permanent lockups to the images loading.
I am still getting continual error messages, though. It's this one over and over (the same as I had before):
_void CellHexesPhotoBuilder.BuildHexStruct(): System.InvalidOperationException: Species is not initialized_

src/engine/PhotoStudio.cs
```diff
@@ -451,4 +464,12 @@ private Node3D CreateNewRoot(IWorldSimulation worldSimulation)
 
         return node;
     }
+
+    private class TaskComparer : IComparer<(int, int)>
+    {
+        public int Compare((int, int) x, (int, int) y)
+        {
+            return x.CompareTo(y);
```
hhyyrylainen, line 472: I tried to find the memory allocations using benchmark dot net, but I didn't see anything. Though that isn't conclusive as I was unable to get it to detect memory usage when I intentionally added stuff.

For maximum clarity in this critical system, I'd prefer the following a bit more verbose approach here:

```suggestion
            if (x.Item1 < y.Item1)
                return -1;

            if (x.Item1 > y.Item1)
                return 1;

            return x.Item2.CompareTo(y.Item2);
```

hhyyrylainen:
@HexapodPhilosopher 

Still no one else has been able to trigger:
>  void CellHexesPhotoBuilder.BuildHexStruct(): System.InvalidOperationException: Species is not initialized

So I think it is either really rare on everyone else's computer or something is wrong for you locally causing it. If you can find some way to consistently trigger it on someone else's computer I can then try to find out what goes wrong (it might be related to things taking really long to generate on your system so maybe something is deleting / reusing the species object before it is photographed or something like that).

hhyyrylainen (approved):
I think this now works, and my suggestion for the new comparer *was* the right way around (I was slightly worried I had a thought error and had written the code in reverse).

PR #5563, opened 2024-10-03
Add new SVG icons

### LINKED ISSUE #3776: Icons for Compounds and Environment panel buttons
Breaking this off from the "Allow closing environment panel" issue: https://github.com/Revolutionary-Games/Thrive/issues/3693

Now that the environment and compounds panels will have separate icons, it will be much easier to create appropriate icons for each. 

For the Compounds Panel, the panel simply represents the food/nutrients that the cell has stored. So we could represent this with an icon for food/nutrients. 

We could use an apple and wheat stalk like so:
image

Or a chicken drumstick icon like so:
image

And such an icon would then be reused for the Nutrient Panel in the macroscopic stages for continuity.

For environment, @camilomcatasus created some good icons in the linked issue above, but I feel like a petri dish is not the best since it evokes a small artificial environment, and though the globe icon is good we may want to use a globe icon somewhere else (so would not want repeat icons). So I was thinking perhaps we could use a weather icon like so (since I don't imagine it being used elsewhere):
image

Any other thoughts?

### LINKED ISSUE #5329: Create custom icon for siderophore shooting action
as right now it doesn't have an icon for it

HexapodPhilosopher:

Adds .svg versions of existing compound and environment icons.
Adds icons for Compound and Environment windows
Adds siderophore icon.
Adds macro Wood icon.

**Related Issues**

close #3776 
close #5329

**Progress Checklist**

      break existing features:
      members reviewing this PR)


assets/textures/gui/bevel/Siderophore.svg
hhyyrylainen, line 1: As this is an action it should be in the actions sub folder.

simulation_parameters/microbe_stage/compounds.json
```diff
@@ -31,47 +17,36 @@
       "g": 0.39,
       "b": 0.0
     },
-    "IconPath": "res://assets/textures/gui/bevel/ammonia.png"
+    "IconPath": "res://assets/textures/gui/bevel/Ammonia.svg"
   },
-  "phosphates": {
-    "Name": "PHOSPHATE",
+  "atp": {
```
hhyyrylainen, line 22: Could you keep the same order in this file that there was previously? That'd make it much easier to see what was actually changed. (even if technically it doesn't impact anything in the game)

HexapodPhilosopher: I thought it would be worth alphabetizing the compounds, but I see your point.

hhyyrylainen: I guess that could theoretically help, but the file is so long that it is basically the only good way to navigate it is to use "search forward" feature in a text editor.

hhyyrylainen:
Oh yeah, it came up on discord that the old files these replace should be deleted. Because otherwise someone will accidentally use one of the old files somewhere, and then we'll be in a situation where a mix of files from old and new are used at the same time and I don't want to get into that mess.

hhyyrylainen (approved):
At 720p game resolution I'm seeing slight "haze" around a few of the icons like iron and hydrogen sulfide (but not glucose). But I think this is an improvement, and it is very nice to have SVGs of the icons.

PR #5564, opened 2024-10-04
Flagella scaling autoevo

adQuid:

Adds functionality for auto-evo to scale flagella. Right now only sets to one of three sizes, but we can iterate on this moving forward.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
+++ b/src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
@@ -1,0 +1,42 @@
+﻿using AutoEvo;
+using System;
+using System.Collections.Frozen;
+using System.Collections.Generic;
+using System.Linq;
+
+internal class UpgradeOrganelle : IMutationStrategy<MicrobeSpecies>
+{
+
+    private readonly FrozenSet<OrganelleDefinition> allOrganelles;
+    private IComponentSpecificUpgrades upgrade;
+
+    public UpgradeOrganelle(Func<OrganelleDefinition, bool> criteria, IComponentSpecificUpgrades upgrade)
+    {
+        allOrganelles = SimulationParameters.Instance.GetAllOrganelles().Where(criteria).ToFrozenSet();
+        this.upgrade = upgrade;
+    }
+
+    public bool Repeatable => true;
+
+    public List<Tuple<MicrobeSpecies, float>>? MutationsOf(MicrobeSpecies baseSpecies, float mp, bool lawk,
+    Random random)
+    {
+        var mutated = new List<Tuple<MicrobeSpecies, float>>();
+
+        MicrobeSpecies newSpecies = (MicrobeSpecies)baseSpecies.Clone();
+
+        foreach (OrganelleTemplate organelle in newSpecies.Organelles.Where(x => allOrganelles.Contains(x.Definition)))
+        {
+
+            if (organelle.Upgrades == null)
+            {
+                organelle.Upgrades = new OrganelleUpgrades();
+            }
+            organelle.Upgrades.CustomUpgradeData = upgrade;
+        }
+
+        mutated.Add(new Tuple<MicrobeSpecies, float>(newSpecies, mp));
+
+        return mutated;
+    }
+}
--- a/src/microbe_stage/MicrobeInternalCalculations.cs
+++ b/src/microbe_stage/MicrobeInternalCalculations.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using AngleSharp;
 using Godot;
 using Systems;
 
@@ -212,6 +213,13 @@ public static class MicrobeInternalCalculations
 
                 float movementConstant =
                     Constants.FLAGELLA_BASE_FORCE * organelle.Definition.Components.Movement!.Momentum;
+
+                var configuration = organelle.Upgrades?.CustomUpgradeData;
+                if (configuration != null)
+                {
+                    if (configuration is FlagellumUpgrades flagellumUpgrades)
+                        movementConstant += Constants.FLAGELLA_MAX_UPGRADE_FORCE * flagellumUpgrades.LengthFraction;
+                }
 
                 if (!isBacteria)
                     movementConstant *= Constants.EUKARYOTIC_MOVEMENT_FORCE_MULTIPLIER;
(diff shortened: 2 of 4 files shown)


src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Frozen;
+using System.Collections.Generic;
+using System.Linq;
+using AutoEvo;
+
+internal class UpgradeOrganelle : IMutationStrategy<MicrobeSpecies>
```
hhyyrylainen, line 7: Why is this internal? Other mutation strategies are public.

src/microbe_stage/MicrobeInternalCalculations.cs
hhyyrylainen, line 1: I assume you added this code here due to it being forgotten in the initial flagellum upgrades PR?

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    private readonly FrozenSet<OrganelleDefinition> allOrganelles;
+    private IComponentSpecificUpgrades upgrade;
```
hhyyrylainen, line 10: Can't this also be readonly?

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -0,0 +1,42 @@
...
+        Random random)
+    {
+        if (allOrganelles.Count == 0)
+        {
+            return null;
+        }
+
+        var mutated = new List<Tuple<MicrobeSpecies, float>>();
+
+        MicrobeSpecies newSpecies = (MicrobeSpecies)baseSpecies.Clone();
```
hhyyrylainen, line 30: This still has the problem where this clones a species before knowing that it would actually apply any upgrades. So I'd prefer an approach where a copy is only made *after* it is known that it is possible. Also allocating the `mutated` list once needed would be a good memory use optimization.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        foreach (OrganelleTemplate organelle in newSpecies.Organelles.Where(x => allOrganelles.Contains(x.Definition)))
```
hhyyrylainen, line 32: This `.Where` should be written out manually instead of with LINQ as this is a hot path in auto-evo so the performance and memory use of LINQ is not preferable here.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            organelle.Upgrades ??= new OrganelleUpgrades();
+            organelle.Upgrades.CustomUpgradeData = upgrade;
```
hhyyrylainen, line 35: I was going to comment that it would be slightly more efficient to not clone the custom upgrade data here, but that would be quite longer code and require adding new configuration for the organelle clone method to not clone the upgrades. As all organelles are cloned anyway this probably doesn't matter so I won't ask you to do that unless you reall want to.

hhyyrylainen (changes requested):
I didn't test how it feels in game when all flagellum of a species may get modified by this mutation, but here are some code technical comments.

src/microbe_stage/MicrobeInternalCalculations.cs
adQuid: Correct. This is also the statistic that is used for the auto-evo calculations, so auto-evo wouldn't work without this functionality.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -0,0 +1,60 @@
...
+    private readonly FrozenSet<OrganelleDefinition> allOrganelles;
+    private readonly IComponentSpecificUpgrades upgrade;
+
+    public UpgradeOrganelle(Func<OrganelleDefinition, bool> criteria, IComponentSpecificUpgrades upgrade)
+    {
+        allOrganelles = SimulationParameters.Instance.GetAllOrganelles().Where(criteria).ToFrozenSet();
+        this.upgrade = upgrade;
+    }
+
+    public bool Repeatable => true;
```
hhyyrylainen, line 18: Is this step repeatable BTW because this doesn't do anything different if this is called multiple times on the same species?

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            var mutated = new List<Tuple<MicrobeSpecies, float>>();
+
+            MicrobeSpecies newSpecies = (MicrobeSpecies)baseSpecies.Clone();
+
+            foreach (OrganelleTemplate organelle in newSpecies.Organelles)
+            {
+                if (allOrganelles.Contains(organelle.Definition))
+                {
+                    organelle.Upgrades ??= new OrganelleUpgrades();
```
hhyyrylainen, line 48: Could you add a TODO comment here that once upgrades cost MP this should take the cost into account?

hhyyrylainen (changes requested):
The code is now almost perfect, I just have 2 tiny comments on a few things I noticed now.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
adQuid: honestly I had just assumed that "repeatable" just mean it wouldn't break if you ran it twice. I'll set it to false.

hhyyrylainen: That's true, but the auto-evo system will burn unnecessary CPU cycles doing a mutation again that doesn't result in any changes (if I understood the new core code correctly, any repeatable mutation modifier can be tried again on the same species once it has been already applied if there is any MP left, which there will be as your code doesn't reduce the MP count).

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
adQuid: comment added (doesn't show your comment as outdated since it's a different line technically)

hhyyrylainen (approved):

PR #5571, opened 2024-10-06
Implement a UV unwrapper for 3D creatures

dligr:

Implements a UV unwrapper to apply textures to 3D creatures.
Currently uses a placeholder iron texture.

**To do:**
- [ ] <s>Make a procedural texture generator</s> (to be done in future)
- [x] Probably make it run asynchronously after mesh generation, without stopping the game
- [ ] <s>Reduce amount of calculations for bigger meshes</s> (doesn't seem to be possible because of how xatlas works)

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/extension/atlas/ExtendedArrayMesh.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "atlas_unwrap.hpp"
+#include "ExtendedArrayMesh.hpp"
+
+#include <godot_cpp/core/class_db.hpp>
+
+#include <godot_cpp/classes/array_mesh.hpp>
+
+namespace Thrive
+{
+	void ExtendedArrayMesh::_bind_methods()
```
hhyyrylainen, line 12: You should run clang-format on all the cpp and hpp files added / changed by this PR.

src/extension/atlas/atlas_unwrap.cpp
```diff
@@ -0,0 +1,202 @@
...
+#include <godot_cpp/classes/mesh.hpp>
+#include <godot_cpp/classes/array_mesh.hpp>
+#include <godot_cpp/classes/surface_tool.hpp>
+#include <godot_cpp/classes/ref.hpp>
+#include "godot_cpp/templates/pair.hpp"
+#include "godot_cpp/templates/local_vector.hpp"
+
+using namespace godot;
+
+int Thrive::TestFunc(int test)
```
hhyyrylainen, line 46: This seems like leftover debugging code.

src/extension/atlas/atlas_unwrap.cpp
```diff
@@ -0,0 +1,202 @@
+/**************************************************************************/
```
hhyyrylainen, line 1: Is this actually code copy-pasted from the engine? Is there no way to call this method through the GDExtension interface?

src/extension/interop/ExtensionInterop.h
```diff
@@ -6,6 +6,13 @@
 
 #include "interop/CStructures.h"
 
+#include "atlas/atlas_unwrap.hpp"
+
+#include <godot_cpp/classes/mesh.hpp>
+#include <godot_cpp/classes/array_mesh.hpp>
+
```
hhyyrylainen, line 13: This file is meant to be pure C code, so this is wrong to include these files here at all.

src/engine/ExtendedArrayMesh.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Godot;
+
+/// <summary>
+///   Wrapper of an <see cref="ArrayMesh"/> with additional functionality
+/// </summary>
+public class ExtendedArrayMesh
```
hhyyrylainen, line 6: On reviewing how this is used, this seems like a totally unnecessary thing. The unwrap method just adjusts the internal ArrayMesh, so a freestanding C++ method exposed through the interop way should be able to perform the exact same operation without needing to allocate an extra object instance.

src/late_multicellular_stage/MulticellularConvolutionDispayer.cs
```diff
@@ -82,7 +81,7 @@ protected override void Dispose(bool disposing)
     {
         if (disposing)
         {
-            material?.Dispose();
+            // ...
```
hhyyrylainen, line 84: This dispose method likely ends up empty now and it can be just entirely removed.

src/engine/ExtendedArrayMesh.cs
dligr: Would `mesh.NativeInstance` be the right way to get an ArrayMesh's pointer in C#?

src/extension/atlas/atlas_unwrap.cpp
dligr: It is copy-pasted from the engine; it isn't possible to call the original method, because it's heavily tied to lightmaps. Additionally, we might want to make some changes to its functionality in future, so I think that it makes to keep it as a separate method

src/engine/ExtendedArrayMesh.cs
hhyyrylainen: I'm not sure such a property exists, but internally godot calls `GodotObject.GetPtr(mesh)` to get the native instance pointer when it is calling the C# bound methods. We *probably* need to use a bit of reflection to get into the godot internals to be able to grab the pointer.

src/extension/atlas/atlas_unwrap.cpp
```diff
@@ -0,0 +1,197 @@
...
+/*  register_types.cpp                                                    */
```
hhyyrylainen, line 2: Are these the exact copyright headers? Because I think it is supposed to have the file name here which doesn't match the name of the file (so I'm just checking that you didn't accidentally grab the wrong headers).

src/extension/atlas/atlas_unwrap.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include "Include.h"
+
+#include "core/NativeLibIntercommunication.hpp"
+
+#include <godot_cpp/core/class_db.hpp>
+
+#include <godot_cpp/classes/mesh.hpp>
+#include <godot_cpp/classes/array_mesh.hpp>
+#include <godot_cpp/classes/surface_tool.hpp>
+
```
hhyyrylainen, line 44: It looks like there's an absolute ton of unnecessary includes, array_mesh.hpp seems to me like the only one that should be here.

src/extension/atlas/atlas_unwrap.cpp
hhyyrylainen: Alright, yep it looks like it is not exposed from the engine (looking in the godot-cpp project). So I guess this needs to be done like this.

src/extension/core/RegisterThriveTypes.cpp
```diff
@@ -14,7 +14,7 @@ void InitializeThriveModule(godot::ModuleInitializationLevel level)
     {
         return;
     }
-
+
```
hhyyrylainen, line 17: non-clang format conforming change here

src/extension/interop/ExtensionInterop.h
```diff
@@ -22,6 +22,8 @@ extern "C"
     /// \return The version number
     [[maybe_unused]] API_EXPORT int32_t ExtensionGetVersion(ThriveConfig* thriveConfig);
 
+    [[maybe_unused]] API_EXPORT bool Unwrap(float texelSize, ArrayMesh* mesh);
```
hhyyrylainen, line 25: For consistency I'd prefer to swap the arguments here so that the object the method is called on is always first. Also as this is a C file the function name "Unwrap" is the *entire* thing that resolves the method (not the parameter names) so having this short of a function name is not great here. I'd suggest naming this `ArrayMeshUnwrap(ArrayMesh* mesh, float texelSize)`

src/extension/interop/ExtensionInterop.cpp
```diff
@@ -34,6 +37,11 @@ void DebugDrawerAddLine(DebugDrawer* drawerInstance, JVecF3* from, JVecF3* to, J
             *reinterpret_cast<godot::Color*>(colour));
 }
 
+bool Unwrap(float texelSize, ArrayMesh* mesh)
+{
+	return Thrive::Unwrap(texelSize, *reinterpret_cast<godot::ArrayMesh*>(mesh));
```
hhyyrylainen, line 42: I'm pretty sure this is a tab character which clang-format shouldn't put according to our configured style.

src/extension/atlas/atlas_unwrap.hpp
```diff
@@ -0,0 +1,50 @@
...
+namespace Thrive
+{
+
+bool Unwrap(float texelSize, godot::ArrayMesh& mesh);
```
hhyyrylainen, line 48: I'd swap the argument order here as well even if this is not absolutely necessary.

src/extension/atlas/atlas_unwrap.cpp
```diff
@@ -0,0 +1,197 @@
...
+	PackedFloat32Array rweights = arrays[Mesh::ARRAY_WEIGHTS];
+	
+	for (int i = 0; i < ic; i += 3) {
+		
+		for (int j = 0; j < 3; j++) {
+			// Index of an original vertex corresponding to the generated one.
+			int vertex_id = output.vertexArray[output.indexArray[i + j]].xref;
+			
+			if (surface_format & Mesh::ARRAY_FORMAT_COLOR) {
+                surfaces_tools->set_color(rcolors[vertex_id]);
```
hhyyrylainen, line 162: I'd run clang format on this file even if this is meant to be mostly copy-pasted code but I assume you've made some changes here.

If you've made multiple changes that it is unrealistic to expect to copy-paste the engine source code over this again, that's fine to then keep the formatting mostly as-is. But if you have made a lot of changes already, I'd ask to remove the trailing comments on lines (our style is to put comments on their own lines before the line they comment on).

hhyyrylainen (changes requested):

src/extension/atlas/atlas_unwrap.cpp
dligr: Those are the right copyright headers, I just decided to change the file name.

Here's the original file:
https://github.com/godotengine/godot/blob/master/modules/xatlas_unwrap/register_types.cpp

dligr: Thinking about it, this license probably doesn't apply to atlas_unwrap.hpp, considering that no parts of it were copied from Godot

hhyyrylainen: I see, it is certainly common that a bunch of godot files are called "register_types".

src/engine/ExtendedArrayMesh.cs
dligr: This field does exist, and ultimately it seems to do the same as `GetPtr()`:
`public IntPtr NativeInstance => NativePtr;`

However, the C++ code crashes when called with such a pointer. From using some debug returns, I can conclude that it crashes when a mesh's function is called. 
So, it seems like there's some kind of issue with the pointer passed.

hhyyrylainen: Could it be that the pointers passed to the C# side are adjusted somehow? That's about the only thing I can think off. It should be possible to use a C++ debugger to add a breakpoint in the extension and then inspect the passed in pointer as if it was a mesh to see if it is correct or not. If the pointer is not good then I see just three possibilities: the pointer access in C# side is wrong, the Godot engine does some pointer offsetting that are passed to/from C# so we need to read the godot source code to find out how to undo that ourselves, or I think you can script gdb for example to just try incrementing and decrementing the pointer to see if any small-ish offset results in valid data appearing.

src/extension/interop/ExtensionInterop.cs
```diff
@@ -121,4 +121,7 @@ internal static partial class NativeMethods
 {
     [DllImport("thrive_extension")]
     internal static extern int ExtensionGetVersion(IntPtr thriveConfig);
+
+    [DllImport("thrive_extension")]
+    internal static extern int ArrayMeshUnwrap(IntPtr mesh, float texelSize);
```
hhyyrylainen, line 126:
```suggestion
    internal static extern bool ArrayMeshUnwrap(IntPtr mesh, float texelSize);
```
Just noticed this return value mismatch with the C++ code.

src/extension/atlas/atlas_unwrap.cpp
```diff
@@ -0,0 +1,210 @@
...
+/*                                                                        */
+/* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,        */
+/* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF     */
+/* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. */
+/* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY   */
+/* CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,   */
+/* TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE      */
+/* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                 */
+/**************************************************************************/
+
```
hhyyrylainen, line 30: As this file is modified, I think we should put extra text here mentioning that this file is based on the godot engine source, but is modified for Thrive.

hhyyrylainen:
I guess pretty much all that's left is to figure out how to get the right native pointer to be passed to the extension. And maybe some kind of texture? (because I think just defaulting to iron skin wouldn't be that great)

dligr:
The unwrapping should also be made to run in the background after mesh generation, as otherwise the game freezes for too long.

dligr:
The approach with sending a Variant worked. 
Now this PR can actually be tested in-game (although I haven't yet worked on any of the to-do features).

src/extension/interop/ExtensionInterop.h
```diff
@@ -22,9 +22,12 @@ extern "C"
     /// \return The version number
     [[maybe_unused]] API_EXPORT int32_t ExtensionGetVersion(ThriveConfig* thriveConfig);
 
+    [[maybe_unused]] API_EXPORT bool ArrayMeshUnwrap(ArrayMeshVariant* mesh, float texelSize);
```
hhyyrylainen, line 25: As this method technically takes in just a godot variant, I think the opaqua handle should be called `GodotVariant` rather than `ArrayMeshVariant` as the object type is just `godot::Variant` and nothing specific to the mesh.

src/extension/interop/ExtensionInterop.cpp
```diff
@@ -34,6 +39,19 @@ void DebugDrawerAddLine(DebugDrawer* drawerInstance, JVecF3* from, JVecF3* to, J
             *reinterpret_cast<godot::Color*>(colour));
 }
 
+bool ArrayMeshUnwrap(ArrayMeshVariant* mesh, float texelSize)
+{
+    godot::Variant* var = reinterpret_cast<godot::Variant*>(mesh);
+
+    godot::Object* obj = (godot::Object*)var;
+
+    obj->is_class("ArrayMesh");
```
hhyyrylainen, line 48: Shouldn't this fail if it is not the right class? And above there should be a check if the cast to object failed because someone could pass in a variant that just holds like the number 12.

src/late_multicellular_stage/MulticellularConvolutionDispayer.cs
```diff
@@ -75,17 +76,30 @@ public void DisplayFromLayout(IReadOnlyCollection<MulticellularMetaball> layout)
         Mesh = meshGen.DualContour();
         Mesh.SurfaceSetMaterial(0, material);
 
+        Task uvUnwrap = new Task(() => UVUnwrapAndTexturize((Mesh as ArrayMesh)!));
```
hhyyrylainen, line 79: If the cast must succeed, shouldn't this use the cast syntax that ensures it succeeds or throws (`(ArrayMesh)Mesh`)?

src/late_multicellular_stage/MulticellularConvolutionDispayer.cs
```diff
@@ -75,17 +76,30 @@ public void DisplayFromLayout(IReadOnlyCollection<MulticellularMetaball> layout)
...
     }
 
-    protected override void Dispose(bool disposing)
+    private void UVUnwrapAndTexturize(ArrayMesh mesh)
     {
-        if (disposing)
-        {
-            material?.Dispose();
-        }
+        var nativeVariant = Variant.From(mesh).CopyNativeVariant();
```
hhyyrylainen, line 87: What's the logic with CopyNativeVariant? I've never used that before. Why not just `new Variant(mesh)`?

dligr: I assumed that C++ side wouldn't be able to use a C# Variant, but it actually seems to work

hhyyrylainen: It does make sense that the C# bindings would try to match the memory layout and size of the C++ side. Though for safety it might be nice to add a static assert in the C++ code and a startup check in C# side to make sure the size of the types match a known value.

dligr:
Since there wasn't any concept work on what the 3D creatures' skin textures should look like (and I believe that it would take quite some time to discuss it and to then implement), I propose that the texture generator should be done some time in the future. This PR would then just add a UV unwrapper to serve as a foundation of this future functionality. It would have to be disabled though, since it's not really useful without textures.

hhyyrylainen:
I think that's a fine enough conclusion. Once you are ready for this to be reviewed and merged we'll need to bump the native library version numbers.

dligr:
This is ready for review now.

Note:
- [ ] Disable the unwrapper before merging

dligr:
Regarding making the native libs, I'm not really sure how to do that so I think that I would leave it to you.

hhyyrylainen:
When running to do final checks I'm getting an error like: `Cannot marshal field '_obj' of type 'Godot.Variant': Invalid managed/unmanaged type combination (Marshaling to and from COM VARIANTs isn't supported).`

I'll investigate and report back *if* I can't find a solution.

hhyyrylainen:
I think I found the solution which is: `Variant.From(mesh).CopyNativeVariant()` and then passing that as an `in reference` (which probably causes some extra copies but in terms of how expensive the operation is that probably doesn't matter).

hhyyrylainen:
While testing I saw errors like this, but I'll re-disable the code and merge:

kuva

as that seems maybe an error related to the change to using invoke. Or maybe it is something else but as this code is now disabled it should be fine for now to be in there.

hhyyrylainen (approved):
This should be fine now with the various tweaks I did. Most of the time I spent trying to figure out how to get the code to work on Linux as well.
I think unrelated to this the latest Jolt version no longer works correctly (which I'll investigate tomorrow and make a separate PR).

PR #5581, opened 2024-10-09
Animated Distortion Shaders on Background

Deus-Codes:

Implements distortion shaders on the background, giving a distortion effect to replicate the movement of fluids. Creates motion in the parallax even when the player is static.

Note that the pull request shows more lines added/subtracted than actually occurred, so apologies for any inconvenience. I had to copy and paste the code from another area. Let me know if I should submit another pull request without such a drastic CtrlC+CtrlV.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


shaders/MicrobeBackground.gdshader
```diff
@@ -1,64 +1,77 @@
...
-    else
-    {
-        // Night
+    } else {
         return mix(vec3(0.052f, 0.05f, 0.17f), vec3(0.25f, 0.25f, 0.25f), 4.0f * amount);
     }
 }
 
-void vertex(){
+void vertex() {
```
hhyyrylainen, line 37: Not doing these formatting changes (so undoing these) would make the diff a lot clearer, and avoiding reordering the speed variables etc. at the start of the file would help a lot with seeing what is going on.

shaders/MicrobeBackground.gdshader
```diff
@@ -1,64 +1,77 @@
...
-void vertex(){
+void vertex() { 
     vec2 offset = (repeats - 1.0f) / 2.0f;
     vec2 worldPos = (INV_VIEW_MATRIX * vec4(0.0f, 0.0f, 0.0f, 1.0f)).xz;
 
-    UV = (UV + worldPos / (speed0 * repeats)) * repeats - offset;
-    UV2 = (0.12f + UV + worldPos / (speed1));
-    UV3 = (0.512f + UV + worldPos / (speed2));
-    UV4 = (0.05f + UV + worldPos / (speed3));
+    UV = (UV + worldPos / (speed0 * repeats)) * repeats - offset;
```
hhyyrylainen, line 41: And a lot of these spaces at the end of lines cause sections to be marked as changed. Any good text editor should have a feature to delete all trailing whitespace so these should be easy to get rid of.

Deus-Codes:
Addressed every format change not related to distortion effect expect for one at the end involving a space, which I am not sure exactly what the issue is there.

shaders/MicrobeBackground.gdshader
```diff
@@ -10,6 +10,12 @@ uniform sampler2D layer1 : source_color;
 uniform sampler2D layer2 : source_color;
 uniform sampler2D layer3 : source_color;
 
+// the noise textures, distortion amount, and distortion speed for the distortion shader
+uniform sampler2D noise_tex_1;
+uniform sampler2D noise_tex_2;
```
hhyyrylainen, line 15: Without also a change to the microbe background godot scene, this PR shouldn't be able to work correctly as these textures are both empty. Also we use camelCase naming convention in Thrive so these new variables you've added should be named `noiseTex1`, `distortionFactor` etc. rather than be named with underscores.

shaders/MicrobeBackground.gdshader
```diff
@@ -10,6 +10,12 @@ uniform sampler2D layer1 : source_color;
 uniform sampler2D layer2 : source_color;
 uniform sampler2D layer3 : source_color;
 
+// the noise textures, distortion amount, and distortion speed for the distortion shader
```
hhyyrylainen, line 13:
```suggestion
// The noise textures, distortion amount, and distortion speed for the distortion shader
```

shaders/MicrobeBackground.gdshader
```diff
@@ -48,17 +54,33 @@ void vertex(){
 }
 
 void fragment(){
-    vec4 colour0 = texture(layer0, UV);
-    vec4 colour1 = texture(layer1, UV2);
-    vec4 colour2 = texture(layer2, UV3);
-    vec4 colour3 = texture(layer3, UV4);
+    //Application of noise textures on parallax layers, creating distortion. Subtracting time results in a flip of the direction of distortion, and multiplying by factor amplifies speed
```
hhyyrylainen, line 57: I'm pretty sure this line is too long so I'd do a change like this.
```suggestion
    // Application of noise textures on parallax layers, creating distortion. 
    // Subtracting time results in a flip of the direction of distortion, and multiplying by factor amplifies speed.
```

shaders/MicrobeBackground.gdshader
```diff
@@ -48,17 +54,33 @@ void vertex(){
...
+    vec2 noise_uv1 = UV + TIME * distortion_speed;
+    vec2 noise_uv2 = UV2 + TIME * distortion_speed * .8;
+    vec2 noise_uv3 = UV3 - TIME * distortion_speed * 1.5;
+    vec2 noise_uv4 = UV4 - TIME * distortion_speed;
+
```
hhyyrylainen, line 62:
```suggestion
```

hhyyrylainen:
> Addressed every format change not related to distortion effect expect for one at the end involving a space, which I am not sure exactly what the issue is there.

I don't think I can see which bit this is about. If you are asking about the last red "-" line, that just is saying that you've removed a blank line from between these lines:
```
    ALBEDO.rgb = composition.rgb * LightInfluence(lightLevel);
(like in the original version there's a blank line here, but this line doesn't exist at all in your version)
    ALPHA = 1.0f;
```

shaders/MicrobeBackground.gdshader
Deus-Codes: Confirmed that it doesn't work, and will change the naming convention. 

About the microbe background scene - is there a specific .tscn for the microbe background? I'm trying to find it on godot, but can only find microbestage.tscn and this background shader file.

hhyyrylainen: The background is in the microbe camera scene as the background is attached to the camera (as that's how it stays always visible):

kuva

Deus-Codes:
Playtesting it shows that the backgrounds now are animated. And as a bonus, are also animated during the editor too, which wasn't so when I was just tweaking the shader. 

Just to clarify, should only variables should be camelCase? Some of the uv textures were noted like "uv_texture", but I noticed that this was true in priorly-existing code for the parallax layers as well. 

I'd like input from everyone on if they think the effect should be altered in anyway - I can't do much besides altering distortion strength and speed, but even that has a strong effect. 

Thanks for the help hhyyrylainen!

CI09:
Playtest deems it working as of latest commit.

hhyyrylainen (approved):
I did a few style fixes and disabled this by default (for now, I'll add an option menu entry to turn this on). The bigger change I did is that I added a mode where the layers 1-3 don't distort, but instead move tiny bit based on the elapsed time to give those layers a bit more movement (so only layer 0 of the background moves).

Issue #4548, opened 2023-11-06
Make parts tooltips be generated automatically from JSON


Athriq:
They make up the bulk of ToolTipManager.tscn, unnecessarily making things harder to maintain. Tooltip descriptions and other related contents should be moved to JSON so that everything would be JSON-driven. Anyone's free to help. Comment below if something's not clear.

hhyyrylainen:
This is now done for organelle part tooltips: https://github.com/Revolutionary-Games/Thrive/pull/5585 

but everything else is still missing (at least membranes).

hhyyrylainen:
Though there is still at least something missing from the JSON tooltips as they still cause a *slight* crash when trying to add new organelles.

Issue #1659, opened 2020-09-29
Split ToolTipManager.tscn to multiple files


hhyyrylainen:
to not have one huge file with all of the tooltips, instead sub scenes one for each GUI scene should be used.

hhyyrylainen:
And also none of the actual tooltips should be in a file called "Tooltip**Manager**" as I think it's a bit confusing that the actual tooltips are in there. So instead a new folder like tooltips could be added which contains things like:
- GeneralTooltips.tscn
- MicrobeEditorTooltips.tscn
- MicrobeStageTooltips.tscn

etc.

hhyyrylainen:
It makes sense to do this other issue first as it'll reduce the amount of work in this issue: https://github.com/Revolutionary-Games/Thrive/issues/4548

PR #5585, opened 2024-10-12
Organelle json tooltips

### LINKED ISSUE #2306: Make organelle storage amounts in tooltips use localization
Right now values like "0.5" aren't localized to use locale specific decimal separators. See: https://github.com/Revolutionary-Games/Thrive/pull/2304#discussion_r650359918

This should be changed to use localization. Might also be worth it at the same time to make the values be dynamically generated based on the actual values loaded from the organelles json so that changing the storage values doesn't need to be updated in multiple places.

CI09:

Replaces manually created organelle tooltips with JSON generated thing.

**Related Issues**
~~What about #1659?~~ This doesn't address that -hyyrylainen

half does:  #4548
closes #2306

**Progress Checklist**

      break existing features:
      members reviewing this PR)


simulation_parameters/SimulationParameters.cs
```diff
@@ -230,6 +230,8 @@ public override void _Ready()
         // This is done this way to allow StartupActions to run before SimulationParameters are loaded
         ApplyTranslations();
 
+        CreateOrganelleTooltips();
```
hhyyrylainen, line 233: In my opinion this is the wrong place to do this, my comment from Discord:
> I don't think the tooltip setup should be in the simulation parameters, it should instead be in ToolTipManager

> If the logic gets long I'd put it in a different class that is then just called by the tooltip manager, maybe something like ProceduralToolTipGenerator

src/microbe_stage/OrganelleDefinition.cs
hhyyrylainen, line 1: Isn't the other part of the description missing? Most organelles have a short "processes" description at the top and then at the bottom of the tooltip there's the scientific flavour text description.

src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -516,4 +518,32 @@ private void HideAllToolTips()
             }
         }
     }
+
+    private void CreateOrganelleTooltips()
+    {
+        if (SimulationParameters.Instance == null)
+            return;
```
hhyyrylainen, line 525: This should be a really loud error, arguably this should maybe prevent the game from starting up as this would result in a really broken game if this failed.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -90,6 +95,11 @@ public class OrganelleDefinition : IRegistryType
     /// </summary>
     public OrganelleFeatureTag[] FeatureTags = Array.Empty<OrganelleFeatureTag>();
 
+    /// <summary>
+    ///   Description of processes the organelle performs
+    /// </summary>
+    public string? ProcessesDescription = null;
```
hhyyrylainen, line 101: The default value here being null is unnecessary. Also processes description is more important for an organelle than the description in my opinion. But both of these should be checked in the Check method that they aren't null and aren't empty.

src/microbe_stage/editor/tooltips/SelectionMenuToolTip.cs
```diff
@@ -176,7 +178,6 @@ public override void _Ready()
         requiresNucleusLabel = GetNode<Label>(RequiresNucleusPath);
         descriptionLabel = GetNode<CustomRichTextLabel>(DescriptionLabelPath);
         processesDescriptionLabel = GetNode<CustomRichTextLabel>(ProcessesDescriptionLabelPath);
-        modifierInfoList = GetNode<VBoxContainer>(ModifierListPath);
```
hhyyrylainen, line 179: Shouldn't `ModifierListPath` be deleted now that it is unused?

src/microbe_stage/editor/tooltips/SelectionMenuToolTip.cs
```diff
@@ -213,12 +214,19 @@ public override void _ExitTree()
     /// <summary>
     ///   Instances the UI element for a modifier info
     /// </summary>
-    public void AddModifierInfo(string name, float value)
+    public void AddModifierInfo(string name, string value, float valueColor = 0, string? iconPath = null)
     {
+        // For stuff that loads too early
+        if (modifierInfoScene == null)
+            modifierInfoScene = GD.Load<PackedScene>("res://src/microbe_stage/editor/tooltips/ModifierInfoLabel.tscn");
```
hhyyrylainen, line 221: This shouldn't be like this. I'd say it'd be better to initialize this field in the constructor to ensure that this method works before adding nodes of this type to the scene tree.

src/microbe_stage/OrganelleDefinition.cs
CI09: What about unimplemented organelles that don't have processes description?

src/microbe_stage/editor/tooltips/SelectionMenuToolTip.cs
CI09: I encountered an error where modifierInfoScene was null for no clear reason to me. So I added this to ensure it's not

src/microbe_stage/OrganelleDefinition.cs
hhyyrylainen: I guess that's a fine exception to add to the code, but otherwise the condition should be enforced.

src/microbe_stage/editor/tooltips/SelectionMenuToolTip.cs
hhyyrylainen: Before you add a node to the scene tree, _Ready *will not be called* so if you call this method and *then* add the tooltip to the scene tree, this field will be uninitialized. But instead if the field is initialized in the constructor the field *cannot* be uninitialized when calling C# method irrespective whether the node is in the scene tree or not. I've written some node types that cannot be fully used before they are added to the scene tree. But I don't think such a requirement needs to be added here as a constructor early loading the required packed scene is probably sufficient.

CI09: Oops appearently it was relevant to the unused Path variable

hhyyrylainen: The new direct node references are initialized earlier than when the old NodePath based lookup of child nodes happened. Could you maybe convert the entire selection menu class to use the direct references as you converted one of them and it wouldn't be that much more work to convert the rest (as there is an open issue about switching everything to use the new approach that is less error-prone).

locale/af.po
```diff
@@ -1335,15 +1335,9 @@ msgstr ""
 msgid "DIFFICULTY_PRESET_NORMAL"
 msgstr ""
 
-msgid "DIGESTION_EFFICIENCY"
```
hhyyrylainen, line 1338: Seems like some stuff is not handled by the new tooltips yet.

CI09: Isn't it added back some lines later?

hhyyrylainen: Lines that start like:
```
#~ msgid "DIGESTION_EFFICIENCY"
#~ msgstr "Digestion Efficiency"
```

Are *comments*, so the translation has been removed. But gettext keeps old translations like that as comments so if in the future the key is added back, it will restore the translation. But as things are it is currently in deleted state.

locale/en.po
```diff
@@ -5336,9 +5333,6 @@ msgstr "Patch Map"
 msgid "THRIVE_LICENSES"
 msgstr "Thrive Licenses"
 
-msgid "THYLAKOID"
```
hhyyrylainen, line 5339: Why is thylakoid now gone? And it doesn't seem like anything replaced this.

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -30,6 +30,16 @@ public class OrganelleDefinition : IRegistryType
...
 
+    /// <summary>
+    ///   Description of processes the organelle performs
+    /// </summary>
+    public string? ProcessesDescription;
+
+    /// <summary>
+    ///   Description of the organelle
+    /// </summary>
+    public string Description = null!;
```
hhyyrylainen, line 41: The check method should be updated to ensure there must be a description for an organelle.

src/microbe_stage/editor/tooltips/SelectionMenuToolTip.cs
```diff
@@ -213,12 +192,15 @@ public override void _ExitTree()
     /// <summary>
     ///   Instances the UI element for a modifier info
     /// </summary>
-    public void AddModifierInfo(string name, float value)
+    public void AddModifierInfo(string name, string value, float valueColor = 0, string? iconPath = null)
```
hhyyrylainen, line 195: I think `valueColor` is too unclear of a name, and doesn't follow our naming convention (should use "colour"). I suggest naming it something like `valueForColourApplying` to better describe what it does.

src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -516,4 +519,54 @@ private void HideAllToolTips()
...
+                    "+" + (Constants.ENGULF_BASE_COMPOUND_ABSORPTION_YIELD * 100 *
+                        Constants.ENZYME_DIGESTION_EFFICIENCY_BUFF_FRACTION).ToString(CultureInfo.CurrentCulture) + "%",
+                    0,
+                    "res://assets/textures/gui/bevel/DigestionIcon.png");
+            }
+
+            if (organelle.Components.Movement != null)
+            {
+                tooltip.AddModifierInfo(new LocalizedString("SPEED").ToString(),
+                    "+0.7", 0,
```
hhyyrylainen, line 567: The value should be in Constants so that it can be easily found and it should be documented that it needs to be dynamically calculated in the future (and a new issue should be opened and linked to in the comment about making it be properly calculated).

src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -516,4 +519,54 @@ private void HideAllToolTips()
...
+                tooltip.AddModifierInfo(new LocalizedString("STORAGE").ToString(),
+                    "+" + organelle.Components.Storage.Capacity.ToString(CultureInfo.CurrentCulture), 0,
+                    "res://assets/textures/gui/bevel/StorageIcon.png");
+            }
+
+            if (organelle.Components.Lysosome != null)
+            {
+                tooltip.AddModifierInfo(new LocalizedString("DIGESTION_SPEED").ToString(),
+                    "+" + (Constants.ENGULF_COMPOUND_ABSORBING_PER_SECOND *
+                        Constants.ENZYME_DIGESTION_SPEED_UP_FRACTION).ToString(CultureInfo.CurrentCulture), 0,
```
hhyyrylainen, line 554: Using current culture text formatting here, means that this data has to be re-generated when the selected game language changes.

src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -516,4 +519,54 @@ private void HideAllToolTips()
...
+            tooltip.DisplayName = organelle.Name;
+            tooltip.RequiresNucleus = organelle.RequiresNucleus;
+            tooltip.ThriveopediaPageName = organelle.InternalName;
+            tooltip.ProcessesDescription = organelle.ProcessesDescription == null ?
+                string.Empty :
+                organelle.ProcessesDescription;
+
+            if (organelle.Components.Storage != null)
+            {
+                tooltip.AddModifierInfo(new LocalizedString("STORAGE").ToString(),
```
hhyyrylainen, line 545: I'll note that thanks to this, this other issue will be solved by this PR: https://github.com/Revolutionary-Games/Thrive/issues/2306

locale/en.po
CI09: It's been replaced by chromatophore I think

src/microbe_stage/OrganelleDefinition.cs
CI09: I wanted to add that for Implemented organelles but myofibril and axon doesn't have process description

locale/en.po
hhyyrylainen: Then that's wrong because the thing in the game is called a thylakoid, meaning that chromatophore is an old name that should be scrubbed from everywhere. You don't have to do that in this PR (and I don't think it fits logically here), but you should absolutely make sure that the new tooltips use the proper name of the translation key.

CI09: Oh it actually didn't get replaced, it just is in prular (THYLAKOIDS) instead of singular for some reason

src/microbe_stage/OrganelleDefinition.cs
hhyyrylainen: I'm just asking for the description to be enforced, not the processes description. If some organelles just define the processes description then those should probably be defined to specify the description instead of the processes description. This is good to find these kinds of inconsistencies and fix them with a unified system.

CI09: My bad

src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -516,4 +542,98 @@ private void HideAllToolTips()
...
+
+    private void UpdateModifierInfoWithTranslations(OrganelleDefinition organelle, ICustomToolTip tooltip)
+    {
+        if (tooltip is not SelectionMenuToolTip selectionMenuTooltip)
+            return;
+
+        var modifierInfo = selectionMenuTooltip.GetModifierInfo("storage");
+
+        if (modifierInfo == null)
+            return;
```
hhyyrylainen, line 608: Doesn't this cause this to stop working if an earlier stat is missing (due to this return)?

I'll fix myself all the problems I see with this PR to get this done as reviewing this is starting to take more time than it would have taken me to do this from scratch.

hhyyrylainen (approved):
On testing I noticed this changed quite many things about the tooltips (including somehow breaking a word wrap on the chemosynthesizing proteins which I fixed). So I went and fixed all of those, added some TODOs, and improved the code a bit (including fixing a method that was called now quite a bit that allocated new lambda closures each time).

PR #5599, opened 2024-10-16
Make membranes more opaque

dligr:

- Decreases transparency (especially visible on lighter patches and for bigger microbes)
- <s>Improves UV mapping for bump maps. Specifically, this removes bump map stretching and 
instead makes them repeat</s> (already included in a separate commit)

Before/after:
image

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/MembraneShapeGenerator.cs
hhyyrylainen, line 1: This doubles the size of the UV data for all membranes (which there can be hundreds). Would it be possible to calculate the UV2 value in the vertex shader as it looks like it is just the vertex coordinates multiplied by a constant? I think it is possible to access the vertex position in the vertex shader in godot and write it to the UV2 variable so that it is then available in the fragment shader. This way the GPU can just calculate it (which is probably even faster due to data locality) just before running the pixel shaders.

dligr: Since mesh's vertices are moved real-time, this approach causes normals to 'drift' along the membrane surface.

I think that once new membrane textures are made, we would be able to use the same UVs for both texture and normals, so we wouldn't need two arrays.

hhyyrylainen: But hang on couldn't the animation which is calculated in the shader first copy the original data to the UV2 channel and *then* apply the animation?

shaders/Membrane.gdshader
```diff
@@ -21,6 +21,8 @@ uniform vec4 tint : source_color = vec4(1, 1, 1, 1);
 void vertex(){
     vec3 worldVertex = (MODEL_MATRIX * vec4(VERTEX, 1.0)).xyz;
     float size = length(VERTEX);
+	
+	UV2 = VERTEX.xz * 0.05f;
```
hhyyrylainen, line 25: There's a tab here

hhyyrylainen (changes requested):
Is the visuals still much more opaque in the center of cells? Also one tiny syntax style thing I noticed.

hhyyrylainen:
I assume it is fine if I make a separate simple commit from fixing the UVs for the bump map and merge that in and leave the opaqueness change for later?

hhyyrylainen:
So this is now just a PR increasing the opacity of the membranes. In order for this to be merged, next this needs positive feedback about the graphics change (so far there's been some disagreement that this is a good change).

Nein99:
I definitely support this change, as I've been playing with it a bit and it not only increases visibility of small cells on light backgrounds, but in my opinion it significantly improves the "under a microscope" look that I presume we want for the microbe stage.

Here's a screenshot I took of gameplay without this change. Bacteria are almost invisible in screenshots on light backgrounds, and are mainly only visible in game because they're moving around.
image

Here's some screenshots I took of bacteria while playing with this change (merged into my fork, so ignore the bioluminescence):
image
image
image
image

The extra opacity makes textures more visible, which gives more of an impression of a living organism as opposed to organelles wrapped in a bubble. When these bacteria were moving around on my screen, it felt significantly more like they were real 3D micro-organisms instead of abstract 2D representations. I think that the middle of the membrane being more opaque goes a long way towards making it feel like the membrane wraps all the way around the organism, and not just a circle containing your organism.

hhyyrylainen (approved):

Issue #4917, opened 2024-02-25
Test if compound cloud simulation can be written in C++ for more performance


hhyyrylainen:
Should also add a compound cloud performance benchmark to be able to get the numbers related to that performance so that a new approach can be compared to the old one.

This at least has a soft requirement for upgrade to Godot 4 as this should be much better to do as a gdextension.

hhyyrylainen:
In latest code profiling the compound cloud simulation is almost the last significant time using part of the C# code.

hhyyrylainen:
We got 100% more cloud performance in 1.1 so we can probably consider it good enough now and not need to keep this issue open forever.

PR #5606, opened 2024-10-18
Early C++ implementation for the CompoundClouds' system

AverageNerdz:
This PR introduces a basic implementation of the CompoundClouds system in C++. While it lays the groundwork for future development, there are several known issues and limitations that need to be addressed. Once these issues are resolved, this PR will close [Issue #4917](https://github.com/Revolutionary-Games/Thrive/issues/4917).

### Related Issues:
- **Color Overload Inside the Clouds**: The cloud rendering exhibits excessive color variation leading to unrealistic appearances across the compounds even when they shouldn't be there.
- **Basic Absorption System**: The current absorption mechanics are rudimentary and require further refinement.
- **Overgeneration of Textures**: The texture generation process is inefficient leading to performance issues and potential visual clutter.
- **Unoptimized Code**: The implementation is not optimized, resulting in increased resource usage.
- **Memory Leaks**: There are potential memory leaks that need to be addressed to ensure stable performance.

This implementation serves as a starting point, and I look forward to refining it further with your feedback. 

### Benchmark Results
I'd also like to share benchmark results comparing the CS benchmark with the C++ implementation:

-------------------------------------------------
**Benchmark results for CloudBenchmark CS**
- Cloud spawn score: 139
- Absorber score: 150.706
- Many spawners score: 138.353
- Cloud sim multiplier before under 60 FPS: 3.2615
- Stress test spawners: 38
- Stress test average FPS: 127.221
- Stress test min FPS: 26
- Total test duration: 134.5s
- CPU: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz (used tasks: 4, native: 2, sim threads: True)
- GPU: NVIDIA TITAN Xp
- OS: Windows
-------------------------------------------------
**Benchmark results for CloudBenchmark C++**
- Cloud spawn score: 225
- Absorber score: 277.392
- Many spawners score: 280.235
- Cloud sim multiplier before under 60 FPS: 8.6923
- Stress test spawners: 127
- Stress test average FPS: 269.913
- Stress test min FPS: 168
- Total test duration: 274.8s
- CPU: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz (used tasks: 4, native: 2, sim threads: True)
- GPU: NVIDIA TITAN Xp
- OS: Windows

Here's the visual results of my implentation:


src/saving/serializers/CompoundCloudPlaneConverter.cs
```diff
@@ -1,24 +1,25 @@
...
-        return objectType == typeof(CompoundCloudPlane);
-    }
-
-    protected override bool SkipMember(string name)
-    {
-        if (name == "Mesh")
-            return true;
+        public override bool CanConvert(Type objectType)
+        {
+            // Since CompoundCloudPlane is not a C# type, we return false
```
hhyyrylainen, line 16: How is the data saved then if this converter no longer does it? Because this was the way if I recall correctly how saving and loading didn't clear out all the compound clouds.

src/extension/nodes/CompoundCloudPlane.hpp
hhyyrylainen, line 1: This can be taken care of later, but I'll note that this file doesn't follow our naming conventions: our C++ code should use the same naming conventions as our C# code (so no snake_case naming of variables or methods).

src/extension/nodes/CompoundCloudPlane.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+namespace Thrive {
+
+std::unordered_map<int64_t, CompoundCloudPlane*> CompoundCloudPlane::instances;
+int64_t CompoundCloudPlane::next_instance_id = 1;
+
+void CompoundCloudPlane::_bind_methods() {
+    using namespace godot;
+
+    ClassDB::bind_method(D_METHOD("get_native_instance"), &CompoundCloudPlane::get_native_instance);
```
hhyyrylainen, line 16: But when exposed through Godot the godot naming convention should be used. So this line should be:
```suggestion
    ClassDB::bind_method(D_METHOD("get_native_instance"), &CompoundCloudPlane::GetNativeInstance);

```

src/extension/nodes/CompoundCloudPlane.cpp
```diff
@@ -0,0 +1,479 @@
...
+    compounds.set(2, cloud3);
+    compounds.set(3, cloud4);
+
+    set_process_priority(render_priority);
+
+    resolution = 84;
+    size = 104;
+
+    if (size <= 0) {
+        godot::UtilityFunctions::printerr("Invalid size for CompoundCloudPlane");
```
hhyyrylainen, line 110: The helper error macro should be used instead as that will automatically include the method name and line number the error happens so that's much more helpful way to print errors.

src/extension/nodes/CompoundCloudPlane.cpp
```diff
@@ -0,0 +1,479 @@
...
+    }
+
+    if (index == -1)
+        return;
+
+    std::lock_guard<std::mutex> lock(density_mutex);
+    density[x][y][index] += new_density;
+    
+    texture_needs_update = true;
+    set_process(true);
```
hhyyrylainen, line 237: Have you tested the performance impact of constantly calling the enable processing? Most clouds are basically never empty meaning that I think the overhead from enabling / disabling processing is probably more than what is gained by disabling the processing. I think this probably artificially inflates the benchmark numbers. If required I can modify the benchmark to use all compound cloud types in sequence to ensure fairness against the C# implementation that way.

src/extension/nodes/CompoundCloudPlane.cpp
```diff
@@ -0,0 +1,479 @@
...
+    return compounds;
+}
+
+int CompoundCloudPlane::get_compounds_count() const {
+    return static_cast<int>(compounds.size());
+}
+
+void CompoundCloudPlane::update_compound_colors() {
+    // Define a map of compound IDs to their corresponding colors from compounds.json
+    std::unordered_map<int, godot::Color> compound_color_map = {
```
hhyyrylainen, line 455: Is this temporary code or? If this is not temporary this definitely needs to be sent by the C# side to configure the cloud correctly.

src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -1,210 +1,182 @@
...
+                elapsed = 0;
+                foreach (var cloudInstance in cloudInstances)
+                {
+                    cloudInstance.Call("update_texture");
+                }
+            }
         }
+        catch (Exception e)
+        {
+            GD.PrintErr($"Error in CompoundCloudSystem._Process: {e.Message}\n{e.StackTrace}");
```
hhyyrylainen, line 112: I'm pretty sure there's no way anything here would cause an exception... because Godot Call will just print an error and not throw anything.

src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -1,210 +1,182 @@
...
+        }
+    }
+
+    private int GetCompoundID(CompoundDefinition compoundDefinition)
+    {
+        return (int)compoundDefinition.ID;
```
hhyyrylainen, line 118: A lot of memory I think could be saved by using the underlying type of the ID, which I specifically chose to be uint16 as that gives still plenty of compound types but requires only half as much memory as an int.

src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -1,210 +1,182 @@
...
     }
 
-    /// <summary>
-    ///   Takes compound at world position. This doesn't use locks or interlocked read so this is not thread safe
-    ///   unlike <see cref="AbsorbCompounds"/>, which is basically what should be used instead.
-    /// </summary>
-    /// <param name="compound">The compound type to take</param>
-    /// <param name="worldPosition">World position to take from</param>
-    /// <param name="fraction">The fraction of compound to take. Should be &lt;= 1</param>
     public float TakeCompound(Compound compound, Vector3 worldPosition, float fraction)
```
hhyyrylainen, line 149: I think this method needs to be re-thought out as this is called on many points for a circle for a single absorb operation, meaning that there's a ton of calls going between C# and C++ when absorbing. Also why did you remove the first part of the summary? And also you seemed to add a bunch of locks in the C++ code so it is less parallel than this C# code. You should instead consider trying to use atomic reads and writes in C++.

src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -213,134 +185,167 @@ public float AmountAvailable(Compound compound, Vector3 worldPosition, float fra
...
+        int compoundID = GetCompoundID(compound);
 
-        foreach (var cloud in clouds)
+        foreach (var cloudInstance in cloudInstances)
         {
-            // Skip clouds that don't handle the target compound
-            if (!cloud.HandlesCompound(compound))
+            var compounds = (int[])cloudInstance.Call("get_compounds");
+
+            if (!Array.Exists(compounds, id => id == compoundID))
```
hhyyrylainen, line 309: This allocates a bit of memory, would be really nice if a core operation like this didn't need to constantly allocate memory.

src/microbe_stage/CompoundCloudSystem.cs
```diff
@@ -213,134 +185,167 @@ public float AmountAvailable(Compound compound, Vector3 worldPosition, float fra
...
+                    float distanceSquared = dx * dx + dy * dy;
 
-                    // Negative coordinates are always outside the cloud area
-                    if (x < 0 || y < 0)
+                    if (distanceSquared > localRadiusSquared)
                         continue;
 
-                    // Then just need to check that it is within the cloud simulation array
-                    if (x < cloud.Size && y < cloud.Size)
+                    float amount = (float)cloudInstance.Call("amount_available", compoundID, x, y, 1.0f);
```
hhyyrylainen, line 338: This is in a very big loop calling a ton of methods through the Godot calling interface.

src/microbe_stage/CompoundCloudSystem.cs
hhyyrylainen, line 1: It looks like to me that you have done a ton of unnecessary changes to this file that make things a lot harder (so unrelated to just having the cloud types now be implemented in C++) and including removing documentation on methods. At the same time it looks like the methods I really think should be entirely rewritten for the new approach to reduce the C#<->C++ call count, are not entirely rewritten. So I sadly need to say that I think you've kind of done the opposite here than what I think should have been done.

hhyyrylainen (changes requested):
In general I think the goal with the conversion of the clouds to C++ would be that the visuals or functionality don't change in any significant way, which from your screenshots doesn't seem to be the case, but only performance would increase... Also out of the code architecture changes you've done many seem not necessary (and I didn't find out why you needed to add a global list of existing cloud planes) and when thinking about this conversion many architecture changes I anticipated needing to be done seem to not have been done (like converting the absorb / emit compound methods to an approach that reduces the number of calls between C++ and C#). So I have quite a lot of architecture feedback, some of which I specifically pointed out but there's so many instances that I couldn't comment on all of them.

PR #5611, opened 2024-10-20
Fix flagellum daughter incorrect size

CI09:

This PR does some stuff...

**Related Issues**

https://discord.com/channels/228300288023461893/958598553389903913/1297495322007961693

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/PlacedOrganelle.cs
+++ b/src/microbe_stage/PlacedOrganelle.cs
@@ -357,8 +357,18 @@ public class PlacedOrganelle : IPositionedOrganelle, ICloneable
     {
         var scale = CalculateTransformScale();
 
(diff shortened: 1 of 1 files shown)


dligr (approved):
The code added here is the same as in `CalculateVisualsTransformExternal`, so it should work (I also tested this in game).

Maybe, given that the code for flagellum scaling is used multiple times now, it would be better to incorporate that code into `CalculateTransformScale` (and make that method output a Vector3 instead of a float), to reduce code duplication.

hhyyrylainen (approved):
I quickly checked in game and it seemed that this worked fine when flagella were growing. I fixed the one code problem I saw here.

PR #5616, opened 2024-10-21
Adjust photosynthesis to have same glucose output ratio as before

hhyyrylainen:
the dynamic compounds experimental PR that multiplied the carbon dioxide requirement by a ton but didn't modify the glucose output so this should correct photosynthesis to be a viable glucose source again

**Related Issues**

reports after the RC about photosynthesis output problems

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/microbe_stage/bio_processes.json
+++ b/simulation_parameters/microbe_stage/bio_processes.json
@@ -35,11 +35,11 @@
     "Name": "PHOTOSYNTHESIS",
     "Inputs": {
       "sunlight": 1,
-      "carbondioxide": 3
+      "carbondioxide": 0.27
     },
     "Outputs": {
-      "glucose": 0.04,
-      "oxygen": 3
+      "glucose": 0.12,
+      "oxygen": 0.27
     }
   },
   "cytotoxinSynthesis": {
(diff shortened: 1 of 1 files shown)


simulation_parameters/microbe_stage/bio_processes.json
```diff
@@ -214,11 +214,11 @@
...
     "Inputs": {
       "sunlight": 1,
-      "carbondioxide": 1
+      "carbondioxide": 0.081
     },
     "Outputs": {
-      "glucose": 0.012,
-      "oxygen": 1
+      "glucose": 0.036,
+      "oxygen": 0.081
```
dligr, line 221: I believe that glucose production has to be increased x4 here, as otherwise it matches thylakoids' performance

dligr (commented):
After testing this in game, I can confirm that it works. Photosynthesis is viable now, and it still affects environment despite lower input/output values. Patches' oxygen/carbondioxide seem to end up in good equilibrium values (in surface patches), allowing effective photosynthesis.

simulation_parameters/microbe_stage/bio_processes.json
hhyyrylainen: I multiplied both chloroplast process and thylakoid process output by 3x, so the ratio should be the same. The first photosynthesis process has output of: `0.12` and the thylakoid has: `0.036`. So it is still multiple times more, so isn't it fine or what numbers do you suggest using?

dligr: Glucose production depends on the ratio between carbondioxide and glucose values. So, normal photosynthesis reaches 100% performance at 27% CO2 and 0.12 glucose production.
Chromatophore photosynthesis reaches 333.(3)% performance at the same level of CO2 and 3.(3) * 0.036 = 0.12 glucose, which is the same as normal photosynthesis.

dligr (approved):

PR #5631, opened 2024-10-26
Update Nitrogen icon

HexapodPhilosopher:

Removes shading from middle of nitrogen icon.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


assets/textures/gui/bevel/Nitrogen.svg.import
```diff
@@ -2,7 +2,7 @@
 
 importer="texture"
 type="CompressedTexture2D"
-uid="uid://d0qdyfeacbf31"
+uid="uid://r5ci3tvgwuo"
```
hhyyrylainen, line 5: I think the UID change here will cause issues as this is referenced in other files. I think this PR would be better if this had only the svg file change in it.

hhyyrylainen (changes requested):

hhyyrylainen (approved):

Issue #3207, opened 2022-04-14
Process panel should show all processes in a cell colony


hhyyrylainen:
especially important for multicellular where colony members can have processes the main cell doesn't have

hhyyrylainen:
Based on a comment (https://github.com/Revolutionary-Games/Thrive/pull/5820#discussion_r1934333642) I think we need to sum all of the non-ATP producing processes in the cell colony and show in the process panel (potentially dividing by the cell count to show the average values like happens in a colony). ATP-producing processes should probably be shown as-is as the ATP is not shared between colony members.

hhyyrylainen:
Related comment: when a new cell grows in a colony it shouldn't reset the process panel state.

PR #5632, opened 2024-10-27
Fix processes' toggle buttons not matching their actual state

### LINKED ISSUE #5420: Process toggle buttons sometimes don't match process state
I was able to find two cases when this happens:
1. When exiting editor with a new organelle, the process speed of which is immediately zero (because of having no needed compound or something like that). For example, placing rusticyanin while having no iron or thylakoids in a patch with no light. The process is shown as turned off. After gathering needed resources, however, the process starts to run (the button still shows it as turned off).
2. If a process was manually turned off, it starts to run again after an editor session. However, it is still shown as turned off.

Picture after exiting editor and getting resources required for processes to run:
image
Some processes run, even though they are turned off.

dligr:

Fixes process toggle buttons being de-synced from their processes' actual state.

Does so by making process displays get the toggle state directly from the process, instead of using a local variable. This also allows those states to be saved.

Also, makes TweakedProcesses' speed multipliers save when they are updated (e.g. when an organelle duplicates).

**Related Issues**

Fixes #5420

**Progress Checklist**

      break existing features:
      members reviewing this PR)


dligr:
I have noticed while testing that process toggles don't work on cell colonies (at least, on other members of those colonies). I hope that it's fine if I fixed it too with this PR.

hhyyrylainen:
It's most likely fine, but I have slight concern if that has understandability problems because https://github.com/Revolutionary-Games/Thrive/issues/3207 is not done. But I guess in most situations where like for efficiency reasons players want to toggle processes it's fine as the colony compound distribution will make sure that other colony members can basically do the same processes as the lead cell.

src/gui_common/ChemicalEquation.cs
```diff
@@ -102,14 +100,19 @@ public bool ShowToggle
...
+            if (EquationFromProcess == null)
+                return true;
+
+            return EquationFromProcess.Enabled;
+        }
         set
         {
-            if (value == lastToggle)
+            if (EquationFromProcess == null)
                 return;
```
hhyyrylainen, line 113: I think this should print an error message in case there's a programmer error in the future (so that it'll be quicker to diagnose).

src/microbe_stage/TweakedProcess.cs
```diff
@@ -4,13 +4,7 @@
...
 ///   A concrete process that organelle does. Applies a modifier to the process
 /// </summary>
-/// <remarks>
-///   <para>
-///     This is a struct as this just packs a few floats and a single object reference in here. This allows much tighter
-///     data packing when this is used in lists.
-///   </para>
-/// </remarks>
-public struct TweakedProcess : IEquatable<TweakedProcess>
+public class TweakedProcess : IEquatable<TweakedProcess>
```
hhyyrylainen, line 7: This is a pretty major change to no longer have this as a struct, and will now have temporary memory allocation consequences.

src/microbe_stage/StrictProcessDisplayInfoEquality.cs
```diff
@@ -33,6 +33,9 @@ public bool Equals(StrictProcessDisplayInfoEquality? other)
         var our = DisplayInfo;
         var theirs = other.DisplayInfo;
 
+        if (our.Enabled != theirs.Enabled)
```
hhyyrylainen, line 36: The whole process equality comparisons is a bit of a mess and changing any of the code here can cause bugs in a bunch of places. I think the enabled state probably should not be compared here as having a process speed display show the same item twice but another is disabled while the other is enabled doesn't seem like a good idea to me.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -61,6 +61,13 @@ public static void ComputeActiveProcessList(IReadOnlyList<IPositionedOrganelle>
     {
         result ??= new List<TweakedProcess>();
 
+        var speedMultipliers = new Dictionary<BioProcess, float>();
```
hhyyrylainen, line 64: This code should be rewritten to not use this extra temporary memory.

src/microbe_stage/ProcessStatistics.cs
```diff
@@ -412,6 +418,10 @@ public override bool Equals(object? obj)
         // This also checks for obj being null
         if (obj is AverageProcessStatistics statistics)
         {
+            GD.Print(owner.Equals(statistics.owner));
```
hhyyrylainen, line 421: This looks like a debug print.

src/microbe_stage/ProcessStatistics.cs
```diff
@@ -412,6 +418,10 @@ public override bool Equals(object? obj)
...
+            if (owner.Process.SpeedMultiplier != statistics.owner.Process.SpeedMultiplier)
+                GD.Print("Speed multiplier mismatch!");
```
hhyyrylainen, line 423: Is this an error condition? (also the message should be a bit more specific because if we get logs with this message in them, that will really confuse whoever is trying to determine the problem).

src/microbe_stage/ProcessStatistics.cs
```diff
@@ -312,6 +309,7 @@ public override bool Equals(object? obj)
         // This also checks for obj being null
         if (obj is SingleProcessStatistics statistics)
         {
+            GD.Print(Process.Equals(statistics.Process));
```
hhyyrylainen, line 312: Another debug print.

src/microbe_stage/ProcessStatistics.cs
```diff
@@ -322,6 +320,10 @@ public bool Equals(SingleProcessStatistics? other)
     {
         if (ReferenceEquals(null, other))
             return false;
+
+        if (Process.SpeedMultiplier != other.Process.SpeedMultiplier)
+            GD.Print("Speed multiplier mismatch!");
```
hhyyrylainen, line 325: This seems like a debug print?

src/microbe_stage/ProcessStatistics.cs
```diff
@@ -15,24 +15,19 @@ public class ProcessStatistics
...
     /// </summary>
-    private List<BioProcess>? temporaryRemovedItems;
+    private List<TweakedProcess>? temporaryRemovedItems;
 
     /// <summary>
     ///   The processes and their associated speed statistics
     /// </summary>
     /// <remarks>
     ///   <para>
-    ///     This uses <see cref="BioProcess"/> rather than <see cref="TweakedProcess"/> as the key so that equality
```
hhyyrylainen, line 25: This is another point that I think is causing some pretty major architectural problems here. I didn't yet check every single place, that these changes impact, but my impression is that all of the code depending on the assumptions about the code that has been changed in this PR have not been updated yet.

hhyyrylainen (changes requested):
I think this is a pretty good design for this problem and one that can fix the toggle states getting out of sync. What I have significant concerns about is the change of that one struct to a class which majorly changes all memory usage and allocations related to it, which is something I've relied on greatly in the process system for maximum efficiency. I think as a result the changes here need to be considered more.

src/microbe_stage/TweakedProcess.cs
dligr: Apparently, because of how process updating works, this change was unnecessary

hhyyrylainen: That's actually very good to hear, so this PR might be able to be polished up easily, then.

src/microbe_stage/ProcessStatistics.cs
```diff
@@ -71,13 +66,8 @@ public void RemoveUnused()
         }
     }
 
-    public SingleProcessStatistics GetAndMarkUsed(BioProcess forProcess)
+    public SingleProcessStatistics GetAndMarkUsed(TweakedProcess forProcess)
     {
-#if DEBUG
```
hhyyrylainen, line 76: I think this debug code could be updated to check if the process inside the tweaked process is valid. That would keep this safety check (that only is in debug mode, so active for only programmers) in-place.

src/microbe_stage/TweakedProcess.cs
```diff
@@ -50,7 +50,8 @@ public bool Equals(TweakedProcess other)
             return false;
 
         // ReSharper disable once CompareOfFloatsByEqualityOperator
-        return Rate == other.Rate && ReferenceEquals(Process, other.Process);
+        return Rate == other.Rate && SpeedMultiplier == other.SpeedMultiplier
```
hhyyrylainen, line 53: Adding comparing the speed multiplier here, I think might break some grouping of process displays...

So I would remove this change.

dligr: I don't think it's right, because if processes with different states are considered equal, then they aren't correctly updated on GUI, unless I'm missing something.

hhyyrylainen: Have you playtested and confirmed that is a bug? Kind of unhelpfully the process list detection uses equality to combine items, so too strict checking can result in a situation where duplicates get created into the process list.

hhyyrylainen: It's true that if this isn't the type of equality used in the GUI combining (and that doesn't call this equality in turn), this should be fine. Though, the last time I tried to improve the process equality checks to be more accurate, I totally messed up the process speed panel to show each instance of a process separately (instead of keeping the functionality where all organelles are summed into one process of each type). I'll need to check this carefully when doing the next review (after the extra temporary memory use is solved).

src/microbe_stage/systems/ProcessSystem.cs
dligr: Should the Dictionary be stored outside of the method's domain?

hhyyrylainen: Yes, if there's truly no way to write the logic in a way that doesn't require this store first and then apply approach.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -32,6 +32,8 @@ namespace Systems;
 [RuntimeCost(55)]
 public sealed class ProcessSystem : AEntitySetSystem<float>
 {
+    private static readonly Dictionary<BioProcess, float> TempSpeedMultipliers = new();
```
hhyyrylainen, line 35: I'm pretty sure this is a massive problem as ComputeActiveProcessList has to be thread safe.

dligr: Would it be right to put all of `ComputeActiveProcessList`'s code in a `lock` statement? Only a single thread would then be able to access the dictionary at a time, though that would worsen performance

hhyyrylainen: I wouldn't because this then becomes a bottleneck. If there truly is no way to avoid the temp memory, then the approach where the method caller must provide the working memory should be used (look at for example how the hex adding methods in the hex layout class require multiple working memory parameters to work).

dligr: This is done now. Each microbe spawn now requires an extra memory allocation, though.

src/microbe_stage/systems/EndosymbiontOrganelleSystem.cs
```diff
@@ -74,7 +75,7 @@ protected override void Update(float state, in Entity entity)
                     // times and when not empty mostly just once
                     organelleContainer.OnOrganellesChanged(ref entity.Get<CompoundStorage>(),
                         ref entity.Get<BioProcesses>(), ref entity.Get<Engulfer>(), ref entity.Get<Engulfable>(),
-                        ref entity.Get<CellProperties>());
+                        ref entity.Get<CellProperties>(), tempProcessListMemory);
```
hhyyrylainen, line 78: The list must be locked as this can run multithreaded. See for example line 100 how the hexWorkData has to be locked.

src/microbe_stage/systems/MicrobeReproductionSystem.cs
```diff
@@ -623,7 +624,7 @@ private void ReadyToReproduce(in Entity entity, ref OrganelleContainer organelle
                     // Return the first cell to its normal, non duplicated cell arrangement and spawn a daughter cell
                     organelles.ResetOrganelleLayout(ref entity.Get<CompoundStorage>(),
                         ref entity.Get<BioProcesses>(),
-                        entity, species, species, worldSimulation, workData1, workData2);
+                        entity, species, species, worldSimulation, workData1, workData2, tempProcessListMemory);
```
hhyyrylainen, line 627: This also has to lock the data.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -57,10 +57,15 @@ public ProcessSystem(World world, IParallelRunner runner) : base(world, runner,
...
     /// </remarks>
     public static void ComputeActiveProcessList(IReadOnlyList<IPositionedOrganelle> organelles,
-        [NotNull] ref List<TweakedProcess>? result)
+        [NotNull] ref List<TweakedProcess>? result, Dictionary<BioProcess, float> tempStorage)
     {
         result ??= new List<TweakedProcess>();
 
+        foreach (var process in result)
+        {
+            tempStorage.TryAdd(process.Process, process.SpeedMultiplier);
```
hhyyrylainen, line 66: For the hex work data the pattern I usually use is that the data is cleared just before use, this way if an exception breaks the game and skips the data clear, that won't cause compounding errors after that.

src/general/world_effects/PhotosynthesisProductionEffect.cs
```diff
@@ -50,6 +50,8 @@ private void ApplyCompoundsAddition()
             float oxygenBalance = 0;
             float co2Balance = 0;
 
+            Dictionary<BioProcess, float> tempStorage = new();
```
hhyyrylainen, line 53: This should be allocated outside the loop.

src/microbe_stage/ProcessStatistics.cs
```diff
@@ -71,13 +66,12 @@ public void RemoveUnused()
...
 
-    public SingleProcessStatistics GetAndMarkUsed(BioProcess forProcess)
+    public SingleProcessStatistics GetAndMarkUsed(TweakedProcess forProcess)
     {
 #if DEBUG
-        if (forProcess == null!)
+        if (forProcess.Process == null!)
             throw new ArgumentException("Invalid process marked used");
 #endif
-
```
hhyyrylainen, line 80: I think the blank line after a pragma looks much better than code being all bunched up.

src/microbe_stage/ProcessStatistics.cs
```diff
@@ -154,6 +148,8 @@ public float CurrentSpeed
         }
     }
 
+    public bool Enabled => Process.SpeedMultiplier > 0.5f;
```
hhyyrylainen, line 151: Wouldn't it be better to show it enabled if the value is above 0? I think somewhere else there's also a similar threshold so unless they both use zero, I'd put the magic value in Constants to make sure a consistent value is used.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -57,10 +57,15 @@ public ProcessSystem(World world, IParallelRunner runner) : base(world, runner,
     ///   </para>
     /// </remarks>
     public static void ComputeActiveProcessList(IReadOnlyList<IPositionedOrganelle> organelles,
-        [NotNull] ref List<TweakedProcess>? result)
+        [NotNull] ref List<TweakedProcess>? result, Dictionary<BioProcess, float> tempStorage)
     {
         result ??= new List<TweakedProcess>();
 
+        foreach (var process in result)
```
hhyyrylainen, line 64: Could you add a comment here that this block of code is for remembering the existing speed multiplier states?

hhyyrylainen (changes requested):
At least the couple data locks have to be added I commented on at a minimum.

I still have this vague idea in my head that it would be better to do this without the temp memory.

Here's an idea I just had: what if TweakedProcess gets added an `internal bool Marked` field and then the ComputeActiveProcessList method uses that to detect what is a new entry and what is an old one? That way the process speeds can be copied from an existing process entry to a newly created one inside the loop. And then instead of calling `result.Clear` the clearing could be done by removing the entries that didn't have the marked field set. Does that sound doable?

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -106,9 +117,25 @@ public static void ComputeActiveProcessList(IReadOnlyList<IPositionedOrganelle>
...
 
                 // If not found, then create a new result
-                result.Add(new TweakedProcess(process.Process, process.Rate));
+                var newProcess = new TweakedProcess(processKey, process.Rate);
+                newProcess.Marked = true;
+                result.Add(newProcess);
             }
         }
+
+        int writeIndex = 0;
```
hhyyrylainen, line 126: My last comment. This block of code could do with a comment saying this removes the unmarked processes to not keep processes around that don't exist anymore, and also for efficiency this handles unmarking processes that are kept.

hhyyrylainen (approved):
I think the code is perfect now. I'll do some final testing tomorrow, unless someone else gets to it first, before merging this.

dligr:
Apparently, the process button used the `toggled()` signal (Godot-side), which is called when button's state is changed, even if it's changed from code. Which led to the unexpected behavior of the button toggling itself.

src/microbe_stage/TweakedProcess.cs
```diff
@@ -50,7 +60,8 @@ public bool Equals(TweakedProcess other)
...
+            && ReferenceEquals(Process, other.Process);
```
hhyyrylainen, line 64: I think I was able to confirm that this change breaks the ProcessStatistics.GetAndMarkUsed so I will revert this change. (this now causes process panel process order to change when turning processes off and on)

hhyyrylainen (approved):
I did a fix for the process equality bug I found (it caused items in the process panel to jump around as I pressed the toggle buttons). Now everything should be working fine, though I'd greatly appreciate if you could double check my changes. I'm unsure if the logic for ordering the items back into the order they should be by moving the marked items to the start of the process data list is fully working, so especially double checking this bit of code would be good for avoiding bugs:
```c#
for (int l = 0; l < k; ++l)
{
    if (!result[l].Marked)
    {
        // Swap positions of the data, as we will write to the k index (that is updated)
        // we need to only write the moving away data to perform the swap
        result[k] = result[l];
        k = l;
        break;
    }
}
```

dligr:
After some testing, I can confirm that the new changes work. Processes are now kept in a consistent order, and their toggling now changes their speed gradually. The changes to the ProcessSystem's process updating make sense to me too.

PR #5643, opened 2024-10-30
Fixed moving flagellum having default size

### LINKED ISSUE #5640: Moving an organelle with upgrades doesn't correctly display the upgrade visuals on the in-progress moved hex
instead the visuals apply correctly again only after placing the organelle

I'm tagging this as easy as I think this will be just a case of reviewing the hover organelle display code and making sure it uses the same approach as the real visuals code for taking upgrades into account.

CI09:

This PR does some stuff...

**Related Issues**

Closes #5640

as long as falgellum is the only case of custom scaling

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/OrganelleDefinition.cs
+++ b/src/microbe_stage/OrganelleDefinition.cs
@@ -336,6 +336,22 @@ public class OrganelleDefinition : IRegistryType
         return true;
     }
 
+    public Vector3 GetUpgradesSizeModification(OrganelleUpgrades? upgrades)
+    {
+        var scale = Constants.DEFAULT_HEX_SIZE;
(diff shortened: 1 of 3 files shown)


hhyyrylainen (changes requested):
I think this is the wrong approach as `UpdateAlreadyPlacedVisuals` already has logic for handling the special cases, so I think the highligthed organelle method here should rather use `UpdateOrganelleDisplayerTransform` than reimplement the logic.

hhyyrylainen (approved):
This now looks nice with the new helper method. It is probably in a sensible enough place. And I verified in-game that all seems to function correctly.

PR #5649, opened 2024-11-03
Add underwater vent eruption event

CI09:

Thanks dligr and hhyrylainen!

**Related Issues**

Roadmap "Environmental events" (partially)

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I think it would be important to also show an icon on the patch map about the event for visibility and on the report screen if the player has that patch selected. That way the player will notice the event, instead of just noticing the random compound changes and submitting a bug report.

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -33,6 +33,14 @@ public partial class PatchMapNode : MarginContainer
     [Export]
     public string UnknownTextureFilePath = null!;
 
+#pragma warning disable CA2213
+
+    // Is public to easily edit
```
hhyyrylainen, line 38: This should be made private and a method added to control the effect display.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -876,6 +876,9 @@ private void AddPatchNode(Patch patch, Vector2 position)
 
         node.Enabled = patchEnableStatusesToBeApplied?[patch] ?? true;
 
+        if (patch.ActiveDisplayVisuals.Contains(WorldEffectVisuals.UnderwaterVentEruption))
```
hhyyrylainen, line 879: I think the code here should be made extensible to allow a wide variety of events to be added in the future. It's probably okay to not support multiple events at once in a single patch.

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -147,7 +147,8 @@ protected override void InitEditor(bool fresh)
 
         reportTab.UpdateReportTabPatchSelector();
 
-        reportTab.UpdateGlucoseReduction(CurrentGame.GameWorld.WorldSettings.GlucoseDecay);
+        reportTab.UpdateEvents(CurrentGame.GameWorld.EventsLog.Where(
+            c => c.Key == CurrentGame.GameWorld.TotalPassedTime).ToDictionary().Values.ToList()[0]);
```
hhyyrylainen, line 151: We'll probably have events in the future that last multiple generations so I think the events log should be an object that can be asked here if there are any ongoing events at current editor time.

src/general/world_effects/UnderwaterVentEruptionEffect.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using Newtonsoft.Json;
+using Xoshiro.PRNG64;
+
+/// <summary>
+///   An effect reducing the glucose amount (for the microbe stage to make early game easier, and the late game harder)
```
hhyyrylainen, line 6: This seems copy-pasted, so could you double check you don't have accidentally unchanged copy-pasted stuff in your PR?

src/general/GameEventDescription.cs
```diff
@@ -24,4 +25,9 @@ public GameEventDescription(LocalizedString description, string? iconPath, bool
     ///   If true, this event will be highlighted in the timeline UI
     /// </summary>
     public bool Highlighted { get; private set; }
+
+    /// <summary>
+    ///   Some events show up in report tab
```
hhyyrylainen, line 30:
```suggestion
    ///   Some events show up in the report tab to draw extra attention to them
```

src/general/GameWorld.cs
```diff
@@ -338,8 +345,6 @@ public void ReplacePlayerSpeciesWith(Species species, bool deleteOldPlayer = tru
     /// </summary>
     public void GenerateRandomSpeciesForFreeBuild(long seed = 0)
     {
-        Random random;
```
hhyyrylainen, line 341: I think this is a bad change, the freebuild stuff here recreates the random so it is kind of weird how it is a field. I'd not make the random a field unless you consider these overall implications more for this class.

hhyyrylainen (changes requested):
I just submitted some initial feedback on this. I didn't look at everything yet but my initial impression is that the system needs to be made a bit more general so that other events can easily be added in the future.

src/microbe_stage/editor/PatchMapDrawer.cs
CI09: I can do that but are you sure it's good idea to avoid multiple events?

hhyyrylainen: I mean per patch. If there's like 3 events in a vents patch showing all of them in the patch map would make the icon super busy and the player would need to spend quite long looking at the map. So I think it probably makes better gameplay that each patch can have only one event affect it per generation.

src/microbe_stage/editor/MicrobeEditor.cs
CI09: I tried making system show eruption marker on a patch map node if there is event in events log but it didn't work. I think we can ignore possible multi-generations events for now and handle them when we actually implement these

src/microbe_stage/editor/PatchMapNode.tscn
```diff
@@ -84,5 +87,18 @@ label_settings = ExtResource("4_tygy2")
 horizontal_alignment = 1
 vertical_alignment = 1
 
+[node name="Margin" type="VBoxContainer" parent="."]
+custom_minimum_size = Vector2(16, 64)
+layout_mode = 2
+size_flags_vertical = 0
+
+[node name="Eruption" type="TextureRect" parent="Margin"]
```
hhyyrylainen, line 95: Would be pretty nice to have a tooltip explaining on hover what the icon means. Something like `EVENT_TOOLTIP_ERUPTION` with text like 
```
Eruption Patch Event
Creates extra hydrogen sulfide and carbon dioxide
```

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -269,6 +273,12 @@ public void UpdateVisibility()
         }
     }
 
+    public void ChangeMarginIconVisibility(WorldEffectVisuals type, bool value)
```
hhyyrylainen, line 276:
```suggestion
    public void UpdateEventVisuals(WorldEffectVisuals type)
```
That's still my suggestion on how the public API for managing the visual events should be for future extensibility.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -876,6 +876,12 @@ private void AddPatchNode(Patch patch, Vector2 position)
 
         node.Enabled = patchEnableStatusesToBeApplied?[patch] ?? true;
 
+        foreach (var effect in Enum.GetValues(typeof(WorldEffectVisuals)))
+        {
+            if (patch.ActiveDisplayVisuals.Contains((WorldEffectVisuals)effect))
```
hhyyrylainen, line 881: I think a more extensible code would be to just have `patch.ApplyActiveEffectVisuals(node)` and then put the more complex code into the patch class.

src/microbe_stage/editor/MicrobeEditorReportComponent.tscn
```diff
@@ -240,8 +241,12 @@ layout_mode = 2
 size_flags_horizontal = 3
 size_flags_vertical = 3
 
-[node name="GlucoseReduction" type="Label" parent="MarginContainer/HSplitContainer/PanelContainer/MarginContainer/VBoxContainer/EvoResultsPanel/MarginContainer/ScrollContainer/VBoxContainer"]
+[node name="MajorEvents" type="VBoxContainer" parent="MarginContainer/HSplitContainer/PanelContainer/MarginContainer/VBoxContainer/EvoResultsPanel/MarginContainer/ScrollContainer/VBoxContainer"]
+layout_mode = 2
+
+[node name="Event" type="Label" parent="MarginContainer/HSplitContainer/PanelContainer/MarginContainer/VBoxContainer/EvoResultsPanel/MarginContainer/ScrollContainer/VBoxContainer/MajorEvents"]
 editor_description = "PLACEHOLDER"
```
hhyyrylainen, line 248: Shouldn't the editor description be cleared here? It seems unrelated to what this node was turned into.

src/general/world_effects/WorldEffectVisuals.cs
```diff
@@ -0,0 +1,5 @@
+﻿public enum WorldEffectVisuals
+{
+    Default,
```
hhyyrylainen, line 3:
```suggestion
    None,
```

src/microbe_stage/editor/MicrobeEditorReportComponent.cs
```diff
@@ -209,13 +212,27 @@ public void UpdateTimeIndicator(double value)
...
+        }
 
-        // The amount of glucose has been reduced to {0} of the previous amount.
-        glucoseReductionLabel.Text = Localization.Translate("THE_AMOUNT_OF_GLUCOSE_HAS_BEEN_REDUCED")
-            .FormatSafe(percentage);
+        foreach (var registeredEvent in events)
+        {
+            if (registeredEvent.ShowInReport)
+            {
+                var duplicate = (Label)eventTemplate.Duplicate();
```
hhyyrylainen, line 229: Packed scenes should be used instead of duplicating nodes for instantiating things.

src/general/world_effects/UnderwaterVentEruptionEffect.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    [JsonProperty]
+    private GameWorld targetWorld;
+
+    private XoShiRo256starstar random;
+
+    public UnderwaterVentEruptionEffect(GameWorld targetWorld, XoShiRo256starstar random)
+    {
+        this.targetWorld = targetWorld;
+        this.random = random;
```
hhyyrylainen, line 19: The random should be made into a read-only property with the JsonProperty attribute as it needs to be loaded from saves. Also as random instances aren't super safe to share, usually when a method takes in a random in Thrive, then in the method body I do `random = new Xosiro...(random.Next...)` to just take in a seed for the random from the other random but then not keep it around for safety reasons.

src/microbe_stage/Patch.cs
```diff
@@ -16,6 +16,9 @@
 [UseThriveSerializer]
 public class Patch
 {
+    [JsonProperty]
+    public List<WorldEffectVisuals> ActiveDisplayVisuals;
```
hhyyrylainen, line 20: I think this should be renamed *and* made private and read-only. Then methods like add event and a method for this patch to send the active events for rendering / text display should be added. That way the internal details are nicely encapsulated and can be extended in the future and we avoid extra memory allocations (which would be necessary if a method like `GetActiveEvents` was created).

CI09: How read-only if we want to change it via methods?

hhyyrylainen: I mean `readonly` so that the list object cannot be reassigned, but its contents will still be modifiable.

src/microbe_stage/editor/PatchMapDrawer.cs
hhyyrylainen, line 1: I'm pretty sure applying things just here doesn't work for the auto-evo exploring tool. So this approach only works for the normal game. I'm currently fixing all the minor style problems still remaining so I'll open a separate issue, which I'd appreciate if you could work on it as this is your feature: https://github.com/Revolutionary-Games/Thrive/issues/5666

Oh and I noticed that the exploring tool doesn't have a textual list of the events anywhere.

src/general/world_effects/UnderwaterVentEruptionEffect.cs
```diff
@@ -0,0 +1,67 @@
...
+
+                // Percentage is density times amount, so clamp to the inversed amount (times 100)
+                currentHydrogenSulfide.Density = Math.Clamp(currentHydrogenSulfide.Density, 0, 1
+                    / currentHydrogenSulfide.Amount * 100);
+                currentCarbonDioxide.Ambient = Math.Clamp(currentCarbonDioxide.Ambient, 0, 1);
+
+                patch.Biome.ModifyLongTermCondition(Compound.Hydrogensulfide, currentHydrogenSulfide);
+                patch.Biome.ModifyLongTermCondition(Compound.Carbondioxide, currentCarbonDioxide);
+
+                patch.LogEvent(new LocalizedString("UNDERWATER_VENT_ERUPTION"),
```
hhyyrylainen, line 54: Shouldn't this be inside an else clause? I'll fix this assuming that is the case.

src/microbe_stage/Patch.cs
```diff
@@ -516,6 +526,21 @@ public void ApplyVisibility(MapElementVisibility visibility)
...
     }
 
+    public void ApplyPatchNodeVisuals(WorldEffectVisuals visual)
+    {
+        activeWorldEffectVisuals.Add(visual);
+    }
+
+    public void ClearPatchNodeVisuals()
+    {
+        activeWorldEffectVisuals = new List<WorldEffectVisuals>();
```
hhyyrylainen, line 536: Why is a new list made here instead of the previous one cleared? This looks like a big mistake in terms of code efficiency. I'll fix this, but I'm commenting here for a learning opportunity.

src/general/world_effects/UnderwaterVentEruptionEffect.cs
hhyyrylainen: It was in fact incorrect, I didn't realize the difference when quickly reading. So I ended up reverting my change to this part.

src/microbe_stage/editor/PatchMapDrawer.cs
hhyyrylainen: I made a helper method for updating the map drawer events state, so it could be as simple as calling that one method in the right place in the exploring tool.

PR #5657, opened 2024-11-05
Optimized CMake Build Process 

AverageNerdz:

This PR refines the CMake configuration for our GDExtension, optimizing build performance across platforms. It addresses specific compilation issues on Windows, bringing improvements in speed and reliability it should mantain robust performances on Linux.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/extension/CMakeLists.txt
```diff
@@ -5,54 +5,201 @@
...
-  "${PROJECT_BINARY_DIR}/Include.h"
-  core/RegisterThriveTypes.cpp core/RegisterThriveTypes.hpp
-  core/ThriveConfig.cpp core/ThriveConfig.hpp
-  interop/ExtensionInterop.cpp interop/ExtensionInterop.h
-  nodes/DebugDrawer.cpp nodes/DebugDrawer.hpp
-  core/GodotJoltConversions.hpp
+add_library(thrive_extension SHARED)
+
+# Collect all source files recursively
+file(GLOB_RECURSE EXTENSION_SOURCES
```
hhyyrylainen, line 11: Again, I really dislike GLOB because it doesn't automatically re-run cmake when new files are added.

third_party/CMakeLists.txt
```diff
@@ -6,26 +6,38 @@
...
+# Force dynamic runtime for all libraries
+set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
+
+# JoltPhysics configuration
 set(DOUBLE_PRECISION ON)
 
 if(THRIVE_AVX)
-  set(USE_AVX ON)
-  set(USE_AVX2 ON)
+    set(USE_AVX ON)
```
hhyyrylainen, line 24: Could you re-indent all the files using the existing formatting rules (two spaces per indentation level)?

third_party/windows/godot-cpp-windows.cmake
```diff
@@ -0,0 +1,219 @@
+# Windows-specific godot-cpp configuration (temporary fix)
```
hhyyrylainen, line 1: I'd like to not accept a temporary fix into the main repo as it'll then just be here forever...

CMakeLists.txt
```diff
@@ -1,10 +1,275 @@
...
+        JPH_USE_SSE4_1=1
+        JPH_USE_SSE4_2=1
+        JPH_TRACK_BROADPHASE_STATS=0
+        JPH_STATISTICS=0
+        JPH_DISABLE_CUSTOM_ALLOCATOR
+    )
+endfunction()
+
+# Source files print function
+function(print_source_files target)
```
hhyyrylainen, line 183: Isn't this like debugging code? I'd remove this.

CMakeLists.txt
```diff
@@ -1,10 +1,275 @@
 # Native code side of Thrive
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.13)
+
+set(CMAKE_CONFIGURATION_TYPES "Debug;Release;RelWithDebInfo;Distribution" CACHE STRING "" FORCE)
```
hhyyrylainen, line 4: You should copy the comment from where you moved this code (the reason why the non-standard distribution type is here).

CMakeLists.txt
```diff
@@ -1,10 +1,275 @@
...
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
```
hhyyrylainen, line 5: This should be just removed as this is default behaviour and I think it is unnecessarily complication to be at the top of the file.

CMakeLists.txt
```diff
@@ -1,10 +1,275 @@
...
+endif()
 
 project(Thrive)
 
 # If you want to get compile commands run cmake with
 # "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
+# Options
+# Building either the faster variant with AVX or without for older CPU support
+
+# CPU core optimization
```
hhyyrylainen, line 19: All of this stuff is super long, so I think it should be put into a separate file like `Scripts/CMakeTargetConfiguration.cmake` and included here.

CMakeLists.txt
```diff
@@ -1,10 +1,275 @@
...
+        /FI"cstdint"
+        /FI"atomic"
+        $<$<CONFIG:Debug>:/MDd /Od /Ob1 /RTC1>
+        $<$<NOT:$<CONFIG:Debug>>:/MD /O2 /Ob2 /DNDEBUG>
+        # Critical warning suppressions
+        /wd4005  # Macro redefinition
+        /wd4273  # Inconsistent dll linkage
+        /wd4141  # 'inline' used more than once
+        # Other warning disables
+        /wd4068  # Unknown pragma
```
hhyyrylainen, line 81: If the thrive code doesn't trigger any of these warnings it would be, much much better to not include these. In general you have made a ton of changes here, whereas I'd prefer to *only* have the things that are truly, truly needed and not just list everything just in case at it makes it super hard to see what actually is needed.

CMakeLists.txt
```diff
@@ -1,10 +1,275 @@
...
+    if(WIN32)
+        return()
+    endif()
+
+    target_compile_options(${target} PRIVATE
+        -Wall
+        -Wextra
+        -Wpedantic
+        -Wno-unknown-pragmas
+        -fPIC
```
hhyyrylainen, line 141: This shouldn't be needed because we use the POSITION_INDEPENDENT_CODE from Cmake which ensures it is correctly set for all compilers cmake supports.

CMakeLists.txt
```diff
@@ -1,10 +1,275 @@
...
+        target_include_directories(${target} BEFORE PRIVATE 
+            ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES}
+            ${PROJECT_BINARY_DIR}
+            ${PROJECT_SOURCE_DIR}/third_party/JoltPhysics
+        )
+    else()
+        configure_linux_target(${target})
+    endif()
+
+    target_compile_definitions(${target} PRIVATE
```
hhyyrylainen, line 170: JPH stuff should all be configured in the third party cmakelists folder where that target is imported from, keeps stuff much cleaner.

CMakeLists.txt
```diff
@@ -1,10 +1,275 @@
...
+    endif()
+endfunction()
+
+# OS detection
+if(WIN32)
+    set(THRIVE_OS "windows")
+    set(THRIVE_ARCH "x64")
+else()
+    set(THRIVE_OS "linux")
+    execute_process(COMMAND uname -m OUTPUT_VARIABLE THRIVE_ARCH OUTPUT_STRIP_TRAILING_WHITESPACE)
```
hhyyrylainen, line 213: Did you add support for compiling as 32-bit? If you didn't I think this line is totally useless.

third_party/CMakeLists.txt
AverageNerdz: done :]

third_party/windows/godot-cpp-windows.cmake
AverageNerdz: I should have probably dvelved into more an yeah that temporary fix is a stretch but i found out that the main culprit is the cmakelist from the godot-cpp folder, and i don't know if they may release a better cmake which is also suitable for windows uses in the future that is why i put this in the comment

CMakeLists.txt
AverageNerdz: the problem isn't in the thrive code pal, the problem is in the thousands of files from the godot library which tended to give a lot of warnings and sometimes errors making the compile task quite tedious

CMakeLists.txt
AverageNerdz: done :]

CMakeLists.txt
AverageNerdz: this helps printing the files that are being compiled in the unity instead of just having the unity being compiled it is really important for the compile imho

CMakeLists.txt
AverageNerdz: i did pal

src/extension/CMakeLists.txt
AverageNerdz: if you have another way to do it recursively i'm all ears pal, i added GLOB because it seemed the easier to implement and the one that was pretty easy on the CPUs plus it allows anyone to add anything they want in the folder and just start working on it without editing addittional code other than the interops

CMakeLists.txt
AverageNerdz: you mean like putting all this code just in the third party?

CMakeLists.txt
```diff
@@ -1,10 +1,275 @@
...
+            VS_GLOBAL_UseUnitySupport "true"
+            VS_GLOBAL_UnitySupport "true"
+        )
+    endif()
+endfunction()
+
+# Windows detection function
+function(configure_windows_target target)
+    if(NOT WIN32 OR NOT MSVC)
+        return()
```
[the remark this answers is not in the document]
AverageNerdz: have you tested it? if you mean the cpu part i would be really glad to know how it is done in Linux i tried to make something that works on both sides and doesn't impact the compile speed but as i don't have those platforms in hand it is difficult to make something without any feedback

CMakeLists.txt
AverageNerdz: you mean keeping all the compile as is for Linux and just use .cmakes for windows in the implementation you suggested?

CMakeLists.txt
AverageNerdz: done, i added something to it too :]

src/extension/CMakeLists.txt
hhyyrylainen: The only good way is to manually list each file. That way cmake is forced to re-run when a new file is defined and everything works fine. With GLOB you have to *manually* re-run cmake whenever you add a file so it is a major pain and you cannot easily see what all gets compiled in. So I'm totally against the use of GLOB at all for defining cmake files to compile.

CMakeLists.txt
hhyyrylainen: Long helper code like this should be moved to a separate file (probably fine to have in the scripts folder like `Scripts/CMakeHelperMethods.cmake`) for cleaning things up in the main files.

CMakeLists.txt
hhyyrylainen: No, I mean putting all of the long code you plan to keep into a separate file so that reading the main cmake file doesn't require you to read through a ton of specific details before getting to the options etc. actually important parts of the file.

CMakeLists.txt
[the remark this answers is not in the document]
hhyyrylainen: This is literally just an early return that seems to not do anything. So if someone accidentally calls this method, then they *silently* don't get any feedback that nothing happened. That's very unintuitive which is why error conditions should print error messages.

CMakeLists.txt
hhyyrylainen: Then make it so that the godot library gets added these as public compile flags, that's the modern way to handle this in cmake. I see quite a lot of cmake usage where the old way to do things is used instead of the modern and more clean cmake way.

CMakeLists.txt
hhyyrylainen: Yes, in the third party cmakelists, I've already gathered quite a lot of JPH config there so having it all in one place just makes way too much sense (and when the next time JPH build breaks I'll be able to find all of the stuff related to it easily instead of having to hunt around for it).

third_party/windows/godot-cpp-windows.cmake
hhyyrylainen: This is going to be really annoying to maintain as each specific godot-cpp update *might* break our build, and then I wouldn't notice it until someone tries using MSVC the next time and finds it broken.

If at all possible it would be better to try to modify any variables this script uses or modify the resulting targets it exports, so that any forceful customizations could still be kept outside this file (so that we can use the upstream file as-is so that we don't have to forever maintain our separate version).

Though one more alternative would be to open an issue / talk to the godot-cpp repo people and ask if they would accept changes to their file that make it MSVC compatible.

AverageNerdz:
I made a few updates to the code, addressing the 32/64 bit exchange problem and removing the glob features, hope this one is better than the previous version

CMakeLists.txt
hhyyrylainen: I still don't see the comment text about Jolt requiring the Distribution type...

CMakeLists.txt
```diff
@@ -1,10 +1,72 @@
...
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.13)
 
-project(Thrive)
+# Must be defined before project() call
+set(CMAKE_CONFIGURATION_TYPES "Debug;Release;RelWithDebInfo;Distribution" CACHE STRING "" FORCE)
 
-# If you want to get compile commands run cmake with
-# "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
+project(Thrive VERSION 1.0 LANGUAGES CXX)
```
hhyyrylainen, line 7: Specifying a version *here* I think is really bad as it will never be updated so it makes no sense to have it here.

CMakeLists.txt
```diff
@@ -1,10 +1,72 @@
...
+
+# Add our custom CMake scripts directory
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/Scripts")
+include(CMakeConfiguration)
+include(CMakeHelperFunctions)
+
+if(WIN32)
+    set(CMAKE_GENERATOR_PLATFORM x64)
```
hhyyrylainen, line 15: Indentation is still wrong, should be exactly 2 spaces per level.

CMakeLists.txt
```diff
@@ -1,10 +1,72 @@
...
+endif()
+
+message(STATUS "Building for ${THRIVE_OS} on ${THRIVE_ARCH}")
+
+# Detect library version
```
hhyyrylainen, line 20: I think this code still very much needs to be after the user-configurable option variables which I think should be as close to the top of the file as possible. I'd put them even way before including any of the helper functions etc.

CMakeLists.txt
```diff
@@ -39,172 +97,66 @@ option(USE_ATOMIC_COLLISION_WRITE
...
+if(MSVC AND CMAKE_BUILD_TYPE STREQUAL "Debug")
+    install(FILES 
+        $<TARGET_PDB_FILE:thrive_native>
+        $<TARGET_PDB_FILE:thrive_extension>
+        DESTINATION bin 
+        OPTIONAL
+    )
+endif()
+
+message(STATUS "Building Thrive version ${NATIVE_LIBRARY_VERSION}")
```
hhyyrylainen, line 160:
```suggestion
message(STATUS "Building ThriveNative version ${NATIVE_LIBRARY_VERSION}")
```

also pretty weird that only one library's version is mentioned.

Scripts/CMakeConfiguration.cmake
hhyyrylainen, line 1: Thisi file has wrong indentation and doesn't end with a newline, same with the other added cmake file in this folder.

Scripts/CMakeConfiguration.cmake
```diff
@@ -0,0 +1,134 @@
...
+    set_target_properties(${target} PROPERTIES
+        MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL"
+    )
+
+    configure_unity_build(${target})
+endfunction()
+
+function(configure_linux_target target)
+    if(WIN32)
+        return()
```
hhyyrylainen, line 87: This is missing an error

Scripts/CMakeConfiguration.cmake
```diff
@@ -0,0 +1,134 @@
...
+        target_include_directories(${target} BEFORE PRIVATE 
+            ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES}
+            ${PROJECT_BINARY_DIR}
+        )
+    else()
+        configure_linux_target(${target})
+    endif()
+endfunction()
+
+# Function to configure godot-cpp target without modifying its CMakeLists
```
hhyyrylainen, line 120: This stuff is just needed for the third party stuff (right?) so I would totally like to see this only be in the third_party folder cmakelists file.

src/extension/CMakeLists.txt
```diff
@@ -14,45 +14,192 @@ add_library(thrive_extension SHARED
...
 
-if(MSVC)
-  target_compile_options(thrive_extension PRIVATE /W4 /wd4068)
+target_sources(thrive_extension PRIVATE ${EXTENSION_SOURCES})
 
-  if(WARNINGS_AS_ERRORS)
-    target_compile_options(thrive_extension PRIVATE /WX)
-  endif()
+# Print files and configure build
+print_source_files(thrive_extension)
```
hhyyrylainen, line 20: I really think it should be a special `option` to print the source files and not do it by default.

src/extension/CMakeLists.txt
```diff
@@ -14,45 +14,192 @@ add_library(thrive_extension SHARED
...
 
-  # TODO debug symbols in release for MSVC (and Distribution mode)
+# Common compile definitions for both OS
```
hhyyrylainen, line 22: Any comments etc. referring to Thrive being just for TWO OSes is not going to be good because in the not too distant future *all* of this stuff also needs to be on mac. For this reason, because I think I have to do it, I don't want this huge set of cmake changes that aren't what I'm used to.

src/extension/CMakeLists.txt
```diff
@@ -14,45 +14,192 @@ add_library(thrive_extension SHARED
...
+target_compile_definitions(thrive_extension PRIVATE 
+    THRIVE_NATIVE_BUILD
+    _HAS_EXCEPTIONS=1
+    JPH_PROFILE_ENABLED=0
+    JPH_DEBUG_RENDERER=0
```
hhyyrylainen, line 27: Why is this here? When compiling in debug mode this has to be defined to one or everything will break (well at least for the thrive native library, but it would be super weird to not have consistent options).

AverageNerdz:
I tried to work from the advices you gave me and reworked the environment a little bit, any feedback is appreciated pal :]

hhyyrylainen:
I took the parts I had no qualms about and put those separately onto master branch: https://github.com/Revolutionary-Games/Thrive/commit/7b3481dce0c0478fc5cc19c037bccfe9f79047f5

AverageNerdz:
> I took the parts I had no qualms about and put those separately onto master branch: [7b3481d](https://github.com/Revolutionary-Games/Thrive/commit/7b3481dce0c0478fc5cc19c037bccfe9f79047f5)

Thanks pal i'll close this pr then as it fullfilled its purpose if the solution needs further reworking i'll work from the branch you merged into the master, have a nice one :)

PR #5653, opened 2024-11-04
Balance Changes to Glucose Consumption Rate, Storage Values 

Deus-Codes:
**Brief Description of What This PR Does**

This PR does some stuff...

* Slightly reduced photosynthesis glucose output for chloroplasts, though maintained for thylakoids. 
* Slightly reduced mitochondria (aerobic respiration) glucose input to burn less glucose while maintaining same amount of ATP burn. This should make glucose last a bit longer, making respiration a bit more viable
* Generally halved storage of most internal parts, except for lysosome, binding agents, and toxin vacoules. Maintained storage values of external parts. This should make strategizing around storage more of an intentional series of decisions for the player, though worry about it being too harsh. Vacuole storage has been reduced, but the scarcity of storage makes it even more important for certain playstyles. 
* Slightly reduced energy cost of flagella, allowing them to be a bit more accessible - especially as they duplicate. 

The result should allow for more build diversity away from photosynthesis, as players now burn through the glucose they worked so hard to get less quickly. This hopefully should allow predation to be more rewarding, which when paired with a less energy-demanding flagella should  encourage these builds more. Slightly reduced chloroplast output should hopefully make players be more concise about photosynthesis-related adaptations, making sure not to push their glucose budget.

Storage reduction makes cytoplasms and vacoules more important, especially for photosynthetic organisms. Now, players have to more intentionally create photosynthetic organisms focused on storage, which could alter player behavior and strategy towards being more reflective of real-life evolutionary trends. I think the storage changes to most parts are reasonable, but my reduction of nucleus storage could be something I bump up a bit if playtesters think it is too severe of a drop. 

**Progress Checklist**

Note: before starting this checklist the PR should be marked as non-draft.

- [ ] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [ ] Initial code review passed (this and further items should not be checked by the PR author)
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author. Merging must follow our
[styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md#git).

PR #5659, opened 2024-11-05
Glucose Consumption & Storage Balance Tweaks

Deus-Codes:

Note that this is the same pull request as: https://github.com/Revolutionary-Games/Thrive/pull/5653
I just realized that the branch was out of sync, so I needed to re-sync the branch and thus undo the previous pull request.

* Slightly reduced photosynthesis glucose output for chloroplasts, though maintained for thylakoids. 
* Slightly reduced mitochondria (aerobic respiration) glucose input to burn less glucose while maintaining same amount of ATP burn. This should make glucose last a bit longer, making respiration a bit more viable
* Generally halved storage of most internal parts, except for photosynthesis-related parts, nucleus, lysosome, binding agents, and toxin vacuoles. Maintained storage values of external parts. This should make strategizing around storage more of an intentional series of decisions for the player, though worry about it being too harsh. Vacuole storage has been reduced, but the scarcity of storage makes it even more important for certain playstyles. 
* Slightly reduced energy cost of flagella, allowing them to be a bit more accessible - especially as they duplicate. 

The result should allow for more build diversity away from photosynthesis, as players now burn through the glucose they worked so hard to get less quickly. This hopefully should allow predation to be more rewarding, which when paired with a less energy-demanding flagella should  encourage these builds more. Slightly reduced chloroplast output should hopefully make players be more concise about photosynthesis-related adaptations, making sure not to push their glucose budget.

Storage reduction makes cytoplasm and vacuoles more important, especially for photosynthetic organisms. Now, players have to more intentionally create photosynthetic organisms focused on storage, which could alter player behavior and strategy towards being more reflective of real-life evolutionary trends. I think the storage changes to most parts are reasonable, but my reduction of nucleus storage could be something I bump up a bit if playtesters think it is too severe of a drop. 

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Deus-Codes:
Adding notes from previous patch: investigate photosynthesis glucose output to see if it is too nerfed with dynamic compounds and check and see if vacuoles are too underpowered. Note that I tweaked vacuoles to have a bit more storage now, with 3 instead of 2.

hhyyrylainen:
I'm pretty sure the old PR was automatically updated so you shouldn't have opened a new one... oh well this is now opened so might as well continue here.

kuva

Like github there saw your push and updated the PR just after you closed it.

Deus-Codes:
> I'm pretty sure the old PR was automatically updated so you shouldn't have opened a new one... oh well this is now opened so might as well continue here.
> 
> kuva
> 
> Like github there saw your push and updated the PR just after you closed it.
> 
> Edit: okay I see that other PR ended up empty now, but it should have worked if you just pushed your changes normally as this PR is opened from the exact same branch name as the previous one.

I was having some issues with my original branch, so I decided to make another branch and delete the old one. As you can imagine, my software management skills are 
image

hhyyrylainen:
If you deleted and made a local branch with the same name as before, force pushing that to github should have been able to update your PR with the all-new commits (in effect replacing the content of it entirely). Well now you know for the future.

Also BTW I didn't have time to check if the files are the same as last I checked but that last time you only seemed to update the chloroplast and mitochondrion numbers, and didn't update the thykaloid or metabolosomes. Was that intended or should those have also been part of the balance (they do a different process variant so updating just "respiration" only affects mitochondria for example).?

Deus-Codes:
Yes, that is intentional. A lot of the complaints from the community focused on eukaryotic gameplay, whereas prokaryotic gameplay in general seems to be okay. Most commentary about transitioning towards photosynthesis did note that this transition occurred after a nucleus was put on and even when the player had multiple mitochondria on their organism.

hhyyrylainen:
> Most commentary about transitioning towards photosynthesis did note that this transition occurred after a nucleus was put on and even when the player had multiple mitochondria on their organism.

Then the problem is that you cannot get chloroplasts without first using thylakoids for many generations. [I recently balanced](https://github.com/Revolutionary-Games/Thrive/pull/5646) that duration to be lower, but still it requires many generations of thylakoid use. So nobody is going to get to chloroplasts unless thylakoids seem like a good direction *and* the chloroplast needs to be much more powerful in relation to the thylakoids to feel like a worthwhile upgrade.

Deus-Codes:
Did some playtesting on this looking specifically at photosynthesis.

with a thylakoid to nucleus progression as a low movement organism, I think that existence was pretty well balanced. Chloroplasts also demonstrated a significant level-up in photosynthetic ability, so I'm not worried there.

I want to investigate more eukaryotic gameplay though, specifically how it feels to add a few flagella. We should select against highly motile photosynthetic life-forms, but there's a difference in gameplay between those hypercarnivorous organisms with chloroplasts and a painfully slow-moving organism.

simulation_parameters/microbe_stage/organelles.json
```diff
@@ -608,7 +608,7 @@
     "Processes": {},
     "Components": {
       "Storage": {
-        "Capacity": 8
+        "Capacity": 3
```
hhyyrylainen, line 611: I'm going to bump this back up so that vacuole would actually feel really impactful, because I'm doubting that when storage is made into a problem but the solution is also nerfed really heavily, it still won't feel worth it to add a vacuole.

hhyyrylainen (approved):
I did a few further balance tweaks and applied the tweaks more consistently (made the prokaryotic photosynthesis again less than 1/3 the power of the eukaryotic part). I bumped up the vacuole storage up to 5 so that it is a noticeable jump when you run out of storage and need to place one. Maybe it could be lowered further to require multiple vacuoles to be placed but for an initial storage balance pass I think the solution to the introduced storage problems should be relatively strong.

PR #5662, opened 2024-11-06
Redo organelles with lower mesh count

HexapodPhilosopher:

Adjusts models and shaders for organelles. Each organelle now uses a single model and texture. Several textures are now shared between models.

* Remade Nucleus - model adjustment and retexturing. Added pores and visible DNA. Made the Golgi body and ER stand out more. Now has a single model and texture. Poly count increased from 1438 to 2170.
* Redid Binding Agent with a single model and texture
* Remade Cilia as a single model, with fading at the base. Redid animation, and combined texture into Pilus texture.
* Adjusted Chemoplast. Fits better on the hex.
* Cleaned up Chemorecepter and added fading to base.
* Reduced Ferroplast from 3 models to 1. Removed Membrane texture.
* Added fading to Flagellum base. Made animation more curvy. Set to same texture as Pilus.
* Redid Lysosome as a single model.
* Remade Mitochondrion. Now more wiggly and more textured.
* Redid Nitroplast with half the polygons.
* Made shared texture for Nitoplast and Nitrogenase.
* Added fading to pilus and injectisome
* Redid Rusticyanin, sharing texture with Ferroplast.
* Reduced Thermoplast to single model and texture. Particles made more visible, and emission added.
* Redid Thylakoids and Chloroplast using a single texture. Chloroplast texturing is now more consistent, and poly count reduced by 40%. Thylakoid poly count reduced by 20%.
* Changed Toxisome to use same texture as Toxin Vacuole.

* Re-oriented all organelles to match the shader animation.

	
Shader improvements:
* Changed the angle of the shader vertex animation, so that default behavior (jiggling about in all directions when seen from above) works without the organelles having to be rotated first. This will make it easier to add new parts.
* Added scaling to dissolve effect, causing smaller or larger pieces of the mesh to be visible while dissolving.
* Clamped min and max values for emission, to remove bright spots that sometimes appear due to the dissolve shader
* Redid Fresnel effect on organelles, to give more control over visible insides. Now 0 = no effect, 1 = full effect.
* Added emissive organelle shader, used for thermoplast particle glow.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


assets/models/organelles/Cytoplasm.tscn
```diff
@@ -5,13 +5,14 @@
...
 resource_local_to_scene = true
 render_priority = 0
 shader = ExtResource("3")
-shader_parameter/dissolveValue = null
-shader_parameter/fresnelValue = 1.0
-shader_parameter/fresnelActive = false
+shader_parameter/dissolveValue = 0.0
+shader_parameter/dissolveScale = 1.0
+shader_parameter/fresnelValue = 0.8
+shader_parameter/fresnelActive = true
```
hhyyrylainen, line 15: Is it intentional that this model has fresnel mode active? If I remember right that didn't used to be enabled for anything in the game.

assets/textures/ExternalOrganelle.png
hhyyrylainen, line 1: Why is this named external organelle? Wasn't `Pilus.png` a good name?

assets/textures/Nucleus2.png.import
hhyyrylainen, line 1: What's the difference between this and the other nucleus texture? If both are needed could the names be clarified a bit?

assets/textures/PhotosynthesisOrganelle.png
hhyyrylainen, line 1: What's the reasoning with these very generic names? Usually it makes it much more clear to name the files after the organelle they are used in.

shaders/OrganelleOpaque.gdshader
```diff
@@ -4,11 +4,11 @@ render_mode depth_draw_opaque;
 uniform sampler2D texture : source_color;
 
 uniform sampler2D dissolveTexture : source_color;
-uniform float dissolveValue : hint_range(0, 1);
+uniform float dissolveValue : hint_range(0, 1) = 0.0;
+uniform float dissolveScale = 1.0;
```
hhyyrylainen, line 8: I assume the scale is 0-1 here? If so could you add a `hint_range` for this? That makes the value show up more intuitively in the Godot editor.

shaders/OrganelleTransparent.gdshader
```diff
@@ -1,16 +1,17 @@
 shader_type spatial;
-render_mode depth_prepass_alpha;
+render_mode depth_prepass_alpha, diffuse_lambert_wrap;
 
 uniform sampler2D texture : source_color;
 
 uniform sampler2D dissolveTexture : source_color;
-uniform float dissolveValue : hint_range(0, 1);
+uniform float dissolveValue : hint_range(0, 1) = 0.0;
+uniform float dissolveScale = 1.0;
```
hhyyrylainen, line 8: I think these other shader files also have the same thing about the hint range.

shaders/OrganelleEmissive.gdshader
```diff
@@ -0,0 +1,74 @@
...
+        vec4(0.f, 1.f, 0.f, 0.f),
+        vec4(sin(angle),  0.f, cos(angle), 0.f),
+        vec4(0.f, 0.f, 0.f, 1.f)
+    );
+
+    VERTEX = (rotation * vec4(VERTEX, 1.0f)).xyz;
+}
+
+void fragment() {
+    vec2 base_uv = UV;
```
hhyyrylainen, line 49: Variables should be named in camelCase and not snake_case

hhyyrylainen: Though is there a reason why this needs to be a variable? It looks like the UV is not modified so why not just use `UV` in the places this is used?

shaders/OrganelleEmissive.gdshader
```diff
@@ -0,0 +1,74 @@
...
+        float(-0.5 + clamp(dissolveValue, 0, 1));
+
+    vec3 dissolveOutline = vec3(round(1.0 - float(cutoff - outlineWidth))) *
+        growColor.rgb;
+
+    ALBEDO = final.rgb;
+
+    SPECULAR = specular;
+
+    vec4 roughness_texture_channel = vec4(1.0, 0.0, 0.0, 0.0);
```
hhyyrylainen, line 64: There's multiple snake_case variable names here.

assets/models/organelles/Flagellum.tscn
```diff
@@ -1,45 +1,47 @@
...
+bones/15/position = Vector3(0.282216, -0.030774, -4.40677)
 
-[node name="flagella" parent="Armature_001/Skeleton3D" index="0"]
-material_override = SubResource("ShaderMaterial_ff6tk")
+[node name="FlagellumMesh" parent="Armature_001/Skeleton3D" index="0"]
+material_override = SubResource("ShaderMaterial_provs")
 
 [node name="AnimationPlayer" parent="." index="1"]
 autoplay = "swim"
-speed_scale = 0.3
```
hhyyrylainen, line 44: Why is the default speed scale gone? The intention with that is that the flagellum only plays the full speed swimming animation when the C# code commands it, and otherwise is like in the "idle" state, as there's many more cases where it is useful to have the idle animation than needing to show the max speed.

assets/models/organelles/Nucleus.tscn
```diff
@@ -1,69 +1,27 @@
...
+[ext_resource type="Texture2D" uid="uid://baxuoyeo83r2u" path="res://assets/textures/dissolve_noise.tres" id="2_tdl0s"]
+[ext_resource type="Texture2D" uid="uid://bh1e6qrcmeoux" path="res://assets/textures/Nucleus.png" id="3_1w7t3"]
+[ext_resource type="ArrayMesh" uid="uid://buw78wgilybsy" path="res://assets/models/organelles/Nucleus.mesh" id="4_tmwr2"]
 
-[sub_resource type="ShaderMaterial" id="1"]
+[sub_resource type="ShaderMaterial" id="ShaderMaterial_ngmrq"]
 resource_local_to_scene = true
-render_priority = 0
-shader = ExtResource("2")
+render_priority = 1
```
hhyyrylainen, line 10: Is this required? This may cause unintended rendering on top of organelles that are in front o the nucleus. The reason this was used before is that the nucleus was multiple parts and it was required to get the draw order for them correct.

hhyyrylainen (changes requested):
Overall this looks to be in pretty good shape already, but I had a few correction suggestions. Though I didn't test in game yet to see if there's some unexpected issues I couldn't see from just the text data.

Also did you test in opengl3 mode? That needs to at least look decent after these major graphics updates. The one new shader is the most likely thing that *could* have issues there.

assets/models/organelles/Cytoplasm.tscn
HexapodPhilosopher: That's just the parameter that's been in OrganelleOpaque.gdshader since last year at least. Some of the membranes (including the Ferroplast one) were using it previously.
I put it here just to make it more transparent, but I could just make the texture more transparent instead.

assets/textures/ExternalOrganelle.png
HexapodPhilosopher: I changed it because both forms of pilus as well as the flagellum and cilia now all use that texture.

assets/textures/Nucleus2.png.import
HexapodPhilosopher: Damn, that should have been removed.

shaders/OrganelleOpaque.gdshader
HexapodPhilosopher: 1.0 is the default, but values above 1 decrease the size of the noise. Putting it like that, maybe I should have called it 'dissolveResolution'...
Values from 0.5 to 5.0 are fine.

shaders/OrganelleEmissive.gdshader
HexapodPhilosopher: I saw that used somewhere (outside Thrive) and thought it might reduce the number of times it has to check the mesh. I didn't find info on that.

assets/models/organelles/Flagellum.tscn
HexapodPhilosopher: That was unintended. It seemed to work without it, though.

HexapodPhilosopher:
I didn't test it in OpenGL3 mode. I'll have to do that now.

assets/models/organelles/Nucleus.tscn
HexapodPhilosopher: I was using it while trying out different ways to do the nucleus. It's unnecessary now.

assets/models/organelles/Cytoplasm.tscn
hhyyrylainen: Yeah I know the parameter was added ages ago but it was removed from all of the organelles. I suppose if it looks fine and the dissolve effect also looks fine with it on for specific organelles, it can be kept on.

assets/textures/ExternalOrganelle.png
hhyyrylainen: The slime jet is also an external organelle, so I still think this should be have a better name as that doesn't use it. Maybe like `StrandOrganelle.png` as that would slightly narrow it down?

assets/textures/PhotosynthesisOrganelle.png
hhyyrylainen: That sounds like good justification, it just isn't immediately clear to me what it is for. But now that I know, I approve of the new name.

shaders/OrganelleOpaque.gdshader
hhyyrylainen: I think the name is clear enough, but you should add `hint_range(0.5, 5)` here to make the Godot editor show a slider for picking the value.

shaders/OrganelleEmissive.gdshader
hhyyrylainen: I think the way it works is that godot just setups the variable once before starting the fragment shader. So I highly think it doesn't matter. And if I remember right in like GLSL or HLSL you need to specify the UV as a method parameter so it is pretty clear there that the data is copied from the mesh into local memory for the shader when setting it up so again I highly doubt it affects used memory bandwidth.

assets/models/organelles/Flagellum.tscn
hhyyrylainen: I'm 90% sure that the editor would end up displaying the animation playing at 100% speed which looks very busy and not very passive in the editor. So this change would increase visual business in the editor (if I'm right in how I expect this to affect things).

assets/models/organelles/Nucleus.tscn
hhyyrylainen: Alright so it can be removed, I'll wait for your commit changing that.

assets/models/organelles/Cytoplasm.tscn
HexapodPhilosopher: Was there a specific reason it was removed before?
Also, I forgot to mention that I altered the calculation for it slightly.
It's now used only by lysosome, Signalling Agent, Vacuole and the endosome.

hhyyrylainen: I can't remember / know, but I think it was removed as it looked bad. So it's fine to put it back in if it looks good now.

HexapodPhilosopher:
The Thermoplast issue seems to be because of the lighting issue on my system. The emission is working outside the editor, and is inactive in the editor.
I think I'll stick with it, and I'll just have to get feedback on if anyone else has issues.

assets/models/organelles/Cilia.tscn
```diff
@@ -1,102 +1,57 @@
...
+bones/23/rotation = Quaternion(0.0341753, 0.119834, -0.0883902, 0.988261)
+bones/24/rotation = Quaternion(-0.0142192, -0.114817, 0.051429, 0.991953)
+
+[node name="CiliaMesh" parent="Armature/Skeleton3D" index="0"]
+material_override = SubResource("ShaderMaterial_jjmap")
+mesh = ExtResource("5_uvfp1")
+
+[node name="AnimationPlayer" parent="." index="1"]
 autoplay = "wiggle"
-speed_scale = 0.2
```
hhyyrylainen, line 102: This also has the changed speed scale like flagellum. I think you fixed that there, but this is still like this...

assets/models/organelles/Lysosome.tscn
```diff
@@ -1,49 +1,26 @@
...
+[ext_resource type="ArrayMesh" uid="uid://vc7ghifv2htv" path="res://assets/models/organelles/Lysosome.mesh" id="1_ov5jm"]
+[ext_resource type="Texture2D" uid="uid://baxuoyeo83r2u" path="res://assets/textures/dissolve_noise.tres" id="2_g5t7g"]
+[ext_resource type="Texture2D" uid="uid://dmwdx5hfmr0ip" path="res://assets/textures/Lysosome.png" id="3_dbsut"]
 
-[sub_resource type="ShaderMaterial" id="3"]
+[sub_resource type="ShaderMaterial" id="ShaderMaterial_78kbi"]
 resource_local_to_scene = true
-render_priority = 0
-shader = ExtResource("4")
+render_priority = 1
```
hhyyrylainen, line 10: Another render priority I spotted...

assets/models/organelles/Nucleus.tscn
```diff
@@ -1,69 +1,27 @@
...
-mesh = ExtResource("4")
-
-[node name="ER" type="MeshInstance3D" parent="."]
-transform = Transform3D(-0.630648, 0.776067, 1.26437e-07, -0.776067, -0.630648, -2.65666e-07, -1.26437e-07, -2.65666e-07, 0.999999, 0.0103581, -0.0146696, -0.00807817)
-material_override = SubResource("3")
-mesh = ExtResource("7")
+rotation_edit_mode = 2
+material_override = SubResource("ShaderMaterial_ngmrq")
+mesh = ExtResource("4_tmwr2")
+skeleton = NodePath("")
```
hhyyrylainen, line 27: Skeleton? Why does the Nucleus have a skeleton set.

hhyyrylainen:
What's the problem with the thermoplast? Does the new emission shader not work or? I noticed just a few more small technical details but all of the other files I didn't comment on now seem fine to me so this is very close to being done. BTW if you have the models as .blend files for example (or another modelling program format) could you upload those files to https://dev.revolutionarygamesstudio.com/files/Raw%20Assets/Models/Organelles so that in the future there's the raw source files if any future tweaks are required for the organelle models?

assets/models/organelles/Nucleus.tscn
HexapodPhilosopher: It doesn't have a skeleton. I'm not sure how that got added.

HexapodPhilosopher:
The emission shader works fine, except in the cell editor when in OpenGL mode (where the emission seems to be inactive).
After the Bloom slider PR was added, the maximum effect of the emission in terms of the glow around the emitting object seems much lower. That doesn't make a noticeable difference for the Thermoplast, though.

hhyyrylainen:
This is probably super hard to find but Godot doesn't fully like one of the new models I think as I got this when first importing all the new changes: `scene/resources/3d/importer_mesh.cpp:521 - Ignoring face with non-finite normal in LOD generation.`

(I'm now doing some final checks on this before merging)

hhyyrylainen (approved):
I tested in opengl mode and only really the thermoplast looked bad, kind of like a disgusting biological thing actually but I think that's good enough for not officially supported mode (which is still totally playable). I looked at all the models in game and I think they work fine. I noticed a slight edge in the slime jet model at some very specific angles in the game, but other than that I don't have anything really comment on the models so good job! 

I'll merge this now but if at all possible I'd still like to see the source 3D models added to the devcenter so that if someone else in like 5 years needs to modify an organelle slightly that would be possible to do (https://dev.revolutionarygamesstudio.com/files/Raw%20Assets/Models/Organelles).

PR #5664, opened 2024-11-07
Made gas updates recalculate total fractions and made nitrogen vary

hhyyrylainen:
Improves the dynamic gas compounds

Also fixes a bug with the compound distribution that could cause too much compound to move resulting in the target having more than the source in the end

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/BiomeConditions.cs
+++ b/src/microbe_stage/BiomeConditions.cs
@@ -241,8 +241,6 @@ public class BiomeConditions : IBiomeConditions, ICloneable
 
         float previousOther = 1 * biomeDetails.GasVolume - previousTotal;
 
-        // TODO: handling for keeping some space for other compounds?
-
         foreach (var change in changes)
         {
             if (!simulationParameters.GetCompoundDefinition(change.Key).IsGas)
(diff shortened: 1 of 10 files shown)


src/microbe_stage/BiomeConditions.cs
```diff
@@ -170,6 +172,124 @@ public bool TryGetCompound(Compound compound, CompoundAmountType amountType, out
...
+        // TODO: handling for keeping some space for other compounds?
+
+        foreach (var change in changes)
+        {
+            if (!simulationParameters.GetCompoundDefinition(change.Key).IsGas)
+                continue;
+
+            gases.TryGetValue(change.Key, out var updatedValue);
+
+            updatedValue = Math.Clamp(updatedValue + change.Value, 0, 1);
```
dligr, line 253: Shouldn't the maximum value for the clamp be `biomeDetails.GasVolume`? Otherwise, with a high enough gas volume, the `updatedValue` may be higher than 1 and cause a clamp when it shouldn't.

hhyyrylainen: Yeah, I think that is true on second read of that code so I'll make that change. Good catch.

PR #5691, opened 2024-11-16
Add endosymbiosis requirement

### LINKED ISSUE #5134: Add text in organelle unlock conditions popup to mention endosymbiosis option
so that players will have one more natural way to discover the potential of endosymbiosis

Patryk26g:

Add endosymbiosis to organell unlocks list

**Related Issues**

Closes #5134

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
This isn't exactly what I had in mind in relation to the issue. My thinking was that by default the organelle unlocks tooltip would have a sentence added saying that unlocks through endosymbiosis are also possible. I didn't test in the game but from the code this seems like this adds the text specifically into the unlock conditions of 2 organelles (which isn't all of the ones that can be gained through endosymbiosis).

Patryk26g:
image
Hopefully this is fine :D

hhyyrylainen:
It's still not what I'm exactly wanting, because the text now repeats and makes the tooltip a lot longer than it needs to be. What I meant by my previous comment is that you'd literally just extend the text at the top to say: Unlock conditions are checked only once per generation when entering the editor. Some organelles can alternatively be unlocked through endosymbiosis.

locale/en.po
```diff
@@ -4015,7 +4015,7 @@ msgid "ORGANELLES_BUTTON"
 msgstr "Organelles"
 
 msgid "ORGANELLES_WILL_BE_UNLOCKED_NEXT_GENERATION"
-msgstr "Unlock conditions are checked only once per generation when entering the editor"
+msgstr "Unlock conditions are checked only once per generation when entering the editor. Some organelles can alternatively be unlocked through endosymbiosis"
```
hhyyrylainen, line 4018:
```suggestion
msgstr "Unlock conditions are checked only once per generation when entering the editor. Some organelles can alternatively be unlocked through endosymbiosis."
```

For consistency I'd end the second sentence with a dot as well.

With that change I think this might just be fully done, which is probably why I marked this as an easy issue as it doesn't take much to do.

hhyyrylainen (approved):
I think this is fine now. Sorry if the original issue description was not clear enough.

PR #5693, opened 2024-11-18
Oxygen and Carbon Dioxide Rebalance

adQuid:

Balances O2 and CO2 more tightly in the special case of that (mostly) closed loop between them. Also simplifies the diffusion simulation to let those compounds spread a little more, and modifies the auto-evo to make photosynthesisers more common to allow for more oxygen in the first place.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/world_effects/VolcanismEffect.cs
```diff
@@ -32,6 +33,8 @@ private void ApplyVolcanism()
     {
         foreach (var patchKeyValue in targetWorld.Map.Patches)
         {
+            GD.Print(patchKeyValue.Value + " volcanism");
```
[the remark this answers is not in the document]
adQuid: I knew I forgot something!

adQuid:
I want to make a future PR doing the diffusion better, but right now I think this is good enough for surface patches. Would appreciate lots of testing, since that gets very time-intensive.

src/general/world_effects/VolcanismEffect.cs
```diff
@@ -72,7 +72,7 @@ private void ProduceCO2(Patch patch, float co2Strength, float threshold)
             return;
 
         // TODO: should this clamp or not?
-        addedCo2[Compound.Carbondioxide] = Math.Clamp(amount.Ambient + co2Strength, 0, threshold);
+        addedCo2[Compound.Carbondioxide] = Math.Clamp(co2Strength, 0, threshold);
```
hhyyrylainen, line 75: I think this is a good catch (fixes a slight bug I've made).

src/general/world_effects/CompoundDiffusionEffect.cs
```diff
@@ -27,14 +26,12 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
         HandlePatchCompoundDiffusion();
     }
 
-    private static (float Ambient, float Density) CalculateWantedMoveAmounts(Patch sourcePatch, Patch adjacent,
+    private static (float Ambient, float Density) CalculateWantedMoveAmounts(Patch adjacent,
         KeyValuePair<Compound, BiomeCompoundProperties> compound)
     {
         // Apply patch distance to diminish how much to move (to make ocean bottoms receive less surface
         // resources like oxygen)
```
hhyyrylainen, line 33: Isn't this comment now incorrect?

src/general/world_effects/CompoundDiffusionEffect.cs
```diff
@@ -27,14 +26,12 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
...
-        // TODO: improve the formula here as sqrt isn't the best
-        float moveModifier = Constants.COMPOUND_DIFFUSE_BASE_MOVE_AMOUNT /
-            MathF.Sqrt(Constants.COMPOUND_DIFFUSE_BASE_DISTANCE + Math.Abs(sourcePatch.Depth[0] - adjacent.Depth[0]));
+        float moveModifier = Constants.COMPOUND_DIFFUSE_BASE_MOVE_AMOUNT;
```
hhyyrylainen, line 34: Just using a flat modifier here makes it so that there's nothing preventing oxygen from going to the deepest ocean very fast now, which is one of the major reasons I added the distance modifier. Maybe instead we could rewrite the code to only use the distance modifier if the distance is like at least 500?

src/general/world_effects/PhotosynthesisProductionEffect.cs
```diff
@@ -114,27 +113,48 @@ private void ApplyCompoundsAddition()
...
+            patch.Biome.TryGetCompound(Compound.Carbondioxide, CompoundAmountType.Biome, out var existingCo2);
+
+            float total = existingOxygen.Ambient + existingCo2.Ambient;
 
-            changesToApply.Clear();
+            if (oxygenOut == 0)
+            {
+                if (co2Out == 0)
+                {
+                    // in the rare event we aren't making either compound, do nothing
```
hhyyrylainen, line 138: But now consuming the compounds also doesn't do anything? I think this is a wrong conclusion to make in the code.

src/general/world_effects/PhotosynthesisProductionEffect.cs
```diff
@@ -114,27 +113,48 @@ private void ApplyCompoundsAddition()
...
+            }
+            else
+            {
+                oxygenTarget = oxygenOut / (oxygenIn + co2In) * total;
+                co2Target = co2Out / (oxygenIn + co2In) * total;
+            }
 
-            if (co2Balance != 0)
-                changesToApply[Compound.Carbondioxide] = co2Balance;
+            changesToApply[Compound.Oxygen] = (oxygenTarget - existingOxygen.Ambient) / 2.0f;
```
hhyyrylainen, line 156: It's more efficient to multiply by 0.5f than to do a float division.

src/general/world_effects/PhotosynthesisProductionEffect.cs
```diff
@@ -114,27 +113,48 @@ private void ApplyCompoundsAddition()
...
+                co2Target = total;
+            }
+            else if (co2Out == 0)
+            {
+                oxygenTarget = total;
+                co2Target = 0;
+            }
+            else
+            {
+                oxygenTarget = oxygenOut / (oxygenIn + co2In) * total;
```
hhyyrylainen, line 152: Could you add a comment here explaining the logic? I don't really understand this kind of scaling because ApplyLongTermCompoundChanges *already* does a relative applying of gases. So higher oxygen output reduces the effect of the co2 adding and vice versa. So to me it looks like this is now doing the same effect twice, which doesn't logically make a ton of sense as ApplyLongTermCompoundChanges needs to be given the absolute change wanted and then it automatically makes gases related to each other and caps it to a 100%.

hhyyrylainen (changes requested):

src/general/world_effects/CompoundDiffusionEffect.cs
adQuid: The number of patches between the epipelagic and sea floor already do that pretty well, but based on my quick research even that isn't accurate. As Buckly mentioned in the last Thrivestream, oxygen is actually pretty well spread in the modern oceans: https://en.wikipedia.org/wiki/Oxygen_minimum_zone

adQuid: Based on what I'm seeing, Oxygen in the bathypelagic or abyssopelagic should be around a third of the surface, while in this current code it's more like a tenth at best

hhyyrylainen: I guess I have to accept that as much as I think that will lead to players submitting bug reports I have to then read (as I'm almost the only one dealing with bug reports)... Though the comment that was left above is then obviously incorrect as it no longer related to the changed code.

src/general/world_effects/PhotosynthesisProductionEffect.cs
adQuid: You are correct. If microbes had only bio processes that destroyed O2 or CO2 without producing anything in return, this code doesn't account for that. Right now the PhotosynthesisProductionEffect class is dependent on the special case of conservation of mass (and that CO2 and O2 are only being converted into each other)

src/general/world_effects/PhotosynthesisProductionEffect.cs
adQuid: Fair. Rather than explain here, I will add the comment and see if the explanation makes sense to you.

src/general/world_effects/PhotosynthesisProductionEffect.cs
hhyyrylainen: The code shouldn't depend on that then as someone will make a mod eventually that breaks the laws of physics and will get confused why it doesn't do things correctly.

adQuid: alright, I've modified the code to handle matter sinks, and while a matter generator will result in absurd numbers, I've protected against divide-by-zero exceptions.

src/general/world_effects/PhotosynthesisProductionEffect.cs
```diff
@@ -114,27 +114,57 @@ private void ApplyCompoundsAddition()
...
+                oxygenTarget = oxygenProduced / (oxygenConsumed + co2Consumed + MathUtils.EPSILON) * total;
+                co2Target = co2Produced / (oxygenConsumed + co2Consumed + MathUtils.EPSILON) * total;
+            }
 
-            if (co2Balance != 0)
-                changesToApply[Compound.Carbondioxide] = co2Balance;
+            changesToApply[Compound.Oxygen] = (oxygenTarget - existingOxygen.Ambient) * 0.5f;
+            changesToApply[Compound.Carbondioxide] = (co2Target - existingCo2.Ambient) * 0.5f;
 
             if (changesToApply.Count > 0)
```
hhyyrylainen, line 169: This if no longer works correctly after your changes because even if the change to apply is less than EPSILON the value is added to the dictionary so this check no longer works. Either it needs to be reworked that small values (so close to target) are not processed or this needs to loop here to check if any of the changes are above EPSILON. I think the first approach will be easier to do and clearer to understand here.

hhyyrylainen (changes requested):
I'd like to see a few more small changes (I forgot to post my last comment as part of this review).

HexapodPhilosopher:
On testing this today: the compounds are a lot more stable, but it's difficult to get the oxygen level to stay above 11%. My plant cells had to rely on hydrogenase to avoid using most of the oxygen in most patches. And even after 20 generations of maximizing my population in all surface patches, half of these had metabolosomes functioning worse than hydrogenase.

hhyyrylainen:
I'll close this PR as pretty stale at this point to get this out of the incomplete list for 0.8.1.

Issue #5582, opened 2024-10-10
Improve membrane textures to look more distinct with the 3D meshes for them


hhyyrylainen:
as they now look very similar, also the damaged texture doesn't look super visible anymore (maybe due to: https://github.com/Revolutionary-Games/Thrive/pull/5579)

hhyyrylainen:
Partially done by: https://github.com/Revolutionary-Games/Thrive/pull/5699

hhyyrylainen:
I think all of the membrane textures ended up being worked on in that PR so I'll guess it's fine to now consider this issue done.

PR #5699, opened 2024-11-20
Improve visibility of membrane textures

HexapodPhilosopher:

Adds new membrane textures, making different cell types more visually distinct.

* Currently includes all membranes except chitin.

**Related Issues**

#5582 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


assets/textures/CelluloseMembrane.png.import
```diff
@@ -0,0 +1,34 @@
...
+
+[params]
+
+compress/mode=0
+compress/high_quality=false
+compress/lossy_quality=0.7
+compress/hdr_compression=1
+compress/normal_map=0
+compress/channel_pack=0
+mipmaps/generate=false
```
hhyyrylainen, line 24: 3D textures should have mipmaps enabled for them, even though this is less critical than for 2D GUI images, I think it is still required ot make sure the texture looks good at small sizes (for example when viewing from far away).

shaders/Membrane.gdshader
```diff
@@ -63,9 +64,18 @@ float fresnel(float power, vec3 normal, vec3 view)
...
 
-    vec4 normalmap = texture(normalTexture, UV2);
-    NORMAL_MAP = normalmap.xyz;
-    NORMAL_MAP_DEPTH = 0.5f;
+    vec3 normalmap1 = texture(normalTexture, UV).rgb;
+    vec3 normalmap2 = texture(wiggleNormalTexture, UV).rgb;
+    float blend_amount = 0.45f;
+
+    vec3 a = normalmap1*2.0 - 1.0;
+    vec3 b = normalmap2*2.0 - 1.0;
```
hhyyrylainen, line 72:
```suggestion
    vec3 a = normalmap1 * 2.0f - 1.0f;
    vec3 b = normalmap2 * 2.0f - 1.0f;
```
Spaces should be used around operators to improve readability. Also shaders should really work with floats so numbers should be written preferrably with the `f` suffix (though I'm not always in the mood to nitpick about this).

shaders/Membrane.gdshader
```diff
@@ -10,9 +10,10 @@ uniform sampler2D albedoTexture : source_color;
 uniform sampler2D damagedTexture : source_color;
 
 uniform sampler2D normalTexture;
+uniform sampler2D wiggleNormalTexture;
```
hhyyrylainen, line 13: Why is this called "wiggleNormal"? As far as I can see there isn't a difference between the first normal map and the second in terms of wiggle (both seem to be now affected by it). Is there a typo perhaps or did you just mean to have two normal maps? In the latter case I think the variable needs renaming (and the relevant variables in the membrane C# code).

hhyyrylainen (changes requested):
I could have fixed the minor style problems myself but I noticed this one major thing I don't understand so I have to ask you to clarify / do some changes on this.

assets/textures/CelluloseMembrane.png.import
HexapodPhilosopher: Okay. I was trying to get these to match the pre-existing ones, which had mipmaps disabled.

shaders/Membrane.gdshader
HexapodPhilosopher: The idea was to have them move separately, although I'm not sure there's a need for it. I'll look at making it move relative to the other normal. The alternative would be reverting most of the shader changes and making the new normals more bumpy.

hhyyrylainen: If you didn't mean to break a fix by @dligr the normal bump map is supposed to use UV**2** and not just the first UV. So if you redo that change to preserve that, then this would all make sense as the second bump map would be affected by the wiggles and the first one wouldn't be.

hhyyrylainen:
This PR is waiting for those couple of technical fixes.

shaders/Membrane.gdshader
HexapodPhilosopher: I've experimented with this, and I just don't see any difference in using the UV2. The shader was scaling the UV2 and applying the normal map on the basis of that. The other effects on the normal were based on the value of the normal map as applied, and it makes no difference if it's UV1 or UV2.
What I've now done is removed the second normal map, replaced the normal map texture with one that includes any extra bumpiness required, and applied it via UV1. The effect is the same whichever channel I use, so I might as well use the same one as the albedo.

dligr: The way UV2 was calculated is actually not scaling, but rather something like a projection from a flat plane. UV1 is different in that it's kind of wrapped around the cell.

You can check the difference on larget cells, where bumps would be more streched, while with UV2 they would repeat.

HexapodPhilosopher: Okay, I see your point. I hadn't fully compared larger and smaller cells. I'll have to change it back to split normals.
What's the reason for setting the UV2 scale in vertex()?

assets/textures/CarbonateMembrane.png.import
hhyyrylainen, line 1: I have one question: why are the textures compressed losslessly? This will eat up so, so much more VRAM (that basically no games allow ordinary textures in 3D scenes to be in lossless encoding).

hhyyrylainen (approved):
I fixed the membrane textures to be in VRAM mode (and compressed) so that this can be merged now as this has been open for a while.

assets/textures/CarbonateMembrane.png.import
HexapodPhilosopher: I'm really not sure how I ended up doing them that way. While experimenting with how to get everything rendering properly, I undid changes multiple times. I then forgot to check for this in the import files. I'd also previously adjusted the shader variable names to fit the convention, which I then undid. I'll have to be more systematic with commits in future.

hhyyrylainen: Well with commits luckily it is always possible to go back / compare with an older version that worked better if further changes turn out to be the wrong direction to go.

PR #5706, opened 2024-11-24
Mucocyst use prototype

### LINKED ISSUE #5290: Update microbe AI to know how to use mucocyst
so that the player also needs to adapt to trying to attack something that has a mucocyst

Patryk26g:

Allows AI to use mucus to defend itself. The pr is still WIP.

**Related Issues**

closes #5290 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/components/MicrobeControl.cs
hhyyrylainen, line 1: I see this is still marked as a draft but can I ask why you are adding these two new variables? Mucus shouldn't need a secretion time as it is a toggle ability if I remember right. And it doesn't have a cooldown as again if I remember right the way it works is that it eats up some amount of mucilage before it can activate.

Patryk26g: Well I thought that it eats some mucus in time. If I get it correctly in the toggle version it would be able to stay in mucus layer indefinately if endangered by a predator? That might be a better solution tbh, I didn't think about that

hhyyrylainen: MucocystSystem handles consuming mucocyst and disabling the mode when running out. And there's a method `SetMucocystState` in the state helpers that ensures mucocyst can't be re-enabled if not enough mucilage is stored.

src/microbe_stage/systems/MicrobeEmissionSystem.cs
```diff
@@ -345,11 +356,61 @@ private void HandleSlimeSecretion(in Entity entity, ref MicrobeControl control,
...
+            }
         }
 
-        control.QueuedSlimeSecretionTime -= delta;
-        if (control.QueuedSlimeSecretionTime < 0)
-            control.QueuedSlimeSecretionTime = 0;
+        DecreaseTimeCountdownValue(ref control.QueuedSlimeSecretionTime, delta);
+    }
+
+    private void HandleMucusSecretion(in Entity entity, ref MicrobeControl control,
```
hhyyrylainen, line 367: What's this code about? It seems like you are partially reimplementing the MucocystSystem here, which is not where this code belongs as there's a dedicated system handling mucocyst operations.

src/microbe_stage/components/OrganelleContainer.cs
```diff
@@ -640,7 +640,11 @@ public static void FetchLayoutOrganelleComponents(this ref OrganelleContainer co
             {
                 if (organelleComponent is SlimeJetComponent slimeJetComponent)
                 {
-                    if (!slimeJetComponent.IsMucocyst)
+                    if (slimeJetComponent.IsMucocyst)
+                    {
+                        ++container.MucocystCount;
```
hhyyrylainen, line 645: Isn't this updated from somewhere else? Because if this code is now newly added wouldn't that mean that mucocysts couldn't have functioned before, but they did (`CalculateOrganelleLayoutStatistics`)?

src/microbe_stage/systems/MicrobeEmissionSystem.cs
```diff
@@ -82,15 +82,8 @@ protected override void Update(float delta, in Entity entity)
     {
         ref var control = ref entity.Get<MicrobeControl>();
 
-        // Reduce agent emission cooldown
-        // TODO: is it faster to check than to just blindly always subtract delta here?
```
hhyyrylainen, line 86: I think this comment should be copied to `DecreaseTimeCountdownValue`.

src/microbe_stage/systems/MicrobeEmissionSystem.cs
```diff
@@ -117,7 +110,6 @@ protected override void Update(float delta, in Entity entity)
             control.QueuedSiderophoreToEmit = false;
         }
 
-        // This method itself checks for the preconditions on emitting slime
```
hhyyrylainen, line 120: Why was this comment removed?

src/microbe_stage/systems/MicrobeEmissionSystem.cs
```diff
@@ -348,8 +340,16 @@ private void HandleSlimeSecretion(in Entity entity, ref MicrobeControl control,
...
 
-        control.QueuedSlimeSecretionTime -= delta;
-        if (control.QueuedSlimeSecretionTime < 0)
-            control.QueuedSlimeSecretionTime = 0;
+        DecreaseTimeCountdownValue(ref control.QueuedSlimeSecretionTime, delta);
+    }
+
+    private void DecreaseTimeCountdownValue(ref float countdownValue, float delta)
+    {
+        if (countdownValue > 0)
```
hhyyrylainen, line 348: Is this if actually good for performance? Checking branches is pretty slow in terms of CPU speed, so it might be faster to always just decrement the value and clamp it to 0 if it is negative.

src/microbe_stage/systems/MicrobeEmissionSystem.cs
```diff
@@ -348,8 +340,16 @@ private void HandleSlimeSecretion(in Entity entity, ref MicrobeControl control,
                 jet.Active = false;
         }
 
-        control.QueuedSlimeSecretionTime -= delta;
-        if (control.QueuedSlimeSecretionTime < 0)
-            control.QueuedSlimeSecretionTime = 0;
+        DecreaseTimeCountdownValue(ref control.QueuedSlimeSecretionTime, delta);
+    }
+
+    private void DecreaseTimeCountdownValue(ref float countdownValue, float delta)
```
hhyyrylainen, line 346: I think this should have MethodImpl aggressive inlining set as this is basically de-inlining this method which could reduce performance quite a bit due to the extra method calls to do one small thing.

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -307,16 +309,28 @@ private void ChooseActions(in Entity entity, ref MicrobeAI ai, ref CompoundAbsor
...
+        if (predator.HasValue && position.Position.DistanceSquaredTo(predator.Value.Position) <
             1500.0 * speciesFear / Constants.MAX_SPECIES_FEAR)
         {
-            FleeFromPredators(ref position, ref ai, ref control, ref organelles, compounds, entity, predator.Value,
-                speciesFocus, speciesActivity, speciesAggression, speciesFear, strain, random);
+            // If microbe secretes mucus and predator is still there skip taking any action
+            if (control.State == MicrobeState.MucocystShield)
+                return;
+
+            // MicrobeState predatorState = predator.Value.Entity.Get<MicrobeControl>().State;
```
hhyyrylainen, line 321: Commented out code should be removed.

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -773,25 +787,42 @@ private void PursueAndConsumeChunks(ref WorldPosition position, ref MicrobeAI ai
...
     {
+        var ourCompounds = compoundStorage.Compounds;
         control.SetStateColonyAware(entity, MicrobeState.Normal);
+        MicrobeState predatorState = predatorEntity.Get<MicrobeControl>().State;
 
         ai.TargetPosition = (2 * (position.Position - predatorLocation)) + position.Position;
 
         control.LookAtPoint = ai.TargetPosition;
 
+        // Shouldn't this distance value scale with predator's and prey's cell radius?
```
hhyyrylainen, line 802:
```suggestion
        // TODO: shouldn't this distance value scale with predator's and prey's cell radius?
```

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -773,25 +787,42 @@ private void PursueAndConsumeChunks(ref WorldPosition position, ref MicrobeAI ai
...
             bool shouldSprint = true;
+
             if ((organelles.SlimeJets?.Count ?? 0) > 0 &&
                 RollCheck(speciesFear, Constants.MAX_SPECIES_FEAR, random))
             {
                 // There's a chance to jet away if we can
                 control.SecreteSlimeForSomeTime(ref organelles, random);
             }
+            else if (ourCompounds.GetCompoundAmount(Compound.Mucilage) > Constants.MUCOCYST_MINIMUM_MUCILAGE
+                     && organelles.MucocystCount > 0 && (strain >= Constants.MAX_STRAIN_PER_ENTITY * 0.70
```
hhyyrylainen, line 815: The preferred style in Thrive is to have the `&&` and `||` at the end of lines rather than at the beginning of lines.

hhyyrylainen (changes requested):
The overall architecture is now in good shape I think. I have a few technical comments about some of the changes but all of them are minor. I haven't checked in game yet but it seems the logic change makes sense and should allow the AI to use mucocyst defensively.

src/microbe_stage/components/OrganelleContainer.cs
Patryk26g: There are 2 functions: `CalculateOrganelleLayoutStatistics` and  `FetchLayoutOrganelleComponents`. It was updated in the first one but someone forgot to add it in the second one, so many microbes didn't know they had mucocyst until their organelles split. We were able to use the mucocyst as the player because the first function was called upon leaving the editor

hhyyrylainen: I'm almost certain that cannot be the case, because that would mean that a bunch of other stuff like toxin vacuoles would be bugged similarly. `MicrobeReproductionSystem` calls `OnOrganellesChanged` which in turn calls `CalculateOrganelleLayoutStatistics`.

Patryk26g: Yes, sorry, of course I meant `MucocystCount` instead of `isMucocyst`. And you're right about counting it twice, i need to add `container.MucocystCount = 0;` at the beggining of the `FetchLayoutOrganelleComponents`

src/microbe_stage/systems/MicrobeEmissionSystem.cs
Patryk26g: Yeah your right, it's quite difficult to determine which one will be faster. I removed the first condition branch

src/microbe_stage/components/OrganelleContainer.cs
hhyyrylainen: I did some testing and I'm nearly certain that this change is unnecessary. I put breakpoints and checked that `CalculateOrganelleLayoutStatistics` already sets the mucocyst count, and when the code for running `FetchLayoutOrganelleComponents` is triggered I see this:
Kuvakaappaus - 2024-11-29 12-43-52

So the count is already set and your changed code just calculates it again, which is unnecessary.

So either the current code without these changes is correct or you've found a rare bug I was unable to trigger. So please re-check this. I've now checked the rest of the PR and I think it's fine for merge after the changes to OrganelleContainer.cs are finished.

hhyyrylainen (changes requested):

src/microbe_stage/components/OrganelleContainer.cs
Patryk26g: I also did some testing. I put a print into the ChooseAction function to show number of mucocyst of the species that do have it. At the start of the game (first screenshot) all of them show 0 (some of them show 2 because their mucocysts split). After running the game for a longer time most species finally "understand" that they have mucocyst (second screenshot).
image
image

With my changes all of these cells know that they have mucocysts at the start (all of them show 1).
I also debugged the same way as you did. This function is called in the loading screen and all of them showed 0. When the game loaded only then did this value show a different number than 0.

Anyway in the `CalculateOrganelleLayoutStatistics` there is a line `container.OrganelleComponentsCached = false;` which makes it so that the `FetchLayoutOrganelleComponents` is called again... so maybe we should either remove it or move all organelles caching from `CalculateOrganelleLayoutStatistics` to `FetchLayoutOrganelleComponents` to reduce the number of recalculations

hhyyrylainen: I think you are seeing the species gain the mucocyst (so it at first doesn't have it but then it gets it and as old cell entities are despawned and freshly spawned ones replace them the mucocyst count goes up). I've debugged more today and I'm still convinced the existing code is correct. To verify, I started from master and added the following code to `MicrobeAIASystem.Update` and played the game for many minutes with the mucocyst:

```c#
        var species = (MicrobeSpecies)entity.Get<SpeciesMember>().Species;
        var mucocystCount = entity.Get<OrganelleContainer>().MucocystCount;
        var slime = SimulationParameters.Instance.GetOrganelleType("slimeJet");

        var speciesMucocystCount = species.Organelles.Count(o =>
            o.Definition.InternalName == slime.InternalName && o.Upgrades != null &&
            o.Upgrades.UnlockedFeatures.Contains("mucocyst"));

        if (mucocystCount < speciesMucocystCount)
        {
            GD.PrintErr($"Found entity with incorrect mucocyst count: {entity}");
            Debugger.Break();
        }
```

And I didn't hit the debugger break (or the print) so I'm very certain that mucocyst count calculation is correct. I did also verify that the species mucocyst count does go above 1 so that my testing code wasn't actually the bugged one. If you can show the problem appearing with that debugging code I wrote or if I have a bug in the debugging code then I'll believe the mucocyst count code has an issue. Though I think my testing code also has a potential "bug" where a species that gains a mucocyst while there are still old cells swimming around with the previous bodyplan. So I think it can report false positives, but it shouldn't be able to report a false negative.

Patryk26g: Okay, probably you're right about the spawning and despawning then, I will remove this mucocyst count update :)

hhyyrylainen (approved):
The code change is now good. I haven't really seen in my playtesting the AI using the mucocyst but I'll assume that part works at least a bit and merge this (I'm always low on time to test complex PRs that would take really long to verify the gameplay).

PR #5710, opened 2024-11-26
Added disk caching for generated images

### LINKED ISSUE #3156: Create a cache that is stored on disk for PhotoStudio

hhyyrylainen:

Adds an on-disk cache in addition to memory caching to make caching more efficient

I haven't done final testing but I'll mark this as ready for review

**Related Issues**

closes #3156 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/HexEncoding.cs
```diff
@@ -0,0 +1,179 @@
...
+
+    /// <summary>
+    ///   Decodes hex encoded UTF8 text in place to raw bytes
+    /// </summary>
+    /// <param name="data">Input and the result</param>
+    /// <param name="length">How many bytes the input is</param>
+    /// <param name="outBytes">How many bytes have been written</param>
+    /// <returns>False if invalid characters are encountered (or input is empty)</returns>
+    public static bool DecodeFromHexInPlace(byte[] data, int length, out int outBytes)
+    {
```
dligr, line 54: With prints, I determined that this function fails when reading a ':' symbol, probably at the start of the path.
It seems like this function isn't intended to read a full path, so maybe there should be a `startIndex` argument here?

dligr (commented):
Testing on Windows, I get errors like this:
`Failed to decode calculate hash from path: C:/Users/[user name]/AppData/Roaming/Thrive/cache/img\7\dd5105b15525c2e.png`

src/general/HexEncoding.cs
hhyyrylainen: I found one problem preventing things from working on Linux and have now fixed that (there was one extra character that needed to be removed from the prefix). Could you re-test this on Windows as I don't have time today to do that myself?

PR #5712, opened 2024-11-28
Fix non-lawk unlocks showing up after loading a save

### LINKED ISSUE #5517: Loading a save causes LAWK disabled organelles to appear in list of possible unlocks
Steps to reproduce:
- Start a game with LAWK on
- Go to the editor, look at the list of organelle unlocks, there is no thermoplast
- Make a save
- Load that save
- Look at the list again, thermoplast has appeared (but funnily enough the prokaryotic variant seems to not appear)

I didn't test if this actually allows unlocking them, which would be a bigger bug.

Patryk26g:

Hide non-lawk unlocks when non-lawk is enabled

**Related Issues**

Closes #5517

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Could you update the PR title to be more clear what this fixes? Just "unlocks bugfix" doesn't really specify anything this does so it is not a descriptive title.

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -370,6 +370,9 @@ private void CreateUndiscoveredOrganellesButtons(bool refresh = false, bool auto
 
         var worldAndPlayerArgs = GetUnlockPlayerDataSource();
 
+        // Apply LAWK settings so that no-unexpected organelles are shown
+        UpdateOrganelleLAWKSettings();
```
hhyyrylainen, line 374: If this change works, I think line 387 comment needs to be deleted so that there's no outdated comment left in the code.

hhyyrylainen: Upon playtesting I discovered that the below comment is indeed important, as the following steps lead to a bug:

- Start a new non-LAWK game
- Get to the editor
- Use the cheat to unlock all organelles
- Exit the editor and re-enter it
- Notice how the thermoplast organelle can now be placed even though LAWK is still on

So I think the code needs to be tweaked. I think easiest thing is to revert this change and instead modify the loop below to take LAWK into account and not adding something to the unlock conditions text when LAWK mode wouldn't allow it.

hhyyrylainen (changes requested):

hhyyrylainen (approved):
I simplified this fix quite a bit. I think this still does the same thing so I'll merge once the CI checks succeed.

PR #5713, opened 2024-11-28
Rigidity slider tooltip fix (to hide osmoregulation)

### LINKED ISSUE #5635: Rigidity slider tooltip shouldn't show osmoregulation change (as it doesn't affect)
A slight regression in 0.7.1

So the bottom stat here should just be set invisible:

kuva

Patryk26g:

Bugfix

**Related Issues**

Closes #5635

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (changes requested):
I think that instead this should be done with the `ShowOsmoregulation` export variable which I explicitly added for the functionality of hiding osmoregulation in specific tooltips (this is already used for the upgrade tooltips, though those are dynamically generated so a bit different) . Or is there a reason why that isn't usable in this case?

hhyyrylainen (approved):
Code change looks good and I tested quickly in-game and this seems to work perfectly. Thanks for the fix.

PR #5714, opened 2024-11-28
Stop checking equations constantly

### LINKED ISSUE #5590: Organelle tooltips shouldn't have ChemicalEquation `AutoRefreshProcess` on
as the numbers only change when switching patches, so this unnecessarily eats up performance in the editor

Patryk26g:

This PR does some stuff...

**Related Issues**

Closes #5590

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Instead of letting the placeholder text "This PR does some stuff..." stay in the PR template, could you please write a very short summary of the changes you did? For examples you can look at my recent PRs.

src/gui_common/ChemicalEquation.cs
```diff
@@ -143,7 +143,7 @@ public bool ShowFullSecondText
     /// <summary>
     ///   If true this will automatically check the set process for changes
     /// </summary>
-    public bool AutoRefreshProcess { get; set; } = true;
+    public static bool AutoRefreshProcess { get; set; } = false;
```
hhyyrylainen, line 146: Why is this static? This means that *all* chemical equations in the game share this, including the process panel.

src/gui_common/ChemicalEquation.cs
```diff
@@ -205,6 +208,7 @@ private void UpdateSecondLabel()
 
     private void UpdateEquation()
     {
+        GD.Print("Updating equation");
```
hhyyrylainen, line 211: Debugging code

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -175,6 +175,7 @@ public Patch? CurrentPatch
         {
             currentPatch = value;
             playerHere.Visible = CurrentPatch == SelectedPatch;
+            ChemicalEquation.AutoRefreshProcess = true;
```
hhyyrylainen, line 178: This looks like it just sets the auto refresh to be on when this value is written. Instead I think what needs to be done is triggering the refresh when editor selected patch or current patch has changed.

src/gui_common/ChemicalEquation.cs
```diff
@@ -182,7 +182,10 @@ public override void _Process(double delta)
         }
 
         if (AutoRefreshProcess)
+        {
+            AutoRefreshProcess = false;
```
hhyyrylainen, line 186: As this is static this will cause only a *single* chemical equation to refresh which will then set this to false and not allow any of the other instances to see this variable being true.

hhyyrylainen (changes requested):
I don't think this can work correctly implemented like this.

I imagined more like there'd be a method in ChemicalEquation that says it should refresh once, then that is exposed through wrapper methods all the way up to the tooltip level, and then the editor would have a method to tell all the tooltips they need to refresh.

PR #5720, opened 2024-11-30
Add background blur

### LINKED ISSUE #5641: Add blur to the microbe background distortion shader effect (and enable by default)
The effect looks pretty good, but we agreed that it needs a bit of blur to really elevate the look

dligr:

Adds a Gaussian blur to microbe backgrounds (only enabled if distortion is on).

**Related Issues**

Closes #5641 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


dligr:
Some screenshots:
image
image
image
image

shaders/MicrobeBackground.gdshader
```diff
@@ -22,6 +22,8 @@ uniform float distortionSpeed : hint_range(0.0, 0.01, 0.0001) = 0.001f;
 
 uniform float layerAnimateSpeed : hint_range(0.0, 2.0, 0.1) = 1.0f;
 
+uniform vec2 worldPos;
```
hhyyrylainen, line 25: Why does world position now need to be passed in? couldn't the math extracting it from the view matrix be just altered?

shaders/Blur.gdshader
hhyyrylainen, line 1: If this is gaussian blur I think the shader file name should have "gaussian" in it.

src/microbe_stage/MicrobeCamera.tscn
```diff
@@ -48,12 +72,51 @@ current = true
...
+current = true
 
 [node name="BackgroundPlane" type="CSGMesh3D" parent="."]
-transform = Transform3D(1, 0, 0, 0, -1.62921e-07, -1, 0, 1, -1.62921e-07, 0, 5.96247e-08, -15)
+transform = Transform3D(1, 0, 0, 0, -1.62921e-07, -1, 0, 1, -1.62921e-07, 0, 0, -20)
+cast_shadow = 0
+mesh = SubResource("1")
+material = SubResource("ShaderMaterial_1sr76")
+
+[node name="SubViewport2" type="SubViewport" parent="."]
```
hhyyrylainen, line 88: Could this be implemented as a second render pass or directly baked into the background shader? I ask because we'll hit a rendering performance problem at some point if we just throw subviewports at any problem.

dligr: Actually, it seems possible to re-implement this without viewports. I've just found out that it's possible for a shader to get it's texture from a CanvasLayer instead of a viewport ([the docs](https://docs.godotengine.org/en/stable/tutorials/shaders/custom_postprocessing.html#multi-pass-post-processing)), which should be more performant.

shaders/MicrobeBackground.gdshader
hhyyrylainen: Oh yeah, I kind of suspected after getting to the subviewport. So that part needs to be resolved before talking about this part...

src/microbe_stage/MicrobeCamera.tscn
hhyyrylainen: In case that's not usable, I think putting the blur code directly into the background shader would be the right way to code. That would also quite neatly I think expose how expensive in terms of texture reads it is actually to perform a blur.

dligr: Putting the blur code into the background shader would be bad for performance, because it would then need n*n texture reads (for blur diameter of n), and for each of those reads it would also need to calculate distortion.
The current implementation does blur in 2 passes (vertical and horizontal), and because of that only needs 2*n reads. This difference in calculation costs should probably compensate for viewports' performance.

hhyyrylainen: Still it will now blur the moving bubble layers that are separate from the furthest background layer as well, which is probably not the best. Also it could maybe be possible to use a cheaper blur method combined into the distortion math to do a reasonable enough blur with high performance at the same time as applying the distortion math?

I guess I'll let you investigate first what is actually possible to do before dumping more ideas.

dligr: Apparently the way of making multi-pass shaders with CanvasLayers requires reading the screen texture, so there still needs to be at least one viewport. 
The background plane can also be optimized into a ColorRect (which needs a scale argument passed to it, but it's probably still more performant than rendering 3D geometry).

This is what the resulting scene hierarchy looks like:
image

assets/textures/blank.png.import
hhyyrylainen, line 1: This doesn't seem good as the blank texture is a png meant for GUI stuff. If it is not usable as-is a separate blank one for 3D needs to be added...

src/engine/Settings.cs
```diff
@@ -111,7 +111,21 @@ public enum StrainBarVisibility
     ///   enabled.
     /// </summary>
     [JsonProperty]
-    public SettingValue<float> MicrobeDistortionStrength { get; private set; } = new(0);
+    public SettingValue<float> MicrobeDistortionStrength { get; private set; } = new(0.6f);
+
+    /// <summary>
+    ///   The amount of blur applied to microbe backgrounds when distortion is on.
```
hhyyrylainen, line 117: This comment seems to be duplicated from the below variable. So maybe this should be removed and blur strength of 0 should mean off?

src/general/OptionsMenu.cs
```diff
@@ -853,6 +859,9 @@ public void ApplySettingsToControls(Settings settings)
         bloomEffectToggle.ButtonPressed = settings.BloomEnabled;
         bloomSlider.Value = settings.BloomStrength;
         bloomSlider.Editable = settings.BloomEnabled || !DisableInactiveSliders;
+        blurEffectToggle.ButtonPressed = settings.MicrobeBackgroundBlurEnabled;
+        blurSlider.Value = settings.MicrobeBackgroundBlurStrength;
+        blurSlider.Editable = settings.MicrobeBackgroundBlurEnabled || !DisableInactiveSliders;
```
hhyyrylainen, line 864: Disabled sliders cause issues with the focus flow (as we currently have no visual way disabled sliders show they have focus), so based on my previous comment I kind of recommend we'd just have a single blur strength slider that when dragged to 0 acts as the disabled state.

src/microbe_stage/MicrobeCamera.cs
```diff
@@ -94,6 +77,11 @@ public partial class MicrobeCamera : Camera3D, IGodotEarlyNodeResolve, ISaveLoad
     [JsonProperty]
     private float lightLevel = 1.0f;
 
+#pragma warning disable CA2213
+    [JsonIgnore]
+    private BackgroundPlane backgroundPlane = null!;
```
hhyyrylainen, line 82: I prefer putting the Godot variables before the "normal" variables so I'd prefer if this bit was kept around line 65 like where ` private Node3D? backgroundPlane;` was previously.

src/microbe_stage/MicrobeCamera.cs
```diff
@@ -164,22 +152,14 @@ public Vector3 CursorVisualWorldPos
...
     {
-        var material = GetNode<CsgMesh3D>("BackgroundPlane").Material;
-        if (material == null)
-        {
-            GD.PrintErr("MicrobeCamera didn't find material to update");
-            return;
-        }
-
-        materialToUpdate = (ShaderMaterial)material;
+        backgroundPlane = GetNode<BackgroundPlane>("BackgroundPlane");
```
hhyyrylainen, line 155: I think it would be better to make the backgroundPlane variable an `[Export]` and set it directly in the editor than like this.

src/microbe_stage/MicrobeCamera.cs
```diff
@@ -223,13 +192,15 @@ public override void _Process(double delta)
...
-            UpdateLightLevel((float)delta);
+            backgroundPlane.UpdateLightLevel((float)delta);
         }
 
         if (AutoProcessWhilePaused && PauseManager.Instance.Paused)
         {
             UpdateCameraPosition(delta, null);
         }
+
+        backgroundPlane.SetWorldPosition(new Vector2(Position.X, Position.Z));
```
hhyyrylainen, line 203: Shouldn't the background plane handle this or? (so it can just look at its own position to update the shader parameters)

src/microbe_stage/MicrobeCamera.cs
```diff
@@ -330,61 +299,46 @@ public bool Zoom(double delta, float value)
...
+        backgroundPlane.SetBackground(background);
     }
 
     protected override void Dispose(bool disposing)
     {
         if (disposing)
         {
-            lightLevelParameter.Dispose();
-            distortionStrengthParameter.Dispose();
+            // Nothing to dispose
```
hhyyrylainen, line 309: This method should be removed entirely if there's nothing to do here.

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,263 @@
...
+public partial class BackgroundPlane : Node3D, IGodotEarlyNodeResolve
+{
+    private readonly StringName blurAmountParameter = new("blurAmount");
+    private readonly StringName textureAlbedoParameter = new("textureAlbedo");
+    private readonly StringName worldPositionParameter = new("worldPos");
+    private readonly StringName lightLevelParameter = new("lightLevel");
+    private readonly StringName distortionStrengthParameter = new("distortionFactor");
+
+    [Export]
+    private NodePath? blurPlanePath;
```
hhyyrylainen, line 16: With Godot4 we switched to the style where nodepaths aren't used anymore but direct node setting (there is still a lot of old code not updated but I enforce that proper usage in new code).

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,263 @@
+using System;
+using Godot;
+
+/// <summary>
+///   Manages the microbe background plane, optionally applies blur.
+/// </summary>
+public partial class BackgroundPlane : Node3D, IGodotEarlyNodeResolve
```
hhyyrylainen, line 7: Why does this use `IGodotEarlyNodeResolve` as this shouldn't be able to be loaded from saves due to being ignored in the camera class? As a result any saved state wouldn't work in this class whatsoever so this needs to rely on the camera to remember and re-set everything after load (as this is currently setup).

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,263 @@
...
+    private ShaderMaterial currentBackgroundMaterial = null!;
+
+    private SubViewport subViewport1 = null!;
+
+    private SubViewport subViewport2 = null!;
+#pragma warning restore CA2213
+
+    public bool NodeReferencesResolved { get; private set; }
+
+    public float PlaneOffset
```
hhyyrylainen, line 48: Why is this not the direct offset but 15 is added on top? Isn't this quite unexpected, like for a variable like "plane offset" I'd expect it to set the plane offset, not plane offset with extra modifier on top...

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,263 @@
...
+
+    public override void _Ready()
+    {
+        var material = GetNode<CsgMesh3D>(backgroundPlanePath).Material;
+        var planeBlurMaterial = GetNode<CsgMesh3D>(blurPlanePath).Material;
+        var colorRectBlurMaterial = GetNode<CanvasItem>(blurColorRectPath).Material;
+
+        if (material == null || planeBlurMaterial == null || colorRectBlurMaterial == null)
+        {
+            GD.PrintErr("MicrobeCamera didn't find material to update");
```
hhyyrylainen, line 68: The classname in the error is no longer accurate.

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,263 @@
...
+        NodeReferencesResolved = true;
+
+        if (HasNode(backgroundPlanePath))
+            backgroundPlane = GetNode<Node3D>(backgroundPlanePath);
+
+        if (HasNode(blurPlanePath))
+            blurPlane = GetNode<Node3D>(blurPlanePath);
+
+        if (HasNode("SubViewport"))
+            subViewport1 = GetNode<SubViewport>("SubViewport");
```
hhyyrylainen, line 96: These should also be done with direct export node references. Also what would be the use case where the subviewports are allowed to be missing?

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        base.Dispose(disposing);
+    }
+
+    private void OnDisplayBackgroundParticlesChanged(bool displayed)
+    {
+        if (backgroundParticles == null)
+        {
+            GD.PrintErr("MicrobeCamera didn't find background particles on settings change");
```
hhyyrylainen, line 191: Previous class name here as well.

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,263 @@
...
+    {
+        ApplyBlurEffect();
+    }
+
+    private void ApplyBlurEffect()
+    {
+        float blurStrength = Settings.Instance.MicrobeBackgroundBlurStrength;
+        bool enabled = blurStrength > 0 && Settings.Instance.MicrobeBackgroundBlurEnabled.Value;
+
+        if (enabled)
```
hhyyrylainen, line 233: This seems relatively expensive operation, so this should remember the previous state and not re-apply if currently done.

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,263 @@
...
+        {
+            blurPlane.Visible = true;
+
+            canvasBlurMaterial.SetShaderParameter(blurAmountParameter, blurStrength);
+            spatialBlurMaterial.SetShaderParameter(blurAmountParameter, blurStrength);
+
+            if (backgroundPlane.GetParent() == this)
```
hhyyrylainen, line 240: Rather than doing this I think there should be a bool flag to remember what was done to avoid this check.

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,263 @@
...
+            {
+                RemoveChild(backgroundPlane);
+                subViewport1.AddChild(backgroundPlane);
+            }
+
+            subViewport1.RenderTargetUpdateMode = SubViewport.UpdateMode.WhenVisible;
```
hhyyrylainen, line 246: I think disabling the update entirely and force enabling it when wanted would be better than this as that would logically avoid a check the engine needs to do to see if the viewport is actually visible in the scene tree each frame.

assets/textures/blank.png.import
hhyyrylainen, line 1: This is still not in the old GUI icon texture settings...

src/microbe_stage/MicrobeCamera.cs
```diff
@@ -269,9 +237,7 @@ public void UpdateCameraPosition(double delta, Vector3? followedObject)
 
         if (backgroundPlane != null)
```
hhyyrylainen, line 238: `backgroundPlane` cannot be null so this if condition should be removed (I'll be surprised a bit if the CI didn't catch this).

src/microbe_stage/BackgroundPlane.tscn
```diff
@@ -0,0 +1,114 @@
...
+subViewport1 = NodePath("SubViewport")
+subViewport2 = NodePath("SubViewport2")
+
+[node name="BlurPlane" type="CSGMesh3D" parent="."]
+transform = Transform3D(1, 0, 0, 0, -1.62921e-07, -1, 0, 1, -1.62921e-07, 0, 0, -20)
+cast_shadow = 0
+mesh = SubResource("1")
+material = SubResource("ShaderMaterial_1sr76")
+
+[node name="SubViewport2" type="SubViewport" parent="."]
```
hhyyrylainen, line 82: I think these viewports should be set to not render by default in the scene just as a precaution if the C# script fails to run correctly that would protect against a slight performance hit.

src/engine/Settings.cs
```diff
@@ -111,7 +111,14 @@ public enum StrainBarVisibility
...
+    ///   enabled.
+    /// </summary>
+    [JsonProperty]
+    public SettingValue<float> MicrobeBackgroundBlurStrength { get; private set; } = new(3.0f);
```
hhyyrylainen, line 121: This value is more than the max value set in `src/general/OptionsMenu.tscn`

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,249 @@
...
+    private ShaderMaterial currentBackgroundMaterial = null!;
+
+    private ShaderMaterial spatialBlurMaterial = null!;
+
+    private ShaderMaterial canvasBlurMaterial = null!;
+#pragma warning restore CA2213
+
+    private bool blurEnabledLastTime = true;
+
+    public bool NodeReferencesResolved { get; private set; }
```
hhyyrylainen, line 46: With this no longer implementing the interface this property seems unnecessary.

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,249 @@
...
+        for (int i = 0; i < 4; ++i)
+        {
+            // TODO: switch this loop away to reuse StringName instances if this causes significant allocations
+            currentBackgroundMaterial.SetShaderParameter($"layer{i:n0}", GD.Load<Texture2D>(background.Textures[i]));
+        }
+
+        backgroundParticles?.DetachAndQueueFree();
+
+        backgroundParticles = background.ParticleEffectScene.Instantiate<GpuParticles3D>();
+        backgroundParticles.Rotation = Rotation;
```
hhyyrylainen, line 123: Is this rotation setting new or why does this not seem familiar to me?

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,249 @@
...
+        currentBackgroundMaterial.SetShaderParameter(distortionStrengthParameter,
+            Settings.Instance.MicrobeDistortionStrength.Value);
+    }
+
+    private void OnBackgroundBlurStrengthChanged(float value)
+    {
+        ApplyBlurEffect();
+    }
+
+    private void OnBackgroundBlurToggleChanged(bool value)
```
hhyyrylainen, line 199: This seems like an unused method now.

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,249 @@
...
+    {
+        ApplyBlurEffect();
+    }
+
+    private void ApplyBlurEffect()
+    {
+        float blurStrength = Settings.Instance.MicrobeBackgroundBlurStrength;
+        bool enabled = blurStrength > 0;
+
+        if (blurEnabledLastTime == enabled)
```
hhyyrylainen, line 209: Doesn't this work incorrectly if blur strength is changed from 1.0 to 1.1 (i.e. it would early return and not apply the strength)?

hhyyrylainen: I meant to add that I suspect that due to the code in `OnBackgroundBlurStrengthChanged`

src/microbe_stage/BackgroundPlane.cs
hhyyrylainen: Nevermind, this is not new code, just moved.

src/microbe_stage/MicrobeCamera.cs
dligr: Apparently the background plane's world position doesn't change when it's in a subviewport, so the camera has to handle that

dligr: Actually, I've just realized that the position of the `BackgroundPlane` does change, just not relative to the camera. So, `GlobalPosition` has to be used instead of `Position`. Otherwise, the implementation that you've described is in fact what already is done.

hhyyrylainen: Ah of course. I didn't notice you had made that mistake. It took me a while but I've almost entirely stopped making the mistake of using Position / WorldPosition when I should have used the other one.

src/engine/Settings.cs
```diff
@@ -111,7 +111,14 @@ public enum StrainBarVisibility
     ///   enabled.
     /// </summary>
     [JsonProperty]
-    public SettingValue<float> MicrobeDistortionStrength { get; private set; } = new(0);
+    public SettingValue<float> MicrobeDistortionStrength { get; private set; } = new(0.6f);
+
+    /// <summary>
+    ///   The amount of blur applied to microbe backgrounds when distortion is on.
+    ///   enabled.
+    /// </summary>
```
hhyyrylainen, line 119:
```suggestion
    ///   The amount of blur applied to microbe backgrounds when distortion is on.
```

I think that was a typo in the comment.

src/microbe_stage/MicrobeCamera.cs
```diff
@@ -292,7 +255,7 @@ public void ResetHeight()
     public void SetCustomCurrentStatus(bool current)
     {
         Current = current;
-        UpdateBackgroundVisibility();
+        backgroundPlane.SetVisibility(Current);
```
hhyyrylainen, line 258:
```suggestion
        UpdateBackgroundVisibility();
```

Why doesn't this just call that method that still exists and does the SetVisibility call?

src/microbe_stage/BackgroundPlane.tscn
```diff
@@ -0,0 +1,117 @@
...
+visible = false
+cast_shadow = 0
+mesh = SubResource("1")
+material = SubResource("ShaderMaterial_1sr76")
+
+[node name="SubViewport2" type="SubViewport" parent="."]
+disable_3d = true
+handle_input_locally = false
+size = Vector2i(1280, 720)
+render_target_clear_mode = 1
```
hhyyrylainen, line 93: This is still the wrong property, this is `Clear Mode` whereas the right thing is to set `Update Mode` to `Disabled`

So in this text view it should be `render_target_update_mode = 0` (and this clear mode should be left to default).

src/microbe_stage/BackgroundPlane.cs
hhyyrylainen: There still wasn't a bool flag added and there is still a GetParent call (which is almost always a sign of too much interconnectness in Godot, so an early sign of spaghetti forming). It's not super serious in this case but I see a very simple way to avoid it here so I'd like that approach to be taken.

src/microbe_stage/BackgroundPlane.tscn
```diff
@@ -0,0 +1,117 @@
...
+anchor_right = 0.5
+anchor_bottom = 0.5
+offset_left = -640.0
+offset_top = -360.0
+offset_right = 640.0
+offset_bottom = 360.0
+grow_horizontal = 2
+grow_vertical = 2
+
+[node name="SubViewport" type="SubViewport" parent="."]
```
hhyyrylainen, line 109: I think I finally understood the logic after thinking for a while, so for future understandability I would like to ask the viewports to be named something like `BlurPhase1` and `BlurPhase2` and maybe `BlurPlane` as `BlurResultDisplay` (if I understood what that was used for correctly).

src/microbe_stage/BackgroundPlane.cs
hhyyrylainen, line 1: Also one thing in general is that the subviewports default to 720p. So even when running the game at 4K the backgrounds are at 720p so they are extra blurred due to the resolution. I'm not sure if anything needs to be done about that but I think at least a big code comment should be added to explain this situation.

An example fix would be to just check the game window size in the _Process here and resize the viewport sizes to match the window size if changed.

src/engine/Settings.cs
```diff
@@ -111,7 +111,14 @@ public enum StrainBarVisibility
     ///   enabled.
     /// </summary>
     [JsonProperty]
-    public SettingValue<float> MicrobeDistortionStrength { get; private set; } = new(0);
+    public SettingValue<float> MicrobeDistortionStrength { get; private set; } = new(0.6f);
```
hhyyrylainen, line 114: As there's a new default value, there should be `[JsonProperty(PropertyName = "MicrobeDistortionStrengthV2")]` on this property to ensure all players get the new default value.

hhyyrylainen (changes requested):
On the code there's just a few small readability and style enhancements I'd like to see. I playtested this now and noticed that it seems like the background light level applying is broken. I'm here in epipelagic and despite waiting for basically a full day the background brightness never reaches the usual value:

kuva

Besides that one more minor point is that if the blur only applies when distortion is enabled the slider should be disabled when distortion is off to clearly indicate to the player what it does (and the checkbox should be moved immediately below the slider). Alternatively I think the blur could be made to also work with the normal backgrounds (though in this case I think the render resolution problem is a more important one). Also depending on how expensive the blur is when rendering at bigger size there might need to be a separate toggle about using low resolution blur on the backgrounds.

So yeah sorry to dump so much work still on you but I think this feature still needs polish before adding to the game.

src/microbe_stage/BackgroundPlane.cs
dligr: The problem with dynamically updating subviewports' resolution is that it causes screen artifacts to appear when scaling windows in real time, such as this:
image

src/microbe_stage/BackgroundPlane.tscn
dligr: I have additionally renamed the subviewports in code for more consistency

src/microbe_stage/BackgroundPlane.cs
hhyyrylainen: Could it work so that the viewport change is applied only once two frames the window size is the same? Or maybe there's like a 0.3 second interval that the resolution must be the same before switching (can be easily made with just a single double variable keeping time)?

src/microbe_stage/BackgroundPlane.tscn
hhyyrylainen: That's a good idea to match the code and the scene.

src/microbe_stage/BackgroundPlane.cs
dligr: Would it be a good idea to simply reuse the `blurEnabledLastTime` variable for that, since this hierarchy change is done along with all the other changes associated with `blurEnabledLastTime`?

hhyyrylainen: I mean yeah, it probably works to keep it all behind that one variable.

src/microbe_stage/BackgroundPlane.cs
dligr: It still results in a flicker, though it's better than continuous artifacts

hhyyrylainen: Interesting. I suppose maybe hiding the stuff entirely for the frame it is resized would remove the artifacting? But if that's worth the time to develop that kind of complex workaround, I'm not sure.

dligr: It seems like the background viewport is the one that causes those arifacts to appear, as rescaling only the blur viewport works fine. Could the problem be with rescaling specifically a 3D viewport?

hhyyrylainen: Doesn't the normal part of the game display fine though when resizing? So the issue is likely more complex with the different viewports depending on each other, which is likely the interaction that shows the non-optimal behaviour. Though of course  I can't say anything for certain without checking Godot open and closed issues to see if this is something seen by other people as well.

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -0,0 +1,229 @@
...
+        backgroundParticles = background.ParticleEffectScene.Instantiate<GpuParticles3D>();
+        backgroundParticles.Rotation = Rotation;
+        backgroundParticles.LocalCoords = false;
+
+        AddChild(backgroundParticles);
+
+        OnDisplayBackgroundParticlesChanged(Settings.Instance.DisplayBackgroundParticles);
+    }
+
+    public void SetWorldPosition(Vector2 position)
```
hhyyrylainen, line 122: I think this should be made private.

hhyyrylainen:
I had just that one minor comment about the code for now, so the code so far written now looks good to me. I didn't notice any of the viewport resizing etc. code or allowing blur even without distortion enabled, so I guess those aren't done yet? I think it would take a bit of crunching if this is to make it into 0.8.0 so it is fine if this needs more time in a relaxed manner to get done, which I'd also prefer as I'm still not fully done with the re-added mac support so it would be a pretty big squeeze to fit this into 0.8.0.

Deus-Codes:
Looks really nice. I think it might be worth it to increase background distortion speed now, as the movement of the background is even less obvious now. The blur should help alleviate worries of disorientation we had before, where the background might have moved too fast for the eye to be comfortable with.

This isn't relevant to this pull request, but in the future, it would be nice to create some sense of depth, as it can sometimes seem obvious that the background is the background and the plane the cell is on is the foreground.

hhyyrylainen:
Yeah that probably makes sense. Translation keys should still be clear and include the context that they are about a game background for translators to be able to translate correctly.

src/general/OptionsMenu.cs
```diff
@@ -853,6 +856,8 @@ public void ApplySettingsToControls(Settings settings)
         bloomEffectToggle.ButtonPressed = settings.BloomEnabled;
         bloomSlider.Value = settings.BloomStrength;
         bloomSlider.Editable = settings.BloomEnabled || !DisableInactiveSliders;
+        blurSlider.Value = settings.MicrobeBackgroundBlurStrength;
+        blurSlider.Editable = Settings.Instance.MicrobeDistortionStrength > 0 || !DisableInactiveSliders;
```
hhyyrylainen, line 860: Does the effect look really bad without distortion, or why was this ultimately left so that the blur cannot be enabled unless distortion is also on?

dligr: It's honestly debatable if blur should only be possible to enable with distortion. My reasoning was that the background were originally not designed to be used with blur, so they might look bad with it.
But if blur is to be applied to them without distortion, it's probably better to just switch the images to pre-blurred ones for more performance. The cost of that would be blur strength being unconfigurable, though.

hhyyrylainen: I'd argue the distortion already changes the original artistic vision, and I allowed that so with the same logic blur should be allowed. And as it would be just a tiny bit of code change to not lock out that option while distortion is off, I think it is worth it to allow it. And blurring the background images themselves would need a lot of artistic touch to do it well (it was tried once before and I really disliked the effect), and no one would be able to play with the old backgrounds if they preferred. So all in all I think it is just much superior option to allow to enable the blur separately from the distortion effect.

dligr: I meant that the pre-blurred images would only be shown when blur is on (otherwise the original backgrounds would remain); though it might feel weird that the slider doesn't control blur strength when distortion is off.

hhyyrylainen: I wouldn't want to bloat the game filesize a bunch just to have this feature. So I'm against including both variants of the backgrounds, which is why I think this should be a dynamically rendered effect.

src/microbe_stage/BackgroundPlane.cs
hhyyrylainen: Was this flicker thing solved? I see this as the last comment not marked as resolved...

hhyyrylainen (changes requested):
I've now again reviewed all of the code here and it all looks good to me. There's just the problem of the missing render target size update code. I think it would be good enough to update it like 5 times per second if the window size has changed and call that good enough so that this PR can finally get merged.

dligr:
I have added the resolution updating. I decided that it's better to update the resolution at 1 second intervals though, so that there is less artifact flickering.

hhyyrylainen (approved):
I fixed up one comment I saw that was outdated still when reading through the code one last time. I think the code is now perfect. I tested in game and it seems all is working fine (one thing I couldn't verify is that the viewports are not causing a rendering speed penalty when the settings are disabled, for performance on low end systems but that probably work).

PR #5721, opened 2024-12-01
Make engulfing info last in membrane tooltips

### LINKED ISSUE #5717: Membrane tooltips should have the engulfing info after osmoregulation
As a slight side effect of the tooltip refactoring, the osmoregulation stat is always last, which in the case of the membrane tooltips I don't think is optimal, as that looks like this:

So for that tooltip type it'd be nice to override the feature that osmoregulation is always last. I'm marking this intermediate as I expect a bit of care needing to be taken when designing a clean architecture way to add this exception.

Patryk26g:

Makes so that the ifno about engulfing is displayed after osmoregulation. It seems that just changing the ordering was enough and it did not need any other refactoring
image

**Related Issues**

Closes #5717

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/tooltip/ToolTipManager.tscn
+++ b/src/gui_common/tooltip/ToolTipManager.tscn
@@ -349,13 +349,13 @@ layout_mode = 2
 DisplayName = "PHYSICAL_RESISTANCE"
 ModifierIcon = ExtResource("15")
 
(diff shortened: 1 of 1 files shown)


hhyyrylainen (approved):
I was expecting this to be harder, but apparently this seemingly is perfectly fine way to fix the issue. So good job.

PR #5734, opened 2024-12-08
Make stop checking for equations constantly

### LINKED ISSUE #5590: Organelle tooltips shouldn't have ChemicalEquation `AutoRefreshProcess` on

Patryk26g:

Redid my  pr because I messed up the previous one.
Adds ChemicalEquations to the global group and call ls the UpdateEquation function on patch change. Hopefully this solution is acceptable

**Related Issues**

Closes #5590

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:

src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -595,6 +595,7 @@ private void MoveToPatchClicked()
             return;
         }
 
+        GetTree().CallGroup("ChemicalEquations", "UpdateEquation");
```
hhyyrylainen, line 598: I'm sorry to say but I don't like this solution. We don't use Godot groups for anything so adding it for just this one thing seems not good.

My preferred solution is still to trigger a GUI tooltip update in the editor in the patch changed callback. It needs a few levels of indirection but it shouldn't take that long to write all the proxy methods that get the notice to update where it is needed.

hhyyrylainen (changes requested):

src/microbe_stage/gui/PatchDetailsPanel.cs
Patryk26g: Do we have some similar solution somewhere in the code? Would be easier to get inspired by it :D

hhyyrylainen: I can't immediately think of an exact match but `UpdateMPCost` in `CellEditorComponent.GUI.cs` has a loop that updates a property on all tooltips. So basically that but instead just calling a method on each tooltip that would then in turn call a method on all chemical equations that the tooltip contains (this is likely the only slightly tricky part as the tooltips need to be updated to know about the chemical equations inside them).

Patryk26g:
Well just turning off the automatic process update on the organelleSelectionTooltip was enough.

I was also thinking about the processes that are displayed during the microbe game. When we display them they start updating constantly even when the processes list is hidden so we probably would need an issue for that

hhyyrylainen:
> Well just turning off the automatic process update on the organelle selection tooltip was enough.

What updates the tooltips then now to make sure they don't show outdated information?

> I was also thinking about the processes that are displayed during the microbe game. When we display them they start updating constantly even when the processes list is hidden so we probably would need an issue for that

Easiest thing would be if the process panel just deleted its contents when it is hidden but I don't consider this much of a problem as after all there's a limited set of possible processes the player cell might have so that caps the performance impact and many players never open the process panel so they wouldn't be impacted.

Patryk26g:
> What updates the tooltips then now to make sure they don't show outdated information?

`WriteOrganelleProcessList` is called in `UpdateOrganelleEfficiencies` which is called in 'CalculateOrganelleEffectivenessInCurrentPatch' which in turn is called whenever the patch or light level is changed

hhyyrylainen (approved):
I did quite a bit of testing and I think I can confirm that the process tooltip rebuild method is always called when necessary so the changes here to disable the auto-refresh on the tooltips should be exactly right and this PR should now be good to go as solving the issue.

PR #5751, opened 2024-12-15
Fix the lag when zooming out in the patch map

### LINKED ISSUE #4722: Patch map initial scroll out can hitch a bit (lag spike when zooming out)
I just noticed that in a early multicellular save in which I enter the editor and try to fully scroll out the patch map, I always get a slight lag spike when scrolling out at the exact same spot. So there's something non-smooth about the zoom out in there.

dligr:

Fixes the lag (actually just zoom returning to the default value) when initially zooming out in the patch map.

**Related Issues**

Fixes #4722

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (approved):
I think I was able to replicate the bug conditions and verify that this fix works.

PR #5753, opened 2024-12-15
Fix some patches not appearing in the editor dropdown

### LINKED ISSUE #5668: Patch dropdown selector in editor doesn't always contain all patches
I started a freebuild game, played a few generations, saved and loaded in the editor, and then played a bit more. Then I noticed an event in a patch that seemingly didn't exist. I was able to select it from the map but it is totally missing from the selector for some reason:

kuva

This might be freebuild specific, but in case it isn't I'm putting it into high priority issues.

dligr:

Fixes the issue with some patches not appearing in the editor report's patch dropdown

**Related Issues**

Fixes #5668

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (approved):
Seems like a good fix. This was much simpler than I was expecting. I don't know why that other method would have been used originally to not list all of the patches.

PR #5759, opened 2024-12-17
Fix photostudio's distance calculation formula

dligr:

Fixes photostudio's formula for calculating the distance to the photographed object.

Additionally, changes a constant to compensate for the formula being changed; otherwise the distance calculated is slightly less than it should be.

The reasoning for the new formula:
image

**Related Issues**

A TODO in the code

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -1558,7 +1558,7 @@ public static class Constants
 
     public const float PHOTO_STUDIO_CAMERA_FOV = 70;
     public const float PHOTO_STUDIO_CAMERA_HALF_ANGLE = PHOTO_STUDIO_CAMERA_FOV / 2.0f;
-    public const float PHOTO_STUDIO_CELL_RADIUS_MULTIPLIER = 0.80f;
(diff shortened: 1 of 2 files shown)


hhyyrylainen:
I'm unsure about this change, especially this close to release. On one hand this makes many cells match the image size more closely, but caused many simple cells to have their images cutoff, which doesn't look good (also I checked the thriveopedia and this seems to help a bit there, I think).

So here's a screenshot with these changes:


And here's without (I'll note that most of the cutoff cells are no longer cutoff so this gives a better impression on how well the game is working):

dligr:
I have managed to calculate a constant value that results in the visible scale being similar to what it was. It should stay the same for different FOVs, so there shouldn't be a problem with this.

I also want to emphasize that this PR isn't really supposed to change the current visuals, but rather fix the way distance scales with FOV.

hhyyrylainen (approved):
Comparing things now I can't really tell any difference in cell images so I'll merge this.

PR #5763, opened 2024-12-18
Fixed auto-evo migration using wrong patch to calculate

hhyyrylainen:

the resulting miche (it used the source patch instead of the target patch)

**Related Issues**

When adding a miche that uses chunks only present in some patches I noticed errors where it tried to be processed for chunks that didn't exist in a patch. Eventually after adding a lot of other error checking code I noticed that the migration miche accidentally used the source patch data and the target miche rather than pairing up the miche with the target patch. So this fixes that but also adds a bit of other safety code.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/Miche.cs
+++ b/src/auto-evo/Miche.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using Newtonsoft.Json;
 
 /// <summary>
@@ -46,6 +47,8 @@ public class Miche
     /// </remarks>
     [JsonProperty]
     public Species? Occupant;
+
+    private bool locked;
 
     public Miche(SelectionPressure pressure)
     {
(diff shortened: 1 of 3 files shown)


CountingBeeps:
Oops. Also the miche system probably should have more safety code in general, so LGTM.

CountingBeeps:
Actually one other thing. A non dry call to insert species should probably call ThrowIfLocked(); too.

hhyyrylainen:
> A non dry call to insert species should probably call ThrowIfLocked(); too.

I wasn't sure as I thought that the point of the tree was that it is allowed to be modified (at least after the initial build to fill it with the species) so I didn't add that; I was only sure the tree *structure* is not allowed to change so I added only that part.

CountingBeeps:
Ah. I think I misunderstood the purpose. I assumed it was locking all changes, not just structure changes. What you have makes sense then.

hhyyrylainen:
Yeah. Separately locking all changes could be useful, but that would be I think a bit harder for me to find the right place to insert that locking so that the locking is applied only after the miche modification steps are done.

PR #5766, opened 2024-12-19
Try to fix Bathypelagic Background

### LINKED ISSUE #1265: Fix background image artifacts
The backgrounds have very slight artifacts (horizontal line here):
Kuvakaappaus - 2020-05-15 23-13-00

Probably something to do with the shader / layer texture repeats.

HexapodPhilosopher:

I was not able to replicate the issue with a horizontal line appearing on the bathypelagic patch background, but I noticed that the only patch backgrounds not set to detect_3d/compress_to=1 were for Bathypelagic and Vents. Hopefully this fixes the issue.
(I also checked the texture and found no issues with it.)

**Related Issues**

Hopefully closes #1265

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/assets/textures/background/Thrive_bathy0.png.import
+++ b/assets/textures/background/Thrive_bathy0.png.import
@@ -31,4 +31,4 @@ process/normal_map_invert_y=false
 process/hdr_as_srgb=false
 process/hdr_clamp_exposure=false
 process/size_limit=0
-detect_3d/compress_to=0
+detect_3d/compress_to=1
--- a/assets/textures/background/Thrive_vent0.png.import
+++ b/assets/textures/background/Thrive_vent0.png.import
@@ -31,4 +31,4 @@ process/normal_map_invert_y=false
 process/hdr_as_srgb=false
 process/hdr_clamp_exposure=false
 process/size_limit=0
-detect_3d/compress_to=0
+detect_3d/compress_to=1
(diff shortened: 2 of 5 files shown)


hhyyrylainen:
That shouldn't really have an effect as the detect option only triggers in the godot editor when you use the texture in a way that warrants different options, it will auto-apply those. And once detected it turns off the autodetection, which is probably why the other backgrounds have the property unset.

HexapodPhilosopher:
All the other ones have the property on. These are the only ones that didn't.

hhyyrylainen:
Probably what happened is that for these backgrounds they were used in the godot editor to try something which triggered the 3D detection, messing up their import settings and turning the detection off. Then when that was noticed the import settings were fixed but 3D detection was not put on because that's going to do totally the wrong thing with these backgrounds as these are meant for a 2D effect kind of thing.

At least that's my understanding that these "detect 3D" options don't impact anything outside the Godot editor whatsoever, so these changes shouldn't be able to do anything, other than set up a landmine for if someone accidentally does something in the Godot editor that triggers the detection and causes the import settings to change to unsuitable ones.

hhyyrylainen:
I'll close this PR for now as I don't think this should affect anything. If I'm wrong absolutely reopen this and let me know.

PR #5771, opened 2024-12-22
Make compound clouds and 3D objects' lighting depend on the light level

### LINKED ISSUE #3920: Day/night cycle should affect cell brightness (and probably also chunk)
It seems like only compound clouds and the backgrounds are affected by night. To improve the effect of it being night various other game objects should react, this can probably be achieved by adjusting the game world light, though that probably leaves out the floating particles still

dligr:

Makes the lighting of 3D objects (cells, clouds, etc.) and compound clouds depend on the current light level (if the patch has day-night cycle).

**Related Issues**

Closes #3920

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/PatchManager.cs
+++ b/src/microbe_stage/PatchManager.cs
@@ -145,8 +145,7 @@ public class PatchManager : IChildPropertiesLoadCallback
 
         gameWorld.UpdateGlobalLightLevels();
 
-        var multiplier = biome.Conditions.GetCompound(Compound.Sunlight, CompoundAmountType.Current).Ambient *
-            gameWorld.LightCycle.DayLightFraction;
+        var multiplier = biome.Conditions.GetCompound(Compound.Sunlight, CompoundAmountType.Current).Ambient;
         compoundCloudSystem.SetBrightnessModifier(multiplier * (compoundCloudBrightness - 1.0f) + 1.0f);
         UpdateLight(biome);
     }
--- a/src/microbe_stage/PatchManager.cs
+++ b/src/microbe_stage/PatchManager.cs
@@ -151,14 +151,13 @@ public class PatchManager : IChildPropertiesLoadCallback
 
         if (maxLightLevel > 0.0f)
         {
-            multiplier = biome.Conditions.GetCompound(Compound.Sunlight, CompoundAmountType.Current).Ambient * biome.;
+            multiplier = biome.Conditions.GetCompound(Compound.Sunlight, CompoundAmountType.Current).Ambient *
+                CurrentGame!.GameWorld.LightCycle.DayLightFraction;
         }
         else
         {
             multiplier = 1.0f;
         }
-
-        GD.Print(multiplier, biome.Conditions.GetCompound(Compound.Sunlight, CompoundAmountType.Current).Ambient);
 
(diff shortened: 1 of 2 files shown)


Patryk26g (changes requested):
I was testing it in a cave patch with 0% light and lighting was still changing with the day -night cycle

dligr:
This should be fixed now (apparently the patch manager used raw light level, which changes regardless of patch)

Patryk26g:
Okay now lighting in patches with no light is independent from the day-night cycle but the lighting looks the same as if there was light but the chunks and cells should be dark just as they are in other patches when there is night.

And I think that it is to dark anyway for the night lightning

src/microbe_stage/PatchManager.cs
```diff
@@ -145,9 +143,24 @@ public void UpdateAllPatchLightLevels()
...
         gameWorld.UpdateGlobalLightLevels();
+
+        var maxLightLevel = biome.Conditions.GetCompound(Compound.Sunlight, CompoundAmountType.Biome).Ambient;
+
+        float multiplier;
+
+        if (maxLightLevel > 0.0f)
+        {
+            multiplier = biome.Conditions.GetCompound(Compound.Sunlight, CompoundAmountType.Current).Ambient *
+                CurrentGame!.GameWorld.LightCycle.DayLightFraction;
```
hhyyrylainen, line 155: Doesn't this end up applying the light level twice? Because the current sunlight compound should already follow the light cycle, and here that is then further multiplied by data from the LightCycle.

dligr:
The issue with day light fraction multiplying should be fixed now. The `biome` variable was actually a biome template, so its ambient light value should be static, but anyway I changed the `UpdateAllPatchLightLevels` function to accept a patch as an argument to conform it to the way `MicrobeStage` gets its light value.

@Patryk26g
> Additionally, the night lighting feels too dark overall and playing 0% lux patches with these settings wouldn't be a great experience

I changed the world lighting to have a minimum value of 20%, so the night should be a bit brighter now.
> Also the compound clouds seem to be uneffected

The compound clouds should be affected, though it's not very visible in some patches. The effect should be very visible in the estuary patch.
> Okay now lighting in patches with 0% lux is independent from the day-night cycle but the lighting looks the same as if there was light but the chunks and cells should be dark just as they are in other patches when there is night.

I think that lighting in 0% lx patches should be left as is, because adding light to those patches was an artistic choice, so it probably shouldn't be changed.

> and I think that the lightning level should also be changed in editor when we go to "membrane" tab and change day time to night

I'm not sure if this should be done, since it might worsen the editor's clarity

hhyyrylainen:
> I think that lighting in 0% lx patches should be left as is, because adding light to those patches was an artistic choice, so it probably shouldn't be changed.

:+1: 

> I'm not sure if this should be done, since it might worsen the editor's clarity

I don't really have an opinion on this. I'd need to see how it looks, but the idea with the editor is that the cell is easily visible (which is why it has more light by default than the stage).

hhyyrylainen (approved):
The code looks straightforward to me now and I tested in game and all seemed to work fine.

hhyyrylainen:
Thanks for adding this long-open feature.

PR #5774, opened 2024-12-28
Fix compound balance mode index

0HyperCube:

The id for the compound balance mode option button now corresponds to the values given in the `BalanceDisplayType` enum.
https://github.com/Revolutionary-Games/Thrive/blob/dce65afc2961d5c00c1b17b3227a01834f03a15f/src/microbe_stage/BalanceDisplayType.cs#L5-L15

**Related Issues**

I noticed that the compound balance mode option button didn't work properly and also produced this log:
```
Unknown compound balance type: 2
```

**Progress Checklist**

      break existing features:
      members reviewing this PR)


dligr (commented):
I can confirm the outlined issue. Regarding the changes, I have noticed that the order of the items in `CompoundBalanceDisplay.tscn` is different than that of the enum in the code, which leads to the dropdown options corresponding to wrong display types. So, I think that the order should be reversed, either in the .tscn file or in the C# code.

dligr (approved):

PR #5777, opened 2024-12-29
Make Auto-Evo able to evolve slime jets

### LINKED ISSUE #4128: Make auto-evo take slime jets into account (more)
I just checked and it seems like auto-evo has no clue whatsoever that slime jets exist. Would be nice if they had some impact for example related to predation.

### LINKED ISSUE #5776: Make auto-evo take slime jets into account
Opening this issue since #4128 was closed although the mucilage isn't taken into account

Patryk26g:

Makes it so that slime jets are taken into account in auto-evo

Screenshot_20241229_125618
Screenshot_20241229_125633

**Related Issues**

Closes #4128 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -379,44 +379,73 @@ public void Clear()
...
-                if (process.Process.Outputs.TryGetValue(mucilage, out var mucilageAmount))
-                {
-                    mucilageScore += mucilageAmount * Constants.AUTO_EVO_MUCILAGE_PREDATION_SCORE;
-                }
+        // Make sure that slime jets are positioned at the back of the cell, because otherwise they will push the cell
+        // backwards (into the predator or away from the prey) or to the side
+        foreach (var slimeJetPosition in slimeJetsHorizontalPositions)
+        {
+            // Range of [0,1] (front to back)
+            var mucilageMultiplier = (float)(slimeJetPosition - topPosition) / (bottomPosition - topPosition);
```
adQuid, line 434: wouldn't it be possible for bottomPosition to equal topPosition, causing a divide by zero error?

adQuid:
I haven't tested yet, but I can verify that this is using auto-evo correctly.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -379,44 +379,73 @@ public void Clear()
...
         var oxytoxyScore = 0.0f;
-        var mucilageScore = 0.0f;
+        var slimeJetScore = 0.0f;
 
         var organelles = microbeSpecies.Organelles.Organelles;
         var organelleCount = organelles.Count;
+        var slimeJetsCount = 0;
+        var topPosition = 0;
+        var bottomPosition = 0;
+        List<int> slimeJetsHorizontalPositions = new List<int>();
```
adQuid, line 397: is this named correctly? Based on the code below it seems like these are vertical positions.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -379,44 +379,73 @@ public void Clear()
...
+        {
+            // Range of [0,1] (front to back)
+            var mucilageMultiplier = (float)(slimeJetPosition - topPosition) / (bottomPosition - topPosition);
+            if (mucilageMultiplier <= 0.5)
+            {
+                slimeJetScore *= 0;
+            }
+            else
+            {
+                slimeJetScore += mucilageMultiplier;
```
adQuid, line 441: It seems like this logic is dependent on the order in which organelles are traversed. For example, if the first three mucilage were facing forward, but a fourth was facing backwards, it would set the `slimeJetScore` to zero three times, then add the fourth one, right?

I might suggest something like this:
```suggestion
                slimeJetScore += Constants.AUTO_EVO_MUCILAGE_PREDATION_SCORE * (mucilageMultiplier - 0.5f);
```

Patryk26g: well the problem was with the operation - it was meant to be multiplication instead of addition :sweat_smile:

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -277,10 +277,10 @@ public float GetPredationScore(Species species, Species preySpecies, IBiomeCondi
...
         }
 
-        var (pilusScore, oxytoxyScore, mucilageScore) = GetPredationToolsRawScores(microbeSpecies);
-
-        // TODO: Support mucilage in predation score
-        mucilageScore *= 0;
+        var (pilusScore, oxytoxyScore, predatorSlimeJetScore) = GetPredationToolsRawScores(microbeSpecies);
+        var (_, _, preySlimeJetScore) = GetPredationToolsRawScores(prey);
+        var slimeJetScore = predatorSlimeJetScore - preySlimeJetScore;
+        slimeJetScore = slimeJetScore < 0 ? 0 : slimeJetScore;
```
dligr, line 283: Maybe it would be better to use Math.Max() here?

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -379,44 +379,78 @@ public void Clear()
...
         var oxytoxyScore = 0.0f;
-        var mucilageScore = 0.0f;
+        var slimeJetScore = 0.0f;
 
         var organelles = microbeSpecies.Organelles.Organelles;
         var organelleCount = organelles.Count;
+        var slimeJetsCount = 0;
+        var topPosition = 0;
+        var bottomPosition = 0;
+        List<int> slimeJetsVerticalPositions = new List<int>();
```
hhyyrylainen, line 397: Using temporary lists should be avoided for memory allocation efficiency.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -379,44 +379,78 @@ public void Clear()
...
-                }
+        // Make sure that slime jets are positioned at the back of the cell, because otherwise they will push the cell
+        // backwards (into the predator or away from the prey) or to the side
+        foreach (var slimeJetPosition in slimeJetsVerticalPositions)
+        {
+            var mucilageMultiplier = 1.0f;
+            if (bottomPosition != topPosition)
+            {
+                // Range of [0,1] (front to back)
+                mucilageMultiplier = (float)(slimeJetPosition - topPosition) / (bottomPosition - topPosition);
```
hhyyrylainen, line 437: Isn't this slightly wrong? Like any external organelles the slime jets jump to the closest membrane edge. So the efficiency here should just be calculated as the angle between the Y-axis and a line from the slime jet hex position to the middle hex (which is always 0,0). For examples you can take a look at the movement component movement force calculation.

Patryk26g: Now the angle between the slimejet and y-axis is calculated instead of this

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -496,4 +508,17 @@ private float CalculateStorageScore(MicrobeSpecies species, BiomeConditions biom
...
     }
+
+    /// <summary>
+    ///   Calculates cos of angle between the organelle and vertical axis
+    /// </summary>
+    private float CalculateAngleMultiplier(Hex pos)
+    {
+        Vector3 organellePosition = Hex.AxialToCartesian(pos);
+        Vector3 verticalAxis = Hex.AxialToCartesian(new Hex(0, 1));
+        float angleCos = organellePosition.Dot(verticalAxis) / (organellePosition.Length() * verticalAxis.Length());
```
hhyyrylainen, line 519: What's the math here? I'm having a bit of trouble trying to understand this... Why is 0,1 in polar coordinates converted like this and then multiple lengths are calculated? Why not just do the same as the MovementComponent or `organellePosition.AngleTo(Hex.AxialToCartesian(new Hex(0, 0));`?

Patryk26g: I changed the maths a little bit so that it is easier to understand. I'm just calculating the cosine of the vector pointing downwards and the vector pointing to the organelle, so that the smaller the angle the bigger the multiplier

hhyyrylainen: I guess it is fine now, though there's an unnecessary Normalized call as vector (0, 0, 1) has to be normalized already. Also just to confirm: this will work even if the organelle position is 0,0,0?

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -496,4 +508,17 @@ private float CalculateStorageScore(MicrobeSpecies species, BiomeConditions biom
...
     }
+
+    /// <summary>
+    ///   Calculates cos of angle between the organelle and vertical axis
+    /// </summary>
+    private float CalculateAngleMultiplier(Hex pos)
+    {
+        Vector3 organellePosition = Hex.AxialToCartesian(pos);
+        Vector3 downVector = new Vector3(0, 0, 1);
+        float angleCos = organellePosition.Normalized().Dot(downVector.Normalized());
```
hhyyrylainen, line 519:
```suggestion
        float angleCos = organellePosition.Normalized().Dot(downVector);
```

One more tiny suggestion, a slight performance improvement from not normalizing something that always has to be already normalized.

src/auto-evo/simulation/SimulationCache.cs
Patryk26g: It will just result as cos(angle) = 0

hhyyrylainen: Doesn't that then lead to 0 score? If I remember right the slime jet is biased to go backwards meaning that actual 0,0 position should mean a multiplier of 1...

Patryk26g: Yes but that is a really rare case and I think adding an if statement would just add some unnecessary calculations which won't result in any noticeable difference in the evolution

hhyyrylainen: I would add the if for two reasons:
- Energy calculations impact the player
- And we already have multiple open bug reports about inconsistent stuff when placed at 0,0 (so we'd eventually have someone find out this bug and report it, at which point it will be super hard to diagnose as I won't remember this discussion)

Patryk26g: You're right, that's an easy fix but I'll do it in the evening cause I don't have access to my laptop rn unfortunately

CountingBeeps:
This looks pretty good overall. 26% feels really high, especially for what should be a rarer organelle. It would probably be best to lower the score for it a bit.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -277,10 +278,10 @@ public float GetPredationScore(Species species, Species preySpecies, IBiomeCondi
...
         }
 
-        var (pilusScore, oxytoxyScore, mucilageScore) = GetPredationToolsRawScores(microbeSpecies);
-
-        // TODO: Support mucilage in predation score
-        mucilageScore *= 0;
+        var (pilusScore, oxytoxyScore, predatorSlimeJetScore) = GetPredationToolsRawScores(microbeSpecies);
+        var (_, _, preySlimeJetScore) = GetPredationToolsRawScores(prey);
+        var slimeJetScore = predatorSlimeJetScore - preySlimeJetScore;
+        slimeJetScore = Math.Max(slimeJetScore, 0);
```
CountingBeeps, line 284: I'm not actually sure there is any reason score can't be negative. It would probably be good to penalize predation versus prey with slime jets

PR #5778, opened 2024-12-29
Rigidity slider fix

### LINKED ISSUE #4150: Rigidity slider breaks when trying to move it at exactly 0 MP

Patryk26g:

Hopefully this fixes the rigidity slider issue. The problem was mainly with the division by 0.

**Related Issues**

Closes #4150

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -1272,6 +1272,7 @@ public partial class CellEditorComponent :
         var theoreticalCost = Editor.WhatWouldActionsCost(new[] { data });
         var cost = (int)Math.Ceiling(Math.Ceiling(theoreticalCost / costPerStep) * costPerStep);
 
+        // Cases where mutation points are equal 0 are handled below in the next "if" statement
         if (cost > Editor.MutationPoints && Editor.MutationPoints != 0)
         {
             int stepsToCutOff = (int)Math.Ceiling((cost - Editor.MutationPoints) / costPerStep);
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -1270,7 +1270,7 @@ public partial class CellEditorComponent :
 
         var cost = Editor.WhatWouldActionsCost(new[] { data });
 
-        if (cost > Editor.MutationPoints)
+        if (cost > Editor.MutationPoints && Editor.MutationPoints != 0)
         {
             int stepsToCutOff = (int)Math.Ceiling((cost - Editor.MutationPoints) / costPerStep);
             data.NewRigidity -= (desiredRigidity - previousRigidity > 0 ? 1 : -1) * stepsToCutOff /
@@ -1283,7 +1283,7 @@ public partial class CellEditorComponent :
 
         // Make sure that if there are no mutation points the player cannot drag the slider
         // when the cost is rounded to zero
-        if (Editor.MutationPoints == 0 && cost == 0)
+        if (cost >= 0 && (Editor.MutationPoints - cost < 0 || costPerStep > Editor.MutationPoints))
         {
             UpdateRigiditySlider(previousRigidity);
             return;
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -1281,6 +1281,8 @@ public partial class CellEditorComponent :
             return;
         }
 
+        // Make sure that if there are no mutation points the player cannot drag the slider
+        // when the cost is rounded to zero
         if (Editor.MutationPoints == 0 && cost == 0)
         {
             UpdateRigiditySlider(previousRigidity);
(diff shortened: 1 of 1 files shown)


src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1281,6 +1281,12 @@ public void OnRigidityChanged(int desiredRigidity)
             return;
         }
 
+        if (Editor.MutationPoints == 0 && cost == 0)
```
dligr, line 1284: I think that it would be good to add a comment here clarifying why this check is needed (if I get it right, it handles the corner case of the cost being rounded to zero, while adding up with other costs in such a way that MP ends up at -1)

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1281,6 +1281,14 @@ public void OnRigidityChanged(int desiredRigidity)
             return;
         }
 
+        // Make sure that if there are no mutation points the player cannot drag the slider
+        // when the cost is rounded to zero
+        if (Editor.MutationPoints == 0 && cost == 0)
```
dligr, line 1286: I noticed that when playing with the mutation cost of 1.9, dragging the rigidity slider might bring the MP down to -1. So, I think that this check should be updated to be able to handle any mutation cost. Something like `Editor.MutationPoints < cost + 1 && CostMultiplier % 1 != 0` should probably work (the second part checking if the number is non-integer and might have rounding issues).

Patryk26g: Went with a different approch, tested it and couldn't find any more issues

dligr:
I have found a new corner case:
1. Set mutation cost at 1.9
2. Place two cytoplasms
3. Drag the rigidity slider to get the MP to 7
4. Skip two slider steps with a mouse press; the MP ends up at -1

It seems like the issue is that the step cutoff calculations don't take into account the extra 1 that may be added due to rounding

Patryk26g:
I think this corner case is working properly now

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -1261,18 +1261,20 @@ public void OnRigidityChanged(int desiredRigidity)
...
             CostMultiplier = CostMultiplier,
         };
 
-        var cost = Editor.WhatWouldActionsCost(new[] { data });
+        // In some cases "theoreticalCost" might get rounded improperly
+        var theoreticalCost = Editor.WhatWouldActionsCost(new[] { data });
+        var cost = (int)Math.Ceiling(Math.Ceiling(theoreticalCost / costPerStep) * costPerStep);
 
-        if (cost > Editor.MutationPoints)
+        if (cost > Editor.MutationPoints && Editor.MutationPoints != 0)
```
hhyyrylainen, line 1275: I think adding a comment here as well explaining that the mutation points being zero corner case is handled below.

BTW unrelated to this exact change, but what would you expect to happen once mutation points are converted to floats? That's a stalled PR and a pretty long-wanted fix for the code architecture.

dligr:
I have found another corner case:
1. Set mutation cost at 0.2
2. Get to editor
3. Place some hydrogenases to get MP to 1
4. Move the rigidity slider three steps and it becomes immovable

src/microbe_stage/editor/CellEditorComponent.cs
Patryk26g: That's hard to tell but I guess the slider code would need some refactoring

hhyyrylainen: That's about what I expected... Hopefully this bugfix doesn't make that eventual work harder.

Patryk26g:
> The only issue I could find now is that when choosing rigidity though clicking while having low enough MP, there might be more steps cut off than needed. For example, with the MP cost modifier of 1.9:
> 
> 1. Switch to any membrane type
> 2. Move rigidity one step left
> 3. Click somewhere on the right part of the slider
> 
> The slider's value would then be one step away from the right value due to excessive cut off

Well that seems like a really rare case and not that problematic. I think that in order to cover this one I would actually need to rewrite the whole function from scratch because I don't see any other solution

hhyyrylainen (approved):
This fixes the primary bug but as @dligr found out this leaves 1 or 2 MP when trying to move too far in one step, which is a pretty noticeable bug when playing with the slider.

As this has been open for a while and fixes the game locking up, I'll merge this but open a follow up issue.

hhyyrylainen:
Opened the new issue: https://github.com/Revolutionary-Games/Thrive/issues/5821

Issue #5536, opened 2024-09-24
Add MP cost for modifying flagellum length upgrade in the editor


hhyyrylainen:
so that it is not free. This is the latest in a long series of issues about making upgrades cost MP, like: 
- https://github.com/Revolutionary-Games/Thrive/issues/4093
- https://github.com/Revolutionary-Games/Thrive/issues/4092

Patryk26g:
I was wondering what the best option to display the cost would be. I came up wtih something like this. The receptor has many different customization options so I think it would be reasonable to make it so that each options has a separate cost
![image](https://github.com/user-attachments/assets/583c2ca9-b755-4cdc-aabe-83c32e4f3938)

hhyyrylainen:
Does the cost have to be displayed? For example the rigidity slider has no such thing, it just immediately updates the main MP bar to show how. So I think that initially the same thing could be extended to the modify popup to show how much MP the upgrade as configured would cost (similarly to showing how much placing the current organelle with the -> arrow is shown). Though I do agree that the chemoreceptor should have different parts costing different amount of MP so maybe that would be good to display the different costs of the different settings.

Patryk26g:
Well the rigidity slider has cost displayed in the header. It costs 2MP for every tick of the slider. The problem with the chemoreceptor and flagellum sliders is that they are too smooth (continous instead of discretized) so I would suggest that I add that "jumping" effect to these sliders as well so that if we decide to display the cost it would be more logical for the players
![image](https://github.com/user-attachments/assets/4e0df21d-571b-4ed2-b4fd-f2bd1ed33b6d)

hhyyrylainen:
Oh yeah it certainly shows the cost there. And not all of the upgrades have tooltips on them so it'd be pretty hard to require them first before being able to show the cost so yeah maybe what you showed first is the best design that can fit the current constraints. I'll note though that I already added costs to the toxin organelle upgrades and those wouldn't be visually consistent with your approach. I'd like a consistent way to show the costs as otherwise the player could see the lack of polish due to inconsistent MP cost displays in the different upgrade GUIs.

Issue #4095, opened 2023-02-02
Undoing organelle upgrades in later editor cycle should not refund the full amount


hhyyrylainen:
as otherwise it is just a free bank for keeping a bunch of MP and I'd consider that an exploit. Instead the MP refund should be like only half of the price of an upgrade when moving back to the base form, or another solution can be come up with as well.

84634E1A607A:
I'd suggest a "*unlock*" approach, which, after you use a certain amount of MP you unlock a certain function, and switching back and forth cost a few MP.

hhyyrylainen:
Tracking that state would require extra data in the organelle template, and is such a more complicated thing, but I guess that'd be a fine approach to try to implement.

hhyyrylainen:
There's a chance that this is fixed by: https://github.com/Revolutionary-Games/Thrive/pull/5779

hhyyrylainen:
I think this should now work with the latest MP recalculation system.

Issue #4093, opened 2023-02-01
Make chemoreceptor upgrades cost MP


hhyyrylainen:
should be really low cost like just a couple of MP

Same caveat that the base system needs slight changes applies as here: https://github.com/Revolutionary-Games/Thrive/issues/4092

PR #5779, opened 2024-12-29
Fix double refunding of organelle upgrade MP

### LINKED ISSUE #5524: Reverting an organelle upgrade and then selecting it again incorrectly gives back the used MP (infinite MP glitch)
Steps to reproduce:
- Start a normal game
- Cheat to the editor
- Use the cheat to unlock all organelles
- Place a slime jet
- Modify it into a mucocyst
- Hit ok
- Modify it again back to a normal state
- Hit ok
- And now when it is modified again it costs 0 MP to select the mucocyst upgrade again.
- Hit ok
- Now finally modify one last time to switch back to the default state of the organelle which now gives more MP than before and these steps can be repeated for an infinite MP exploit

This bug is probably caused by the GUI somehow mistaking the organelle upgrade state, allowing freely to select an upgrade when it should still cost the normal amount as it was reversed by a later action. Could be a bug in the actions cancelling out system incorrectly calculating things as cancelling out.

If this is fixed, this might also fix: https://github.com/Revolutionary-Games/Thrive/issues/5518

0HyperCube:

When changing the organelle upgrade repeatedly in the same editor cycle, MP would be refunded twice (as detailed in #5524). This is because the cost of an upgrade also includes the negated cost of the current upgrade, meanwhile the editor refunds MP where `ActionInterferenceMode.Combinable` is returned as the interference mode.

I initially fixed this by not combining the actions. However I have now changed the implementation to avoid the negated cost.

**Related Issues**
Closes #5524

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/action_data/OrganelleUpgradeActionData.cs
+++ b/src/microbe_stage/editor/action_data/OrganelleUpgradeActionData.cs
@@ -41,7 +41,7 @@ public class OrganelleUpgradeActionData : EditorCombinableActionData<CellType>
 
     protected override ActionInterferenceMode GetInterferenceModeWithGuaranteed(CombinableActionData other)
     {
-        // Organelle upgrades can never combine since the user will have already been refunded for the cost of the old choice.
+        // Organelle upgrades can never combine since the old choice will have already been refunded.
         // https://github.com/Revolutionary-Games/Thrive/issues/5524
 
         return ActionInterferenceMode.NoInterference;
--- a/src/microbe_stage/editor/action_data/OrganelleUpgradeActionData.cs
+++ b/src/microbe_stage/editor/action_data/OrganelleUpgradeActionData.cs
@@ -31,6 +31,11 @@ public class OrganelleUpgradeActionData : EditorCombinableActionData<CellType>
         var newUpgrades = NewUpgrades.UnlockedFeatures.Except(OldUpgrades.UnlockedFeatures)
             .Where(u => availableUpgrades.ContainsKey(u)).Select(u => availableUpgrades[u]);
 
+        // TODO: Removals should cost MP: https://github.com/Revolutionary-Games/Thrive/issues/4095
+        // var removedUpgrades = OldUpgrades.UnlockedFeatures.Except(NewUpgrades.UnlockedFeatures)
+        //     .Where(u => availableUpgrades.ContainsKey(u)).Select(u => availableUpgrades[u]);
+        // ? removedUpgrades.Sum(u => u.MPCost);
+
         cost += newUpgrades.Sum(u => u.MPCost);
 
         return cost;
--- a/src/microbe_stage/editor/action_data/OrganelleUpgradeActionData.cs
+++ b/src/microbe_stage/editor/action_data/OrganelleUpgradeActionData.cs
@@ -41,13 +41,8 @@ public class OrganelleUpgradeActionData : EditorCombinableActionData<CellType>
 
     protected override ActionInterferenceMode GetInterferenceModeWithGuaranteed(CombinableActionData other)
     {
-        if (other is OrganelleUpgradeActionData upgradeActionData)
-        {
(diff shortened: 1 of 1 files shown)


dligr (approved):
This PR fixes the issue based on my testing; the solution makes sense to me too.

src/microbe_stage/editor/action_data/OrganelleUpgradeActionData.cs
```diff
@@ -41,13 +41,8 @@ protected override int CalculateCostInternal()
...
     {
-        if (other is OrganelleUpgradeActionData upgradeActionData)
-        {
-            if (ReferenceEquals(UpgradedOrganelle, upgradeActionData.UpgradedOrganelle))
-                return ActionInterferenceMode.Combinable;
-        }
-
-        // The replacing action is in the remove organelle action
+        // Organelle upgrades can never combine since the old choice will have already been refunded.
+        // https://github.com/Revolutionary-Games/Thrive/issues/5524
```
hhyyrylainen, line 45: I'm a bit hesitant on this as if you have an upgrade like the chemoreceptor and it is made to cost MP, then wouldn't the combine be required for MP to work correctly?

For example:
- Player edits detection distance from 200 to 500 and it costs 3 MP
- The player then edits the distance from 500 to 200 again costing 3 MP
- The combined action should then be the change being from 200 to 200 and not costing any MP, but if the combining system is told that upgrade changes cannot combine, then this would cost 6 MP.

hhyyrylainen: Just to clarify that is not the case yet, but is a desired feature: https://github.com/Revolutionary-Games/Thrive/issues/4093 also relevant for the flagellum: https://github.com/Revolutionary-Games/Thrive/issues/5536

dligr: I think that the original problem comes down to the fact that removing upgrades refunds MP, which causes weirdness with combined upgrades.
```C#
cost += newUpgrades.Sum(u => u.MPCost);
cost -= removedUpgrades.Sum(u => u.MPCost);
```
Since #4095 is an issue, I believe that this refunding wasn't a desired behavior in the first place, as refunding MP by reverting upgrades in the same editor cycle should be done by action combining.

hhyyrylainen: > reverting upgrades in the same editor cycle should be done by action combining.

That's probably true and a mistake I didn't realize at the time when I was writing the general upgrades system for the editor.

0HyperCube: I have changed it so that it won't refund MP from previous editor sessions (since I assume that is what you wanted?).

When I previously asked about the desired behaviour, nobody gave any particular opinion either way.

dligr: I meant more an approach like this:
```C#
cost += newUpgrades.Sum(u => u.MPCost);
cost += removedUpgrades.Sum(u => u.MPCost);
```
So that any changes to upgrades cost MP. It shouldn't be possible to refund actions from previous editor sessions too.
I don't think there will be any problems with reverting (=refunding) upgrades within the same session because of action combining.

hhyyrylainen (approved):
This fixes the infinite MP exploit (as long as I didn't accidentally mess up the steps). So I think this is good to merge now. I added a further TODO about a related issue that this doesn't fix (and I kept the deleted code as commented out for future reference).

src/microbe_stage/editor/action_data/OrganelleUpgradeActionData.cs
hhyyrylainen: I merged this for now, but I edited the issue https://github.com/Revolutionary-Games/Thrive/issues/4095 to be about making the upgrades cost MP to also remove.

PR #4593, opened 2023-11-28
Display mucilage in specialized vacuole upgrade menu

CI09:
**Brief Description of What This PR Does**

Now code for displaying compounds in vacuole upgrade menu excludes agents except for mucilage.

Probably could have been done much better than that so I am open for suggestions.

**Related Issues**

Closes #4589

**Progress Checklist**

Note: before starting this checklist the PR should be marked as non-draft.

- [x] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [ ] Initial code review passed (this and further items should not be checked by the PR author)
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author. Merging must follow our
[styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md#git).

PR #4553, opened 2023-11-12
Split microbe hud to its own scene

CI09:
**Brief Description of What This PR Does**

This PR makes MicrobeHUD.tscn a seperate scene. HotBar mouse pressed signals couldn't connect directly to PlayerMicrobeInput.cs so I connected them to MicrobeHUD.tscn and then relayed them to PlayerMicrobeInput.cs instead. I hope the changes are ok, but don't hesitate to edit / suggest me something. 

**Related Issues**

Closes #3741

**Progress Checklist**

Note: before starting this checklist the PR should be marked as non-draft.

- [x] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [x] Initial code review passed (this and further items should not be checked by the PR author)
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author. Merging must follow our
[styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md#git).

PR #5780, opened 2024-12-30
Add simple energy balance info without dictionaries

### LINKED ISSUE #5426: Create variant of GetEnergyBalanceForSpecies for auto-evo that just has the overall balance
and doesn't try to track the individual consumptions and productions as calculating and storing those requires extra time which slows down auto-evo unnecessarily

This is mostly just copy-pasting but I'd like to see in any PR solving this some good architectural combining of the two method variants as much as possible to reduce real code duplication (but the more important consideration is raw performance, so some dirtyness here would be fine as this is a pretty well contained part that's critical for speed).

0HyperCube:

Create an `EnergyBalanceInfoSimple` class that doesn't record the consumption and production. This is used for the auto evo since it doesn't require the consumption and production information.

Also add an `EnergyBalanceInfoFull` class that inherits from `EnergyBalanceInfoSimple` but with the dictionaries for consumption and production. This is used for the editor display.

Since there are virtual methods now, this will likely be detrimental to performance. However I wasn't able to test the significance of this compared to removing the dictionary allocations for auto-evo since there doesn't see to be a great way to benchmark auto-evo.

**Related Issues**
Closes #5426

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/simulation/SimulationCache.cs
+++ b/src/auto-evo/simulation/SimulationCache.cs
@@ -79,7 +79,6 @@ public class SimulationCache
         // different species but with same organelles to be able to use the same cache value) would be nice here
         var key = (species, biomeConditions);
 
-        // TODO: check if caching instances of these objects would be better than always recreating
         if (cachedSimpleEnergyBalances.TryGetValue(key, out var cached))
         {
             return cached;
@@ -87,6 +86,7 @@ public class SimulationCache
 
         var maximumMovementDirection = MicrobeInternalCalculations.MaximumSpeedDirection(species.Organelles);
 
+        // TODO: check if caching instances of these objects would be better than always recreating
         cached = new EnergyBalanceInfoSimple();
 
         // Auto-evo uses the average values of compound during the course of a simulated day
--- a/src/auto-evo/simulation/SimulationCache.cs
+++ b/src/auto-evo/simulation/SimulationCache.cs
@@ -28,7 +28,10 @@ public class SimulationCache
     private readonly WorldGenerationSettings worldSettings;
 
     private readonly Dictionary<(Species, SelectionPressure, Patch), float> cachedPressureScores = new();
-    private readonly Dictionary<(MicrobeSpecies, IBiomeConditions), EnergyBalanceInfo> cachedEnergyBalances = new();
+
+    private readonly Dictionary<(MicrobeSpecies, IBiomeConditions), EnergyBalanceInfoSimple>
+        cachedSimpleEnergyBalances = [];
+
     private readonly Dictionary<MicrobeSpecies, float> cachedBaseSpeeds = new();
     private readonly Dictionary<MicrobeSpecies, float> cachedBaseHexSizes = new();
 
@@ -69,27 +72,29 @@ public class SimulationCache
         return cached;
     }
 
-    public EnergyBalanceInfo GetEnergyBalanceForSpecies(MicrobeSpecies species, IBiomeConditions biomeConditions)
+    public EnergyBalanceInfoSimple GetEnergyBalanceForSpecies(MicrobeSpecies species,
+        IBiomeConditions biomeConditions)
     {
         // TODO: this gets called an absolute ton with the new auto-evo so a more efficient caching method (to allow
         // different species but with same organelles to be able to use the same cache value) would be nice here
         var key = (species, biomeConditions);
 
-        if (cachedEnergyBalances.TryGetValue(key, out var cached))
+        // TODO: check if caching instances of these objects would be better than always recreating
+        if (cachedSimpleEnergyBalances.TryGetValue(key, out var cached))
         {
             return cached;
         }
 
         var maximumMovementDirection = MicrobeInternalCalculations.MaximumSpeedDirection(species.Organelles);
 
-        cached = new EnergyBalanceInfo();
+        cached = new EnergyBalanceInfoSimple();
 
         // Auto-evo uses the average values of compound during the course of a simulated day
-        ProcessSystem.ComputeEnergyBalance(species.Organelles, biomeConditions, species.MembraneType,
+        ProcessSystem.ComputeEnergyBalanceSimple(species.Organelles, biomeConditions, species.MembraneType,
             maximumMovementDirection, true, species.PlayerSpecies, worldSettings, CompoundAmountType.Average, this,
             cached);
 
-        cachedEnergyBalances.Add(key, cached);
+        cachedSimpleEnergyBalances.Add(key, cached);
         return cached;
     }
 
@@ -373,7 +378,7 @@ public class SimulationCache
     public void Clear()
     {
         cachedPressureScores.Clear();
-        cachedEnergyBalances.Clear();
+        cachedSimpleEnergyBalances.Clear();
         cachedBaseSpeeds.Clear();
         cachedBaseHexSizes.Clear();
         cachedCompoundScores.Clear();
--- a/src/microbe_stage/EnergyBalanceInfoFull.cs
+++ b/src/microbe_stage/EnergyBalanceInfoFull.cs
@@ -1,82 +1,34 @@
 ﻿using System.Collections.Generic;
 
 /// <summary>
-///   Info regarding a microbe's energy balance in a patch
+///   Info regarding a microbe's energy balance in a patch, including consumption and production
 /// </summary>
 /// <remarks>
 ///   <para>
 ///     This does not take special modes the microbe can be into account, for example engulfing or binding
 ///   </para>
+///   <para>
+///     This inherits from <see cref="EnergyBalanceInfoSimple"/> whilst also tracking the consumption and production
+///   </para>
 /// </remarks>
-public class EnergyBalanceInfo
+public class EnergyBalanceInfoFull : EnergyBalanceInfoSimple
 {
     /// <summary>
     ///   The raw list of all energy consuming things related to the microbe. The key is the action name and the
     ///   value is the total consumption of that action.
     /// </summary>
-    public Dictionary<string, float> Consumption { get; } = new();
+    public Dictionary<string, float> Consumption { get; } = [];
 
     /// <summary>
     ///   The same as <see cref="Consumption"/> but for energy production
     /// </summary>
-    public Dictionary<string, float> Production { get; } = new();
+    public Dictionary<string, float> Production { get; } = [];
 
     /// <summary>
     ///   If setup to collect required compounds to run, then this collects compounds each <see cref="Production"/>
     ///   entry requires to produce the energy.
     /// </summary>
     public Dictionary<string, Dictionary<Compound, float>>? ProductionRequiresCompounds { get; private set; }
-
-    /// <summary>
-    ///   The cost of base movement (only when moving)
-    /// </summary>
-    public float BaseMovement { get; set; }
-
-    /// <summary>
-    ///   The cost of having all flagella working at the same time (only when moving)
-    /// </summary>
-    public float Flagella { get; set; }
-
-    /// <summary>
-    ///   The cost of having all cilia working at the same time at max rotation (only when rotating)
-    /// </summary>
-    public float Cilia { get; set; }
-
-    /// <summary>
-    ///   Sum of <see cref="BaseMovement"/>, <see cref="Flagella"/>, and <see cref="Cilia"/>
-    /// </summary>
-    public float TotalMovement { get; set; }
-
-    /// <summary>
-    ///   The total osmoregulation cost for the microbe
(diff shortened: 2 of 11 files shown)


src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -68,25 +72,27 @@ public float GetPressureScore(SelectionPressure pressure, Patch patch, Species s
...
 
-        if (cachedEnergyBalances.TryGetValue(key, out var cached))
+        if (cachedSimpleEnergyBalances.TryGetValue(key, out var cached))
         {
             return cached;
         }
 
         var maximumMovementDirection = MicrobeInternalCalculations.MaximumSpeedDirection(species.Organelles);
 
+        cached = new EnergyBalanceInfoSimple();
```
hhyyrylainen, line 88: Could you add a TODO comment here like `// TODO: check if caching instances of these objects would be better than always recreating`? I expect there to be no big difference but for future reference I'd like a comment here pointing out this possibility.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -193,20 +193,25 @@ public static Dictionary<string, OrganelleEfficiency> ComputeOrganelleProcessEff
...
+        ComputeEnergyBalance(result, organellesList, biome, membrane, maximumMovementDirection, includeMovementCost,
             isPlayerSpecies, worldSettings, amountType, calculateRequiredResources, null);
+        return result;
     }
 
     /// <summary>
     ///   Computes the energy balance for the given organelles in biome
     /// </summary>
+    /// <param name="result">
+    ///   The energy balance info that is returned. This may be an EnergyBalanceSimple or an EnergyBalanceFull.
```
hhyyrylainen, line 213: `<see cref=""/>` should be used when referring to code in the XML to make it navigable and Jetbrains will check to make sure the reference is valid.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -227,15 +232,11 @@ public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemp
     ///   If true, then the required input compounds to run at the given energy balance are stored per energy producer
     /// </param>
     /// <param name="cache">Auto-Evo Cache for speeding up the function</param>
-    public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemplate> organelles,
+    public static void ComputeEnergyBalance(EnergyBalanceInfoSimple result, IReadOnlyList<OrganelleTemplate> organelles,
```
hhyyrylainen, line 235: I think this method needs to be split into two even if that results in code duplication as I think it will be much cleaner to remove the dummy methods from the lite version of energy balance that do absolutely nothing, like `SetupTrackingForRequiredCompounds`. That's not optimal but I think it might be warranted in this very core piece of code that needs maximum performance. We just need to add big enough warning comments about the duplicated code that needs to all be changed if one copy is changed.

src/microbe_stage/EnergyBalanceInfoFull.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System.Collections.Generic;
+
+/// <summary>
+///   Info regarding a microbe's energy balance in a patch
```
hhyyrylainen, line 4: The summary comments should be made distinct between this and the simple variant.

src/microbe_stage/EnergyBalanceInfoSimple.cs
```diff
@@ -0,0 +1,79 @@
...
+    public virtual void AddConsumption(string groupName, float amount)
+    {
+    }
+
+    public virtual void AddProduction(string groupName, float amount,
+        Dictionary<Compound, float> requiredInputCompounds)
+    {
+    }
+
+    public virtual void SetupTrackingForRequiredCompounds()
```
hhyyrylainen, line 76: I'm quite suspicious of this kind of "trap" in the code where a method does absolutely nothing when used on an object in a specific state even though the name of the method is very clearly implying it should do something.

hhyyrylainen (changes requested):
This is already a very good start, but I kind of want the architecture split to go a bit deeper even if that means a bit of code duplication (but on balance removes the virtual methods).

hhyyrylainen:
Is there a reason why this was closed?

0HyperCube:
@hhyyrylainen the changes requested around avoiding the virtual methods would involve rewriting the PR. I'm happy to do this when I have some free time.

hhyyrylainen:
Yeah, a bit of refactoring would be needed, but I thought it wouldn't require rewriting all of the PR so I was wondering about the close.

> I'm happy to do this when I have some free time.

Totally fine. This isn't a super high priority thing so you can help out again once you have time.

0HyperCube:
@hhyyrylainen I believe that I have refactored this PR as desired?

src/microbe_stage/EnergyBalanceInfoFull.cs
```diff
@@ -0,0 +1,80 @@
...
+/// </summary>
+/// <remarks>
+///   <para>
+///     This does not take special modes the microbe can be into account, for example engulfing or binding
+///   </para>
+///   <para>
+///     This is a wrapper around <see cref="EnergyBalanceInfoEvaluated"/> also tracking the consumption and production
+///   </para>
+/// </remarks>
+public class EnergyBalanceInfoFull
```
hhyyrylainen, line 14: Would it be crazy to make this inherit `EnergyBalanceInfoEvaluated`? That way the code changes related to `balance.Evaluated.whatever` don't need to be done. So the rest of the codebase can be left more unchanged.

src/microbe_stage/EnergyBalanceInfoEvaluated.cs
```diff
@@ -0,0 +1,64 @@
...
+/// </summary>
+/// <remarks>
+///   <para>
+///     This does not take special modes the microbe can be into account, for example engulfing or binding
+///   </para>
+///   <para>
+///     The consumption and production is not tracked, use <see cref="EnergyBalanceInfoFull"/> to store it
+///   </para>
+/// </remarks>
+public class EnergyBalanceInfoEvaluated
```
hhyyrylainen, line 12: I don't really get why the name of this is `Evaluated` I'd name this like "simple" or "base" (if that other info is made to inherit from this).

hhyyrylainen:
Overall it's a lot better now but some of the naming stuff doesn't fully make sense to me and the doc comments didn't make it clear why such a name was chosen.

src/microbe_stage/EnergyBalanceInfoEvaluated.cs
0HyperCube: Since it contains the final evaluated balance rather than a dictionary of the contribution from each organelle.

src/microbe_stage/EnergyBalanceInfoFull.cs
0HyperCube: I have refactored this to avoid the inheritance as suggested.

src/microbe_stage/EnergyBalanceInfoEvaluated.cs
hhyyrylainen: But where is the non-evaluated variant? That's what's confusing me is that if that doesn't actually exist, then it is just confusing to specify that this class is "evaluated" when there isn't a non-evaluated variant.

src/microbe_stage/EnergyBalanceInfoFull.cs
hhyyrylainen: But like I only asked the inheritance to be removed *because of virtual methods*. I think that this pattern where there is some general info that many things need and then more specialized info is absolutely perfect to utilize inheritance here. It should simplify code a lot when there's no need to put `.Evaluated.` before property names you want to access from the full info. Or am I missing something?

0HyperCube: If you desire, I will refactor this again.

hhyyrylainen: I think using inheritance in that one place will make it much clearer how the code works and makes other code need fewer changes in this PR. So yes, I think it would be be a big improvement to refactor that.

src/microbe_stage/EnergyBalanceInfoEvaluated.cs
0HyperCube: I have renamed it for your benefit.

src/microbe_stage/EnergyBalanceInfoFull.cs
0HyperCube: I have refactored it all again for your benefit. Please note that I will not refactor it again. Thanks.

CI09: I mean, if we want to keep going the project for an amount of years to go, it needs to be pretty clean. And to avoid making performace related updates and post-cleanups, we might just as well do these stuff on implementation.
Here are some of my PRs, they are not clean of any change-requests either:
https://github.com/Revolutionary-Games/Thrive/pull/4553
https://github.com/Revolutionary-Games/Thrive/pull/4593
It just takes time to learn styleguide to be something that comes naturally.
In the end of the day it's up to you if you want to contribute more. You don't have to, you can, but not must.

0HyperCube: Hello @CI09. Having contributed to a couple open source projects before, I totally understand that it is up to the maintainer what code is merged. They are entitled to set whatever requirements that they desire. For a large project is is obviously necessary to set code quality requirements.

I have read and attempted to follow the style guide. I have also attempted to keep my code changes maintainable and well documented. However I have now rewritten the code twice due to uncommunicated requirements around the architecture. I found this rather tedious and don't really feel motivated enough by this change to do it again, considering that I haven't managed to measure any particular performance improvements from this. I hope that clarifies my position. Thanks.

hhyyrylainen: Sorry that the issue has been so unclear. This has been a case of where I didn't know what the optimal way to implement the change was when this was started. So work on this PR has been experimental in nature and searching for the right implementation rather than being able to immediately reach the goal. I realize I should have probably communicated that more clearly when asking for changes. I think the approach is now finally the right one. I'll help out by doing a few more polishing things and removing one added memory allocation in the latest refactoring (`float?` return type causes boxed objects to be created on each call of the method). Thanks for sticking out with it. 

I know it's not optimal to have these kinds of issues where the implementation path is not clear. Sadly the only way for me to have had a clear idea from the start on how to do this, I would have needed to like program already 90% of the change to validate what is a sensible approach. And by that point I might as well finish the changes. So unfortunately I don't really see a way to make the situation better. I can try to remember to write on issues when I don't have a clear implementation architecture in mind as a warning. I originally tagged the issue easy as I had the idea in my head that the balance calculation method would just be copy-pasted, which is not clean, but it was at least easy (and I would have accepted a PR doing that as long as it had warning comments slapped on about the duplication). The architecture changes required in this PR, if I had known in advance, I would have tagged the issue at least intermediate difficulty. If I didn't let you know about that easier path, I'm sorry.

> I have read and attempted to follow the style guide. I have also attempted to keep my code changes maintainable and well documented. However I have now rewritten the code twice due to uncommunicated requirements around the architecture.

I realize now that when you point that out our style guide should highlight that we are making a game at the top and the special considerations needed. Many of the changes you tried would have been totally fine for another type of program (for example if you sent it to https://github.com/Revolutionary-Games/RevolutionaryWebApp I would have had no problems with the use of virtual methods). So I think I will need to touch up our styleguide a bit to put some general points about Thrive *game* programming at the top (avoiding memory allocations, LINQ, virtual methods in hot paths, and overall considering the speed of code), and only then would the styleguide sections be. I've put at least most of those key points somewhere in the list of style rules but as they are it is pretty hard to find among the more specific style rules.

> considering that I haven't managed to measure any particular performance improvements from this.

That's about what I expected, because this was about improving memory allocations rather than running speed. Hopefully if everything is fine, this should result in measurable decrease in memory allocations in the auto-evo exploring tool. If you profiled memory allocations then sorry for misunderstanding. I'll skip on profiling mainly because I don't want to work on too much on Thrive today (as I need to get some stuff done on the ThriveDevCenter). Also I'm not going to reject this PR at this point even if the memory allocations aren't really reduced significantly.

> I hope that clarifies my position. Thanks.

It does. Thank you for the contribution and the insight on what we could do better to improve the contributor experience.

I'll merge this PR after pushing my last set of small changes and testing.

hhyyrylainen (approved):
This should now be all good to go.

PR #5786, opened 2025-01-01
Connections between patches and regions should be shown

### LINKED ISSUE #3556: Patches that link map regions together should be more obvious
as it is right now hard to tell which patch in a region you need to move to to be able to move outside the region. Maybe the region connection lines should continue all the way to the patches they actually connect?

Patryk26g:

This PR adds connections between the patches and regions sides. I increased the margin between the patches and also changed the color of the connections to gray (subject to change) because otherwise it would get a little bit to crowded with all of the connections.
image
image
Screenshot_20250101_131756

**Related Issues**

Closes #3556

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/PatchMapGenerator.cs
+++ b/src/microbe_stage/PatchMapGenerator.cs
@@ -533,6 +533,7 @@ public static class PatchMapGenerator
             {
                 region.Width += offset;
                 region.Height += offset;
+                region.Columns = region.Rows = 1;
 
                 var cave = region.Patches.FirstOrDefault(p => p.BiomeType == BiomeType.Cave);
 
@@ -542,15 +543,22 @@ public static class PatchMapGenerator
                 {
                     --waterPatchCount;
                     region.Width += offset;
+                    region.Columns += 1;
                 }
 
-                // Have 2 columns
+                // Have 3 columns
                 if (waterPatchCount > 1)
+                {
                     region.Width += offset;
+                    region.Columns += 1;
+                }
 
                 // Have 2 rows
                 if (waterPatchCount > 2)
+                {
                     region.Height += offset;
+                    region.Rows += 1;
+                }
 
                 break;
             }
@@ -560,12 +568,17 @@ public static class PatchMapGenerator
                 int verticalPatchCount = region.Patches.Count(IsWaterPatch);
 
                 region.Width += offset;
+                region.Columns += 1;
 
                 // If a cave or vent is present
                 if (verticalPatchCount != region.Patches.Count)
+                {
                     region.Width += offset;
+                    region.Columns += 1;
+                }
 
                 region.Height += verticalPatchCount * offset;
+                region.Rows += verticalPatchCount;
 
                 break;
             }
--- a/src/microbe_stage/editor/PatchMapDrawer.cs
+++ b/src/microbe_stage/editor/PatchMapDrawer.cs
@@ -12,6 +12,9 @@ public partial class PatchMapDrawer : Control
     public bool DrawDefaultMapIfEmpty;
 
     [Export(PropertyHint.ColorNoAlpha)]
+    public Color InterConnectionColor = Colors.DimGray;
+
+    [Export(PropertyHint.ColorNoAlpha)]
     public Color DefaultConnectionColor = Colors.ForestGreen;
 
     [Export(PropertyHint.ColorNoAlpha)]
@@ -33,7 +36,7 @@ public partial class PatchMapDrawer : Control
     /// <summary>
     ///   The representation of connections between regions, so we won't draw the same connection multiple times
     /// </summary>
-    private readonly Dictionary<Vector2I, Vector2[]> connections = new();
+    private readonly Dictionary<Vector2I, (Vector2[], bool, bool)> connections = new();
 
 #pragma warning disable CA2213
     private PackedScene nodeScene = null!;
@@ -171,12 +174,15 @@ public partial class PatchMapDrawer : Control
         // Create connections between regions if they dont exist.
         if (connections.Count == 0)
         {
+            // Clear existing connection lines
+            lineContainer.FreeChildren();
+
             CreateRegionLinks();
             RebuildRegionConnections();
         }
 
-        DrawRegionBorders();
         DrawPatchLinks();
+        DrawRegionBorders();
 
         // Scroll to player patch only when first drawn
         if (!alreadyDrawn)
@@ -384,13 +390,42 @@ public partial class PatchMapDrawer : Control
             pos.Y + Constants.PATCH_NODE_RECT_LENGTH * 0.5f);
     }
 
-    private Line2D CreateConnectionLine(Vector2[] points, Color connectionColor)
+    private static Vector2 ClosestPatchSideCenter(Vector2 patchPos, Vector2 pointPos, bool isVertical)
+    {
+        float halfLength = Constants.PATCH_NODE_RECT_LENGTH * 0.5f;
+        float posX = patchPos.X;
+        float posY = patchPos.Y;
+        List<Vector2> sideCenterPoints = new();
+        if (!isVertical)
+        {
+            // Right
+            sideCenterPoints.Add(new Vector2(posX + 2 * halfLength, posY + halfLength));
+
+            // Left
+            sideCenterPoints.Add(new Vector2(posX, posY + halfLength));
+        }
+        else
+        {
+            // Top
+            sideCenterPoints.Add(new Vector2(posX + halfLength, posY));
+
+            // Bottom
+            sideCenterPoints.Add(new Vector2(posX + halfLength, posY + 2 * halfLength));
+        }
+
+        return sideCenterPoints.OrderBy(point => point.DistanceSquaredTo(pointPos)).First();
+    }
+
+    private Line2D CreateConnectionLine(Vector2[] points, Color connectionColor, int zIndex = 1, float width = Constants.PATCH_REGION_CONNECTION_LINE_WIDTH)
     {
         var link = new Line2D
         {
             DefaultColor = connectionColor,
             Points = points,
-            Width = Constants.PATCH_REGION_CONNECTION_LINE_WIDTH,
+            Width = width,
+            BeginCapMode = Line2D.LineCapMode.Round,
+            EndCapMode = Line2D.LineCapMode.Round,
+            ZIndex = zIndex,
         };
 
         lineContainer.AddChild(link);
@@ -410,11 +445,6 @@ public partial class PatchMapDrawer : Control
         gradient.SetColor(reversed ? 2 : 0, color);
         gradient.SetColor(reversed ? 0 : 2, transparent);
         line.Gradient = gradient;
-    }
-
-    private void DrawNodeLink(Vector2 center1, Vector2 center2, Color connectionColor)
-    {
-        DrawLine(center1, center2, connectionColor, Constants.PATCH_REGION_CONNECTION_LINE_WIDTH, true);
     }
 
     private PatchMapNode? GetPatchNode(Patch patch)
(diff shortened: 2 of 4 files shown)


src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -533,6 +533,7 @@ private static void BuildRegionSize(PatchRegion region)
             {
                 region.Width += offset;
                 region.Height += offset;
+                region.Columns = region.Rows = 1;
```
hhyyrylainen, line 536: Does the columns and rows actually need to be stored like this? I get the impression that before the region patch positions was implicitly known based on the size, couldn't the same thing be done still (i.e. don't add the rows and columns variables but calculate them from the width and height of the region)?

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -33,7 +36,7 @@ public partial class PatchMapDrawer : Control
     /// <summary>
     ///   The representation of connections between regions, so we won't draw the same connection multiple times
     /// </summary>
-    private readonly Dictionary<Vector2I, Vector2[]> connections = new();
+    private readonly Dictionary<Vector2I, (Vector2[], bool, bool)> connections = new();
```
hhyyrylainen, line 39: I think this should use a named tuple here as otherwise it is very unclear what the two added booleans mean.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -384,13 +390,43 @@ private static Vector2 PatchCenter(Vector2 pos)
...
     {
         var link = new Line2D
         {
             DefaultColor = connectionColor,
             Points = points,
-            Width = Constants.PATCH_REGION_CONNECTION_LINE_WIDTH,
+            Width = width,
+            BeginCapMode = Line2D.LineCapMode.Round,
+            EndCapMode = Line2D.LineCapMode.Round,
+            ZIndex = zIndex,
```
hhyyrylainen, line 429: ZIndex can mess with other layers than purely just this. For example even the pause menu might not work correctly. So I'd really prefer an approach where ZIndex is not used. Reordering the children in the line container can be used to set the draw order. If just a couple of ZIndex values are needed, they could be made as separate containers for the lines. Or what is it needed for? I can probably come up with some better suggestions if I know that.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -472,52 +503,166 @@ private void CreateRegionLinks()
...
+        var lower = startRect.End.Y > endRect.End.Y ? startRect : endRect;
+        var left = startRect.Position.X < endRect.Position.X ? startRect : endRect;
+        var right = startRect.End.X > endRect.End.X ? startRect : endRect;
+        Vector2 intermediate1;
+        Vector2 intermediate2;
+        int priority = 0;
+        switch (shape)
+        {
+            case "lower":
+                intermediate1 = new Vector2(startCenter.X, lower.End.Y + i * 50);
```
hhyyrylainen, line 587: Case statements with multiple lines should be wrapped in braces according to our styleguide.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -892,22 +1041,47 @@ private void AddPatchNode(Patch patch, Vector2 position)
...
+        bool isVertical, MapElementVisibility regionVisibility)
     {
-        // Clear existing connection lines
-        lineContainer.FreeChildren();
+        foreach (var adjacent in region1.PatchAdjacencies[region2.ID])
+        {
+            if (adjacent.Visibility == MapElementVisibility.Hidden ||
+                (adjacent.Visibility == MapElementVisibility.Unknown &&
+                    regionVisibility == MapElementVisibility.Unknown))
+                continue;
```
hhyyrylainen, line 1060: Multiline if statements should always use braces according to the styleguide.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -472,52 +503,166 @@ private void CreateRegionLinks()
...
     /// <returns>Path represented in a Vector2 array</returns>
-    private Vector2[] GetLeastIntersectingPath(PatchRegion start, PatchRegion end)
+    private (Vector2[] Path, bool StartVertical, bool EndVertical) GetLeastIntersectingPath(PatchRegion start,
+        PatchRegion end)
     {
         var startCenter = RegionCenter(start);
         var startRect = new Rect2(start.ScreenCoordinates, start.Size);
         var endCenter = RegionCenter(end);
         var endRect = new Rect2(end.ScreenCoordinates, end.Size);
+        int priority;
```
hhyyrylainen, line 635: Variables should be defined as close to where they are initialized as possible (so not all at the start of a method like ancient C styles mandated).

CountingBeeps:
This is a great start, but I don't always like how these look. I think the visuals could use a little work before this is merged. Specifically in this case:
image

hhyyrylainen:
I did a partial review of this PR. I left a big chunk of the code unreviewed but it was probably mostly just the path algorithm changes that caused the internal API to change, which looks long but isn't actually that complicated of a change...

This is already in pretty good shape I think overall design wise.

Anyway I'd like to first get the biggest issues out of the way before reviewing everything in here:
- The columns and rows variables which will break save compatibility
- Use of zindex in the line rendering on the map

Ones those are taken care of I think this will be close to done already. (though I just saw @CountingBeeps comment)

Patryk26g:
> This is a great start, but I don't always like how these look. I think the visuals could use a little work before this is merged. Specifically in this case: image

It should look better now

Patryk26g:
> I did a partial review of this PR. I left a big chunk of the code unreviewed but it was probably mostly just the path algorithm changes that caused the internal API to change, which looks long but isn't actually that complicated of a change...
> 
> This is already in pretty good shape I think overall design wise.
> 
> Anyway I'd like to first get the biggest issues out of the way before reviewing everything in here:
> 
> * The columns and rows variables which will break save compatibility
> * Use of zindex in the line rendering on the map
> 
> Ones those are taken care of I think this will be close to done already. (though I just saw @CountingBeeps comment)

I have removed z-indeces and the column and row variables. I also adjusted the look of the patches

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -472,52 +504,183 @@ private void CreateRegionLinks()
...
+        var upper = startRect.Position.Y < endRect.Position.Y ? startRect : endRect;
+        var lower = startRect.End.Y > endRect.End.Y ? startRect : endRect;
+        var left = startRect.Position.X < endRect.Position.X ? startRect : endRect;
+        var right = startRect.End.X > endRect.End.X ? startRect : endRect;
+        Vector2 intermediate1;
+        Vector2 intermediate2;
+        int priority = 0;
+        switch (shape)
+        {
+            case "lower":
```
hhyyrylainen, line 583: I think this should be an enum as using strings with magic values like this isn't very good style.
Also the CI has found unused variables so, could you please fix those before I take a full new look at the code?

Patryk26g: I updated the code

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -384,15 +401,43 @@ private static Vector2 PatchCenter(Vector2 pos)
             pos.Y + Constants.PATCH_NODE_RECT_LENGTH * 0.5f);
     }
 
-    private Line2D CreateConnectionLine(Vector2[] points, Color connectionColor)
+    private static Vector2 ClosestPatchSideCenter(Vector2 patchPos, Vector2 pointPos, bool isVertical)
+    {
+        float halfLength = Constants.PATCH_NODE_RECT_LENGTH * 0.5f;
+        float posX = patchPos.X;
+        float posY = patchPos.Y;
+        List<Vector2> sideCenterPoints = new();
```
hhyyrylainen, line 409: This method should be rewritten to not use this temporary memory. As we are working on a realtime game, memory allocations should be avoided whenever possible. Here it looks like it will be possible by writing more code to manually compare the values to get the actual single result value.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -472,52 +512,181 @@ private void CreateRegionLinks()
...
 
+    private int AdjustPriority(int size1, int priority)
+    {
+        // Prioritize connections to the longer sides
+        priority += size1 * size1;
+
+        return priority;
+    }
+
+    private int GetIShapePathCharacteristics(int startRows, int startColumns, int endRows, int endColumns,
```
hhyyrylainen, line 532: Shouldn't the name be `GetIShapePathPriority`? As this returns the priority and not some other unspecified characteristic.

Also the the other "characteristics" methods seem to have this same problem with them.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -472,52 +512,181 @@ private void CreateRegionLinks()
...
+    }
+
+    private (Vector2 Intermediate1, Vector2 Intermediate2, int Priority) GetZShapePathCharacteristics(int startRows,
+        int startColumns, int endRows, int endColumns, Vector2 startCenter, Vector2 endCenter,
+        Vector2 middlePoint, bool isTrueMiddle, bool flip = false)
+    {
+        Vector2 intermediate1;
+        Vector2 intermediate2;
+        int priority = isTrueMiddle ? 2 : 1;
+        if (!flip)
```
hhyyrylainen, line 558: I'd add a blank line here to give better readability for the if (as the variables above this aren't used in the if condition, I'd logically separate them with a blank line)

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -472,52 +512,181 @@ private void CreateRegionLinks()
...
+        Vector2 endCenter, Side side, int i)
+    {
+        var upper = startRect.Position.Y < endRect.Position.Y ? startRect : endRect;
+        var lower = startRect.End.Y > endRect.End.Y ? startRect : endRect;
+        var left = startRect.Position.X < endRect.Position.X ? startRect : endRect;
+        var right = startRect.End.X > endRect.End.X ? startRect : endRect;
+        Vector2 intermediate1;
+        Vector2 intermediate2;
+        int priority = 0;
+        switch (side)
```
hhyyrylainen, line 587: Same here with the empty line. I like to have room in the code for it to breathe rather than mushing all lines together like empty lines cost money.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -472,52 +512,181 @@ private void CreateRegionLinks()
...
 
-                connections.Add(connectionKey, pathToAdjacent);
+                connections.Add(connectionKey, (pathToAdjacent, startVertical, endVertical));
             }
         }
 
         AdjustPathEndpoints();
     }
 
+    private int AdjustPriority(int size1, int priority)
```
hhyyrylainen, line 524: For understandability I'd add a general comment before this group of priority methods for what the priority means, something like:

```
// The next few methods are for calculating path priorities. They are needed because ...
```

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -472,52 +512,181 @@ private void CreateRegionLinks()
...
         var endRect = new Rect2(end.ScreenCoordinates, end.Size);
+        Vector2 intermediate1;
+        Vector2 intermediate2;
+
+        const float offset = Constants.PATCH_NODE_RECT_LENGTH + Constants.PATCH_AND_REGION_MARGIN;
+        const float margin = Constants.PATCH_REGION_BORDER_WIDTH + Constants.PATCH_AND_REGION_MARGIN;
+        int startRows = (int)Math.Round((start.Height - margin) / offset);
+        var startColumns = (int)Math.Round((start.Width - margin) / offset);
+        int endRows = (int)Math.Round((end.Height - margin) / offset);
+        var endColumns = (int)Math.Round((end.Width - margin) / offset);
```
hhyyrylainen, line 658: Why are int and var used inconsistently in this block of variables?

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -526,45 +695,61 @@ private Vector2[] GetLeastIntersectingPath(PatchRegion start, PatchRegion end)
         var right = startRect.End.X > endRect.End.X ? startRect : endRect;
 
         // 3-segment line, Z shape
-        var middlePoint = new Vector2(left.End.X + right.Position.X, upper.End.Y + lower.Position.Y) / 2.0f;
+        for (int i = -3; i <= 3; ++i)
+        {
+            var middlePoint =
+                new Vector2(left.End.X + right.Position.X + i * 25, upper.End.Y + lower.Position.Y + i * 25) / 2.0f;
```
hhyyrylainen, line 701: The 25 here and the 3 few lines below seem like magic numbers I have no clue about, could you make those named const values in this class to clarify their reason why they are like that?

hhyyrylainen, line 701: Also I know that the old code did this, but could you update any `/ 2.0f` pieces of code to be `* 0.5f` where they are safe to change? Floating point multiplies are faster to calculate than divides so it should be preferred whenever a divide is not absolutely required.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -899,15 +1086,48 @@ private void AddPatchNode(Patch patch, Vector2 position)
...
+                continue;
+            }
+
+            bool highlight = (ContainsSelectedPatch(region1) && SelectedPatch?.Adjacent?.Contains(adjacent) == true) ||
+                (ContainsSelectedPatch(region2) && SelectedPatch?.ID == adjacent.ID);
+            var closesPatchSideCenter = ClosestPatchSideCenter(adjacent.ScreenCoordinates, regionPoint, isVertical);
+            Vector2[] path = [PatchCenter(adjacent.ScreenCoordinates), closesPatchSideCenter, regionPoint];
+
+            // If the path is highlighted then create it later to be on top of other paths
+            if (!highlight)
```
hhyyrylainen, line 1107: Multiline ifs should always use braces according to our styleguide.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -917,15 +1137,18 @@ private void RebuildRegionConnections()
                 continue;
 
             // Check if connections should be highlighted
-            // (Unknown patches should not highlight connections)
```
hhyyrylainen, line 920: What's the reason behind this change, making unknown patches now be able to highlight?

hhyyrylainen (changes requested):
I think overall architecture-wise this is now in good shape. I did a full readthrough of the code and I commented on a few places that could do with some improved clarity or slight changes to match our styleguide. I didn't test in game yet but I liked how the visual screenshots looked. Even if the extra space looks unfamiliar, but to avoid like overlaps it looks like the empty space is required.

src/microbe_stage/editor/PatchMapDrawer.cs
Patryk26g: That is true but I wanted to be consistent in all these paths, but I will change it to priority

src/microbe_stage/editor/PatchMapDrawer.cs
Patryk26g: I think that it now makes more sense since it is connected anyway and user can click on it

Patryk26g: image
Like, in my opinion if I click this patch then the connections should be highlighted even if they fade away

hhyyrylainen: Yeah, that makes sense to me (the code was probably originally made for things to not look weird when something was partially visible). BTW that screenshot shows some aliasing artifacts, is the antialias option on for the lines already or not? If it isn't on already it might help a bit with the visuals.

src/microbe_stage/editor/PatchMapDrawer.cs
hhyyrylainen: The change you did is slightly different in semantics, an XML comment *only* applies to the language element that it is before. So now this comment only applies to AdjustPriority. So I suggest it is reformatted to be a normal comment `//`  instead of `/// <summary>` and a blank line to be added after it and before AdjustPriority to make it clear the comment applies to multiple things and not just this one method.

hhyyrylainen:
I've now tested this in-game and the results are really good.

I tried that antialias option and it doesn't really seem to do much, sadly. So I wouldn't turn it on.

As I tested this in-game I noticed there's some pretty unclear situations. In these kind of situations I think it would be more clear and hopefully easy to make the links between patches highlight whenever a patch is selected (so not just the between region links):

kuva

Patryk26g:
Do not review it yet, I'm not finished

Patryk26g:
Ready for testing now

hhyyrylainen (approved):
I think the code is still probably fine since I last looked, and the visuals have been tweaked a lot already. There is one known problem but I'll open a separate issue for that after merging.

hhyyrylainen:
The follow up issue: https://github.com/Revolutionary-Games/Thrive/issues/5822

PR #5788, opened 2025-01-06
Fix lost click event when opening popup

### LINKED ISSUE #5217: Fix needing to click twice sometimes in the editor to get the mouse click to register (after canceling an action / popup)

0HyperCube:

When opening a popup such as the organelle context menu, the mouse release event will be marked as handled by the popup. This means that the input manager won't call the OnInput function:
https://github.com/Revolutionary-Games/Thrive/blob/dce65afc2961d5c00c1b17b3227a01834f03a15f/src/engine/input/InputManager.cs#L573-L575
This caused the HeldDown field of RunOnKeyAttribute to get stuck in the true state.

I changed the RunOnKeyDownAttribute to rely on `Input.IsActionJustPressed` rather than the `HeldDown` state to avoid this issue.

**Related Issues**
I think this closes #5217?

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/input/RunOnKeyDownAttribute.cs
+++ b/src/engine/input/RunOnKeyDownAttribute.cs
@@ -18,11 +18,8 @@ public class RunOnKeyDownAttribute : RunOnKeyAttribute
 
     public override bool OnInput(InputEvent @event)
     {
-        // Only trigger if the input was pressed this frame. It isn't possible to use !HeldDown since its value is
-        // incorrect if a mouse release event is marked as handled e.g. when opening a popup.
-        // https://github.com/Revolutionary-Games/Thrive/issues/5217
-        var justPressed = Input.IsActionJustPressed(InputName);
-        if (base.OnInput(@event) && justPressed && HeldDown)
+        var before = HeldDown;
+        if (base.OnInput(@event) && !before && HeldDown)
         {
             if (TrackInputMethod)
             {
@@ -41,5 +38,13 @@ public class RunOnKeyDownAttribute : RunOnKeyAttribute
 
     public override void OnProcess(double delta)
     {
+    }
+
+    public override void OnUnprocesedInput(InputEvent @event)
+    {
+        // Update the state when the key is released, even if the event is marked as handled.
+        // This can occur for example when a popup is opened when the key is pressed.
+        if (@event.IsActionReleased(InputName, false))
+            HeldDown = false;
     }
 }
--- a/src/engine/input/RunOnKeyDownAttribute.cs
+++ b/src/engine/input/RunOnKeyDownAttribute.cs
@@ -18,8 +18,11 @@ public class RunOnKeyDownAttribute : RunOnKeyAttribute
 
     public override bool OnInput(InputEvent @event)
     {
-        var before = HeldDown;
-        if (base.OnInput(@event) && !before && HeldDown)
+        // Only trigger if the input was pressed this frame. It isn't possible to use !HeldDown since its value is
+        // incorrect if a mouse release event is marked as handled e.g. when opening a popup.
+        // https://github.com/Revolutionary-Games/Thrive/issues/5217
+        var justPressed = Input.IsActionJustPressed(InputName);
+        if (base.OnInput(@event) && justPressed && HeldDown)
         {
             if (TrackInputMethod)
             {
--- a/src/engine/input/InputAttribute.cs
+++ b/src/engine/input/InputAttribute.cs
@@ -59,6 +59,19 @@ public abstract class InputAttribute : Attribute
     public abstract bool OnInput(InputEvent input);
 
     /// <summary>
+    ///   Called for input that is not processed by the <see cref="OnInput"/> method.
+    /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     This is useful for updating state such as HeldDown which would otherwise become incorrect.
+    ///   </para>
+    /// </remarks>
+    /// <param name="input">The event fired by the user</param>
+    public virtual void OnUnprocesedInput(InputEvent input)
+    {
+    }
+
+    /// <summary>
     ///   Processes input actions that aren't triggered on key events directly
     /// </summary>
     /// <param name="delta">The time since the last call of OnProcess</param>
--- a/src/engine/input/InputManager.cs
+++ b/src/engine/input/InputManager.cs
@@ -582,6 +582,10 @@ public partial class InputManager : Node
                         break;
                 }
             }
+            else
+            {
+                attribute.OnUnprocesedInput(@event);
+            }
         }
 
         // Define input as consumed to Godot if something reacted to it


src/engine/input/RunOnKeyDownAttribute.cs
```diff
@@ -18,8 +18,11 @@ public RunOnKeyDownAttribute(string inputName) : base(inputName)
 
     public override bool OnInput(InputEvent @event)
     {
-        var before = HeldDown;
-        if (base.OnInput(@event) && !before && HeldDown)
+        // Only trigger if the input was pressed this frame. It isn't possible to use !HeldDown since its value is
+        // incorrect if a mouse release event is marked as handled e.g. when opening a popup.
```
hhyyrylainen, line 22: Wouldn't the actual fix be to make sure that up events are *always* handled? Because that's how I tried to design the input system in the first place. I'm sure I put in at least some code that should feed handled inputs that are release type events into the system so that the held down statuses wouldn't get stuck.

hhyyrylainen (changes requested):

src/engine/input/RunOnKeyDownAttribute.cs
0HyperCube: As far as I can see, `HeldDown` is only set to false in the `OnInput` and `FocusLost` methods. `OnInput` is only called if the input is unhandled (or the attribute has `OnlyUnhandled`). 'FocusLost' is only called when the window has focus lost.

Could you please point me towards the existing code? Thanks.

hhyyrylainen: That's probably exactly the bug, `OnInput` only passes events it sees but it has code that further key handlers are given all events if they are up events. So this situation likely needs a bit of refactoring so that handled up events would still be fed as up events into all the listeners (so I'd split OnInput into two parts, one that does the normal stuff and one that has just the event forwarding, and then I'd pass handled events that are key up events to the  second part of the method).

0HyperCube: I'm struggling to implement your proposed solution.

The [Godot input code](https://docs.godotengine.org/en/stable/tutorials/inputs/inputevent.html#how-does-it-work) calls the `_Input()` method before the `_UnhandledInput()`. This means that inputs that end up being unhandled are processed twice: first as a handled input and then as an unhandled input. You proposed to update the `HeldDown` state for handled inputs. Therefore `HeldDown` would never be updated for the unhandled input event (due to the aforementioned double event processing).

Therefore I think that it is not possible to rely on the `HeldDown` state to determine if the key is pressed. The only solution I have thought of is the `Input.IsActionJustPressed` (as implemented in this PR). Although perhaps I have misunderstood your proposed approach?

hhyyrylainen: Sorry I don't fully understand what you mean. Cannot the input method be modified to *only* process key up events? (and for efficiency the unhandled input could be made to ignore up events to not process them twice)

0HyperCube: I'm not sure I understand? If only key up events are processed, then it is not possible to run a method on key down.

hhyyrylainen: That would be true if only key up events were processed but my suggestion is:
- Add a new method that runs for all key events that are released (even consumed ones)
- Keep the old method that runs on not consumed events

That way passing the events from that new method and the old method to a handling place, would correctly handle both situations. 

When I talked about only processing key up events it was specifically for that new listener, because otherwise it would totally break the input system design as it would react to also consumed input.

0HyperCube: Wouldn't that approach break the `RunOnKeyUpAttribute`? Since both that and the key down one inherit from `RunOnKeyAttribute` which actually handles the `HeldDown` state.

Although I suppose I can change it so that the `HeldDown` state is updated only in the child classes.

hhyyrylainen: I did foresee that but I think it is probably better to check if anything would break and then add a big warning for the key up that it will trigger even for consumed inputs so it needs to be used carefully.

0HyperCube: I propose that it is better to make all of the attributes work properly by using the `Input.IsActionJustPressed` and Input.IsActionJustReleased`.

However I also see it is probably advantageous to have the `HeldDown` state correct, which would involve calling the `OnInput` for handled input but have it only dispatch the methods if the input is unhandled.

hhyyrylainen: > I propose that it is better to make all of the attributes work properly by using the Input.IsActionJustPressed and Input.IsActionJustReleased`.

I don't prefer this approach mainly because then I would need to review a big PR with a lot of potential implications all over the codebase. That way I suggested it works with much more minimal changes. And I think that it is already the case that events that are up are already fed to all of the input listeners so it wouldn't be that big of a change to also send GUI-consumed inputs if they are up events to everywhere.

hhyyrylainen:
Are you remaking this PR or why was this closed? Or should I try to make the changes based on my idea that I *think* will work?

0HyperCube:
When I was first trying to implement your approach, I was unable to make the KeyUpAttribute work at all (breaking functionality such as ending sprinting). This was because I had added the unprocessed event listener to the `RunOnKeyAttribute`. Therefore the event ordering (handled input events always precede unhandled input events) made it so HeldDown was false before the OnInput in the key up attribute could run.

I have now realized that it is necessary to put this method into the `RunOnKeyDownAttribute` only.

If this is still not the desired fix, I think maybe it is beneficial for you to make the changes yourself since I'm still rather struggling to understand what you have in mind. Thanks for your time working with me on this.

hhyyrylainen:
Reading the code I realized there was already that "always grab events" method. So my suggestion actually required just adding `|| !isDown` to one if condition. I've made a commit of that (plus some documentation changes, as well as I verified that all of the usages of the key up attribute should be fine after the change):

https://github.com/Revolutionary-Games/Thrive/commit/b816b862f627a5f036185077c4ff0b1f65ce186f

Let me know what you think of that fix. I only very quickly checked that the organelle context menu seemed to not mess with the placement clicks, but I didn't test anywhere close to long enough to know for sure that my fix would work.

If that works I'll merge my version to master through a separate PR (as I didn't end up using any of your changes). Thanks for working on this with me.

0HyperCube:
Your fix seems to work well @hhyyrylainen. It does obviously mean the equivalent of `OnlyUnhandled = false` for all key up attributes, however that doesn't appear to have any effect on the functionality.

hhyyrylainen:
Yeah. Like I said before the input system was already internally setup to pass key up events to everything, so that change makes it just pass a few more up events to everything. So logically the change is not that much, but I did check all the uses of the key up attribute to confirm they are correctly written. Thanks for double checking my work.

PR #5794, opened 2025-01-08
Add Banana Biome

HexapodPhilosopher:

Adds the banana biome as a rare replacement for coastal patches.
The biome has a much more yellow background, and 4 different banana-shaped chunks. Features a lot of marine snow, all of which is Nitroplasts.

Adds a new profile in biomes.json
Alters existing banana rock model, and adds 2 smaller models.

* Frequency of patch appearing needs to be reset to 1/100 before merging.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -825,4 +838,39 @@ private static LocalizedString GetPatchLocalizedName(string regionName, string b
     {
         return new LocalizedString("PATCH_NAME", regionName, new LocalizedString(biomeType));
     }
+
+    private static void AddBananaBiome(PatchMap map, Random random)
+    {
+        // Check Easter eggs
+        if (!ICurrentGameInfo.CurrentGame.GameWorld.WorldSettings.EasterEggs)
```
hhyyrylainen, line 845: This is the line that causes this PR to not compile. The problem is that you are trying to access `CurrentGame` but you specified the interface name `ICurrentGameInfo` instead of a variable holding an instance of an object. This doesn't work as to access a property you need an instance of an object type not the type itself that defines how that object looks like. So here you would need to get the current game properties to access this data.

Though in the map generator the current game has not been created yet. Instead the `Generate` method receives `WorldGenerationSettings settings` as a parameter, which will have this info. So that needs to be used to check the easter egg status either in this method or maybe this method could be not called in the first place if easter eggs are not enabled.

assets/models/easter_eggs/BigBanan.shape
hhyyrylainen, line 1: Is the typo in the filename intentional or?

simulation_parameters/microbe_stage/biomes.json
```diff
@@ -3119,6 +3119,239 @@
       }
     }
   },
+  "banana": {
+    "Name": "COASTAL",
```
hhyyrylainen, line 3123: Shouldn't the name be different so people who aren't super familiar with the banana legend would still be able to realize the easter egg? (something like `BANANA_BIOME`)

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -27,6 +27,20 @@ public static PatchMap Generate(WorldGenerationSettings settings, Species defaul
...
         Patch? vents = null;
         Patch? tidepool = null;
+        Patch? banana = null;
+
+        bool useRarePatch = false;
+
+        if (settings.EasterEggs)
+        {
+            // 1% chance for rare biome to spawn (!!CURRENTLY SET TO 1 IN 3 FOR TESTING!!)
+            int rand = random.Next(0, 3);
```
hhyyrylainen, line 37: I'll put this random call outside the if check to ensure that changing the easter eggs setting doesn't cause the entire patch map to become different with the same seed.

hhyyrylainen (approved):
I did a little bit of code renaming and slight improving of the code, should be good to merge now.

PR #5820, opened 2025-01-16
Implement a multicellular statistics panel

### LINKED ISSUE #3170: Create an organism statistics panel for the body plan editor
right now the space used for it in the cell editor is blank. Some kind of panel should be created there showing some overall multicellular species statistics. And probably also total compound balance of the colony. And maybe some kind of ATP balance bar?

dligr:

Implements a multicellular statistics panel, mostly porting features from the similar panel in the cell editor.
In terms of stats, it only features speed, storage and rotatation speed as the other microbial stats don't make sense for a colony.

image

TODO:
- Add daylight configuration, similar to the cell editor
- Make the process panel display all cells' processes (could also help address #3207)
- Add a warning if the player's organism doesn't produce enough ATP to survive

**Related Issues**

Closes #3170

**Progress Checklist**

      break existing features:
      members reviewing this PR)


dligr:
While this PR is not complete, it would be good to hear some feedback regarding the general architecture, especially the changes made to the ProcessSystem to make the energy balance calculation work.

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -54,45 +54,36 @@ public static float GetTotalNominalCapacity(IEnumerable<OrganelleTemplate> organ
...
 
-        var capacities = new Dictionary<Compound, float>();
+        return capacities;
+    }
 
-        // Update the variant of this logic in UpdateSpecificCapacities if changes are made
-        for (int i = 0; i < count; ++i)
-        {
-            var organelle = organelles[i];
+    public static Dictionary<Compound, float> GetTotalSpecificCapacity(IEnumerable<CellTemplate> cells,
```
hhyyrylainen, line 67: I think this is different enough in purpose / stage this belongs in that I'd move this to a new file in `src/multicellular_stage/CellBodyPlanInternalCalculations.cs`

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -796,4 +787,31 @@ private static Vector3 ChooseFromSymmetricFlagella(IReadOnlyList<OrganelleTempla
...
+
+            // If this is updated the code in CompoundBag.AddSpecificCapacityForCompound must also be updated
+            if (capacities.TryGetValue(specificCapacity.Compound, out var currentCapacity))
+            {
+                capacities[specificCapacity.Compound] = currentCapacity + specificCapacity.Capacity;
+            }
+            else
+            {
+                capacities.Add(specificCapacity.Compound, specificCapacity.Capacity + totalNominalCap);
+            }
```
hhyyrylainen, line 814:
```suggestion
            capacities.TryGetValue(specificCapacity.Compound, out var currentCapacity);
            capacities[specificCapacity.Compound] = currentCapacity + specificCapacity.Capacity;
```
Can't this be simplified as that?

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
```diff
@@ -0,0 +1,345 @@
...
+using System.Text;
+using Godot;
+using Systems;
+
+/// <summary>
+///   The partial class containing GUI updating actions
+/// </summary>
+public partial class CellBodyPlanEditorComponent
+{
+    private StringBuilder atpToolTipTextBuilder = new();
```
hhyyrylainen, line 14: Should this be readonly?

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -227,18 +228,25 @@ public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemp
     ///   If true, then the required input compounds to run at the given energy balance are stored per energy producer
     /// </param>
     /// <param name="cache">Auto-Evo Cache for speeding up the function</param>
+    /// <param name="result">
+    ///   Energy balance info to add the new info to. To be used for calculating multiple cells' balances
```
hhyyrylainen, line 232: Would this incidentally do anything related to https://github.com/Revolutionary-Games/Thrive/issues/5426 ?

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -227,18 +228,25 @@ public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemp
...
+    /// </param>
     public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemplate> organelles,
```
hhyyrylainen, line 234: I think the return value should be removed and this refactored so that the balance info is mandatory value to pass in. That way the usage would be clearer.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -294,33 +302,36 @@ public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemp
             hexCount += organelle.Definition.HexCount;
         }
 
-        result.BaseMovement = Constants.BASE_MOVEMENT_ATP_COST * hexCount;
+        float baseMovement = Constants.BASE_MOVEMENT_ATP_COST * hexCount;
```
hhyyrylainen, line 305: I'd use `var` for all the local variables here so that there can't be any future problems about changing types of things in the math.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -350,9 +361,9 @@ public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemp
     /// </remarks>
     public static Dictionary<Compound, CompoundBalance> ComputeCompoundBalance(
         IEnumerable<OrganelleDefinition> organelles, IBiomeConditions biome, CompoundAmountType amountType,
-        bool requireInputCompoundsInBiome)
+        bool requireInputCompoundsInBiome, Dictionary<Compound, CompoundBalance>? result = null)
```
hhyyrylainen, line 364: Also same here and multiple methods afterwards: I think the result parameter should be required so that the callsites can then more intelligently be fixed to reuse objects (like is the goal for improving the game allocation counts).

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
```diff
@@ -0,0 +1,345 @@
...
+        if (!storageLabel.IsToolTipRegistered(tooltip))
+            storageLabel.RegisterToolTipForControl(tooltip, true);
+
+        var description = new LocalizedStringBuilder(100);
+
+        bool first = true;
+
+        var simulationParameters = SimulationParameters.Instance;
+
+        foreach (var entry in storage)
```
hhyyrylainen, line 48: I think that a lot of the stuff here is ending up very similar than in the microbe stage so I think a general base class should be created for microbe editor and the cell body plan editor that could contain the code for generating this tooltip, the ATP comparer class, the special cases in balance tooltip handling, and UpdateEnergyBalance.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -477,13 +549,96 @@ public bool ShowCellOptions()
         return true;
     }
 
-    protected CellType CellTypeFromName(string name)
+    public float CalculateSpeed()
     {
-        return Editor.EditedSpecies.CellTypes.First(c => c.TypeName == name);
+        var leader = editedMicrobeCells[0].Data!;
```
hhyyrylainen, line 554: This should also go into the new cell body plan calculation helpers class as this is so long and I assume auto-evo will also need this info.

hhyyrylainen:
> especially the changes made to the ProcessSystem to make the energy balance calculation work.

I think that's fundamentally a fine change but I think it needs a bit more tweaking. I had a quick look at most of  the code and I think a bigger architecture problem is that a lot of the quite complex calculation logic for the GUI display seems to be mostly duplicated from the microbe version. So finding ways to avoid the code duplication would be excellent.

src/microbe_stage/systems/ProcessSystem.cs
dligr: It shouldn't, since the method's internal logic wasn't changed, apart from making energy balance stats additive

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
dligr: Perhaps it would be better to make a single `OrganismStatistics` class and move all related GUI logic out of the editor components? The latter would then just calculate the data (the calculations are already very different in the two classes) and call `OrganismStatistics`' methods to update the shown stats. The light level buttons and the display options would send signals for the editor component to recalculate the data.

src/microbe_stage/systems/ProcessSystem.cs
hhyyrylainen: But still technically wouldn't this (after the few refactorings I suggested), make it much easier to do that other issue? As things would already be almost perfectly setup for reusing the data objects and avoiding allocations.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
hhyyrylainen: That actually seems like a very good design (even better than the idea I had).

hhyyrylainen:
Hopefully my PR doesn't cause conflicts with this, but it might: https://github.com/Revolutionary-Games/Thrive/pull/5829

dligr:
Now that the architecture was updated (i.e. the organism statistics panel made into its own class) and the TODO list completed, I think that this PR is ready for review.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -227,18 +228,23 @@ public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemp
...
-        // further items
-        var result = new EnergyBalanceInfo();
-
-        if (calculateRequiredResources)
+        if (result == null)
         {
-            result.SetupTrackingForRequiredCompounds();
+            result = new EnergyBalanceInfo();
+
+            if (calculateRequiredResources)
```
hhyyrylainen, line 244: I think it would be a lot clearer to remove the `calculateRequiredResources` parameter from this method as the parameter *only* works if the result passed in is null, so that's a very tricky and error-prone usage of this method.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -404,11 +409,11 @@ public static Dictionary<Compound, CompoundBalance> ComputeCompoundBalance(
...
     ///   </para>
     /// </remarks>
-    public static Dictionary<Compound, CompoundBalance> ComputeCompoundBalanceAtEquilibrium(
+    public static void ComputeCompoundBalanceAtEquilibrium(
         IEnumerable<OrganelleDefinition> organelles, IBiomeConditions biome, CompoundAmountType amountType,
-        EnergyBalanceInfo energyBalance)
+        EnergyBalanceInfo energyBalance, Dictionary<Compound, CompoundBalance>? result)
     {
-        var result = new Dictionary<Compound, CompoundBalance>();
+        result ??= new Dictionary<Compound, CompoundBalance>();
```
hhyyrylainen, line 416: Same problem with the ref here, for clarity I would make the result in this method chain non-nullable because it doesn't make that much sense why a caller wouldn't be able to allocate a dictionary for the results.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.tscn
```diff
@@ -330,20 +401,130 @@ CannotDeleteInUseTypeDialogPath = NodePath("CannotDeleteInUseType")
...
+tooltip_text = "SUNLIGHT"
+texture = ExtResource("5_e4mej")
+expand_mode = 1
+stretch_mode = 5
+
+[node name="Control" type="Control" parent="TopPanel/PanelContainer/MarginContainer/HBoxContainer"]
+custom_minimum_size = Vector2(10, 0)
+layout_mode = 2
+
+[node name="CurrentLightButton" type="Button" parent="TopPanel/PanelContainer/MarginContainer/HBoxContainer"]
```
hhyyrylainen, line 450: If the light bar is being added to this editor as well, I think it would make sense to split this out as well as its own scene to reduce this button with theme overrides scene content duplication.

src/multicellular_stage/CellBodyPlanInternalCalculations.cs
```diff
@@ -0,0 +1,99 @@
...
+public static class CellBodyPlanInternalCalculations
+{
+    public static Dictionary<Compound, float> GetTotalSpecificCapacity(IEnumerable<CellTemplate> cells,
+        out float nominalCapacity)
+    {
+        nominalCapacity = 0.0f;
+
+        var capacities = new Dictionary<Compound, float>();
+
+        foreach (var cell in cells)
```
hhyyrylainen, line 14: I'd add a TODO comment here about checking if this can be sped up by going by cell type and then multiplying that by how many cells of that type there are in the bodyplan.

src/multicellular_stage/CellBodyPlanInternalCalculations.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        var speed = MicrobeInternalCalculations.CalculateSpeed(leader.Organelles, leader.MembraneType,
+            leader.MembraneRigidity, leader.IsBacteria);
+
+        speed *= cells.Count * Constants.CELL_COLONY_MOVEMENT_FORCE_MULTIPLIER;
+        var seriesValue = 1 - 1 / (float)Math.Pow(2, cells.Count - 1);
+        speed -= speed * 0.15f * seriesValue;
+
+        if (cells.Count == 1)
+            return speed;
```
hhyyrylainen, line 37: Shouldn't this return be before the speed is adjusted? Because if a single cell I don't think multicellular stage creates the colony, so it should have the exact same speed as in the microbe stage.

src/multicellular_stage/CellBodyPlanInternalCalculations.cs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public static float CalculateSpeed(IReadOnlyList<HexWithData<CellTemplate>> cells)
+    {
+        var leader = cells[0].Data!;
+
+        var speed = MicrobeInternalCalculations.CalculateSpeed(leader.Organelles, leader.MembraneType,
+            leader.MembraneRigidity, leader.IsBacteria);
+
+        speed *= cells.Count * Constants.CELL_COLONY_MOVEMENT_FORCE_MULTIPLIER;
```
hhyyrylainen, line 32: The math in this method seems very complicated and similar to what the MicrobeColony helpers has. Was that removed or could there be some common refactoring to share as much of the complex core math as possible? If not possible them both methods need to be plastered with big warning comments to not change one if the other is not changed.

src/multicellular_stage/CellBodyPlanInternalCalculations.cs
```diff
@@ -0,0 +1,99 @@
...
+    public static float CalculateRotationSpeed(IReadOnlyList<HexWithData<CellTemplate>> cells)
+    {
+        var leader = cells[0].Data!;
+
+        var colonyRotation = MicrobeInternalCalculations
+            .CalculateRotationSpeed(leader.Organelles);
+
+        Vector3 leaderPosition = Hex.AxialToCartesian(leader.Position);
+
+        foreach (var colonyMember in cells)
```
hhyyrylainen, line 86: Or maybe it is this math that should already exist in the cell colony helpers...

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
```diff
@@ -0,0 +1,118 @@
...
+    {
+        // Empty list to later fill
+        var processStatistics = new List<ProcessSpeedInformation>();
+
+        var processes = new List<TweakedProcess>();
+
+        int cellCount = cells.Count;
+        for (int i = 0; i < cellCount; ++i)
+        {
+            ProcessSystem.ComputeActiveProcessList(cells[i].Data!.Organelles, ref processes);
```
hhyyrylainen, line 44: Isn't this code wrong? The active compute list will have items removed on each subsequent cell types that don't happen to run a process. Instead I think this code ought to be refactored to count how many of each cell type are in use and then run on each cell type and multiply it by the count to get to the final result.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
```diff
@@ -0,0 +1,118 @@
...
+        organismStatisticsPanel.UpdateCompoundBalances(balances, warningTime);
+    }
+
+    private void UpdateCompoundLastingTimes(Dictionary<Compound, CompoundBalance> normalBalance,
+        Dictionary<Compound, CompoundBalance> nightBalance, float nominalStorage,
+        Dictionary<Compound, float> specificStorages)
+    {
+        float lightFraction = Editor.CurrentGame.GameWorld.WorldSettings.DaytimeFraction;
+
+        var warningTime = Editor.CurrentGame.GameWorld.LightCycle.DayLengthRealtimeSeconds * (1 - lightFraction);
```
hhyyrylainen, line 88: If these calculations are the same for each editor type? Wouldn't it make sense to put this into the statistics panel class so that there's no future problem where someone just remembers to update one place with the same math?

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -419,10 +452,12 @@ public override bool CanFinishEditing(IEnumerable<EditorUserOverride> userOverri
         if (!base.CanFinishEditing(editorUserOverrides))
             return false;
 
-        if (editorUserOverrides.Contains(EditorUserOverride.NotProducingEnoughATP))
-            return true;
-
-        // TODO: warning about not producing enough ATP if entire body plan would be negative
+        if (IsNegativeAtpProduction() &&
+            !editorUserOverrides.Contains(EditorUserOverride.NotProducingEnoughATP))
```
hhyyrylainen, line 456: Isn't it wrong that a specific editorUserOverrides field was added to this *component*? As the overrides should be editor-wide and is already passed as a parameter to this method.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -477,13 +514,47 @@ public bool ShowCellOptions()
...
+            Editor.CurrentGame.GameWorld.WorldSettings.DayNightCycleEnabled && Editor.CurrentPatch.HasDayAndNight);
+    }
+
+    public override void OnLightLevelChanged(float dayLightFraction)
+    {
+        var maxLightLevel = Editor.CurrentPatch.Biome.GetCompound(Compound.Sunlight, CompoundAmountType.Biome).Ambient;
+        var templateMaxLightLevel =
+            Editor.CurrentPatch.GetCompoundAmountForDisplay(Compound.Sunlight, CompoundAmountType.Template);
+
+        // Currently, patches whose templates have zero sunlight can be given non-zero sunlight as an instance. But
```
hhyyrylainen, line 538: Seems like a pretty long explanation to copy-paste so I think a some kind of refactoring is needed to have this long comment and bug "workaround" in a single place in the code base.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -979,11 +1081,120 @@ private void ClearSelectedAction()
...
+        if (balanceMode != ResourceLimitingMode.AllResources)
+        {
+            conditionsData = new BiomeResourceLimiterAdapter(balanceMode, conditionsData);
+        }
+
+        var energyBalance = new EnergyBalanceInfo();
+        energyBalance.SetupTrackingForRequiredCompounds();
+        foreach (var hex in cells)
+        {
+            ProcessSystem.ComputeEnergyBalance(hex.Data!.Organelles, conditionsData, hex.Data.MembraneType, moving,
```
hhyyrylainen, line 1138: I guess it would get pretty complicated to change but I'd at least add a TODO about the potential to also improve this code's performance (if needed, things are probably pretty fast with 20 or fewer cells in the editor) by calculating things per type.

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2025,7 +1852,7 @@ private void CalculateEnergyAndCompoundBalance(IReadOnlyList<OrganelleTemplate>
     {
         biome ??= Editor.CurrentPatch.Biome;
 
-        bool moving = calculateBalancesWhenMoving.ButtonPressed;
+        bool moving = organismStatisticsPanel.CalculateBalancesWhenMoving();
```
hhyyrylainen, line 1855: Shouldn't this be a property in the statistics? Pretty weird that a method with an imperative name would just return the status of something...

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2034,33 +1861,36 @@ private void CalculateEnergyAndCompoundBalance(IReadOnlyList<OrganelleTemplate>
...
-                organelles, conditionsData, energyBalance,
+            CalculateCompoundBalanceWithMethod(organismStatisticsPanel.GetBalanceDisplayType(),
+                organismStatisticsPanel.GetCompoundAmountType(), organelles, conditionsData, energyBalance,
                 ref specificStorages, ref nominalStorage);
 
         UpdateCompoundBalances(compoundBalanceData);
 
         // TODO: should this skip on being affected by the resource limited?
-        var nightBalanceData = CalculateCompoundBalanceWithMethod(compoundBalance.CurrentDisplayType,
+        var nightBalanceData = CalculateCompoundBalanceWithMethod(organismStatisticsPanel.GetBalanceDisplayType(),
```
hhyyrylainen, line 1887: `GetBalanceDisplayType` shouldn't that also be a property?

src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -2017,6 +1832,18 @@ private void StartAutoEvoPrediction()
...
     }
 
+    private void OnEnergyBalanceOptionsChanged()
+    {
+        CalculateEnergyAndCompoundBalance(editedMicrobeOrganelles, Membrane);
+    }
+
+    private void OnResourceLimitingModeChanged(int index)
+    {
+        balanceMode = (ResourceLimitingMode)index;
```
hhyyrylainen, line 1842: Wouldn't it be better to just have a property in the statistics panel that returns the current mode (similarly to `GetBalanceDisplayType`)? That way things don't break in all the editors if the order of GUI items is changed, which seems like a very good protection to have.

src/microbe_stage/OrganismStatisticsPanel.cs
```diff
@@ -0,0 +1,629 @@
...
+    public LabelSettings ATPBalanceNormalText = null!;
+
+    [Export]
+    public LabelSettings ATPBalanceNotEnoughText = null!;
+
+#pragma warning restore CA2213
+
+    private readonly StringBuilder atpToolTipTextBuilder = new();
+
+    private LightLevelOption selectedLightLevelOption = LightLevelOption.Current;
```
hhyyrylainen, line 46: In my style I prefer to put normal variables after the godot ones, so I'd move this variable down a bunch.

src/microbe_stage/OrganismStatisticsPanel.cs
```diff
@@ -0,0 +1,629 @@
...
+        float maxValue = Math.Max(energyBalance.TotalConsumption, energyBalance.TotalProduction);
+        atpProductionBar.MaxValue = maxValue;
+        atpConsumptionBar.MaxValue = maxValue;
+
+        atpProductionBar.UpdateAndMoveBars(SortBarData(energyBalance.Production));
+        atpConsumptionBar.UpdateAndMoveBars(SortBarData(energyBalance.Consumption));
+
+        if (Visible)
+        {
+            TutorialState?.SendEvent(TutorialEventType.MicrobeEditorPlayerEnergyBalanceChanged,
```
hhyyrylainen, line 220: Does this tutorial still work? Just double checking because when reading the code I didn't see the place where the tutorial state is forwarded to this component...

src/microbe_stage/OrganismStatisticsPanel.cs
```diff
@@ -0,0 +1,629 @@
...
+        {
+            tooltip.Description = description.ToString();
+        }
+        else
+        {
+            GD.PrintErr("Can't update digestion efficiency tooltip");
+        }
+    }
+
+    public BalanceDisplayType GetBalanceDisplayType()
```
hhyyrylainen, line 470: Do these change automatically when the game language is changed? I think with this being a separate component it would be better design to have properties for all the stats, and then this component would handle reacting to language change and updating all the GUI nodes.

src/microbe_stage/systems/ProcessSystem.cs
dligr: Right, I initially changed this for the speed calculation code

src/multicellular_stage/CellBodyPlanInternalCalculations.cs
dligr: It's the rotation speed calculation that is present in the MicrobeColony, though I still improved this code by separating the initial speed modification into its own method

src/microbe_stage/OrganismStatisticsPanel.cs
dligr: The tutorial didn't work apparently, but it should now, as I have added `TutorialState` forwarding to the `CellEditorComponent`'s setter

hhyyrylainen: Nice, that's about what I expected as I didn't remember seeing the code that would have been needed to make it work (and didn't just forget seeing it if it was actually there).

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
dligr: The `editorUserOverrides` is in fact a local variable created in this method: 
```C#
var editorUserOverrides = userOverrides.ToList();
```
Though it seems redundant in both the body plan and the cell editor components, so I suppose it makes sense to remove it.

hhyyrylainen: CI will probably complain about multiple enumeration, which is why that line exists.

I did see some CI failure about shadowing a field so I think at some point there was some duplicate field for the info (but I didn't look very closely)?

src/microbe_stage/OrganismStatisticsPanel.cs
dligr: The GUI is updated automatically when the language changes; the ATP balance panel's localization should update too now.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
dligr: The problem here is that those calculations rely on `Editor`'s data, which is unaccessible to the organism stats panel. So, either it has to be given the access or the editor components would have to pass this data as parameters.

hhyyrylainen: I guess with this kind of architecture problem I'd just slap some TODO comments on here about trying to fix the code duplication in the future.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -193,15 +205,16 @@ public static Dictionary<string, OrganelleEfficiency> ComputeOrganelleProcessEff
     ///   Computes the energy balance for the given organelles in biome and at a given time during the day (or type
     ///   can be specified to be a different type of value)
     /// </summary>
-    public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemplate> organelles,
+    public static void ComputeEnergyBalance(IReadOnlyList<OrganelleTemplate> organelles,
         IBiomeConditions biome, MembraneType membrane, bool includeMovementCost, bool isPlayerSpecies,
-        WorldGenerationSettings worldSettings, CompoundAmountType amountType, bool calculateRequiredResources)
+        WorldGenerationSettings worldSettings, CompoundAmountType amountType,
+        [NotNull] ref EnergyBalanceInfo? result)
```
hhyyrylainen, line 211: If this is constrained to be not null, what is the point with this being marked nullable and a ref? I really think the simplest thing is to always just require the called to already have created the instance...

src/microbe_stage/editor/LightConfigurationPanel.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        Day,
+        Night,
+        Average,
+        Current,
+    }
+
+    public void OnLightButtonClicked(string type)
+    {
+        EmitSignal(SignalName.OnLightButtonClick, type);
```
hhyyrylainen, line 38: I think it would be better if this component protected other code from potential string misconfigurations, so I'd like to see the enum LightLevelOption be put in a separate file and this would then translate the GUI data to the C# enum. It might be the case that the event handler needs to use `int` as the type to properly pass the data (as I'm not sure if godot engine detects and supports enums automatically).

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
dligr: The CI isn't complaining (and the editor overrides work in game), so the change is probably fine

src/microbe_stage/editor/LightConfigurationPanel.cs
```diff
@@ -0,0 +1,65 @@
...
+    [Export]
+    private Button nightButton = null!;
+
+    [Export]
+    private Button averageLightButton = null!;
+
+    [Export]
+    private Button currentLightButton = null!;
+
+#pragma warning disable CA2213
```
hhyyrylainen, line 23:
```suggestion
#pragma warning restore CA2213
```

I think that's a typo.

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -54,17 +54,20 @@ public static float GetTotalNominalCapacity(IEnumerable<OrganelleTemplate> organ
...
-            totalNominalCap += GetNominalCapacityForOrganelle(organelle.Definition, organelle.Upgrades);
-        }
+        AddSpecificCapacity(organelles, capacities, totalNominalCap);
 
-        var capacities = new Dictionary<Compound, float>();
+        return capacities;
+    }
+
+    public static void AddSpecificCapacity(IReadOnlyList<OrganelleTemplate> organelles,
+        Dictionary<Compound, float> capacities, float totalNominalCap)
```
hhyyrylainen, line 68: Doesn't this method *not* use totalNominalCap? So this parameter could be removed.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -223,27 +236,19 @@ public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemp
     /// <param name="isPlayerSpecies">Whether this microbe is a member of the player's species</param>
     /// <param name="worldSettings">The world generation settings for this game</param>
     /// <param name="amountType">Specifies how changes during an in-game day are taken into account</param>
-    /// <param name="calculateRequiredResources">
-    ///   If true, then the required input compounds to run at the given energy balance are stored per energy producer
-    /// </param>
     /// <param name="cache">Auto-Evo Cache for speeding up the function</param>
-    public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemplate> organelles,
+    /// <param name="result">
+    ///   The resulting energy balance. If not null, the results are added to the passed value.
```
hhyyrylainen, line 241:
```suggestion
    ///   The resulting energy balance.
```
Cannot be null anymore...

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -294,33 +299,36 @@ public static EnergyBalanceInfo ComputeEnergyBalance(IReadOnlyList<OrganelleTemp
...
 
         // Add osmoregulation
-        result.Osmoregulation = Constants.ATP_COST_FOR_OSMOREGULATION * hexCount *
+        var osmoregulation = Constants.ATP_COST_FOR_OSMOREGULATION * hexCount *
             membrane.OsmoregulationFactor;
 
         if (isPlayerSpecies)
         {
-            result.Osmoregulation *= worldSettings.OsmoregulationMultiplier;
+            osmoregulation *= worldSettings.OsmoregulationMultiplier;
```
hhyyrylainen, line 322: Doesn't this result in the results not combining correctly? So for example if we have two cells that have the balance calculated for the overall one, and it is the player species. The first cell might have 20 osmoregulation which is adjusted to 16. Then we calculate the second cell as 20+16=36 and again apply the osmoregulation we get: 28.8 **but** we should have actually ended up with 32 (2 * 16).

So I think this has to "unadjust" the osmoregulation value before this is added to `result.Osmoregulation`.

Something like:
```
if (isPlayerSpecies && result.Osmoregulation != 0)
    result.Osmoregulation *= 1 / worldSettings.OsmoregulationMultiplier
```
(I may have gotten that last line of math wrong as I didn't test it).

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
```diff
@@ -0,0 +1,128 @@
...
+using Systems;
+
+/// <summary>
+///   The partial class containing GUI updating actions
+/// </summary>
+public partial class CellBodyPlanEditorComponent
+{
+    protected override void OnTranslationsChanged()
+    {
+        organismStatisticsPanel.OnTranslationsChanged();
```
hhyyrylainen, line 13: Doesn't the cell editor also need this method? I checked in my IDE and it sees only one call to `organismStatisticsPanel.OnTranslationsChanged()` which is from here so that would mean the microbe editor is missing this part...

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private void ConfirmFinishEditingWithNegativeATPPressed()
+    {
+        if (OnFinish == null)
+        {
+            GD.PrintErr("Confirmed editing for cell editor when finish callback is not set");
```
hhyyrylainen, line 20: Has this error been updated for this new editor it is in?

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
```diff
@@ -0,0 +1,128 @@
...
+        IBiomeConditions biome)
+    {
+        // Empty list to later fill
+        var processStatistics = new List<ProcessSpeedInformation>();
+
+        var processes = new List<TweakedProcess>();
+
+        foreach (var cellType in GetCellTypes())
+        {
+            var newProcesses = new List<TweakedProcess>();
```
hhyyrylainen, line 40: This list should be moved outside the loop so that it can be reused in each loop iteration by just clearing it. This will ever so slightly save on memory allocations.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
```diff
@@ -0,0 +1,128 @@
...
+                newProcesses[i] = new TweakedProcess(newProcesses[i].Process, newProcesses[i].Rate * cellType.Count)
+                {
+                    SpeedMultiplier = newProcesses[i].SpeedMultiplier,
+                };
+            }
+
+            ProcessSystem.MergeProcessLists(processes, newProcesses);
+        }
+
+        float consumptionProductionRatio = energyBalance.TotalConsumption / energyBalance.TotalProduction;
```
hhyyrylainen, line 55: This is going to be a bit of a bigger problem, but I think there's a mistake here related to ATP balance. ATP balance is *per cell* not the overall creature. So any negative ATP check and this code here has to work on per-cell level and then only after that aggregate the results.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -620,13 +676,44 @@ protected override void Dispose(bool disposing)
...
+            }
+
+            default:
+                throw new Exception("Invalid light level option");
+        }
+    }
+
+    private bool IsNegativeAtpProduction()
+    {
+        return energyBalanceInfo != null &&
```
hhyyrylainen, line 715: Here, it doesn't help if there's one cell with massively positive ATP, because if there's any negative cell in the colony, that will just die when it grows in the body plan...

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -979,11 +1066,145 @@ private void ClearSelectedAction()
...
+
+        specificStorages ??= CellBodyPlanInternalCalculations.GetTotalSpecificCapacity(cells.Select(o => o.Data!),
+            out nominalStorage);
+
+        return ProcessSystem.ComputeCompoundFillTimes(compoundBalanceData, nominalStorage, specificStorages);
+    }
+
+    private IEnumerable<(CellType Type, int Count)> GetCellTypes()
+    {
+        cellTypesWorkMemory.Clear();
```
hhyyrylainen, line 1188: This seems a bit weird to me... this method always allocates an enumerator, so I think it would make a bit more sense to just rework the temporary storage to just act as the result. Something like `Dictionary<..., int>` should work, I think.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -979,11 +1066,145 @@ private void ClearSelectedAction()
...
+                true, Editor.CurrentGame.GameWorld.WorldSettings, organismStatisticsPanel.CompoundAmountType,
+                energyBalance);
+        }
+
+        energyBalanceInfo = energyBalance;
+
+        organismStatisticsPanel.UpdateEnergyBalance(energyBalance);
+
+        float nominalStorage = 0;
+        Dictionary<Compound, float>? specificStorages = null;
```
hhyyrylainen, line 1134: I had a bit of trouble understanding the ref usage here, so could you add a comment before these two variables that they are passed by ref so that they can be set by the first call to the method that uses them and then reused?

src/microbe_stage/OrganismStatisticsPanel.cs
```diff
@@ -0,0 +1,613 @@
...
+    [Export]
+    private CustomWindow processListWindow = null!;
+
+    [Export]
+    private LightConfigurationPanel lightConfigurationPanel = null!;
+
+#pragma warning restore CA2213
+
+    private LightLevelOption selectedLightLevelOption =
+        LightLevelOption.Current;
```
hhyyrylainen, line 129:
```suggestion
    private LightLevelOption selectedLightLevelOption = LightLevelOption.Current;
```
Wouldn't this fit on a single line?

src/microbe_stage/OrganismStatisticsPanel.cs
```diff
@@ -0,0 +1,613 @@
...
+        var selection = (LightLevelOption)option;
+
+        selectedLightLevelOption = selection;
+
+        ApplyLightLevelSelection();
+    }
+
+    private List<KeyValuePair<string, float>> SortBarData(Dictionary<string, float> bar)
+    {
+        var comparer = new ATPComparer();
```
hhyyrylainen, line 573: I think this should be saved as a readonly field.

hhyyrylainen (changes requested):
I think I've now reviewed all the code. So after those few bugs are solved this should be really close to done with just a few code cleanups to do. I didn't read the new OrganismStatisticsPanel too closely as I assumed it was mostly copy-pasted code (with hopefully no problems getting picked up at the same time).

(and I again got carried away with thinking about making a few minor comments before starting this review)

hhyyrylainen:
Also if it would help any (there's about 5 merge conflicts with master on this branch) I can push my local version of this branch that I rebased on master today.

src/microbe_stage/systems/ProcessSystem.cs
hhyyrylainen: Ah, you are right of course I had a thought mixup when reading the earlier line and already thinking it was using the result.osmoregulation. So as written it should work. Sorry about the confusion. I'll mark this discussion as resolved.

MatterNoise:
It has nothing to do with that but, will it be possible to run the cheat panel after that Pull?

hhyyrylainen:
I'm pretty sure I didn't see any code changes related to that so it will keep working as usual in the microbe editor (and still be missing from multicellular).

src/multicellular_stage/editor/CellBodyPlanEditorComponent.GUI.cs
dligr: I suppose that makes displaying the ATP balance bars and the process list obsolete (the latter because is doesn't make sense to show ATP-producing processes in the list, when they affect one cell each). The compound balance shouldn't be affected, since the cells have a shared compound storage (to my knowledge).

The ATP balance bars could be moved to the cell type buttons on the left panel and work on a per-cell basis, but that's probably outside the scope of this PR; for now they could just be removed from the multicellular editor.

Regarding the process list, what if it was made to have a dropdown for choosing a cell type, and then only display the processes of the chosen cell type? A similar approach can then be used for #3207.

hhyyrylainen: Yeah, that is kind of a core issue. I think for now showing ATP balance like in microbe will be better than not showing it at all, but it is obviously not fully suited for multicellular gameplay. We can go with the current approach for this PR, but in the future I think some other display might make more sense to show in the multicellular editor as surviving in multicellular is different than in micorbe.

> The compound balance shouldn't be affected, since the cells have a shared compound storage (to my knowledge).

Yes, this is correct ATP is a special case.

> Regarding the process list, what if it was made to have a dropdown for choosing a cell type, and then only display the processes of the chosen cell type? 

That would actually make a lot of sense.

> A similar approach can then be used for https://github.com/Revolutionary-Games/Thrive/issues/3207.

I feel like a slightly different approach needs to be used. I think a hybrid approach there where all except ATP producing processed are summed across the colony. But good thinking (I would have totally not remembered that issue exists).

hhyyrylainen:
For the more complex situation of the ATP balance bar not being fully suitable for multicellular, let's just put a few TODO comments in the code. That and adding some defensive programming around the cell counts are the last things I need doing. I can help with those tomorrow if at least the merge conflicts are solved on this branch.

dligr:
The submodule issue should be fixed now.

Regarding the ATP balance problem, wouldn't it be better to open an issue to track it, so that it can't be buried in the code?

hhyyrylainen:
I suppose that would be good so that the TODO comments can then reference also the issue URL.

hhyyrylainen (approved):
I think the code should be good now. I'll do a quick playtest tomorrow before merging this.

hhyyrylainen (changes requested):
So in playtesting I got this bug:




I can't merge this with that bug. It isn't immediately obvious what triggers the problem to me.

To trigger that bug all I did was start a new game with tutorials on, go to the editor and follow the tutorials by placing a hydrogenase, exiting, and then when getting back that wrong tutorial immediately pops up. I just confirmed that on master branch the same steps don't lead to that specific tutorial triggering too early.

src/microbe_stage/OrganismStatisticsPanel.cs
```diff
@@ -0,0 +1,612 @@
...
+        float maxValue = Math.Max(energyBalance.TotalConsumption, energyBalance.TotalProduction);
+        atpProductionBar.MaxValue = maxValue;
+        atpConsumptionBar.MaxValue = maxValue;
+
+        atpProductionBar.UpdateAndMoveBars(SortBarData(energyBalance.Production));
+        atpConsumptionBar.UpdateAndMoveBars(SortBarData(energyBalance.Consumption));
+
+        if (Visible)
+        {
+            TutorialState?.SendEvent(TutorialEventType.MicrobeEditorPlayerEnergyBalanceChanged,
```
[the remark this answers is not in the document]
hhyyrylainen: I suppose that makes sense. Or this could be updated to use `IsVisibleInTree` which should fix the Visible property being different in this child node rather than the direct component node.

hhyyrylainen (approved):
I think everything is fine now with that bugfix.

PR #5827, opened 2025-01-17
Make Auto-Evo able to evolve mucocyst

### LINKED ISSUE #5291: Auto-evo should take mucocyst into account
so that it has a chance to evolve, additionally that depends on:

https://github.com/Revolutionary-Games/Thrive/issues/3038

Patryk26g:

Makes Auto-Evo be able to evolve mucocyst in preys and also adds more info about organelles' upgrades in AutoEvoExploringTool 

**Related Issues**

closes #5291

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/auto-evo/selection_pressure/AvoidPredationSelectionPressure.cs
```diff
@@ -25,6 +25,9 @@ public AvoidPredationSelectionPressure(Species predator, float weight) : base(we
             CommonMutationFunctions.Direction.Rear),
         new MoveOrganelleBack(organelle => organelle.HasSlimeJetComponent),
         new MoveOrganelleBack(organelle => organelle.HasMovementComponent),
+        new UpgradeOrganelle(organelle => organelle.HasSlimeJetComponent, "mucocyst"),
```
hhyyrylainen, line 28: Shouldn't this use the magic constant from the slime jet component? `MUCOCYST_UPGRADE_NAME` (you'll need to change it t public).

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -410,12 +412,21 @@ public void Clear()
 
             if (organelle.Definition.HasSlimeJetComponent)
             {
-                slimeJetScore += Constants.AUTO_EVO_MUCILAGE_PREDATION_SCORE;
+                // Since slime jets have only one upgrade available we don't need to check what these Upgrades have
```
hhyyrylainen, line 415: This doesn't seem futureproof so please update this code to actually check the real upgrades and confirm it is actually the mucocyst.

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -1186,8 +1213,14 @@ public AutoEvoExploringToolWorld(IAutoEvoConfiguration configuration)
             foreach (var organelle in SimulationParameters.Instance.GetAllOrganelles())
             {
                 MicrobeSpeciesOrganelleStatistics.Add(organelle, (0, 0));
+                foreach (var upgrade in organelle.AvailableUpgrades.Keys)
+                {
+                    organelle.AvailableUpgrades.TryGetValue(upgrade, out var upgradeName);
+                    MicrobeSpeciesUpgradesStatistics.TryAdd(upgrade, (upgradeName?.Name ?? string.Empty, 0, 0));
```
hhyyrylainen, line 1219: Shouldn't this have error handling and printing an error rather than adding an empty upgrade name? Doesn't it imply major json config error if the TryGetValue fails here?

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -7,12 +7,19 @@
...
     {
         allOrganelles = SimulationParameters.Instance.GetAllOrganelles().Where(criteria).ToFrozenSet();
-        this.upgrade = upgrade;
+        this.customUpgrade = customUpgrade;
+    }
+
+    public UpgradeOrganelle(Func<OrganelleDefinition, bool> criteria, string upgradeName)
+    {
+        allOrganelles = SimulationParameters.Instance.GetAllOrganelles().Where(criteria).ToFrozenSet();
+        this.upgradeName = upgradeName;
```
hhyyrylainen, line 22: Shouldn't this verify the upgrade name actually exists in the organelle? And throw an exception as that's a really big game config error.

src/auto-evo/simulation/SimulationCache.cs
CountingBeeps: Also honestly, the prey having mucocyst should subtract from the predators score, not the predator getting a bonus for having one.

hhyyrylainen: Though wouldn't that mean there's no auto-evo benefit for the species itself, meaning that it would never evolve? So there also needs to be some bonus calculation from having the mucocyst (I guess there's the "avoid predation" thing in auto-evo where this could be added).

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -431,6 +442,9 @@ public void Clear()
         slimeJetScore *= MathF.Sqrt(slimeJetsCount);
         slimeJetScore *= mucilageMultiplier;
 
+        mucocystsScore *= MathF.Sqrt(mucocystsCount);
+        slimeJetScore += mucocystsScore;
+
         var predationToolsRawScores = (pilusScore, oxytoxyScore, slimeJetScore);
```
CountingBeeps, line 448: Mucocysts really should be their own "predation tool" instead of adding to slime jet, as they do very different things.

src/auto-evo/simulation/SimulationCache.cs
CountingBeeps: The avoid predation pressure works by taking 1 / the score the predator has to eat the prey, so subtracting from it actually would give a benefit.

Patryk26g: Talking about that score = (f(x) = 1/x for x > 0) and (f(x) = 1 for x < 0) - I don't like that function, it is not continous, get's messy for x in range [0,1] and doesn't increase the prey's score if they get better from some point so I'll change it a little bit probably for a different function
image
image

src/microbe_stage/organelle_components/SlimeJetComponent.cs
```diff
@@ -7,8 +7,8 @@
 /// </summary>
 public class SlimeJetComponent : IOrganelleComponent
 {
+    public const string MUCOCYST = "mucocyst";
```
hhyyrylainen, line 10: As this is now public I think it is even *more* important for this to have a clear name so could you please rename this bac kto MUCOCYST_UPGRADE_NAME?

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -47,7 +62,15 @@ public UpgradeOrganelle(Func<OrganelleDefinition, bool> criteria, IComponentSpec
...
                     organelle.Upgrades ??= new OrganelleUpgrades();
-                    organelle.Upgrades.CustomUpgradeData = upgrade;
+                    if (customUpgrade != null)
+                    {
+                        organelle.Upgrades.CustomUpgradeData = customUpgrade;
+                    }
+
+                    if (upgradeName != null)
+                    {
+                        organelle.Upgrades.UnlockedFeatures.Add(upgradeName);
```
hhyyrylainen, line 72: This should check if the unlocked features already contains the upgrade and not add it if it is there already. This is because the UnlockedFeatures is a list (and not a set that would automatically prevent this).

src/auto-evo/AutoEvoExploringTool.cs
```diff
@@ -1153,6 +1174,12 @@ private class AutoEvoExploringToolWorld
         public readonly Dictionary<OrganelleDefinition, (double Percentage, double Average)>
             MicrobeSpeciesOrganelleStatistics = new();
 
+        /// <summary>
+        ///   Used to generate world statistics
```
hhyyrylainen, line 1178: Shouldn't this clarify that this is for calculating world *upgrade* statistics? To me this comment like it is right now doesn't really seem to match the variable this is on...

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -410,12 +412,18 @@ public void Clear()
 
             if (organelle.Definition.HasSlimeJetComponent)
             {
-                slimeJetScore += Constants.AUTO_EVO_MUCILAGE_PREDATION_SCORE;
+                if (organelle.Upgrades?.UnlockedFeatures.Contains(SlimeJetComponent.MUCOCYST) ?? false)
```
hhyyrylainen, line 415:
```suggestion
                if (organelle.Upgrades?.UnlockedFeatures.Contains(SlimeJetComponent.MUCOCYST) == true)
```

I prefer this way of writing the nullable bool check.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
hhyyrylainen: Arguably it could be a set. I didn't make it initially a set for a few reasons:
- It's simpler to just have a list
- There could be save complications with a set as I'm pretty sure we don't have any sets that are deserialized yet
- Most of unlock feature lists are at most 2-3 items so always looping the entire thing is probably not any slower than a set
- There could be some performance overhead of a set, though I suppose be basically would only need to call `.Contains` most of the time

Checking all of those factors will be a bit of work which is why I haven't bothered changing it whenever I've again worked on the unlocked features variable.

src/auto-evo/selection_pressure/AvoidPredationSelectionPressure.cs
```diff
@@ -25,6 +25,9 @@ public AvoidPredationSelectionPressure(Species predator, float weight) : base(we
             CommonMutationFunctions.Direction.Rear),
         new MoveOrganelleBack(organelle => organelle.HasSlimeJetComponent),
         new MoveOrganelleBack(organelle => organelle.HasMovementComponent),
+        new UpgradeOrganelle(organelle => organelle.HasSlimeJetComponent, SlimeJetComponent.MUCOCYST),
+        new MoveOrganelleBack(organelle => organelle.HasSlimeJetComponent),
+        new MoveOrganelleBack(organelle => organelle.HasMovementComponent),
```
CountingBeeps, line 30: This creates a second instance of these when there only needs one each

```
new MoveOrganelleBack(organelle => organelle.HasSlimeJetComponent),
new MoveOrganelleBack(organelle => organelle.HasMovementComponent),
```

hhyyrylainen (approved):
Code changes also look good now. I tested in auto-evo exploring tool that nothing seemed wrong and some species got the mucocyst.

PR #5835, opened 2025-01-21
Increased auto-evo strength on the player (and made it a difficulty option)

### LINKED ISSUE #2907: Increase player auto-evo strength to 80% once the algorithm feels good
currently auto-evo is turned *way* down for the player because the algorithm isn't very good in terms of player experience. Once that is fixed the strength should be turned back up.

hhyyrylainen:

Makes auto-evo strength on the player configurable instead of a constant

**Related Issues**

closes #2907

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/IDifficulty.cs
```diff
@@ -63,13 +70,12 @@ public interface IDifficulty : IRegistryAssignable
     public bool LimitGrowthRate { get; }
 
     /// <summary>
-    ///   How intense should the fog-of-war be
+    ///   Sets how intense should the fog-of-war be
```
dligr, line 73: It's a nitpick, but I believe that this comment isn't grammatically correct ('should' should go after 'the fog-of-war', per subject-verb-object order)

dligr (approved):
I can confirm that this PR works as intended. I haven't noticed any problems with the code, apart from one grammar mistake.

src/general/IDifficulty.cs
hhyyrylainen: According to https://plugins.jetbrains.com/plugin/16136-grazie-pro the old way was flagged as incorrect, but this variant of the text is fine (this plugin has been absolutely brutal on the game comments with it flagging most of them as needing grammar improvements). And I think it is possible to make this kind of construction grammatically where the should and be are separated like this.

dligr: This kind of construction is possible, but only if inversion is applied (e.g. in questions). I don't think it's applicable in this comment ([here are its rules](https://dictionary.cambridge.org/grammar/british-grammar/inversion)).

hhyyrylainen: Alright, I've changed this comment.

hhyyrylainen:
I'll assume there weren't other problems and merge this in a few minutes.

PR #5841, opened 2025-01-23
Fix patch notes not grabbing focus

### LINKED ISSUE #4714: Viewing patch notes no longer grabs focus correctly
Maybe a regression from: https://github.com/Revolutionary-Games/Thrive/pull/4523 as that's the only thing I can think of that has recently changed in regards to the view patch notes button in the options menu. Might be a bit tricky if fixes are needed for the focus grab system to be new-style popup aware.

CI09:

I don't know if it's the right way to do it, I made patch notes focus mode "Click" and made popups do GrabFocus().
It doesn't immediately grab focus when opened but after navigation I think any way it does.

**Related Issues**

Closes #4714

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09:
Actually I tried without GrabFocus() and it also works, let me know if it needs to be deleted.

hhyyrylainen:
There's the FocusGrabber system I put in the game, that should be used instead of trying to go around it. So the popup needs a focus grabber with a higher priority than the menu one, and set the property to grab on becoming visible.

CI09:
Seems you solved the issue without opening the engine.

hhyyrylainen:
I don't have time to playtest this today, but is there a reason why it seems like you didn't enable the option to force grab focus when becoming visible? As I think that would make the popup potentially slightly more reliable.

CI09:
Appearently it worked without so, but I can enable that, just give me a sign

hhyyrylainen:
I would enable it for safety reasons as long as that doesn't cause any problems elsewhere.

hhyyrylainen (approved):
I think this is all fine and working correctly.

PR #5843, opened 2025-01-23
Improved speed of organelle duplication position finding and natural auto-evo positions

### LINKED ISSUE #3273: Fix lag spikes from organelle duplication in very large cells

hhyyrylainen:

As when working on #3273 I found quite a few inefficiencies there (like trying multiple rotations that don't affect the placement possibility). My original idea of adding a cache to resume checking positions didn't end up being really useful, but these changes should help. Auto-evo seems to be about 25% or so faster to get to generation 15 and playing with a pretty much the hugest plausible cell (without cheats) there's no lag spikes when an organelle duplicates anymore.

**Related Issues**

closes #3273 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/auto-evo/mutation_strategy/AddOrganelleAnywhere.cs
```diff
@@ -86,15 +86,17 @@ public static AddOrganelleAnywhere ThatConvertBetweenCompounds(Compound fromComp
 
         var mutated = new List<Tuple<MicrobeSpecies, float>>();
 
+        // TODO: reuse this memory somehow
         var workMemory1 = new List<Hex>();
         var workMemory2 = new List<Hex>();
+        var workMemory3 = new HashSet<Hex>();
```
CountingBeeps, line 92: Would making them attributes of the class and just clearing them at the start work? It would keep the rules of no state due to them being reset each time the function is called.

PR #5847, opened 2025-01-25
Evolve lysosomes in auto-evo

### LINKED ISSUE #3358: Make auto-evo take lysosomes (new digestion) into account
lysosome effect on predation needs to be taken into account in the auto-evo, also the fact that prokaryotes get less food from predation needs to be taken into account

Description from #3512:
> Such as number of lysosomes present in a species affecting digestion speed and efficiency and what type of enzymes are contained by the species for deciding what can or can't be eaten. All these should be taken into account by the auto-evo.

Patryk26g:

- Evolve lysosomes in auto-evo
- Add membrane and enzymes statitics to auto-evo tool
- Adjust pilus score calculation in auto-evo for more costincency
- Small auto-evo score refactor

I also tried to make lysosomes evolve their enzymes but didn't get any result probably because there almost weren't any preys with chitine or cellulose membranes

**Related Issues**

Closes #3358 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -320,7 +321,11 @@ public float GetPredationScore(Species species, Species preySpecies, IBiomeCondi
             scoreMultiplier *= Constants.AUTO_EVO_CHUNK_LEAK_MULTIPLIER;
         }
 
-        cached = (scoreMultiplier * behaviourScore * (pilusScore + engulfScore + oxytoxyScore + predatorSlimeJetScore) -
+        var enzymesScore = predatorEnzymes
+            .Sum(enzyme => Constants.AutoEvoLysosomeEnzymesScores.TryGetValue(enzyme, out var value) ? value : 0.0f);
+
+        cached = (scoreMultiplier * behaviourScore *
+                (pilusScore + engulfScore + oxytoxyScore + predatorSlimeJetScore + enzymesScore) -
                 (preySlimeJetScore + preyMucocystsScore)) /
```
CountingBeeps, line 329: realistically this should probably be a multiplier on engulf score instead of just a flat score added on

CountingBeeps:
I would argue that this probably should effect the energy gained by predation, but the energy system really isn't setup that way, so that is probably better left to a rework of that system. Might be good to open an issue on it.

locale/en.po
```diff
@@ -3308,6 +3311,12 @@ msgstr "Place flagella or slime jets symmetrically at the back of your cell. Oth
 msgid "MICROBE_LOADING_TIP_9"
 msgstr "It matters where you place parts. The shape of your cell impacts your movement and turning speed."
 
+msgid "MICROBE_MEMBRANE_PERCENTAGE_STATISTICS"
+msgstr "  {0}: Found in {1} species"
```
hhyyrylainen, line 3315: It's very undesirable to have padding whitespace directly in the translated text...

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -433,15 +443,23 @@ public void Clear()
                     oxytoxyScore += oxytoxyAmount * Constants.AUTO_EVO_TOXIN_PREDATION_SCORE;
                 }
             }
+
+            if (organelle.Definition.HasLysosomeComponent)
+            {
+                enzymes.UnionWith(organelle.Definition.Enzymes
```
hhyyrylainen, line 449: This LINQ in core logic needs to be avoided...

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -43,7 +44,8 @@ public class SimulationCache
     private readonly Dictionary<(TweakedProcess, IBiomeConditions), ProcessSpeedInformation> cachedProcessSpeeds =
         new();
 
-    private readonly Dictionary<MicrobeSpecies, (float, float, float, float)> cachedPredationToolsRawScores = new();
+    private readonly Dictionary<MicrobeSpecies, (float, float, float, float, HashSet<string>)>
```
hhyyrylainen, line 47: I think that adding a HashSet here is starting to make this way too bloated to keep using this current architecture...

I think the enzymes need to be saved differently. Maybe splitting the data into multiple parts? Like one general predation tools score and then relying on caching the enzyme scores per predated-on species (in predationScores).

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -435,18 +442,48 @@ public void Clear()
...
 
+        slimeJetScore *= slimeJetsMultiplier;
+
         var predationToolsRawScores = (pilusScore, oxytoxyScore, slimeJetScore, mucocystsScore);
 
         cachedPredationToolsRawScores.Add(microbeSpecies, predationToolsRawScores);
         return predationToolsRawScores;
     }
 
+    public (float EnzymesScore, bool IsMembranedigestable) GetEnzymesScores(MicrobeSpecies predator,
```
hhyyrylainen, line 458: Is there a reason why this doesn't just return 0 when indigestible?

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -435,18 +442,48 @@ public void Clear()
...
+
+        var organelles = predator.Organelles.Organelles;
+        var isMembraneDigestable = dissolverEnzyme == Constants.LIPASE_ENZYME;
+        var enzymesScore = 1.0f;
+
+        foreach (var organelle in organelles)
+        {
+            if (organelle.Definition.HasLysosomeComponent)
+            {
+                foreach (var enzyme in organelle.Definition.Enzymes.Where(enzyme => enzyme.Value > 0))
```
hhyyrylainen, line 473: This LINQ usage should be avoided as this allocates a bunch of temporary memory. I'm doing some style changes already for this PR, so I can change this as well. I'll just comment for letting you know to watch out for this in the future.

src/auto-evo/simulation/SimulationCache.cs
Patryk26g: The engulfment score is muliplied by enzymes score so its default value is 1, if that's what you mean

hhyyrylainen: No, I meant that as there's only use of the score inside a check with the bool, you might as well just return 0 if the bool would be false. In fact I've done that change and I'll push what I have in a minute or so for you to look at.

Patryk26g: but maybe you're right and I should return only 0 if IsMembranedigestable is false , otherwise return the enzyme score

src/auto-evo/simulation/SimulationCache.cs
hhyyrylainen: Also I think I found a major problem here, using `Name` instead of `InternalName` means that the scores were *never* read from the score dictionary. I'll push what I have but I think I'll need your help to rebalance all the numbers to work.

hhyyrylainen (approved):
This does work now, but it seems the chance is pretty low. In my testing 50% of my auto-evo exploring tool worlds created (each up to generation 15) had lipase in them (and no other enzyme). So I'll merge this as a working implementation but I think a follow up PR tweaking the balance would be great.

PR #5852, opened 2025-01-28
Visuals apply only when patch is visible

CI09:

For testers, make a world with any seed, find another hydrothermal vents patch, make another world with same seed and go near it without "discovering" it. You can also make eruption chance 100% to be sure.

**Related Issues**

https://discord.com/channels/228300288023461893/958598553389903913/1333539369696362537

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Patch.cs
+++ b/src/microbe_stage/Patch.cs
@@ -540,7 +540,8 @@ public class Patch
 
     public void ApplyPatchEventVisuals(PatchMapNode node)
     {
-        node.ShowEventVisuals(activeWorldEffectVisuals);
+        if (Visibility == MapElementVisibility.Shown)
+            node.ShowEventVisuals(activeWorldEffectVisuals);
     }
 
     public override string ToString()


src/microbe_stage/Patch.cs
```diff
@@ -540,7 +540,8 @@ public void ClearPatchNodeEventVisuals()
 
     public void ApplyPatchEventVisuals(PatchMapNode node)
     {
-        node.ShowEventVisuals(activeWorldEffectVisuals);
+        if (Visibility == MapElementVisibility.Shown)
+            node.ShowEventVisuals(activeWorldEffectVisuals);
```
hhyyrylainen, line 544: If the patch becomes visible in the current editor cycle, does the event icon appear? When seeing this PR that was my one concern that event icons might stay missing if a patch becomes later visible.

hhyyrylainen:
I'll also correct the title here as it is very unclear what "visuals" apply.

src/microbe_stage/Patch.cs
CI09: Method is called every patch map draw, so yes, it will check again for event visuals when discovering patch. Sorry for the late reply

hhyyrylainen (approved):
I didn't through a full playtest but I verified the code changes that they should work as intended.

PR #5864, opened 2025-01-30
Clear endosymbiosis when entering multicellular

### LINKED ISSUE #5844: Cancel endosymbiosis when becoming multicellular
Someone finally tested: https://github.com/Revolutionary-Games/Thrive/issues/5131 and while not super bad it'd be better to cancel the endosymbiosis action. Mostly because the colonies are not setup to change shape by adding new organelles so it'd be much harder to keep the endosymbiosis functionality working while in multicellular.

https://community.revolutionarygamesstudio.com/t/mutation-suggestion-recommending-myofibril-in-microbe-stage/8106/13?u=hhyyrylainen

CI09:

This PR does some stuff...

**Related Issues**

Closes #5844

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/MicrobeStage.cs
```diff
@@ -566,6 +566,11 @@ public void MoveToMulticellular()
             throw new Exception("failed to keep the current scene root");
         }
 
+        var endosymbiosis = GameWorld.PlayerSpecies.Endosymbiosis;
+
+        if (endosymbiosis.StartedEndosymbiosis != null)
+            endosymbiosis.CancelEndosymbiosisTarget((int)endosymbiosis.StartedEndosymbiosis.Species.ID);
```
hhyyrylainen, line 572: I think it would be better to write a new method in EndosymbiosisData called something like `CancelAllTargets` (or something similar) that would just always be called here, and that would then clear up any endoymbiosis targets. That way the code here will keep working if ever it is added support for to have multiple ongoing endosymbiosis processes at once.

hhyyrylainen (changes requested):

hhyyrylainen (approved):
I didn't test specifically in game but the code changes look perfect to me now.

PR #5867, opened 2025-01-31
Add radioactive chunk miche and make AI to use and omit radioactive chunks

### LINKED ISSUE #5855: Microbe AI should react to taking radiation damage
And try to run from the damage (probably enough to change movement direction by at least 90 degrees)

### LINKED ISSUE #5856: Microbe AI should know how to use the melanosome to gather energy
So that AI cells with it aren't totally useless. There might be some kind of detection based on the compounds in the chunk that match ATP-generating resource, but that likely still would need tweaking.

### LINKED ISSUE #5857: Auto-evo needs a new miche added for radiotrophy
So that AI species can evolve it and it would help the player's population

Kind of related issue: https://github.com/Revolutionary-Games/Thrive/issues/5856 as if AI cells are really bad at using the melanosome in-game evolving it more will cause some obvious bad behaviour

Patryk26g:

- Adds radioactive chunk miche. The miche fills only after the nucleus is developed just as expected.
- AI omits the radioactive chunks when it starts taking damage
- AI with melanosomes stays near the chunk to produce ATP

Here's the link to the test save where there are some species that have melanosomes: https://dev.revolutionarygamesstudio.com/files/Public/Testing%20Resources/Saves/Radioactive%20Chunk%20AI%20Test%20Save.thrivesave

**Related Issues**

Closes #5857 
Closes #5855 
Closes #5856

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CountingBeeps:
Didn't test, but the code all checks out Auto-Evo wise

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -497,6 +510,16 @@ private void ChooseActions(in Entity entity, ref MicrobeAI ai, ref CompoundAbsor
         // There is no reason to be engulfing at this stage
         control.SetStateColonyAware(entity, MicrobeState.Normal);
 
+        // If the microbe has radiation protection it means it has melanosomes and can stay near tha radioactive chunks
```
hhyyrylainen, line 513: Typo in the comment

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -510,6 +533,82 @@ private void ChooseActions(in Entity entity, ref MicrobeAI ai, ref CompoundAbsor
...
+        ref MicrobeControl control)
+    {
+        var chosenChunk = GetNearestRadioactiveChunk(ref position, 500.0f);
+
+        if (chosenChunk == null)
+        {
+            return false;
+        }
+
+        var oppositeDirection = position.Position + (position.Position - chosenChunk.Value.Position);
```
hhyyrylainen, line 575: Shouldn't this be normalized and then multiplied by a certain distance? I think that this math may end up picking a point really near if the distance between the chunk and the current position gets too low.

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -510,6 +533,82 @@ private void ChooseActions(in Entity entity, ref MicrobeAI ai, ref CompoundAbsor
...
+        var maxDistance = 30000.0f * speciesFocus / Constants.MAX_SPECIES_FOCUS + 3000.0f;
+        var chosenChunk = GetNearestRadioactiveChunk(ref position, maxDistance);
+
+        if (chosenChunk == null)
+        {
+            return false;
+        }
+
+        // If the microbe is close to the chunk it doesn't need to go any closer
+        if (position.Position.DistanceSquaredTo(chosenChunk.Value.Position) < 700.0f)
```
hhyyrylainen, line 598: Maybe a slight bit of randomness in the distance could be used here to make cells not always stop at an exact distance away from a chunk? (the microbe AI should already have access to a random instance that is fetched for each entity)

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -1320,34 +1419,45 @@ private void BuildChunksCache()
         // To allow multithreaded AI access safely
         lock (chunkDataCache)
         {
-            if (chunkCacheBuilt)
-                return;
-
-            foreach (ref readonly var chunk in chunksSet.GetEntities())
+            lock (terrainChunkDataCache)
```
hhyyrylainen, line 1422: I think this second level of locking is unnecessary and could be replaced with a comment like:
```
// As the chunk lock is always held when building all of the chunk caches, the other individual cache objects don't need separate locks to protect them
```

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -1320,34 +1419,45 @@ private void BuildChunksCache()
...
             {
-                if (chunk.Has<TimedLife>())
+                if (chunkCacheBuilt)
+                    return;
+
+                foreach (ref readonly var chunk in chunksSet.GetEntities())
                 {
-                    // Ignore already despawning chunks
-                    ref var timed = ref chunk.Get<TimedLife>();
+                    if (chunk.Has<TimedLife>())
```
hhyyrylainen, line 1429: Is this a new code? Is this really that useful to add here? This means that relatively the chunk cache will take quite a bit longer to build. And my hunch is that this doesn't help much as it takes just a few seconds for chunks to dissolve, so microbes don't see that many disappearing chunks even without this check.

hhyyrylainen (changes requested):
I think this is already in very good shape but I had a few small technical comments on the code (I haven't playtested).

src/microbe_stage/systems/MicrobeAISystem.cs
Patryk26g: Yes, this code was already there, but I can remove it

hhyyrylainen: If the code was there already, then fine it can stay.
The diff should get a bit clearer once you remove that one lock and deindent a lot of the lines back to what they were.

hhyyrylainen (approved):
Code changes seem fine to me. I doubt there's anyone who would do playtesting even if this was left open for longer so I'll just merge this.

hhyyrylainen:
Actually this needs merge conflicts to be resolved before this can be merged.

PR #5869, opened 2025-02-01
Make microbes stop sprinting when engulfed

### LINKED ISSUE #5569: Sprinting needs to stop if engulfed
Report from discord:
> If I'm sprinting and get engulfed, the sprint bar keeps filling, going into the red.

So there probably isn't logic to turn off the under strain variable when a cell gets engulfed / is engulfed.

dligr:

Stops microbes from sprinting after they get engulfed. 
The `IsUnderStrain` marker only gets changed in the movement system, which doesn't run when the microbe gets engulfed. As such, it has to be set to false in the `OnBecomeEngulfed` method.

**Related Issues**

Fixes #5569

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/components/Engulfable.cs
+++ b/src/microbe_stage/components/Engulfable.cs
@@ -182,6 +182,11 @@ public static class EngulfableHelpers
             entity.Get<MicrobeControl>().State = MicrobeState.Normal;
         }
 
+        if (entity.Has<StrainAffected>())
+        {
+            entity.Get<StrainAffected>().IsUnderStrain = false;
+        }
+
         // Disable compound venting
         if (entity.Has<UnneededCompoundVenter>())
         {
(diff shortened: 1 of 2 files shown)


src/microbe_stage/components/Engulfable.cs
```diff
@@ -182,6 +182,11 @@ public static void OnBecomeEngulfed(this ref Engulfable engulfable, in Entity en
             entity.Get<MicrobeControl>().State = MicrobeState.Normal;
         }
 
+        if (entity.Has<StrainAffected>())
```
hhyyrylainen, line 185: Did you check all the usages of OnBecomeEngulfed and marked all of those systems with `WritesToComponent[typeof(StrainAffected)]`? I ask because those changes aren't in this PR, but it could be that all of the relevant systems already are marked as modifying the strain affected component. If those changes are not done then also the "generate threaded run" script should be re-ran to make sure there's no data read/write conflicts with this new component access.

Other than that the change here seems very logical.

dligr: I added `[WritesToComponent(typeof(StrainAffected))]` to the engulfing system, as it's the only one that calls the modified method.
I also ran the generate threaded systems, though it did no changes.

hhyyrylainen: That's perfect then.

hhyyrylainen (approved):
The code changes look perfect to me now. I'll assume enough testing was done that this does fix the marked issue.

PR #5880, opened 2025-02-06
Add dots indicating player population on patch map

CI09:

Might need some visual stuff and other tweaks but it theoretically can be merged.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/PatchMapDrawer.cs
+++ b/src/microbe_stage/editor/PatchMapDrawer.cs
@@ -1075,7 +1075,7 @@ public partial class PatchMapDrawer : Control
         }
 
         // Hide excess cached indicators
-        for (var i = nextIndicatorIndex; i < playerSpeciesPopulationIndicators.Count - 1; i++)
+        for (var i = nextIndicatorIndex; i < playerSpeciesPopulationIndicators.Count - 1; ++i)
         {
             playerSpeciesPopulationIndicators[i].Hide();
         }


CountingBeeps:
An image would be nice, so we can give graphics feedback.

CI09:
image

hhyyrylainen:
I think the dots should be more desaturated to look like more as just background stuff. That would help with them not messing with the readability of the connection lines.

src/microbe_stage/editor/PatchMapPopulationIndicator.cs
```diff
@@ -0,0 +1,35 @@
+﻿using Godot;
+
+/// <summary>
+///   Dots scattered around patch map nodes, indication population of player species
+/// </summary>
+public partial class PatchMapPopulationIndicator : Control
```
hhyyrylainen, line 6: There's going to be a ton of these, so I think that actually this shouldn't be a node type with a script attached. And instead just I'd inline this code to the patch map drawer class (though it should be cleanly put into a separate section as it is already a pretty long file). I think the performance impact will be much less when there's just simple graphics used for the indicators rather than nodes with scripts attached.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -141,6 +142,11 @@ public Patch? SelectedPatch
     /// </summary>
     public Action<PatchMapDrawer>? OnSelectedPatchChanged { get; set; }
 
+    /// <summary>
+    ///   Player species ID for player population indicator (dots on patch map)
+    /// </summary>
+    public uint PlayerSpeciesID { get; set; }
```
hhyyrylainen, line 148: I think this could default to -1 (or uint.MaxValue if C# complains) so that this doesn't accidentally match anything if there's a bug elsewhere in the code that causes a species with ID = 0 to be created.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1109,6 +1117,24 @@ private void AddPatchNode(Patch patch, Vector2 position)
...
         patch.ApplyPatchEventVisuals(node);
 
+        var playerSpecies = patch.FindSpeciesByID(PlayerSpeciesID);
+        if (playerSpecies != null)
+        {
+            var playerPopulation = patch.GetSpeciesSimulationPopulation(playerSpecies);
+
+            for (var i = 0; i < playerPopulation * 0.001; ++i)
+            {
+                var indicator = populationIndicatorScene.Instantiate<PatchMapPopulationIndicator>();
```
hhyyrylainen, line 1127: I think it would be worthwhile probably to have a cache of these objects as I think otherwise this will increase the amount of memory allocations quite a lot here.

Overall the architecture isn't far off from being pretty much exactly what I would have done for this feature. So if this is slightly tweaked and the visuals improved a bit, this PR will be good to go.

CI09: By cache you mean a list created in this create-patch-node method, a list in the patch map drawer class, or something else completely? I'm not questioning why, since you are much more experienced but I'm curious how it reduces allocations

hhyyrylainen: A Queue or a Stack in the patch map drawer class, so it can just create them once and then reuse them.

Once the queue is filled up once with unused objects, the next time the dots need to be put in the map, they can be pulled from the queue and refilled. So this way the dots are instantiated just once, instead of being instantiated on each map draw.

hhyyrylainen:
Seems like pretty normal Godot shutdown errors to me. There's probably an open bug report to godot about those. At least I remember looking up the rendering server error and seeing there was an open bug with it. I think I'm watching it on github and I don't remember seeing an email about it being closed, so the issue is still probably open.

src/microbe_stage/editor/PatchMapDrawer.cs
hhyyrylainen: What if they weren't? What if they were added as children to the patch map itself? AFAIK the patch nodes are added as children to the drawer node in order to position them and make them display at the right position on screen.

CI09:
image

On below image is volcanic vents patch and only adjacent patch inhabited by player:
image

All 3 patched inhabited by player. Uppermost has about 6k population, rest 2-3k:
image

I don't know if the caching was done the right way though.

hhyyrylainen:
> Edit: Oh and dots are behind the panel. I wanted to give other patch map elements ZIndex (greater than zero) but it would be on top of pause menu. So I put indicators underneath all else then.

You can reorder the node that contains the dot in relation to its siblings to control what is drawn on top of what without messing with zindex or draw behind parent properties.

CI09:
Hmm weird, I changed indicator container to be at top in the tree, yet it behaves weirdly:
image

CI09:
Oh I just reminded myself that tutorial will be needed. It can be done in a seperate PR, or in this. I will be doing else for today though so I won't implement tutorial today.

hhyyrylainen:
As long as the effect is made simple enough, I think we can do without having a tutorial. Well as long as there's some thriveopedia page or something that explains the feature.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -11,6 +11,8 @@ public partial class PatchMapDrawer : Control
     [Export]
     public bool DrawDefaultMapIfEmpty;
 
+    public uint PlayerSpeciesID = int.MaxValue;
```
hhyyrylainen, line 14: Non-godot properties shouldn't be mixed in so this field should be after LineContainerPath

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -149,6 +157,8 @@ public override void _Ready()
         lineContainer = GetNode<Control>(LineContainerPath);
 
         nodeScene = GD.Load<PackedScene>("res://src/microbe_stage/editor/PatchMapNode.tscn");
+        populationIndicatorScene = GD.Load<PackedScene>(
```
hhyyrylainen, line 160: This is just a single node so for performance I don't think this should be made as a separate scene, but just instantiate the nodes with the right texture (load the texture here) in this class.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1047,6 +1057,13 @@ private void DrawPatchLinks()
     /// </summary>
     private void RebuildMap()
     {
+        playerSpeciesPopulationIndicators = new List<Control>();
```
hhyyrylainen, line 1060: The list should be stored persistently as otherwise this allocates memory unnecessarily each time the map is drawn.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1047,6 +1057,13 @@ private void DrawPatchLinks()
...
+        foreach (var node in populationIndicatorContainer.GetChildren())
```
hhyyrylainen, line 1061: GetChildren is still here, this needs to go I think GetChildren is a sign of an architecture problem. Use a persistent list for knowing all the existing indicators that way you can also avoid the type cast on line 1063.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1109,6 +1126,58 @@ private void AddPatchNode(Patch patch, Vector2 position)
 
         patch.ApplyPatchEventVisuals(node);
 
+        var playerSpecies = patch.FindSpeciesByID(PlayerSpeciesID);
+        if (playerSpecies != null)
+        {
+            var playerPopulationIndicatorAmount = (int)Math.Ceiling(
```
hhyyrylainen, line 1132: As this whole long block is inside this if, I'd recommend making this as a separate helper method.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1109,6 +1126,58 @@ private void AddPatchNode(Patch patch, Vector2 position)
...
+                patch.GetSpeciesSimulationPopulation(playerSpecies) * 0.004);
+
+            var indicatorExcess = Math.Clamp(playerSpeciesPopulationIndicators.Count - playerPopulationIndicatorAmount,
+                0,
+                playerSpeciesPopulationIndicators.Count);
+
+            for (int i = 0; i < indicatorExcess; ++i)
+            {
+                playerSpeciesPopulationIndicators.Last().QueueFree();
```
hhyyrylainen, line 1141: I think it would be better to just hide the excess indicators, I don't expect us to have so many excess ones that they'd cause an issue during the lifetime of the map drawer node.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1109,6 +1126,58 @@ private void AddPatchNode(Patch patch, Vector2 position)
 
         patch.ApplyPatchEventVisuals(node);
 
+        var playerSpecies = patch.FindSpeciesByID(PlayerSpeciesID);
+        if (playerSpecies != null)
+        {
+            var playerPopulationIndicatorAmount = (int)Math.Ceiling(
+                patch.GetSpeciesSimulationPopulation(playerSpecies) * 0.004);
```
hhyyrylainen, line 1133: Making the constant here into a value in Constants would make it easier for someone else to tweak how many circles appear.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1109,6 +1126,58 @@ private void AddPatchNode(Patch patch, Vector2 position)
...
+                }
+
+                indicator.MouseFilter = MouseFilterEnum.Ignore;
+
+                var nodeModifier = node.Position.LengthSquared();
+                var modifierSinus = MathF.Sin(i);
+
+                if (noCached)
+                {
+                    populationIndicatorContainer.AddChild(indicator);
```
hhyyrylainen, line 1170: This could be combined with the above if where this is instantiated for easier reading of this block of code.

hhyyrylainen (changes requested):
I think this is in better shape than when I last looked at this. There's still some minor stuff in the code I'd like to see improved.

hhyyrylainen:
That's why I asked it to be made lower contrast so that it just looks like a bit of background flair and not really anything the player needs to pay attention to.

CI09:
The idea was to persuade player to use migration tool more. It's safer for player to spread onto as many patches as possible but migration can be easily forgotten in my opinion as main feature of editor is, well, cell editor. If it gets into RC, we can ask what others think of the visual look and maybe come up with something better.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1109,10 +1116,67 @@ private void AddPatchNode(Patch patch, Vector2 position)
...
+    {
+        var playerPopulationIndicatorAmount = (int)Math.Ceiling(patch.GetSpeciesSimulationPopulation(playerSpecies) *
+            Constants.PLAYER_POPULATION_POPULATION_FOR_PER_INDICATOR);
+
+        var indicatorExcess = Math.Clamp(playerSpeciesPopulationIndicators.Count - playerPopulationIndicatorAmount,
+            0,
+            playerSpeciesPopulationIndicators.Count);
+
+        // Hide excess from the end of the list
+        for (int i = 0; i < indicatorExcess; ++i)
```
hhyyrylainen, line 1139: If you want to keep things this way, I'll accept it, but I think a slightly easier way to program this would be to use a variable like, `nextIndicatorIndex` that starts at 0 and goes up for each used indicator. Then the excess indicators can be hidden by looping from the ending nextIndicatorIndex value up to the indicators count. And the `noCached` variable can be calculated as `nextIndicatorIndex >= playerSpeciesPopulationIndicators.Count`.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1109,10 +1116,67 @@ private void AddPatchNode(Patch patch, Vector2 position)
...
+                populationIndicatorContainer.AddChild(indicator);
+            }
+            else
+            {
+                indicator = playerSpeciesPopulationIndicators[i];
+
+                indicator.Show();
+            }
+
+            indicator.MouseFilter = MouseFilterEnum.Ignore;
```
hhyyrylainen, line 1167: This should be in the indicator creation as it doesn't seem like this value is ever updated so it wastes time to reapply the value on each loop iteration.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1109,10 +1116,67 @@ private void AddPatchNode(Patch patch, Vector2 position)
...
+
+            var nodeModifier = node.Position.LengthSquared();
+            var modifierSinus = MathF.Sin(i);
+
+            playerSpeciesPopulationIndicators.Add(indicator);
```
hhyyrylainen, line 1172: Isn't this a major bug where the same indicator objects are added again and again to the same list?

assets/textures/gui/bevel/MapDotIndicator.svg
hhyyrylainen, line 1: I think that maybe if the dot was fully filled in and made slightly less in opacity, it would be a lot more subtle effect? That way I think there wouldn't be really any complaining about the added visuals.

src/microbe_stage/editor/PatchMapDrawer.cs
hhyyrylainen: I think this is probably what caused this error when I moved patches in the editor:

```
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at PatchMapDrawer.AddPlayerPopulationIndicators(Patch patch, Species playerSpecies, Control node, Vector2 position) in /home/hhyyrylainen/Projects/Thrive/src/microbe_stage/editor/PatchMapDrawer.cs:line 1141
   at PatchMapDrawer.AddPatchNode(Patch patch, Vector2 position) in /home/hhyyrylainen/Projects/Thrive/src/microbe_stage/editor/PatchMapDrawer.cs:line 1122
   at PatchMapDrawer.RebuildMap() in /home/hhyyrylainen/Projects/Thrive/src/microbe_stage/editor/PatchMapDrawer.cs:line 1069
   at PatchMapDrawer._Process(Double delta) in /home/hhyyrylainen/Projects/Thrive/src/microbe_stage/editor/PatchMapDrawer.cs:line 175
   at Godot.Node.InvokeGodotClassMethod(godot_string_name& method, NativeVariantPtrArgs args, godot_variant& ret) in /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Generated/GodotObjects/Node.cs:line 2395
   at Godot.CanvasItem.InvokeGodotClassMethod(godot_string_name& method, NativeVariantPtrArgs args, godot_variant& ret) in /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Generated/GodotObjects/CanvasItem.cs:line 1505
   at Godot.Control.InvokeGodotClassMethod(godot_string_name& method, NativeVariantPtrArgs args, godot_variant& ret) in /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Generated/GodotObjects/Control.cs:line 2912
   at PatchMapDrawer.InvokeGodotClassMethod(godot_string_name& method, NativeVariantPtrArgs args, godot_variant& ret) in /home/hhyyrylainen/Projects/Thrive/.godot/mono/temp/obj/Debug/Godot.SourceGenerators/Godot.SourceGenerators.ScriptMethodsGenerator/PatchMapDrawer_ScriptMethods.generated.cs:line 318
   at Godot.Bridge.CSharpInstanceBridge.Call(IntPtr godotObjectGCHandle, godot_string_name* method, godot_variant** args, Int32 argCount, godot_variant_call_error* refCallError, godot_variant* ret) in /root/godot/modules/mono/glue/GodotSharp/GodotSharp/Core/Bridge/CSharpInstanceBridge.cs:line 24
```

hhyyrylainen (changes requested):
I playtested this now and noticed a bug where on the initial generation (normal and freeplay) the population indicators do not display at all. I think that needs to be fixed. Also using the "move to this patch" button causes an error immediately upon press (I commented the callstack separately). So there's a few gameplay bugs and some quite small remaining code related problems.

src/microbe_stage/editor/PatchMapDrawer.cs
CI09: Good you caught that.

CI09:
image

src/microbe_stage/editor/PatchMapDrawer.cs
hhyyrylainen: Now with the persistent index for used indicators, I think this hiding excess indicators code needs to be put into a single place that is called *after* the entire map is drawn. Otherwise the last indicators get hidden over and over (and I think your indicatorExcess math was not updated anyway so that would lead to a bug in any case).

CI09:
Excess indicators are now hidden after all patch nodes are dones.
Though I get this error:
image

hhyyrylainen:
Isn't that again just the usual Godot texture errors on game exit that happen very often? I think it's unlikely any of your changes in this PR would affect that.

hhyyrylainen:
This probably won't make it in for 0.8.1 so I'll push this back to release 0.8.2.

hhyyrylainen:
Was this updated after my last review so that I should look at this again?

PR #5889, opened 2025-02-09
Add organelles cost display

### LINKED ISSUE #3631: Add showing the nutrient (initial composition) costs of organelles in the editor
https://forum.revolutionarygamesstudio.com/t/nutrient-cost-trait/914

so add the ammonia and phosphate costs of organelles in their tooltips (should be loaded from the JSON). Also maybe good to then show the overall nutrient cost in the organism statistics panel as well then.

Patryk26g:

Add info about nutrient cost to tooltips
Widened the organelle tooltip a little bit to better fit the info 
Screenshot_20250210_001409

**Related Issues**

Closes #3631

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I recently saw a video of someone playing the game in Russian, which made the organelle tooltips longer causing issues with the bottom being cutoff. But I suppose that's a problem for later solving rather than needing to be solved in this PR.

Patryk26g:
So I should revert the width changes?

hhyyrylainen:
> So I should revert the width changes?

I don't think that's necessary. I didn't even notice in the screenshot.

hhyyrylainen:
> I shoved it at the end of the tooltip

Tooltip? Sorry if I was not clear I only meant the right side organism statistics that really need to put it last. I thought the tooltips... well it doesn't really make sense to do the same fix there. Because if a tooltip is too tall to appear on screen there is no way to currently scroll it so it is anyway a pretty bad issue if the tooltip doesn't fit entirely. So there I think it is much less use to reorder the tooltip, though I guess even there putting it down near the "press F1 for more info" maybe makes some sense?

Patryk26g:
Ahhh sorry, of course I meant right statistics panel and not the tooltip :)

Patryk26g:
Sometimes I forget what tooltip really means and I use it too broadly, my bad

src/multicellular_stage/CellBodyPlanInternalCalculations.cs
```diff
@@ -78,6 +78,28 @@ public static float CalculateSpeed(IReadOnlyList<HexWithData<CellTemplate>> cell
         return speed / cells.Count + addedSpeed / (massEstimate * 1.4f);
     }
 
+    /// <summary>
+    ///   Calculates a colony's nutrient cost
+    /// </summary>
+    public static (int AmmoniaCost, int PhosphatesCost) CalculateOrganellesCost(
+        IReadOnlyList<HexWithData<CellTemplate>> cells)
+    {
+        float ammoniaCostTotal = 0, phosphatesCostTotal = 0;
```
hhyyrylainen, line 87: Each variable should be declared on its own line.

src/microbe_stage/MicrobeInternalCalculations.cs
```diff
@@ -409,6 +409,20 @@ public static Dictionary<Enzyme, float> CalculateDigestionEfficiencies(IEnumerab
         return result;
     }
 
+    public static (int AmmoniaCost, int PhosphatesCost) CalculateOrganellesCosts(
+        IEnumerable<OrganelleTemplate> organelles)
+    {
+        float ammoniaCostTotal = 0, phosphatesCostTotal = 0;
```
hhyyrylainen, line 415: More variables on a single line here

hhyyrylainen (changes requested):
Functionality-wise this seems already very good. There's just a few small technical details and typos I think need fixing. As a final point could you add a tooltip for hovering over the ammonia and phosphates costs? Because the average player might have no clue what they mean. I think the tooltip for both should be the same and be something along the lines of "This shows how many compounds your cell needs to divide. Higher values mean that reproduction takes more time if there are no compensating extra compound generation or gathering."

hhyyrylainen (approved):
This should be fine now, I just did a couple of really tiny changes. Two of the tooltips are now too tall, but I'll open a separate issue for that.

PR #5892, opened 2025-02-10
Fix membrane stats

### LINKED ISSUE #5266: Comparing cellulose to chitin membrane stats doesn't display all stats correctly
Instead it only shows one of the stats like this:

Switching to the other type then doesn't even compare to itself correctly, only when switching from another membrane type to cellulose does the tooltip return to displaying correctly:

Patryk26g:

Fixes membrane differences stats

**Related Issues**

Closes #5266

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/tooltip/ToolTipManager.tscn
+++ b/src/gui_common/tooltip/ToolTipManager.tscn
@@ -283,6 +283,16 @@ layout_mode = 2
 DisplayName = "HEALTH"
 ModifierIcon = ExtResource("13")
 
+[node name="physicalResistance" parent="GroupHolder/membraneSelection/single/MarginContainer/VBoxContainer/ModifierList" index="3" instance=ExtResource("2")]
+layout_mode = 2
+DisplayName = "PHYSICAL_RESISTANCE"
+ModifierIcon = ExtResource("15")
+
+[node name="toxinResistance" parent="GroupHolder/membraneSelection/single/MarginContainer/VBoxContainer/ModifierList" index="4" instance=ExtResource("2")]
+layout_mode = 2
+DisplayName = "TOXIN_RESISTANCE"
+ModifierIcon = ExtResource("16")
+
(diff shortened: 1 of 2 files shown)


hhyyrylainen (changes requested):
Besides the one code thing I think it is a bit weird that when you click on a membrane now it almost always results in the tooltip suddenly getting more stuff in it once it is the active one. I think that needs fixing. Either we always show all of the stats, which I guess gets a bit overwhelming as there must have been reason some are hidden, or we don't add those +0% when comparing a membrane with itself.

Patryk26g:
Then it will be just empty

hhyyrylainen:
Well on master branch the other than just that one bugged pair of tooltips manages to display correctly... So maybe there's some other kind of fix that could be made than this approach? I don't have a concrete suggestion though.

I think it would be better to just not display any of the +0 stats than having one appear when you click something... (in case there's no other fix possible).

Patryk26g:
image
image

hhyyrylainen:
If there's no better fix available, I think that's preferable to one random extra modifier appearing in the tooltip after click.

Patryk26g:
The other alternative is to always display all modifiers no matter if their change is 0%

hhyyrylainen:
What's the reason that one pair of membranes can't be special cased or something to force fix it just for that one situation? As much as I don't like hardcoding it like that, the other fix solutions don't seem to be any better.

Patryk26g:
Because the issue was not just with this one pair of membranes but almost all of them, example of silica-single:

Pre-fix:

Screenshot_20250210_153412

Post-fix:

Screenshot_20250210_153315

hhyyrylainen (approved):
I guess this now makes the most sense in how to solve the issue with the membrane comparison tooltips.

Edit: I also now tested in game.

PR #5901, opened 2025-02-12
Added support for exclusive fullscreen mode

### LINKED ISSUE #5204: Weird border artifact when fullscreen on Windows
Caused by Godot bug: https://github.com/godotengine/godot/issues/63500

Also causes the fullscreen resolution to be slightly off from the actual monitor resolution

### LINKED ISSUE #5205: Add the new Godot 4 fullscreen modes to be selectable
This adds a workaround for: https://github.com/Revolutionary-Games/Thrive/issues/5204

axQuadratic:

This PR implements support for Godot 4's exclusive fullscreen display mode, replacing the previous fullscreen toggle with a drop-down menu. The behaviour of the Alt+Enter hotkey is preserved.

**Related Issues**

Closes #5205. The exclusive fullscreen mode also appears to resolve #5204.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/gui_common/FocusGrabber.tscn
hhyyrylainen, line 1: Thanks for the PR. I don't have time to do a full review yet (and might not this week). But I have a little bit of time to comment on a few specific things. Some other people from the team might be able to do a full review sooner.

Is this file related to the full screen functionality? Or could these changes be left out from the PR?

src/engine/Settings.cs
```diff
@@ -63,10 +63,11 @@ public enum StrainBarVisibility
     // Graphics Properties
 
     /// <summary>
-    ///   Sets whether the game window is in fullscreen mode
+    ///   Sets window mode of the game window
     /// </summary>
     [JsonProperty]
-    public SettingValue<bool> FullScreen { get; private set; } = new(true);
+    public SettingValue<DisplayServer.WindowMode> DisplayMode { get; private set; } =
```
hhyyrylainen, line 69: I think we should have our own enum to guard against Godot changing theirs and messing stuff up.
I think we'll need just Windowed, Fullscreen and ExclusiveFullscreen values.

src/gui_common/FocusGrabber.tscn
axQuadratic: I am actually unsure of why it was modified in the first place. I doubt leaving it out would break anything; I will do that.

src/engine/FullScreenToggle.cs
```diff
@@ -22,8 +22,17 @@ public void ToggleFullScreenPressed()
...
+            Settings.Instance.DisplayMode.Value == DisplayServer.WindowMode.ExclusiveFullscreen)
+        {
+            Settings.Instance.DisplayMode.Value = DisplayServer.WindowMode.Windowed;
+        }
+        else
+        {
+            Settings.Instance.DisplayMode.Value = DisplayServer.WindowMode.Fullscreen;
+        }
+
+        GD.Print("Toggling fullscreen with keypress, new value: ", Settings.Instance.DisplayMode.Value);
```
hhyyrylainen, line 35: Would it make sense if the fullscreen toggle switched now between the 3 states? Because otherwise this won't know which fullscreen mode to return to... (though I suppose logic could also be added to detect when the first time fullscreen mode is entered and then remember that, but that wouldn't work if the user just selected windowed mode in the options menu and then restarted the game)

src/gui_common/FocusGrabber.tscn
hhyyrylainen: Looks like it was an automatic upgrade to Godot 4 syntax, but I'd prefer to keep that out of this PR to make the diff smaller. I'll put on my TODO list for next week to reimport that file so that the same thing doesn't happen in the future.

src/engine/Settings.cs
```diff
@@ -63,10 +63,11 @@ public enum StrainBarVisibility
...
+        new(DisplayServer.WindowMode.Fullscreen);
```
hhyyrylainen, line 70: Also as basically everyone on Windows is impacted, I think we might just want to have exclusive fullscreen be the default mode, because we'd still get bug reports about the border and wrong resolution probably if it isn't picked by default.

src/engine/FullScreenToggle.cs
axQuadratic: As in, it would transition in order like Windowed -> Fullscreen -> Exclusive, or similar? I feel like that might be unintuitive considering how little visual difference there is between the Exclusive and Fullscreen modes.

hhyyrylainen: Yeah, but otherwise it would also be pretty weird if the fullscreen toggle button selected the fullscreen mode that didn't work well on your computer...

axQuadratic: I suppose that's a good point. I'll change that, then.

axQuadratic:
This new commit should address all of the earlier points.

hhyyrylainen (approved):
I added a few error checks and corrected one possible bug with maximized handling. In my testing this seems to be fine so this should be good to merge now. I decided to prioritize getting this fix in as a lot of Windows players have been reporting the fullscreen problem. I didn't test on Windows but according to the docs this should fix the border around the game.

PR #5922, opened 2025-02-20
Water current visuals

### LINKED ISSUE #4739: Test if FastNoiseLite would be faster in the C++ module
this might give some performance improvement for the game as chunk pushing current calculation (at least used to) take quite a bit of processing time

dligr:

Implements water current visuals using GPU particles. Also makes currents differ per patch and affect cells.

<details>
  <summary>Current demonstration</summary>
  
   https://github.com/user-attachments/assets/7a9ed9c1-ddd4-483e-9cf0-bc1fa1e757c6
</details>

**Related Issues**

A roadmap item (https://wiki.revolutionarygamesstudio.com/wiki/Release_Roadmap#0.8.x_(world_and_auto-evo_updates))

Closes #4739 (no longer necessary as the current system now uses noise textures)

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Patryk26g:
Looks great! But I think it would be good idea to add a little bit of variation in the size of these particles. And I think that the currents should stop when you pause the game

dligr:
> GPU particles seems pretty bad in terms of performance in general in Godot. Might be better to switch to CPU particles but that might not also perform that well. In the worst case we need some kind of custom visuals shader on a plane to represent the currents (which kind of was my initial guess as to how to implement the visuals).

CPU particles might be unusable here, because they don't support particle shaders, which I used to apply currents to particles.

It's also possible to slightly optimize GPU particles by setting their FPS to a smaller value, but turning on interpolation, thus avoiding some calculations.

hhyyrylainen:
Has the visuals now been changed or is there a new thing to look at here? Now that we are done with the release I can have a proper look at this once the approach is locked in (performance is good / acceptable, and graphics are satisfactory).

dligr:
I think that the general approach is locked in. Performance doesn't seem awful (in fact, it might be a bit better, as the currents system doesn't have to calculate noise in real-time now). The programming side of visuals (i.e. particles) is done, though we might want to experiment with the particles' sprites, maybe also making them vary per-patch.

src/microbe_stage/components/CurrentAffected.cs
```diff
@@ -11,4 +11,15 @@
 [JSONDynamicTypeAllowed]
 public struct CurrentAffected
 {
+    public float EffectStrength;
+
+    public CurrentAffected()
+    {
+        EffectStrength = 1.0f;
+    }
+
```
hhyyrylainen, line 20: Isn't a parameterless constructor a bad idea in structs? I think I recall seeing errors about this or something...

src/microbe_stage/components/CurrentAffected.cs
```diff
@@ -11,4 +11,15 @@
...
+    public CurrentAffected(float strength)
```
hhyyrylainen, line 21: For JSON safety the strenght parameter name should be `effectStrength`

src/microbe_stage/NoiseDisturbanceX.tres
hhyyrylainen, line 1: This file and the Y variant need more descriptive names.

src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -888,15 +886,16 @@ private void PartialDiffuseEdges(int x0, int y0, int width, int height, float de
...
+    private void PartialAdvectCenter(int x0, int y0, int width, int height, float delta)
     {
         for (int x = x0; x < x0 + width; ++x)
         {
             for (int y = y0; y < y0 + height; ++y)
             {
                 if (OldDensity[x, y].LengthSquared() > 1)
                 {
-                    var velocity = fluidSystem!.VelocityAt(pos + new Vector2(x, y) * Resolution) * VISCOSITY;
+                    var worldPos = ConvertToWorld(x, y);
```
hhyyrylainen, line 897: ConvertToWorld is a much more expensive operation than just adding a `Vector2 pos` here. So could you explain why this is needed? And why this is fine in terms of performance. I didn't profile the performance but the cloud performance is a *major* problem for low end laptops so I really dislike any change that might make the situation worse.

simulation_parameters/microbe_stage/biomes.json
```diff
@@ -1511,7 +1535,11 @@
         "z": 0.75
       }
     },
+    "WaterCurrentSpeed": 1.5,
+    "WaterCurrentChaoticness": 2.0,
+    "WaterCurrentScale": 1.0,
     "CompoundCloudBrightness": 2.5,
```
hhyyrylainen, line 1541: This block is inconsistent with the other ones where all of the new properties are next to each other without gaps.

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -15,11 +18,31 @@
...
+render_priority = -2
+shader = ExtResource("4_ced6l")
+shader_parameter/warmColour = Color(1, 0.359565, 0.239655, 1)
+shader_parameter/coldColour = Color(0.233659, 0.0244511, 0.882812, 1)
+shader_parameter/uvOffset = Vector2(0, 0)
+shader_parameter/heatThreshold = 0.5
+shader_parameter/alphaMultiplier = 0.077
+shader_parameter/noiseMultiplier = 0.099
+shader_parameter/heat = ExtResource("5_jrgee")
+shader_parameter/noise = SubResource("NoiseTexture2D_d28l8")
```
hhyyrylainen, line 38: These seem like unused resources so you should restart godot editor and then resave this file which should get rid of these.

src/microbe_stage/systems/FluidCurrentsSystem.cs
```diff
@@ -73,34 +86,84 @@ public FluidCurrentsSystem(float currentsTimePassed) : base(TemporarySystemHelpe
...
+            var disturbancesX = noiseDisturbancesX.GetData();
+            var disturbancesY = noiseDisturbancesY.GetData();
+            var currentsX = noiseCurrentsX.GetData();
+            var currentsY = noiseCurrentsY.GetData();
+
+            noiseWidth = disturbancesX[0].GetWidth();
+            noiseHeight = disturbancesY[0].GetHeight();
+
+            int noiseDepth = noiseDisturbancesX.Depth;
+            noiseDisturbancesXImage = new Image[noiseDepth];
```
hhyyrylainen, line 135: Why does this use a *bunch* of images? Wouldn't one be enough? This seems like much worse for performance than this could be.

src/microbe_stage/systems/FluidCurrentsSystem.cs
```diff
@@ -116,7 +179,18 @@ protected override void Update(float delta, in Entity entity)
         var pos = new Vector2(position.Position.X, position.Position.Z);
         var vel = VelocityAt(pos) * Constants.MAX_FORCE_APPLIED_BY_CURRENTS;
 
-        physicsControl.ImpulseToGive += new Vector3(vel.X, 0, vel.Y) * delta;
+        physicsControl.ImpulseToGive += new Vector3(vel.X, 0, vel.Y) * delta
+            * entity.Get<CurrentAffected>().EffectStrength;
         physicsControl.PhysicsApplied = false;
     }
+
+    private float GetPixel(float x, float y, float z, Image[] array)
```
hhyyrylainen, line 187: This should be marked as aggressive inlining as I think that might help a bit with the performance.

src/microbe_stage/FluidCurrentDisplay.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        GlobalPosition = new Vector3(GlobalPosition.X, 1.0f, GlobalPosition.Z);
+
+        if (!PauseManager.Instance.Paused)
+        {
+            time += (float)delta;
+
+            material.SetShaderParameter(gameTimeParameterName, time);
+
+            SpeedScale = 1.0f;
```
hhyyrylainen, line 38: This *might* need to be given the current game speed from the microbe world simulation as otherwise going to 2x mode might look a bit weird if a bunch of particles keep moving at the fixed speed while microbes speed up.

shaders/CurrentsParticles.gdshader
```diff
@@ -0,0 +1,122 @@
+shader_type particles;
+render_mode disable_velocity;
+
+uniform vec3 emission_box_extents;
```
hhyyrylainen, line 4: I assume this name needing to be like this comes from godot.

shaders/CurrentsParticles.gdshader
```diff
@@ -0,0 +1,122 @@
...
+// The following constants should be the same as in FluidCurrentsSystem.cs
+const float disturbanceTimescale = 1.0f;
+const float currentsTimescale = 1.0f / 500.0f;
+const float currentsStretchingMultiplier = 1.0f / 10.0f;
+const float disturbanceToCurrentsRatio = 0.15f;
+const float minCurrentIntensity = 0.4f;
+const float positionScaling = 0.9f;
+
+// Generates a random float and modifies the seed
+float rand_from_seed(inout uint seed) {
```
hhyyrylainen, line 29: This shouldn't have naming convention limitations, right? So the name should be `RandFromSeed`

shaders/CurrentsParticles.gdshader
```diff
@@ -0,0 +1,122 @@
...
+
+uint hash(uint x) {
+    x = ((x >> uint(16)) ^ x) * uint(73244475);
+    x = ((x >> uint(16)) ^ x) * uint(73244475);
+    x = (x >> uint(16)) ^ x;
+    return x;
+}
+
+void start() {
+    uint base_number = NUMBER;
```
hhyyrylainen, line 52: There's also a few variables here that need renames to be consistent with our naming convention.

hhyyrylainen:
I ended up commenting on quite many technical details... I think the general approach is fine and if those minor problems (as well as performance) are corrected I think this could be merged. As long as gameplay balance testing is positive enough (we should ask for team-wide testing once the code is good).

hhyyrylainen:
Also before testing I think it would be nice to have a currents strength slider in the new game planet section, that way it'll be much easier test the game balance when other team members can easily try other values, and in the worst case if the feature ends up problematic the default currents strength could be set low enough to not affect the gameplay as much.

This can then be a final touch but I think for low end devices we need an option to turn off the current visuals in the options menu.

src/microbe_stage/components/CurrentAffected.cs
dligr: As far as I know, parameterless constructors weren't allowed for structs before C# 10.0 (shouldn't be a problem because we're using a newer version), and writing those caused compilation errors

HyperbolicHadron:
I really like this change, it makes the different environments dynamic and visually interesting. I do have a few concerns though:
It seems like the compounds tend to collect in lines/square borders quite often, like this image of the tidepool:
image
Also, I think that the cell membranes should only move when the organism is being controlled to move, and not just being moved by the currents. (It makes it kind of hard to tell how the organisms are moving, especially when you are trapped in a strong current. It also makes it look like your organism is being autonomously controlled instead of passively moved)

Anyway, I have only tried the early game so far, but it has made that more interesting and fun. I think that these problems are mostly visual and that the gameplay at least is quite improved by this change in its current state.

src/microbe_stage/systems/FluidCurrentsSystem.cs
dligr: The fluid current system uses 3D noise, which means using an array of images; 3D noise has to be used so that currents change over time.

dligr: (Each of the 4 3D noises are 64x64x64, which shouldn't be too bad performance-wise)

src/microbe_stage/CompoundCloudPlane.cs
dligr: I optimized this a bit by moving `ConvertToWorld` out of the loop, and then just adding x,y with the modifier of `Resolution`; this should be equivalent to what was calculated before

shaders/CurrentsParticles.gdshader
dligr: It actually doesn't, I just forgot to rename it initially

shaders/CurrentsParticles.gdshader
dligr: I have now renamed this method, as well as other methods and variables according to the style guide

dligr:
> It seems like the compounds tend to collect in lines/square borders quite often, like this image of the tidepool

This might have to do with this line of code:
```
var currentsVelocity = new Vector2(
    Math.Abs(currentsX) > 0.4 ? currentsX : 0.0f,
    Math.Abs(currentsY) > 0.4 ? currentsY : 0.0f);
```
It floors currents that are less than 0.4, but it works per-axis, so it might result in currents going along one of the axes.

This can probably be fixed by making the clamp use true distance

HyperbolicHadron:
I did a fair bit of testing with my low end system, a Surface Pro 3. The performance with the latest version seems to have dropped the performance from 10-12 fps to 7-8 fps.This is using the opengl rendering mode.

HyperbolicHadron:
Actually, that computer is currently getting 7-8 fps on 0.8.1.1, which is basically the same as I am getting on this PR, so it probably isn't this PR that decreased the performance.

Patryk26g:
Works great! The only issue I have is that there tends to be this square pattern from compound clouds

Screenshot_20250304_215550
Screenshot_20250304_215337

src/microbe_stage/components/CurrentAffected.cs
hhyyrylainen: Reading the specification: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/parameterless-struct-constructors this seems pretty error prone as there are situations when default initialization happens but this constructor *will not run*. For that reason I'd like this component to follow the same approach as other components, i.e. default constructor must initialize everything to 0. In the strength system, you should then assume that 0 means the same things as 1. I've had to do a similar thing for other components where a modifier value of 0 (which is default initialized) needs to be considered the "default speed".

src/microbe_stage/systems/FluidCurrentsSystem.cs
hhyyrylainen: Ah, I see that's probably the case. Did you measure the RAM usage impact? I would expect that it's just some dozens of megabytes so it doesn't matter, but verifying my expectation with actual numbers would be good.

src/microbe_stage/CompoundCloudPlane.cs
hhyyrylainen: That sounds fine. I'll mark this discussion as resolved but I didn't get the info on why this change was necessary...

src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -888,15 +886,17 @@ private void PartialDiffuseEdges(int x0, int y0, int width, int height, float de
...
+        var worldPos = ConvertToWorld(0, 0);
         for (int x = x0; x < x0 + width; ++x)
         {
             for (int y = y0; y < y0 + height; ++y)
             {
                 if (OldDensity[x, y].LengthSquared() > 1)
                 {
-                    var velocity = fluidSystem!.VelocityAt(pos + new Vector2(x, y) * Resolution) * VISCOSITY;
+                    var velocity = fluidSystem!.VelocityAt(
+                        new Vector2(worldPos.X + x * Resolution, worldPos.Z + y * Resolution)) * VISCOSITY;
```
hhyyrylainen, line 899: Sorry to nitpick, but as the clouds is one of the most performance critical systems, I think the Resolution should be read to a local variable outside the loop, this way the property (reading of which emits a method call operation) is read only once. (same in the below diff block).

src/microbe_stage/systems/FluidCurrentsSystem.cs
```diff
@@ -73,34 +87,84 @@ public FluidCurrentsSystem(float currentsTimePassed) : base(TemporarySystemHelpe
...
+        return currentsVelocity.Lerp(disturbancesVelocity, DISTURBANCE_TO_CURRENTS_RATIO) * speed;
     }
 
     protected override void PreUpdate(float delta)
     {
         base.PreUpdate(delta);
 
+        if (!imagesInitialized)
+        {
+            var disturbancesX = noiseDisturbancesX.GetData();
```
hhyyrylainen, line 127: Can't this return null before the noise is generated? I think I had to take that into account in the thermal heat system (so the system needs to retry reading the data on each PreUpdate (and maybe skip running) until the noise image data is ready).

src/microbe_stage/systems/FluidCurrentsSystem.cs
```diff
@@ -116,7 +180,19 @@ protected override void Update(float delta, in Entity entity)
         var pos = new Vector2(position.Position.X, position.Position.Z);
         var vel = VelocityAt(pos) * Constants.MAX_FORCE_APPLIED_BY_CURRENTS;
 
-        physicsControl.ImpulseToGive += new Vector3(vel.X, 0, vel.Y) * delta;
+        physicsControl.ImpulseToGive += new Vector3(vel.X, 0, vel.Y) * delta
+            * entity.Get<CurrentAffected>().EffectStrength;
```
hhyyrylainen, line 184: This would be the place to read effect strength into a variable and if it is 0 set it to 1 and if it is negative *then* skip applying any force (as -1 seems like a good value to indicate disabling as 0 can't be used for that).

src/microbe_stage/systems/FluidCurrentsSystem.cs
```diff
@@ -116,7 +180,19 @@ protected override void Update(float delta, in Entity entity)
...
         physicsControl.PhysicsApplied = false;
     }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private float GetPixel(float x, float y, float z, Image[] array)
+    {
+        if (x < 0.0f)
+            x = noiseWidth + x % noiseWidth;
```
hhyyrylainen, line 192: Doesn't this run into negative value modulo differences if x is somehow less than `-noiseWidth`?

HyperbolicHadron:
The lines/square pattern seems to be significantly reduced now, though not entirely gone:
Screenshot from 2025-03-05 14-53-02

I think it is not too visually distracting now if it can't be fixed entirely.

At the beginning of the game, when the compounds just start to be mixed, it is most prominent:
Screenshot from 2025-03-06 10-54-31
Though it usually smooths out more after the initial movement to look more natural.

src/microbe_stage/systems/FluidCurrentsSystem.cs
```diff
@@ -116,7 +160,55 @@ protected override void Update(float delta, in Entity entity)
...
+        float effectStrength = entity.Get<CurrentAffected>().EffectStrength;
+
+        if (effectStrength == 0)
+        {
+            effectStrength = 1;
+        }
+        else if (effectStrength == -1)
+        {
+            effectStrength = 0;
+        }
```
hhyyrylainen, line 172:
```suggestion
        else if (effectStrength < 0)
        {
            return;
        }
```

src/microbe_stage/systems/FluidCurrentsSystem.cs
```diff
@@ -116,7 +160,55 @@ protected override void Update(float delta, in Entity entity)
...
     }
+
+    private void TryGetNoiseImages()
+    {
+        var disturbancesX = noiseDisturbancesX.GetData();
+        var disturbancesY = noiseDisturbancesY.GetData();
+        var currentsX = noiseCurrentsX.GetData();
+        var currentsY = noiseCurrentsY.GetData();
+
+        if (disturbancesX == null || disturbancesY == null || currentsX == null || currentsY == null)
```
hhyyrylainen, line 185: I think it would be ever so slightly more efficient to check the values one by one and return early when one is missing.

hhyyrylainen:
The technical implementation side is now very close to perfect. Though hopefully Godot 4.4 doesn't break anything with this.

src/microbe_stage/systems/FluidCurrentsSystem.cs
Patryk26g: Will there be any difference? If the first statement is true the rest won't be evaluated because there are only || operators

hhyyrylainen: Yes because the variables are already retrieved before we start checking any of them. So by the time the if statement starts, the expensive work of fetching data from Godot C++ side is already done. So if the first one has already failed, we've tried to also fetch 3 useless pieces of data before we get to checking whether the first GetData call failed or not.

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -15,11 +18,31 @@
...
 [ext_resource type="PackedScene" uid="uid://dwkek50fju0fu" path="res://src/microbe_stage/PatchNameOverlay.tscn" id="85"]
 
-[node name="MicrobeStage" type="Node" node_paths=PackedStringArray("heatViewOverlay")]
+[sub_resource type="FastNoiseLite" id="FastNoiseLite_q12wg"]
+frequency = 0.2401
+
+[sub_resource type="NoiseTexture2D" id="NoiseTexture2D_d28l8"]
+noise = SubResource("FastNoiseLite_q12wg")
+
+[sub_resource type="ShaderMaterial" id="ShaderMaterial_3c4mw"]
```
hhyyrylainen, line 27: This has seemingly appeared again so could you again reset the shader on the heat gradient plane to default?

src/microbe_stage/FluidCurrentDisplay.cs
```diff
@@ -0,0 +1,74 @@
...
+#pragma warning disable CA2213
+    private ShaderMaterial material = null!;
+
+    private MicrobeStage stage = null!;
+#pragma warning restore CA2213
+
+    private StringName gameTimeParameterName = new("gameTime");
+    private StringName speedParameterName = new("speed");
+    private StringName chaoticnessParameterName = new("chaoticness");
+    private StringName scaleParameterName = new("scale");
```
hhyyrylainen, line 17: I think these should be made read-only

src/microbe_stage/FluidCurrentDisplay.cs
```diff
@@ -0,0 +1,74 @@
...
+
+    private float time;
+
+    public void Init(MicrobeStage microbeStage)
```
hhyyrylainen, line 21: Giving this the entire stage feels pretty wrong to me, so I think it would be better to just give a `IWorldSimulation worldTimeSource` as parameter.

src/microbe_stage/systems/FluidCurrentsSystem.cs
```diff
@@ -116,7 +160,63 @@ protected override void Update(float delta, in Entity entity)
...
         var vel = VelocityAt(pos) * Constants.MAX_FORCE_APPLIED_BY_CURRENTS;
 
-        physicsControl.ImpulseToGive += new Vector3(vel.X, 0, vel.Y) * delta;
+        float effectStrength = entity.Get<CurrentAffected>().EffectStrength;
+
+        if (effectStrength == 0)
+        {
+            effectStrength = 1;
+        }
+        else if (effectStrength == -1)
```
hhyyrylainen, line 169:
```suggestion
        else if (effectStrength < 0)
```
I'll just commit this change so that there's no negative forces being applied as I think that could end up with really silly results.

src/microbe_stage/FluidCurrentDisplay.cs
```diff
@@ -0,0 +1,74 @@
...
+
+    public override void _Process(double delta)
+    {
+        base._Process(delta);
+
+        GlobalPosition = new Vector3(GlobalPosition.X, 1.0f, GlobalPosition.Z);
+
+        if (!PauseManager.Instance.Paused)
+        {
+            material.SetShaderParameter(gameTimeParameterName, time);
```
hhyyrylainen, line 41: Is it intentional that this passes in the *old* time? I think I'll swap this around the way I think this should be.

hhyyrylainen (changes requested):
I fixed the technical code style problems, but now when I playtested for the first time I noticed two graphical problems.

There's these really ugly lines that show up in the compound clouds:



I think those need to be solved. Also the clouds seem to be mixed up a lot more than before so maybe the cloud system should scale back how much currents it applies as now the clouds are a really mixed blob, more than before.

And one minor graphical thing I noticed is that when riding a current the particles in your "current" are jittery. This is probably the classical problem of the particles updating on each frame, but the player position updates on only each game simulation update.

hhyyrylainen (approved):
I think this is now ready. I did a final test and didn't see the jitter and the clouds didn't mix into a total soup but I still noticed instances of: https://github.com/Revolutionary-Games/Thrive/issues/1924 (which is probably that other issue but the currents mixing things up more makes this more obvious).

I'll merge this now but I think some additional tweaks could be done later. I think that maybe the particle colours could be adjusted per patch so that for example the vent current particles could be orange like the background particles that are already present in the patch. And I think it would be maybe better to slightly bump up the current threshold again so that there are more currents, but then adjust the current velocity down for the compound clouds to still make those behave well.

PR #5923, opened 2025-02-20
Radiation glow

HexapodPhilosopher:

Adds emission and pulsing particles to radioactive rock chunks.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


shaders/RadioactiveChunk.gdshader
hhyyrylainen, line 1: What's the difference between this and the emissive chunk material? We should try to use as few materials as possible as each shader will be compiled separately and will cause lag spike on the first spawn (because we don't have shader preloading).

HexapodPhilosopher: The crystal shader always makes things look transparent. It's the reason I had to make the large phosphate chunk with the rock 2 meshes, as otherwise the rock was see-through.

hhyyrylainen: Could you then maybe rename this shader EmissiveChunk.gdshader? That way there could be future uses by other chunk types that also want to have emissive properties.

hhyyrylainen (approved):
Technical side of things is good now. And I think the visual effect is an improvement in the game, but there's a tiny bit of lighting weirdness that  has a harsh line jumping up and down the chunk. Though I'm not fully sure if that is truly a problem, but at least to my eyes that looks a bit like graphics glitching.

PR #5938, opened 2025-02-23
Tutorial Improvements

adQuid:

Updates a few sequences and a lot of text of tutorials, aiming to avoid leading the player down "rabbit holes" of the more complicated bits of the game, improve some clarity, optimize use of limited player attention, and add a little bit a scientific narrative poetry to do some story-telling while trying to explain the game.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


simulation_parameters/Constants.cs
```diff
@@ -1356,11 +1356,11 @@ public static class Constants
 
     public const float TIME_BEFORE_TUTORIAL_CAN_PAUSE = 0.01f;
 
-    public const float MICROBE_MOVEMENT_EXPLAIN_TUTORIAL_DELAY = 12.0f;
+    public const float MICROBE_MOVEMENT_EXPLAIN_TUTORIAL_DELAY = 3.5f;
     public const float MICROBE_MOVEMENT_EXPLAIN_TUTORIAL_DELAY_CONTROLLER = 1.0f;
-    public const float MICROBE_MOVEMENT_TUTORIAL_REQUIRE_DIRECTION_PRESS_TIME = 2.2f;
+    public const float MICROBE_MOVEMENT_TUTORIAL_REQUIRE_DIRECTION_PRESS_TIME = 1.6f;
```
hhyyrylainen, line 1361: I think this should not be reduced until we have a popup for the player to pick their preferred movement mode.

simulation_parameters/Constants.cs
```diff
@@ -1356,11 +1356,11 @@ public static class Constants
...
     public const float TUTORIAL_ENTITY_POSITION_UPDATE_INTERVAL = 0.2f;
-    public const float GLUCOSE_TUTORIAL_TRIGGER_ENABLE_FREE_STORAGE_SPACE = 0.14f;
+    public const float GLUCOSE_TUTORIAL_TRIGGER_ENABLE_FREE_STORAGE_SPACE = 0.03f;
```
hhyyrylainen, line 1363: Doesn't this cause problems with the player being able to find glucose way too soon? I've balanced the tutorial so that at most you need to spend like one second cluelessly in the glucose cloud wondering if it is working or not. With less storage space the player may need to spend a lot longer inside a cloud attempting to absorb it even when their storage is full.

src/tutorial/microbe_stage/GlucoseCollecting.cs
```diff
@@ -69,9 +69,17 @@ public override bool CheckEvent(TutorialState overallState, TutorialEventType ev
             case TutorialEventType.MicrobeCompoundsNearPlayer:
             {
                 var data = (EntityPositionEventArgs)args;
+                var compounds = ((EntityPositionEventArgs)args).Compounds;
+                var hud = ((EntityPositionEventArgs)args).HUD;
 
                 if (!HasBeenShown && data.EntityPosition.HasValue && CanTrigger && !overallState.TutorialActive())
                 {
+                    // force player microbe to have the right amount of Glucose missing
+                    compounds.TakeCompound(Compound.Glucose, compounds.GetCapacityForCompound(Compound.Glucose));
```
hhyyrylainen, line 78: I see this change. I personally would prefer to keep the approach where the tutorial only triggers after the player can reasonable be expected to have enough missing storage space.

I might be convinced to see this is better but I think 0.2 glucose buffer is pretty low safety margin and we should not make the player more likely to die during the tutorial.

src/tutorial/microbe_stage/DayNightTutorial.cs
```diff
@@ -21,8 +21,11 @@ public override bool CheckEvent(TutorialState overallState, TutorialEventType ev
         {
             case TutorialEventType.MicrobePlayerEnterSunlightPatch:
             {
+                MicrobeHUD hud = ((HUDEventArgs)args).HUD;
```
hhyyrylainen, line 24: I think it is pretty dirty to pass the HUD like this.

Also the environment panel has to be shown for temperature change with thermosynthesis.

I don't immediately have a good idea here, but I think that some kind of system like "wants pause" going through the tutorial state would be better. Maybe some kind of event system where the TutorialState offers an event callback any active HUD registers to, and individual tutorials can then send events through like "EnvironmentPanelRequired"?

src/tutorial/microbe_editor/EditorUndoTutorial.cs
```diff
@@ -32,9 +32,13 @@ public override bool CheckEvent(TutorialState overallState, TutorialEventType ev
         {
             case TutorialEventType.MicrobeEditorOrganellePlaced:
             {
-                if (!HasBeenShown && CanTrigger && !overallState.TutorialActive())
+                OrganelleDefinition placedOrganelle = ((OrganellePlacedEventArgs)args).Definition;
+
+                if (!HasBeenShown && CanTrigger && !overallState.TutorialActive()
```
hhyyrylainen, line 37: I really think we need to tutorialize the undo feature earlier than when placing the nucleus.

We could have another tutorial reminding about undo when placing the nucleus (if the player ATP balance / resource balance is not good). But I don't think we should give up the undo tutorial, because back in the dark ages people used to often ask on our discord how to undo. And I don't want to go back to that.

src/tutorial/microbe_editor/EditorRedoTutorial.cs
hhyyrylainen, line 1: I think the redo tutorial had a pretty nice flow to it and it didn't add too much time. And it synergizes really well with the general undo tutorial that is early in the process.

src/tutorial/TutorialEventArgs.cs
```diff
@@ -62,12 +62,16 @@ public void ReuseEvent(bool usesScreenRelativeMovement, Vector3 movementDirectio
 
 public class EntityPositionEventArgs : TutorialEventArgs
 {
-    public EntityPositionEventArgs(Vector3? position)
+    public EntityPositionEventArgs(Vector3? position, CompoundBag compounds, MicrobeHUD hud)
```
hhyyrylainen, line 65: Which compounds are these, the entity's or the player's? I think that needs naming clarification. Also I'd totally remove the HUD as I think it just doesn't fit at all into the idea of event args. At most I think a hack with the sender of the event could be done but even that seems not very clean design.

src/microbe_stage/MicrobeStage.cs
```diff
@@ -51,6 +51,9 @@ public partial class MicrobeStage : CreatureStageBase<Entity, MicrobeWorldSimula
     [JsonProperty]
     private bool wonOnce;
 
+    [JsonProperty]
+    private bool tutorialEndedOnce;
```
hhyyrylainen, line 55: I think this should be documented what this does. Also from the name I thought this would be set to true when tutorial has been completed, but this seems like it is set to true when the tutorial is skipped?

Also I think there is a missing detection of immediately turning off the tutorials which should return the usual default panel configuration.

hhyyrylainen (changes requested):
I didn't play with the new tutorial yet, so this is just an architectural review for now. I think the HUD control from individual tutorials needs a bit more work, but other than that other problems were just minor.

simulation_parameters/Constants.cs
adQuid: This variable could be put back probably given the other tutorials, but there's a new change later to set the glucose manually so that there's always enough free space to finish that tutorial phase.

hhyyrylainen: Yeah, I did see that code later. To me it doesn't seem exactly good that the player data is passed to the tutorial like that to be modified, that's part of the major architectural problem I think exists here I commented on in my review.

src/tutorial/microbe_stage/GlucoseCollecting.cs
adQuid: I was toying with the idea of having a check to make sure that the glucose never falls below 0.1 during this stage, but I didn't for two reason:

1.  it would be complex and likely error prone messaging code (and you said you wanted small incremental changes)
2. This is immediately after the player reads a message that says "if you don't have ATP you die". 5-10 seconds into the tutorial, I'm more afraid of players wondering what they are supposed to do after 15 seconds of nothing happening and getting lost in menus than dying from the thing the tutorial was talking about killing you and having to respawn

It's all subjective, but I think a key problem in the early tutorial is a lack of responsiveness. All the streamers I watched are eager to do things and see results, and our first tutorials don't do that very well.

src/tutorial/microbe_stage/DayNightTutorial.cs
adQuid: I agree that the args is a little abused in our tutorials. If you have a better pattern you prefer, however, you'll probably need to explain to me how it's going to work...

src/tutorial/microbe_editor/EditorUndoTutorial.cs
adQuid: This isn't the undo tutorial waiting until the player is ready for the nucleus, this is the editor tutorial telling the player to put a nucleus on _first_, then explaining why that would be a terrible idea and explaining the undo button.

It is entirely possible for a player to ignore this advice and skip this phase by doing different modifications, but I opted to not force the player to read more walls of text if they are aren't ignoring instructions.

src/tutorial/microbe_stage/DayNightTutorial.cs
hhyyrylainen: I did lay out a rough idea. After having it stew in my head for a while, I think it is probably the best option, so adding a method in TutorialState to trigger a tutorial event that is listened to by active HUD instances, would be the way to go in my opinion. Then the tutorials can send events like "compounds panel should be visible" etc. and the relevant HUD then reacts to it.

src/tutorial/microbe_editor/EditorRedoTutorial.cs
adQuid: I don't think anyone sees an undo button and can't guess what the button next to it does, and every time you interrupt flow to read a box of text is another couple players who get bored and turn off the tutorial before the important parts.

src/tutorial/microbe_editor/EditorUndoTutorial.cs
hhyyrylainen: But then the player is allowed to skip like half of the hands on tutorial for the first editor cycle, which I don't think is good...

src/tutorial/TutorialEventArgs.cs
adQuid: I'll rename the compounds. As for the HUD I have no attachment to this pattern, I just didn't see a cleaner way to send it.

src/microbe_stage/MicrobeStage.cs
adQuid: That's fair, I could have called this "canceled" instead...

src/tutorial/TutorialEventArgs.cs
hhyyrylainen: After reading the code it seems like it is the current player's data, which kind of doesn't fit logically as the position is for another entity and the player data is just tacked on... So I would really prefer if that data was also stripped so that we follow more separation of concerns (disallowing the tutorial from directly messing with the player state).

simulation_parameters/Constants.cs
adQuid: The problem here was that the buttons last a little longer than a lot of player's intuition on when a task should be completed, which I think is leading to players assuming they are supposed to be pressing combinations or something. To help with your concerns, I added a new change to both highlight the option to change the movement controls, and make the movement tutorial unclosable, in case someone needs more time (which I've seen happen once).

src/tutorial/microbe_editor/EditorUndoTutorial.cs
adQuid: The player is allowed to ship all of the phases by closing them (or just not paying attention). Hopefully reducing four consecutive text boxes to just three will let us hold on to that attention longer for the more important information.

Plus, this isn't canceling the other editor tutorial buttons; two boxes (showing non-vital information) are just not shown until a player chooses to try the nucleus.

hhyyrylainen: Is there safety against the player just placing another kind of organelle and then confirming to exit the editor? That would be really bad if the player can skip the most important tutorial in the game like that. I think that the changes here make it much more likely because if the undo tutorial doesn't trigger for every placed organelle and the nucleus needs to be found by scrolling down the list of available organelles, I think we are in a similar situation as the glucose balance display which is not visible by default.

src/tutorial/microbe_stage/DayNightTutorial.cs
```diff
@@ -23,6 +30,8 @@ public override bool CheckEvent(TutorialState overallState, TutorialEventType ev
             {
                 if (!HasBeenShown && CanTrigger && !overallState.TutorialActive())
                 {
+                    EnvironmentPanel.ShowPanel = true;
+                    HUDBottomBar.EnvironmentPressed = true;
```
hhyyrylainen, line 34: I think GUI state applying should more logically be in the ApplyGUIState method. Also I think it would be much safer to make a single method that ensures whatever the GUI wants to show is visible correctly, rather than trying to set the panel visibility and related button here.

src/tutorial/microbe_stage/GlucoseCollecting.cs
hhyyrylainen: I still think this is logically the wrong place to edit the player state. I think a better thing would be just to trigger the tutorial once there's enough space to absorb something.

locale/en.po
```diff
@@ -3437,17 +3433,16 @@ msgstr ""
 
 msgid "MICROBE_STAGE_COLLECT_TEXT"
 msgstr ""
-"Collect [thrive:compound type=\"glucose\"][/thrive:compound] (white clouds) by moving over it.\n"
-"\n"
-"Your cell needs [thrive:compound type=\"glucose\"][/thrive:compound] in order to produce energy ([thrive:compound type=\"atp\"][/thrive:compound]) to stay alive.\n"
+"To survive, a microbe will need [thrive:compound type=\"atp\"][/thrive:compound]. For your species, that means finding [thrive:compound type=\"glucose\"][/thrive:compound].\n"
```
hhyyrylainen, line 3436: I think this is a regression in understandability as this uses the more complex technical term before mentioning energy.

locale/en.po
```diff
@@ -3470,41 +3465,41 @@ msgstr ""
...
 
 msgid "MICROBE_STAGE_INITIAL_POND"
 msgstr ""
-"On a distant alien planet, eons of chemical evolution building up to self-replicating molecules have led to the development of a new phenomenon in the universe.\n"
+"On a distant alien planet, eons of chemical evolution have built up to self-replicating molecules, and the development of a new phenomenon.\n"
 "\n"
 "Life.\n"
 "\n"
-"Simple microbes reside in shallow warm ponds. You are the Last Universal Common Ancestor (LUCA) on this planet.\n"
+"This simple bubble has the ability to maintain a different environment inside of itself than in the water outside. In time, it may grow into a complex species that dominates the stars.\n"
```
hhyyrylainen, line 3496: This edited text makes the different start flavour texts less distinct...

locale/en.po
```diff
@@ -3470,41 +3465,41 @@ msgstr ""
...
 "\n"
-"To survive in this hostile world, you will need to collect any compounds that you can find and evolve across generations to compete against the other species of microbes."
+"But first, you will need to survive."
 
 msgid "MICROBE_STAGE_ORGANELLE_DIVISION"
 msgstr ""
-"One of your organelles has divided.\n"
+"In the process of growing enough to reproduce, each of your organelles will divide. These new organelles provide bonuses (and osmoregulation costs) just like the organelles your microbe started with.\n"
```
hhyyrylainen, line 3502: I take a bit of issue with calling new organelles providing bonuses, I'd cut that part to focus on that the new grown organelles act similarly to the ones the cell starts off with, there's just more of them.

locale/en.po
```diff
@@ -5859,9 +5854,9 @@ msgid "TUTORIAL_MICROBE_EDITOR_CELL_TEXT"
 msgstr ""
 "This is the cell editor, where you can evolve your species by spending [b]mutation points[/b] (MP) [thrive:icon]MP[/thrive:icon]. Each generation you will always have [b]100 MP[/b] [thrive:icon]MP[/thrive:icon] to spend, so don't bother saving up!\n"
 "\n"
-"The hex in the center of your screen is your [b]cell[/b], which is composed of a single cytoplasm part.\n"
+"The hex in the center of your screen is your species.\n"
 "\n"
-"To continue, select a part from the left panel. Then left-click next to the hex to place it. You can rotate parts with [thrive:input]e_rotate_left[/thrive:input] and [thrive:input]e_rotate_right[/thrive:input]."
+"To continue, scroll down on the left panel until you find the nucleus. Click on it, then left-click next to the hex to place it. You can rotate parts with [thrive:input]e_rotate_left[/thrive:input] and [thrive:input]e_rotate_right[/thrive:input]."
```
hhyyrylainen, line 5859: Yeah, I really think the scrolling here should absolutely be automatic to make it easier for the player. Nucleus should also be underlined or bolded as it is the most important word in this whole text.

locale/en.po
```diff
@@ -5873,13 +5868,13 @@ msgstr ""
 
 msgid "TUTORIAL_MICROBE_EDITOR_ENDING_TEXT"
 msgstr ""
-"Those are the basics of editing your species. As a finishing touch, you can rename your species by clicking on the name on the bottom left and editing the text.\n"
+"Adaptation is hard. There are many sources of [thrive:compound type=\"atp\"][/thrive:compound], but not all of them are in every patch. You can use all the tabs and panels here to make the most informed choice you can, or just try something and see how it works!\n"
```
hhyyrylainen, line 5871: I think if we are skipping the patch map, a tutorial that is seen before the patch map should not refer to patches at all as the player will have no clue what patches are.

locale/en.po
```diff
@@ -6018,9 +6009,9 @@ msgstr ""
 
 msgid "TUTORIAL_MICROBE_STAGE_REPRODUCE_TEXT"
 msgstr ""
-"To reproduce, you must duplicate your organelles. This happens gradually as you survive, but you can speed things up by gathering [thrive:compound type=\"ammonia\"][/thrive:compound] and [thrive:compound type=\"phosphates\"][/thrive:compound].\n"
+"Life needs more molecules than just sugar. See which other kinds of clouds your microbe can collect!\n"
```
hhyyrylainen, line 6012: We consistently use "compounds" word in the game to mean molecules + other stuff so I think for consistent messaging this should not use the word molecules.

locale/en.po
```diff
@@ -1765,13 +1765,9 @@ msgstr ""
 
 msgid "EDITOR_TUTORIAL_EDITOR_TEXT"
 msgstr ""
-"Welcome to the [b]Microbe Editor[/b].\n"
+"This early in the history of life, the biosphere is pretty simple. Later, the menus you see here will be essential to understanding the interconnected story unfolding on your alien world.\n"
```
hhyyrylainen, line 1768: I feel slightly bad about removing the part that tells the player that they are in the microbe editor as that's a pretty core term that we developers would have then a hard time talking to newer fans of the game if they never see the term anywhere.

locale/en.po
```diff
@@ -1765,13 +1765,9 @@ msgstr ""
...
 "\n"
-"Here you can review what has happened during previous generations, and then make changes to your species.\n"
-"\n"
-"The Report Screen gives you a list of changes in the patch you currently occupy. Try exploring the various tools at your disposal to learn more about events in your world!\n"
-"\n"
-"When you are ready, press the \"next\" button in the bottom right to continue to the Patch Map Screen."
+"For now, just click the \"Editor\" tab on top left to open the [b]Microbe Editor[/b]"
```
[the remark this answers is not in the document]
adQuid: I admit this is kind of a shortcut because I didn't want to add yet MORE GUI changes to this PR. Still, I'm not sure when I would feel comfortable opening the report screen to the player after locking it off. It is never "harmful" for a curious player to look at it, but it's not really "important" at any point in your first game, particularly if you aren't playing a heterotroph. I'm trying to avoid getting too intrusive on a first pass.

hhyyrylainen:
I started with a rebase and next I'll revert some changes from this PR until I think I can just merge this without further changes to act as a base for a bigger tutorial rework branch I'll work on next.

src/microbe_stage/gui/CompoundPanels.cs
hhyyrylainen, line 1: I added a temporary debugging checkbox to easily toggle the compound agents panel mode as well and I think I got the logic right now. The problem was not the animations, they were fine, rather it was a problem with the code not taking into account the difference between if agents are currently shown/hidden and whether something needs to be done to the agents visibility state.

hhyyrylainen: Actually, I realized again that that tutorial disable is the difficult case. I think I tracked down the problem in starting a new animation just a few frames after the first one. I added code so that the panels cannot start a new animation while another is in progress, instead it just queues the method to run again. This way the compounds show animation plays correctly first and then the agents animation plays after that. I think this is a fine solution rather than starting tracking of the *completed* animation state and based on that re-enqueueing animations.

PR #5941, opened 2025-02-24
Add a toggle for low resolution background blur

### LINKED ISSUE #5910: Add separate toggle for low quality background blur
As the blur can get expensive to render it'd be nice to have a toggle to force that to stay at 720p and not rescale with the game window

dligr:

Adds a toggle in the options menu that, when enabled, sets background blur viewports' resolution to the fixed value of 1280x720, improving performance.

**Related Issues**

Closes #5910

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/OptionsMenu.tscn
```diff
@@ -477,6 +478,11 @@ layout_mode = 2
 size_flags_horizontal = 0
 text = "DISPLAY_BACKGROUND_DISTORTION_EFFECT"
 
+[node name="LowQualityBlur" type="CheckBox" parent="CenterContainer/VBoxContainer/VBoxContainer/VBoxContainer/PanelContainer/Graphics/ScrollContainer/MarginContainer/VBoxContainer"]
+layout_mode = 2
+size_flags_horizontal = 0
+text = "LOW_QUALITY_BACKGROUND_BLUR"
```
hhyyrylainen, line 484: Maybe this could do with a tooltip? To explain turning this on improves performance by using a lower resolution blurring for the background (when blur is used).

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -166,9 +171,8 @@ protected override void Dispose(bool disposing)
         base.Dispose(disposing);
     }
 
-    private void UpdateSubViewportResolution()
+    private void UpdateSubViewportResolution(Vector2I newSize)
     {
-        var newSize = GetWindow().Size;
```
hhyyrylainen, line 171: I think it would be clearer to put the logic here to determine which size to use. That way accidentally calling UpdateSubViewportResolution from the wrong place in the code would still work correctly. And would only require `GetWindow().Size` to be in one place which I think is a relatively "complex" operation. 

Also I think in _Ready if low quality blur is on already it should use the fixed resolution, so this refactoring I think would fix a bug as well.

hhyyrylainen (changes requested):

hhyyrylainen (approved):
Everything looks good now.

PR #5954, opened 2025-02-24
Make fossilization buttons' positioning account for screen distortion

dligr:

Applies chromatic abberation (barrel distortion) to fossilization buttons' positions, placing them where microbes visually are.

This requires an inverse function. The true inverse function for barrel distortion would be way too complex ([something like this](https://www.wolframalpha.com/input?i=solve+for+a+and+b+%7B+c+%3D+a+*+%28a+*+a+%2B+b+*+b+%2B+%28a+*+a+%2B+b+*+b%29%5E2%29%3B+d+%3D+b+*+%28a+*+a+%2B+b+*+b+%2B+%28a+*+a+%2B+b+*+b%29%5E2%29+%7D)), so this PR uses an approximation, though its inaccuracy seems almost negligible. It breaks when the given position is outside of the screen, but it's fixable with clamping.

**Related Issues**

Fossilization buttons being too far from their microbes

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/MicrobeHUD.cs
+++ b/src/microbe_stage/MicrobeHUD.cs
@@ -267,6 +267,7 @@ public partial class MicrobeHUD : CreatureStageHUDBase<MicrobeStage>
 
             var button = FossilisationButtonScene.Instantiate<FossilisationButton>();
             button.AttachedEntity = entity;
+            button.IsMicrobeStage = true;
             button.Connect(FossilisationButton.SignalName.OnFossilisationDialogOpened, new Callable(this,
                 nameof(ShowFossilisationDialog)));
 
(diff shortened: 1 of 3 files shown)


src/microbe_stage/MicrobeHUD.cs
```diff
@@ -267,6 +267,7 @@ protected override void ShowFossilisationButtons()
 
             var button = FossilisationButtonScene.Instantiate<FossilisationButton>();
             button.AttachedEntity = entity;
+            button.IsMicrobeStage = true;
```
hhyyrylainen, line 270: Just me wondering: I guess as multicellular doesn't seem to have its own HUD so no change related to that is required.

src/thriveopedia/fossilisation/FossilisationButton.cs
hhyyrylainen, line 1: Is there a performance impact from running the inverse distortion for all buttons (all the time, I  assume, but I might misremember when the position update code is called)?

hhyyrylainen (approved):
I playtested and this seems to work well. And I didn't really notice anything related to the performance but I'll wait a bit in case you want to reply before I merge this.

src/thriveopedia/fossilisation/FossilisationButton.cs
dligr: There must be some kind of performance impact (the buttons are updated each frame), but it is probably unnoticable, considering that there isn't much else running at pause (as far as I know).

It's possible to make a small optimization here and not run the function on the buttons that are outside of the screen, though.

Issue #3563, opened 2022-07-19
Change patch map generator to not need to abandon regions


hhyyrylainen:
the generator should not need to do "Region abandoned" if the map is too big. Need to tweak the generation logic. Should be easy enough to find the place in the generator where the abandoning occurs by searching that string.

84634E1A607A:
Currently the generator tries to fit up to 9 patches into a map randomly, ensuring each patch has a distance of at least 190. In some edge cases, 7 or 8 patch already fills up the map, leaving no space for a next one.

The problem arose before my change, but in my pr I aligned patch position to 100 so I'm not sure if this problem persists.

Theoretical calculation (math) is needed to ensure it.

Patryk26g:
What would be the best solution for this? Would we allow to increase the map size?

hhyyrylainen:
I think the algorithm should just be updated so that it doesn't try to place too many regions that won't fit. This issue is mainly just to remove the error print that the patch generator sometimes causes even when things seemingly work fine.

PR #5960, opened 2025-02-25
Added in World Size parameter to World Editor

HyperbolicHadron:

Adds in a World Size parameter to the Planet Editor.

**Related Issues**

Part of a roadmap item (https://wiki.revolutionarygamesstudio.com/wiki/Release_Roadmap#0.8.x_(world_and_auto-evo_updates))

Reduces problem #3563 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/NewGameSettings.cs
+++ b/src/general/NewGameSettings.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
@@ -143,7 +143,10 @@ public partial class NewGameSettings : ControlWithInput
 
     [Export]
     public NodePath GameSeedAdvancedPath = null!;
-
+    
+    [Export]
+    public NodePath WorldSizeButtonPath = null!;
+    
     [Export]
     public NodePath IncludeMulticellularButtonPath = null!;
 
@@ -221,6 +224,7 @@ public partial class NewGameSettings : ControlWithInput
     private VBoxContainer dayLengthContainer = null!;
     private LineEdit gameSeed = null!;
     private LineEdit gameSeedAdvanced = null!;
+    private OptionButton worldSizeButton = null!;
 
     // Misc controls
     private Button includeMulticellularButton = null!;
@@ -316,6 +320,7 @@ public partial class NewGameSettings : ControlWithInput
         dayLengthReadout = GetNode<LineEdit>(DayLengthReadoutPath);
         gameSeed = GetNode<LineEdit>(GameSeedPath);
         gameSeedAdvanced = GetNode<LineEdit>(GameSeedAdvancedPath);
+        worldSizeButton = GetNode<OptionButton>(WorldSizeButtonPath);
         includeMulticellularButton = GetNode<Button>(IncludeMulticellularButtonPath);
         easterEggsButton = GetNode<Button>(EasterEggsButtonPath);
         backButton = GetNode<Button>(BackButtonPath);
@@ -462,6 +467,8 @@ public partial class NewGameSettings : ControlWithInput
         gameSeedAdvanced.Text = seedText;
         SetSeed(seedText);
 
+        worldSizeButton.Selected = (int)settings.WorldSize;
+
         // Always set prototypes to true as the player must have been there to descend
         includeMulticellularButton.ButtonPressed = true;
 
@@ -537,6 +544,7 @@ public partial class NewGameSettings : ControlWithInput
                 DayLengthReadoutPath.Dispose();
                 GameSeedPath.Dispose();
                 GameSeedAdvancedPath.Dispose();
+                WorldSizeButtonPath.Dispose();
                 IncludeMulticellularButtonPath.Dispose();
                 EasterEggsButtonPath.Dispose();
                 BackButtonPath.Dispose();
@@ -656,6 +664,7 @@ public partial class NewGameSettings : ControlWithInput
         settings.DayNightCycleEnabled = dayNightCycleButton.ButtonPressed;
         settings.DayLength = (int)dayLength.Value;
         settings.Seed = latestValidSeed;
+        settings.WorldSize = (WorldGenerationSettings.WorldSizeEnum)worldSizeButton.Selected;
 
         settings.IncludeMulticellular = includeMulticellularButton.ButtonPressed;
         settings.EasterEggs = easterEggsButton.ButtonPressed;
@@ -1040,6 +1049,11 @@ public partial class NewGameSettings : ControlWithInput
         gameSeed.Text = seed;
         gameSeedAdvanced.Text = seed;
         SetSeed(seed);
+    }
+    
+    private void OnWorldSizeSelected(int index)
+    {
+        worldSizeButton.Selected = index;
     }
 
     private void OnIncludeMulticellularToggled(bool pressed)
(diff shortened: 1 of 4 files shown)


HyperbolicHadron:
This should be basically done, just waiting on #5961 to be merged first. Or if I need to do something different with the localization files so that merging doesn't break anything.

src/general/WorldGenerationSettings.cs
```diff
@@ -74,6 +86,11 @@ public enum LifeOrigin
     /// </summary>
     public long Seed { get; set; } = new XoShiRo256starstar().Next64();
 
+    /// <summary>
+    ///   Size of World
+    /// </summary>
+    public WorldSizeEnum WorldSize { get; set; } = WorldSizeEnum.Medium;
```
hhyyrylainen, line 92: Is there a future where we want the world size to have much more flexible size controls? If yes, I think this would maybe make more sense as a float? Then the map generator can do the mapping from the float value to an actual "size".

HyperbolicHadron: As of right now, I do not know how flexible we will have the world size be. I am thinking that just having a few possible sizes will help with limiting the amount of fine tuning needed for the patch partitioning system. I don't think having a more smooth world size selection will be done in the foreseeable future. I do have plans for it, but they are at the very end of the priority list.

HyperbolicHadron: This will probably be similar to how we want some of the other world generation parameters such as land-ocean ratio. We want to limit the amount of combinations possible so that we don't have to test out lots of different combinations to see if the patch partitioning still works properly.

hhyyrylainen: I guess it is then fine to keep this as an enum, but you should add explicit enum values and a comment saying that it is not safe to insert values between, as it looks like you are using the C# enum values directly to access the GUI, which adds a risk that someone changes it and blows it up accidentally.

HyperbolicHadron: I will add a comment about it, though the enum values are already specified unless I am mistaken on what you are referring to.

hhyyrylainen:
The other PR is now merged.

HyperbolicHadron:
I changed the enum comment as requested, and I finished testing so everything seems good now. (#5961 did fix the issue)

src/general/NewGameSettings.cs
```diff
@@ -144,6 +144,9 @@ public partial class NewGameSettings : ControlWithInput
     [Export]
     public NodePath GameSeedAdvancedPath = null!;
 
+    [Export]
+    public NodePath WorldSizeButtonPath = null!;
```
hhyyrylainen, line 148: Could you use the new Godot 4 approach here where the `worldSizeButton` is directly given the `Export` attribute? As we are slowly trying to move away from this old NodePath approach.

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -19,7 +19,22 @@ public static PatchMap Generate(WorldGenerationSettings settings, Species defaul
 
         // Initialize the graph's random parameters
         var regionCoordinates = new List<Vector2>();
-        int vertexCount = random.Next(6, 10);
+
+        int vertexCount = 4;
+
+        if (settings.WorldSize == WorldGenerationSettings.WorldSizeEnum.Small)
```
hhyyrylainen, line 25: I think this should use a switch statement and print an error if the value is unhandled.

locale/en.po
```diff
@@ -8699,6 +8699,26 @@ msgstr ""
 msgid "WORLD_RELATIVE_MOVEMENT"
 msgstr "World relative"
 
+msgid "WORLD_SIZE"
+msgstr "World Size"
+
+msgid "WORLD_SIZE_EXPLANATION"
+msgstr ""
+"Changes the number of patches in the world.\n"
```
hhyyrylainen, line 8707: I think this text needs reworking for two reasons: all the other options use a format like `(explanation goes here)` and the second is that the text is two sentences long which is not used by the other explanations. So this should be rewritten as a single sentence and not use a period at the end.

Optionally I think the performance concern could be put in the tooltip, which is *usually* where more in-depth information should be put.

src/general/NewGameSettings.tscn
```diff
@@ -951,6 +952,41 @@ texture_disabled = ExtResource("7")
...
+
+[node name="Label2" type="Label" parent="CenterContainer/VBoxContainer/AdvancedOptions/Planet/VBoxContainer/HBoxContainer4/VBoxContainer"]
+custom_minimum_size = Vector2(200, 0)
+layout_mode = 2
+size_flags_horizontal = 3
+text = "WORLD_SIZE_EXPLANATION"
+label_settings = ExtResource("5_rnvau")
+autowrap_mode = 3
+
+[node name="WorldSize" type="OptionButton" parent="CenterContainer/VBoxContainer/AdvancedOptions/Planet/VBoxContainer/HBoxContainer4"]
```
hhyyrylainen, line 977: You need to modify the Planet container to have a scrollbar, because adding this makes the menu too tall in the GUI, so that causes a few pixels to be cut off the screen (and this will be worse with translations as many are way longer than the English text).

hhyyrylainen (changes requested):
Testing in-game the functionality seems good already (other than one GUI not fitting on screen). When reviewing the technical implementation I noticed quite a few tweak I'd like to see made to get the changes into a polished state.

src/general/NewGameSettings.tscn
HyperbolicHadron: Perhaps we could instead remove the "Planet generation coming soon!" picture, I think that could fit at least one more world parameter options before we add in a scrollbar, and the picture should soon be unnecessary?

hhyyrylainen: I think a scrollbar needs to be added eventually anyway, so might as well add it now, because otherwise someone will manage to break the layout with translations and then report a bug when it inevitably missed in testing.

HyperbolicHadron: I will add a scrollbar then, should I also remove the picture or wait until some other point for that?

hhyyrylainen: I'd keep the picture for now because a world size selection doesn't really fulfil the purpose of planet customization. If we had a couple more sliders then I think it could be removed.

HyperbolicHadron:
All the requested changes should be made now

hhyyrylainen:
I'll take a proper look on Monday, but I saw that you changed the default size to medium? I think for performance reasons we should keep the default map size in a new game the same it is currently, as otherwise people will be waiting forever for auto-evo on laptops.

HyperbolicHadron:
The medium map size creates 7-9 regions, which is closest to the previous settings bounds of 6-10. When the region connection highlighting change was added recently though, it caused more regions to be abandoned so it was ending up creating a number of regions on the lower end of the bounds. I can change it to the Small map as default, which is 4-6 regions, if we want the default size to be smaller for performance reasons. It would be different to what we had before the 0.8.1 update came out though. (I just tested with the 0.8.0 update, and I got 7 region maps most often).

hhyyrylainen:
Reading the code and your explanation it seems like medium should be the default as it is the closest to the world sizes previously generated.

src/general/NewGameSettings.cs
```diff
@@ -1042,6 +1048,11 @@ private void OnRandomisedGameSeedPressed()
         SetSeed(seed);
     }
 
+    private void OnWorldSizeSelected(int index)
+    {
+        worldSizeButton.Selected = index;
```
hhyyrylainen: Why does this option button select callback just apply the index back to the control this originates from?

hhyyrylainen: I've removed this the same way the other callbacks that don't actually need to do anything.

src/general/NewGameSettings.cs
```diff
@@ -1042,6 +1048,11 @@ private void OnRandomisedGameSeedPressed()
         SetSeed(seed);
     }
 
+    private void OnWorldSizeSelected(int index)
+    {
+        _ = index;
```
hhyyrylainen, line 1053: I fixed this callback assigning the index immediately back to itself.

hhyyrylainen (approved):
I tested this now in game and other than slightly lower than preferable map performance this seems good. I did do a few small technical fixes to get this to a mergeable state.

I've opened a follow up issue about the map performance: https://github.com/Revolutionary-Games/Thrive/issues/5975

PR #5961, opened 2025-02-25
Patch map lag fix

Patryk26g:

Fixes lag when selecting a node

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1328,7 +1328,10 @@ private void CheckNodeSelectionUpdate()
 
             // TODO: would be really nice to be able to just update the line objects without redoing them all
             // Clear existing connection lines
-            lineContainer.FreeChildren();
+            foreach (Node child in lineContainer.GetChildren())
```
hhyyrylainen, line 1331: We have a helper method that does this: `NodeHelpers.QueueFreeChildren` so that should be used as an extension method here instead of reimplementing that code. (potentially with `detach = false` if detaching is not required)

hhyyrylainen (changes requested):
This was a surprisingly simple PR. And if this reduces the lag then I'm all for this, though this should use an existing method we have for performing this operation.

src/microbe_stage/editor/PatchMapDrawer.cs
Patryk26g: For some reason when I used `NodeHelpers.QueueFreeChildren` the lag was back. I tested it multiple times and there really was a difference. I decided to change the body of this method from
`while (true)
        {
            int count = node.GetChildCount();

            if (count < 1)
                break;

            var child = node.GetChild(count - 1);

            if (detach)
                node.RemoveChild(child);

            child.QueueFree();
        }`

to

`foreach (Node child in node.GetChildren())
        {
            if (detach)
                node.RemoveChild(child);

            child.QueueFree();
        }`

and it really inreased its spead even though it seems like it's doing the same thing

hhyyrylainen: Did you set detach to false? That would mean the other variant does more stuff if you didn't.

Could you try rewriting the body so that the GetChildCount method is used just once? And also trying if removing the first child or last child is faster? If none of those help I suppose the magic is in asking the engine just once to return the child list and allocate memory for it is faster than retrieving individual children. Then we need to analyze the first approach to see if it actually allocates memory still or not, if it also allocates memory we probably maybe should just switch to the GetChildren method and warn in the name / document the helper method that it allocates a temporary list.

hhyyrylainen (approved):
Code change looks good to me now.

PR #5967, opened 2025-02-27
Make high fire-rate toxins consume less per shot

### LINKED ISSUE #5366: Investigate if high fire-rate toxins should consume less per shot
as it's a pretty good point that cells cannot fire that fast even with maximum firerate as each shot can consume a ton of toxin at once

I'm putting this in the 0.7.0 but this might not get done by then so this'll be one of those issues that hang around in the release milestones for a really long time

dligr:

Makes toxins consume less toxin per shot when they are modified to have a higher fire-rate. This should allow fast-firing microbes to fire more shots making the lower cooldown more meaningful.

**Related Issues**

Closes #5366

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/MicrobeEmissionSystem.cs
+++ b/src/microbe_stage/systems/MicrobeEmissionSystem.cs
@@ -228,7 +228,9 @@ public sealed class MicrobeEmissionSystem : AEntitySetSystem<float>
                 // aren't used per agent type)
 
                 // Emit as much as you have, but don't start if there's way too little toxin
-                float amountEmitted = Math.Min(amountAvailable, Constants.MAXIMUM_AGENT_EMISSION_AMOUNT);
+                float amountEmitted = Math.Min(amountAvailable,
+                    EmissionAmountWithToxicity(organelles.AverageToxinToxicity));
+
                 if (amountEmitted < Constants.MINIMUM_AGENT_EMISSION_AMOUNT)
                     return;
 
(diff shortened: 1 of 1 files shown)


src/microbe_stage/systems/MicrobeEmissionSystem.cs
```diff
@@ -285,6 +287,23 @@ private float ToxinCooldownWithToxicity(int vacuoleCount, float toxicity)
         return Constants.AGENT_EMISSION_COOLDOWN / vacuoleCount;
     }
 
+    private float EmissionAmountWithToxicity(float toxicity)
+    {
+        if (toxicity < 0)
+        {
+            // High-firerate, low emission amount
+            return Constants.MAXIMUM_AGENT_EMISSION_AMOUNT * (1.0f + toxicity * 0.5f);
```
hhyyrylainen, line 295: Doesn't this also increase the amount? Shouldn't there be like 1 **-** toxicity?

hhyyrylainen: Oh yeah, I missed that part, it is negative here so subtracting it wouldn't be the right thing to do.

I think this is fine. So this makes the technical side all fine. Let's wait a bit and see if we get any game designer or tester to comment on the balance change of this.

hhyyrylainen (approved):
Seems like no one objected to the balance change so I'll merge this now.

src/microbe_stage/systems/MicrobeEmissionSystem.cs
```diff
@@ -285,6 +287,23 @@ private float ToxinCooldownWithToxicity(int vacuoleCount, float toxicity)
         return Constants.AGENT_EMISSION_COOLDOWN / vacuoleCount;
     }
 
+    private float EmissionAmountWithToxicity(float toxicity)
```
hhyyrylainen, line 290: Though now that I merged this, I do have one potential concern that this affects the effectiveness of toxins in an unexpected way.

Because after calculating the amount it is further adjusted by this piece of code:

```c#
                var damagingToxinAmount = amountEmitted *
                    ToxinAmountMultiplierFromToxicity(organelles.AverageToxinToxicity, selectedToxinType);
```

So could you check that, just to make sure things are fine related to that?

dligr: Since the `damagingToxinAmount` only affects the projectile's damage, this code line just calculates the damage based on the toxin amount multiplied by a toxicity modifier. 

The amount of toxin taken isn't modified by this calculation as it depends only on the `amountEmitted`:
```C#
compounds.TakeCompound(agentType, amountEmitted);
```

So, the damage per unit of toxin should stay the same with this PR's changes; the damage per shot is different however.

hhyyrylainen: > So, the damage per unit of toxin should stay the same with this PR's changes; the damage per shot is different however.

I think that was exactly my concern. If this makes individual shots weaker and that was not realized / not tested enough, we might get reports after the next release that toxins feel weaker than they were before. So I want this to be addressed now while we still remember what is being changed (as I'll probably have totally forgotten this conversation in 60 days).

PR #5972, opened 2025-03-02
Added a tutorial about placing the nucleus

### LINKED ISSUE #5428: Add a tutorial for getting the binding agents for advancing in the game
It should probably trigger like around editor cycle 11-12 if the player doesn't have a nucleus already to hint the player that what they need to place to advance in the game.

La-ryssa:

Adds a tutorial after 10 editor cycles, reminding the player to place a nucleus to continue progressing, so they can unlock binding agents and become multicellular. If the player has placed a nucleus the tutorial will not trigger.

**Related Issues**

partially does #5428

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/tutorial/microbe_editor/MicrobeEditorTutorialGUI.tscn
```diff
@@ -554,6 +555,55 @@ Description = "EDITOR_TUTORIAL_EARLY_GOAL"
...
+layout_mode = 2
+size_flags_vertical = 3
+
+[node name="Button" type="Button" parent="NucleusTutorial/ScrollContainer/VBoxContainer"]
+custom_minimum_size = Vector2(110, 0)
+layout_mode = 2
+size_flags_horizontal = 4
+theme_override_fonts/font = ExtResource("9_jkhed")
+theme_override_font_sizes/font_size = 18
+text = "GOT_IT"
```
hhyyrylainen, line 605: I think this tutorial should close only when the nucleus is placed or with the explicit X, so I don't think this tutorial needs an explicit close button.

src/tutorial/TutorialEventArgs.cs
```diff
@@ -154,6 +154,26 @@ public OrganellePlacedEventArgs(OrganelleDefinition definition)
     public OrganelleDefinition Definition { get; }
 }
 
+public class UndoEventArgs : TutorialEventArgs
```
hhyyrylainen, line 157: Why not have just a single class called "EditorActionArgs"? (assuming the action type is also refactored to be the editor action which, if I remember right adds extra data on top of the ReversibleAction class which may be useful in the tutorial).

src/general/ActionHistory.cs
```diff
@@ -15,10 +15,10 @@ public class ActionHistory<T>
     ///   possibly redone), is 0 if there is none.
     /// </summary>
     [JsonProperty]
-    protected int ActionIndex { get; private set; }
+    public List<T> Actions { get; private set; } = new();
 
     [JsonProperty]
-    protected List<T> Actions { get; private set; } = new();
+    public int ActionIndex { get; private set; }
```
hhyyrylainen, line 21: These should not be made public.

Instead there should be a helper method to inspect the latest action. That way any code in the project cannot modify the action history however they like and end up with a massive mess.

src/tutorial/TutorialState.cs
```diff
@@ -389,6 +392,7 @@ private List<TutorialPhase> BuildListOfAllTutorials()
     {
         return new List<TutorialPhase>
         {
+            NucleusTutorial,
```
hhyyrylainen, line 395: I'd put this after the first editor tutorials.

src/tutorial/microbe_editor/NucleusTutorial.cs
```diff
@@ -0,0 +1,114 @@
+﻿namespace Tutorial;
+
+using System;
+using Newtonsoft.Json;
+
+/// <summary>
+///   Prompts the player to eventually place a nucleus for further progression
+/// </summary>
+public class NucleusTutorial : TutorialPhase
```
hhyyrylainen, line 9: Why does this not inherit from the editor entry counting tutorial? I thought that this would benefit from the existing functionality.

src/tutorial/microbe_editor/NucleusTutorial.cs
```diff
@@ -0,0 +1,114 @@
...
+{
+    private const int TriggersOnNthEditorSession = 11;
+
+    private readonly string cellEditorTab = EditorTab.CellEditor.ToString();
+
+    private bool hasNucleus;
```
hhyyrylainen, line 15: This has to be JsonProperty otherwise the game will forget if the player has a nucleus or not when they save/load.

src/tutorial/microbe_editor/NucleusTutorial.cs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public override bool CheckEvent(TutorialState overallState, TutorialEventType eventType, EventArgs args,
+        object sender)
+    {
+        switch (eventType)
+        {
+            case TutorialEventType.MicrobeEditorOrganellePlaced:
+            {
+                var eventArgs = (OrganellePlacedEventArgs)args;
```
hhyyrylainen, line 39: This cast doesn't look safe to me so there should be a check to make sure it is actually a placement action.

src/tutorial/microbe_editor/NucleusTutorial.cs
```diff
@@ -0,0 +1,114 @@
...
+                var isNucleus = eventArgs.Definition.Name == "Nucleus";
```
hhyyrylainen, line 40: This compares the *translated* name, which is going to break immediately for someone using any other language than English. Instead the right way is to compare the *internal* name. And for extra safety the nucleus shouldn't be hardcoded. Instead there should be a readonly field to fetch the nucleus definition from SimulationParameters.

src/tutorial/microbe_editor/NucleusTutorial.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            case TutorialEventType.MicrobeEditorUndo:
+            {
+                var eventArgs = (UndoEventArgs)args;
+                var combinedAction = (CombinedEditorAction)eventArgs.Action;
+
+                foreach (var data in combinedAction.Data)
+                {
+                    if (data is OrganellePlacementActionData { PlacedHex.ReadableName: "Nucleus" })
```
hhyyrylainen, line 62: Same here.

src/tutorial/microbe_editor/NucleusTutorial.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            case TutorialEventType.MicrobeEditorRedo:
+            {
+                var eventArgs = (RedoEventArgs)args;
+                var combinedAction = (CombinedEditorAction)eventArgs.Action;
+
+                foreach (var data in combinedAction.Data)
+                {
+                    if (data is OrganellePlacementActionData { PlacedHex.ReadableName: "Nucleus" })
```
hhyyrylainen, line 78: And here.

src/tutorial/microbe_editor/NucleusTutorial.cs
```diff
@@ -0,0 +1,114 @@
...
+                    Show();
+                }
+
+                if ((ShownCurrently && ((StringEventArgs)args).Data != cellEditorTab) || hasNucleus)
+                {
+                    Hide();
+                }
+
+                break;
+            }
```
hhyyrylainen, line 109: I see this is slightly different than the base class, but I think an approach where using the base class is done, but then this one specific event is not passed to the `base` call and handled separately to override the behaviour.

hhyyrylainen (changes requested):
The nucleus tutorial you added looks to be already on the right track. However I noticed some slight technical problems that should be resolved to improve the code quality and fix some bugs. 

Also I noticed you didn't add a binding agent tutorial, which is fine, but then this PR doesn't fully close the related issue as it *also* calls for an explicit binding agents tutorial.

La-ryssa:
> Also I noticed you didn't add a binding agent tutorial, which is fine, but then this PR doesn't fully close the related issue as it _also_ calls for an explicit binding agents tutorial.

I'll be taking the approach of keeping this as a nucleus only tutorial and not close the issue and leave the binding for a separate PR so I or someone else can do it at another time

src/tutorial/microbe_editor/MicrobeEditorTutorialGUI.tscn
La-ryssa: What I had in mind was that this tutorial would be just a reminder that the player needs the nucleus to continue progressing, but wouldn't force them to do so in case they aren't ready yet

hhyyrylainen: That's fair, which is why I would leave the X button enabled on the tutorial window, but I think that adding the extra button to close the tutorial is not really necessary in this case, I think.

hhyyrylainen:
>  I'll be taking the approach of keeping this as a nucleus only tutorial and not close the issue and leave the binding for a separate PR so I or someone else can do it at another time

That's fine. I renamed the PR and removed the closes issue text to compensate for that. It looks like not all of my earlier comments were addressed yet so I'll wait a bit until doing another full review.

La-ryssa:
> That's fine. I renamed the PR and removed the closes issue text to compensate for that. It looks like not all of my earlier comments were addressed yet so I'll wait a bit until doing another full review.

I haven't addressed any of them yet. I just got on so I remarked it as a draft

src/tutorial/microbe_editor/NucleusTutorial.cs
La-ryssa: One of the recent tutorials, `EarlyGameGoalTutorial` hadn't, and I followed suit. I'm going to implement it, but if it isn't being worked on yet, I think the other tutorials that use counting should be also retroactively changed to inherit from it

src/tutorial/microbe_editor/NucleusTutorial.cs
La-ryssa: I followed the how the other tutorial scripts did it so they should also be changed if they aren't already

src/tutorial/microbe_editor/NucleusTutorial.cs
hhyyrylainen: The early game tutorial is special because it doesn't wait until the cell editor tab. I *think* that in this case it makes more sense to use the base class as the nucleus tutorial is one that waits n number of editor cycles and then triggers in the cell editor tab (this second point is what distinguishes the early game tutorial in my opinion and why I didn't want to use the base class as I'd need to avoid like 50% of the functionality in it).

hhyyrylainen: Though now that I think about it, it could make sense to refactor one more level of inheritance into the stuff to allow the early goal to also inherit something. But I don't think there's going to be another tutorial like the early game goal so that refactor might be irrelevant forever...

src/tutorial/microbe_editor/NucleusTutorial.cs
hhyyrylainen: You'll need to do that in this PR because you refactored where the events are triggered. The previous code was safe only as long as the events can *only* be triggered by the microbe editor, but I think you changed things in this PR so that other editor types will now also automatically send the edit events, which is good for the future but requires some extra safety checks now.

La-ryssa: > but I think you changed things in this PR so that other editor types will now also automatically send the edit events

Could you link to this change? Afaik I only made it so undo and redo events are triggered in a different script

src/tutorial/TutorialEventArgs.cs
La-ryssa: > assuming the action type is also refactored to be the editor action

I'm not sure what you mean by editor action and action type. Do you mean event type?

src/tutorial/microbe_editor/NucleusTutorial.cs
hhyyrylainen: That's exactly the problem. You moved the tutorial event triggering to the base editor class, which means that multicellular and macroscopic editors now also trigger the tutorial event, and those use vastly different editor actions. I didn't test in game but I would expect this to cause the tutorials that were written to only work with the microbe editor to explode due to this change.

src/tutorial/TutorialEventArgs.cs
hhyyrylainen: Undo and redo under the hood *contain* all the action types. So it is already the case that both Undo and redo event args can both contain organelle placement, metaball placement, membrane type change etc.

So I think it would make sense to have a single class for both undo and redo as those *already have the same data* no need in my opinion to differentiate in the data type layer those two actions as they contain the same data. The event type they are sent as can be used to differentiate in the tutorials if undo or redo is happening.

src/tutorial/microbe_editor/NucleusTutorial.cs
La-ryssa: How does this affect other editor tutorials such as organelle placement? It should only interfere when you undo or redo changes, yes?

hhyyrylainen: Under the hood the way any action in the editor is triggered is that the action is created and then redo is called on it, which means that any actions happening (well at least the ones that are hooked up into the action history system) in the editor go always through either the redo or undo call.

Though I may have slightly misread your code and now misremembering where you hooked up into the system.

La-ryssa: Would maybe change it so the tutorials that use editor actions check in which stage the player is in, fix this?

hhyyrylainen: That seems way more complicated than the tutorials just checking the type of data is expected and only react to the expected data. The tutorials *already* have to cast the data to the type they want to process so they might as well use a cast that can return null when it didn't actually succeed.

La-ryssa: Like so https://github.com/Revolutionary-Games/Thrive/pull/5972/commits/e4ec988518c1f5a50fcdb8f2633fd6d094233087?

I made it so if the args is not the type, it breaks the switch. I prefer this over nested if statements for readability, but `FlagellumPlacementTutorial` uses nested if statements instead for its cast checking:

```C#
            case TutorialEventType.MicrobeEditorOrganellePlaced:
            {
                if (args is OrganellePlacedEventArgs organellePlacedArgs)
                {
                    if (organellePlacedArgs.Definition.InternalName != flagellum.InternalName)
                        break;

                    if (!HasBeenShown && CanTrigger && !overallState.TutorialActive())
                    {
                        Show();
                        return true;
                    }
                }

                break;
            }
```

So if that's preferable please tell me and I'll change it.

hhyyrylainen: I think both approaches are fine, if I read correctly that the difference is just the breaking early or using one more level of indentation. I think either is fine (but my personal preference is to use the style in flagellum tutorial as I think it makes it slightly easier what the scope of `organellePlacedArgs` is).

Though I'll mention one slight thing that could be improved in your version is naming the variable more clearly. You've named the casted variable as `eventArgs` in the second block of code, which doesn't really imply that it is a different type from `args`. So I would rename that one variable similarly to how you used `organellePlacedEventArgs` in the upper diff block, which I think makes the code much more easy to read as the name of the new casted variable includes the fact that it is now a more specific type.

But in terms of style I think both approaches are fine and you can use the one you prefer.

locale/en.po
```diff
@@ -1772,6 +1772,14 @@ msgstr ""
 "\n"
 "When you are ready, press the \"next\" button in the bottom right to continue to the Patch Map Screen."
 
+msgid "EDITOR_TUTORIAL_MICROBE_EDITOR_NUCLEUS"
+msgstr ""
+"To continue progressing and unlocking new organelles, you will want to eventually place a [b]nucleus[/b].\n"
+"\n"
+"Adding a [b]nucleus[/b] will unlock more advanced organelles, such as a [b]binding agent[/b], which will allow you to bind with other cells of your species and become [b]multicellular[/b].\n"
```
hhyyrylainen, line 1779:
```suggestion
"Adding a [b]nucleus[/b] will unlock more advanced organelles, such as the [b]binding agent[/b], which will allow you to bind with other cells of your species and become [u]multicellular[/u].\n"
```

src/tutorial/microbe_editor/MicrobeEditorTutorialGUI.tscn
```diff
@@ -554,6 +555,47 @@ Description = "EDITOR_TUTORIAL_EARLY_GOAL"
 ShowDelay = 0.7
 PreventsMouseCaptureWhileOpen = true
 
+[node name="NucleusTutorial" parent="." instance=ExtResource("9")]
```
hhyyrylainen, line 558: As this doesn't have the button anymore, couldn't this use the default tutorial scene so this doesn't need a manual setup of the child nodes in here? Like the "editor tutorial early goal" tutorial.

src/tutorial/microbe_editor/ModifyOrganelleTutorial.cs
```diff
@@ -21,9 +21,13 @@ public override bool CheckEvent(TutorialState overallState, TutorialEventType ev
         {
             case TutorialEventType.MicrobeEditorOrganellePlaced:
             {
-                var eventArgs = (OrganellePlacedEventArgs)args;
-                var upgradable = eventArgs.Definition.AvailableUpgrades.Count > 0 ||
-                    !string.IsNullOrEmpty(eventArgs.Definition.UpgradeGUI);
+                if (args is not OrganellePlacedEventArgs organellePlacedEventArgs)
```
hhyyrylainen, line 24: I don't think this change is necessary as the organelle placed event is not changed in this PR. So I feel like it'd be better if this causes an error if some new code change is done that causes organelle placed events that are invalid.

src/tutorial/microbe_editor/NucleusTutorial.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    public override void ApplyGUIState(MicrobeEditorTutorialGUI gui)
+    {
+        gui.NucleusTutorialVisible = ShownCurrently;
+    }
+
+    public override bool CheckEvent(TutorialState overallState, TutorialEventType eventType, EventArgs args,
+        object sender)
+    {
+        if (!hasNucleus)
```
hhyyrylainen, line 35: I don't fully understand the logic why the base event check is only done if there is no nucleus. So could you either explain this with code comments or refactor this to always pass the base event but then override the result in this method for the situations where the base class did something incorrect?

src/tutorial/microbe_editor/NucleusTutorial.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            if (base.CheckEvent(overallState, eventType, args, sender))
+                return true;
+        }
+
+        switch (eventType)
+        {
+            case TutorialEventType.MicrobeEditorOrganellePlaced:
+            {
+                if (args is not OrganellePlacedEventArgs eventArgs)
```
hhyyrylainen, line 45: Like the chemoreceptor tutorial I don't think this should do a safe cast, better to fail loudly if someone messes up MicrobeEditor**OrganellePlaced** event.

hhyyrylainen (changes requested):
I have a few more small technical comments on this about the code clarity and error handling.

hhyyrylainen:
One more thought I just had: I think the nucleus tutorial should remind the player about the undo function if it turns out they are not ready to evolve the nucleus yet.

src/tutorial/microbe_editor/NucleusTutorial.cs
La-ryssa: I call the base class event check because it already does all the checks I need, and then displays the tutorial. But I don't want it to display if the player has a nucleus. Just checking if that makes sense first before I add comments

hhyyrylainen: I suppose that does make sense yes, but without a comment it's impossible to see from this bit of the code what it does.

src/tutorial/microbe_editor/ModifyOrganelleTutorial.cs
La-ryssa: Is that not what you meant with https://github.com/Revolutionary-Games/Thrive/pull/5972#discussion_r1980910160 and https://github.com/Revolutionary-Games/Thrive/pull/5972#discussion_r1977204900? Apologies if I misunderstood it

hhyyrylainen: No, sorry that's unrelated. The event for organelle placing should have no way to trigger in the base editor class. So that still is purely only in the microbe editor so code using that event does not need to be changed.

hhyyrylainen (approved):
I did some refactoring on the tutorial inheritance tree and various other fixes and improvements on this. I also tested this tutorial. I did that to get this into mergeable shape to reduce the number of open PRs and thus my mental load.

PR #5973, opened 2025-03-02
Add optimal value markers to tolerance sliders

dligr:

Adds markers which indicate optimal tolerance values of the patch. Also adds min/max value labels at the ends of sliders.

This is supposed to improve the clarity of the tolerance panel and make it easier to find good tolerance values.

image

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I'll wait checking this until I'm done with the feature of organelles affecting the tolerances as that needs quite a lot of GUI changes (which sadly will then mean there will be merge conflicts here).

hhyyrylainen:
This can now continue with this being merged: https://github.com/Revolutionary-Games/Thrive/pull/5969

dligr:
Now that the merge conflict is fixed, this PR is ready for review

src/microbe_stage/editor/ToleranceInfo.cs
```diff
@@ -0,0 +1,56 @@
+﻿using Godot;
+
+/// <summary>
+///   Displays additional tolerance slider info, such as the optimal value and value labels.
+/// </summary>
+public partial class ToleranceInfo : Control
```
hhyyrylainen, line 6: I think this class is a bit misnamed because this isn't anything to do with calculated tolerance info, or at least not directly the calculated data. I'd rather name this something like SliderOptimalMarkers or something along those lines.

src/microbe_stage/editor/ToleranceInfo.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            optimalValueMarker.Visible = value;
+        }
+    }
+
+    public override void _Ready()
+    {
+        base._Ready();
+
+        optimalValueMarker.Texture = markerTexture;
```
hhyyrylainen, line 42: If this cannot be changed later, then why isn't this directly just setup in the editor to be in the scene directly? Seems a bit weird to have to setup a separate export variable just to set a texture that from what I've seen of the code here so far could very well ahve been done directly in Godot.

src/microbe_stage/editor/ToleranceInfo.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    public void UpdateBoundaryLabels(string start, string end)
+    {
+        startValue.Text = start;
+        endValue.Text = end;
+    }
+
+    public void UpdateMarker(float optimalValue)
+    {
+        optimalValueMarker.AnchorLeft = optimalValue;
```
hhyyrylainen, line 53: Anchors shouldn't be used for dynamic position, just directly set the positions instead for manual positioning of nodes as that's how it is done everywhere else in the game.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -99,6 +99,21 @@ public partial class TolerancesEditorSubComponent : EditorComponentBase<ICellEdi
     [Export]
     private Label uvResistanceModifierLabel = null!;
 
+    [Export]
+    private ToleranceInfo temperatureToleranceInfo = null!;
```
hhyyrylainen, line 103: These need better naming along with the ToleranceInfo class.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -188,6 +203,7 @@ public override void OnEditorSpeciesSetup(Species species)
         }
 
         UpdateToolTipStats();
+        UpdateAdditionalSliderInfo();
```
hhyyrylainen, line 206: Like here "AdditionalSliderInfo" is much more descriptive than "ToleranceInfo" in regards to what it actually is.

src/microbe_stage/editor/ToleranceInfo.cs
hhyyrylainen: Okay, I read the main scene file. Maybe this could instead be named "overrideMarkerTexture"? And only apply it if it isn't null?

Also as that's meant to be changed when this scene is placed, I think you should sort out the imports like this:

```c#
    [Export]
    [ExportCategory("Visuals")]
    private Texture2D markerTexture = null!;

    [Export]
    [ExportCategory("Internal")]
    private Label startValue = null!;

    [Export]
    private Label endValue = null!;

    [Export]
    private TextureRect optimalValueMarker = null!;
```

To make it clear what scenes that place this scene are supposed to edit and what not.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -905,6 +923,44 @@ private void UpdateCurrentValueDisplays()
...
+    private void UpdateAdditionalSliderInfo()
+    {
+        var patch = Editor.CurrentPatch;
+        var patchTemperature = patch.Biome.GetCompound(Compound.Temperature, CompoundAmountType.Biome).Ambient;
+        var patchPressure = patch.Biome.Pressure;
+        var requiredOxygenResistance = patch.Biome.CalculateOxygenResistanceFactor();
+        var requiredUVResistance = patch.Biome.CalculateUVFactor();
+
+        var unitFormat = Localization.Translate("VALUE_WITH_UNIT");
+        var percentageFormat = Localization.Translate("PERCENTAGE_VALUE");
```
hhyyrylainen, line 935: Could any of this code be combined with existing code in this class? I ask because this seems like duplicated code.

src/microbe_stage/editor/ToleranceInfo.tscn
```diff
@@ -0,0 +1,65 @@
+[gd_scene load_steps=3 format=3 uid="uid://dccswprlmgtel"]
```
hhyyrylainen, line 1: I think everything should have mouse ignore / pass here because right now this adds a "break" area between the value slider and the current value readout where the relevant stat tooltip is not shown due to this messing with the mouse passing.

hhyyrylainen (changes requested):
Here's a few technical comments. I posted on the gui discord channel about general feedback on the look of this (and asked if other team members also want to comment).

hhyyrylainen:
For some reason I can't get this PR to run, I get this error:

image

Is that expected or has something gone wrong for me locally (git doesn't show any scene changes for me locally).

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
dligr: I merged the code into `UpdateCurrentValueDisplays()`. It should be a bit worse for performance, because it's called more often than `UpdateSliderAdditionalInfo()`, but I suppose it's acceptable for cleaner code.

src/microbe_stage/editor/ToleranceInfo.cs
dligr: I have remade this function to update positions rather than markers, but there's a problem now: on the first editor entry, tolerance markers are significantly offset (which fixes itself when changing a patch, for example). The cause of this is that prior to entering the tolerance tab, its elements have different sizes than after doing so.

dligr:
Regarding the exception, I have apparently forgotten to set tolerance marker exports after updating their names, making them null, but it should be fixed now.

src/microbe_stage/editor/ToleranceInfo.cs
hhyyrylainen: Maybe connect to the size changed signal and reposition the markers?

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -903,6 +920,33 @@ private void UpdateCurrentValueDisplays()
...
             uvResistanceModifierLabel.Visible = false;
         }
+
+        // Update markers
+        temperatureToleranceMarker.UpdateBoundaryLabels(unitFormat.FormatSafe(0, temperature.Unit),
+            unitFormat.FormatSafe(100, temperature.Unit));
+        temperatureToleranceMarker.UpdateMarker(patchTemperature / 100.0f);
+
+        string minPressure = unitFormat.FormatSafe(0, "kPa");
+        string maxPressure = unitFormat.FormatSafe(70000, "kPa");
```
hhyyrylainen, line 930: If these value labels are kept, these should actually read the real values from the sliders because otherwise these numbers might not actually match what is configured in the GUI and at that point these would be totally useless.

hhyyrylainen (commented):
This now functionally works.

However, I still don't personally like having the numbers above the sliders for min / max value as they make the GUI not fit and instead add an unnecessary scrollbar in my opinion.

Also I think that the optimal values should be parented to the sliders themselves so that they aren't above them but can actually touch the slider bar, and that would obviously also take up less vertical space so that my dream of not having a scrollbar could be fulfilled.

Though that's just my opinion, opinions from other team members could be sought out about the GUI look.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
dligr: I have removed the labels, so this shouldn't be necessary now

src/microbe_stage/editor/ToleranceOptimalMarker.cs
```diff
@@ -0,0 +1,61 @@
+﻿using Godot;
+
+/// <summary>
+///   Displays additional tolerance slider info, such as the optimal value and value labels.
```
hhyyrylainen, line 4: I think this comment is now outdated.

src/microbe_stage/editor/ToleranceOptimalMarker.cs
```diff
@@ -0,0 +1,61 @@
...
+            if (value == showMarker)
+                return;
+
+            showMarker = value;
+
+            optimalValueMarker.Visible = value;
+        }
+    }
+
+    public float OptimalValue
```
hhyyrylainen, line 34: So this seems like the horizontal position in pixel units? I think the name should be changed as this doesn't seem to relate to the slider values?

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -903,6 +918,20 @@ private void UpdateCurrentValueDisplays()
         {
             uvResistanceModifierLabel.Visible = false;
         }
+
+        // Update markers
+        temperatureToleranceMarker.OptimalValue = patchTemperature / 100.0f;
```
hhyyrylainen, line 923: Like continuing from my previous comment, I have 0 clue why this math works at all.

Though I guess maybe these are the fraction of how far along the control (in range 0-1) this marker should be? I guess that's understandable, but again all of the divisors here should read the actual slider max values because when those are changed (and they surely will be, at least the pressure) these markes will get totally desynced from reality.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -903,6 +918,20 @@ private void UpdateCurrentValueDisplays()
...
+
+        minPressureToleranceMarker.OptimalValue = patchPressure / 70000000;
+        maxPressureToleranceMarker.OptimalValue = patchPressure / 70000000;
```
hhyyrylainen, line 926: I now like the approach to the markers. But I still don't like that the pressure has both the marker on the min and max sliders. Isn't that more confusing than if the marker was only on the upper pressure slider? I could very well be wrong in which is the more understandable option. What's your reasoning for adding the optimal marker for both pressure sliders?

hhyyrylainen (changes requested):
I like the implementation now (though maybe one slight tweak could be made). Other than that I saw just a fewsmall  technical problems like hardcoding slider max values and unclear naming.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
dligr: Yes, the optimal value is the fraction of where the value is between the min and max values. I added a comment clarifying what the variable is.

I have also updated the code block to use slider's min/max values and also account for the case when min value isn't 0.

hhyyrylainen (approved):
The code looks mostly understandable to me now. I still kind of disagree with the pressure slider markers, but I guess it is easy enough to disable later if in further testing other people also find it confusing it so for now it can stay enabled.

PR #5974, opened 2025-03-02
Global glaciation event

Patryk26g:

Adds global glaciation event. Still needs the new icon and translation files

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/world_effects/WorldEffectVisuals.cs
```diff
@@ -4,6 +4,6 @@
     ///   Used as a placeholder value instead of null
     /// </summary>
     None,
-
-    UnderwaterVentEruption,
+    UnderwaterVentEruptionEvent,
+    GlaciationEvent,
```
hhyyrylainen, line 8: I don't think the "Event" suffix is required here...

simulation_parameters/Constants.cs
```diff
@@ -1890,6 +1894,16 @@ public static class Constants
         [CELLULASE_ENZYME] = 4.5f,
     };
 
+    public static readonly HashSet<BiomeType> SurfaceBiomes = new()
```
hhyyrylainen, line 1897: Instead of hardcoding surface patches should be detected from their depth values.

src/microbe_stage/BiomeConditions.cs
```diff
@@ -199,7 +199,7 @@ public bool TryGetCompound(Compound compound, CompoundAmountType amountType, out
     ///   Specifies the compound cloud spawn sizes for all new non-environmental compounds
     /// </param>
     public void ApplyLongTermCompoundChanges(Biome biomeDetails, Dictionary<Compound, float> changes,
-        IReadOnlyDictionary<Compound, float> newCloudSizes)
+        IReadOnlyDictionary<Compound, float>? newCloudSizes)
```
hhyyrylainen, line 202: I think this shouldn't be made nullable because it signifies a major problem if an existing cloud size is not known. So I wouldn't want to make it easier to make bugs happen by making this a nullable parameter.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,218 @@
...
+    private const string Background = "iceshelf";
+
+    private static readonly string[] IceChunksConfigurations =
+        ["iceShard", "iceChunkSmall", "iceChunkBig", "iceSnowflake"];
+
+    [JsonProperty]
+    private readonly XoShiRo256starstar random;
+
+    [JsonProperty]
+    private readonly bool hasEventAlreadyHappened;
```
hhyyrylainen, line 20: How does this work if this is a readonly field?

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    [JsonConstructor]
+    public GlobalGlaciationEvent(GameWorld targetWorld, XoShiRo256starstar random)
+    {
+        this.targetWorld = targetWorld;
+        this.random = random;
+    }
+
+    public void OnRegisterToWorld() { }
```
hhyyrylainen, line 54: I'm not sure if our styleguide mentions it explicitly but Thrive code shouldn't have any methods on a single line, so this should be split out properly:

```
public void OnRegisterToWorld()
{
}
```

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,218 @@
...
+
+        if (generationsLeft > 0)
+            MarkPatches(totalTimePassed);
+
+        if (hasEventAlreadyHappened)
+            return;
+
+        if (generationsLeft == -1)
+            TryToTriggerEvent(totalTimePassed);
+        else if (generationsLeft == 0)
```
hhyyrylainen, line 69: Multiline if statements must use braces according to our styleguide.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,218 @@
...
+            }
+        }
+    }
+
+    private void TryToTriggerEvent(double totalTimePassed)
+    {
+        if (!AreConditionsMet())
+            return;
+
+        generationsLeft = random.Next(2, 6);
```
hhyyrylainen, line 89: I think the duration range would be interesting to allow tweaking through `Constants`

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,218 @@
...
+    private bool AreConditionsMet()
+    {
+        var numberOfSurfacePatches = 0;
+        var patchesExceedingOxygenLevel = 0;
+        foreach (var patch in targetWorld.Map.Patches.Values)
+        {
+            if (!Constants.SurfaceBiomes.Contains(patch.BiomeType))
+                continue;
+
+            var oxygenLevel = patch.Biome.ChangeableCompounds[Compound.Oxygen];
```
hhyyrylainen, line 109: If this doesn't want to change the compound amounts, this should use the proper interface for reading the compounds (`BiomeConditions.TryGetCompound`).

simulation_parameters/Constants.cs
```diff
@@ -1314,6 +1314,10 @@ public static class Constants
     public const float VENT_ERUPTION_HYDROGEN_SULFIDE_INCREASE = 0.00004f;
     public const float VENT_ERUPTION_CARBON_DIOXIDE_INCREASE = 0.3f;
 
+    public const float OXYGEN_THRESHOLD = 0.07f;
+    public const float OXYGEN_PATCHES_THRESHOLD = 0.7f;
+    public const int GLOBAL_GLACIATION_CHANCE = 50;
```
hhyyrylainen, line 1319: I think it is much cleaner if all of our changes are in the range of 0-1 and not use implicit percentages.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,218 @@
...
+    {
+        AdjustBackground(index, patch);
+        AdjustEnvironment(index, patch);
+        AddIceChunks(patch);
+        LogEvent(patch, totalTimePassed);
+    }
+
+    private void AdjustBackground(int index, Patch patch)
+    {
+        previousBackground.Add(index, patch.BiomeTemplate.Background);
```
hhyyrylainen, line 133: I think this needs to be refactored. All of the changed patch properties should be put into the patch snapshot system. I think it is the wrong approach to try to keep track of what every event has changed independently and try to then resolve the mess of 3 different events trying to restore a patch back to the way it should be.

hhyyrylainen (changes requested):
I started commenting and ended up doing an almost full review... though once I saw this major architecture problem I stopped as I think that needs to be the first thing to be actually fixed before it makes sense to polish the rest of the code.

simulation_parameters/microbe_stage/compounds.json
```diff
@@ -228,7 +228,7 @@
     "Volume": 1,
     "IsCloud": false,
     "IsAlwaysUseful": true,
-    "IsEnvironmental": false,
+    "IsEnvironmental": true,
```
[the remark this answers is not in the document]
Patryk26g: I'm adjusting the temperature using  ApplyLongTermCompoundChanges, so I needed it set as enviromental compound

locale/en.po
```diff
@@ -2293,6 +2293,25 @@ msgstr "Visit our GitHub repository"
...
+
+msgid "GLOBAL_GLACIATION_EVENT_LOG"
+msgstr "Global glaciation event in {0}"
+
+msgid "GLOBAL_GLACIATION_EVENT_TOOLTIP"
+msgstr ""
+"Global glaciation event\n"
+"The event has been caused by rising concentration of oxygen in the atmosphere.\n"
+"It will persist for several generations, reducing sunlight and temperature.\n"
+"Snow chunks also appear in these patches now."
```
hhyyrylainen, line 2310:
```suggestion
"Ice chunks also appear in these patches now."
```

simulation_parameters/microbe_stage/compounds.json
```diff
@@ -209,7 +209,7 @@
     "Volume": 1,
     "IsCloud": false,
     "IsAlwaysUseful": false,
-    "IsEnvironmental": true,
+    "IsEnvironmental": false,
```
hhyyrylainen, line 212: This will break photosynthesis entirely... as the process system won't anymore pick up the sunlight from the environment.

src/general/world_effects/WorldEffectVisuals.cs
```diff
@@ -4,6 +4,6 @@
     ///   Used as a placeholder value instead of null
     /// </summary>
     None,
-
```
hhyyrylainen, line 7: Isn't this blank line required by the formatter as there's an XML comment so None is a actually a multiline block meaning that there must be a blank line after it?

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -257,6 +257,7 @@ protected override void ElapseEditorEntryTime()
     {
         // TODO: select which units will be used for the master elapsed time counter
         CurrentGame.GameWorld.OnTimePassed(1);
+        cellEditorTab.UpdateBackgroundImage(CurrentPatch.BiomeTemplate);
```
hhyyrylainen, line 260: I think a comment here would be helpful about the events being able to change the properties of the current patch.

src/multicellular_stage/editor/MulticellularEditor.cs
```diff
@@ -304,6 +304,7 @@ protected override void ElapseEditorEntryTime()
     {
         // TODO: select which units will be used for the master elapsed time counter
         CurrentGame.GameWorld.OnTimePassed(1);
+        cellEditorTab.UpdateBackgroundImage(CurrentPatch.BiomeTemplate);
```
hhyyrylainen, line 307: Or if every single editor needs to implement this, then I think it might be worth considering putting this into the base editor class.

src/microbe_stage/Patch.cs
```diff
@@ -68,6 +69,8 @@ public Patch(LocalizedString name, int id, Biome biomeTemplate, PatchSnapshot cu
         Name = name;
         ID = id;
         BiomeTemplate = biomeTemplate;
+        BiomeTemplate.Background = currentSnapshot.Background;
+        BiomeTemplate.Sunlight.Colour = currentSnapshot.LightColour;
```
hhyyrylainen, line 73: Hang on, this cannot be safe, right? This is modifying the Biome class which is a registry type, which is absolutely forbidden from being modified after loading by SimulationParameters.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,258 @@
...
+
+    public void OnTimePassed(double elapsed, double totalTimePassed)
+    {
+        if (generationsLeft > 0)
+            generationsLeft -= 1;
+
+        if (generationsLeft > 0)
+            MarkPatches(totalTimePassed);
+
+        if (hasEventAlreadyHappened)
```
hhyyrylainen, line 63: Could you clarify the logic why those generation things are done before checking if the event has happened? I think either this needs a clarifying comment or be moved to the start of this method to look like a normal guard clause.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,258 @@
...
+        {
+            FinishEvent();
+        }
+    }
+
+    private void MarkPatches(double totalTimePassed)
+    {
+        foreach (var patch in targetWorld.Map.Patches.Values)
+        {
+            if (patch.Depth[0] == 0)
```
hhyyrylainen, line 80: This is not consistent with the check in TryToTriggerEvent? Is there a reason? I think having a check method if something qualifies as a surface patch would be better to make into a helper method to ensure one place doesn't check things inconsistently.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,258 @@
...
+        AdjustBackground(patch);
+        AdjustEnvironment(patch);
+        AddIceChunks(patch);
+        LogEvent(patch, totalTimePassed);
+    }
+
+    private void AdjustBackground(Patch patch)
+    {
+        patch.BiomeTemplate.Background = Background;
+        patch.CurrentSnapshot.Background = Background;
```
hhyyrylainen, line 142: Modifying the current snapshot is the safe operation, but modifying the template should be absolute not done here.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,258 @@
...
+        patch.BiomeTemplate.Sunlight.Colour = new Color(0.8f, 0.8f, 1, 1);
+        patch.CurrentSnapshot.LightColour = new Color(0.8f, 0.8f, 1, 1);
+    }
+
+    private void AdjustEnvironment(Patch patch)
+    {
+        var currentTemperature = patch.Biome.ChangeableCompounds[Compound.Temperature];
+        var currentSunlight = patch.Biome.ChangeableCompounds[Compound.Sunlight];
```
hhyyrylainen, line 150: I think for safety these should TryGet

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,258 @@
...
+        currentTemperature.Ambient = random.Next(-1, 5) - currentTemperature.Ambient;
+        currentSunlight.Ambient = 0.5f - currentSunlight.Ambient;
+
+        var changes = new Dictionary<Compound, float>
+        {
+            [Compound.Temperature] = currentTemperature.Ambient,
+            [Compound.Sunlight] = currentSunlight.Ambient,
+        };
+
+        patch.Biome.ApplyLongTermCompoundChanges(patch.BiomeTemplate, changes, new Dictionary<Compound, float>());
```
hhyyrylainen, line 161: As this method is applied to multiple patches the changes and the blank created dictionary here should both be class fields so that they don't need to be re-allocated for each method call.

assets/textures/gui/bevel/GlobalGlaciationEvent.svg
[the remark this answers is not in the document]
Patryk26g: Yes, I didn't have time to update it yet

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,258 @@
...
+            true, true, "GlobalGlaciationEvent.svg");
+    }
+
+    private void FinishEvent()
+    {
+        hasEventAlreadyHappened = true;
+        foreach (var index in modifiedPatchesIds)
+        {
+            if (!targetWorld.Map.Patches.TryGetValue(index, out var patch))
+                continue;
```
hhyyrylainen, line 210: Shouldn't this give a warning that a patch has exited the world as that's not expected to happen?

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            PatchSnapshot patchSnapshot = patch.History[eventDuration];
+
+            ResetBackground(patch, patchSnapshot);
+            ResetEnvironment(patch, patchSnapshot);
+            RemoveChunks(patch);
```
hhyyrylainen, line 216: For multiple events to work concurrently, I think we need like a flag in the history snapshot to tell which events are active, that way the history can be scanned for when no event is active. Imagine that we add like a volcano event, and things go like:

- gen 1 no event
- gen 2 volcano event starts
- gen 3 glaciation starts
- gen 4 volcano ends and it reverts things (this can be figured out later but I think the history format should be fixed now)
- gen 6 glaciation ends and it reverts things back to gen 2 *which is incorrect* as the light values will be lower than the default in gen 1

So I think the history needs a flag for active events and probably a helper method for finding conditions without events to revert to by looking backwards.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,258 @@
...
+    {
+        var currentTemperature = patch.Biome.ChangeableCompounds[Compound.Temperature];
+        var currentSunlight = patch.Biome.ChangeableCompounds[Compound.Sunlight];
+        var previousTemperature = patchSnapshot.Biome.ChangeableCompounds[Compound.Temperature];
+        var previousSunlight = patchSnapshot.Biome.ChangeableCompounds[Compound.Sunlight];
+
+        currentTemperature.Ambient = previousTemperature.Ambient - currentTemperature.Ambient;
+        currentSunlight.Ambient = previousSunlight.Ambient - currentSunlight.Ambient;
+
+        var changes = new Dictionary<Compound, float>
```
hhyyrylainen, line 242: I think it would be much easier to use the `ModifyLongTermCondition(Compound compound, BiomeCompoundProperties newValue)` overload here to just apply the old values back.

hhyyrylainen (changes requested):
I only saw one major problem left, which is the modification of the biome template which needs to be gotten rid of. Other than that I commented on mostly just smaller details that should be easier to fix. Though I suppose me asking for patch snapshot system support for more events at once is a bit complicated change to make but makes it much more manageable for future event writers.

Patryk26g:
> I only saw one major problem left, which is the modification of the biome template which needs to be gotten rid of. Other than that I commented on mostly just smaller details that should be easier to fix. Though I suppose me asking for patch snapshot system support for more events at once is a bit complicated change to make but makes it much more manageable for future event writers.

How would you then approach the modifIcation o the background image if the template cannot be changed? The only solution I see is getting the background image from the current snapshot, is that the right approach?

hhyyrylainen:
I would make it so that the template variable can be swapped. So the template objects are immutable, but which template is in use can be changed. So the patches would swap their template from whatever to the ice shelf to apply the background etc. properties.

Patryk26g:
> I would make it so that the template variable can be swapped. So the template objects are immutable, but which template is in use can be changed. So the patches would swap their template from whatever to the ice shelf to apply the background etc. properties.

Right, but that would also change the icon of the patch and besides that I want to create a custom light color, maybe Biome should implement ICloneable so that I can clone it and change its properties?

hhyyrylainen:
I think the map icons changing is fine. And if you want a custom light property then a new biome template could be added that is only used for the glaciation event (and not generated on the map by default).

Patryk26g:
I think I would prefer the icons to stay as they are but if there is no simple solution then I'll do it as you saying. The only other solutuion I have in mind is creating a new template based on the base one this way:

```
private void CreateGlaciatedBiomeTemplate(Biome biomeTemplate)
    {
        Biome glaciatedBiomeTemplate = new Biome
        {
            ActiveMusicContexts = biomeTemplate.ActiveMusicContexts,
            Background = "iceshelf",
            Conditions = (BiomeConditions)biomeTemplate.Conditions.Clone(),
            Icon = biomeTemplate.Icon,
            InternalName = biomeTemplate.InternalName,
            Name = biomeTemplate.Name,
            Sunlight = new Biome.LightDetails
            {
                Colour = new Color(0.8f, 0.8f, 1, 1),
            },
            Panorama = biomeTemplate.Panorama,
        };
    }
```

Is that an acceptable approach?

hhyyrylainen:
Cloning the biome template again has the issue that it will make save compatibility a nightmare as any slight change to the biomes data will not propagate that config change to the cloned templates. So the choice is either a full template switch system where patches can swap between templates due to events, or an addition of data to the snapshots that allow overriding the properties (background image, sunlight, though there will probably be even more so this will bloat the snapshots).

If the snapshots approach is taken even the background images need to be carefully designed how they are saved as it is once again bad for save compatibility if they store the exact paths to the background images. So another layer of indirection will be needed there. That's also one reason why I think that allowing the template type to switch dynamically is the easiest solution in terms of allowing everything to change but at the same time not introducing save compatibility gotchas.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
Patryk26g: I mostly agree with you but I think that the system for the patch events overlapping should be added and thought about when we are going to add a new patch event. After this event I'll probably start working on another one so it will be much easier and clearer for me to do in the next PR

src/multicellular_stage/editor/MulticellularEditor.cs
Patryk26g: I know, but the base class doesn't have `cellEditorTab` or `CurrentPatch`, I think it's not worth changing it for two lines of code multiplicated three times

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
hhyyrylainen: Alright.

I realized after reading my previous message again that my design would be insufficient as we need like a full system that is given the set of active patch events and then determines what patch conditions should be set to. Such a system should be able to correctly handle events starting and stopping at any time and be able to apply any mix of active patch events.

src/multicellular_stage/editor/MulticellularEditor.cs
hhyyrylainen: I was thinking more along the lines of the base class triggering a method about applying the current patch details to the environment after calling ElapseEditorEntryTime, that way we get this working "for free" without needing to add a bunch of extra logic to handle anything that might change in the current patch based on the elapsed time. The logic here would get pretty complex once sunlight can be reduced and the editor should show numbers calculated with the latest patch data. Though I guess that must already exists? Otherwise there's a bug. So I would actually recommend just making sure the latest image is also applied in that method.

src/general/world_effects/patch_events/UnderwaterVentEruptionEffect.cs.uid
hhyyrylainen, line 1: If this file is new then did the old location still have an unused .uid file?

src/microbe_stage/Patch.cs
```diff
@@ -648,12 +652,14 @@ public class PatchSnapshot : ICloneable
     public Dictionary<Species, SpeciesInfo> RecordedSpeciesInfo = new();
 
     public BiomeConditions Biome;
+    public string Background;
```
hhyyrylainen, line 655: How is old save compatibility going to be handled here? Maybe this could be nullable and the Patch.Background property could fallback to the old biome template read in that situation?

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,257 @@
...
+
+        if (!hasTemperature)
+        {
+            GD.PrintErr("Patch has no temperature");
+            return;
+        }
+
+        if (!hasSunlight)
+        {
+            GD.PrintErr("Patch has no sunlight");
```
hhyyrylainen, line 164: Is this actually an error? Patches do exist that are allowed to have no sunlight.

Patryk26g: Yes, but all surface patches must have light. I will update the text to make it more obvious

hhyyrylainen: Yeah, I did see that the code calling this makes sure these are surface patches. Though it isn't totally optimal that an error would only be triggered once the event tries to start, if there is a new bug introduced in the map generator that accidentally leaves out sunlight.

src/microbe_stage/BiomeConditions.cs
```diff
@@ -207,7 +207,14 @@ public void ApplyLongTermCompoundChanges(Biome biomeDetails, Dictionary<Compound
         foreach (var entry in changes)
         {
             var definition = simulationParameters.GetCompoundDefinition(entry.Key);
-            if (!definition.IsEnvironmental)
+            if (!definition.IsEnvironmental && !definition.IsGas)
```
hhyyrylainen, line 210: What's the purpose behind this change in this file?

src/general/base_stage/EditorBase.cs
```diff
@@ -827,6 +828,11 @@ protected virtual void ElapseEditorEntryTime()
         throw new GodotAbstractMethodNotOverriddenException();
     }
 
+    protected virtual void UpdatePatchDetails()
```
hhyyrylainen, line 831: I think this should have a summary comment explaining when this is triggered and what the purpose is because there's no way to tell from the name alone what this method is about.

src/microbe_stage/BiomeConditions.cs
Patryk26g: for temperature update, which can go under value of 0

Patryk26g: nevermind, I forgot that I started using a different mothed, I will revert that

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
Patryk26g: So do you want the error print removed and to just leave the `return` statement?

Patryk26g: I mean, don't event depth patcch have light level which equals 0? I think that this error print should be there

hhyyrylainen: I guess it is better to leave the error in but I think it should be worded more like "Glaciation event encountered patch with unexpectedly no sunlight" like that to emphasise it's a guard clause kind of thing and based on assumptions in the glaciation event code.

Patryk26g:
@hhyyrylainen does this PR need any other work besides master merge? Or is it ready to be merged in your opinion?

hhyyrylainen:
Does this yet have the fix for multiple events being able to start and stop at different times and still be able to correctly restore the old state?

I was kind of waiting for that change to be done before reviewing the code again. Sorry if I missed it being added already.

Patryk26g:
I thought that we came to conclusion in previous comments that such system would come in another PRs with more events as now it is not needed. And I changed the glaciation event a little bit in my meteor event PR to be able to work properly with that one so that previous states would be restored properly

hhyyrylainen:
I got the opposite impression, that you wanted to prepare for the meteor event in *this* PR, oops.

I guess I can re-review this PR, but if there is still the problem here that it cannot restore stuff correctly, I'll ask for changes here.

Patryk26g:
It restores the state correctly and the PR is ready for review

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,284 @@
...
+            return;
+        }
+
+        switch (generationsToTrigger)
+        {
+            case > 0:
+                LogHeadUpEventWarning();
+                generationsToTrigger -= 1;
+                break;
+            case 0:
```
hhyyrylainen, line 113: Seems a bit strange to me to use a switch statement for 2 possible branches, an if would be a lot shorter.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,284 @@
...
+        numberOfSurfacePatches = numberOfSurfacePatches == 0 ? 1 : numberOfSurfacePatches;
+
+        return (float)patchesExceedingOxygenLevel / numberOfSurfacePatches >=
+            Constants.GLOBAL_GLACIATION_PATCHES_THRESHOLD
+            && random.NextFloat() <= Constants.GLOBAL_GLACIATION_CHANCE;
+    }
+
+    private void ChangePatchProperties(int index, Patch patch, double totalTimePassed)
+    {
+        modifiedPatchesIds.Add(index);
```
hhyyrylainen, line 159: Index is not the ID. So this code is incorrect: ` ChangePatchProperties(index, patch, totalTimePassed);`

You should actually use the patch ID.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,284 @@
...
+        targetWorld.LogEvent(new LocalizedString("GLOBAL_GLACIATION_END_EVENT_LOG"),
+            true, true, "GlobalGlaciationEvent.svg");
+    }
+
+    private void FinishEvent()
+    {
+        hasEventAlreadyHappened = true;
+        foreach (var index in modifiedPatchesIds)
+        {
+            if (!targetWorld.Map.Patches.TryGetValue(index, out var patch))
```
hhyyrylainen, line 247: Actually related to my earlier comment, it kind of looks like you are using the IDs? But call them indexes??

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
hhyyrylainen: This will be handled in the meteor event PR.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,284 @@
...
+
+    private void ResetBackground(Patch patch, PatchSnapshot patchSnapshot)
+    {
+        patch.CurrentSnapshot.Background = patchSnapshot.Background;
+    }
+
+    private void ResetEnvironment(Patch patch, PatchSnapshot patchSnapshot)
+    {
+        var previousTemperature = patchSnapshot.Biome.ChangeableCompounds[Compound.Temperature];
+        var previousSunlight = patchSnapshot.Biome.ChangeableCompounds[Compound.Sunlight];
```
hhyyrylainen, line 271: Shouldn't these read the long term here? If I recall correctly there's a method that clearly reads that. For example `TryGetCompound(Compound.Sunlight, CompoundAmountType.Biome, out var sunlight)`

hhyyrylainen:
Okay, I reviewed this again. I think I noticed one bug but other than that there's just some code cleanliness stuff I commented on.

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -0,0 +1,284 @@
...
+        hasEventAlreadyHappened = true;
+        foreach (var index in modifiedPatchesIds)
+        {
+            if (!targetWorld.Map.Patches.TryGetValue(index, out var patch))
+            {
+                GD.PrintErr("Patch exited the world");
+                continue;
+            }
+
+            PatchSnapshot patchSnapshot = patch.History[eventDuration];
```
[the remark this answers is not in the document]
Patryk26g: Well, no, history snapshots are put at the beggining of the history queue, not at the end. In this case the event lasted 2 generations. 0 is current, 1 is previous and 2 is the one before the event triggered.
image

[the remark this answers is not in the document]
hhyyrylainen: Ah, sorry I missed that. I just quickly read `IReadOnlyList<PatchSnapshot> History => history;` and assumed stuff was added to the end, but you are right it is actual a Deque and `history.AddToFront(snapshot);` is used.

So you are correct and this isn't actually a problem.

hhyyrylainen (changes requested):
I was doing final testing and I noticed a bug. The glaciation events *per patch* are put in the global timeline and not the patch local timelines. So this is incorrect:



The patch local timelines list "Global glaciation event" already so I think that is fine, just need to remove that huge spam of events from the global timeline.

hhyyrylainen (approved):
With the event log change this should be perfect now. I'll assume from the code change that it works without re-testing.

PR #5986, opened 2025-03-06
Fix hotbar overlap with heat view button

PollastreGH:

Decreases the size of the `ScrollContainer` containing the `HotBar` so that it no longer covers the heat view button. The size 415px was chosen so that scrolling isn't required if there are up to 8 available abilities while not overlapping with any buttons

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I think the same fix also needs to be applied to the macroscopic HUD file as that doesn't use the same file as microbe and multicellular.

hhyyrylainen (approved):
The size change looks fine. I didn't playtest it, though.

PR #5702, opened 2024-11-22
Added realistic movement trail

AverageNerdz:
**Brief Description of What This PR Does**

This PR adds a visual enhancement for microbe movements, the core class of this implementation is the c++ "Ripple" class that generates smooth, fading trails behind all moving microbes, then the scene emits bubble particles around the trail, all particles are generated by a shader with the same name as the new class. Under the hood the implementation uses triangle strips for the mesh generation it automatically cleans old trail points based on age and distance. 

**Related Issues**

**Progress Checklist**

Note: before starting this checklist the PR should be marked as non-draft.

- [ ] PR author has checked that this PR works as intended and doesn't
      break existing features:
      https://wiki.revolutionarygamesstudio.com/wiki/Testing_Checklist
      (this is important as to not waste the time of Thrive team
      members reviewing this PR)
- [ ] Initial code review passed (this and further items should not be checked by the PR author)
- [ ] Functionality is confirmed working by another person (see above checklist link)
- [ ] Final code review is passed and code conforms to the 
      [styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md).

Before merging all CI jobs should finish on this PR without errors, if
there are automatically detected style issues they should be fixed by
the PR author. Merging must follow our
[styleguide](https://github.com/Revolutionary-Games/Thrive/blob/master/doc/style_guide.md#git).

PR #5996, opened 2025-03-09
Cell Fluid Dynamics trail effect

AverageNerdz:
Adds a transparent water ripple effect that creates dynamic fluid disturbances as cells move through the environment.

**Related Issues**
Closes #5702

**Progress Checklist**
      break existing features:
      members reviewing this PR)


Patryk26g:
Looks good, but the effect is barely visible for large cells with nucleus. Is it possible to somehow scale the range of the ripples with the cell size?

AverageNerdz:
> Looks good, but the effect is barely visible for large cells with nucleus. Is it possible to somehow scale the range of the ripples with the cell size?

yes pal, you can play around with the parameters in the editor just look it up on the Membrane.tscn files ;)

shaders/MembraneWaterRipple.gdshader
```diff
@@ -0,0 +1,182 @@
+shader_type spatial;
+render_mode blend_mix, depth_prepass_alpha, cull_back, diffuse_lambert;
+
+uniform vec4 water_color : source_color = vec4(0.0, 0.0, 0.0, 0.02); // Almost completely transparent background
```
hhyyrylainen, line 4: Our shaders should also use our C# naming convention so this should be converted to use pascalCase naming convention rather than snake_case. Also the trailing comments should be changed to be comments on the line before what they are commenting on to match Thrive style.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,452 @@
...
+    public Color WaterColor = new(0.0f, 0.0f, 0.0f, 0.02f);
+
+    [Export]
+    public float WaterSize = 18.0f; // Base water size
+
+    [Export]
+    public float Phase = 0.2f;
+
+    /*Controls how quickly ripples dissipate
+    higher values = less viscous appearance*/
```
hhyyrylainen, line 26: Should this be formatted as an XML summary comment?

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,452 @@
...
+    public bool DynamicScaling = true;
+
+    [Export]
+    public float ScaleCheckInterval = 0.5f;
+
+    [Export]
+    public float RadiusMultiplier = 2.5f;
+
+    [Export]
+    public int MaxMeshSubdivision = 100; // Reduced for better performane can be increased for a more detailed effect
```
hhyyrylainen, line 44: Trailing comments shouldn't be used

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,452 @@
...
+
+    [Export]
+    public float VerticalOffset = -0.15f; // How far below the membrane the water plane should be
+
+    // Default radius if we can't determine it from the membrane
```
hhyyrylainen, line 49: This definitely needs to be an XML summary comment.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,452 @@
...
+    private float lastScaleCheckTime;
+    private float lastCheckedMembraneRadius;
+    private bool isSetup;
+
+    // Called when the node enters the scene tree for the first time
+    public override void _Ready()
+    {
+        // Initialize the effect
+        SetProcess(true);
+        SetPhysicsProcess(true);
```
hhyyrylainen, line 72: These should be automatic, so not required to be explicitly written like this.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,452 @@
...
+
+        // Find the parent membrane
+        parentMembrane = GetParent<Membrane>();
+        if (parentMembrane == null)
+        {
+            GD.PrintErr("MembraneWaterRipple must be a child of a Membrane");
+            QueueFree();
+            return;
+        }
```
hhyyrylainen, line 81: This should be done the other way around, i.e. the membrane initializing the component inside it and not calling up to the parent. That's basically always a sign of bad Godot scene tree usage.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,452 @@
...
+
+        // Initialize position tracking for movement calculation
+        lastPosition = parentMembrane.GlobalPosition;
+        previousPosition = lastPosition;
+        lastMovementTime = Time.GetTicksMsec() / 1000.0f;
```
hhyyrylainen, line 86: Time shouldn't be done like this, only rely on the delta values given in _Process or _PhysicsProcess.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,452 @@
...
+        CreateWaterPlane();
+
+        // Set initial position
+        UpdatePosition();
+    }
+
+    // Called every frame for visual updates
```
hhyyrylainen, line 93: This is like a default comment that should be deleted. Or if you really want to keep this, this must be reformatted as an XML comment.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,452 @@
...
+            if (waterMaterial != null)
+            {
+                float currentTime = Time.GetTicksMsec() / 1000.0f;
+                waterMaterial.SetShaderParameter("time_offset", currentTime);
+            }
+        }
+        catch (Exception e)
+        {
+            GD.PrintErr("Error in water ripple process: " + e.Message);
+        }
```
hhyyrylainen, line 114: There shouldn't be a catch like this in a component.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,452 @@
...
+
+        try
+        {
+            // Update position every frame to ensure perfect alignment
+            UpdatePosition();
+
+            // Update shader time with exact original timing
+            if (waterMaterial != null)
+            {
+                float currentTime = Time.GetTicksMsec() / 1000.0f;
```
hhyyrylainen, line 107: Time passing should be taken from delta and not like this.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,452 @@
...
+            }
+        }
+        catch (Exception e)
+        {
+            GD.PrintErr("Error in water ripple process: " + e.Message);
+        }
+    }
+
+    // Called on physics process for movement and simulation updates
+    public override void _PhysicsProcess(double delta)
```
hhyyrylainen, line 118: Is there an actual performance improvement from splitting the processing into two parts? I would guess that as you already check the time elapsed here, it would be more performance friendly to put all of the content from here into _Process and using a double variable to keep track when the extra checks need to run.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,452 @@
...
+
+            // Fallback: estimate from scale if membrane is Node3D
+            if (parentMembrane is Node3D node3D)
+            {
+                return node3D.Scale.X * 2.0f;
+            }
+
+            return DEFAULT_MEMBRANE_RADIUS;
+        }
+        catch (Exception)
```
hhyyrylainen, line 447: And this silently ignores errors in reading the data??

hhyyrylainen (changes requested):
I didn't read all of the shader but other than that this is a full code review. Sorry if the tone of the comments gets worse by the end of the file but by that point I was actually surprised about how the implementation was done.

AverageNerdz:
> I didn't read all of the shader but other than that this is a full code review. Sorry if the tone of the comments gets worse by the end of the file but by that point I was actually surprised about how the implementation was done.

Pal you shouldn’t hold back you’re just giving me suggestions to improve the code and you as always are very polite so there’s nothing to worry about, originally the dispose part wasn’t in the original code I tried to experiment with it after I had to change the implementation the shader is really expensive as you said so the solution you proposed is probably the right way to do it

hhyyrylainen:
Thanks for the encouragement. I do worry about coming across as too demanding in reviews, but that of course also depends on how the recipient feels about the review.

> originally the dispose part wasn’t in the original code I tried to experiment with it after I had to change the implementation

Yeah, I can see how that happened. Unless I'm misremembering the style guide should actually have the advice on how to deal with the Dispose stuff.

hhyyrylainen:
One more thing I just realized, the new .uid files should be committed to git otherwise the game causes unknown uid warnings on everyone else's computer.

image

Should be committed to this branch.

HyperbolicHadron:
I think the effect looks really good when going in one direction, though when turning the effect looks strange for the older ripples, as it does not quite take into account the real history of the organism's position. I haven't looked into the trail calculation code too much, but if it is not feasible to pass in the real history of the organism's position then perhaps a more accurate approximation could be calculated, such as using the microbe's instantaneous turning angle rate to calculate an approximation of the path traced with a turn. Or also pass in the average acceleration of the turning angle to give it more history. These may or may not be good suggestions, its just what I immediately thought of.

dligr:
I tested it and the effect looks very good. Though, in my opinion, it should be made more visible in the surface patches, where it would add a lot to the atmosphere with the way it interacts with lighting.

I also agree with HyperbolicHadron in that the effect's direction gets weird when quickly turning. Maybe the C# side could pass a few past position arguments, and the shader interpolate between them?

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,572 @@
...
+                    timeScale = 1.0f + currentSpeed * 0.12f;
+                    break;
+            }
+        }
+
+        timeAccumulator += (float)delta * timeScale;
+        waterMaterial?.SetShaderParameter(timeOffsetParam, timeAccumulator);
+
+        // Update movement parameters (with reduced frequency for distant objects)
+        if (currentLodLevel == LodLevel.VeryLow)
```
hhyyrylainen, line 223: Does this actually help anything? As this is a parented node then I think the full transform matrix has to be calculated each frame basically anyway so I'd expect this extra checking to just eat up more performance than it has any chance of helping with.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,572 @@
...
+        }
+        else
+        {
+            UpdateMovementParameters((float)delta);
+        }
+    }
+
+    private void InitializeShaderParameters()
+    {
+        if (waterMaterial == null)
```
hhyyrylainen, line 239: Water material seems important enough to make it non-nullable and then throw big errors in _Ready if it is missing.

Suppressing errors without even a GD.PrintErr is almost the worst thing to do as someone will run into a bug like this at some point and then we'll be totally clueless as to what went wrong.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,572 @@
...
+
+        // Quick distance check first
+        if (isCameraPositionValid && lastCameraDistance > GetMembraneRadius() * DistanceCullThreshold)
+        {
+            // If very far away consider not visible
+            return false;
+        }
+
+        Vector3 membranePos = parentMembrane.GlobalPosition;
+        Vector2 screenPos = camera.UnprojectPosition(membranePos);
```
hhyyrylainen, line 277: How much all of this LOD and visibility calculation complexity actually help with performance?

This is the kind of code that should be generalized or we should use the Godot Engine methods for this, because we will end up with an unimaginable mess if we end up with even 10 objects all implementing their own custom visibility and LOD systems.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,572 @@
...
+            case LodLevel.High:
+                subdivision = HIGH_LOD_SUBDIVISION;
+                break;
+
+            default:
+                subdivision = MEDIUM_LOD_SUBDIVISION;
+                break;
+        }
+
+        planeMesh.SubdivideWidth = subdivision;
```
hhyyrylainen, line 389: How much performance does this take? I imagine this regenerates the mesh which I can't imagine is good for performance. Probably better to have just static instances of each plane mesh and swap between them dynamically.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,572 @@
...
+        if (parentMembrane == null || waterPlane == null)
+            return;
+
+        Vector3 membranePos = parentMembrane.GlobalPosition;
+        Vector3 waterPos = membranePos;
+        waterPos.Y += VerticalOffset;
+        waterPlane.GlobalPosition = waterPos;
+
+        // Force water plane to have zero rotation (flat horizontal plane)
+        waterPlane.GlobalRotation = new Vector3(0, 0, 0);
```
hhyyrylainen, line 407: Setting these separately uses two method calls into the engine. Would be much preferable to just build a Transform and set that with one operation.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,572 @@
...
+            step = 2;
+
+        // Access circular buffer with wrapping index
+        // Transforms world positions to local XZ offsets for shader
+        for (int i = 0; i < count; i += step)
+        {
+            int index = (currentPositionIndex - 1 - i + PositionHistorySize) % PositionHistorySize;
+
+            if (index < 0)
+                index += PositionHistorySize;
```
hhyyrylainen, line 547: Can't this be flipped to make the math a lot more obvious?

shaders/MembraneWaterRipple.gdshader
```diff
@@ -0,0 +1,272 @@
+shader_type spatial;
+render_mode blend_mix, depth_prepass_alpha, cull_back, diffuse_lambert;
+
+// Almost completely transparent background
+uniform vec4 WaterColor : source_color = vec4(0.0, 0.0, 0.0, 0.02);
+// Intensity of ripple displacement
```
hhyyrylainen, line 6: There should be a blank line before each line comment so that this isn't such a singular blob of stuff.

shaders/MembraneWaterRipple.gdshader
```diff
@@ -0,0 +1,272 @@
...
+// Constants for mesh sizing and effect configuration
+const float MeshSize = 8.0;
+const float EdgeFadeStart = 6.0; 
+const float EdgeFadeEnd = 0.4;
+const int MaxPathSamples = 12;
+const int SecondaryStart = 2;
+const int SecondaryEnd = 10;
+
+// Direct linear interpolation between two positions
+vec2 InterpolatePositions(vec2 pos1, vec2 pos2, float factor) {
```
hhyyrylainen, line 32: Is there no in-built Godot way to do this?

shaders/MembraneWaterRipple.gdshader
```diff
@@ -0,0 +1,272 @@
...
+    return InterpolatePositions(PastPositions[lowerIndex], PastPositions[upperIndex], factor);
+}
+
+// Calculate wave height at given position relative to a center point
+float WaveHeight(vec2 position, vec2 center, float time, float elapsed) {
+    // Distance from wave origin
+    float dist = length(position - center);
+    
+    // Early exit for distant points
+    if (dist > 10.0) return 0.0;
```
hhyyrylainen, line 66:
```suggestion
    if (dist > 10.0) 
        return 0.0;

```

shaders/MembraneWaterRipple.gdshader
```diff
@@ -0,0 +1,272 @@
...
+    // Simplified decay calculation
+    return combined * exp(-elapsed * 0.3) * exp(-0.10 * dist);
+}
+
+// Calculate wake trail effect between two position points
+float CalculateWakeTrail(vec2 position, vec2 pos1, vec2 pos2, float timeAgo, float time) 
+{
+    // Skip if positions are too close
+    float segmentLength = length(pos2 - pos1);
+    if (segmentLength < 0.01) return 0.0;
```
hhyyrylainen, line 98:
```suggestion
    if (segmentLength < 0.01)
        return 0.0;

```

shaders/MembraneWaterRipple.gdshader
```diff
@@ -0,0 +1,272 @@
...
+    if (PastPositionsCount >= 2) {
+        // Determine sample step based on distance
+        int sampleStep = 1;
+        if (centerDist > 8.0) sampleStep = 3;
+        else if (centerDist > 4.0) sampleStep = 2;
+         int maxSamples = min(MaxPathSamples, PastPositionsCount);
+        
+        // Sample the path
+        for (int i = 0; i < maxSamples; i += sampleStep) {
+            if (i >= PastPositionsCount) break;
```
hhyyrylainen, line 161: I'll now stop commenting on this specific style problem.

shaders/MembraneWaterRipple.gdshader
```diff
@@ -0,0 +1,272 @@
...
+            float blendFactor = min(1.0, abs(displacement) * 8.0);
+            NORMAL = normalize(mix(NORMAL, gradientNormal, blendFactor));
+        }
+    }
+    // For distant vertices, we don't modify VERTEX or NORMAL
+}
+
+void fragment() {
+    // Calculate lighting parameters
+    float nDotL = dot(NORMAL, normalize(vec3(0.4, 0.8, 0.3)));
```
hhyyrylainen, line 261: nDotL especially arcane to me, especially as the variable on the next line seems to have a descriptive name...

I'm not a graphics programmer so I couldn't really say anything else about the shader other than the small style things.

hhyyrylainen (changes requested):

src/microbe_stage/MembraneWaterRipple.cs
dligr: Without changing the logic, it could be simplified to this by using MathUtils' `PositiveModulo`:
```C#
for (int i = 0; i < count; i += step)
{
    int index = (currentPositionIndex - 1 - i).PositiveModulo(PositionHistorySize);
    ...
```

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,572 @@
...
+        timeAccumulator += (float)delta * timeScale;
+        waterMaterial?.SetShaderParameter(timeOffsetParam, timeAccumulator);
+
+        // Update movement parameters (with reduced frequency for distant objects)
+        if (currentLodLevel == LodLevel.VeryLow)
+        {
+            // For very distant membranes, update less frequently
+            if ((int)(timeAccumulator * 10) % 3 == 0)
+            {
+                UpdateMovementParameters((float)delta);
```
dligr, line 228: Since this is only called once every 0.3 seconds instead of every update, doesn't this call have to pass 0.3f instead of delta as the parameter?

dligr (commented):
Having tested this PR, I can say that I generally like the effect. It's good to see that after the last commits ripples only change their direction gradually when the cell does.

In my opinion, it would be better if non-moving cells emitted no ripples, and generally if ripples' visibility depended on the cell's speed.

I have also noticed one visual bug: after swimming around for some time, it's possible for the effect to completely freeze its direction. So, even when the cell changes its direction, ripples' direction is the same. It returns to normal after a few seconds however.

hhyyrylainen:
I'm not sure what I'm supposed to be reviewing as multiple of my previous comments are still not solved, including the Initialize method taking in a Membrane object and the custom LOD system being still baked into the membrane class, and a few smaller style things.

hhyyrylainen:
That suggestion was about removing the `?` and didn't imply anything else. It was just an example of how stuff that is considered initialized in _Ready should be considered non-nullable as that massively improves the clarity when working with Godot. Because otherwise like all Godot variables would have to be nullable, leading to massive overcomplication of all methods that would need to start like 4+ lines of null checks, which I hope everyone can agree is going to be really annoying to write and read.

AverageNerdz:
> That suggestion was about removing the `?` and didn't imply anything else. It was just an example of how stuff that is considered initialized in _Ready should be considered non-nullable as that massively improves the clarity when working with Godot. Because otherwise like all Godot variables would have to be nullable, leading to massive overcomplication of all methods that would need to start like 4+ lines of null checks, which I hope everyone can agree is going to be really annoying to write and read.

I see and what about the LOD system?

hhyyrylainen:
The problem with that is still in my opinion that such a thing does not belong to a single object's class, because otherwise we'll end up with 20 varying LOD implementations in different objects once more objects that need LOD are made. So that's why I think that this code should either use an engine feature, or separate out the LOD code to a helper class that can be used by other objects in the future.

src/microbe_stage/Membrane.cs
```diff
@@ -79,17 +84,29 @@ public MembranePointData MembraneData
             if (ReferenceEquals(membraneData, value))
                 return;
 
-            bool reapply = membraneData != null!;
+            float oldRadius = 0;
```
hhyyrylainen, line 87: Could you elaborate on why this bool juggling is needed to be added?

src/microbe_stage/Membrane.cs
```diff
@@ -370,9 +389,12 @@ private void ApplyTextures()
         albedoTexture = Type.LoadedAlbedoTexture;
 
         // This is called rarely enough that this just plain re-creates StringName instances here each time
-        MembraneShaderMaterial!.SetShaderParameter("albedoTexture", albedoTexture);
-        MembraneShaderMaterial.SetShaderParameter("normalTexture", Type.LoadedNormalTexture);
-        MembraneShaderMaterial.SetShaderParameter("damagedTexture", Type.LoadedDamagedTexture);
+        if (MembraneShaderMaterial != null)
```
hhyyrylainen, line 392: What's the reason why this code is required now? If this kind of error checking is needed then a bigger refactor is needed as this would just *silently ignore* applying a material and pretend that it applied correctly, which is not the right way to do error handling.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,666 @@
...
+    public float ResumeMovementThresholdSqr = 0.0003f;
+
+    /// <summary>
+    ///   Maximum delta time to prevent jittering
+    /// </summary>
+    private const float MAX_DELTA_TIME = 0.1f;
+
+#pragma warning disable CA2213
+
+    private readonly StringName timeOffsetParam = new("TimeOffset");
```
hhyyrylainen, line 71: These should be outside the pragma as StringNames need to be disposed.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,666 @@
...
+    public bool EnableEffect = true;
+
+    [Export]
+    public float RippleStrength = 0.8f;
+
+    [Export]
+    public float VerticalOffset = -0.05f;
+
+    /// <summary>
+    ///   Number of past positions to store for the movement history
```
hhyyrylainen, line 20: Does this need to match what is in the shader code? If so it would be pretty good to have warnings here not to change things without updating the shader as well.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,666 @@
...
+    /// </summary>
+    private Camera3D? currentCamera;
+
+    private float lastCameraDistance;
+    private Vector3 lastCameraPosition;
+    private bool isCameraPositionValid;
+
+    // initialization flag
+    private bool isInitialized;
+#pragma warning restore CA2213
```
hhyyrylainen, line 167: There's a bunch of normal primitive type variables that should absolutely not be inside the warning disable block

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,666 @@
...
+            targetAlpha = 0.0f;
+            stillnessValue = 1.0f;
+            targetStillness = 1.0f;
+            isForming = false;
+            formingTime = 0.0f;
+            stillnessFactor = 0.0f;
+            stillnessTimer = 0.0f;
+            timeWithoutMovement = 0.0f;
+            currentPositionIndex = 0;
+            isPositionHistoryFull = false;
```
hhyyrylainen, line 198: Ton of this init code seems like it would be just much better to be directly in as field initializers...

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,666 @@
...
+        {
+            visibilityCheckTimer = 0.0f;
+            UpdateCameraCache();
+            isCurrentlyVisible = IsVisible();
+
+            // Toggle water plane visibility
+            if (waterPlane.Visible != (isCurrentlyVisible && isEffectEnabled))
+            {
+                waterPlane.Visible = isCurrentlyVisible && isEffectEnabled;
+            }
```
hhyyrylainen, line 242: Isn't this kind of unnecessary? If I remember right IsVisible just returns the Godot property `Visible` which is basically a visibility flag on this node itself, which doesn't seem all that important to check? And also the water plane will be invisible anyway if this node (its parent) is invisible. Or did I misunderstand something here?

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,666 @@
...
+
+        if (!isCurrentlyVisible || !isEffectEnabled)
+            return;
+
+        if (positionTarget.IsInsideTree())
+        {
+            Transform3D transform = new Transform3D(Basis.Identity,
+                new Vector3(positionTarget.GlobalPosition.X,
+                    positionTarget.GlobalPosition.Y + VerticalOffset,
+                    positionTarget.GlobalPosition.Z));
```
hhyyrylainen, line 253: Isn't this just `positionTarget.GlobalPosition + new Vector3(0, VerticalOffset, 0)`? That avoid reading the Godot property multiple times which is unnecessarily expensive call to make to the engine here.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,666 @@
...
+    }
+
+    /// <summary>
+    ///   Updates the target and size from an external source or initializes the effect.
+    /// </summary>
+    public void UpdateTarget(Node3D? target, float radius)
+    {
+        // Early return if not properly initialized
+        if (!isInitialized)
+            return;
```
hhyyrylainen, line 271: Seems like the Membrane will not retry this call. So again, it is incorrect to silently ignore the error as then this is left with inconsistent state in regards to the actual data source.

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,666 @@
...
+    /// <summary>
+    ///   Check if the effect is currently visible
+    /// </summary>
+    private bool IsVisible()
+    {
+        if (positionTarget == null || currentCamera == null || !positionTarget.IsInsideTree())
+            return false;
+
+        return GetViewport().GetVisibleRect().HasPoint(currentCamera.UnprojectPosition(positionTarget.GlobalPosition));
+    }
```
hhyyrylainen, line 414: Is updating the effect on the C# side already very expensive or why does this basically get around the normal frustrum culling that should already be in the engine?

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,666 @@
...
+        Vector3 currentPos = positionTarget.GlobalPosition;
+
+        // Calculate local positions relative to current position
+        int count = isPositionHistoryFull ? PositionHistorySize : currentPositionIndex;
+
+        // Optimize based on camera distance so uses fewer samples for distant objects
+        int step = 1;
+        if (lastCameraDistance > 120.0f)
+            step = 3;
+        else if (lastCameraDistance > 80.0f)
```
hhyyrylainen, line 635: Multiline if should always use braces

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,666 @@
...
+            {
+                pastPositions[targetIndex] = position;
+                godotPastPositions[targetIndex] = position;
+            }
+        }
+
+        // Fill remaining positions with zeros for lower detail sampling
+        if (step > 1)
+        {
+            for (int i = count / step + 1; i < count && i < pastPositions.Length && i < godotPastPositions.Count; ++i)
```
hhyyrylainen, line 659: Doesn't this end early? If any of the lengths are not the same, then this will fill only positions with zero until the first length check that fails?

I really think you ought to make a constant for how many elements each of the positions arrays must always have and then rewrite the code a bit to take advantage of that to simplify the bounds checking and ensure there won't be strange bugs if someone ever changes the size of one of the arrays.

src/microbe_stage/Membrane.cs
AverageNerdz: At first i used this logic:
bool reapply = membraneData != null;
float oldRadius = reapply && membraneData != null ? membraneData.Radius : 0;
----
which generated the jetbrain error: ConditionIsAlwaysTrueOrFalseAccordingToNullableAPIContract

i then switched to this:
bool reapply = membraneData != null!;
float oldRadius = reapply ? membraneData.Radius : 0;

which was way simpler, but this generates a CS8602 warning, i used this juggling cause it was the closest one to the solutions above that didn't have problems unless its predecessors

hhyyrylainen: You probably have defined the `membraneData` field as non-nullable. So the compiler is helpfully noticing that you are doing an absolutely silly thing: comparing a thing that is never null to null (which should never be true). If the thing *can be* null, then it has to be marked as a nullable type (`Membrane?` for example, I'm not sure what the variable's type is).

AverageNerdz: I did, problem is that control was needed regardless of the nature of the field removing it caused problems with the code's execution

hhyyrylainen: So then you actually have a much deeper-rooted architecture problem in the code.

Usually these cases with some data not being available yet in Thrive are handled by making the field nullable and *skipping* operations that depend on the data before it is available. But then *also* there's code to retrigger all of those actions when _Ready is called to ensure that a key operation is not silently just skipped potentially forever if there's a bug in the code.

shaders/MembraneWaterRipple.gdshader
```diff
@@ -0,0 +1,291 @@
...
+{
+    // Calculate lighting parameters
+    float nDotL = dot(NORMAL, normalize(vec3(0.4, 0.8, 0.3)));
+    float viewDotNormal = max(0.0, dot(normalize(VIEW), NORMAL));
+
+    // Highlights and fresnel calculation
+    float enhancedHighlight = pow(max(0.0, nDotL), 8.0) * 0.25;
+    float enhancedFresnel = pow(1.0 - viewDotNormal, 2.0) * 0.4;
+
+    ALBEDO = vec3(0.0, 0.0, 0.0) + vec3(enhancedHighlight) + vec3(0.8, 0.9, 1.0) * enhancedFresnel;
```
dligr, line 286: Should we calculate lighting here (enchancedHighlight in particular), given that this shader already has `diffuse_lambert` enabled (which does almost the same)?

src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -0,0 +1,730 @@
...
+
+        float sizeScale = Math.Clamp(membraneRadius / 5.0f, 1.0f, 2.0f);
+        float scaledDistance = lastCameraDistance / sizeScale;
+
+        // Map distance (0-120) to subdivision (120-40) with a non-linear curve
+        float t = Math.Clamp(scaledDistance / 120.0f, 0.0f, 1.0f);
+        t = MathF.Pow(t, 1.5f);
+
+        // Calculate subdivision from min (40) to max (120)
+        int subdivision = (int)Mathf.Lerp(120, 40, t);
```
dligr, line 516: So, based on my testing this part of the code (i.e. changing subdivision) is still somewhat computationally intensive. This could be fixed if the subdivision count was set to a static value. To address the main issue that the LOD system was made to fix, the subdivision count would have to be low, but I believe that it's possible not to sacrifice visual quality in this case. The main graphical issue that would arise is unsmooth lighting, but if the lighting model was switched to `diffuse_lambert_wrap` in the shader, the lighting would look smooth (based on my observations) without requiring high poly count.

src/microbe_stage/Membrane.cs
hhyyrylainen: I'll start doing work on this branch to finish off these comments that have not been addressed (as far as I can tell only master has been merged into this branch recently which is useless for everything else except spamming me for emails, a branch that doesn't have all review comments solved *does not need to be in good merge state* as it cannot be merged anyway due to the review comments).

hhyyrylainen (approved):
I did as much refactoring as I could without really understanding the core math in the riple cs file (or the shader math really). There's still quite a bit of room to improvement probably and skipping more duplicate data writes but this is as much as I could do as not the author of the code.

src/microbe_stage/MembraneWaterRipple.cs
hhyyrylainen: I think this feature works well enough to merge, but I'd still like to see these few technical comments to be addressed (and the above diffuse_lambert comment). I don't know enough to really know what to do. I did some refactoring so that we can merge this PR today well in time before my vacation starts as that would have really delayed this PR *again*.

PR #5998, opened 2025-03-09
Update setup instructions

skyace65:
**Update Outdated information**

Does several things for the setup instructions md file:
- Change "mono version" to ".net version". Yes the downloaded executable is still named mono, but the website calls it .net. I feel like this change can make things a bit less confusing
- Remove note about if the project was previously opened in Godot 3. Given how long Thrive has been on Godot 4 I don't think this needs to be included anymore
- Replace a comma with the word "the", the comma just doesn't make that sentence read right
- Remove monodevelop from the list of IDE's as well as the part saying not to install it and mono. It does not work with up to date versions of Godot anymore, we actually just removed it from the engine docs. I also wouldn't expect anyone to try and download it since its been retired for 4 years now.

**Related Issues**

None

### DIFF
--- a/doc/setup_instructions.md
+++ b/doc/setup_instructions.md
@@ -21,7 +21,7 @@ Thank you!
 Prerequisites
 =============
 
-Godot .Net version
+Godot .NET version
 ------------------
 
 The currently used Godot version is __4.3 .NET__. The regular version
--- a/doc/setup_instructions.md
+++ b/doc/setup_instructions.md
@@ -21,7 +21,7 @@ Thank you!
 Prerequisites
 =============
 
-Godot mono version
+Godot .Net version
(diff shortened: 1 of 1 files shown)


doc/setup_instructions.md
```diff
@@ -21,7 +21,7 @@ Thank you!
 Prerequisites
 =============
 
-Godot mono version
+Godot .Net version
```
hhyyrylainen, line 24:
```suggestion
Godot .NET version
```

.NET is written uppercase (on the Godot website) so this is more consistent with the wording.

hhyyrylainen (approved):
Seems like good changes. Thanks. I made .NET uppercase in the title to be consistent with how that is usually written.

PR #6013, opened 2025-03-12
Fix patch map lag

### LINKED ISSUE #5975: Improve patch map rendering performance with large maps
With the new option to have larger worlds, the map performance on my computer falls to 30 FPS when zooming out. So that isn't exactly great so I think the performance needs some improving.

Interestingly I found out that if I click on something on the map (so that connections are redrawn) that increases the performance a ton, so that's pretty weird.

Image

After clicking:
Image

Patryk26g:

Closes #5975

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/PatchMapDrawer.cs
+++ b/src/microbe_stage/editor/PatchMapDrawer.cs
@@ -188,7 +188,7 @@ public partial class PatchMapDrawer : Control
         if (connections.Count == 0)
         {
             // Clear existing connection lines
-            lineContainer.QueueFreeChildren();
+            lineContainer.QueueFreeChildren(false);
 
             CreateRegionLinks();
             RebuildRegionConnections();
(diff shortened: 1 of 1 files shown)


HyperbolicHadron:
This appears to have completely fixed the issue when tested on my main computer

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -188,14 +188,14 @@ public override void _Draw()
         if (connections.Count == 0)
         {
             // Clear existing connection lines
-            lineContainer.FreeChildren();
+            lineContainer.QueueFreeChildren();
```
hhyyrylainen, line 191: Does this need to detach the nodes here? That I think costs more performance than passing in false here.

hhyyrylainen (approved):
This improves the performance quite a bit. I tested the normal map view and also the patch extinction view and couldn't see any bugs that these changes would have introduced.

src/microbe_stage/editor/PatchMapDrawer.cs
Patryk26g: I forgot to add `false` :sob:

hhyyrylainen: Seems like a simple enough change so I'll merge this without playtesting again as if anything the performance should be even better with this.

PR #6022, opened 2025-03-13
Fix some of the new currents' flaws

dligr:

- Fixes the grid issue (by increasing noise resolution* and making noise seamless)
- Improves visibility in bright patches
- Makes strong currents more common

*This increased the memory usage to 1 MB per one 3D noise texture, adding up to 4 MB total

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/FluidCurrentDisplay.cs
+++ b/src/microbe_stage/FluidCurrentDisplay.cs
@@ -11,6 +11,7 @@ public partial class FluidCurrentDisplay : GpuParticles3D
     private readonly StringName speedParameterName = new("speed");
     private readonly StringName chaoticnessParameterName = new("chaoticness");
     private readonly StringName scaleParameterName = new("scale");
+    private readonly StringName brightnessParameterName = new("brightness");
 
 #pragma warning disable CA2213
     private ShaderMaterial material = null!;
@@ -76,6 +77,8 @@ public partial class FluidCurrentDisplay : GpuParticles3D
         material.SetShaderParameter(chaoticnessParameterName, biome.WaterCurrentChaoticness);
         material.SetShaderParameter(scaleParameterName, biome.WaterCurrentScale);
 
+        material.SetShaderParameter(brightnessParameterName, biome.CompoundCloudBrightness);
+
         Amount = biome.WaterCurrentParticleCount;
     }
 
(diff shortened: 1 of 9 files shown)


src/microbe_stage/FluidCurrentDisplay.cs
```diff
@@ -76,6 +77,8 @@ public void ApplyBiome(Biome biome)
         material.SetShaderParameter(chaoticnessParameterName, biome.WaterCurrentChaoticness);
         material.SetShaderParameter(scaleParameterName, biome.WaterCurrentScale);
 
+        material.SetShaderParameter(brightnessParameterName, biome.CompoundCloudBrightness);
```
dligr, line 80: To increase the particles' visibility, I made their brightness scale with that of the clouds. This works well, but the parameter name is now not entirely correct. Should it be renamed to something like "SpriteBrightness" or "GraphicsBrightness"?

HyperbolicHadron:
This seems to improve upon the currents look even more. This fix would definitely be worth applying now.

However, it is still possible to notice the boundaries if you look for them. I have noticed it more in the hydrothermal vents than the warm pool interestingly. It seems that the compounds starts to collect at the borders of the chunk, and it takes a bit for the compounds to diffuse across the border which makes the borders apparent. (Even though the currents themselves are nicely continuous across the borders, the compound diffusion seems to have a brief halt at the border, introducing the visual discontinuity)

src/microbe_stage/FluidCurrentDisplay.cs
hhyyrylainen: I think the name is fine for now. I'm not particularly keen on those new names as it doesn't apply to all graphics so clearly graphics brightness is wrong and we don't really use sprites but some stuff that are kind of like sprites (like the environmental particles) aren't affected so I think that would also be an unclear name.

The clearest solution I think would be to just have a new variable for "CurrentParticleBrightness" separately.

hhyyrylainen (approved):
I think this is a good big improvement, but I also noticed still those slight cloud artifacts present. Though some of those are clearly caused by the old bug about edge lines existing in the clouds.

I also noticed that the noise generation now seems to take a bit of time when starting the microbe stage as presumably the bigger noise textures take longer to generate.

I'll merge this now to put this in the BOTD, but hopefully further improvements can still be made to the currents.

PR #6029, opened 2025-03-18
Meteor impact event

Patryk26g:

Adds 6 meteor impact event variations.

**[Global glaciation event](https://github.com/Revolutionary-Games/Thrive/pull/5974) need to be merged first!**

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I marked this as a draft as this depends on the glaciation event being merged first.

Also you'll unfortunately get a lot of merge conflicts due to the squash merge that will happen to master from the glaciation once that PR is done.

hhyyrylainen:
Is this now feature complete with that other PR being merged and I saw this got quite many "fixes" commits?

Patryk26g:
> Is this now feature complete with that other PR being merged and I saw this got quite many "fixes" commits?

Yes, It is complete

locale/en.po
```diff
@@ -3287,6 +3324,15 @@ msgstr "Turns [thrive:compound type=\"glucose\"][/thrive:compound] into [thrive:
...
 msgstr "Visit our Threads profile"
 
+msgid "METEOR_IMPACT_EVENT"
+msgstr "Meteor impact event"
+
+msgid "METEOR_STRIKE_START_EVENT_LOG_PLURAL"
+msgstr "Meteor impact event affected {0} patch"
+
+msgid "METEOR_STRIKE_START_EVENT_LOG_SINGULAR"
+msgstr "Meteor impact event affected {0} patches"
```
hhyyrylainen, line 3334: Aren't these reversed?

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -44,6 +44,24 @@ public partial class PatchMapNode : MarginContainer
     [Export]
     private TextureRect glaciationEventIndicator = null!;
 
+    [Export]
+    private TextureRect meteorPlainEventIndicator = null!;
```
hhyyrylainen, line 48: I think there's way too many static indicator textures...

I think this system should be refactored to dynamically load the textures *as needed* so that each patch map node doesn't all keep references to 8+ textures on them that are rarely used.

src/saving/serializers/RegistryTypeStringConverter.cs
```diff
@@ -201,6 +201,14 @@ public class EnzymeStringConverter : RegistryTypeStringSingleTypeConverter<Enzym
     protected override string TypeName => "enzyme";
 }
 
+/// <summary>
+///   Specific converter for <see cref="Meteor"/>
+/// </summary>
+public class MeteorStringConverter : RegistryTypeStringSingleTypeConverter<Enzyme>
```
hhyyrylainen, line 207:
```suggestion
public class MeteorStringConverter : RegistryTypeStringSingleTypeConverter<Meteor>
```

src/saving/serializers/RegistryTypeStringConverter.cs
```diff
@@ -201,6 +201,14 @@ public class EnzymeStringConverter : RegistryTypeStringSingleTypeConverter<Enzym
...
+{
+    protected override string TypeName => "meteor";
```
hhyyrylainen, line 209: It seems you didn't add this to `SupportedRegistryTypes` so this converted to me seems like it isn't hooked up correctly.

src/microbe_stage/Meteor.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    public string InternalName { get; set; } = null!;
+
+    public void Check(string name)
+    {
+    }
+
+    public void ApplyTranslations()
+    {
+        TranslationHelper.ApplyTranslations(this);
```
hhyyrylainen, line 25: This doesn't seem to have a name (just an internal name) so it looks to me like this does nothing?

src/microbe_stage/Meteor.cs
```diff
@@ -0,0 +1,32 @@
...
+/// <summary>
+///   Contains definitions of meteors for meteor impact event
+/// </summary>
+public class Meteor : IRegistryType
+{
+    public WorldEffectVisuals VisualEffect;
+
+    public List<string> Chunks = new();
+
+    public Dictionary<Compound, double> Compounds = new();
```
hhyyrylainen, line 13:
```suggestion
    public Dictionary<Compound, float> Compounds = new();
```

In Thrive compound amounts are always stored as floats.

src/microbe_stage/Meteor.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    public double Probability;
+
+    public string InternalName { get; set; } = null!;
+
+    public void Check(string name)
+    {
```
hhyyrylainen, line 20: What if someone puts in a negative probability, is that valid? If not at least that one thing would warrant being here in the check method.

simulation_parameters/microbe_stage/biomes.json
```diff
@@ -3437,5 +3437,391 @@
         }
       }
     }
+  },
+  "patch_event_template_biome": {
```
hhyyrylainen, line 3441: One tiny naming thing: I've been trying to move us towards consistent camelCase naming for internal names of stuff, so could you rename this and the meteor internal names to be like `patchEventTemplateBiome`?

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -214,6 +218,12 @@ private void LogEvent(Patch patch, double totalTimePassed)
         patch.LogEvent(new LocalizedString("GLOBAL_GLACIATION_EVENT"),
             true, true, "GlobalGlaciationEvent.svg");
 
+        if (patch.Visibility == MapElementVisibility.Shown)
+        {
+            targetWorld.LogEvent(new LocalizedString("GLOBAL_GLACIATION_EVENT_LOG", patch.Name),
+                true, false, "GlobalGlaciationEvent.svg");
```
hhyyrylainen, line 224: Doesn't this add back the spam of glaciation messages about all patches?

src/general/world_effects/patch_events/MeteorImpactEvent.cs
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    private int GetRandomElementByProbability(List<double> chances, double probability)
+    {
+        double totalSum = chances.Sum();
+        if (Math.Abs(totalSum - 1.0) > 1e-6)
+        {
+            GD.PrintErr($"Probability sum mismatch: {totalSum}. Expected: 1.0");
+            return 0;
+        }
```
hhyyrylainen, line 87: Isn't this static info? This should be checked by SimulationParameters when loading, not during gameplay.

src/general/world_effects/patch_events/MeteorImpactEvent.cs
```diff
@@ -0,0 +1,308 @@
...
+
+        LogBeginningOfMeteorStrike();
+    }
+
+    private void ChooseMeteorType()
+    {
+        var meteors = SimulationParameters.Instance.GetAllMeteors().ToList();
+        var index = GetRandomElementByProbability(meteors.Select(meteor => meteor.Probability).ToList(),
+            random.NextDouble());
```
hhyyrylainen, line 76: This allocates a temporary list. The randomness pick should probably be put into the SimulationParamaters where it can precalculate the list to make the random picking method not allocate memory.

src/general/world_effects/patch_events/MeteorImpactEvent.cs
```diff
@@ -0,0 +1,308 @@
...
+
+        throw new ArgumentException("Chances list is empty");
+    }
+
+    private void ChooseAffectedPatches()
+    {
+        var impactSize = random.Next(0, 3);
+        var surfacePatches = targetWorld.Map.Patches.Values.Where(IsSurfacePatch).ToList();
+        var selectedPatch = surfacePatches[random.Next(surfacePatches.Count)];
+        var adjacentRegion = selectedPatch.Region.Adjacent.ToList()[random.Next(selectedPatch.Region.Adjacent.Count)];
```
hhyyrylainen, line 105: A bunch of LINQ usage that should be avoided...

src/general/world_effects/patch_events/MeteorImpactEvent.cs
```diff
@@ -0,0 +1,308 @@
...
+    {
+        bool hasSunlight = patch.Biome.ChangeableCompounds.TryGetValue(Compound.Sunlight, out var currentSunlight);
+
+        if (!hasSunlight)
+        {
+            GD.PrintErr("Meteor impact event encountered patch with unexpectedly no sunlight");
+            return;
+        }
+
+        currentSunlight.Ambient *= Constants.METEOR_IMPACT_SUNLIGHT_MULTIPLICATION;
```
hhyyrylainen, line 182: I see this is how you handle multiple events overlapping. I suppose mathwise this does check out.

But is the gameplay impact of for example glaciation and a meteor impact good at the same time? Reducing sunlight even below 50% is probably very punishing.

src/microbe_stage/editor/PatchMapNode.cs
Patryk26g: So would you prefer to have it load textures for ALL events? Like for example have the texture as a separate scene?

hhyyrylainen: No, I'd prefer a system where none of the icons exist in the scene, just the container. And then during runtime when the list of active events is updated the code will create and destroy the texture rects as required.

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -276,29 +279,32 @@ public void UpdateVisibility()
...
         {
-            switch (list[i])
+            var eventIcon = eventIconScene.Instantiate<PatchMapEventIcon>();
+
+            if (!WorldEffectVisuals.EventsTooltips.TryGetValue(effectType, out var tooltipText))
+            {
+                throw new Exception($"Missing tooltip for {effectType}");
+            }
+
+            eventIcon.TooltipText = Localization.Translate(tooltipText);
```
Patryk26g, line 298: I can't force the localization tool to include these tooltip texts :/

hhyyrylainen: The translation extraction doesn't run the code. It uses just a syntax tree parser, so it can only see `.Translate("STRING_LITERAL")` style patterns. The tooltip text here is a variable so the translation extraction has no clue what it should extract here. You need to put the Localization.Translate call to where the actual string literal is (I'm assuming, I haven't read the code above this yet).

src/microbe_stage/editor/PatchMapNode.cs
```diff
@@ -276,29 +279,32 @@ public void UpdateVisibility()
         }
     }
 
-    public void ShowEventVisuals(IReadOnlyList<WorldEffectVisuals> list)
+    public void ShowEventVisuals(IReadOnlyList<WorldEffectVisuals.WorldEffectTypes> list)
     {
-        eruptionEventIndicator.Visible = false;
-        glaciationEventIndicator.Visible = false;
+        if (eventIconsContainer == null)
```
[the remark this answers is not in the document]
hhyyrylainen: You should swap from the NodePath approach to the new direct node [Export] as those are initialized *way* earlier. This old approach only initializes on _Ready, which means adding a bunch of extra checking code to ensure node operations don't run before _Ready is called. (though there are still situations where the old "skip if too early" code is needed, but from a cursory look at this specific situation that isn't necessary but I'm mentioning this anyway to give you the complete picture which you may need in doing future changes)

simulation_parameters/Constants.cs
```diff
@@ -1769,8 +1773,8 @@ public static class Constants
     public const float MAX_AUTO_EVO_STRENGTH_MULTIPLIER = 1.0f;
 
     // Constants for the procedural patch map
-    public const float PATCH_NODE_RECT_LENGTH = 64.0f;
-    public const float PATCH_AND_REGION_MARGIN = 20.0f;
+    public const float PATCH_NODE_RECT_LENGTH = 80.0f;
+    public const float PATCH_AND_REGION_MARGIN = 14.0f;
```
hhyyrylainen, line 1777: How big of a visual change this is to the map?

simulation_parameters/SimulationParameters.cs
```diff
@@ -864,4 +886,9 @@ private List<Enzyme> ComputeHydrolyticEnzymes()
     {
         return enzymes.Where(e => e.Value.Property == Enzyme.EnzymeProperty.Hydrolytic).Select(e => e.Value).ToList();
     }
+
+    private List<double> ComputeMeteorChances()
+    {
+        return meteors.Values.Select(meteor => meteor.Probability).ToList();
```
hhyyrylainen, line 892:
```suggestion
        return meteors.Values.Select(m => m.Probability).ToList();
```

src/general/world_effects/WorldEffectVisuals.cs
```diff
@@ -1,10 +1,45 @@
...
+        [WorldEffectTypes.MeteorIronImpact] = "res://assets/textures/gui/bevel/MeteorIron.svg",
+        [WorldEffectTypes.MeteorPhosphatesImpact] = "res://assets/textures/gui/bevel/MeteorPhosphates.svg",
+        [WorldEffectTypes.MeteorRadioactiveImpact] = "res://assets/textures/gui/bevel/MeteorRadioavtive.svg",
+        [WorldEffectTypes.MeteorGlucoseImpact] = "res://assets/textures/gui/bevel/MeteorGlucose.svg",
+        [WorldEffectTypes.MeteorSulfurImpact] = "res://assets/textures/gui/bevel/MeteorSulfur.svg",
+    };
+
+    public static readonly Dictionary<WorldEffectTypes, string> EventsTooltips = new()
+    {
+        [WorldEffectTypes.UnderwaterVentEruption] = "EVENT_ERUPTION_TOOLTIP",
```
hhyyrylainen, line 19: This kind of pattern can't be picked up by the localization tool. So something else is needed. It almost seems like it'd be worth it to have a separate world events JSON file that would specify the tooltips, icon paths, *and* the meteors. That way this PR would add just one new JSON info file (which could be used for environmental events in general).

src/microbe_stage/Patch.cs
```diff
@@ -41,7 +41,7 @@ public class Patch
     ///   The current effects on patch node (shown in the patch map)
     /// </summary>
     [JsonProperty]
-    private readonly List<WorldEffectVisuals> activeWorldEffectVisuals = new();
+    private readonly List<WorldEffectVisuals.WorldEffectTypes> activeWorldEffectVisuals = new();
```
hhyyrylainen, line 44: Seeing as this nested enum is used all over, I think it would be better to not have the enum be nested....

Though I'm not sure why there needs to be a Visuals enum and then a type enum inside that. To me that seems the opposite, the visual (as seen on line 605) seems to still be the primary data model...

src/microbe_stage/editor/PatchMapEventIcon.cs
```diff
@@ -0,0 +1,8 @@
+﻿using Godot;
+
+/// <summary>
+///   Texture for patch event
+/// </summary>
+public partial class PatchMapEventIcon : TextureRect
```
hhyyrylainen, line 6: Why is there a blank script here? This will unnecessarily eat up performance...

src/saving/serializers/RegistryTypeStringConverter.cs
hhyyrylainen: This is marked as resolved but I still do not see the change that should be in the diff earlier in this file...

hhyyrylainen (changes requested):

src/microbe_stage/editor/PatchMapEventIcon.cs
Patryk26g: Itsn't needed here: `var eventIcon = eventIconScene.Instantiate<PatchMapEventIcon>();`?

src/general/world_effects/WorldEffectVisuals.cs
Patryk26g: What if I just use Localization.Translate() here instead?

src/saving/serializers/RegistryTypeStringConverter.cs
Patryk26g: I removed it completely

src/general/world_effects/WorldEffectVisuals.cs
hhyyrylainen: When will you trigger re-translation then? If the user changes language after the game is started then the old text will stay here. I suppose having `LocalizedString` here could be a viable alternative.

simulation_parameters/Constants.cs
```diff
@@ -1772,6 +1772,8 @@ public static class Constants
     public const float MIN_AUTO_EVO_STRENGTH_MULTIPLIER = 0.01f;
     public const float MAX_AUTO_EVO_STRENGTH_MULTIPLIER = 1.0f;
 
+    public const float HALF_BLINK_INTERVAL = 0.5f;
```
hhyyrylainen, line 1786: I think this constant name is not clear enough what this is for...

Patryk26g: well I just moved this variable because it was said so in the TODO. I don't what it was for, I just kept the name

hhyyrylainen: Well the name should be prefixed with the class. I assume this is like from the patch node class? So the name should be something like `CURRENT_MAP_PATCH_INDICATOR_HALF_BLINK_INTERVAL`

hhyyrylainen (approved):
I did quite a bit of code cleanup and refactoring so the code should be ready now. My changes shouldn't affect the actual functionality, though I only did quick testing in game to make sure the event still functions at least.

hhyyrylainen:
I ended up having to put quite many general fixes here (due to the other commits to master today) so sorry for slightly cluttering up this PR, but now the checks hopefully will pass and I can merge.

PR #6043, opened 2025-03-27
#5023 spinner shader

### LINKED ISSUE #5023: Implement 2D spinner shader as alternative to the removed pivots in Godot 4
A "hack" fix for the process panel spinner and loading screen spinner is here: https://github.com/Revolutionary-Games/Thrive/pull/5022

The only way I think to really cleanly try to do this, would be to have a custom shader that rotates the UVs around to cause the visuals of the node to spin without hacking around with plain control nodes.

Dynline:

Adds a canvas item shader to the icon used in the ChemicalEquation.tscn that makes it rotate.

**Related Issues**

closes #5023 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/shaders/Spinner.gdshader
+++ b/shaders/Spinner.gdshader
@@ -5,8 +5,8 @@ Vertex is used for better performance, however it requires the size property val
 Fragment is less performant but can be used to avoid passing the size in case the assets using this shader become too many.
 */
 
-
-uniform float rotationSpeed = 1.0; //Use negative numbers to rotate counterclockwise
+//Use negative numbers to rotate counterclockwise
+uniform float rotationSpeed = 1.0;
 
 //This is done to ensure this shader will work with assets that have been scaled down in engine. Comment to use fragment
 uniform float sizeX = 0;
--- a/shaders/Spinner.gdshader
+++ b/shaders/Spinner.gdshader
@@ -1,0 +1,46 @@
+shader_type canvas_item;
+
+/*
+Vertex is used for better performance, however it requires the size property values to be passed again through shader parameters.
+Fragment is less performant but can be used to avoid passing the size in case the assets using this shader become too many.
+*/
+
+
+uniform float rotationSpeed = 1.0; //Use negative numbers to rotate counterclockwise
+
+//This is done to ensure this shader will work with assets that have been scaled down in engine. Comment to use fragment
+uniform float sizeX = 0;
+uniform float sizeY = 0;
+
+//Matrix to calculate the rotation
+vec2 rotate (vec2 uv, vec2 pivot, float angle) {
+    mat2 rotationMatrix = mat2(vec2(cos(angle),-sin(angle)),
+                                vec2(sin(angle),cos(angle)));
+    
+    uv -= pivot;
+    uv = uv * rotationMatrix;
+    uv += pivot;
+    return uv;
+
+}
+
+//Comment to use fragment
+void vertex() {
+    //Mke a vector2 with the passed size values
+    vec2 size = vec2(sizeX, sizeY);
+    
+    //Pass rotation speed
+    float time = TIME * rotationSpeed;
+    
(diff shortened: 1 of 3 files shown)


hhyyrylainen:
Thanks for taking the time to help out with Thrive development. 
I noticed the CLA check is failing. Was there a problem with signing the CLA or did you just not get around to it yet? I ask because there's been some recent problems with the CLA signature *detection* not working correctly.

Dynline:
> Thanks for taking the time to help out with Thrive development. I noticed the CLA check is failing. Was there a problem with signing the CLA or did you just not get around to it yet? I ask because there's been some recent problems with the CLA signature _detection_ not working correctly.

Hi! Yes, I'm having problems signing it, I actually planned to ask about it in the forum. I tried a few times, but I get this error:

Error, server responded with: {"type":"https://tools.ietf.org/html/rfc9110#section-15.6.1","title":"An error occurred while processing your request.","status":500,"detail":"Failed to upload signed document to remote storage. Please try again.","traceId":"00-071bad2311f82778246098db00c7a441-f9e04ec63450d331-00"}, InternalServerError

Same error today.

Dynline:
> I see. I checked the server logs and yeah there's a problem uploading the signatures to storage. I did a few small changes to the process that maybe did something ([Revolutionary-Games/RevolutionaryWebApp@0e81629](https://github.com/Revolutionary-Games/RevolutionaryWebApp/commit/0e8162978c04d0cfb6252a3c8ead9b7badace5b3)) and restarted the server. I just tried signing the CLA again myself and that seemed to work now.
> 
> So could you please retry the CLA signing now? It should hopefully work without issue this time.

Thanks for checking that! It works now.

shaders/Spinner.gdshader
hhyyrylainen, line 1: I think for clarity it would be better to split this file into two: the more hands off pixel variant and the current vertex variant that needs explicit size.

Also could this shader be also applied to the loading screen? As I think that's the one other place that Thrive uses a spinner that this PR doesn't modify yet.

Dynline: The shader is now splitted into two.

Also yes ! Just added it to the loading screen icon. It replaces the parts of code that were making it spin before.

I also wanted to ask about the image size of some icons, for example the one used for the chemical reaction is 800px scaled down to 32. Usually 800 for an icon is pretty big, do you need them to be this size for future parts of the game?

hhyyrylainen: Thanks. We've had issues in the past where there's an icon that's 24x24 but we need it higher resolution to not look blurry on a 4K screen. And as most artists on the team become inactive after a while, there's no way for us to get updated versions for the icons. So it is better to get such a big resolution that there's no way we'd ever need to redo the icon (rather than having icons that are 5 years old now but need to be redone for higher resolution). 

Though, we are switching our icons over to SVG, but obviously that would be a ton of work to redo all old icons so icons are only slowly being replaced. SVGs don't have the same resolution issues so those are imported at smaller scale into the game (as it is very easy to just change the import scale later if required).

Dynline: I see! thank you for letting me know that. I wanted to make sure before making any unwanted changes, as I plan to contribute some more shader or graphics related things.

hhyyrylainen (changes requested):
Testing in game, the loading screen spinner is good. So the shader works. However, the process speed icon (in ChemicalEquation) has two issues:
- There is still the old code that tries to rotate things so when I removed the pivot offset (which should be unnecessary with these changes) there's now a bug where the texture rotates around its top left as well. So there's code that would need to be removed.
- But also the process speed spinner *must pause* when the game is paused and this shader approach doesn't do that. So I think that what actually needs to be done is one more shader variant that allows *manual* control from the C# on the current rotation value. That way the C# code can still control the rotation amount and pause correctly but use the shader to apply the rotation.

I committed a few slight style fixes also to your branch, hope you don't mind.

PR #6047, opened 2025-03-29
Update MulticellularGrowthSystem not to use LINQ Reverse

### LINKED ISSUE #4879: Update MulticellularGrowthSystem to not use LINQ Reverse for growth
switch to the new approach in ProcessBaseReproductionCost that avoids memory allocations. When doing the conversion I was too lazy to update this later part of the game, but this should be very easy to follow the same architecture and update this multicellular specific variant of the code as well.

dligr:

Updates MulticellularGrowthSystem not to use LINQ reverse. Instead, the compound order is now reversed by using a temporary list.

**Related Issues**

Fixes #4879

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/utils/ListUtils.cs
+++ b/src/general/utils/ListUtils.cs
@@ -100,7 +100,7 @@ public static class ListUtils
     {
         var comparer = EqualityComparer<TKey>.Default;
 
-        for (int i = 0; i < list.Count; i++)
+        for (int i = 0; i < list.Count; ++i)
         {
             if (comparer.Equals(key, list[i].Key))
                 return i;
--- a/src/general/utils/ListUtils.cs
+++ b/src/general/utils/ListUtils.cs
@@ -87,6 +87,27 @@ public static class ListUtils
     }
 
     /// <summary>
+    ///   Finds the index of the first pair, the first element of which is equl to <paramref name="key"/>
+    /// </summary>
+    /// <typeparam name="TKey">Type of the pairs' first element</typeparam>
+    /// <typeparam name="TValue">Type of the pairs' second element</typeparam>
+    /// <param name="list">The list to seartch through</param>
+    /// <param name="key">The first element of the pair that needs to be found</param>
+    /// <returns>The index of the found element or -1 if not found</returns>
+    public static int FindIndexByKey<TKey, TValue>(this IReadOnlyList<(TKey Key, TValue Value)> list, TKey key)
+    {
+        var comparer = EqualityComparer<TKey>.Default;
+
+        for (int i = 0; i < list.Count; i++)
+        {
+            if (comparer.Equals(key, list[i].Key))
+                return i;
+        }
+
+        return -1;
+    }
+
+    /// <summary>
     ///   Removes an item from a list at index without preserving the list order, this should be faster than normal
     ///   list remove that preserves order
     /// </summary>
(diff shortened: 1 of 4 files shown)


src/multicellular_stage/systems/MulticellularGrowthSystem.cs
```diff
@@ -197,57 +197,64 @@ private void HandleMulticellularReproduction(ref MulticellularGrowth multicellul
...
-        foreach (var entry in microbeStatus.ConsumeReproductionCompoundsReverse ?
-                     multicellularGrowth.CompoundsNeededForNextCell.Reverse() :
-                     multicellularGrowth.CompoundsNeededForNextCell)
+        lock (temporaryWorkData)
         {
-            var amountNeeded = entry.Value;
+            temporaryWorkData.Clear();
+            foreach (var entry in multicellularGrowth.CompoundsNeededForNextCell)
+            {
+                temporaryWorkData.Add(entry.Key);
```
hhyyrylainen, line 205: Sorry if I'm missing something, but this change seems very strange to me. The point about the TODO is that the list duplication each update should be removed. But here the list is copied each update anyway, so this just saves on one list object allocation. And also introduces a lock meaning that only one multicellular entity can update its growth state at once. So I'm left wondering if there is some other way to do it... maybe even changing CompoundsNeededForNextCell to be a List to allow more efficient handling of it (it should be pretty minimal cases, I think, that will need a linear scan over the list to find wanted values).

hhyyrylainen (changes requested):

src/multicellular_stage/systems/MulticellularGrowthSystem.cs
dligr: I tried making `CompoundsNeededForNextCell` a List, and this approach seems to work. The only place that could suffer from performance loss is `CalculateTotalComposition`, which has to repeatedly use linear search (using a lambda).

I didn't update a similar function with list reversal in `MicrobeReproductionSystem`, but I think it could use a similar approach, and so there would be no need for temporary storage.

src/general/utils/ListUtils.cs
```diff
@@ -86,6 +86,27 @@ public static int FindIndex<T>(this IReadOnlyList<T> list, Predicate<T> match)
...
 
+    /// <summary>
+    ///   Finds the index of the first pair, the first element of which is equl to <paramref name="key"/>
+    /// </summary>
+    /// <typeparam name="TKey">Type of the pairs' first element</typeparam>
+    /// <typeparam name="TValue">Type of the pairs' second element</typeparam>
+    /// <param name="list">The list to seartch through</param>
+    /// <param name="key">The first element of the pair that needs to be found</param>
+    /// <returns>The index of the found element or -1 if not found</returns>
+    public static int FindIndexByKey<TKey, TValue>(this IReadOnlyList<(TKey Key, TValue Value)> list, TKey key)
```
hhyyrylainen, line 97: AggressiveInlining attribute would probably be nice to have here.

hhyyrylainen (approved):
Code looks perfect to me now and I also did a quick test in game to confirm multicellular growth still seems to work.

hhyyrylainen:
Just need to fix that one small format problem and this is good to merge.

PR #6050, opened 2025-03-31
Some slight tutorial size and wording fixes

hhyyrylainen:

To correct the few problems we noticed in the tutorial with the live play session

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


CI09 (approved):
Tested except for nucleus part but looks good to me.

I'm kinda late with tutorial rework, but my feedback would be generally positive, I like the way stuff is told now. Though, sometimes tutorial says to do certain action to close tutorial, but there is X button anyways. That's a minor thing that doesn't even have to be removed in my opinion.
2nd and 3rd editor session you still can't go back to previous tabs, which could be potentially annoying if someones moves their species out of starting region and later  decides it was a bad idea.
And finally, one small thing from my side, would be, 3rd editor session I think, when tutorial says about specializing, not getting big asap if it's not necessary, etc. when it mentions that photosynthesis is easier but running after other microbes is more fun, I would add something like "photosynthesis may settle you down for later stages/for good" or "the latter would serve as a keystone to later motile lifestyle" or something like that if you know what I mean.

That's it from my feedback I think, implement what you want to implement.

Or you can give me green light for some fix if you have other plans; it doesn't seem too hard.

hhyyrylainen:
> 2nd and 3rd editor session you still can't go back to previous tabs, which could be potentially annoying if someones moves their species out of starting region and later decides it was a bad idea.

It should be the case that once you finish the tutorial, it will re-enable the tabs. So you should be able to go back to the map tab once you're done in the editor tab. Could you double check, because if this is the case then a regression has slipped in?

>  "photosynthesis may settle you down for later stages/for good" or "the latter would serve as a keystone to later motile lifestyle" or something like that if you know what I mean.

I think that's maybe a bit too much extra text to add. Isn't that text in already one of the longest tutorials already?. The photosynthesis is meant to be like an early game clue what the player can try if they simply can't make a successful cell. So telling like how to do a full microbe stage playthrough feels like maybe a bit too much guidance? Like it takes maybe too much discovery out of the game?

> but there is X button anyways. 

I mean if the player absolutely wants to skip some tutorials, they can. Only a few of the tutorials are truly forced with the X disabled. This is an intentional choice that's carried over from the previous tutorial version.

---

Apparently I forgot to merge this today so I guess I'll do it now.

> Or you can give me green light for some fix if you have other plans; it doesn't seem too hard.

If you want you can do some further tutorial change suggestions but you'll need to handle everything related to the change.

CI09:
3rd session brings up patch map for the first time, but it's at this point tutorial ends in the entire session without bringing back tabs, so at cell editor there is no trigger to bring tabs back either and that locks ability to go back to patch map in that session.

hhyyrylainen:
> 3rd session brings up patch map for the first time, but it's at this point tutorial ends in the entire session without bringing back tabs, so at cell editor there is no trigger to bring tabs back either and that locks ability to go back to patch map in that session.

I think you probably had a tutorial already marked as completed, and that tutorial would have been the one to trigger the tabs to come back.

You are supposed to have a tutorial in the cell editor which when over shows the tabs. I didn't consider all of the potential cases where the editor can be in an unexpected state due to partial tutorial completion from other playthroughs. So I kind of consider those bugs fine as long as they don't softlock the game, they are just a very minor inconvenience. Though if someone wants to specifically add cases to handle those, I'll accept that as a bugfix PR.

PR #6066, opened 2025-04-06
Fix visual artifacts around toxins when chromatic abberation is enabled

dligr:

Fixes the weird outlines near toxins (and some other objects), which appear when chromatic abberation is enabled.
Apparently, the cause of their appearance was a wrong random number generation method, which used screen texture as a noise texture. This PR uses a different RNG function ([taken from here](https://www.shadertoy.com/view/XssGz8)).

This changes the chromatic aberration visuals, adding flickering noise. If this is undesirable, randomness could be removed and the shader would retain its former look (just without artifacts).

https://github.com/user-attachments/assets/0d792f9a-5172-4bff-bacf-75781ccbb8ff

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/shaders/Chromatic.gdshader
+++ b/shaders/Chromatic.gdshader
@@ -123,11 +123,6 @@ vec3 render(vec2 uv, sampler2D tex)
     return srgb2lin(texture(tex, uv).rgb);
 }
 
-float nrand( vec2 n )
-{
-    return fract(sin(dot(n.xy, vec2(12.9898, 78.233))) * 43758.5453);
-}
-
 void fragment()
 {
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
I think it probably would be better without the random noise as I think the goal of the chromatic aberration is to be just a slight effect in Thrive, but that noise makes it really noticeable.

hhyyrylainen (approved):
I think this is a big improvement to the shader. I can't find an open issue about it but I've noticed the clouds not blending well near the screen edges and I think this is what finally fixes that.

I'll merge this normally in order to keep a historical record of the noise intermediate step that was not kept in the final code.

PR #6074, opened 2025-04-09
Fixed second ground plane size in industrial stage and eliminated the gap

### LINKED ISSUE #6057: The extra ground plane is now sized incorrectly in the prototypes
which leaves a gap like this:

Image

Should be very easy to change the size of that in the Godot editor (or if specified in C# then it's just editing a couple of numbers)

RankornCZ:

I fixed the bug 6057 . The ground planes are now the same size and are next to each other as they are in Society stage.

**Related Issues**

fixes #6057

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/industrial_stage/IndustrialStage.tscn
```diff
@@ -20,7 +20,7 @@ uv1_scale = Vector3(105, 105, 105)
 
 [sub_resource type="PlaneMesh" id="2"]
 material = SubResource("3")
-size = Vector2(1200, 800)
+size = Vector2(400, 400)
```
hhyyrylainen, line 23: Doesn't this reduce the size of the area a ton? That seems kind of an unintended effect... I'd prefer to keep the ground planes at least as big as they were before, and just position or increase the size of the second one to make sure there's no gap.

RankornCZ: I see then i will return it to the original size and without that gap. And i will try to be more descriptive make other nessessary changes in the PR.

PR #6089, opened 2025-04-19
Fix a crash when entering an editor scene directly from Godot

dligr:

Fixes a crash that happened when an editor scene was entered directly from the Godot Editor. It happened because a field initialization in `TolerancesEditorSubComponent` depended on `SimulationParameters` being initialized, which only happens when the game is started from the main menu.

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (approved):
This change makes sense to me. And I'll assume this fixes things for directly running so I'll merge this without testing myself as I can't imagine this would break something.

hhyyrylainen:
I guess I should have tested as I now have an error when working on another branch and trying to start a new freebuild editor. I can probably do a fix quick, but I'll just bundle that up with my changes going into: https://github.com/Revolutionary-Games/Thrive/pull/6029

Issue #5994, opened 2025-03-07
Make MetabolicStabilityPressure not punish stationary species for not being able to sustain movement


GameDungeon:
Right now MetabolicStabilityPressure severely punishes species that can't sustain constant movement, but this penalty should not apply to sedentary species.

PR #6093, opened 2025-04-25
MetabolicStabilityPressure Adjustement

CritikalH1T-pon3:
For issues:

https://github.com/Revolutionary-Games/Thrive/issues/5994

I worked on making MetabolicStabilityPressure not punish stationary species for not being able to sustain movement


src/auto-evo/selection_pressure/MetabolicStabilityPressure.cs
```diff
@@ -41,8 +41,14 @@ public override float Score(Species species, Patch patch, SimulationCache cache)
...
-            // Punish microbes that can't move continuously severely
-            return 0.25f;
+            // Only punish non-sessile species for not being able to move continuously
+            if (microbeSpecies.Behaviour.Activity > 0)
+            {
+                return 0.25f;
+            }
+
+            // Sessile species get full score if they can maintain energy while stationary
+            return 1.0f;
```
GameDungeon, line 51:
```suggestion
            // Only punish non-sessile species for not being able to move continuously
            return 1.0 - microbeSpecies.Behaviour.Activity / Constants.MAX_SPECIES_ACTIVITY;
```
This hasn't actually been tested, but it should scale the score based on the activity so it isn't just a huge drop off.

GameDungeon:
at the top:
```cs
    public MetabolicStabilityPressure(float weight) : base(weight, [
        AddOrganelleAnywhere.ThatCreateCompound(Compound.ATP),
        RemoveOrganelle.ThatUseCompound(Compound.ATP),
        new UpgradeOrganelle(organelle => organelle.HasMovementComponent, new FlagellumUpgrades(-1.0f)),
    ])```
is the mutations.

Adding this would make it so the auto-evo mutations better respect this new scoring.
```cs
new ChangeBehaviorScore(ChangeBehaviorScore.BehaviorAttribute.Activity, -150.0f),```
will encourage creatures to be sessile so they don't get the penalty

Patryk26g:
Please, add description to the PR and link the issue if it exists

PR #6116, opened 2025-05-04
Alter process indicator speed during speed mode

### LINKED ISSUE #6048: Process panel spinners should take game timescale into account
as they don't speed up when 2x simulation speed mode is used. The reason for that is that the GUI speed isn't doubled as I expect that to like break all animations etc. to make it look bad. But the process panel might be a bit confusing if it doesn't speed up.

This is a relatively easy thing to read and multiply with the simulation speed, however I'm not quickly doing this myself as it will be tiny bit annoying to pass through the data all the way to the chemical equations from where the speed is available. Still, I'm tagging this as easy as I don't foresee any difficulties besiedes just finding the right place to add a bit of code to forward the variable to the ChemicalEquation class.

CI09:

CreatureHUDBase now has method for updating process list speed modifier that is called when speed mode is changed.
ChemicalEquations now know their ProcessList parent and said ProcessList has speed modifier property that is read by chemical equations. ProcessPanel uses ProcessList's speed modifier property. It's just ported into seperate one so it can be accessed by HUD.

**Related Issues**

Closes #6048

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/ChemicalEquation.cs
+++ b/src/gui_common/ChemicalEquation.cs
@@ -7,8 +7,6 @@ using Godot;
 /// </summary>
 public partial class ChemicalEquation : VBoxContainer
 {
-    public ProcessList ParentList = null!;
-
 #pragma warning disable CA2213
     [Export]
     public LabelSettings DefaultTitleFont = null!;
@@ -126,6 +124,8 @@ public partial class ChemicalEquation : VBoxContainer
     /// </summary>
     public bool ShowPerSecondLabel { get; set; } = true;
 
+    public float ExternalSpeedModifier { get; set; } = 1.0f;
+
     /// <summary>
     ///   Only if this is true, the text "/ second" is shown fully, instead of being abbreviated to ` / s`
     /// </summary>
@@ -175,7 +175,7 @@ public partial class ChemicalEquation : VBoxContainer
     {
         if (ShowSpinner && EquationFromProcess != null)
         {
-            currentSpinnerRotation += (float)delta * EquationFromProcess.CurrentSpeed * SpinnerBaseSpeed * ParentList.ExternalSpeedModifier;
+            currentSpinnerRotation += (float)delta * EquationFromProcess.CurrentSpeed * SpinnerBaseSpeed * ExternalSpeedModifier;
 
             // TODO: should we at some point subtract like 100000*360 from the spinner rotation to avoid float range
             // exceeding?
--- a/src/gui_common/ChemicalEquation.cs
+++ b/src/gui_common/ChemicalEquation.cs
@@ -7,6 +7,8 @@ using Godot;
 /// </summary>
 public partial class ChemicalEquation : VBoxContainer
 {
+    public ProcessList ParentList = null!;
+
 #pragma warning disable CA2213
     [Export]
     public LabelSettings DefaultTitleFont = null!;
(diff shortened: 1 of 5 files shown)


src/gui_common/ChemicalEquation.cs
```diff
@@ -7,6 +7,8 @@
 /// </summary>
 public partial class ChemicalEquation : VBoxContainer
 {
+    public ProcessList ParentList = null!;
```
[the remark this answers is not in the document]
CI09: Chemical equations are procedurally instantiated though + that way all chemical equation modifier values never will be desynced.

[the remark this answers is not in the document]
hhyyrylainen: Doesn't that make it simpler then? On instantiate apply the current speed, and on speed change just apply the new speed to all of the equation instances (I'm sure there's a list of them somewhere just for exactly this situation, and that's the reason why I always ask to refactor anything where a list of dynamic items isn't kept).

src/microbe_stage/ProcessList.cs
```diff
@@ -33,6 +35,21 @@ public IEnumerable<IProcessDisplayInfo>? ProcessesToShow
...
+    /// <summary>
+    ///   External, more technical modifiers for speed, like 2x gameplay speed modifier
+    /// </summary>
+    public float ExternalSpeedModifier
+    {
+        get => externalSpeedModifier;
+
+        set
+        {
+            externalSpeedModifier = value;
```
hhyyrylainen, line 47: This should have an optimization where if the modifier didn't change this just returns and doesn't call `UpdateEquationsExternalSpeedModifier`.

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -247,7 +247,10 @@ public override void ApplySpeedMode(bool fastModeEnabled)
             return;
         }
 
-        stage.WorldSimulation.WorldTimeScale = fastModeEnabled ? 2 : 1;
+        var resultingModifier = fastModeEnabled ? 2 : 1;
+
+        stage.WorldSimulation.WorldTimeScale = resultingModifier;
+        UpdateProcessPanelExternalSpeedModifier(resultingModifier);
```
hhyyrylainen, line 253: I think calling this here is not enough, because the speed mode is preserved when making and loading a save, so the process speed actually needs to be checked and applied constantly, otherwise the GUI state might get desynced from the real simulation state when loading a save.

src/microbe_stage/ProcessList.cs
```diff
@@ -103,6 +121,14 @@ private ChemicalEquation CreateEquation(StrictProcessDisplayInfoEquality process
         return equation;
     }
 
+    private void UpdateEquationsExternalSpeedModifier()
+    {
+        foreach (var equation in createdProcessControls.GetChildren())
```
hhyyrylainen, line 126: This needs protection if this is called before _Ready (when the createdProcessControls is created), otherwise this may cause a problem if someone updates the property too early.

hhyyrylainen (changes requested):
The approach looks better now, though I see a few bugs here.

src/microbe_stage/ProcessList.cs
CI09: For some reason check thinks guard is redundant

hhyyrylainen: That's because it's marked non-nullable, but the way we use that it only means that the variable is fine to use after _Ready is called. So either mark the field as nullable or use `== null!` check here.

src/microbe_stage/ProcessList.cs
```diff
@@ -103,6 +125,22 @@ private ChemicalEquation CreateEquation(StrictProcessDisplayInfoEquality process
         return equation;
     }
 
+    private void UpdateEquationsExternalSpeedModifier()
+    {
+        if (createdProcessControls == null!)
+            return;
+
+        if (ExternalSpeedModifier == previousExternalSpeedModifier)
```
hhyyrylainen, line 133: This is unnecessary, the check already exists in the setter that avoids duplicate values being set and calling UpdateEquationsExternalSpeedModifier again.

hhyyrylainen: I'll change this myself so I can do a final test on this PR (as I need to soon wrap up work for today).

hhyyrylainen (approved):
I tested that this works in-game, including with saving and loading. I did a small code improvement so this should be good to go now.

Issue #4805, opened 2024-01-15
Investigate making the pause menu into a singleton and not separate component each stage has


hhyyrylainen:
this will reduce the amount of times the thriveopedia page tree needs to be generated, so eventually we might need to switch to this approach to reduce the lag on opening the pause menu. As an added bonus the thriveopedia link click handler can be merged into the singleton.

hhyyrylainen:
This would presumably also help with https://github.com/Revolutionary-Games/Thrive/issues/5075 as we would do less scene loading.

And example problematic callstack that this change would help with:

```
ERROR: System.NullReferenceException: Object reference not set to an instance of an object.
   at ThriveopediaPatchMapPage.UpdatePlayerPatch(Patch patch) in /home/hhyyrylainen/Projects/Thrive/src/thriveopedia/pages/ThriveopediaPatchMapPage.cs:line 91
   at ThriveopediaPatchMapPage.SetGameForMap() in /home/hhyyrylainen/Projects/Thrive/src/thriveopedia/pages/ThriveopediaPatchMapPage.cs:line 112
   at ThriveopediaPatchMapPage.UpdateCurrentWorldDetails() in /home/hhyyrylainen/Projects/Thrive/src/thriveopedia/pages/ThriveopediaPatchMapPage.cs:line 61
   at Thriveopedia.set_CurrentGame(GameProperties value) in /home/hhyyrylainen/Projects/Thrive/src/thriveopedia/Thriveopedia.cs:line 138
   at PauseMenu.set_GameProperties(GameProperties value) in /home/hhyyrylainen/Projects/Thrive/src/general/PauseMenu.cs:line 94
   at StageBase.SetupStage() in /home/hhyyrylainen/Projects/Thrive/src/general/base_stage/StageBase.cs:line 307
   at CreatureStageBase`2.SetupStage() in /home/hhyyrylainen/Projects/Thrive/src/general/base_stage/CreatureStageBase.cs:line 300
   at MicrobeStage.SetupStage() in /home/hhyyrylainen/Projects/Thrive/src/microbe_stage/MicrobeStage.cs:line 1087
   at MicrobeStage._Ready() in /home/hhyyrylainen/Projects/Thrive/src/microbe_stage/MicrobeStage.cs:line 197
```

PR #6134, opened 2025-05-11
Basic planet customizer

Patryk26g:

Creates planet customizer tool which allows for fast planet generation and adds new options to new game settings. 

Current customization is vary basic but I will work more on it later.

I have moved some existing files to "menus" folder to order them in some way. 

After the review and merge I will create another PR where I will make the new game settings GUI more consistent 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/gui_common/menus/PlanetSettings.cs
```diff
@@ -0,0 +1,165 @@
...
+    public void SetDayLength(float value)
+    {
+        dayLength.Value = value;
+    }
+
+    protected override void Dispose(bool disposing)
+    {
+        if (disposing)
+        {
+            lifeOriginButton.Dispose();
```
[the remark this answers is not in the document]
Patryk26g: Okay, I got rid of thge dispose :)

Patryk26g:
Okay, I merged the newest changes and thankfully it still works :D. Sorry it took me so long but I was away for some time. It should be ready for code review now

Deus-Codes:
Well organized, looks good. Great start for us to expand on.

 Some things I wanted to point out: 
image
I noticed that the tooltip for the Climate Instability measure copies the tooltip for the World Size, which can be confusing. Also would be worthwhile to rename the Climate Instability settings ("Stable, Average, Unstable" instead of "Small, Medium, and Large").

The panspermia option is greyed out; is there a particular reason why/was this also true before, or is it new? 

Instead of having the planet editor be listed under Advanced Options, I think it should either be it's own specific tab or just be the starting menu when creating a new save. Advanced Options has the baggage of being optional and fine-tuney, while a Planet Generator is an inherent part of the experience separate from that sort of detail.

Patryk26g:
Thanks for the review. I think that with all of the changes I just forgot to set the Climate Instability settings names :D.

The paspermia is only available when the LAWK only is turned off, that's how it always worked (although I didn't know that, I discoverd it doing the planet generator :sweat_smile:).

Do you think that the solution would be to get rid of the current basic options menu and set the planet generator tab as default? I was planning on doing that but the basic options contain the "difficulty setting" which is located in advanced difficulty tab and not in the planet generator...

So maybe we should have 2 basic tabs? The one current (maybe without the options for starting place and LAWK) and the planet generator?

There is also third option that comes to my mind - get rid of the basic options menu. Have these 3 tabs and also a button for advanced options. If advanced options are turned off then only 2 tabs are selectable - planet and difficulty. If it is selected then all tabs are available and we also add the more complicated stuff to the "difficulty" tab

hhyyrylainen:
I haven't looked at the latest state yet, but I'm still kind of torn with the toggle buttons (versus checkboxes) as they do look visually more appealing, but I think it makes it slightly harder to quickly see which options are selected as checkboxes are visually immediately next to their explanation. With the options menu also changed to use the toggle buttons I suppose it will become a common enough paradigm in the game to have no chance of player confusion after initially seeing them. Still that's just two places in the game so we'll still have places with checkboxes leading to some less visual cohesion.

But ultimately I guess I can concede to the more visually appealing look with checkboxes gone if that's what everyone else wants.

project.godot
```diff
@@ -16,7 +16,7 @@ Play as a microbe on an alien world trying to survive and adapt to your surround
 
 https://revolutionarygamesstudio.com/"
 config/version="0.8.2.1"
-run/main_scene="res://src/general/MainMenu.tscn"
+run/main_scene="res://src/gui_common/menus/MainMenu.tscn"
```
hhyyrylainen: I don't think it is a good idea to move the main menu to the gui common folder, as it is not a shared part of the GUI, but rather a specific game state, maybe even the most important game state is it is the entry point of the program.

locale/en.po
```diff
@@ -382,6 +382,9 @@ msgstr "run status:"
 msgid "AUTO_EVO_STATUS_COLON"
 msgstr "Auto-evo Status:"
 
+msgid "AUTO_EVO_TOOL_BUTTON"
+msgstr "Use in Auto-Evo Tool"
```
hhyyrylainen: Is this a new tool or referring to the auto-evo explorer tool? If the latter this should use the proper name for the tool.

src/space_stage/SpaceStage.tscn
```diff
@@ -5,7 +5,7 @@
 [ext_resource type="PackedScene" uid="uid://c2h4vpmx0juq8" path="res://assets/models/menu_backgrounds/StarsMesh.tscn" id="3"]
 [ext_resource type="PackedScene" uid="uid://dokew8lhrri5" path="res://assets/models/menu_backgrounds/Comet.tscn" id="4_aqo7l"]
 [ext_resource type="PackedScene" uid="uid://i8evqpkyb68h" path="res://src/space_stage/gui/SpaceHUD.tscn" id="7"]
-[ext_resource type="PackedScene" uid="uid://devtugnggmuol" path="res://src/general/PauseMenu.tscn" id="8"]
+[ext_resource type="PackedScene" uid="uid://devtugnggmuol" path="res://src/gui_common/menus/PauseMenu.tscn" id="8"]
```
hhyyrylainen: I can kind of see why this change makes logical sense, *however* my future plan is that the pause menu would be a persistent scene that is always loaded and not a part of all the separate scenes (https://github.com/Revolutionary-Games/Thrive/issues/4805). So this is again a change I would like to not have in this PR.

src/gui_common/menus/NewGameSettings.tscn
hhyyrylainen, line 1: I kind of think that we don't have a situation where the new game settings menu is reused anywhere. So I think this should also stay in the general folder. I can picture situations where the planet customizer is used in the space / ascension stage for example so I think that's a fine fit to be in the GUI common folder.

src/general/world_effects/patch_events/UnderwaterVentEruptionEffect.cs
```diff
@@ -42,7 +42,7 @@ public void OnTimePassed(double elapsed, double totalTimePassed)
             if (patch.BiomeType != BiomeType.Vents)
                 continue;
 
-            if (random.Next(100) > Constants.VENT_ERUPTION_CHANCE)
+            if (random.Next(100) > GetVentEruptionChance())
```
hhyyrylainen: Doesn't this now basically always fail? Because you changed the vent eruption constant from 15 to 0.15, so it is now extremely unlikely (or maybe impossible) for this random check here to pass... This should be swapped out for `random.NextFloat()` and I think then it would again work correctly.

Patryk26g: Yes, my intention was to change it to random.NextFloat(), I don't know how I forgot it

Patryk26g:
I have moved the files back to their previous folders :)

src/auto-evo/AutoEvoExploringTool.tscn
```diff
@@ -485,7 +485,8 @@ theme_override_constants/margin_bottom = 10
 
 [node name="DraggableScrollContainer" parent="GUI/VBoxContainer/HistoryReportSplit/PatchMap/HBoxContainer/PatchMap/MarginContainer" node_paths=PackedStringArray("content") instance=ExtResource("16")]
 layout_mode = 2
-content = NodePath("PatchMapDrawer")
+ContentPath = NodePath("PatchMapDrawer")
```
hhyyrylainen: Isn't this actually incorrect change as the node_paths on line 486 doesn't match? (I noticed a similar change in the editor scenes)

src/general/world_effects/patch_events/GlobalGlaciationEvent.cs
```diff
@@ -213,6 +213,37 @@ private void AddIceChunks(Patch patch)
...
+                return random.Next((int)(Constants.GLOBAL_GLACIATION_MIN_DURATION * 1.5f),
+                    (int)(Constants.GLOBAL_GLACIATION_MAX_DURATION * 1.5f));
+            case WorldGenerationSettings.ClimateInstabilityEnum.Medium:
+            default:
+                return random.Next(Constants.GLOBAL_GLACIATION_MIN_DURATION,
+                    Constants.GLOBAL_GLACIATION_MAX_DURATION);
+        }
+    }
+
+    private float GetSunlightMultiplier()
```
hhyyrylainen, line 242: I think this code should have a warning that this may not change otherwise the ending the event blows up. I didn't read the world settings code change yet, but there should also be a warning there that ClimateInstability may not be changed due to the glaciation event requiring it to stay constant.

src/general/WorldGenerationSettings.cs
```diff
@@ -100,6 +160,26 @@ public enum WorldSizeEnum
...
+    /// </summary>
+    public WorldSeaLevelEnum WorldSeaLevel { get; set; } = WorldSeaLevelEnum.Moderate;
+
+    /// <summary>
+    ///   Geological activity of World
+    /// </summary>
+    public GeologicalActivityEnum GeologicalActivity { get; set; } = GeologicalActivityEnum.Average;
+
+    /// <summary>
+    ///   Climate instability of World
```
hhyyrylainen, line 179: This should have the warning that changing this will break glaciation event restoring the world. If there's not an issue yet about the problems with the ending event reverting data not always working, a new issue could be created for that and a link put here so people of the future can read the relevant info about what could fix this and allow this to be safely modified.

src/gui_common/menus/PlanetCustomizerTool.cs
```diff
@@ -0,0 +1,117 @@
...
+    private void OnRegeneratePressed()
+    {
+        InitNewWorld(SimulationParameters.Instance.AutoEvoConfiguration);
+    }
+
+    private void OnAutoEvoToolPressed()
+    {
+        var scene = SceneManager.Instance.LoadScene("res://src/auto-evo/AutoEvoExploringTool.tscn").Instantiate();
+
+        if (scene is AutoEvoExploringTool tool)
```
hhyyrylainen, line 113: I think this should give a pretty loud error if this fails, as otherwise players will be confused why nothing seemed to go wrong if they get a different planet instead in the exploring tool.

src/gui_common/menus/PlanetCustomizerTool.cs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    private void OnRegeneratePressed()
+    {
+        InitNewWorld(SimulationParameters.Instance.AutoEvoConfiguration);
+    }
+
+    private void OnAutoEvoToolPressed()
+    {
+        var scene = SceneManager.Instance.LoadScene("res://src/auto-evo/AutoEvoExploringTool.tscn").Instantiate();
```
hhyyrylainen: This doesn't match our usual pattern of `GD.Load<PackedScene>("stuff")` so I'd change this for code consistency purposes.

locale/en.po
```diff
@@ -3085,7 +3124,7 @@ msgid "LIFE_ORIGIN"
 msgstr "Life origin"
 
 msgid "LIFE_ORIGIN_EXPLANATION"
-msgstr "(starting location)"
+msgstr "(Starting location)"
```
hhyyrylainen: I'm kind of against this change as we use this pattern in many places in the game where short explanations in parentheses are not started with a capital letter.

src/gui_common/menus/PlanetCustomizerTool.cs
Patryk26g: Why would it fail? And if so, maybe we just shouldn't let it go to the Auto-Evo explorer tool, so I would put the transition code into that if statement. But if you think that warning popup would be more appropriate then I will do it

src/general/WorldGenerationSettings.cs
Patryk26g: I don't think it needs any warnings as it would not break the glaciation. Its duration is only calculated once and then saved and this setting is not used anymore. Anyway, why would that setting be modified after the world is created? You cannot and should not be able to change the planet's properties once the game starts

src/gui_common/menus/PlanetCustomizerTool.cs
hhyyrylainen: I think a warning popup would be overkill, I just think that there should be a GD.PrintErr at least here about the failure.

src/general/WorldGenerationSettings.cs
hhyyrylainen: But it *is* used by the code ending the event, I commented here: https://github.com/Revolutionary-Games/Thrive/pull/6134/files#r2206823694 thus meaning that if the value changes while the event is in progress that would be a sure way to end up with not 100% light at the end.

I can imagine that maybe we would want to make world conditions vary over time, like maybe the instability is high for the first n generations and then gets lower or something like that.

Patryk26g: Okay I'm sorry your right, so I guess we only need a warning in the comment above?

hhyyrylainen: Yes, and also in the relevant code in the event so that if someone wants to resolve the warning they can find the info needed rather than needing to investigate the whole thing from scratch.

Patryk26g:
Okay, I think I address the rest of the issues. I dont know why the check has failed

hhyyrylainen:
I hope the test failure is just that the latest master needs to be merged in. I updated the unit testing framework very recently, and the master builds have been succeeding.

hhyyrylainen:
First time playtesting this, I noticed this new error:
```
ERROR: Node not found: "OptionsMenu" (relative to "/root/MainMenu").
   at: get_node (scene/main/node.cpp:1877)
```

I'm not sure yet what caused that.

hhyyrylainen:
On my initial testing in my game the GUI looks good and seems to work. However the sea level option isn't very intuitive to me. It seems more like it should be about how connected the ocean areas are or something like that that would more explain what it does. I was about to comment about a bug with the sea level not affecting how deep the ocean floor is from the surface, but then I read the explanation text on it and realized I had gotten it wrong. I bet that some players will also make the same reading mistake. So I think it would be great to rename that option to something that more clearly communicates immediately what it does by its name.

hhyyrylainen:
I found a bug with the planet customizer, if I hit the "generate world" button (which BTW I think should be "Generate World" to match our casing in buttons) then go back to the setting sand press that again, I get this error each time:

image

(the given key changes numbers but otherwise the error is the same)

Edit: the "Regenerate world" button also triggers the issue (and that too should be capitalized as "Regenerate World")

hhyyrylainen:
One more small thing, we could finally remove the "???" button from the main menu tools menu, as I think there's enough buttons there already to not need to have placeholders to pad out the menu.

src/gui_common/menus/PlanetSettings.cs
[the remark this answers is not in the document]
hhyyrylainen: Isn't this still in the latest diff for this PR? I don't think I'm accidentally looking at older commit range and these disposes are still here.

src/gui_common/menus/PlanetCustomizerTool.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            InitNewWorld(SimulationParameters.Instance.AutoEvoConfiguration);
+            generateButton.Text = Localization.Translate("BACK_TO_SETTINGS");
+        }
+        else
+        {
+            generateButton.Text = Localization.Translate("GENERATE_BUTTON");
+        }
+
+        settingsPanel.Visible = !settingsPanel.Visible;
```
hhyyrylainen: Could you add a comment explaining this line of code? I've read this entire file and I just can't reason why this toggling of visibility is here and specifically after the button text has been updated.

src/general/NewGameSettings.tscn
hhyyrylainen, line 1: This is a lot of changes to look through so I mostly skimmed and then verified in-game the functions. I didn't notice any other problems except the planet seed is now gone from the advanced view, I think it should still be there similarly to how the LAWK option is duplicated from the simple view.

hhyyrylainen: Also while I do kind of get why the LAWK option is now lower than before, the reason why it was at the top is that the life origin setting has one option that depends on LAWK. So to me it makes a lot of sense to put these options next to each other as turning off LAWK allows more starting choices to be selected.

hhyyrylainen: Apparently I'm just commenting now as I test... sorry about the individual comments spam, but could you turn on the follow focus option for the scroll boxes in the new settings? That would with quite minimal effort make the keyboard navigation work a ton better on this screen (there is the same problem on master, but this should take only like 30 seconds to improve things a lot so that's why I'm asking).

src/general/NewGameSettings.cs
```diff
@@ -284,6 +264,11 @@ public override void _Ready()
             backButton.Visible = false;
             checkOptionsMenuAdviceContainer.Visible = false;
         }
+
+        planetSettings.Connect(PlanetSettings.SignalName.LawkSettingsChanged,
+            new Callable(this, nameof(OnLawkPlanetSettingChanged)));
+        planetSettings.Connect(PlanetSettings.SignalName.LifeOriginSettingsChanged,
+            new Callable(this, nameof(OnLifeOriginPlanetSettingChanged)));
```
hhyyrylainen: We prefer making connections like this through the Godot editor so could you move these into the NewGameSettings.tscn file?

src/gui_common/menus/PlanetSettings.cs
```diff
@@ -0,0 +1,165 @@
...
+    private LineEdit dayLengthReadout = null!;
+
+    [Export]
+    private VBoxContainer dayLengthContainer = null!;
+
+    [Signal]
+    public delegate void LawkSettingsChangedEventHandler(bool lawkOnly);
+
+    [Signal]
+    public delegate void LifeOriginSettingsChangedEventHandler(int lifeOrigin);
```
hhyyrylainen: As this is exposed as a signal from a comment, I think this needs to be refactored to use a proper enum (or string values) so that GUI indexes aren't passed between components. To me that seems very brittle and against the principal of making GUI components' interfaces insulated from their internal implementation details.

Patryk26g:
I still haven't done 3 things: enums, comment explaining the buttonPush and name change for sea level

src/gui_common/menus/PlanetCustomizerTool.cs
Patryk26g: Well pushing the button switches between settings with button labeled as "generate" and patch map with button labeled as "regenerate". I will add a comment there

hhyyrylainen: Wouldn't it make more sense if the visibility state was changed *first* and then the current data applied to the button text etc? That way this would look like a toggle button callback and then easy enough to understand applying of current state to the GUI.

locale/en.po
```diff
@@ -4799,6 +4838,9 @@ msgstr "Redo the last action"
 msgid "REFRESH"
 msgstr "Refresh"
 
+msgid "REGENERATE_BUTTON"
+msgstr "Regenerate world"
```
hhyyrylainen:
```suggestion
msgstr "Regenerate World"
```

Patryk26g:
I think I might be done :)

hhyyrylainen:
I started testing the changes again, and I noticed a new bug: the random seed button on the planet customizer screen doesn't do anything, the seed stays as "123456789". I think it also would be better if a first random seed was automatically selected that way the player wouldn't need to press the random button whenever they want to try out the planet customizer.

hhyyrylainen:
Also it seems like the displayed seed is not used as if I use the "Back to Settings" button and then hit generate again without changing the seed, I get different maps even though the GUI claims the seed is the same.

src/general/NewGameSettings.cs
```diff
@@ -855,51 +846,41 @@ private void OnMapTypeSelected(int index)
...
         {
             lifeOriginButton.Selected = (int)WorldGenerationSettings.LifeOrigin.Vent;
-            lifeOriginButtonAdvanced.Selected = (int)WorldGenerationSettings.LifeOrigin.Vent;
         }
     }
 
-    private void OnGameSeedChangedFromBasic(string text)
+    private void OnLifeOriginPlanetSettingChanged(int index)
     {
-        // Need different methods to handle each view; otherwise we overwrite caret position
```
hhyyrylainen, line 894: This bug with the caret position is back.

If I go to the advanced screen, select the whole seed, and then type in "42", that actually results in "24" being displayed due to the caret repositioning bug.

src/general/NewGameSettings.cs
```diff
@@ -269,7 +248,8 @@ public override void _Ready()
 
         var seed = GenerateNewRandomSeed();
         gameSeed.Text = seed;
-        gameSeedAdvanced.Text = seed;
+
+        // gameSeedAdvanced.Text = seed;
```
hhyyrylainen: I think it is not worth it to keep this commented out text anymore with the separate component handling the advanced seed.

hhyyrylainen (changes requested):
The new game screen seems mostly bug free now. I did notice a slight caret positioning bug when typing in a seed. The code other than one small nitpick looks good now entirely.

The planet customizer screen has some more bugs I noticed (and commented on before I started this review).

So unfortunately there are still bugs I could find within a few minutes of testing.

src/general/NewGameSettings.tscn
hhyyrylainen: Seems like the focus follow option is still not set? At least on the advanced planet configuration tab doesn't follow focus.

Patryk26g:
I have fixed the bugs with the carets and planet generator

hhyyrylainen:
I just realized the keyboard navigation is still broken. I'll try to quickly fix it.

PR #6148, opened 2025-05-21
Make organelle names update when the language is changed

### LINKED ISSUE #5897: Organelle names in editor tooltips don't update when selecting a new language
instead they require restarting the game

Image

dligr:

Makes organelle names update when the language is changed by passing the new name to the tooltip.

**Related Issues**

Fixes #5897

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/tooltip/ToolTipManager.cs
+++ b/src/gui_common/tooltip/ToolTipManager.cs
@@ -378,12 +378,14 @@ public partial class ToolTipManager : CanvasLayer
         // Organelle tooltips
         foreach (var organelle in SimulationParameters.Instance.GetAllOrganelles())
         {
-            var tooltip = GetToolTip(organelle.InternalName, "organelleSelection");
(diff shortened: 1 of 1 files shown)


src/gui_common/tooltip/ToolTipManager.cs
```diff
@@ -378,12 +378,14 @@ public void OnLanguageChanged()
         // Organelle tooltips
         foreach (var organelle in SimulationParameters.Instance.GetAllOrganelles())
         {
-            var tooltip = GetToolTip(organelle.InternalName, "organelleSelection");
+            var tooltip = GetToolTip(organelle.InternalName, "organelleSelection") as SelectionMenuToolTip;
```
hhyyrylainen, line 381: I think it would be good to do the cast in a second step and print a warning if the cast fails, as because written like this, this can silently fail as the tooltip doesn't exist and wrong type will now be both handled just by:

```
            if (tooltip == null)
                continue;
```

so a future bug may be hard to find.

PR #6161, opened 2025-05-30
Limit unlock tooltip size

### LINKED ISSUE #5936: Do something about too long list of unlock conditions that goes offscreen
My idea here is still that we could hide all but the closest 2-3 unlocks for the player, and then in the tooltip say how many other organelles there are still to discover, but other solutions are also welcome.

0HyperCube:
In order to rank the unlocks, we probably need a function to return the progress (0 to 1) of each unlock condition. Where multiple unlock conditions are required, the average is used. Where one is needed, the maximum can be taken.

Unlock conditions can then be sorted by the progress and only the first 4 displayed.

**Related Issues**
Closes #5936

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/OrganelleDefinition.cs
+++ b/src/microbe_stage/OrganelleDefinition.cs
@@ -791,6 +791,13 @@ public class OrganelleDefinition : IRegistryType
         }
     }
 
+    public float Progress(WorldAndPlayerDataSource worldAndPlayerData)
+    {
+        if (UnlockConditions == null)
+            return 0;
+        return UnlockConditions.Select(entry => entry.Progress(worldAndPlayerData)).Max();
+    }
+
     public void ApplyTranslations()
     {
         TranslationHelper.ApplyTranslations(this);
--- a/src/microbe_stage/organelle_unlocks/IUnlockCondition.cs
+++ b/src/microbe_stage/organelle_unlocks/IUnlockCondition.cs
@@ -15,6 +15,16 @@ public interface IUnlockCondition
     public bool Satisfied(IUnlockStateDataSource data);
 
     /// <summary>
+    ///   Gets the current progress on the unlock condition
+    /// </summary>
+    /// <param name="data">
+    ///   The data about the world, if this condition doesn't handle the data of the given type this always
+    ///   returns false
+    /// </param>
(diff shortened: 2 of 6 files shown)


hhyyrylainen:
What still needs to be done for this PR?

src/microbe_stage/organelle_unlocks/IUnlockCondition.cs
```diff
@@ -14,6 +14,16 @@ public interface IUnlockCondition
     /// <returns>True if condition satisfied</returns>
     public bool Satisfied(IUnlockStateDataSource data);
 
+    /// <summary>
+    ///   Gets the current progress on the unlock condition
+    /// </summary>
+    /// <param name="data">
+    ///   The data about the world, if this condition doesn't handle the data of the given type this always
+    ///   returns false
```
hhyyrylainen, line 22:
```suggestion
    ///   returns 0
```
Should it say this instead?

src/microbe_stage/OrganelleDefinition.cs
```diff
@@ -791,6 +791,14 @@ public void GenerateUnlockRequirementsText(LocalizedStringBuilder builder,
         }
     }
 
+    public float Progress(WorldAndPlayerDataSource worldAndPlayerData)
+    {
+        if (UnlockConditions == null)
+            return 0;
+
+        return UnlockConditions.Select(entry => entry.Progress(worldAndPlayerData)).Max();
```
hhyyrylainen, line 799: This might allocate a temporary delegate...

src/microbe_stage/organelle_unlocks/WorldBasedUnlockCondition.cs
```diff
@@ -87,24 +99,30 @@ public class SpeedBelow : WorldBasedUnlockCondition
...
-    private float GetPlayerSpeed(ICellDefinition playerData)
+    private float? GetPlayerSpeed(IUnlockStateDataSource data)
     {
+        if (data is not WorldAndPlayerDataSource worldArgs)
+            return null;
+
+        var playerData = worldArgs.PlayerData;
+
+        if (playerData == null)
+            return null;
```
hhyyrylainen, line 124: And also this I think causes boxing on the return value of this method.

PR #6163, opened 2025-06-01
Solving the Nucleus Paradox

domin2ktr:

This PR gives a buff of +30% bioprocess speed and -10% osmoregulation cost. This is done so that the nucleus finally has a meaning other than "this part is needed for more advanced organelles"
As a result, this gives a break-even point for the core when the cell size reaches approximately 25 hexagons.

This has biological justifications in the form of the fact that the nucleus reduces the likelihood of errors in the production of proteins, and also accelerates the production of proteins, which reduces the requirements of osmoregulation, as well as increases the efficiency of organelles (since proteins are used to produce energy, accordingly, increasing the rate of protein production will increase the rate of energy production).

**Related Issues**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/OsmoregulationAndHealingSystem.cs
+++ b/src/microbe_stage/systems/OsmoregulationAndHealingSystem.cs
@@ -127,6 +127,18 @@ public sealed class OsmoregulationAndHealingSystem : AEntitySetSystem<float>
             osmoregulationCost *= 20.0f / (20.0f + colonySize);
         }
 
+        // 10% osmoregulation bonus if have nucleus
+        if (!cellProperties.IsBacteria)
+        {
+            osmoregulationCost *= 0.9f;
+        }
+
+        // 30% bioprocess speed bonus if have nucleus
+        if (!cellProperties.IsBacteria)
+        {
+            microbeEnvironmentalEffects.ProcessSpeedModifier *= 1.3f;
+        }
+
         // TODO: remove this check on next save breakage point
         if (entity.Has<MicrobeEnvironmentalEffects>())
         {


0HyperCube:
Hi @domin2ktr,
You seem to be having a bit of trouble updating the ProcessSpeedModifier and OsmoregulationModifier. You might be interested to review the following patch:
```diff
diff --git a/src/auto-evo/simulation/SimulationCache.cs b/src/auto-evo/simulation/SimulationCache.cs
index e8528674f..630a44c68 100644
--- a/src/auto-evo/simulation/SimulationCache.cs
+++ b/src/auto-evo/simulation/SimulationCache.cs
@@ -536,6 +536,15 @@ public class SimulationCache
 
         var result = MicrobeEnvironmentalToleranceCalculations.ResolveToleranceValues(tolerances);
 
+        if (!species.IsBacteria)
+        {
+            // 30% bioprocess speed bonus if have nucleus
+            result.ProcessSpeedModifier *= 1.3f;
+
+            // 10% osmoregulation bonus if have nucleus
+            result.OsmoregulationModifier *= 0.9f;
+        }
+
         cachedResolvedTolerances.Add(key, result);
         return result;
     }
```

P.S. Mostly you do development on your local computer with git rather than editing directly on GitHub.
P.P.S. Also I recommend you don't commit to the master branch of your fork. I suggest you instead create a different branch allowing you to have multiple PRs at the same time.

I hope that helps.

domin2ktr:
> Hi @domin2ktr, You seem to be having a bit of trouble updating the ProcessSpeedModifier and OsmoregulationModifier. You might be interested to review the following patch:
> 
> ```diff
> diff --git a/src/auto-evo/simulation/SimulationCache.cs b/src/auto-evo/simulation/SimulationCache.cs
> index e8528674f..630a44c68 100644
> --- a/src/auto-evo/simulation/SimulationCache.cs
> +++ b/src/auto-evo/simulation/SimulationCache.cs
> @@ -536,6 +536,15 @@ public class SimulationCache
>  
>          var result = MicrobeEnvironmentalToleranceCalculations.ResolveToleranceValues(tolerances);
>  
> +        if (!species.IsBacteria)
> +        {
> +            // 30% bioprocess speed bonus if have nucleus
> +            result.ProcessSpeedModifier *= 1.3f;
> +
> +            // 10% osmoregulation bonus if have nucleus
> +            result.OsmoregulationModifier *= 0.9f;
> +        }
> +
>          cachedResolvedTolerances.Add(key, result);
>          return result;
>      }
> ```
> 
> P.S. Mostly you do development on your local computer with git rather than editing directly on GitHub. P.P.S. Also I recommend you don't commit to the master branch of your fork. I suggest you instead create a different branch allowing you to have multiple PRs at the same time.
> 
> I hope that helps.

Will this have an effect on the player's cells as well?

0HyperCube:
@domin2ktr I have created an updated diff that also alters the cell editor:
```diff
diff --git a/src/microbe_stage/MicrobeEnvironmentalToleranceCalculations.cs b/src/microbe_stage/MicrobeEnvironmentalToleranceCalculations.cs
index 429bc84c1..953d1ef02 100644
--- a/src/microbe_stage/MicrobeEnvironmentalToleranceCalculations.cs
+++ b/src/microbe_stage/MicrobeEnvironmentalToleranceCalculations.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 
 /// <summary>
 ///   Helper class that contains all the math for environmental tolerances in one place (though the microbe editor and
@@ -9,6 +10,8 @@ using System.Diagnostics;
 /// </summary>
 public static class MicrobeEnvironmentalToleranceCalculations
 {
+    public static OrganelleDefinition Nucleus = SimulationParameters.Instance.GetOrganelleType("nucleus");
+
     /// <summary>
     ///   Calculates the total overall score of environmental tolerance without giving the sub-scores
     /// </summary>
@@ -60,6 +63,7 @@ public static class MicrobeEnvironmentalToleranceCalculations
             resolvedTolerances.UVResistance = 0;
 
         CalculateTolerancesInternal(resolvedTolerances, noExtraEffects, environment, result);
+        result.HasNucleus = organelles.Any(orgnaelle => orgnaelle.Definition == Nucleus);
 
         return result;
     }
@@ -206,6 +210,15 @@ public static class MicrobeEnvironmentalToleranceCalculations
             result.OsmoregulationModifier *= Math.Min(Constants.TOLERANCE_UV_OSMOREGULATION_MAX, 2 - data.UVScore);
         }
 
+        if (data.HasNucleus)
+        {
+            // 30% bioprocess speed bonus if have nucleus
+            result.ProcessSpeedModifier *= 6f;
+
+            // 10% osmoregulation bonus if have nucleus
+            result.OsmoregulationModifier *= 0.1f;
+        }
+
 #if DEBUG
         if (result.OsmoregulationModifier <= MathUtils.EPSILON)
         {
@@ -435,4 +448,5 @@ public class ToleranceResult
 
     public float UVScore;
     public float PerfectUVAdjustment;
+    public bool HasNucleus;
 }
```

microscopic editor showing reduction in cost

domin2ktr:
> Hi @domin2ktr, You seem to be having a bit of trouble updating the ProcessSpeedModifier and OsmoregulationModifier. You might be interested to review the following patch:
> 
> ```diff
> diff --git a/src/auto-evo/simulation/SimulationCache.cs b/src/auto-evo/simulation/SimulationCache.cs
> index e8528674f..630a44c68 100644
> --- a/src/auto-evo/simulation/SimulationCache.cs
> +++ b/src/auto-evo/simulation/SimulationCache.cs
> @@ -536,6 +536,15 @@ public class SimulationCache
>  
>          var result = MicrobeEnvironmentalToleranceCalculations.ResolveToleranceValues(tolerances);
>  
> +        if (!species.IsBacteria)
> +        {
> +            // 30% bioprocess speed bonus if have nucleus
> +            result.ProcessSpeedModifier *= 1.3f;
> +
> +            // 10% osmoregulation bonus if have nucleus
> +            result.OsmoregulationModifier *= 0.9f;
> +        }
> +
>          cachedResolvedTolerances.Add(key, result);
>          return result;
>      }
> ```
> 
> P.S. Mostly you do development on your local computer with git rather than editing directly on GitHub. P.P.S. Also I recommend you don't commit to the master branch of your fork. I suggest you instead create a different branch allowing you to have multiple PRs at the same time.
> 
> I hope that helps.

image
Why it chek failed?

0HyperCube:
@domin2ktr the check failed because SimulationCache.cs line 547 has trailing white space.

domin2ktr:
> @domin2ktr the check failed because SimulationCache.cs line 547 has trailing white space.

image
It remains to wait until someone allows the merging.

(And by the way, will they add me to the credits?)

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -536,6 +536,15 @@ public ResolvedMicrobeTolerances GetEnvironmentalTolerances(MicrobeSpecies speci
...
 
+        if (!species.IsBacteria)
+        {
+            // 30% bioprocess speed bonus if have nucleus
+            result.ProcessSpeedModifier *= 1.3f;
+
+            // 10% osmoregulation bonus if have nucleus
+            result.OsmoregulationModifier *= 0.9f;
+        }
+
```
0HyperCube: You apply the modifier twice here I think? Since it is already done in `MicrobeEnvironmentalToleranceCalculations.cs`.

Deus-Codes:
Good initiative to take this on yourself. I will take a look at it. My initial reaction is that I'd prefer it to be attached to a blatant mechanic such as a size-related cost just to make its functionality more apparent, but this could definitely do otherwise.

adQuid:
I would like to add that we already have a 50% damage reduction for eucaryotes that we never tell the players about, that was our first attempt at addressing this issue https://github.com/Revolutionary-Games/Thrive/pull/3341

Personally I was never in love with that idea (even though I made the first PR about it), so I'm fine if we want to replace it, just posting for awareness when balancing.

domin2ktr:
> I would like to add that we already have a 50% damage reduction for eucaryotes that we never tell the players about, that was our first attempt at addressing this issue #3341
> 
> Personally I was never in love with that idea (even though I made the first PR about it), so I'm fine if we want to replace it, just posting for awareness when balancing.

As deus said, "we need to focus on more general game mechanics", so he added the sunlight tolerance mechanic first, rather than the preferred flowers for photosynthesis mechanic.
That's why I think the energy bonus is better, it gives more universal, meaningful and general, while the health increase is only useful in some situations.

hhyyrylainen (changes requested):
This has the technical problem of using the tolerance system to force the buff. That will cause incorrect GUI displays and incorrect messages shown to the player, not to mention how surprising that is considering there's no added text explaining the mechanic change.

So next I think the game design question should be answered if this is a desired change. If it is then we can talk about how technically the buffs should be implemented.

hhyyrylainen:
Why did you request me to re-review this? As far as I can tell nothing has changed since my previous comment.

domin2ktr:
> This has the technical problem of using the tolerance system to force the buff. That will cause incorrect GUI displays and incorrect messages shown to the player, not to mention how surprising that is considering there's no added text explaining the mechanic change.
> 
> So next I think the game design question should be answered if this is a desired change. If it is then we can talk about how technically the buffs should be implemented.

Regarding the first, I can say that the screenshot above shows that everything is displayed correctly.

Regarding the second, I can say that I think you can simply add to the description of the core that it has energy bonuses for large cells.

> Why did you request me to re-review this? As far as I can tell nothing has changed since my previous comment.

Sorry, I accidentally pressed the wrong button.

hhyyrylainen:
Okay, I accept that I may have been mistaken, but I just tested in game and I'm pretty sure this messes with things.

Here I'm in a patch I'm not adapted to:




And then I add a nucleus:




And boom, my debuff display changes. That should not happen if this feature was made as its own feature and not inserted as a change to tolerances.

Also for some reason having *bad* pressure tolerance gives a buff:




That doesn't make sense. Also when I correct that, the pressure suddenly gives a bonus, which should not be able to happen unless perfectly adapted (and I have not changed the pressure range so I cannot be perfectly adapted):




So unless I made a mistake in my testing, that shows that this is the wrong technical approach to making this buff and my earlier comment is correct that the approach needs to be redone if this kind of change is to be accepted.

domin2ktr:
> Okay, I accept that I may have been mistaken, but I just tested in game and I'm pretty sure this messes with things.
> 
> Here I'm in a patch I'm not adapted to:
> 
> 
> 
> And then I add a nucleus:
> 
> 
> 
> And boom, my debuff display changes. That should not happen if this feature was made as its own feature and not inserted as a change to tolerances.
> 
> Also for some reason having _bad_ pressure tolerance gives a buff:
> 
> 
> 
> That doesn't make sense. Also when I correct that, the pressure suddenly gives a bonus, which should not be able to happen unless perfectly adapted (and I have not changed the pressure range so I cannot be perfectly adapted):
> 
> 
> 
> So unless I made a mistake in my testing, that shows that this is the wrong technical approach to making this buff and my earlier comment is correct that the approach needs to be redone if this kind of change is to be accepted.

Can you remind me which file contains the condition that increases the health of a eukaryotic cell by 2 times?

hhyyrylainen:
> Can you remind me which file contains the condition that increases the health of a eukaryotic cell by 2 times?

It does not affect the health. At least I cannot recall such a thing and when double checking in-game just now placing a nucleus didn't change my cell's health from 60 to anything else.

This is where the 50% damage resistance for eukaryotes is applied:

https://github.com/Revolutionary-Games/Thrive/blob/a074cc7e0a365943ad1636e7e1e02a7ce6c3edaa/src/engine/common_components/Health.cs#L205

Knowing that place is not very helpful for other changes. Like I said before, I think each individual effect you'd like to add for the nucleus needs to put into its relevant system that governs that game mechanic.

domin2ktr:
> > Can you remind me which file contains the condition that increases the health of a eukaryotic cell by 2 times?
> 
> It does not affect the health. At least I cannot recall such a thing and when double checking in-game just now placing a nucleus didn't change my cell's health from 60 to anything else.
> 
> This is where the 50% damage resistance for eukaryotes is applied:
> 
> https://github.com/Revolutionary-Games/Thrive/blob/a074cc7e0a365943ad1636e7e1e02a7ce6c3edaa/src/engine/common_components/Health.cs#L205
> 
> Knowing that place is not very helpful for other changes. Like I said before, I think each individual effect you'd like to add for the nucleus needs to put into its relevant system that governs that game mechanic.

Can I just paste this code into "process.cs" to make it work?

        if (!species.IsBacteria)
        {
            // 30% bioprocess speed bonus if have nucleus
            result.ProcessSpeedModifier *= 1.3f;
        }

And also insert this into "OsmoregulationAndHealingSystem.cs"?

        if (!species.IsBacteria)
        {
            // 20% osmoregulation bonus if have nucleus
            result.OsmoregulationModifier *= 0.8f;
        }

hhyyrylainen:
Kind of. That code will not compile as-is. But theoretically it would work to check the species properties. However, neither of those systems currently access the microbe species information as the systems are designed to not be hardcoded for the microbe stage, which is one of the core concepts of an ECS is that the systems shouldn't be too complicated and specific that they cannot be reused.

So if you want you could modify the osmoregulation and healing system to read species information. But I won't allow the same edit for the process system as it is intentionally kept disconnected from specific microbe stage details.

domin2ktr:
> Kind of. That code will not compile as-is. But theoretically it would work to check the species properties. However, neither of those systems currently access the microbe species information as the systems are designed to not be hardcoded for the microbe stage, which is one of the core concepts of an ECS is that the systems shouldn't be too complicated and specific that they cannot be reused.
> 
> So if you want you could modify the osmoregulation and healing system to read species information. But I won't allow the same edit for the process system as it is intentionally kept disconnected from specific microbe stage details.

What is the name of the parameter responsible for the cost of movement?

hhyyrylainen:
This is where the base movement cost (so excluding flagella) is calculated:

https://github.com/Revolutionary-Games/Thrive/blob/3f597a23b926ffcc9267821f3fbdef86f450197d/src/microbe_stage/systems/MicrobeMovementSystem.cs#L257

domin2ktr:
> This is where the base movement cost (so excluding flagella) is calculated:
> 
> https://github.com/Revolutionary-Games/Thrive/blob/3f597a23b926ffcc9267821f3fbdef86f450197d/src/microbe_stage/systems/MicrobeMovementSystem.cs#L257

    if (!species.IsBacteria)
    {
        // 35% moving bonus if have nucleus
        cost *= 0,65f;
    }

If I paste this code into "MicrobeMoventSystem.cs" will it work?

hhyyrylainen:
Why did you not try it? From quickly looking at that file, it looks like it isn't setup to check the species, so that code should not even compile.

Sadly I do not have time to suggest a step-by-step solution to this problem or check each time if code you ask about would even compile.

domin2ktr:
> Why did you not try it? From quickly looking at that file, it looks like it isn't setup to check the species, so that code should not even compile.
> 
> Sadly I do not have time to suggest a step-by-step solution to this problem or check each time if code you ask about would even compile.



> Why did you not try it? From quickly looking at that file, it looks like it isn't setup to check the species, so that code should not even compile.
> 
> Sadly I do not have time to suggest a step-by-step solution to this problem or check each time if code you ask about would even compile.

If I move the Hypercube code to MicrobeMoventSystem and change a couple of parameters in it, will it compile?

(And for some reason I now don’t have the opportunity to compile and test these changes yet.)

hhyyrylainen:
> If I move the Hypercube code to MicrobeMoventSystem and change a couple of parameters in it, will it compile?

What code are you talking about? This: https://github.com/Revolutionary-Games/Thrive/pull/6163#issuecomment-2927860632 ? 

It would be theoretically possible to calculate whether something is an eukaryote based on the organelle list, which is already processed in the movement system. But performance needs to be considered so the nucleus property should actually be cached in `struct OrganelleContainer` which slightly complicates that approach.

domin2ktr:
> > If I move the Hypercube code to MicrobeMoventSystem and change a couple of parameters in it, will it compile?
> 
> What code are you talking about? This: [#6163 (comment)](https://github.com/Revolutionary-Games/Thrive/pull/6163#issuecomment-2927860632) ?
> 
> It would be theoretically possible to calculate whether something is an eukaryote based on the organelle list, which is already processed in the movement system. But performance needs to be considered so the nucleus property should actually be cached in `struct OrganelleContainer` which slightly complicates that approach.

Will the changes in "MicrobeMoventSystem.cs" be work?

hhyyrylainen:
You should be able to see yourself that the changes don't even compile by checking the "build" CI job.

https://dev.revolutionarygamesstudio.com/ci/1/build/14721/jobs/1

domin2ktr:
> You should be able to see yourself that the changes don't even compile by checking the "build" CI job.
> 
> https://dev.revolutionarygamesstudio.com/ci/1/build/14721/jobs/1



> You should be able to see yourself that the changes don't even compile by checking the "build" CI job.
> 
> https://dev.revolutionarygamesstudio.com/ci/1/build/14721/jobs/1

Maybe there is still a place in the code where one could insert this file without errors?

    if (!species.IsBacteria)
    {
        // 30% bioprocess speed bonus if have nucleus
        result.ProcessSpeedModifier *= 1.3f;
    }

If not, then I just don't know how it could be implemented.

domin2ktr:
> Kind of. That code will not compile as-is. But theoretically it would work to check the species properties. However, neither of those systems currently access the microbe species information as the systems are designed to not be hardcoded for the microbe stage, which is one of the core concepts of an ECS is that the systems shouldn't be too complicated and specific that they cannot be reused.
> 
> So if you want you could modify the osmoregulation and healing system to read species information. But I won't allow the same edit for the process system as it is intentionally kept disconnected from specific microbe stage details.



> Kind of. That code will not compile as-is. But theoretically it would work to check the species properties. However, neither of those systems currently access the microbe species information as the systems are designed to not be hardcoded for the microbe stage, which is one of the core concepts of an ECS is that the systems shouldn't be too complicated and specific that they cannot be reused.
> 
> So if you want you could modify the osmoregulation and healing system to read species information. But I won't allow the same edit for the process system as it is intentionally kept disconnected from specific microbe stage details.

How to make this script work in the osmoregulation system?

    if (!species.IsBacteria)
    {
        // 30% bioprocess speed bonus if have nucleus
        result.ProcessSpeedModifier *= 1.3f;
    }

hhyyrylainen:
The system would need to be converted from using `SpeciesMember` to using `MicrobeSpeciesMember` though that will break the system in the multicellular stage so the system also needs to be made to work with `MulticellularSpecies` which is the kind of complexity I was talking about earlier by making systems too specific to one stage instead of using just general information. That's such a complex edit that I can't really describe it without doing it myself.

And sadly I cannot put in the time to fully babysit this code change. So you need to either find someone else to help you out or improve your own C# skills.

domin2ktr:
@0HyperCube Can you make a script to speed up bioprocesses with a core, without bugs with the environmental tolerance system?

domin2ktr:
> The system would need to be converted from using `SpeciesMember` to using `MicrobeSpeciesMember` though that will break the system in the multicellular stage so the system also needs to be made to work with `MulticellularSpecies` which is the kind of complexity I was talking about earlier by making systems too specific to one stage instead of using just general information. That's such a complex edit that I can't really describe it without doing it myself.
> 
> And sadly I cannot put in the time to fully babysit this code change. So you need to either find someone else to help you out or improve your own C# skills.

@dligr Can you code the 40 percent moving cost reduction please?

dligr:
So, I implemented the movement cost reduction, here's the git diff that I got:

<details>
  <summary>
    git diff
  </summary>
  
  ```diff
  ---------------------- simulation_parameters/Constants.cs ----------------------
  index 28bb9647f..93d4afaa7 100644
  @@ -215,6 +215,8 @@ public static class Constants
     /// </remarks>
     public const float BASE_MOVEMENT_ATP_COST = 1.0f;
   
  +    public const float NUCLEUS_MOVEMENT_COST_MODIFIER = 0.6f;
  +
     public const float BASE_MOVEMENT_FORCE = 900.0f;
   
     public const float MAX_STRAIN_PER_ENTITY = 400.0f;

  -------------- src/microbe_stage/systems/MicrobeMovementSystem.cs --------------
  index 794d121d5..8a0b07ba6 100644
  @@ -256,6 +256,11 @@ public sealed class MicrobeMovementSystem : AEntitySetSystem<float>
       // movement cost
       var cost = Constants.BASE_MOVEMENT_ATP_COST * organelles.HexCount * length * delta * strainMultiplier;
   
  +        if (!cellProperties.IsBacteria)
  +        {
  +            cost *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
  +        }
  +
       var got = compounds.TakeCompound(Compound.ATP, cost);
   
       // Halve base movement speed if out of ATP

  ------------------ src/microbe_stage/systems/ProcessSystem.cs ------------------
  index b27f9d1ae..4132b7ef8 100644
  @@ -33,6 +33,8 @@ using World = DefaultEcs.World;
   [RuntimeCost(55)]
   public sealed class ProcessSystem : AEntitySetSystem<float>
   {
  +    public static OrganelleDefinition Nucleus = SimulationParameters.Instance.GetOrganelleType("nucleus");
  +
   #if CHECK_USED_STATISTICS
     private readonly List<ProcessStatistics> usedStatistics = new();
   #endif
  @@ -849,6 +851,8 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
   
       int hexCount = 0;
   
  +        bool hasNucleus = false;
  +
       int organelleCount = organelles.Count;
       for (int i = 0; i < organelleCount; ++i)
       {
  @@ -877,9 +881,18 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
   
         // Store hex count
         hexCount += organelle.Definition.HexCount;
  +
  +            if (organelle.Definition == Nucleus)
  +                hasNucleus = true;
       }
   
       var baseMovement = Constants.BASE_MOVEMENT_ATP_COST * hexCount;
  +
  +        if (hasNucleus)
  +        {
  +            baseMovement *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
  +        }
  +
       result.BaseMovement += baseMovement;
   
       if (includeMovementCost)
  ```
</details>

hhyyrylainen:
It looks pretty correct to me, but that will only apply to base movement, which in the case of cells with flagellum (which is most eukaryotes that can move) means that a significant portion of the movement cost is not reduced.

dligr:
> that will only apply to base movement

True, I slightly changed the approach to fix this

<details>
  <summary>
    git diff
  </summary>
  
  ```diff
  ---------------------- simulation_parameters/Constants.cs ----------------------
  index 28bb9647f..93d4afaa7 100644
  @@ -215,6 +215,8 @@ public static class Constants
     /// </remarks>
     public const float BASE_MOVEMENT_ATP_COST = 1.0f;
   
  +    public const float NUCLEUS_MOVEMENT_COST_MODIFIER = 0.6f;
  +
     public const float BASE_MOVEMENT_FORCE = 900.0f;
   
     public const float MAX_STRAIN_PER_ENTITY = 400.0f;

  --------- src/microbe_stage/organelle_components/MovementComponent.cs ---------
  index e9a266f75..43c68169b 100644
  @@ -141,6 +141,11 @@ public class MovementComponent : IOrganelleComponent
       var requiredEnergy = (Constants.FLAGELLA_ENERGY_COST + Constants.FLAGELLA_MAX_UPGRADE_ATP_USAGE
         * flagellumLength) * elapsed;
   
  +        if (!isBacteria)
  +        {
  +            requiredEnergy *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
  +        }
  +
       var availableEnergy = compounds.TakeCompound(Compound.ATP, requiredEnergy);
   
       if (availableEnergy < requiredEnergy)

  -------------- src/microbe_stage/systems/MicrobeMovementSystem.cs --------------
  index 794d121d5..8a0b07ba6 100644
  @@ -256,6 +256,11 @@ public sealed class MicrobeMovementSystem : AEntitySetSystem<float>
       // movement cost
       var cost = Constants.BASE_MOVEMENT_ATP_COST * organelles.HexCount * length * delta * strainMultiplier;
   
  +        if (!cellProperties.IsBacteria)
  +        {
  +            cost *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
  +        }
  +
       var got = compounds.TakeCompound(Compound.ATP, cost);
   
       // Halve base movement speed if out of ATP

  ------------------ src/microbe_stage/systems/ProcessSystem.cs ------------------
  index b27f9d1ae..21e75e392 100644
  @@ -33,6 +33,8 @@ using World = DefaultEcs.World;
   [RuntimeCost(55)]
   public sealed class ProcessSystem : AEntitySetSystem<float>
   {
  +    public static OrganelleDefinition Nucleus = SimulationParameters.Instance.GetOrganelleType("nucleus");
  +
   #if CHECK_USED_STATISTICS
     private readonly List<ProcessStatistics> usedStatistics = new();
   #endif
  @@ -246,8 +248,10 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
       }
   #endif
   
  +        bool hasNucleus = organelles.Any(orgnaelle => orgnaelle.Definition == Nucleus);
  +
       CalculateSimplePartOfEnergyBalance(organelles, biome, environmentTolerances, membrane, onlyMovementInDirection,
  -            includeMovementCost, isPlayerSpecies, worldSettings, amountType, cache, result);
  +            includeMovementCost, isPlayerSpecies, worldSettings, amountType, cache, hasNucleus, result);
     }
   
     /// <summary>
  @@ -278,8 +282,10 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
       bool includeMovementCost, bool isPlayerSpecies, WorldGenerationSettings worldSettings,
       CompoundAmountType amountType, SimulationCache? cache, EnergyBalanceInfoFull result)
     {
  +        bool hasNucleus = organelles.Any(orgnaelle => orgnaelle.Definition == Nucleus);
  +
       CalculateSimplePartOfEnergyBalance(organelles, biome, environmentTolerances, membrane, onlyMovementInDirection,
  -            includeMovementCost, isPlayerSpecies, worldSettings, amountType, cache, result);
  +            includeMovementCost, isPlayerSpecies, worldSettings, amountType, cache, hasNucleus, result);
   
       // Once simple balance is calculated we add the extra info on top, this approach loops the organelles twice
       // but reduces code duplication
  @@ -293,7 +299,14 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
   
         // Take special cell components that take energy into account
         if (TryGetMovementCostForOrganelle(includeMovementCost, organelle, onlyMovementInDirection, out var cost))
  +            {
  +                if (hasNucleus)
  +                {
  +                    cost *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
  +                }
  +
           result.AddConsumption(organelle.Definition.InternalName, cost);
  +            }
   
         if (includeMovementCost && organelle.Definition.HasCiliaComponent)
         {
  @@ -841,7 +854,7 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
       IBiomeConditions biome, in ResolvedMicrobeTolerances environmentTolerances, MembraneType membrane,
       Vector3 onlyMovementInDirection, bool includeMovementCost, bool isPlayerSpecies,
       WorldGenerationSettings worldSettings, CompoundAmountType amountType,
  -        SimulationCache? cache, EnergyBalanceInfoSimple result)
  +        SimulationCache? cache, bool hasNucleus, EnergyBalanceInfoSimple result)
     {
       var processATPProduction = 0.0f;
       var processATPConsumption = 0.0f;
  @@ -880,6 +893,15 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
       }
   
       var baseMovement = Constants.BASE_MOVEMENT_ATP_COST * hexCount;
  +
  +        if (hasNucleus)
  +        {
  +            baseMovement *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
  +
  +            movementATPConsumption -= result.Flagella * (1.0f - Constants.NUCLEUS_MOVEMENT_COST_MODIFIER);
  +            result.Flagella *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
  +        }
  +
       result.BaseMovement += baseMovement;
   
       if (includeMovementCost)
  ```
</details>

domin2ktr:
> It looks pretty correct to me, but that will only apply to base movement, which in the case of cells with flagellum (which is most eukaryotes that can move) means that a significant portion of the movement cost is not reduced.

Will we have time to add this before the update comes out?

hhyyrylainen:
> Will we have time to add this before the update comes out?

There's no time for new features before the 0.8.3 release.

domin2ktr:
> > Will we have time to add this before the update comes out?
> 
> There's no time for new features before the 0.8.3 release.



> So, I implemented the movement cost reduction, here's the git diff that I got:
> 
> git diff
> ```diff
> ---------------------- simulation_parameters/Constants.cs ----------------------
> index 28bb9647f..93d4afaa7 100644
> @@ -215,6 +215,8 @@ public static class Constants
>    /// </remarks>
>    public const float BASE_MOVEMENT_ATP_COST = 1.0f;
>  
> +    public const float NUCLEUS_MOVEMENT_COST_MODIFIER = 0.6f;
> +
>    public const float BASE_MOVEMENT_FORCE = 900.0f;
>  
>    public const float MAX_STRAIN_PER_ENTITY = 400.0f;
> 
> -------------- src/microbe_stage/systems/MicrobeMovementSystem.cs --------------
> index 794d121d5..8a0b07ba6 100644
> @@ -256,6 +256,11 @@ public sealed class MicrobeMovementSystem : AEntitySetSystem<float>
>      // movement cost
>      var cost = Constants.BASE_MOVEMENT_ATP_COST * organelles.HexCount * length * delta * strainMultiplier;
>  
> +        if (!cellProperties.IsBacteria)
> +        {
> +            cost *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
> +        }
> +
>      var got = compounds.TakeCompound(Compound.ATP, cost);
>  
>      // Halve base movement speed if out of ATP
> 
> ------------------ src/microbe_stage/systems/ProcessSystem.cs ------------------
> index b27f9d1ae..4132b7ef8 100644
> @@ -33,6 +33,8 @@ using World = DefaultEcs.World;
>  [RuntimeCost(55)]
>  public sealed class ProcessSystem : AEntitySetSystem<float>
>  {
> +    public static OrganelleDefinition Nucleus = SimulationParameters.Instance.GetOrganelleType("nucleus");
> +
>  #if CHECK_USED_STATISTICS
>    private readonly List<ProcessStatistics> usedStatistics = new();
>  #endif
> @@ -849,6 +851,8 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
>  
>      int hexCount = 0;
>  
> +        bool hasNucleus = false;
> +
>      int organelleCount = organelles.Count;
>      for (int i = 0; i < organelleCount; ++i)
>      {
> @@ -877,9 +881,18 @@ public sealed class ProcessSystem : AEntitySetSystem<float>
>  
>        // Store hex count
>        hexCount += organelle.Definition.HexCount;
> +
> +            if (organelle.Definition == Nucleus)
> +                hasNucleus = true;
>      }
>  
>      var baseMovement = Constants.BASE_MOVEMENT_ATP_COST * hexCount;
> +
> +        if (hasNucleus)
> +        {
> +            baseMovement *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
> +        }
> +
>      result.BaseMovement += baseMovement;
>  
>      if (includeMovementCost)
> ```

Why doesn't your code work?

dligr:
> Why doesn't your code work?

There seem to be typos on lines 251 and 285 in ProcessSystem.cs:
`bool hasNucleus = organelles.Any(organelle => orgnaelle.Definition == Nucleus);` the second `organelle` in both lines is misspoken as `orgnaelle`, while both names should be the same

Also, there should be a blank line after the line 309 (in the same file)

domin2ktr:
> > Will we have time to add this before the update comes out?
> 
> There's no time for new features before the 0.8.3 release.

Merge this pull request

domin2ktr:
> > Will we have time to add this before the update comes out?
> 
> There's no time for new features before the 0.8.3 release.

Hey, we can start merging this pr now?

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -536,6 +536,11 @@ public ResolvedMicrobeTolerances GetEnvironmentalTolerances(MicrobeSpecies speci
 
         var result = MicrobeEnvironmentalToleranceCalculations.ResolveToleranceValues(tolerances);
 
+        if (!species.IsBacteria)
+        {
+            result.OsmoregulationModifier *= 0.2f;
```
hhyyrylainen: Wouldn't this still cause the problem with incorrect tolerances being displayed in the GUI?

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -880,6 +893,15 @@ private static void CalculateSimplePartOfEnergyBalance(IReadOnlyList<OrganelleTe
         }
 
         var baseMovement = Constants.BASE_MOVEMENT_ATP_COST * hexCount;
+
+        if (hasNucleus)
+        {
+            baseMovement *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
+
+            movementATPConsumption -= result.Flagella * (1.0f - Constants.NUCLEUS_MOVEMENT_COST_MODIFIER);
+            result.Flagella *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
```
hhyyrylainen: This only modifies the calculated energy balance and doesn't actually modify the ATP flagella drain during gameplay. So this causes a difference between calculated (and shown to the player) ATP usage and the actual ATP usage.

hhyyrylainen: As far as I can see, this change is still in the code of this PR?

domin2ktr: > As far as I can see, this change is still in the code of this PR?

It is planned that this should only affect the standard cost of movement.

src/auto-evo/simulation/SimulationCache.cs
domin2ktr: > Wouldn't this still cause the problem with incorrect tolerances being displayed in the GUI?

I decided to simply remove the impact on auto-evo since auto-evo doesn't take into account the core's energy consumption in its calculations anyway.

src/microbe_stage/systems/ProcessSystem.cs
hhyyrylainen: This line of code here adjusting the flagella cost is still here...

domin2ktr: I removed the line responsible for the cost of movement from the flagellum

domin2ktr: @hhyyrylainen hey, I fixed it.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -293,7 +299,14 @@ public static void ComputeEnergyBalanceFull(IReadOnlyList<OrganelleTemplate> org
 
             // Take special cell components that take energy into account
             if (TryGetMovementCostForOrganelle(includeMovementCost, organelle, onlyMovementInDirection, out var cost))
+            {
+                if (hasNucleus)
+                {
+                    cost *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
```
hhyyrylainen: That other thing I commented on is now fixed. But there's this still this part that modifies the flagella energy in *this* calculation.

This one PR is taking up so much of my reviewing time that I just can't keep reviewing this. So sadly I need to decide to not keep reviewing this.
I'm willing to have a final look at this, *but* I will not do that before there's someone else who has helped you polish this PR and fix all the problems at once rather than taking multiple comments from me for each small change.

domin2ktr:
@dligr Check how well my code works, it should not give errors, incorrectly display the costs of osmoregulation or affect the energy costs of the flagellum.

src/microbe_stage/systems/ProcessSystem.cs
domin2ktr: > That other thing I commented on is now fixed. But there's this still this part that modifies the flagella energy in _this_ calculation.
> 
> This one PR is taking up so much of my reviewing time that I just can't keep reviewing this. So sadly I need to decide to not keep reviewing this. I'm willing to have a final look at this, _but_ I will not do that before there's someone else who has helped you polish this PR and fix all the problems at once rather than taking multiple comments from me for each small change.


I did some tests and now it works without errors.

The editor displays everything correctly
image
image


Outside of the editor it works just as well. (without reducing osmoregulation this cell (this is not the cell I showed above) would die)
image

There are no errors with the flagellum either.
image
image


This function is now fully functional and has no errors.
Please do the merge this PR.

src/microbe_stage/systems/ProcessSystem.cs
```diff
@@ -896,6 +909,11 @@ private static void CalculateSimplePartOfEnergyBalance(IReadOnlyList<OrganelleTe
         var osmoregulation = Constants.ATP_COST_FOR_OSMOREGULATION * hexCount *
             membrane.OsmoregulationFactor;
 
+        if (hasNucleus)
+        {
+            osmoregulation *= Constants.NUCLEUS_MOVEMENT_COST_MODIFIER;
```
dligr: The PR seems to work, but there's an issue with this line: osmoregulation is begin multiplied by the movement cost modifier. I think the modifier should be `0.6f`, like in `OsmoregulationAndHealingSystem` or both should be made into their own constant.

domin2ktr: > The PR seems to work, but there's an issue with this line: osmoregulation is begin multiplied by the movement cost modifier. I think the modifier should be `0.6f`, like in `OsmoregulationAndHealingSystem` or both should be made into their own constant.

i fixed this problem

domin2ktr: I created a new constant responsible for osmoregulation from the core. In game tests showed that this works well.

domin2ktr: > The PR seems to work, but there's an issue with this line: osmoregulation is begin multiplied by the movement cost modifier. I think the modifier should be `0.6f`, like in `OsmoregulationAndHealingSystem` or both should be made into their own constant.

When will this PR be merged?

domin2ktr:
@hhyyrylainen This PR fully works and was tested by me and dligr. 
All that remains is your consent to the merge this PR.

hhyyrylainen:
I see that the code changes that are here are now fine. However the code changes are only about 20-30% of the total work required. Still missing:

- Updates to the nucleus tooltips explaining these effects and the modifier stats list
- Checking the tutorial and other important text in the game whether they correctly characterise the nucleus now
- Reviewing whether there's any Thriveopedia content that must be updated to reflect the new information. I think that at least the strategy and nucleus pages need to be updated
- Checking if these much more limited changes actually still "solve the nucleus paradox"
- Checking what auto-evo does and do these changes actually help in it considering the nucleus useful
- **Getting the OK from the game design team** whether these buff amounts are good or should they be higher/lower and in general if this seems like a good game design change
- And as a bonus (though I suspect there's not much problem here as the nucleus already gives a 50% damage reduction): checking the theory side that whether it makes sense for movement and osmoregulation cost to be reduced from having a nucleus

So this PR is in now ready on a prototype level for gameplay testing, but not close to being ready for general inclusion in the game.

domin2ktr:
> I see that the code changes that are here are now fine. However the code changes are only about 20-30% of the total work required. Still missing:
> 
> * Updates to the nucleus tooltips explaining these effects and the modifier stats list
> * Checking the tutorial and other important text in the game whether they correctly characterise the nucleus now
> * Reviewing whether there's any Thriveopedia content that must be updated to reflect the new information. I think that at least the strategy and nucleus pages need to be updated
> * Checking if these much more limited changes actually still "solve the nucleus paradox"
> * Checking what auto-evo does and do these changes actually help in it considering the nucleus useful
> * **Getting the OK from the game design team** whether these buff amounts are good or should they be higher/lower and in general if this seems like a good game design change
> * And as a bonus (though I suspect there's not much problem here as the nucleus already gives a 50% damage reduction): checking the theory side that whether it makes sense for movement and osmoregulation cost to be reduced from having a nucleus
> 
> So this PR is in now ready on a prototype level for gameplay testing, but not close to being ready for general inclusion in the game.

The first 3 problems I can solve by simply changing the text of the tips, descriptions, etc., which is quite easy to do. 
The other depends on what the thrive game design team decides.

Deus-Codes:
Hey domin2ktr, wanted to let you know that I'll be getting to this hopefully tomorrow, and if not then, sometime this weekend. Good work on sticking with this. My initial impression is that this is a good change, it just requires some balancing. Playtesting it will give me a better feel of how it works

Deus-Codes:
After a quick test: I do think this is a nice mechanic, but it does need more balancing. Because the cost of the nucleus diminishes, it doesn't take very long at all to create an organism able to support a nucleus. Some rather simple cells can easily support the part now (an organism with about 7 cytoplasm and a hydrogenosome and a cytoplasm can fully cover the costs, as seen at the bottom of this comment.)

I think an immediate thing that pops in my head is jacking up the price of the nucleus. That way, you still have to build a big enough cell to make it a manageable cost, but there's still that element of strategy and timing. 

Another thing we should pay attention to when it comes to balancing: since the nucleus benefits size after a while, is there a point later in the stage at a bigger mass where players really don't have to concern themselves at all with adding high energy parts? 

image

image

domin2ktr:
> After a quick test: I do think this is a nice mechanic, but it does need more balancing. Because the cost of the nucleus diminishes, it doesn't take very long at all to create an organism able to support a nucleus. Some rather simple cells can easily support the part now (an organism with about 7 cytoplasm and a hydrogenosome and a cytoplasm can fully cover the costs, as seen at the bottom of this comment.)
> 
> I think an immediate thing that pops in my head is jacking up the price of the nucleus. That way, you still have to build a big enough cell to make it a manageable cost, but there's still that element of strategy and timing.
> 
> Another thing we should pay attention to when it comes to balancing: since the nucleus benefits size after a while, is there a point later in the stage at a bigger mass where players really don't have to concern themselves at all with adding high energy parts?
> 
> image image

after doing a couple of tests I decided to reduce the energy bonus to 33% instead of 40%. This makes getting the core at the beginning of the game much less easy (since you can't just put hydrogenases in a few generations since they still need to be fed with glucose, which isn't that easy), and now the core's breakeven point is 30 hexes in size.
image

Accidental-Explorer:
> after doing a couple of tests I decided to reduce the energy bonus to 33% instead of 40%. This makes getting the core at the beginning of the game much less easy (since you can't just put hydrogenases in a few generations since they still need to be fed with glucose, which isn't that easy), and now the core's breakeven point is 30 hexes in size. 

Even though I haven't tested for myself yet, I have to agree in principle with lowering the bonus being a better balance knob than increasing the size of the nucleus. Decreasing the bonus better preserves the cost for adding individual organelles, and does less to devalue any future reductions in osmo/movement cost we may want to add.

Accidental-Explorer:
Having now actually tested, I do think the basic principle is very good. And I can definitely see this working after some tuning.

Right, now I do think it's still working a bit _too_ well (the bonus is now set to 33%, yes?). Prokaryotes are being displaced by eukaryotes quite rapidly (though this is very hard to balance overall, I know). It also still happens quite early, though Deus' oxygen rebalance will probably interact with that timing.

Maybe something to try: instead of looking to see how small you need to make the bonus for it to not cause problems, maybe look to see how small you can make the bonus and still see at least _some_ effect? For example, in the current development version, auto-evo explorer might produce 1% of species with a nucleus and they die out quickly. How much does a 5% bonus change that? 10%?
Keep in mind that the nucleus _does have_ other benefits. During my playtesting I was able to place it for nett benefit (also according to auto-evo) because the size enabled me to predate on many more targets, even though I was nowhere near the mathematical break-even point on the bonus versus the cost. So it's enough if the bonus is _just_ large enough to make the nucleus worth it _in combination with its other effects_.

I also still believe that having an overall smaller bonus and lower cost on the nucleus is better balance than going the other way around. Having a bigger bonus diminishes the effect of other osmoregulation cost reductions, like membrane types (also 30-40%). I also feel like it makes placing more cell parts too cheap in osmoregulation cost once you get the nucleus. 
Since we're making changes to osmoregulation cost anyway, I would sooner suggest _lowering_ the nucleus cost, rather than raising it, if that is necessary to balance against a smaller bonus.

domin2ktr:
> After a quick test: I do think this is a nice mechanic, but it does need more balancing. Because the cost of the nucleus diminishes, it doesn't take very long at all to create an organism able to support a nucleus. Some rather simple cells can easily support the part now (an organism with about 7 cytoplasm and a hydrogenosome and a cytoplasm can fully cover the costs, as seen at the bottom of this comment.)
> 
> I think an immediate thing that pops in my head is jacking up the price of the nucleus. That way, you still have to build a big enough cell to make it a manageable cost, but there's still that element of strategy and timing.
> 
> Another thing we should pay attention to when it comes to balancing: since the nucleus benefits size after a while, is there a point later in the stage at a bigger mass where players really don't have to concern themselves at all with adding high energy parts?
> 
> image image

What is your decision on this matter?
(I reduced the bonus from 40% to 33%)

@hhyyrylainen 
Where is the file where I can change the nucleus description?

hhyyrylainen:
All game text goes through the translations system:

https://github.com/Revolutionary-Games/Thrive/blob/master/doc/working_with_translations.md

Deus-Codes:
I'm in agreement with Rathalos (Accidental-Explorer) here. The effect should be much more subtle in my opinion. 

The thing with the nucleus is that it represents a significant progression in the Microbe Stage. If the part is extremely beneficial from a really early part of the game, it is much less of a goal the player must build to and much more of a part which has immediate benefit. 

Again, and as Rathalos says, the bonus being this strong completely throws off balancing for the rest of the game, and can encourage the evolution of the nucleus by turn 4 or so. 

On Earth, the nucleus evolved atleast 2 billion years after life first evolved. In Thrive, that amount of time is equal to 20 generations. I'm not saying that generation 20 should be the target for when the player starts placing down the nucleus, but it shows that the nucleus really took a while to place down. Such an expansion of the player's capabilities should be more of a challenge to build up to, and the paradox is meant to be a subtle nudge for the player and AI to encourage some +/- rather than it being just an immense cost sink. But it should still be a bit cost.

domin2ktr:
> I'm in agreement with Rathalos (Accidental-Explorer) here. The effect should be much more subtle in my opinion.
> 
> The thing with the nucleus is that it represents a significant progression in the Microbe Stage. If the part is extremely beneficial from a really early part of the game, it is much less of a goal the player must build to and much more of a part which has immediate benefit.
> 
> Again, and as Rathalos says, the bonus being this strong completely throws off balancing for the rest of the game, and can encourage the evolution of the nucleus by turn 4 or so.
> 
> On Earth, the nucleus evolved atleast 2 billion years after life first evolved. In Thrive, that amount of time is equal to 20 generations. I'm not saying that generation 20 should be the target for when the player starts placing down the nucleus, but it shows that the nucleus really took a while to place down. Such an expansion of the player's capabilities should be more of a challenge to build up to, and the paradox is meant to be a subtle nudge for the player and AI to encourage some +/- rather than it being just an immense cost sink. But it should still be a bit cost.



> I'm in agreement with Rathalos (Accidental-Explorer) here. The effect should be much more subtle in my opinion.
> 
> The thing with the nucleus is that it represents a significant progression in the Microbe Stage. If the part is extremely beneficial from a really early part of the game, it is much less of a goal the player must build to and much more of a part which has immediate benefit.
> 
> Again, and as Rathalos says, the bonus being this strong completely throws off balancing for the rest of the game, and can encourage the evolution of the nucleus by turn 4 or so.
> 
> On Earth, the nucleus evolved atleast 2 billion years after life first evolved. In Thrive, that amount of time is equal to 20 generations. I'm not saying that generation 20 should be the target for when the player starts placing down the nucleus, but it shows that the nucleus really took a while to place down. Such an expansion of the player's capabilities should be more of a challenge to build up to, and the paradox is meant to be a subtle nudge for the player and AI to encourage some +/- rather than it being just an immense cost sink. But it should still be a bit cost.

I've already reduced the bonus to 33%, which I think is quite sufficient. Furthermore, I can say that even without this osmoregulation bonus, it's possible to reach the core in 4-5 generations using photosynthesis + metabolosomes or chemolithotrophy (I even have a screenshot of how I achieved multicellularity in less than a billion years). So I don't think a 33% bonus will change much.

Accidental-Explorer:
I have to admit it is now a bit difficult to judge the effect this would now have in master considering Deus' balance changes now added there (in particular the oxygen and ATP production changes, as well as the change to nucleus mass).

But overall I think 33% is still a very large bonus. Even if you have just 7 hexes, a nucleus adds less to your total cost than 4 cytoplasm. The effect is still stronger than membrane choice (because it effects both osmoregulation and movement cost). (By the way, testing this I noticed that the nucleus' cost modifier does not affect the osmoregulation cost in organelle tooltips. It probably should.) Going forward after placing the nucleus, every hex you place takes up 33% less ATP.

To repeat myself from before: don't think "how much do I need to reduce this bonus to be reasonable" instead think "how small can I make this bonus while still having an effect".

Right now, taking into account its own bonuses, the Nucleus by itself has an ongoing ATP cost of 12. For example, reducing the bonus to 20% but also decreasing the base cost of the nucleus to 8 (instead of 9), will keep that at 12,8. So that would make the bonus less distorting while still making the nucleus a bit more easy to place.

But again, I really need to see this in combination with the rest of current balance to make any judgement.

Deus-Codes:
1. With the new balancing changes to metabolism, rapid progress like that should now be much harder. Oxygen builds up much more slowly, and anaerobic respiration is now less powerful. 
2. Even then, there’s still the added element in this PR of the nucleus reducing all *future costs* you place down. In most video games, global effects like this are scaled to be very subtle unless there’s a huge amount of effort in getting some really powerful loot or something. Because the nucleus is meant to be the first “level up” in Thrive, it can’t really be balanced to be rare enough to warrant such a huge bonus. 
3. With the prior balancing, even though energy generation was extremely generous, you still had to be intentionally aware of all the costs associated with the nucleus and build your organism around that. With this, because the nucleus pays for itself extremely quickly, that layer of decision-making associated with progression is lessened. 

So again, I think Rathalos’ suggestion of scaling this down with the approach of it noticeable would be more appropriate rather than ensuring that the nucleus is extremely beneficial.

domin2ktr:
> I have to admit it is now a bit difficult to judge the effect this would now have in master considering Deus' balance changes now added there (in particular the oxygen and ATP production changes, as well as the change to nucleus mass).
> 
> But overall I think 33% is still a very large bonus. Even if you have just 7 hexes, a nucleus adds less to your total cost than 4 cytoplasm. The effect is still stronger than membrane choice (because it effects both osmoregulation and movement cost). (By the way, testing this I noticed that the nucleus' cost modifier does not affect the osmoregulation cost in organelle tooltips. It probably should.) Going forward after placing the nucleus, every hex you place takes up 33% less ATP.
> 
> To repeat myself from before: don't think "how much do I need to reduce this bonus to be reasonable" instead think "how small can I make this bonus while still having an effect".
> 
> Right now, taking into account its own bonuses, the Nucleus by itself has an ongoing ATP cost of 12. For example, reducing the bonus to 20% but also decreasing the base cost of the nucleus to 8 (instead of 9), will keep that at 12,8. So that would make the bonus less distorting while still making the nucleus a bit more easy to place.
> 
> But again, I really need to see this in combination with the rest of current balance to make any judgement.

I think reducing the bonus to 25% would be optimal, you won't be able to become eukaryotic too quickly, but you can still reach a point where the core saves more energy than it consumes.

Accidental-Explorer:
> I think reducing the bonus to 25% would be optimal, you won't be able to become eukaryotic too quickly, but you can still reach a point where the core saves more energy than it consumes.

Of course, I do think that is moving in the right direction.

Though keep in mind the nucleus does not necessarily have to _save_ energy. It's a development that IRL happened once, and we really only _need_ it to happen once. So it doesn't need to be something that's definitely worth it for every large cell. It needs to be at least _just about_ worth it for large cells in certain situations, where it might not just be about the discount, but also the gained size and storage.

Once it exists, there is all the unlocked advanced organelles that should carry it forward.

But again, comparing it in the current balance would be very useful. Do you think you can update this branch with the latest commits in Thrive master?

domin2ktr:
@hhyyrylainen merge this pr. I consulted about the balance and also changed the descriptions.

Accidental-Explorer:
> @hhyyrylainen merge this pr. I consulted about the balance and also changed the descriptions.

Does the osmoregulation cost in the tooltip when hovering over organelles in the tab now update after you have a nucleus?

hhyyrylainen:
> Does the osmoregulation cost in the tooltip when hovering over organelles in the tab now update after you have a nucleus?

That's also a very good thing to test that every display updates correctly upon placing or undoing the nucleus.

domin2ktr:
> I really think it would be much better for these key stats to be included in the list of buffs and debuffs the organelle has (so a different part of the tooltip as well), because otherwise at a glance these very important stats that affect almost everything about the player's cell will be easy to miss for players who are in general notoriously bad at reading more than a few sentences.
> 
> And for this tooltip text, I think this is something that will be balanced a bunch so I kind of think the numbers should not be hardcoded into the text, but rather put in as placeholders ( `{0}`, `{1}`) and then during runtime updated with the actual values, because otherwise any balance changes will break the translations etc. which would cause a lot of problems for balancing this change (I do realize that the 50% damage reduction is hardcoded text but that's different I think as the exact value of that has never been a topic of debate).

I don't think this is a cause for concern. The percentage values ​​stand out from the rest of the text, so players will notice (plus, I decided to keep the description brief). And the osmoregulation and movement cost reduction values ​​set in the core are unlikely to change more than a few times. Besides, this change isn't significant enough to be too concerning.

Stop wasting time already. I think it's ready for merge.

Deus-Codes:
When we bring things up for implementation, we aren’t just making things up - they are constants and necessities for every part of the game, and are necessary before implementation. All features are to be judged in their quality of implementation by the programming team, and all features are to be judged for their balancing and value by the design team. So when we say things need to be changed, things need to be changed. With all due respect - it is up to us what goes in to our game, and how it is implemented.

There is room for discussion about new features of course, and different interpretations. But when we say that a change is significant and requires proper accessibility/documentation, we mean it. All features get judged by the same standard your PR is getting judged by. 

You’ve displayed a dismissive and combative attitude towards multiple team members of the development team now, taking their input and contributions for granted. If you continue with this disrespectful behavior, we flat out won’t accept this PR, even if it’s the greatest thing we’ve ever seen. Knock it off, and react accordingly to the input of the lead programmer of the team.

hhyyrylainen:
Thank you, Deus. I don't like demanding tone when *asking* for a PR to be merged.

And just to be clear, these aren't just arbitrary requirements, but ones for keeping the game codebase manageable, because Thrive will absolutely turn into a spaghetti mess that would eventually make any significant game progress impossible to do *unless* we stay vigilant and improve the quality of all changes before they are finalized.

Also from my point of view having done some major overhauls to Thrive somewhat recently, it is very bad when I need to update some code that is not sufficiently documented / I don't understand it and end up making mistakes as a result. So that's an extra point why I want contributions from other people to be polished to close to perfection, because if there is a problem *I'm* the one who has to deal with that as most people disappear from the project after a year or two, and many bugs are discovered only after multiple years. 

So as such I'm asking that you would please do those few more changes to the flexibility of this feature so that future developers (and balance tweakers, like Deus) will have a much easier time adjusting this feature. Though, again I'm like really burnt out with reviewing this PR, so could you please then ask someone else to review your work before I would do further checks?

domin2ktr:
I decided to reorganize my fork, since I can't make it so that the pr is copied from the branch, and not the fork itself, now the solution to the kernel paradox is transferred here:
https://github.com/Revolutionary-Games/Thrive/pull/6595

PR #6164, opened 2025-06-02
Make cell type's hash depend on the type's color

dligr:

Fixes the issue in the multicellular editor where a cell icon doesn't change its color after it is changed.

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/multicellular_stage/CellType.cs
+++ b/src/multicellular_stage/CellType.cs
@@ -150,10 +150,12 @@ public class CellType : ICellDefinition, ICloneable
 
     public ulong GetVisualHashCode()
     {
+        var hash = Colour.GetVisualHashCode();
+
         // This code is copied from MicrobeSpecies
         var count = Organelles.Count;
 
-        ulong hash = PersistentStringHash.GetHash(MembraneType.InternalName) * 5743 ^
+        hash ^= PersistentStringHash.GetHash(MembraneType.InternalName) * 5743 ^
             (ulong)MembraneRigidity.GetHashCode() * 5749 ^ (IsBacteria ? 1UL : 0UL) * 5779UL ^ (ulong)count * 131;
 
         var list = Organelles.Organelles;


src/multicellular_stage/CellType.cs
```diff
@@ -150,10 +150,12 @@ public object Clone()
 
     public ulong GetVisualHashCode()
     {
+        var hash = Colour.GetVisualHashCode();
+
         // This code is copied from MicrobeSpecies
         var count = Organelles.Count;
 
-        ulong hash = PersistentStringHash.GetHash(MembraneType.InternalName) * 5743 ^
+        hash ^= PersistentStringHash.GetHash(MembraneType.InternalName) * 5743 ^
```
Patryk26g: maybe lets change it to:
```
// This code is copied from MicrobeSpecies
var count = Organelles.Count;

var hash = Colour.GetVisualHashCode();
hash ^= PersistentStringHash.GetHash(MembraneType.InternalName) * 5743;
hash ^= (ulong)MembraneRigidity.GetHashCode() * 5749;
hash ^= (IsBacteria ? 1UL : 0UL) * 5779UL;
hash ^= (ulong)count * 131;
```
to make it look more readible

src/multicellular_stage/CellType.cs
```diff
@@ -153,8 +153,11 @@ public ulong GetVisualHashCode()
         // This code is copied from MicrobeSpecies
         var count = Organelles.Count;
 
-        ulong hash = PersistentStringHash.GetHash(MembraneType.InternalName) * 5743 ^
-            (ulong)MembraneRigidity.GetHashCode() * 5749 ^ (IsBacteria ? 1UL : 0UL) * 5779UL ^ (ulong)count * 131;
+        var hash = Colour.GetVisualHashCode();
```
hhyyrylainen: I think there's a slight need to reorder things here, because above the comment says "This code is copied from MicrobeSpecies" and now this isn't copied. So I suggest copying the exact hash start, and *then* having code like:

```
// Need to additionally add colour to the hash to allow cell type colour updates to work well
hash ^=  Colour.GetVisualHashCode();
```

hhyyrylainen (approved):
Seems good to me now.

PR #6165, opened 2025-06-02
Fix the engulfing system using capacity information from a wrong cell

### LINKED ISSUE #5098: Multicellular ingestion incorrectly checks capacity in lead cell
I haven't looked at the code so I'm opening this issue purely based on: https://community.revolutionarygamesstudio.com/t/0-6-6-multicellular-ingested-matter-storage-limited-to-core-cell/7450

I suspect there's some place in the ingestion systems that incorrectly checks the wrong cell for capacity when starting the ingestion and fails due to that incorrectly. I'm marking this only an intermediate difficulty as if that turns out to be the bug then it is a relatively simple fix.

Second / third confirmation of this on Discord: https://discord.com/channels/228300288023461893/958598553389903913/1334634349856821299

dligr:

Makes the engulfing system use the proper `Engulfer`. That is, the engulfing cell itself, instead of the colony leader.

In addition, this PR renames a variable to prevent confusion between the variable called `realEngulfer` (`Entity` type) and the `Engulfer` type variables.

**Related Issues**

Fixes #5098

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/EngulfingSystem.cs
+++ b/src/microbe_stage/systems/EngulfingSystem.cs
@@ -1062,7 +1062,7 @@ public sealed class EngulfingSystem : AEntitySetSystem<float>
                 continue;
 
             // Pili don't block engulfing, check starting engulfing
-            var realEngulfer = entity;
+            var actualEntity = entity;
 
             if (resolveColony)
             {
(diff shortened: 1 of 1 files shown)


src/microbe_stage/systems/EngulfingSystem.cs
```diff
@@ -1081,13 +1081,14 @@ private void CheckStartEngulfing(ref CollisionManagement collisionManagement, re
             }
 #endif
 
-            ref var actualEngulfer = ref engulfer;
+            ref var actualEngulfer =
+                ref actualEntity == entity ? ref engulfer : ref actualEntity.Get<Engulfer>();
```
hhyyrylainen, line 1085: It's probably safe to assume that the actual engulfing entity always has the Engulfer component, as anything else would be an extremely weird entity setup. I would have probably tried to add some kind of error checking but that would be likely quite cumbersome with the ref usages here. So I think this is fine for now, especially as I need to get through a ton of PR reviews. I'll assume that this functions correctly as I don't have time for playtesting.

hhyyrylainen (approved):

PR #6172, opened 2025-06-05
Engulfment Rebalancing

Deus-Codes:

Current changes include:

* Increased size discrepancy needed to engulf something 
* Slightly sped up the rate at which engulfed chunks are turned into compounds 
* Slightly reduced amount of compounds retrieved from engulfed items 
* Slightly debuffed amount of digestion time decrease from enzymes

This makes baseline engulfment just a touch less powerful, making lysosomes more of a worthy investment. The effort I'm going for is to make your ingested material go away quicker, so that you get the bonus you deserve for a successful hunt, but don't get to survive off of one or two hunts for the entire generation. This incentivizes adaptations to become a more voracious hunter - incentivizing pulling cilia, lysosomes, etc. -  instead of utilizing digestion as a one-and-done sort of thing. 

I'm experimenting a bit with the compound digested percentage and speed, so that could be subjected to change - regardless, a change to the size discrepancy ratio is needed.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -883,7 +883,7 @@ public static class Constants
     /// <summary>
     ///   The minimum size ratio between a cell and a possible engulfing victim.
     /// </summary>
-    public const float ENGULF_SIZE_RATIO_REQ = 1.5f;
+    public const float ENGULF_SIZE_RATIO_REQ = 2.5f;
 
     public const float EUKARYOTIC_ENGULF_SIZE_MULTIPLIER = 2.0f;
 
@@ -915,12 +915,12 @@ public static class Constants
     /// <summary>
     ///   The speed of which a cell can absorb compounds from digestible engulfed objects.
     /// </summary>
-    public const float ENGULF_COMPOUND_ABSORBING_PER_SECOND = 0.3f;
+    public const float ENGULF_COMPOUND_ABSORBING_PER_SECOND = 0.5f;
 
     /// <summary>
     ///   How much compounds in relation to real compound amount can be absorbed from digestible engulfed objects.
     /// </summary>
-    public const float ENGULF_BASE_COMPOUND_ABSORPTION_YIELD = 0.5f;
+    public const float ENGULF_BASE_COMPOUND_ABSORPTION_YIELD = 0.3f;
 
     /// <summary>
     ///   How long can cell be in engulf mode after activating without ATP
@@ -958,7 +958,7 @@ public static class Constants
     /// <summary>
     ///   Each enzyme addition grants a fraction, set by this variable, increase in digestion speed.
     /// </summary>
-    public const float ENZYME_DIGESTION_SPEED_UP_FRACTION = 0.1f;
+    public const float ENZYME_DIGESTION_SPEED_UP_FRACTION = 0.05f;
 
     /// <summary>
     ///   Each enzyme addition grants this fraction increase in compounds yield.


domin2ktr:
I think we need to increase the amount of glucose we get to make heterotrophy truly playable, since right now heterotrophy feels like it gives very little glucose (this is especially noticeable in the late microbe stage and multicelluar stage).

Deus-Codes:
> I think we need to increase the amount of glucose we get to make heterotrophy truly playable, since right now heterotrophy feels like it gives very little glucose (this is especially noticeable in the late microbe stage and multicelluar stage).

I think if we tweak how fast the nutrients get released, even smaller amounts of glucose feel more meaningful (.5 released over 2 seconds compared to released over 4 seconds makes a sizable difference). Right now, I dropped the amount of compounds released by .1 compared to the original, but this is subject to change. I think it feels fine right now considering the speed increase, but I'm more adept at Thrive compared to other people, so that could be subjective. I welcome playtesting feedback

domin2ktr:
I think we need to increase the amount of glucose we get to make heterotrophy truly playable, since right now heterotrophy feels like it gives very little glucose (this is especially noticeable in the late microbe stage and multicelluar stage).

Deus-Codes:
Intended to make this prior commit its own pull request to neatly organize movement changes, but I guess this will do here for now. Fits in to making predation a bit more viable anyways

Patryk26g:
I really like the size difference changes but in my opinion speeding up the rate of digestion is not a good idea

hhyyrylainen:
Seems like there's still some ongoing discussion about the changes.

I'll say that the community members often have trouble knowing the size difference:

> Increased size discrepancy needed to engulf something

So if it isn't 50% bigger, then nobody will remember it easily and we'll be spreading misinformation... so I'm kind of not on board with that change as it would be a pain to make sure everyone knows that change and going through the Thriveopedia and other in-game text to make sure there's no outdated information.

simulation_parameters/common/difficulty_presets.json
```diff
@@ -3,9 +3,9 @@
     "Name": "DIFFICULTY_PRESET_EASY",
     "Index": 0,
     "MPMultiplier": 0.8,
-    "AIMutationMultiplier": 0.8,
+    "AIMutationMultiplier": 1.3,
```
hhyyrylainen: I think it doesn't make sense for the AI to mutate faster on easy mode than on normal mode.

simulation_parameters/Constants.cs
```diff
@@ -883,9 +883,9 @@ public static class Constants
     /// <summary>
     ///   The minimum size ratio between a cell and a possible engulfing victim.
     /// </summary>
-    public const float ENGULF_SIZE_RATIO_REQ = 1.5f;
+    public const float ENGULF_SIZE_RATIO_REQ = 2.5f;
 
-    public const float EUKARYOTIC_ENGULF_SIZE_MULTIPLIER = 2.0f;
+    public const float EUKARYOTIC_ENGULF_SIZE_MULTIPLIER = 2.5f;
```
hhyyrylainen: This has the potential to cause a visual size inconsistency as this is kind of meant to match the 2x visual size multiplier...

simulation_parameters/common/difficulty_presets.json
hhyyrylainen: Also @GameDungeon mentioned on Discord that we really shouldn't have the mutation multiplier above 1 for any of the preset difficulties.

PR #6174, opened 2025-06-05
Mucocyst activation cost

0HyperCube:

There is now a cost of mucilage for initial activation of the mucocyst shield. This replaces the minimum amount to activate the shield.

If you do not have enough mucilage, a message will appear:
not enough mucilage to activate mucocyst shield

The cost is set at half of the total mucilage capacity. This probably needs further balancing.

**Related Issues**

As suggested by [@Deus-Codes on a developer discussion](https://forum.revolutionarygamesstudio.com/t/wrapping-up-the-microbe-stage/1170/4).

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Deus-Codes:
Awesome work to do this quick. I'll playtest at some point, but off the top of my head, I think something like 2 minimum should be spent. 

Ideally, I think it would be great if something like half of your max storage has to be spent to enter torpor (so if you have 16 max storage, you need atleast 8 to enter, if you have 4, you need atleast 2 to enter, etc.) so that effectively scales across multiple sizes. A flat rate is tough on smaller organisms, but for larger organisms, they can easily eat that up. 

If that is too bothersome however, don't feel the need to take it upon yourself. This is great work regardless, and apologies for not specifying that in my development post, didn't realize someone would be on the job so quickly.

0HyperCube:
I've updated it to half of the capacity as suggested.

Deus-Codes:
Awesome work, great functionality for a shield mechanic.

hhyyrylainen:
From briefly reading the previous discussion, it seems like this is done, but the PR is marked as a draft?

hhyyrylainen (approved):
The only slight thing I'd maybe want to tweak is the wording of the notice message, but other than that this seems a completely ready change to merge.

PR #6176, opened 2025-06-07
Restrict tab switching in the editor when there is an in-progress action

### LINKED ISSUE #3169: Switching editor tabs with an in-progress action should be prevented
*however* loading a save with in-progress action *must* set the right tab, so some special logic regarding preventing changing the tab is needed to allow tabs to change when loading a save with in-progress action.

Also I think this should be done for the microbe editor as well.

dligr:

Removes the possibility to switch tab in an editor component when there's an in-progess action, such as moving a cell.

This should work in the microbe, multicellular, and macroscopic editors.

**Related Issues**

Closes #3169

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/macroscopic_stage/editor/MetaballBodyEditorComponent.cs
```diff
@@ -1009,6 +1009,17 @@ private void SetSelectionMenuTab(string tab)
 
         GUICommon.Instance.PlayButtonPressSound();
 
+        // Don't change the tab if there's an in-progress action
+        if (CanCancelAction)
+        {
+            ToolTipManager.Instance.ShowPopup(Localization.Translate("ACTION_BLOCKED_WHILE_ANOTHER_IN_PROGRESS"),
```
hhyyrylainen: I think this kind of would warrant a special message, something like `TAB_CHANGE_BLOCKED_WHILE_ACTION_IN_PROGRESS`.

And this few lines of code I think could be put into the base editor component so that it would be easy to use.

Also this might not be the case but I have a slight feeling that there might be a save loading problem here. Because if I remember right in-progress actions are saved so when loading such a save the correct tab might not be properly applied before the in-progress action is loaded and that could then result in a softlock of the game.

dligr: I have changed the tooltip text, but I'm wondering if it's actually necessary to separate the code into a function. It's quite simple, and since the `ApplySelectionMenuTab()` call is tied to the function's logic, separating it might obscure the flow.

hhyyrylainen: I mean the line of code is long and it still would have the specific popup text key, which is also long? So I think it has enough unique details to warrant a single helper method that keeps all uses unified.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -1403,6 +1403,17 @@ private void SetSelectionMenuTab(string tab)
...
         GUICommon.Instance.PlayButtonPressSound();
 
+        // Don't change the tab if there's an in-progress action
+        if (CanCancelAction)
+        {
+            ToolTipManager.Instance.ShowPopup(Localization.Translate("TAB_CHANGE_BLOCKED_WHILE_ACTION_IN_PROGRESS"),
+                1.5f);
+
+            ApplySelectionMenuTab();
+
```
hhyyrylainen: This is 4 lines of duplication if counting the comment. So I think that this is enough stuff to warrant making a helper method to ensure that someone doesn't edit one of the comments for example adding more context or something like that. Other than that and the merge conflicts I think this PR is ready.

hhyyrylainen (approved):
Code seems good to me now.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
dligr: I have made a function that shows the tooltip and returns true if the tab switch has to be blocked; on the component realization side, the the tab is only changed if the function returns false.

PR #6179, opened 2025-06-08
Fix iron meteor

### LINKED ISSUE #6178: Iron meteor doesn't work
I just made the meteors all iron and set them to happen in all patches for testing. I assume you will find this if you encounter an iron meteor organically.

Relevant log:
```
Meteor impact event encountered patch with unexpectedly no Iron
```
Printing the `patch.Biome.ChangeableCompounds.Keys` yields:
```
Ammonia,Glucose,Phosphates,Hydrogensulfide,Oxygen,Carbondioxide,Nitrogen,Sunlight,Temperature
```
Notably iron is missing. This means that the meteor will not actually end up adding iron.

This list appears to be specified in `biomes.json` inside the "Compounds" inside "Conditions".

Possible fixes:
- Add the zero level of iron to all biomes.
- Adjust the `MeteorImpactEvent.cs` to automatically create an empty entry for iron inside the ChangeableCompounds so it doesn't fail.

0HyperCube:

The iron meteor previously failed to add iron with the following error:
```
Meteor impact event encountered patch with unexpectedly no Iron
```

This PR aims to fix that by automatically adding iron if it doesn't exist already.

**Related Issues**

Closes #6178

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/world_effects/patch_events/MeteorImpactEvent.cs
+++ b/src/general/world_effects/patch_events/MeteorImpactEvent.cs
@@ -192,14 +192,10 @@ public class MeteorImpactEvent : IWorldEffect
 
         foreach (var (compoundName, levelChange) in selectedMeteor.Compounds)
         {
-            bool hasCompound =
-                patch.Biome.ChangeableCompounds.TryGetValue(compoundName, out var currentCompoundLevel);
+            // Add 0 value for compound if it doesn't exist (e.g. for iron)
+            patch.Biome.ChangeableCompounds.TryAdd(compoundName, default);
 
-            if (!hasCompound)
-            {
-                GD.PrintErr($"Meteor impact event encountered patch with unexpectedly no {compoundName.ToString()}");
-                return;
-            }
+            var currentCompoundLevel = patch.Biome.ChangeableCompounds[compoundName];
 
             var definition = SimulationParameters.Instance.GetCompoundDefinition(compoundName);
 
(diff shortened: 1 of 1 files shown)


src/general/world_effects/patch_events/MeteorImpactEvent.cs
```diff
@@ -192,14 +192,10 @@ private void AdjustCompounds(Patch patch)
 
         foreach (var (compoundName, levelChange) in selectedMeteor.Compounds)
         {
-            bool hasCompound =
-                patch.Biome.ChangeableCompounds.TryGetValue(compoundName, out var currentCompoundLevel);
+            // Add 0 value for compound if it doesn't exist (e.g. for iron)
+            patch.Biome.ChangeableCompounds.TryAdd(compoundName, default);
```
hhyyrylainen: This should not use this like this directly. Instead even when the compound is missing `patch.Biome.ApplyLongTermCompoundChanges` should be used. The only slightly difficult thing is if the compound is a cloud type, then a cloud size value will need to be pulled from somewhere. I think it might be fine to hard code it here, or use the approach of looking in a template biome type (like some other events do when setting up new data).

hhyyrylainen (approved):
I adjusted the approach in this PR. The code was already setup to handle adding new compounds, but wasn't allowed to do so. So I basically just reverted most of the changes and removed that prevention of adding new compounds. I tested in the auto-evo explorer tool and in there everything seems fine with these changes now.

Patryk26g (approved):

PR #6185, opened 2025-06-11
Prevent chunks from spawning too close to the player

### LINKED ISSUE #5094: Floating chunks can spawn inside the player cell colony (which should be prevented)
My guess is that when the entity limit is pretty much full (which more often happens in early multicellular with massive body plans), chunks can get queued and then spawn way later once the player has had time to move over to where the spawn happens.

https://www.reddit.com/r/thrive/comments/1cgrsoh/comment/l1xm0d0/

I've had an idea for a while that the spawn system should skip spawns that end up too close to the player's new position. I think I didn't implement that based on this bug being able to happen but I didn't double check. If that was implemented then this will be a much trickier bug to fix.

dligr:

Implements a distance check before chunks are spawned to prevent them from spawning too close to the player or inside the player colony.

**Related Issues**

Fixes #5094

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/SpawnQueue.cs
+++ b/src/general/SpawnQueue.cs
@@ -106,12 +106,21 @@ public class SingleItemSpawnQueue : SpawnQueue
 {
     private readonly Factory factory;
 
-    public SingleItemSpawnQueue(Factory factory, Spawner fromSpawnType) : base(fromSpawnType)
+    private readonly CheckTooCloseToPlayer cancelCheck;
+
+    private Vector3 spawnLocation;
+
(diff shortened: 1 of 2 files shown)


hhyyrylainen (approved):
This makes quite a lot of sense. I was initially confused as this looked like something that should have already been in the game, but on further reading I realized you hooked up the existing system to the single spawn item class, which I apparently had forgotten to do when writing this too close prevention code years ago. So thanks for the fix. (though I'm not playtesting this again as I have a really big backlog of stuff to look through today)

PR #6186, opened 2025-06-13
Fix the toxin firing conditions

### LINKED ISSUE #3193: Signaling agents only in another colony member should be usable
The player should be able to use a signaling agent from other members in their colony to issue commands.

dligr:

Makes it possible for colonies to use signaling agents, even if they are placed in a non-core cell.

Also refactors a related function, making it check for special organelles' presence instead of counting them.

**Related Issues**

Closes #3193

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/components/MicrobeColony.cs
+++ b/src/microbe_stage/components/MicrobeColony.cs
@@ -230,22 +230,26 @@ public static class MicrobeColonyHelpers
     }
 
     /// <summary>
-    ///   Calculates the total counts of special organelles in a colony
+    ///   Returns the information about the presence of special organelles in a colony
     /// </summary>
-    public static void CalculateColonySpecialOrganelles(this ref MicrobeColony colony, out int agentVacuoles,
-        out int slimeJets, out int mucocysts)
+    public static void GetColonySpecialOrganelles(this ref MicrobeColony colony, out bool agentVacuoles,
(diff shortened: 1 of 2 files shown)


src/microbe_stage/components/MicrobeColony.cs
```diff
@@ -230,22 +230,25 @@ public static float CalculateTotalEngulfStorageSize(this ref MicrobeColony colon
     }
 
     /// <summary>
-    ///   Calculates the total counts of special organelles in a colony
+    ///   Returns the information about the presence of special organelles in a colony
     /// </summary>
-    public static void CalculateColonySpecialOrganelles(this ref MicrobeColony colony, out int agentVacuoles,
-        out int slimeJets, out int mucocysts)
+    public static void GetColonySpecialOrganelles(this ref MicrobeColony colony, out bool hasAgentVacuoles,
```
hhyyrylainen, line 256: For stuff like auto-evo I still think it is useful to get a total count of the organelles. So I'd like to keep a `CalculateColonySpecialOrganelles` method, but for usability from elsewhere I think it is a fine enough idea to create a wrapper called `GetColonySpecialOrganelles`.

hhyyrylainen (approved):
Code change looks good now. I expect this works correctly so I'll merge this now.

PR #6188, opened 2025-06-14
Reset scene after returning from editor

0HyperCube:

Reset scene after returning from editor

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/MicrobeStage.cs
+++ b/src/microbe_stage/MicrobeStage.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Components;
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
This not only resets the scene but also resets any stored compounds the player had, which is pretty much a bug because we have settings for what to do to compounds after reproduction.

I know some people want this kind of mode, but I really don't, so I'll accept this but only if there's a setting added to toggle between consistent environment between generations and the environment always resetting. And for now I'd like that setting to keep the old behaviour, that way we can collect feedback first specifically from players who try playing with the different mode.

0HyperCube:
Well it doesn't make sense for a new individual from a different location to have the same compounds as previously.

Since you are clearly not receptive to resetting the scene, I'll close this PR.

PR #6189, opened 2025-06-14
Setting menu rework 

Patryk26g:

Redign of the oprions nemu to make it fit planet generation style. Probably needs some more testing to ensure that all options work properly

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/OptionsMenu.tscn
```diff
@@ -275,34 +273,47 @@ size_flags_horizontal = 3
...
-[node name="Vsync" type="CheckBox" parent="CenterContainer/VBoxContainer/VBoxContainer/VBoxContainer/PanelContainer/Graphics/ScrollContainer/MarginContainer/VBoxContainer"]
-layout_mode = 2
-size_flags_horizontal = 0
-text = "VSYNC"
-
 [node name="FocusGrabber" parent="CenterContainer/VBoxContainer/VBoxContainer/VBoxContainer/PanelContainer/Graphics/ScrollContainer/MarginContainer/VBoxContainer" instance=ExtResource("13")]
 layout_mode = 2
 Priority = 3
-NodeToGiveFocusTo = NodePath("../Vsync")
+NodeToGiveFocusTo = NodePath("")
```
dligr: This change causes exceptions to be thrown, as FocusGrabber requires this field to be non-empty


<details>
 <summary>The exception</summary>
 E 0:00:05:831   void FocusGrabber._Ready(): System.ArgumentException: Focus grabber must have the node to focus set
  <C# Error>    System.ArgumentException
  <C# Source>   FocusGrabber.cs:60 @ void FocusGrabber._Ready()
  <Stack Trace> FocusGrabber.cs:60 @ void FocusGrabber._Ready()
                Node.cs:2542 @ bool Godot.Node.InvokeGodotClassMethod(Godot.NativeInterop.godot_string_name&, Godot.NativeInterop.NativeVariantPtrArgs, Godot.NativeInterop.godot_variant&)
                CanvasItem.cs:1654 @ bool Godot.CanvasItem.InvokeGodotClassMethod(Godot.NativeInterop.godot_string_name&, Godot.NativeInterop.NativeVariantPtrArgs, Godot.NativeInterop.godot_variant&)
                Control.cs:3017 @ bool Godot.Control.InvokeGodotClassMethod(Godot.NativeInterop.godot_string_name&, Godot.NativeInterop.NativeVariantPtrArgs, Godot.NativeInterop.godot_variant&)
                FocusGrabber_ScriptMethods.generated.cs:108 @ bool FocusGrabber.InvokeGodotClassMethod(Godot.NativeInterop.godot_string_name&, Godot.NativeInterop.NativeVariantPtrArgs, Godot.NativeInterop.godot_variant&)
                CSharpInstanceBridge.cs:24 @ Godot.NativeInterop.godot_bool Godot.Bridge.CSharpInstanceBridge.Call(nint, Godot.NativeInterop.godot_string_name*, Godot.NativeInterop.godot_variant**, int, Godot.NativeInterop.godot_variant_call_error*, Godot.NativeInterop.godot_variant*)

</details>

src/general/OptionsMenu.tscn
```diff
@@ -580,164 +634,275 @@ popup/item_1/id = 1
...
+layout_mode = 2
+size_flags_horizontal = 3
+mouse_filter = 0
 text = "DISPLAY_BACKGROUND_DISTORTION_EFFECT"
 
-[node name="LowQualityBlur" type="CheckBox" parent="CenterContainer/VBoxContainer/VBoxContainer/VBoxContainer/PanelContainer/Graphics/ScrollContainer/MarginContainer/VBoxContainer"]
+[node name="BackgroundDistortion" type="CheckButton" parent="CenterContainer/VBoxContainer/VBoxContainer/VBoxContainer/PanelContainer/Graphics/ScrollContainer/MarginContainer/VBoxContainer/VBoxContainer14/HBoxContainer"]
 layout_mode = 2
-size_flags_horizontal = 0
-tooltip_text = "LOW_QUALITY_BACKGROUND_BLUR_TOOLTIP"
```
dligr, line 651: Some tooltips on the former CheckBoxes appear to have been removed

dligr (commented):
The settings options seem to work based on my playtesting, though I have noticed some issues

src/general/OptionsMenu.tscn
Patryk26g: Thanks, I readded the missing tooltips

src/general/OptionsMenu.tscn
Patryk26g: Fixed this as well, I added fields to focus grabbers

src/general/OptionsMenu.tscn
```diff
@@ -19,31 +19,28 @@
...
 process_mode = 3
 layout_mode = 3
 anchors_preset = 15
 anchor_right = 1.0
 anchor_bottom = 1.0
 grow_horizontal = 2
 grow_vertical = 2
+scale = Vector2(0.976858, 1.00331)
```
hhyyrylainen: Why is there scaling here? I think this is going to cause blurriness issues. Or at least scaling is used nowhere else in the Thrive GUI so I really thinks this part should also be done without scaling for consistency sake to not have weird issues only affecting hte options menu.

hhyyrylainen:
One problem in general I noticed is that the toggle button focused border style colour doesn't match the drop down border colour, and as such it is hard to see now in the options menu when the initial vsync option is focused.

hhyyrylainen:
kuva

I think these added dividers would look better if the spacing above them was the same as below them as they now look a bit weird in my eyes with it seeming to be closer to the above content than the below.

hhyyrylainen:
Also there's now a stronger case for the sliders to be disabled when the relevant check buttons are turned off. For example the chromatic aberration slider could be made disabled when the button for that is turned off. The reason why this was not done before was that that broke the keyboard navigation flow, but now with separate toggle buttons that is no longer relevant. So the sliders could now be made to disable themselves when turned off entirely.

I noticed a few keyboard navigation inconsistencies that should be improved like jumping from the colourblind correction mode down should probably go to the chromatic aberration toggle button rather than the amount slider.

locale/en.po
```diff
@@ -5297,6 +5297,9 @@ msgstr "Silica"
 msgid "SILICA_MEMBRANE_DESCRIPTION"
 msgstr "This membrane has a strong wall of silica. It can resist overall damage well and is very resistant to physical damage. It also requires less energy to maintain its form. However, it slows the cell down by a large factor and the cell absorbs resources at a reduced rate."
 
+msgid "SIMULATION"
```
hhyyrylainen: I think this needs a *slightly* more descriptive name, perhaps "SIMULATION_CONFIG" might be good? Or "SIMULATION_TITLE"

src/general/OptionsMenu.tscn
Patryk26g: To be honest I'm not sure how it got there, sorry for that

Patryk26g:
Thanks for the review, hopefully I adressed all of the issues you mentioned

hhyyrylainen (changes requested):
I think what I commented before is fixed, but for some reason the buttons that are at the bottom of the settings menu are now gone?

kuva

Though additionally I think the CPU count and thread count would look better aligned all the way to the right than being just a bit to the right of the positions they are at.

The focus grabber on the inputs tab is now improperly setup as it should put the focus on the "inverted" checkbox rather than the "bind axes together" which is no longer the top item.

The focus moving from screen effect to the "built at" text label is broken. That's an intentional feature so that controller players can find the date when Thrive was built. Without that feature you cannot scroll the options menu down enough on a controller to ever see that.

Patryk26g:
I fixed the rest of the issues

hhyyrylainen:
It seems like the default tab visibility was left in an incorrect state...

image

PR #6190, opened 2025-06-15
Fix ATP flickering

### LINKED ISSUE #4703: ATP bar no longer stays properly full instead it flickers near the top
Probably our cutoff for when to show the bar full is not good enough?

Or might be fine to swap around the order of the systems so that ATP is consumed first before the process system runs? If this is done then the system running multithreading attributes need to be set correctly for the fix to stick.

0HyperCube:

When there is a large lag spike, the ATP bar decreases randomly.

It is difficult to fix this by snapping the APT value to max since with a large lag spike, the ATP value can fall significantly e.g. 0.5 to 0.2.

I initially considered reordering the existing systems as suggested in the issue. This would involve the `ProcessSystem` running after the `OsmoregulationAndHealingSystem` and the `MicrobeMovementSystem`. However this caused many issues since systems like the `MicrobeEmissionSystem` `MicrobeAISystem` etc like to run between these. I found that it was extremly difficult to rearrange the systems. Perhaps a system like [bevy's system sets](https://bevy-cheatbook.github.io/programming/system-sets.html) to more easily rearrange them.

I decided to stick with the current system order. However I made it so that the process system uses the delta time from the previous frame (which is what the osmoregulation would use). This ensures that the ATP value is always set to exactly the right value at the end of the ProcessSystem; even when a large lag spike occurs.

In order to access the ATP value before it is changed further, I added a new system `PlayerATPRecorderSystem` that just records the ATP value.

This new approach removes the need to round the almost full bar, since the APT value is exactly full.

**Related Issues**

Closes #4703

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/common_systems/PlayerATPRecorderSystem.cs
+++ b/src/engine/common_systems/PlayerATPRecorderSystem.cs
@@ -1,0 +1,36 @@
+﻿namespace Systems;
+
+using Components;
+using DefaultEcs;
+using DefaultEcs.System;
+
+/// <summary>
+///   Records the player ATP for easy access. Run before osmoregulation to ensure it doesn't flicker.
+/// </summary>
+[With(typeof(PlayerMarker))]
+[With(typeof(CompoundStorage))]
+[RunsAfter(typeof(ProcessSystem))]
+[RunsBefore(typeof(OsmoregulationAndHealingSystem))]
+[RunsBefore(typeof(MicrobeMovementSystem))]
+[RuntimeCost(0.5f)]
+[RunsOnMainThread]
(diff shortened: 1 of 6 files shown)


hhyyrylainen:
Is there a specific reason why this was closed?

0HyperCube:
Since this was an extremely large change (adding a new system and 147 lines), I doubt it will ever get merged.

hhyyrylainen:
That doesn't sound like an extremely large change to me. I regularly review similarly sized PRs and they *do* get merged after the review cycle. But if you don't want to continue working on this PR, that's also totally fine.

PR #6193, opened 2025-06-17
Implement multicellular preview images

### LINKED ISSUE #3179: Make body plan editor remember exactly the hex layout created by the player
Right now there's a conversion when exiting from the editor to a layout that contains all of the cells with their hexes. Then when returning to the editor that layout is converted back into one where each cell is represented by a single hex. This isn't always exact so we need some way to ensure it would be exact.

One idea regarding fixing this that I've had is that we could store both layouts and work like that (but we still need code to go from the full layout back to the single layout and from the single layout tot the full layout to make editor work on any species).

https://github.com/Revolutionary-Games/Thrive/blob/ffbe19bfe09646c171f9f56ec5818c82dafc794f/src/early_multicellular_stage/editor/CellBodyPlanEditorComponent.cs#L237

https://github.com/Revolutionary-Games/Thrive/blob/ffbe19bfe09646c171f9f56ec5818c82dafc794f/src/early_multicellular_stage/editor/CellBodyPlanEditorComponent.cs#L262

This should allow the body plan to have holes in it.

### LINKED ISSUE #3899: Preview images for early multicellular species need to be implemented
Portraits for multicellular species are not displayed, they can not be fossilized or viewed in the evolutionary tree

**To Reproduce**
1. Become multicellular
2. View your species in the evolutionary window

**Screenshots**

evwindow

**Additional context**
The "Stage" stat (which is stage: microbe basically every cell) also doesn't display on the specie info tab

dligr:

Implements multicellular preview images.

Also makes the species remember the exact colony layout between editor sessions to make the hex preview work. Though it has no cell pictures for now because I couldn't find a way to implement that.

**Related Issues**

Closes #3899
Closes #3179 (to make the hex layout preview work)

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/CellHexesPreview.cs
+++ b/src/microbe_stage/CellHexesPreview.cs
@@ -1,23 +1,24 @@
 ﻿using System;
+using Godot;
 
 /// <summary>
 ///   Shows a visualization of a cell's hexes in the GUI
 /// </summary>
 public partial class CellHexesPreview : PhotographablePreview
 {
-    private MicrobeSpecies? microbeSpecies;
+    private Species? species;
 
-    public MicrobeSpecies? PreviewSpecies
+    public Species? PreviewSpecies
     {
-        get => microbeSpecies;
+        get => species;
         set
         {
             if (PreviewSpecies == value)
                 return;
 
-            microbeSpecies = value;
+            species = value;
 
-            if (microbeSpecies != null)
+            if (species != null)
             {
                 UpdatePreview();
             }
@@ -28,18 +29,36 @@ public partial class CellHexesPreview : PhotographablePreview
         }
     }
 
-    protected override IImageTask SetupImageTask()
+    protected override IImageTask? SetupImageTask()
     {
-        if (microbeSpecies == null)
+        if (species == null)
             throw new InvalidOperationException("No species set to generate image of hexes for");
 
-        var hash = CellHexesPhotoBuilder.GetVisualHash(microbeSpecies);
+        if (species is MicrobeSpecies microbeSpecies)
+        {
+            var hash = CellHexesPhotoBuilder.GetVisualHash(microbeSpecies);
 
-        var task = PhotoStudio.Instance.TryGetFromCache(hash);
+            var task = PhotoStudio.Instance.TryGetFromCache(hash);
 
-        if (task != null)
-            return task;
+            if (task != null)
+                return task;
 
-        return PhotoStudio.Instance.GenerateImage(new CellHexesPhotoBuilder { Species = microbeSpecies }, Priority);
+            return PhotoStudio.Instance.GenerateImage(new CellHexesPhotoBuilder { Species = microbeSpecies }, Priority);
+        }
+
+        if (species is MulticellularSpecies multicellularSpecies)
+        {
+            var hash = ColonyHexPhotoBuilder.GetVisualHash(multicellularSpecies);
+
+            var task = PhotoStudio.Instance.TryGetFromCache(hash);
+
+            if (task != null)
+                return task;
+
+            return PhotoStudio.Instance.GenerateImage(new ColonyHexPhotoBuilder { Species = multicellularSpecies }, Priority);
+        }
+
+        GD.PrintErr("Unknown species type to preview: ", species, " (", species.GetType().Name, ")");
+        return null;
     }
 }
--- a/src/microbe_stage/CellHexesPreview.cs
+++ b/src/microbe_stage/CellHexesPreview.cs
@@ -55,6 +55,12 @@ public partial class CellHexesPreview : PhotographablePreview
             if (task != null)
                 return task;
 
+            if (multicellularSpecies.EditorCellLayout == null)
+            {
+                GD.PrintErr("No cell layout is remembered, the hex preview can't be generated");
+                return null;
+            }
+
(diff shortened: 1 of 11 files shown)


src/microbe_stage/CellHexesPreview.cs
```diff
@@ -28,18 +29,43 @@ public MicrobeSpecies? PreviewSpecies
...
+            }
 
-        if (task != null)
-            return task;
+            return PhotoStudio.Instance.GenerateImage(new ColonyHexPhotoBuilder { Species = multicellularSpecies },
+                Priority);
+        }
 
-        return PhotoStudio.Instance.GenerateImage(new CellHexesPhotoBuilder { Species = microbeSpecies }, Priority);
+        GD.PrintErr("Unknown species type to preview: ", species, " (", species.GetType().Name, ")");
```
hhyyrylainen:
```suggestion
        GD.PrintErr("Unknown species type to generate hexes view of: ", species, " (", species.GetType().Name, ")");
```

src/multicellular_stage/ColonyHexPhotoBuilder.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        if (species.EditorCellLayout == null)
+        {
+            GD.PrintErr("No cell layout is remembered, the radius can't be calculated for the hex preview");
+            return;
+        }
+
+        foreach (var cell in species.EditorCellLayout)
+        {
+            farthest = MathF.Max(farthest, Hex.AxialToCartesian(cell.Position).DistanceTo(Vector3.Zero));
```
hhyyrylainen: Wouldn't it be slightly more efficient to calculate the farthest point in squared distance, and then only take square root once outside the loop?

src/multicellular_stage/MulticellularSpecies.cs
```diff
@@ -188,6 +199,54 @@ public override float GetPredationTargetSizeFactor()
...
+        foreach (var entity in worldSimulation.EntitySystem)
+        {
+            if (!entity.Has<CellProperties>())
+                continue;
+
+            ref var cellProperties = ref entity.Get<CellProperties>();
+
+            // This uses the membrane as radius is not set as the physics system doesn't run
+            if (!cellProperties.IsMembraneReady())
+                throw new InvalidOperationException("Microbe doesn't have a ready membrane");
```
hhyyrylainen: Shouldn't this be checked in `StateHasStabilized` as this truly may not be ready as when the membranes are calculated is kind of performance dependent?

hhyyrylainen: Though I guess the microbe simulation already has code assuming that. For putting similar code in a similar spot, I think this distance calculation would make sense to be in the MicrobeVisualOnlySimulation class as the microbe variant of this method is there.

src/multicellular_stage/MulticellularSpecies.cs
```diff
@@ -188,6 +199,54 @@ public override float GetPredationTargetSizeFactor()
...
+
+            var cellRadius = cellProperties.CreatedMembrane!.EncompassingCircleRadius;
+
+            var farthestPoint = entity.Get<WorldPosition>().Position.DistanceTo(center) + cellRadius;
```
hhyyrylainen: Same squared distance comparison optimization could be done here as well I think.

src/multicellular_stage/MulticellularSpecies.cs
```diff
@@ -210,6 +269,20 @@ public override object Clone()
...
 
+    public override ulong GetVisualHashCode()
+    {
+        ulong hash = 1099511628211;
+
+        foreach (var cell in Cells)
+        {
+            hash ^= cell.GetVisualHashCode() ^ (ulong)cell.Position.GetHashCode();
+
+            hash = (hash << 7) | (hash >> 57);
```
hhyyrylainen: Doesn't this mean that the order of the cells in the list affects the hash? That seems incorrect to me as the order of the cells shouldn't affect the visuals as they are all eventually placed anyway...

src/multicellular_stage/MulticellularSpecies.cs
```diff
@@ -28,6 +29,12 @@ public MulticellularSpecies(uint id, string genus, string epithet) : base(id, ge
     [JsonProperty]
     public CellLayout<CellTemplate> Cells { get; private set; } = new();
 
+    /// <summary>
+    ///   The cell layout that this species has in the editor. Is null unless this colony created in the editor.
```
hhyyrylainen: I think for eventually we need to have auto-evo also create these layouts as otherwise the player can only really preview their own species which is not the most helpful... So I'd ever so slightly reword this comment.

hhyyrylainen (changes requested):
I think the overall approach is good and should be workable. I did notice a few minor things to clean up and improve efficiency of.

src/multicellular_stage/MulticellularSpecies.cs
hhyyrylainen: One more thought on this I just had: I have a slight recollection that I probably already made the world wait automatically if there are pending membranes, as that's probably needed for microbes. So this might not actually be a place that needs additional safety checks, but I still would like to see this code be moved so that it is easy to find the similar code.

hhyyrylainen (approved):
Code looks good now.

PR #6198, opened 2025-06-27
Increase autoevo diversity

adQuid:

The compound conversion score focuses only on the efficiency of conversion of those organelles, without factoring in osmoregulation and time. The problem with this seems to be that a species with one chloroplast is as efficient as a species focused around photosynthesis, for example. This change isn't a perfect simulation, but it is seeming to provide the desired diversity and specialization, particularly among autotrophs.

**Related Issues**

None

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/AutoEvoGlobalCache.cs
+++ b/src/auto-evo/AutoEvoGlobalCache.cs
@@ -43,20 +43,20 @@ public class AutoEvoGlobalCache
         EnvironmentalTolerancesPressure = new EnvironmentalTolerancePressure(4);
 
         MinorGlucoseConversionEfficiencyPressure =
-            new CompoundConversionEfficiencyPressure(Compound.Glucose, Compound.ATP, 0.75f);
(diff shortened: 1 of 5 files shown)


adQuid:
Having had more time to playtest this for a few runs, I feel comfortable sending this into proper peer review.

GameDungeon:
I don't see anywhere where the new peram `CompoundConversionEfficiencyPressure` is false. What is the main idea behind it?

adQuid:
> I don't see anywhere where the new peram `CompoundConversionEfficiencyPressure` is false. What is the main idea behind it?

the idea was that there might be a time someone wants to use compound conversion efficiency in a situation where the species isn't living off the compound (like if we had one from producing toxin, for example), and the developer didn't want speed to factor in. It is "future-proofing" though, so I see the case against it.

GameDungeon:
I was just confused and assumed I was missing something. Personally I think it is fine.

hhyyrylainen (approved):
The code change looks good to me and the JSON attributes should be right as well. So this can be merged if people are happy with the improved auto-evo results.

PR #6201, opened 2025-07-05
Make the time based reproduction work with colonies

### LINKED ISSUE #3655: Make time based reproduction work with early colony gameplay
right now it just completely halts the progress and the bottom right bars won't move at all

just dump the free compounds needed for growth into storage?
Or allow the first step of organelles preparing to split (can be done once https://github.com/Revolutionary-Games/Thrive/issues/3645 is done)? Maybe needs even something else entirely.

dligr:

Makes early colonies receive free compounds, so that they don't lose out on the free growth progress from the time based reproduction system.

**Related Issues**

Closes #3655

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/MicrobeReproductionSystem.cs
+++ b/src/microbe_stage/systems/MicrobeReproductionSystem.cs
@@ -230,8 +230,11 @@ public sealed class MicrobeReproductionSystem : AEntitySetSystem<float>
 
         if (isInColony)
         {
-            // TODO: should the colony just passively get the reproduction compounds in its storage?
-            // Otherwise multicellular colonies lose the passive reproduction feature
+            var (_, freeCompounds) = CalculateFreeCompoundsAndLimits(gameWorld!.WorldSettings, organelles.HexCount, false,
+                reproductionDelta);
+
+            AddFreeCompoundsToStorage(entity, freeCompounds);
+
             return;
         }
 
@@ -341,6 +344,28 @@ public sealed class MicrobeReproductionSystem : AEntitySetSystem<float>
         }
 
         requiredCompoundsForBaseReproduction[compound] = left;
+    }
+
+    private void AddFreeCompoundsToStorage(in Entity entity, float freeCompounds)
+    {
+        ref var baseReproduction = ref entity.Get<ReproductionStatus>();
+
+        ref var storage = ref entity.Get<CompoundStorage>();
+
+        float requiredCompoundSum = 0.0f;
+
+        if (baseReproduction.MissingCompoundsForBaseReproduction == null)
+            return;
+
+        foreach (var pair in baseReproduction.MissingCompoundsForBaseReproduction)
+        {
+            requiredCompoundSum += pair.Value;
+        }
+
+        foreach (var pair in baseReproduction.MissingCompoundsForBaseReproduction)
+        {
+            storage.Compounds.AddCompound(pair.Key, freeCompounds * (pair.Value / requiredCompoundSum));
+        }
     }
 
     /// <summary>


src/microbe_stage/systems/MicrobeReproductionSystem.cs
```diff
@@ -343,6 +346,28 @@ private static void ProcessBaseReproductionForCompoundType(Compound compound,
...
+        ref var baseReproduction = ref entity.Get<ReproductionStatus>();
+
+        ref var storage = ref entity.Get<CompoundStorage>();
+
+        float requiredCompoundSum = 0.0f;
+
+        if (baseReproduction.MissingCompoundsForBaseReproduction == null)
+            return;
+
+        foreach (var pair in baseReproduction.MissingCompoundsForBaseReproduction)
```
hhyyrylainen: Why does this only add compounds that are missing for base reproduction? Doesn't this stop working once the base reproduction cost is taken? If I remember right the new approach is that base cost is taken first and then organelle costs. So if the player is already duplicating their organelles and joins a colony, that will break this code...

dligr: The added compounds should now be based on species' base reproduction cost and organelles' `InitialComposition`.

dligr:
I have changed the approach to use either the organelle cost or the base reproduction cost as a basis of adding compounds (based on the reproduction stage). 
Though now, thinking about it, do we actually need this kind of precision (in the amounts of compounds given) to warrant this kind of code complexity? In my opinion, just adding `species.BaseReproductionCost` (as much as possible of it with the free compounds) should probably be good enough.

hhyyrylainen:
I agree that it is probably enough, however someone in the future might add a funny organelle that takes in like iron in its base cost. Then that approach would not work and it wouldn't give the right resources. One idea I had is that we could precompute the set of compounds and their ratios that a species needs in total to reproduce (base + all organelles) that precomputed ratio of compounds could be used to give stuff passively. But as I thought that that would be more complex to implement I didn't suggest that idea before. If you want we could still swap to that approach as I foresee some future usage like calculating optimal nutrient ratios in later stages or something like that. Though one additional hard part with that other approach is how to approach that data being missing (though in that case falling back to just base cost should be sufficient) when trying to preserve older save compatibility.

src/microbe_stage/MicrobeSpecies.cs
```diff
@@ -351,6 +351,20 @@ public override string GetDetailString()
                 BaseHexSize);
     }
 
+    protected override Dictionary<Compound, float> CalculateTotalReproductionCost()
+    {
+        var result = new Dictionary<Compound, float>(BaseReproductionCost);
```
hhyyrylainen: One small thing I think I just noticed: shouldn't this call `base.CalculateTotalReproductionCost`? Because now this just skips the base reproduction cost...

So this line should be `var result = base.CalculateTotalReproductionCost()`.

src/microbe_stage/systems/MicrobeReproductionSystem.cs
```diff
@@ -230,8 +230,25 @@ protected override void Update(float state, in Entity entity)
 
         if (isInColony)
         {
-            // TODO: should the colony just passively get the reproduction compounds in its storage?
-            // Otherwise multicellular colonies lose the passive reproduction feature
+            var (_, freeCompounds) = CalculateFreeCompoundsAndLimits(gameWorld!.WorldSettings, organelles.HexCount,
+                false, reproductionDelta);
+
+            var species = entity.Get<SpeciesMember>().Species;
```
hhyyrylainen: Technically this system should be marked as reading `SpeciesMember` component (even though this probably won't have a practical effect as this already is marked as reading both microbe species and multicellular species, still technically this component is a different type so that should be marked for completeness' sake).

hhyyrylainen (changes requested):
I think this is now a very good solution. I think I did see one bug and one minor code technical detail that would be good to still fix.

src/microbe_stage/systems/MicrobeReproductionSystem.cs
dligr: Actually, there's probably no reason not to get a specific `MicrobeSpeciesMember` component (it's the only species type handled by this system), so I made it get that instead.

hhyyrylainen (approved):
Code looks good now.

One minor thing I could say is that I think the foreach loops in MicrobeReproductionSystem probably allocate 2 enumerator objects per cell colony per game update, but I don't see an easy way to avoid those allocations in this situation.

Issue #4981, opened 2024-03-26
Check if physical keys should be used as the basis for rebindable keys instead of key labels


hhyyrylainen:
as maybe those are more reliable or something like that?

Issue #2887, opened 2021-12-01
Check the new Godot 3.4 physical scan codes


hhyyrylainen:
to see if we should update our key handling to take advantage of that

hhyyrylainen:
Seems like 3.4.1 also added some new stuff regarding this: https://github.com/godotengine/godot/pull/55251

Anonymous941:
What's hard about this, does it require a new version of Godot?

hhyyrylainen:
None of our input handling code is made with it in mind. Especially the keybindings rebinding menu might not work correctly at all. Only way to find out what all is broken is try to move to using the physical scan codes, and then fixing everything in the game that ends up being broken.

hhyyrylainen:
With the internal input key support for this, this should be now pretty easy to test and switch to by editing the input map: https://github.com/Revolutionary-Games/Thrive/pull/4979

Issue #2371, opened 2021-06-21
Fix keybinding conflict in perform unbinding with mouse click


hhyyrylainen:
It would be very nice to be able to have fire toxin be the left mouse button, currently can't due to a conflict. This needs to be resolved.

![2021-06-16_17 48 12 2721](https://user-images.githubusercontent.com/3796411/122781480-87420b80-d2b8-11eb-9686-bb6b33d1b9a1.png)

La-ryssa:
I don't think it's necessary to have a key for performing unbinding. This should be whichever button is used to select UI elements (left mouse button, A on xbox, etc)

hhyyrylainen:
I saw this notification a bit ago and have been thinking about this off and on since then, and I'm really not sure. Because while on one hand you are correct that the primary selection action could be used, it is still the case that for a keybinding system to be useful all actions need to be rebindable. For example if someone has some really weird pointing device or something they can use to point with but they don't have the primary mouse buttons on it? Then they'd be stuck unable to use the unbinding action. And for using the primary selection button (which BTW doesn't exist, the closest thing would be the editor primary action) it would logically then need to be added to the gameplay section of inputs as well, which would again mean conflicts.

One thing I came up with when thinking about possible solutions is to add a "non-conflicting keys" property to each input and it can specify which other inputs it doesn't mess with. For example if mouse clicks on different gameplay elements trigger different things those inputs can be bound to multiple actions at once.

I'll increase the difficulty of this issue as it is clear this is not very easy to solve.

Issue #1906, opened 2020-12-09
Make key rebinding controls use Godot signals instead of C#


hhyyrylainen:
It's nicer to use the normal Godot way for the Control hierarchy to communicate with each other. 
It's also fine to keep the current approach *if* the objects sent in the events don't actually make it through properly the serialization through godot internals (I know some object types can't go through there and instead error out).

make InputEventItem instead of calling InputGroupList.ControlsChanged to use Godot signals
move `public event ControlsChangedDelegate OnControlsChanged;` to be a Godot signal as it is contained in a Control

hhyyrylainen:
This seems pretty minor now 5 years later, so I'll mark this as non-gating.

PR #6202, opened 2025-07-05
Fix input rebinding

### LINKED ISSUE #5352: Rebinding a key with conflicts seems to no longer work

dligr:

Now the input actually binds to the action after clicking 'confirm' in the rebinding menu.

**Related Issues**

Fixes #5352

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/input/key_mapping/InputEventItem.cs
+++ b/src/engine/input/key_mapping/InputEventItem.cs
@@ -401,6 +401,13 @@ public partial class InputEventItem : MarginContainer
         OnKeybindingSuccessfullyChanged();
     }
 
+    public void FinishRebind(InputEvent inputEvent)
+    {
+        AssociatedEvent = new SpecifiedInputKey(inputEvent, true);
+
+        OnKeybindingSuccessfullyChanged();
+    }
+
     /// <summary>
     ///   Delete this event from the associated action and update the godot InputMap
     /// </summary>
--- a/src/engine/input/key_mapping/InputGroupList.cs
+++ b/src/engine/input/key_mapping/InputGroupList.cs
@@ -172,7 +172,7 @@ public partial class InputGroupList : VBoxContainer
         latestDialogConflict.Delete();
 
         // Pass the input event again to have the key be set where it was previously skipped
-        latestDialogCaller._Input(latestDialogNewEvent);
+        latestDialogCaller.FinishRebind(latestDialogNewEvent);
     }
 
     public bool IsConflictDialogOpen()
--- a/src/engine/input/key_mapping/SpecifiedInputKey.cs
+++ b/src/engine/input/key_mapping/SpecifiedInputKey.cs
@@ -483,7 +483,7 @@ public class SpecifiedInputKey : ICloneable
         return toStringBuilder.ToString();
     }
 
-    private void ConstructFrom(InputEventWithModifiers @event, bool preferKeyLabel)
+    private void ConstructFrom(InputEventWithModifiers @event, bool preferKeycodes)
     {
         Control = @event.CtrlPressed || @event.MetaPressed;
         Alt = @event.AltPressed;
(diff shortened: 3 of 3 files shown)


hhyyrylainen:
> Also, the input menu now uses key codes instead of key labels for binding, for the reason that otherwise, keybinds made on a non-latin layout wouldn't work on a latin layout (and vice versa). There's also an added difficulty in comparing key labels with key codes in which the default bindings are recorded.

Did you see the open issues related to that? I think you may have accidentally solved them / maybe done a different approach (I didn't read the code in this PR yet):
- https://github.com/Revolutionary-Games/Thrive/issues/2887
- https://github.com/Revolutionary-Games/Thrive/issues/4981

There's also this open issue which might be easy to do while changing the keyrebinding stuff anyway: https://github.com/Revolutionary-Games/Thrive/issues/1906

And one final relevant issue I found which may be a duplicate of the primary issue you marked this PR as solving: https://github.com/Revolutionary-Games/Thrive/issues/2371

src/engine/input/key_mapping/InputEventItem.cs
```diff
@@ -401,6 +401,13 @@ public override void _Input(InputEvent @event)
         OnKeybindingSuccessfullyChanged();
     }
 
+    public void FinishRebind(InputEvent inputEvent)
```
hhyyrylainen: I'm wondering if it would make sense for the above method (around line 387 to call this single helper method) to call this instead of duplicating the event assign and calling `OnKeybindingSuccessfullyChanged`

dligr:
> Did you see the open issues related to that? I think you may have accidentally solved them / maybe done a different approach (I didn't read the code in this PR yet):
> 
> * [Try to switch our input system to using physical scan codes #2887](https://github.com/Revolutionary-Games/Thrive/issues/2887)
> * [Check if physical keys should be used as the basis for rebindable keys instead of key labels #4981](https://github.com/Revolutionary-Games/Thrive/issues/4981)

I had originally made the input system use normal keycodes, but now I've now switched it to physical keys (which should support layouts like AZERTY without the need to rebind any keys).

dligr:
> there is some place that needs conversion from physical key codes to normal key codes for display so that we can show the right symbols for the player's keyboard.

Displayed input strings are currently converted from physical keys like this: `KeyNames.Translate(DisplayServer.KeyboardGetKeycodeFromPhysical(key))`. So there is a conversion, but it seems to consider some characters to be 'localized' ones, and return other symbols instead. E.g. in my testing, German ö was displayed as `

The alternative is to use `KeyboardGetLabelFromPhysical(Key)`, but that would result in all inputs being displayed as their localized versions, which might be undesirable for some languages.

hhyyrylainen:
A potential solution is to then use the non-physical keycodes when the user rebinds stuff. I imagine users don't regularly share their config files with each other so those shouldn't pose an issue when they are generated and used on the same computer / keyboard.

hhyyrylainen:
> The alternative is to use KeyboardGetLabelFromPhysical(Key), but that would result in all inputs being displayed as their localized versions, which might be undesirable for some languages.

We could have an option in the options menu to display those differently, but I think in general that should be the much more accurate way to display a key symbol to the user, as otherwise the key symbol might be completely off as to what key the user should press.

hhyyrylainen (approved):
I fixed that one more thing related to key prompts, and now at least on my computer this seems to now all work perfectly. Hopefully we'll remember these changes if any problems show up during public testing from people with strange locale setups.

hhyyrylainen:
In my testing I did notice a related issue that I think doesn't have yet an open issue, so I just opened this: https://github.com/Revolutionary-Games/Thrive/issues/6206

But I'll merge this PR as that's unrelated to the changes here.

PR #6209, opened 2025-07-12
Optimize GetMutationsForSpecies

GameDungeon:

Refactors the algorithm for GetMutationsForSpecies for a large speed increase.

This does have an effect on auto-evo output, and could contain some regressions due to the complexity of the patch, but everything appears to be working.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/steps/ModifyExistingSpecies.cs
+++ b/src/auto-evo/steps/ModifyExistingSpecies.cs
@@ -35,7 +35,6 @@ public class ModifyExistingSpecies : IRunStep
     private readonly Miche.InsertWorkingMemory insertWorkingMemory = new();
 
     private readonly List<Miche> predatorCalculationMemory2 = new();
-    private readonly List<Species> predatorPressuresTemporary = new();
 
     private readonly List<Mutant> temporaryMutations1 = new();
     private readonly List<Mutant> temporaryMutations2 = new();
@@ -50,8 +49,6 @@ public class ModifyExistingSpecies : IRunStep
 
     private readonly List<Mutation> mutationsToTry = new();
     private readonly HashSet<MicrobeSpecies> handledMutations = new();
-
-    private readonly int expectedSpeciesCount;
 
     private readonly List<Miche> nonEmptyLeafNodes = new();
     private readonly List<Miche> emptyLeafNodes = new();
@@ -74,7 +71,7 @@ public class ModifyExistingSpecies : IRunStep
         random = new XoShiRo256starstar(randomSeed.NextInt64());
 
         // Patch species count is used to know how many steps there are to perform
-        expectedSpeciesCount = patch.SpeciesInPatch.Count;
+        TotalSteps = patch.SpeciesInPatch.Count;
         speciesEnumerator = patch.SpeciesInPatch.GetEnumerator();
     }
 
@@ -95,7 +92,7 @@ public class ModifyExistingSpecies : IRunStep
     /// <summary>
     ///   See <see cref="Step"/> for explanation on the step count
     /// </summary>
-    public int TotalSteps => 4 + expectedSpeciesCount;
+    public int TotalSteps => 4 + field;
 
     public bool CanRunConcurrently => true;
 
@@ -144,8 +141,8 @@ public class ModifyExistingSpecies : IRunStep
                     // Reset this for another enumeration later
                     speciesEnumerator = patch.SpeciesInPatch.GetEnumerator();
 
-                    // Just for safety generate any mutations still missing in case the miche data and species in patch
-                    // are not in sync
+                    // Just for safety, generate any mutations still missing in case the miche data and species in
+                    // the patch are not in sync
                     foreach (var species in speciesWorkMemory)
                     {
                         if (patch.SpeciesInPatch.ContainsKey(species))
@@ -380,6 +377,8 @@ public class ModifyExistingSpecies : IRunStep
         pressureStack.Push(currentMiche.Pressure);
         mutationSorter.Setup(baseSpecies, pressureStack);
 
+        // TODO: avoid this temporary memory allocation somehow (this is slightly tricky as this method is called
+        // recursively)
         var mutations = currentMiche.Pressure.Mutations.ToArray();
         mutations.Shuffle(random);
 
@@ -559,10 +558,12 @@ public class ModifyExistingSpecies : IRunStep
     {
         // This isn't the cleanest, but this class is just optimized for performance, so if someone forgets to set up
         // this, then bad things will happen
-        private IEnumerable<SelectionPressure> pressures = null!;
+
+        // This directly references to the stack type to avoid an enumerator allocation in the foreach loop in Compare
+        private Stack<SelectionPressure> pressures = null!;
         private MicrobeSpecies baseSpecies = null!;
 
-        public void Setup(MicrobeSpecies species, IEnumerable<SelectionPressure> selectionPressures)
+        public void Setup(MicrobeSpecies species, Stack<SelectionPressure> selectionPressures)
         {
             pressures = selectionPressures;
             baseSpecies = species;


src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -382,48 +335,37 @@ private void PredatorsOf(List<Species> result, Miche micheTree, Species species)
...
-        tempMutationStrategies.Shuffle(random);
+        temporaryMutations1.Clear();
+        temporaryMutations1.AddRange(outputSpecies);
 
-        foreach (var mutationStrategy in tempMutationStrategies)
+        foreach (var mutationStrategy in mutations)
         {
-            var inputSpecies = viableVariants;
+            // temporaryMutations1.Clear();
+            // temporaryMutations1.AddRange(outputSpecies);
```
GameDungeon: This is currently a draft as there is currently a bug, where Pressures with multiple mutations only have the first one applied. The commented out lines should fix the issue, but when I uncomment them auto-evo breaks in a way I can't seem to track down. It also seems to break in the same way without the clear. I would be grateful if someone could look at it, as I'm stuck.

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -260,7 +255,7 @@ private static void PruneMutations(List<Tuple<MicrobeSpecies, double>> addResult
                 combinedScores += pastPressure.WeightedComparedScores(newScore, oldScore);
             }
 
-            if (combinedScores > 0)
+            if (combinedScores >= 0)
```
adQuid, line 285: what was the logic behind this change?

GameDungeon: That allows for changes that don't help but also don't hurt to not instantly get pruned. I'll probably revert it though, it was leftover from testing when I was trying to track down the main bug.

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -382,48 +335,37 @@ private void PredatorsOf(List<Species> result, Miche micheTree, Species species)
     }
 
     /// <summary>
-    ///   Returns a new list of all possible species that might emerge in response to the provided pressures,
-    ///   as well as a copy of the original species.
+    ///   Adds a new list of all possible species that might emerge in response to the provided pressures,
+    ///   as well as a copy of the original species to <see cref="mutationsToTry"\>.
```
hhyyrylainen: I think the check is failing here due to the period after the `<see/>` tag. I'll probably have time soon to fix the check, but for now you can probably just remove the ending period on this line and it should make the check happy.

src/auto-evo/steps/ModifyExistingSpecies.cs
hhyyrylainen: I'm not sure what more I can do except try to check if logically the clears and adds work in a sensible order considering where each list is used, but I think I can do that look through next week.

src/auto-evo/steps/ModifyExistingSpecies.cs
GameDungeon: I put a backslash there instead of a forward slash by mistake. That was causing the issue.

src/auto-evo/steps/ModifyExistingSpecies.cs
GameDungeon: From what I can tell, GetTopMutations was actually grabbing the worst mutations instead of the top ones (This should be the case in current master two. Maybe the old code was less susceptible to issues from it?), and the code change I made was allowing the bug to trigger.

Patryk26g:
In terms of performance it works wonderful, good work, but unfortunately there is still something wrong - the species take ages to evolve

Above 10 generations on this branch:
Screenshot_20250713_194607

3 generations on master
Screenshot_20250713_194844

GameDungeon:
All of the duplicate work did have the side effect of trying way more variations. I'll try cranking it up and see if it gives better results.

GameDungeon:
Ok, So I think I figured out the real issue. Before the mutations were applied in a random order, but now the are applied in a strict order based on the miche tree. This is an issue as beneficial mutations are now getting pruned if applied too early. I'm going to try and rework the mutation order to work better now.

With that change this now has a larger gameplay effect and will likely need some sort of play testing to make sure auto-evo isn't completely out of tune.

src/auto-evo/steps/ModifyExistingSpecies.cs
hhyyrylainen: Ah, so the syntax of the XML was incorrect... I guess I don't really need to change anything then (though in an optimal world a check would detect the incorrect syntax first)

GameDungeon:
This now disables the changes in Thim's most recent auto-evo PR. It is incompatible with the patch here. I would like to find a way to have them play nice, but I can't seem to think of one right now. I'm open to suggestions if anyone has any.

Other than that, this PR is complete and works in my testing. I'm opening it for review.

src/auto-evo/steps/ModifyExistingSpecies.cs
GameDungeon: I decided to keep this change as otherwise any mutations on nodes like the root node would be instantly pruned. It allows the miche tree to start with a random cast of species and specialize them as the go through the tree.

Patryk26g:
Now the auto-evo goes a little bit nuts and evolves species with nucleus after only 5 generations :sweat_smile: 
It definately is more "bold" now, but would require to be calmed down a little bit

Edit: I even got a species with nucleus after 3 generations
Screenshot_20250715_232438
Screenshot_20250715_232614

GameDungeon:
Just to have it on the record, that isn't a bug with this patch, and I'll fix it with my next balance patch.

adQuid:
it appears that new species aren't consistently being renamed: 
image

adQuid:
I pushed a change to rebalance the osmoregulation penalty. I don't understand why a square root function makes fewer species get pruned, but eukaryotes now appear at similar rates to before, with some variety of early species still present. It's still your PR, so feel free to mess with it however you need.

GameDungeon:
I'll try and test this tomorrow. Thanks for the commit.

hhyyrylainen:
I merged master here as I *think* this was messing up the CI running unit tests (or at least I hope that was the cause).

hhyyrylainen:
Was the renaming problem already fixed? I saw some commits after that but none directly referencing that.

src/auto-evo/mutations/CommonMutationFunctions.cs
```diff
@@ -444,4 +444,6 @@ private static Hex.HexSide[] SideTraversalOrder(Hex hex, Direction direction, Ra
 
         return TraversalOrder8;
     }
+
+    public record Mutant(MicrobeSpecies Species, double MP);
```
hhyyrylainen, line 441: To not wreck memory allocations, this has to be `public record struct Mutant(MicrobeSpecies Species, double MP);` to make this into a value type.

And also that means `Mutant?` should *not be used* anywhere as that causes boxing. Instead I think that Species here should be made nullable so that null species can stand-in for invalid Mutant instances.

GameDungeon: The only place that `Mutant?` is used is in the sorter, and I don't think it can be replaced with your solution in that context.

hhyyrylainen: I see, that's a bit unfortunate, but it seems it had the same problem with `Tuple`s already. So I guess for now could you add a TODO comment there about this architectural problem that would be nice to solve eventually?

hhyyrylainen:
Well this is critically not ready today, so this kind of can't make it into 0.8.3 anymore...

GameDungeon:
Got it. This one really needs a second balancing pass PR too, so unless both can make it neither should be merged.

Accidental-Explorer:
I am re-opening this and will be trying to balance it out with the current state of the auto-evo scoring system so that we can finally reap the benefits.

GameDungeon:
I've fixed all of the merge conflicts to the best of my ability. I haven't finished getting godot setup again, so I can't actually verify the code runs yet (sorry).

Accidental-Explorer:
So far this seems to be working mostly alright, but I've found one issue (that is also present in the current master version) that I think is blocking. I'll repeat here what I said on the discord:

It seems there is a bug that makes Auto-evo prediction not take into account any energy source that is not already being used by a species (including the player themselves). That is a per-patch condition. This is quite apparent in the auto-evo in-editor prediction if you turn off AI auto-evo. I don't know if that also affects real Auto-evo results (but it has been said that the calculations should be exactly the same), but either way it does not seem ideal.

The only reason it does not stand out so much in master is that there is some method that kind of force-evolves species to fill the available miches (more so for food sources than predation). But even with that, I have noticed this bug in the past, for example when moving to a new surface patch while having or placing a thylakoid before other photosynthesizers exist there. In that case, auto-evo prediction in the editor does not detect any energy gain.

This is not a bug caused by this PR. But this PR does not seem to show the "forced evolution" I mentioned before, and those are masking the bug in master.

So I do suspect that this issue is also affecting "real" auto-evo (not just player predictions), because without the forced evolutions, creation of new species that use the other energy sources is pretty much stalled for the first few generations in this PR.


Long story short, I can probably not properly balance PR unless that bug is fixed or the forced evolution can be included here.

GameDungeon:
This is minor, but `worldSettings.AutoEvoConfiguration.MutationsPerSpecies` is now effectively MutationsPerSpeciesPerLeafNode. Although this should be encouraging more exploration as each energy source is going to have a mutation passed to the mutations to try random pull?

Accidental-Explorer:
> This is minor, but `worldSettings.AutoEvoConfiguration.MutationsPerSpecies` is now effectively MutationsPerSpeciesPerLeafNode. Although this should be encouraging more exploration as each energy source is going to have a mutation passed to the mutations to try random pull?

Interesting! Then without changing that constant it would mean a lot more attempts than before. Might actually want to look into whether to lower that later on then?

So, from what I understand then, the problem I am seeing (lack of diversity early on) could have some causes I can imagine:
* All MP spent at the top of the tree almost every attempt, so nothing left to place f.e. rusticyanin.
* There's still something weird going with selecting which of the attempts actually succeeds.

Any other ideas?

GameDungeon:
Unless they are getting filtered out by GetTopMutations, a copy of the unedited species should be getting passed as a "mutation" to each miche, given it does not fail a pressure.

Accidental-Explorer:
> Unless they are getting filtered out by GetTopMutations, a copy of the unedited species should be getting passed as a "mutation" to each miche, given it does not fail a pressure.

The unedited ones? What about the edited ones? Because if one lonely species (single cytoplasm) tries to add a rusticyanin, it does not matter much that the *un*edited one gets passed to each miche. 

In that scenario, is the edited one guaranteed to get passed to the (unopposed) iron miche and then guaranteed to survive?

Or did I completely misunderstand you?

GameDungeon:
I'm saying the unedited one should still have full MP, so there should be enough left for an organelle. So when it gets to the iron miche there should always be something that can place one. Now if it fails a pressure then it won't get passed anymore, but it will get a chance to mutate on the miche it failed.

GameDungeon:
Is it possible a pressure near the top of the tree is taking too much attention away from the energy sources during the first couple generations?

Accidental-Explorer:
> I'm saying the unedited one should still have full MP, so there should be enough left for an organelle. So when it gets to the iron miche there should always be something that can place one.

Ah, I see! And is this the "iron branch" (which I guess still has several branches) or are we specifically talking about the leaf miche node?

> Is it possible a pressure near the top of the tree is taking too much attention away from the energy sources during the first couple generations?

It's possible, it does usually spawn new species at first, but those only seem to vary in behaviour mainly. On the other hand, that does not cost any MP. (I am also considering setting default behaviour to something closer to what auto-evo thinks is optimal right now)

But if that's the case, I'm not sure the algorithm is working as intended. An entire energy source without competition is such an enormous benefit that it should trump anything else, no? Even if it's inefficient, you'd expect them to split off from their parent species.

GameDungeon:
Can you disable all of the behavior mutations and test? I have a theory that species spend all of their MP on behavior as it provides an immediate benefit at the top of the tree. The second we have more than 50 candidate species the unedited species is probably pushed out by the get top mutations. The behavior mutations each double the number of candidate species early on as they provide an easy win. So unlike I thought before all of the species left with MP get pushed out before it gets to a miche where it would matter. You can see in game that the second all of the behaviors get reasonably optimized the number of other mutations explode.

Oh, also the number of species being tried is also very limited at the start as It's only the 3 per species based on the glucose leaf miche instead of the many more that get added when they have more miche leafs (up to a patch max of 20 of course).

Edit: Behavior does cost MP in auto-evo to stop the species from just editing it non-stop. Although you can try without it, I not sure it's necessary, although every species will probably end up with an optimal behavior basically immediately.

Accidental-Explorer:
> Can you disable all of the behavior mutations and test? I have a theory that species spend all of their MP on behavior as it provides an immediate benefit at the top of the tree.

I can probably get around to this again on Monday.

> Edit: Behavior does cost MP in auto-evo to stop the species from just editing it non-stop. Although you can try without it, I not sure it's necessary, although every species will probably end up with an optimal behavior basically immediately.

Don't we have a not-repeatable setting for this that we use with the other free mutations?

I don't think it's a problem if species get optimal behaviour all the time. The player can do the same for free also.

> Oh, also the number of species being tried is also very limited at the start as It's only the 3 per species based on the glucose leaf miche instead of the many more that get added when they have more miche leafs (up to a patch max of 20 of course).

The limited number of species makes sense. But they're not explicitly aiming for the other unoccupied miches, just the glucose one? Are they somehow just lucking into them?

GameDungeon:
> Don't we have a not-repeatable setting for this that we use with the other free mutations?

We do, and the behavior mutation is set to it. Removing the cost is probably fine. It was a little more reasonable in the old mutation system.

> The limited number of species makes sense. But they're not explicitly aiming for the other unoccupied miches, just the glucose one? Are they somehow just lucking into them?

The "final" species are only added to the "mutationstotry" pool by the leaf nodes, so if all of the species fail a check that "branch" of mutations never get added. So the number of species added scales with the number of leaf nodes. 

When a species enters a miche in the mutation it gets to mutate, and then any past or new mutations that fail the check are purged. So they do mutate in that direction, but only if they manage to pass every check required to get them to the leaf node.

hhyyrylainen:
Tolerances could also eat up the mutation candidates as I believe that is also near the top of the tree and can avoid very large negative effects from tolerance debuffs... and those do cost MP so they can deplete the available pool.

Accidental-Explorer:
> Edit: Behavior does cost MP in auto-evo to stop the species from just editing it non-stop. Although you can try without it, I not sure it's necessary, although every species will probably end up with an optimal behavior basically immediately.

Are you sure about this? I don't see any change to the MP in ChangeBehaviorScore.cs. It might been changed last year? Perhaps even by myself and I forgot.

> Can you disable all of the behavior mutations and test? I have a theory that species spend all of their MP on behavior as it provides an immediate benefit at the top of the tree.

Tried this, and it indeed did not change anything. (other than behaviour obviously staring the same.)

Also tried removing MP cost from membrane rigidity change (which is immediately set to minimum by auto-evo), but that also didn't change things.

> Tolerances could also eat up the mutation candidates as I believe that is also near the top of the tree and can avoid very large negative effects from tolerance debuffs... and those do cost MP so they can deplete the available pool.

Good thing to try, but no effect either. (should be noted that this lack of diversity is a first-few-generations thing and so tolerance mutation probably does not matter much within the confines of the thermal vents.)



Seems to me that evolution mainly gets going once life has spread out over more patches? Almost every significant change seems to happen by splitting species between patches, rather than editing existing species or splitting within the same patch.

hhyyrylainen:
> Seems to me that evolution mainly gets going once life has spread out over more patches? Almost every significant change seems to happen by splitting species between patches

That's quite curious.

That's actually a feature I made for auto-evo so that it fills up the world much faster. It's somewhat arbitrary that we almost force species to appear to fill an empty patch if there's life in any of the adjacent patches. So only that part working seems to be a bit strange, almost like the core of auto-evo is just broken and the forcing aspect only works by skipping the normal mutation stuff (if I remember right).

Accidental-Explorer:
> That's actually a feature I made for auto-evo so that it fills up the world much faster. It's somewhat arbitrary that we almost force species to appear to fill an empty patch if there's life in any of the adjacent patches. So only that part working seems to be a bit strange, almost like the core of auto-evo is just broken and the forcing aspect only works by skipping the normal mutation stuff (if I remember right).

If I remember previous discussions correctly, the "forcing part" was filling miches inside already populated patches? it triggers the "new species emerged to fill "niche" message? That's what causes all food sources to be immediately filled in the master version, but that is not happening in this PR. 

So the options I know of that are left are:
* Actual mutation of existing species.
* A species that exists across multiple patches splitting in half, with some getting the new mutation. (shows as "emerged due to differing selection pressures")

hhyyrylainen:
Maybe it was actually removed because none of these auto-evo config properties seem to be even used anymore:

```
    [JsonProperty]
    public float BiodiversityAttemptFillChance { get; private set; }

    [JsonProperty]
    public bool UseBiodiversityForceSplit { get; private set; }

    [JsonProperty]
    public float BiodiversityFromNeighbourPatchChance { get; private set; }

    [JsonProperty]
    public bool BiodiversityNearbyPatchIsFreePopulation { get; private set; }

    [JsonProperty]
    public int NewBiodiversityIncreasingSpeciesPopulation { get; private set; }
```

Might be good to delete all the auto-evo options that no longer do anything.

Accidental-Explorer:
The change I just pushed is one specific number tweak (discovered through _a _lot_ of trial and error) that at least makes iron oxidising cells show up in the second generation. (the reduction I made is actually larger than necessary)

I still don't quite like how "fragile" the system that selects mutants seems to be, if a factor like this is enough to break it. (logically, it really should not have such an outsized effect).

Accidental-Explorer:
> Maybe it was actually removed because none of these auto-evo config properties seem to be even used anymore:
> 
> ```
>     [JsonProperty]
>     public float BiodiversityAttemptFillChance { get; private set; }
> 
>     [JsonProperty]
>     public bool UseBiodiversityForceSplit { get; private set; }
> 
>     [JsonProperty]
>     public float BiodiversityFromNeighbourPatchChance { get; private set; }
> 
>     [JsonProperty]
>     public bool BiodiversityNearbyPatchIsFreePopulation { get; private set; }
> 
>     [JsonProperty]
>     public int NewBiodiversityIncreasingSpeciesPopulation { get; private set; }
> ```
> 
> Might be good to delete all the auto-evo options that no longer do anything.

So, I take it we want to do this in this PR as well? @GameDungeon, could you confirm that these are things we also would not want to use in the future?

I think it would be nice if we can push this through in the coming week, so that our players can have some better auto-evo performance in the next update (since we've had an increase in complaints about that).

Currently, it's still a little bit slow to get going in the first few generations, but otherwise seems to be working well for me. Of course, I could be missing things, so it would be good to get more eyes on.

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -169,13 +164,21 @@ public bool RunStep(RunResults results)
...
 
-                while (mutationsToTry.Count > TotalMutationsToTry)
+                // Rename and colour selected mutations
+                foreach (var mutation in mutationsToTry.Take(TotalMutationsToTry))
                 {
-                    mutationsToTry.RemoveAt(mutationsToTry.Count - 1);
-                }
+                    MutationLogicFunctions.NameNewMicrobeSpecies(mutation.MutatedSpecies, mutation.ParentSpecies);
+
+                    var oldColour = mutation.MutatedSpecies.Colour;
```
Patryk26g: We should use MutationLogicFunctions.ColourNewMicrobeSpecies as that handles branching off from the player's species... and anyway it's already done in:

if (newPopulation > Constants.AUTO_EVO_MINIMUM_VIABLE_POPULATION)
{
// Only apply a new name and colour to results that are actually kept
MutationLogicFunctions.NameNewMicrobeSpecies(mutation.MutatedSpecies, mutation.ParentSpecies);
MutationLogicFunctions.ColourNewMicrobeSpecies(random, mutation.MutatedSpecies,
mutation.ParentSpecies);

so what's the point of doing it here?

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -289,55 +295,15 @@ private static void GetTopMutations(List<Tuple<MicrobeSpecies, double>> result,
...
-        }
-
-        // This section of the code tries to mutate species into unfilled miches
-        // Not exactly realistic, but more diversity is more fun for the player
-        // TODO: Make this an auto-evo option
-        foreach (var emptyLeafNode in emptyLeafNodes)
-        {
-            currentTraversal.Clear();
-            emptyLeafNode.BackTraversal(currentTraversal);
+        double totalMP = 100 * worldSettings.AIMutationMultiplier;
```
Patryk26g: Would be nice to put it in a constant

GameDungeon: This is technically definitionally correct as MP is supposed to be a percentage, although that still might be more clear?

GameDungeon:
> So, I take it we want to do this in this PR as well? @GameDungeon, could you confirm that these are things we also wold not want to use in the future?

That's probably fine. I think it would be best to remove them now and add back anything as needed. They currently don't serve any purpose beyond confusing people.

src/auto-evo/steps/ModifyExistingSpecies.cs
GameDungeon: I think this is the remnants of a merge conflict for context. It does need to be fixed.

src/auto-evo/steps/ModifyExistingSpecies.cs
hhyyrylainen: Yeah, we have a constant after all for the default MP called `Constants.BASE_MUTATION_POINTS`

src/auto-evo/steps/ModifyExistingSpecies.cs
Accidental-Explorer: I removed this whole section and it indeed does not seem to break anything.

src/auto-evo/steps/ModifyExistingSpecies.cs
Accidental-Explorer: I replaced the static number with the relevant constant.

Accidental-Explorer:
> Might be good to delete all the auto-evo options that no longer do anything.

So what exactly are we talking about here, in practical terms? Deleting those config properties? Finding methods that are now unused?

hhyyrylainen:
Just removing the properties.

What do you mean by "finding methods that are now unused"? AFAIK there should not be (major) methods in auto-evo that are unused.

Accidental-Explorer:
> Just removing the properties.

Alright, this is done now.


> What do you mean by "finding methods that are now unused"? AFAIK there should not be (major) methods in auto-evo that are unused.

Nothing in particular, just was not sure if I was missing something.

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -169,16 +163,6 @@ public bool RunStep(RunResults results)
...
-                while (mutationsToTry.Count > TotalMutationsToTry)
-                {
-                    mutationsToTry.RemoveAt(mutationsToTry.Count - 1);
-                }
-
-                foreach (var mutation in mutationsToTry)
-                {
-                    mutation.MutatedSpecies.OnEdited();
-                }
-
```
Accidental-Explorer, line 196: I am questioning why this section was removed. I can imagine different ways of handling mutations could mean you don't want need or want to set this limit on it.

But isn't removing OnEdited a loss in the accuracy of evaluating the mutants?

hhyyrylainen: > But isn't removing OnEdited a loss in the accuracy of evaluating the mutants?

Maybe it was moved somewhere else? Because it *has* to be called on a modified species at some point (before the final result is taken into use). Otherwise there's a franken species with stats that don't match its cell layout.

GameDungeon: That is almost certainly a merge artifact from how old this PR is now. That really should be there.

Accidental-Explorer: Alright, I added this back in.

GameDungeon:
Apologies, I had meant that commit there to be a suggestion, not a commit, but there it is anyway. You can revert it if you wish. That's a suggestion that I think will reduce the memory usage by a good chunk, but I currently can't test it.

Accidental-Explorer:
> Apologies, I had meant that commit there to be a suggestion, not a commit, but there it is anyway. You can revert it if you wish. That's a suggestion that I think will reduce the memory usage by a good chunk, but I currently can't test it.

Thanks for the suggestion! Looks like it failed some checks, but I will take a look at it tomorrow.

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -253,16 +254,24 @@ private static void PruneMutations(List<Mutant> addResultsTo, MicrobeSpecies bas
...
                     combinedScores = -1;
                     break;
                 }
 
+                // Never prune if the mutation succeeds a new pressure check
+                // Because the score cannot be compared with a parent that does not fill the same niche
+                if (newScore > 0 && oldScore <= 0)
+                {
+                    noPruning = true;
+                }
```
GameDungeon: You could set combined Scores to 1 and break here like above does. It would save having to add a "noPruning" var.

Accidental-Explorer: Oh right, I think this was before I realised the exact score does not matter, because it does not go outside.

Patryk26g:
Besides increased speed, the auto-evo results are much much better now! I even noticed species that had ATP inhibitor toxins, were bigger than me or even had neclues before me (in the 9th generation). Makes it so much more fun to play now, good job

flake.nix
hhyyrylainen, line 1: I assume these are @GameDungeon's changes to this file.

Accidental-Explorer: Correct, and I honestly have no idea what that is for.

hhyyrylainen: It's for building Thrive on NixOS.

src/auto-evo/mutations/CommonMutationFunctions.cs
hhyyrylainen: I'll add the TODO comment.

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -277,22 +272,32 @@ private static void PruneMutations(List<Tuple<MicrobeSpecies, double>> addResult
...
 
-    private static void GetTopMutations(List<Tuple<MicrobeSpecies, double>> result,
-        List<Tuple<MicrobeSpecies, double>> mutated, int amount, MutationSorter sorter)
+    private static void GetTopMutations(List<Mutant> result,
+        List<Mutant> mutated, int amount, MutationSorter sorter)
     {
         result.Clear();
 
-        mutated.Sort(sorter);
+        mutated.Sort((a, b) => sorter.Compare(b, a));
```
hhyyrylainen: Doesn't this just reverse the order? Wouldn't it make more sense to update the custom sorter to be opposite?

hhyyrylainen: Did this.

hhyyrylainen:
With the patch out and no immediate bugs reported, I'm going to merge this now.

PR #6214, opened 2025-07-13
Make the ripple subdivision count static

dligr:

I noticed some lag in the game due to ripple effect subdivision change calls - according to the profiler, they were using about 20% of the game's total CPU time. To address this, this PR makes the subdivision count static. The 80x80 count is arbitrary and can be made higher or lower if necessary.

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Membrane.tscn
+++ b/src/microbe_stage/Membrane.tscn
@@ -67,8 +67,8 @@ shader_parameter/MembraneRadius = 5.0
 [sub_resource type="PlaneMesh" id="PlaneMesh_ripple"]
 resource_local_to_scene = true
 size = Vector2(18, 18)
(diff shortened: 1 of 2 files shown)


hhyyrylainen:
I think it is a good idea to do this if the LOD system meant to save performance caused a lot of CPU usage. If there is a need to allow changing this in the future, I think a graphics option would make sense to add, but maybe we can get away with just a fixed value for now?

dligr:
Regarding the option, maybe it would make sense to add a sort of "Shader quality" setting to avoid adding too many options? But a static number would do for now, though maybe we need to experiment with it a bit to find the optimal value (40x40 is the lowest I found that still looks decent)

PR #6238, opened 2025-07-20
Increase max zoom out for large cell colonies

### LINKED ISSUE #2312: Increase max zoom out when controlling a large cell colony
as a big colony can take up a pretty large amount of the screen.

dligr:

Max zoom out now scales with colony size. There's a hard limit to it though, to prevent the camera from going too far.

Also the max zoom in is now more distant to make cells appear smaller. This is a game feel enchancement, so it's a bit debatable.

**Related Issues**

Closes #2312

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -779,6 +779,17 @@ public static class Constants
 
     public const float SCREEN_DAMAGE_FLASH_DECAY_SPEED = 1.0f;
 
+    public const float MICROBE_CAMERA_MIN_HEGIHT = 3.0f;
+
+    public const float MICROBE_CAMERA_MAX_HEGIHT = 80.0f;
+
+    public const float MULTICELLULAR_CAMERA_MIN_HEGIHT = 8.0f;
+
+    /// <summary>
+    ///   The highest that the dynamic multicellular camera max height can get.
+    /// </summary>
+    public const float MULTICELLULAR_CAMERA_MAX_HEGIHT = 180.0f;
+
     /// <summary>
     ///   Cells need at least this much ATP to regenerate health passively. This is now less than one to allow cells
     ///   with 1 storage to regenerate health. As reaching exactly full storage of ATP is not really possible due to
--- a/src/microbe_stage/MicrobeStage.cs
+++ b/src/microbe_stage/MicrobeStage.cs
@@ -845,6 +845,8 @@ public partial class MicrobeStage : CreatureStageBase<Entity, MicrobeWorldSimula
             }
         }
 
+        UpdateZoomLevels(playerIsMulticellular);
+
         Player.Set(environmentalEffects);
 
         var playerCompounds = Player.Get<CompoundStorage>().Compounds;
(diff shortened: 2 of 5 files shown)


hhyyrylainen:
> Also the max zoom in is now more distant to make cells appear smaller. This is a game feel enchancement, so it's a bit debatable.

I think the max zoom out as a single cell is pretty optimal as even the smallest cells stay visible.

Maybe the default zoom could increase only after becoming eukaryotic? I think that would make sense to give one more bonus from getting the nucleus.

simulation_parameters/Constants.cs
```diff
@@ -779,6 +779,17 @@ public static class Constants
 
     public const float SCREEN_DAMAGE_FLASH_DECAY_SPEED = 1.0f;
 
+    public const float MICROBE_CAMERA_MIN_HEGIHT = 3.0f;
```
hhyyrylainen: Typo in `height`

hhyyrylainen:
On reading the code, I did not find where the default max zoom level was changed? So after all that still seems to be the same as before?

src/microbe_stage/MicrobeStage.cs
```diff
@@ -1342,6 +1344,41 @@ protected override void OnLightLevelUpdate()
...
+                    maxDistance = distance;
+                }
+            }
+
+            maxDistance = MathF.Sqrt(maxDistance);
+
+            // Extra padding, just in case
+            maxDistance += 20.0f;
+
+            Camera.MinCameraHeight = Constants.MULTICELLULAR_CAMERA_MIN_HEGIHT;
```
dligr: This is the line that changes the height limit

hhyyrylainen: Ah, I thought I read "increased the max zoom out distance for early microbe stage as well" but I read wrong.


I suppose that's an alright change, but why do we *need* to remove the ability to inspect the lead cell of a colony as closely as before?

dligr: This is mostly a game feel enchancement, but my reasoning is basically that some mechanics and visuals need to be readapted for the multicellular stage, as they were previously tailored for the microbe stage. Those changes have to be implemented conditionally, because the codebase is shared between the two stages.

Camera zooming is one such mechanic: at minimum zoom, the smallest possible player cells in the both stages now occupy most of the screen.

We might want to adapt other things in future, e.g. the currents display, movement mechanics, etc.

hhyyrylainen: My counter point is that if a player unlocks a new organelle, why shouldn't the player be allowed to admire its 3D model in the stage by zooming in far enough?

dligr: Well, organelles are still very visible at max zoom in
image

hhyyrylainen: This is a bit of a pointless debate, but I'll note that while you can see the organelles, you can get true closeup view of them in the microbe stage if the player prefers enjoying that kind of thing.

PR #6240, opened 2025-07-25
Fix the patch move button not updating correctly

dligr:

After #6239, the patch move button was partially broken, because its `IsPatchMoveValid` property was updated after the button status update. This PR fixes that by reordering the patch property change.

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/PatchMapEditorComponent.cs
+++ b/src/microbe_stage/editor/PatchMapEditorComponent.cs
@@ -250,8 +250,8 @@ public partial class PatchMapEditorComponent<TEditor> : EditorComponentBase<TEdi
 
     protected virtual void UpdateShownPatchDetails()
     {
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
Good that this problem was found. But I kind of expected issues like that with removing that kind of seeming double data update, which is why I usually don't remove them despite there being many such cases in the game as they are nightmares to fully test that it doesn't break some corner case when the double update is removed.

dligr:
It's a good point that changing code like this can result in some unforseen consquences (though I don't see anything else being broken in this case). Should I revert this function to what it was before #6239 instead?

hhyyrylainen:
I think it is better to just do this fix, because if you found this but nothing else wrong, then this might be the only easy to find problem with the change. So basically I don't see the point in reverting when the change is now hopefully fixed up.

PR #6244, opened 2025-07-29
Implement cell type tooltips

### LINKED ISSUE #3212: Add tooltips for the cell type selections similarly to how organelles have them
to give the player some idea what kind of cell types they've created and are thinking of placing.

dligr:

Implements cell type tooltips:
image

ATP balance bars have also been moved to tooltips (to clear some space in the selection menu). Their values are now displayed relative to each other - i.e. cell type ATP production is now shown relative to its ATP consumption and vice versa.

**Related Issues**

Closes #3212

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/multicellular_stage/editor/CellTypeStatLabel.cs
hhyyrylainen, line 1: Is there a reason why an existing stat label control is not used instead of making a new one?

src/multicellular_stage/editor/CellTypeTooltip.cs
```diff
@@ -0,0 +1,161 @@
...
+        get => displayName ?? "CellTypeTooltip_unset";
+        set
+        {
+            displayName = value;
+            UpdateName();
+        }
+    }
+
+    /// <summary>
+    ///   The main message the tooltip contains.
```
hhyyrylainen: Are these docs copied from the interface? If so these should not be just copy pasted here, it is better to just leave them out and I think IDEs automatically grab the interface documentation. That way the two places will not get out of sync after the copy-paste.

src/multicellular_stage/editor/CellBodyPlanEditorComponent.cs
```diff
@@ -918,57 +935,92 @@ private void UpdateCellTypeSelections()
...
+
+        foreach (var button in cellTypeSelectionButtons.Values)
+        {
+            var cellType = button.CellType;
+
+            var tooltip = ToolTipManager.Instance.GetToolTip<CellTypeTooltip>(cellType.TypeName, "cellTypes");
+
+            if (tooltip == null)
+            {
+                GD.PrintErr($"Tooltip not found for species type: {cellType.TypeName}");
```
hhyyrylainen:
```suggestion
                GD.PrintErr($"Tooltip not found for species' cell type: {cellType.TypeName}");
```

src/multicellular_stage/editor/CellTypeStatLabel.cs
dligr: The problem with using `CellStatsIndicator` is that it includes a change indicator, which isn't something that a cell type tooltip needs

hhyyrylainen: I think it is already used in many contexts where the change indicator is unused. There should even be documentation in that class on how to use it without the change indicator triggering. I think that is a cleaner design than having a separate stats indicator what just is missing the change component (though I didn't originally make that design myself, I don't think).

hhyyrylainen (approved):
I think the code change is good now. I quickly tested in-game and everything also seems to function right with this.

Issue #2766, opened 2021-11-17
Implement Steam achievements unlocking


hhyyrylainen:
Requires: https://github.com/Revolutionary-Games/Thrive/issues/2665
After that should be pretty easy to hook into the Steam APIs to load and save current achievement progress.

hhyyrylainen:
This was done already: https://github.com/Revolutionary-Games/Thrive/pull/6258

PR #6246, opened 2025-07-31
Achievements backend implementation

### LINKED ISSUE #2665: Implement achievements system
we need a basic framework for achievements:
- progress tracking and saving-
- popups for progress and unlocking
- some menu to view achievements
- needs to be extensible to easily configure in new achievements
- needs a way to suppress the popups for use with steam

hhyyrylainen:

Implements an achievements system with one achievement put in. Doesn't yet have Steam integration or icons or sounds for the achievement (and the full set of microbe achievements is not done yet)

**Related Issues**

closes #2665

related but not done yet: #2766

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/achievements/AchievementsManager.cs
+++ b/src/general/achievements/AchievementsManager.cs
@@ -571,6 +571,10 @@ public partial class AchievementsManager : Node
         if (string.IsNullOrEmpty(result))
             throw new IOException($"Failed to read achievements file: {path}");
 
+        // Harmonize to unix line endings
+        if (result.Contains("\r\n"))
+            result = result.Replace("\r\n", "\n");
+
         var hash = Convert.ToBase64String(SHA256.HashData(Encoding.UTF8.GetBytes(result)));
 
         if (hash != ACHIEVEMENTS_INTEGRITY)


src/general/achievements/AchievementStatStore.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// </summary>
+    public void Reset()
+    {
+        GD.Print("Resetting tracked stats");
+        statMicrobeKills = 0;
+    }
+
+    public void Save(Dictionary<int, int> intValues)
+    {
+        intValues[STAT_MICROBE_KILLS] = statMicrobeKills;
```
CI09, line 60: What if stat wouldn't be a number but a boolean for an instance?

src/general/achievements/AchievementsManager.cs
```diff
@@ -0,0 +1,698 @@
...
+            if (shownPopupTime > ACHIEVEMENT_DISPLAY_TIME)
+            {
+                // Show the next achievement popup
+                shownPopupTime = 0;
+                var popup = achievementsToPopupQueue.Dequeue();
+
+                if (createdAchievementPopup == null)
+                {
+                    createdAchievementPopup = achievementPopupScene.Instantiate<AchievementPopup>();
+                    achievementsGUIContainer.AddChild(createdAchievementPopup);
```
CI09, line 231: Does the popup ever get removed?

src/general/achievements/AchievementStatStore.cs
hhyyrylainen: Steam API only supports: int, float, and averages. So we have to use just what is compatible with Steam: https://partner.steamgames.com/doc/features/achievements (as I want our achievements system to be also hooked up into Steam).

I'll add the float variant for data once we need it.

src/general/achievements/AchievementsManager.cs
hhyyrylainen: It doesn't before the entire scene tree is destructed. I could have made an achievements manager scene file, but I kind of didn't and then ended up with this architecture. This has the slight added benefit that in Steam mode (once I add that) the graphics never get instantiated as the Steam client library will handle drawing the achievement popups.

hhyyrylainen:
The file should be in the repo. But maybe it is actually caused by Windows line endings? As it is a text file?

I'll make a quick commit to harmonize the line endings upon loading the achievements data. Could you re-test with that?

CI09:
I can run it now, but I tried changing achievements to not have to actually kill 50 microbes to just test but hash check stopped me; I think it would be a worthy addition to put somewhere in documentary on how to do achievements and generate hash. Can be seperate branch though if you don't feel like doing it now

CI09:
Also I quickly noticed that options menu is broken (main menu node stopped referencing to it probably, quick fix)

hhyyrylainen:
> I can run it now, but I tried changing achievements to not have to actually kill 50 microbes to just test but hash check stopped me; I think it would be a worthy addition to put somewhere in documentary on how to do achievements and generate hash. Can be seperate branch though if you don't feel like doing it now

It's very intentional that the achievements data file cannot be changed easily. That's because the config must match what is configured on Steam side, so only very intentional changes should be made. For that reason I don't plan on writing more docs other than the existing comments on how to change that.

> Also I quickly noticed that options menu is broken (main menu node stopped referencing to it probably, quick fix)

It works fine for me so could you confirm if that's just a local problem for you with Godot automatically breaking it?

CI09:
Yeah it works now that I opened Godot editor. I didn't dive too deep but stuff generally works and besides additional cosmetic stuff like greying out ununlocked achievements etc.  I can do myself or someone else can do

hhyyrylainen:
> greying out ununlocked achievements

That should already be in, though I didn't spend that much time thinking about the colours.

Issue #6234, opened 2025-07-17
Compound numbers blend above their bars in grid view


Teashrock:
### Describe the bug

When I enable the grid view, I see some numbers end up over their respective bars. This creates a minor visual inconvenience, as the numbers are hardly visible.

### To Reproduce

1. Open the game.
2. Enter the gameplay routine.
3. Enable grid view of compounds and patch details.
4. See the bug.

### Expected behaviour

Both nummbers and bars being clearly visible in grid view.

### Screenshots

<img width="390" height="599" alt="Image" src="https://github.com/user-attachments/assets/e87e513a-21d0-46a4-87ac-760bb8f4322a" />

### System information

- OS: Gentoo Linux x86_64
- Graphics card: Radeon Vega 8
- Thrive: 0.8.2.1

### Game Logs

I'll add if they're needed here.

### Additional Context

For some reason, "Agents" grid view seems to work well. It can be seen on the screenshot.

hhyyrylainen:
This problem is really only visible once the storage amounts are quite high, as is in your example screenshot. Doing any graphics work on the compact view has been pretty low priority, so I think a simple fix is needed. I think either the icons need to be hidden in the small view once the storage amounts are more than single digits. Or maybe some kind of font shadow could be used? Though as the actual numbers are the same as in normal view, that would also impact the normal view or be quite difficult to program.

> For some reason, "Agents" grid view seems to work well. It can be seen on the screenshot.

The reason is that there's just 2 agents so no rows exist with 3 columns which is what causes the issue when the bars are squeezed into so tight space.

Teashrock:
@hhyyrylainen 
If you look at the screenshot closely, you'll also see the numbers are barely out of the left border. Maybe it would be better to decrease their size a bit, then add a shadow?

hhyyrylainen:
Storage numbers can be up to 3 digits long so that wouldn't work. Also the font size is already 14 which is just a bit larger than the absolute smallest size allowed, which is 12 px.

Teashrock:
So that leaves only putting a shadow and hiding the icons. I think hiding the icons would be inconvenient, as it would mean hiding the only thing pointing at what compound a number is meant for. So if I were to choose, it would be putting a shadow.

PR #6256, opened 2025-08-05
Compound number visibility fix

OmerCD:

This PR adds a new theme to compound text when it is in compact view.
image

 ~~I moved the Compound related files under gui directory to its own folder to make it look more organized.~~

~~I also added a simple functionality to clear selection in cell editor with right click which is really unrelated but annoyed me.~~

**Related Issues**
Closes https://github.com/Revolutionary-Games/Thrive/issues/6234
**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
> I also added a simple functionality to clear selection in cell editor with right click which is really unrelated but annoyed me.

Could you please make that as a separate PR because that is too unrelated to the other changes?

hhyyrylainen:
I see two architecture things I'd like to see done differently:
- I'd prefer a solution where individual theme elements would be changed *or* the label font settings would be changed directly (so instead of swapping a theme, as that is a thing that is not used anywhere in the Thrive GUI currently)
- And I don't really see the point of putting *2* scene files in their own subfolder. With that logic we would end up with half the number of current scene files as separate folders, and I think that is going to be a bigger mess.

So could you please rework those? Or let me know what I missed in my analysis and why the current approach is better.

OmerCD:
Thanks for the quick feedback. I suspected things were out of place but with your feedback I hope it will be proper this time. I still find setting all options at once approach cleaner and revised my PR according to it.

src/microbe_stage/gui/CompoundProgressBar.cs
```diff
@@ -50,6 +50,18 @@ public partial class CompoundProgressBar : Control
 
     private Color fillColour = new(0.6f, 0.6f, 0.6f, 1);
 
+    private StyleBoxFlat amountCompactTheme = new()
```
hhyyrylainen: I think it would be better if this was a `[Export]` variable that is set through the Godot Editor. That way the visuals could be easily tweaked without needing to modify the source code at all.

src/microbe_stage/gui/CompoundProgressBar.cs
```diff
@@ -469,6 +481,7 @@ protected override void Dispose(bool disposing)
             minSizeXReference.Dispose();
         }
 
+        amountCompactTheme.Dispose();
```
hhyyrylainen: Once the theme is set through the Godot editor, this Dispose call should be unnecessary.

src/microbe_stage/gui/CompoundProgressBar.cs
```diff
@@ -617,13 +630,15 @@ private void ApplyCompactMode(bool playAnimation)
                 tween.TweenProperty(this, minSizeXReference,
                     Narrow ? Constants.COMPOUND_BAR_NARROW_COMPACT_WIDTH : Constants.COMPOUND_BAR_COMPACT_WIDTH, 0.3);
 
+                amountLabel.AddThemeStyleboxOverride("normal", amountCompactTheme);
```
hhyyrylainen: I think this allocates a StringName instance each time. Instead of doing that, you should specify a `private readonly StringName normalStyleboxName = new("normal");` in this class and add `normalStyleboxName.Dispose()` into the Dispose method.

OmerCD:
Thanks again for the insightful suggestions. I don't know very much about Godot and appreciate it very much. I am pushing the new changes and hopefully should be good enough.

src/microbe_stage/gui/CompoundProgressBar.cs
```diff
@@ -229,6 +231,19 @@ public bool Compact
...
+    {
+        BgColor = new Color(0.2f, 0.2f, 0.2f, 0.45f),
+        BorderColor = Colors.Transparent,
+        CornerRadiusBottomLeft = 6,
+        CornerRadiusBottomRight = 6,
+        CornerRadiusTopLeft = 6,
+        CornerRadiusTopRight = 6,
+        ContentMarginLeft = 6,
+        ContentMarginRight = 3,
+    };
```
hhyyrylainen:
```suggestion
    private StyleBoxFlat AmountCompactTheme = null!;
```

This data should be set in the .tscn file and be just specified as null here. As that is the way we specify stuff that is initialized by the time `_Ready` is done.

And this should be a private field as it is meant to be set through Godot editor and not through other code.

Other than this one small thing, everything else looks fine now.

OmerCD:
Sorry to bother you with these little code pieces. I hope I will understand the codebase better for future PRs.

hhyyrylainen:
The code change looks perfect now, but I'll test gameplay functionality a bit later before merging.

hhyyrylainen:
If you don't mind I could try finishing this with the font shadow approach today so that this can be included as a fix in 0.8.3 release.

hhyyrylainen:
This is how my attempt looks:

kuva

I think it is good enough to ensure the visibility without causing much noticeable visual change. Most players probably won't notice the text gaining / losing the one pixel shadow.

hhyyrylainen (approved):
I adjusted the visuals and the approach a bit, and I think this is now good. I'll merge this now for this to make it into the next release.

PR #6257, opened 2025-08-05
Add right click to deselect chosen part in cell editor.

OmerCD:

In cell editor, when you select an organelle there is no easy way to deselect it. With this, right clicking on an empty are should clear your selection now.

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -853,7 +853,14 @@ public partial class CellEditorComponent :
             return true;
         }
 
-        return base.CancelCurrentAction();
+        if (base.CancelCurrentAction())
+            return true;
+
+        if (string.IsNullOrWhiteSpace(activeActionName))
+            return false;
+
+        DeselectOrganelleToPlace();
+        return true;
     }
 
     public override void OnEditorSpeciesSetup(Species species)


hhyyrylainen:
Hmm, there's really not ever a situation where you would need to deselect a part in Thrive...
I think I would personally be annoyed by this if I missclick and accidentally deselect the part rather than managing to open the context menu on something.

OmerCD:
Even if it is not right click, I always found not being able to empty your selection very annoying. If I were to left click on something while dragging something unrelated it makes me nervous that it is going to have unpredictable results. And I think this might open up new control options in cell editor (like drag-drop with left click to move organelles around).

hhyyrylainen:
> while dragging

What are you dragging? There's no support for dragging any of the cell parts or anything like that. If you are trying to drag things then I guess that indicates a failure of the GUI to communicate the interaction possibilities clearly.

The only place I can recall dragging working is reordering the organelle growth order.

dylanlee082:
I would be in favor of this change as well. When I did my first play through it was annoying not being able to de-select a organelle after I had clicked on one, especially when I then wanted to go right-click and modify on another organelle. This happened in particular when I was trying to place the nucleus because it was so much bigger than the other things I had added and I couldn't see the other options below very easier.

hhyyrylainen:
If this were to be accepted I think there should be a separate action in the input options JSON config and such so that the action to "clear the current editor action" can be rebound as the player likes.

PR #6259, opened 2025-08-06
Intercellular matrix lines

### LINKED ISSUE #4554: Add intercellular matrix graphics to fill in gaps on cell colonies
Opening an issue as the PR that tried to add this is now closed: https://github.com/Revolutionary-Games/Thrive/pull/4123
and there's a suggestion: https://suggestions.revolutionarygamesstudio.com/posts/33/intercellular-matrix

dligr:

Implements intercellular matrix lines
image

TODO:
- [x] ~Fix scaling when cells are too far apart~ (Probably no way to make it look good when the cells are **really** far apart)
- [x] Don't add connections/links when the two cells are close enough to look connected

**Related Issues**

Closes #4554

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/multicellular_stage/systems/IntercellularMatrixSystem.cs
```diff
@@ -0,0 +1,150 @@
...
+[With(typeof(SpatialInstance))]
+[Without(typeof(MicrobeColony))]
+[ReadsComponent(typeof(MicrobeColonyMember))]
+[ReadsComponent(typeof(SpatialInstance))]
+[ReadsComponent(typeof(CellProperties))]
+[RuntimeCost(1.0f)]
+[RunsOnMainThread]
+public sealed class IntercellularMatrixSystem : AEntitySetSystem<float>
+{
+    private static readonly Lazy<PackedScene> ConnectionScene =
```
hhyyrylainen, line 19: Is this a common pattern in the systems (I may have forgotten), or why was this approach used?

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
```diff
@@ -0,0 +1,150 @@
...
+        var ourMembrane = entity.Get<CellProperties>().CreatedMembrane;
+        var targetMembrane = parentEntity.Get<CellProperties>().CreatedMembrane;
+        if (ourMembrane == null || targetMembrane == null)
+            return;
+
+        var connection = ConnectionScene.Value.Instantiate<Node3D>();
+        instance.AddChild(connection);
+
+        // Get target's relative position (taking rotation into account) using inverse transform multiplication
+        var inverseColonyTransform = entity.Get<WorldPosition>().ToTransform().Inverse();
```
hhyyrylainen: Is this info not available in the colony or helper code? I kind of would like this to leverage existing math so that we don't end up in a situation where some part cannot be then updated correctly when the other place is.

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        var relativePosition = pointB - pointA;
+        float relativePosLength = relativePosition.Length();
+
+        var angle = relativePosition.AngleTo(Vector3.Forward);
+        if (relativePosition.X > 0.0f)
+            angle *= -1.0f;
+
+        connection.Scale = new Vector3(5.0f, 1.0f, relativePosLength + 3.0f);
+        connection.RotateY(angle);
```
hhyyrylainen: I suppose this is safe, but in general I prefer methods that use `Quat`s for rotation as this isn't always a safe operation when using individual axes like this.

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
```diff
@@ -0,0 +1,150 @@
...
+            GD.PrintErr("Intercellular connection is null, can't apply material parameters");
+            return;
+        }
+
+        TempMaterialList.Clear();
+        intercellularMatrix.GeneratedConnection.GetMaterial(TempMaterialList);
+
+        foreach (var material in TempMaterialList)
+        {
+            material.SetShaderParameter("tint", entity.Get<CellProperties>().Colour);
```
hhyyrylainen: "tint" here probably should be saved in a `private readonly StringName` to not need to allocate string names constantly.

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
```diff
@@ -0,0 +1,150 @@
...
+        ref IntercellularMatrix intercellularMatrix)
+    {
+        if (intercellularMatrix.GeneratedConnection == null)
+        {
+            GD.PrintErr("Intercellular connection is null, can't apply material parameters");
+            return;
+        }
+
+        TempMaterialList.Clear();
+        intercellularMatrix.GeneratedConnection.GetMaterial(TempMaterialList);
```
hhyyrylainen: GetMaterial is our helper method, but in this situation where the exact visual instance that is created, I think it would be better to directly read the data from Godot. 

Also with the current `TempMaterialList` there is a thread race condition problem here.

shaders/IntercellularConnection.gdshader
hhyyrylainen, line 1: As this uses a custom shader, this should be added to `stage_resources.json` so that the shader is preloaded before gameplay starts.

Edit or rather `src/multicellular_stage/IntercellularConnection.tscn` should be added (assuming it renders correctly without setup).

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
dligr: It is, e.g. `MicrobeVisualsSystem` uses the same approach

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
dligr: I couldn't find a method that would return this data. There's only one other place where `ToTransform()` is used, so there probably isn't one

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
dligr: It should be safe, given that it's only used once, after spawning the link. Though, maybe it's a good idea to replace this operation with just setting the required rotation? Something like `connection.Quaternion = Quaternion.FromEuler(new Vector3(0.0f, angle, 0.0f))`

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
hhyyrylainen: I meant using code like `MicrobeColonyHelpers.CalculateColonyMemberAttachPosition` that is dealing with similar math than here.

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
hhyyrylainen: (I commented on the wrong thing)

I think that would be a bit better approach. But using the angle in the first place means that this math is made to work only on a 2D plane.

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
hhyyrylainen: Alright, that's I suppose fine then.

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
dligr: Actually, I've realized that it's possible to bypass any calculations by using `AttachedToEntity` data

hhyyrylainen: That sounds about right for what I had in mind, I totally forgot that was a thing I should have suggested earlier.

src/multicellular_stage/IntercellularConnection.tscn
```diff
@@ -0,0 +1,19 @@
...
+[ext_resource type="Texture2D" uid="uid://cwogwp0ip1n3p" path="res://assets/textures/IntercellularMatrix.png" id="1_2enjh"]
+[ext_resource type="Shader" uid="uid://ds2t62ln8y85p" path="res://shaders/IntercellularConnection.gdshader" id="1_ohjgk"]
+[ext_resource type="Texture2D" uid="uid://dy8uqfppw8wxk" path="res://assets/textures/IntercellularMatrixNormal.png" id="2_ohjgk"]
+
+[sub_resource type="ShaderMaterial" id="ShaderMaterial_62jel"]
+render_priority = 0
+shader = ExtResource("1_ohjgk")
+shader_parameter/mainTexture = ExtResource("1_2enjh")
+shader_parameter/normalTexture = ExtResource("2_ohjgk")
+shader_parameter/tint = Vector3(1, 1, 1)
```
hhyyrylainen, line 13: One more problem, this has to be local to scene, because otherwise all instances share the same material parameters and thus the system won't work. I'll do this edit myself if I don't find any other problems so that this can be merged quickly and make it in time for the RC build.

hhyyrylainen (approved):
I corrected that one small thing, and I think this is fine now.

PR #6260, opened 2025-08-07
Rest of basic achievements implementation

hhyyrylainen:

Fills out the list of basic achievements so that we have 12 of them

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/MicrobeStage.cs
```diff
@@ -1806,6 +1842,75 @@ private void SetupPlayerForGlucoseCollecting()
...
+        // Only report a food source when the player is not totally struggling for ATP
+        if (storage.Compounds.GetCompoundAmount(Compound.ATP) > 2)
+        {
+            if (radiationProduction > 1)
+            {
+                AchievementEvents.ReportPlayerUsesRadiation();
+            }
+
+            // Photosynthesis is pretty low for even plants
+            if (photosynthesisProduction > 0.15 && storage.Compounds.GetCompoundAmount(Compound.Glucose) >
```
dligr, line 1905: Is it correct to add a limit like this (and also `storage.Compounds.GetCompoundAmount(Compound.ATP) > 2`)? I think it might block some smaller cells from unlocking the achievement, even if they indeed produce more glucose that they consume

hhyyrylainen: I could do that, but I would kind of argue that if the player wants to just be 3 thylakoids, that doesn't really truly mean that they are a proper plant. So I think any reasonable cell will be able to pass this condition. I added this so that we don't give the achievement for cells that aren't ATP positive and are thus "cheating" in their glucose balance.

dligr: This makes sense. For better player experience though, the achievement description should probably include this limit. Maybe even state the exact minimum amount per second?

hhyyrylainen: The achievement descriptions have quite short max text lengths before they don't really fit where they go. So I wouldn't really want to include this kind of arbitrary limit. The achievement could be reworded as "becoming a photosynthetic plant" which would totally obscure the condition, but I think it is better to at least give the partial condition in the text.

Issue #6110, opened 2025-04-30
Add an option to select on which monitor the game window is on


hhyyrylainen:
One thing that would help quite a bit on Windows would be allowing the player to select the screen the game is on

Can be done with the following APIs:
```
DisplayServer.WindowSetCurrentScreen();
        DisplayServer.WindowGetCurrentScreen();
        DisplayServer.GetScreenCount()
```

I think the default should be kept as automatic window placement (this could be -1 when saved in the settings) and >= 0 values would then *once* apply the window position so that the game still allows manual changing of screens. Though also if the setting value is changed then the window should be moved.

PR #6261, opened 2025-08-07
Monitor selection option

OmerCD:

This PR should create a new option for Windows user in the options scene. It is under Graphics tab, called "Display". It should list all available displays for user to select and immediately move the game window to the selected display.
Maybe there is way but, I couldn't find a good way to list the name of the displays.

image

**Related Issues**
Closes https://github.com/Revolutionary-Games/Thrive/issues/6110

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/engine/Settings.cs
```diff
@@ -1026,6 +1034,32 @@ public void ApplyInputSettings()
     /// </summary>
     public void ApplyWindowSettings()
     {
+        var screenId = -1;
+        bool isOldSelectionDifferent = false;
+        if (OperatingSystem.IsWindows())
```
hhyyrylainen: Why is this only checking for Windows? Shouldn't this work on any desktop OS?

src/engine/Settings.cs
```diff
@@ -1026,6 +1034,32 @@ public void ApplyInputSettings()
...
+        {
+            var currentScreenId = DisplayServer.WindowGetCurrentScreen();
+            if (oldDisplaySelectionIndex == -1)
```
hhyyrylainen: What's the use for this old selection index? Cannot this just call `WindowGetCurrentScreen` and if the result is not what is desired, then call `WindowSetCurrentScreen`?

src/engine/Settings.cs
```diff
@@ -122,6 +124,12 @@ public enum AntiAliasingMode
     [JsonProperty]
     public SettingValue<float> RenderScale { get; private set; } = new(1.0f);
 
+    /// <summary>
+    /// Selected window index for the game window.
+    /// </summary>
+    [JsonProperty]
+    public SettingValue<int> SelectedDisplayIndex { get; set; } = new(DisplayServer.WindowGetCurrentScreen());
```
hhyyrylainen: I really think the default should be -1 and it should mean "auto" as on non-Windows platform it's going to be annoying if the game forces itself on a specific screen.

src/gui_common/menus/OptionsMenu.cs
```diff
@@ -33,6 +33,17 @@ public partial class OptionsMenu : ControlWithInput
         .GetOutputDeviceList().Where(d => d != Constants.DEFAULT_AUDIO_OUTPUT_DEVICE_NAME)
         .Prepend(Constants.DEFAULT_AUDIO_OUTPUT_DEVICE_NAME).ToList();
 
+    private static readonly Lazy<List<DisplayInfo>> DisplaysCache = new(() =>
```
hhyyrylainen: This should not be static because attaching new monitors should work without having to restart the game...

src/gui_common/menus/OptionsMenu.tscn
```diff
@@ -263,6 +265,7 @@ layout_mode = 2
 size_flags_horizontal = 3
 size_flags_vertical = 3
 follow_focus = true
+scroll_vertical = 100
```
hhyyrylainen, line 267: This one line chagne seems suspicious to me...

hhyyrylainen:
Here's the documentation on working with text in the game: https://github.com/Revolutionary-Games/Thrive/blob/master/doc/working_with_translations.md

src/engine/Settings.cs
OmerCD: I thought the original issue mentioned Windows; it would be a Windows specific operation. I think it would be good to remove it too.

src/gui_common/menus/OptionsMenu.tscn
OmerCD: I know. Whenever I open the OptionsMenu scene, Godot just changes this line for me. I always revert it but looks like I missed this one.

src/gui_common/menus/OptionsMenu.cs
OmerCD: I hadn't thought about that. Thanks.

src/engine/Settings.cs
hhyyrylainen: The issue originally mentions Windows, because players on Mac and Linux already have features in their operating system to fix the problem of the game on the wrong screen themselves. Only people who still use Windows have the problem, so the "fix" is mainly aimed at them. However the feature should still be quite workable on Mac and Linux as well, as long as the default behaviour is "auto" and not changing the game monitor.

hhyyrylainen:
I think this would be nice to get in by 1.0 so I'll help finish this today.

hhyyrylainen (approved):
I went through all the changes in this PR and improved everything I thought needed improving and that should have taken care of all of my comments. And I also added translations so this should be good to merge once CI jobs pass. 
Thanks for starting this work.

PR #6269, opened 2025-08-09
Ripple alpha fix

### LINKED ISSUE #6267: Ripple effect sometimes causes a faint square around cells
It looks like this:

Image

I think the solution is to reduce the base visibility of the shader effect so that the places with no waves will be truly fully transparent.

AverageNerdz:

This PR does some stuff...

**Related Issues**

Closes #6267

-->

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/MembraneWaterRipple.cs
```diff
@@ -424,6 +428,23 @@ private float CalculateTimeScale()
         return Mathf.Lerp(timeScale, 0.5f, stillnessFactor);
     }
 
+    // Non-linear alpha scaling function
```
hhyyrylainen:
```suggestion
    /// <summary>
    ///   Non-linear alpha scaling function
    /// </summary>
```

One small thing on the code side, comments on functions should use the XML comment syntax.

AverageNerdz: Omg, so sorry about that i was completely locked on making this fix asap that i overlooked that

hhyyrylainen (commented):
Unfortunately this does not fix the problem. I did kind of try to compare with master branch and I suppose this makes the squares more transparent, but does not remove them. So this helps a bit but I think that even with this change we'll have plenty of people report the squares problem.

AverageNerdz:
> Unfortunately this does not fix the problem. I did kind of try to compare with master branch and I suppose this makes the squares more transparent, but does not remove them. So this helps a bit but I think that even with this change we'll have plenty of people report the squares problem.
> 
> 

Did you use the same save or a different one? And how did it happen?

hhyyrylainen:
The same save as before, and I basically saw it just as soon as I loaded it. 
I've since tried it multiple times and it has been basically instant in every try. I've mostly just played up to like 10 seconds after loading the save and that's been enough to see the problem.

PR #6270, opened 2025-08-10
Make cell links take cell rotation into account

dligr:

Makes intercellular connections take cell rotation into account. Otherwise the connections are completely misaligned (unless all cells have the same rotation)

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/multicellular_stage/systems/IntercellularMatrixSystem.cs
+++ b/src/multicellular_stage/systems/IntercellularMatrixSystem.cs
@@ -66,19 +66,31 @@ public sealed class IntercellularMatrixSystem : AEntitySetSystem<float>
 
         Vector3 targetRelativePos;
 
+        Quaternion ourRotation = Quaternion.Identity;
+        Quaternion targetRotation = Quaternion.Identity;
+
         if (parentEntity == colony.Leader)
         {
-            targetRelativePos = -entity.Get<AttachedToEntity>().RelativePosition;
+            var ourAttachedPosition = entity.Get<AttachedToEntity>();
+
+            targetRelativePos = -ourAttachedPosition.RelativePosition;
+            ourRotation = ourAttachedPosition.RelativeRotation;
         }
         else
         {
-            targetRelativePos = parentEntity.Get<AttachedToEntity>().RelativePosition
-                - entity.Get<AttachedToEntity>().RelativePosition;
+            var ourAttachedPosition = entity.Get<AttachedToEntity>();
+            var targetAttachedPosition = parentEntity.Get<AttachedToEntity>();
+
+            targetRelativePos = targetAttachedPosition.RelativePosition
+                - ourAttachedPosition.RelativePosition;
+
+            ourRotation = ourAttachedPosition.RelativeRotation;
+            targetRotation = targetAttachedPosition.RelativeRotation;
         }
 
         Vector3 pointA, pointB;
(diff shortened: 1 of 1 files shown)


src/multicellular_stage/systems/IntercellularMatrixSystem.cs
```diff
@@ -66,19 +66,31 @@ private static void AddIntercellularConnection(in Entity entity, ref Intercellul
 
         Vector3 targetRelativePos;
 
+        Quaternion ourRotation;
+        Quaternion targetRotation = Quaternion.Identity;
+
         if (parentEntity == colony.Leader)
         {
-            targetRelativePos = -entity.Get<AttachedToEntity>().RelativePosition;
+            var ourAttachedPosition = entity.Get<AttachedToEntity>();
```
hhyyrylainen:
```suggestion
            ref var ourAttachedPosition = ref entity.Get<AttachedToEntity>();
```
Components need to be fetched like this to avoid extra memory copies.

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
```diff
@@ -66,19 +66,31 @@ private static void AddIntercellularConnection(in Entity entity, ref Intercellul
...
+
+            targetRelativePos = -ourAttachedPosition.RelativePosition;
+            ourRotation = ourAttachedPosition.RelativeRotation;
         }
         else
         {
-            targetRelativePos = parentEntity.Get<AttachedToEntity>().RelativePosition
-                - entity.Get<AttachedToEntity>().RelativePosition;
+            var ourAttachedPosition = entity.Get<AttachedToEntity>();
+            var targetAttachedPosition = parentEntity.Get<AttachedToEntity>();
```
hhyyrylainen: Same here

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
```diff
@@ -106,30 +118,33 @@ private static void AddIntercellularConnection(in Entity entity, ref Intercellul
...
-        Vector2 offset2D = new Vector2(membraneBOffset.X, membraneBOffset.Z);
-
         float min = float.MaxValue;
-        Vector2 pointA = Vector2.Zero;
-        Vector2 pointB = offset2D;
+        Vector3 pointA = Vector3.Zero;
+        Vector3 pointB = membraneBOffset;
         foreach (var a in membraneA.Vertices2D)
         {
+            var rotatedA = new Vector3(a.X, 0.0f, a.Y);
```
hhyyrylainen: Why is this variable called "rotated" because no matter how much I look at the math, I cannot see how this is any kind of rotated vector. It just looks like a plain vector with increased dimensions.

hhyyrylainen: Maybe this just be called `convertedA` ?

hhyyrylainen (changes requested):
Functionally this seems fine, but there's a few small code related things to fix.

src/multicellular_stage/systems/IntercellularMatrixSystem.cs
dligr: Yeah, the vector really isn't rotated (it was in the first approach, that's where the name is from). `convertedA` should be a good alternative.

hhyyrylainen (approved):

PR #6278, opened 2025-08-12
Currents overhaul

dligr:

- Makes currents' noise textures 2D instead of 3D for less repetition.
- Current shifting is now done by moving two noise textures and multiplying them with eachother
- Somewhat improves noise textures to make them more interesting (like currents going on for longer and having some whirlpool-like areas)
- Enables trails (for surface patches mostly), also colors some currents

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/FluidCurrentsSystem.cs
+++ b/src/microbe_stage/systems/FluidCurrentsSystem.cs
@@ -101,22 +101,11 @@ public sealed class FluidCurrentsSystem : AEntitySetSystem<float>
 
         if (!imagesInitialized)
         {
-            var newCurrentsNoise1 = currentsNoise1Texture.GetImage();
-            var newCurrentsNoise2 = currentsNoise2Texture.GetImage();
-
-            noiseWidth = newCurrentsNoise1.GetWidth();
-            noiseHeight = newCurrentsNoise2.GetHeight();
+            currentsNoise1 = currentsNoise1Texture.GetImage();
+            currentsNoise2 = currentsNoise2Texture.GetImage();
+            noiseWidth = currentsNoise1.GetWidth();
+            noiseHeight = currentsNoise2.GetHeight();
             imagesInitialized = true;
-
-            if (newCurrentsNoise1 == null || newCurrentsNoise2 == null)
-            {
-                GD.PrintErr("Current noise image created from texture is null");
-            }
-            else
-            {
-                currentsNoise1 = newCurrentsNoise1;
-                currentsNoise2 = newCurrentsNoise2;
-            }
         }
 
         currentsTimePassed += delta;


Patryk26g:
I checked the look of it and I think it looks really nice, but when I go faster with my cell to new areas, the currents are not present there and only appear after a second o two. is it possible for them to cover a bigger area so that they are visible wherever player goes?

dligr:
So, I increased the extents of the particle spawn area by 1.5 and I think that this fixes the problem. Unfortunately, increasing the box extents also requires increasing particle count, so if the issue persists in some way, the solution might be to move the box slightly in front of the player to let particles spawn in advance.

src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -899,7 +899,15 @@ private void PartialAdvectCenter(int x0, int y0, int width, int height, float de
                 if (OldDensity[x, y].LengthSquared() > 1)
                 {
                     var velocity = fluidSystem!.VelocityAt(
-                        new Vector2(worldPos.X + x * resolution, worldPos.Z + y * resolution)) * VISCOSITY;
+                        new Vector2(worldPos.X + x * resolution, worldPos.Z + y * resolution));
+
+                    if (MathF.Abs(velocity.X) + MathF.Abs(velocity.Y) <
```
hhyyrylainen, line 904: What does this math accomplish? To me this looks really bad as this is a nested loop that now has a bunch of extra calculation and branching here. This is surely not good for CPU performance of the clouds. The clouds are already quite a killer for weaker CPUs so I'd like to do everything possible to not increase their CPU impact.

src/microbe_stage/MicrobeStage.tscn
```diff
@@ -1,9 +1,11 @@
-[gd_scene load_steps=17 format=3 uid="uid://dva08tpqyn15k"]
+[gd_scene load_steps=22 format=3 uid="uid://dva08tpqyn15k"]
 
 [ext_resource type="Script" uid="uid://di2y5gnlbsclr" path="res://src/microbe_stage/MicrobeStage.cs" id="1"]
 [ext_resource type="PackedScene" uid="uid://bako5jivv1dji" path="res://src/microbe_stage/MicrobeCamera.tscn" id="2"]
 [ext_resource type="PackedScene" uid="uid://bt2fv82h01ice" path="res://src/microbe_stage/MicrobeHUD.tscn" id="3"]
 [ext_resource type="PackedScene" uid="uid://dew8qwd0buvnb" path="res://src/thriveopedia/fossilisation/FossilisationDialog.tscn" id="4"]
+[ext_resource type="Shader" uid="uid://b12xtoy67lhb" path="res://shaders/HeatGradient.gdshader" id="4_xncm1"]
```
hhyyrylainen: I think you got hit with the Godot automatically duplicating these resources. So could you reset the heat stuff to default values and then save this scene again? That should remove all changes from this file.

src/microbe_stage/systems/FluidCurrentsSystem.cs
```diff
@@ -175,48 +153,14 @@ protected override void Update(float delta, in Entity entity)
...
-            noiseDisturbancesXImage[i] = disturbancesX[i];
-            noiseDisturbancesYImage[i] = disturbancesY[i];
-            noiseCurrentsXImage[i] = currentsX[i];
-            noiseCurrentsYImage[i] = currentsY[i];
-        }
-
-        imagesInitialized = true;
-    }
-
-    [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
hhyyrylainen, line 216: I think this aggressive inline should be kept on the `GetPixel` method.

src/microbe_stage/CompoundCloudPlane.cs
dligr: This cut-off condition is basically a slight modification of the one that was removed from the currents system (to make weaker currents push players/chunks as well). So, I don't think performance would be impacted.

hhyyrylainen: How would performance not be impacted? This is a doubly nested loop for each simulation coordinate of a compound cloud, and it adds 2 Abs calls, one if branch, and a variable assignment. If the branch predictor in a CPU doesn't correctly catch onto when the if passes, that can have a major impact on how fast in total this loop runs.

dligr: There was already a cut-off check in the `VelocityAt()` method; that check was moved here so that it wouldn't affect microbes. As such, compared to what was before this PR, this system isn't more performance-intensive. Might even be less so since it has no multiplication.

hhyyrylainen: Ah, I finally understood. So the code from `VelocityAt` method was just moved here, so that is a neutral operation in terms of computation cost.

So this is all fine then.

src/microbe_stage/systems/FluidCurrentsSystem.cs
```diff
@@ -126,7 +101,15 @@ protected override void PreUpdate(float delta)
 
         if (!imagesInitialized)
         {
-            TryGetNoiseImages();
+            currentsNoise1 = currentsNoise1Texture.GetImage();
+            currentsNoise2 = currentsNoise2Texture.GetImage();
+
+            if (currentsNoise1 != null && currentsNoise2 != null)
```
hhyyrylainen: As these are from a texture, these should never fail, right? So I think this if should be removed. If you want you could leave a GD.PrintErr here if an image is null from a texture.

dligr: In our case, the image should never be null, but just in case, I added an error print (because, according to the documentation, `GetImage()` can return null)

PR #6286, opened 2025-08-17
Membrane graphics additions

CI09:

The bending-on-turn effect and possible internal decoration mesh

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Patryk26g:
I have tested the visuals and I think it looks nice, but I have some comments:
- now the organelles look as if they were on top of the membrane instead of inside the cell, I think the internal decoration mesh might be the cause of that. Maybe the solution would be to add another decoration mesh on top of the organelles that would me more transparent?
- cell previews have less pronouced membrane texture and in some cases it's barely visible
- different types of membranes are less discernible

CI09:
Ah you didn't have to review yet, I'm still debating whether to actually include any changes other than membrane turning. But thanks for review

Patryk26g:
I know, I just wanted to check out of curiosity :D

hhyyrylainen:
> different types of membranes are less discernible

I think that this is a pretty big concern as even now we have complaints about the current membrane graphics being too similar (ever since the 3D membranes implementation). So going even more in the direction of different membranes looking the same has to be avoided.

CI09:
It should work now. The bug where Luca species preview is still with the old membrane is still there though. I found that changing the number to something else at https://github.com/Revolutionary-Games/Thrive/blob/master/src/general/Species.cs#L336 can fix that but I don't know if it's the right place to change something

hhyyrylainen:
That's not the right place. There should be a constant in the Constants file that controls a base hash for cells (it's next to a base value for cell layout views).

CI09:
Should do now, but visibility is still kinda poor in the preview anyways. Main stuff can be tested though

Deus-Codes:
Tried this for a bit - absolutely love it, great work. Makes our microbes look very alive and dynamic. A few pieces of feedback...

- Membranes did look discernable in my eye, though we should always keep an extra eye out here; perhaps a certain color makes this more of a problem. 
- I noticed that the color can look a bit more transparent in the middle of the cell than I remember. That could lead to our cells looking very water-ballooney. That isn't inherently a bad thing - I do like the look, and, well, cells are pouches of goop - but it is definitely a stylistic choice, and it could harm visibility in lighter patches. It might be worthwhile to investigate making the color a bit more pronounced. 
- I do like the speckled texture within the membrane, but I do think it can be improved - how, I'm not sure. More numerous and extending more outwards? More strong, or more transparent? I'd definitely accept the texture as it is, but I wonder if there is room for finding some cool visual enhancements there.

Overall, this is great. Incase you haven't seen it, I'll paste this concept art here, which many people seem to really like. Perhaps it could serve as some inspiration for tweaks - not that I absolutely think there should be an overhaul, and I'm not saying that you should make your pull request look like this. Just a neat visual that might get the brain flowing. 

image

Deus-Codes:
Told you this a bit ago, but wanted to write here that I love this change.

CI09:
I don't know if it looks good enough though

assets/textures/DoubleMembrane.png.import
```diff
@@ -2,22 +2,20 @@
...
 type="CompressedTexture2D"
-uid="uid://cuaou6m07rri7"
-path.s3tc="res://.godot/imported/DoubleMembrane.png-f2a8a15af2da0aea17cabf17eaffed38.s3tc.ctex"
-path.etc2="res://.godot/imported/DoubleMembrane.png-f2a8a15af2da0aea17cabf17eaffed38.etc2.ctex"
+uid="uid://c1y224pma1v5t"
+path="res://.godot/imported/DoubleMembrane.png-f2a8a15af2da0aea17cabf17eaffed38.ctex"
 metadata={
-"imported_formats": ["s3tc_bptc", "etc2_astc"],
-"vram_texture": true
+"vram_texture": false
```
hhyyrylainen, line 8: I think it is wrong to change this to not be a vram texture... Why was this change done?

shaders/EngulfEffect.gdshader
```diff
@@ -24,6 +25,13 @@ void vertex(){
         * wigglyNess * size;
     VERTEX.z += sin(worldVertex.x * movementWigglyNess - TIME / 4.0f) / 10.f
         * wigglyNess * size;
+
+		// Bend when turning
```
hhyyrylainen, line 29: Mixed tabs into the file here.

shaders/MucocystEffect.gdshader
```diff
@@ -1,10 +1,11 @@
 shader_type spatial;
 render_mode unshaded;
 
-// Any changes to these two must be matched in Membrane.gdshader
+// Any changes to these two must be matc  hed in Membrane.gdshader
```
hhyyrylainen, line 4:
```suggestion
// Any changes to these two must be matched in Membrane.gdshader
```

src/microbe_stage/systems/MicrobeVisualsSystem.cs
```diff
@@ -99,8 +99,15 @@ public override void Dispose()
     [Query]
     [All<CellProperties, SpatialInstance, EntityMaterial, RenderPriorityOverride>]
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    private void Update(ref OrganelleContainer organelleContainer, in Entity entity)
+    private void Update([Data] in float delta, ref OrganelleContainer organelleContainer, in Entity entity)
     {
+        ref var cellProperties = ref entity.Get<CellProperties>();
+
+        if (cellProperties.CreatedMembrane != null)
```
hhyyrylainen, line 106: I don't think it should be the visual *creation* system that applies turning property to the visuals, it should be a different system.

PR #6306, opened 2025-08-24
Fix current desync after save load

dligr:

Makes the current display receive information from the system in order to perfectly sync time between them.

My first approach was to make the display save time, but there was still desynchronization, because systems and in-game objects start to run at slightly different time.

**Related Issues**

Current display's time was reset after save load while current system's time was saved between sessions.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen (approved):
Not that I have noticed any problems related to this, this change seems fine to me.

PR #6312, opened 2025-08-26
Planet compounds customization and statistics

Patryk26g:

Adds statistics to planet customization section and also options to customize the compound abundance of the planet. Also applies variation to the starting conditions of planet's patches

image

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/gui/PatchDetailsPanel.cs
```diff
@@ -416,22 +416,22 @@ private void UpdatePatchDetails()
         // Compounds
         hydrogenSulfideLabel.Text =
             Math.Round(GetCompoundAmount(SelectedPatch, Compound.Hydrogensulfide),
-                Constants.PATCH_CONDITIONS_COMPOUND_DISPLAY_DECIMALS).ToString(CultureInfo.CurrentCulture);
+                Constants.COMPOUND_DISPLAY_DECIMALS).ToString(CultureInfo.CurrentCulture);
```
hhyyrylainen: I think this rename is not actually good because many places that display compounds will not use this. So I think the name should keep "patch" in it to be clear that the decimals value is only used in patch displays.

src/microbe_stage/BiomeConditions.cs
```diff
@@ -320,22 +338,34 @@ public void ApplyLongTermCompoundChanges(Biome biomeDetails, Dictionary<Compound
     ///   Modifies the long-term amount of a compound in this biome. This is preferable to directly modifying
     ///   <see cref="ChangeableCompounds"/>. This is only usable for compounds that don't vary along an in-game day.
     /// </summary>
-    /// <param name="compound">The compound to modify</param>
-    /// <param name="newValue">New value to set</param>
-    public void ModifyLongTermCondition(Compound compound, BiomeCompoundProperties newValue)
+    /// <param name="compound">
+    ///   The compound to modify
```
hhyyrylainen: Short param definitions shouldn't be split across multiple lines.

```suggestion
    ///   <param name="compound">The compound to modify</param>
```

src/general/WorldGenerationSettings.cs
```diff
@@ -130,6 +130,27 @@ public enum ClimateInstabilityEnum
         High = 2,
     }
 
+    /// <summary>
+    ///   The possible compound levels settings for the planet
+    /// </summary>
+    public enum CompoundLevelsEnum
```
hhyyrylainen: I think this should be just named `CompoundLevel` as the "enum" suffix in here in the others is used just as a way to avoid name conflicts, but this enum to me doesn't seem to have a name conflict so this should follow the usual naming of *not* having `Enum` suffix in the name.

src/general/NewGameSettings.cs
```diff
@@ -338,6 +349,14 @@ public void OpenFromDescendScreen(GameProperties currentGame)
         planetSettings.SetWorldSeaLevel((int)settings.WorldOceanicCoverage);
         planetSettings.SetWorldGeologicalActivity((int)settings.GeologicalActivity);
         planetSettings.SetWorldClimateInstability((int)settings.ClimateInstability);
+
+        planetSettings.SetSulfideLevel((int)settings.SulfideLevel);
```
hhyyrylainen: Why are these converted to ints here? I think it would be better to consistently use the enum in the settings. (even if just above this that is not done, I think using better patterns in new code is preferable)

locale/messages.pot
```diff
@@ -8077,6 +8191,10 @@ msgstr ""
 msgid "SUICIDE_BUTTON_TOOLTIP"
 msgstr ""
 
+#: ../src/gui_common/menus/PlanetSettings.tscn:437
+msgid "SULFIDE_LEVELS"
```
hhyyrylainen: We don't have sulfide in the game so I think for clarity this translation key should be "HYDROGEN_SULFIDE_..."

src/gui_common/menus/PlanetSettings.cs
```diff
@@ -106,6 +133,36 @@ public void SetWorldClimateInstability(int index)
         worldClimateInstabilityButton.Selected = index;
     }
 
+    public void SetSulfideLevel(int index)
```
hhyyrylainen: I see now why the `(int)` converion is done... These assume the enum value matches the GUI indexes exactly. I would like to see a safer approach here, where these would be given the enum value in and then they internally convert it to indexes. And this should have big warning comments about the indexes needing to always match what is in the tscn file.

Patryk26g:
I have fixed the reviewed issues but still have one problem with CI:

```
    private void OnPlanetSettingsChanged(int index)
    {
        EmitSignal(SignalName.PlanetSettingChanged);
    }
```

This method requires an int argument, because only this way can it be connected to the item_selected(int) signal. However, I don't need that at all and I don't want to emit the signal with it, because I don't in other places and if I do, then it doesn't work...

The only solution I came up with is:

```
#pragma warning disable SA1313
    private void OnPlanetSettingsChanged(int _)
    {
        EmitSignal(SignalName.PlanetSettingChanged);
    }
#pragma warning restore SA1313
```

hhyyrylainen:
To suppress unused variable warnings, use syntax like this:

```c#
_ = index;
```

src/gui_common/menus/PlanetSettings.cs
```diff
@@ -76,34 +103,64 @@ public WorldGenerationSettings GetPlanetSettings()
         };
     }
 
-    public void SetLifeOrigin(int index)
+    public void SetLifeOrigin(WorldGenerationSettings.LifeOrigin value)
+    {
+        lifeOriginButton.Selected = (int)value;
```
hhyyrylainen, line 108: These will still just blow up if someone edits the order of the enum or adds new entries. I think one of the following should be done:
- Very loud comments need to be added to the all relevant enums that new values / reordering *may not be done* without also changing the GUI scenes
- A manual switch case could be added in each of the methods here to convert the values and print errors on unknown values (with safe fallbacks)

simulation_parameters/Constants.cs
```diff
@@ -1826,6 +1827,24 @@ public static class Constants
     public const float TOLERANCE_MAXIMUM_SURVIVABLE_PRESSURE_DIFFERENCE = 4000000;
     public const float TOLERANCE_PERFECT_THRESHOLD_PRESSURE = 350000;
 
+    public const int TEMPERATURE_COLD_MIN = -4;
```
hhyyrylainen: I think these should be placed nearby to planet generator variables, because right now it doesn't make sense to insert these between `TOLERANCE_PERFECT_THRESHOLD_PRESSURE` and `TOLERANCE_INITIAL_PRESSURE_MIN_FRACTION` as those are two *very* closely linked constants to tolerance pressure specifically.

Patryk26g:
I have added comment to the enums and moved the constants

hhyyrylainen (approved):
Everything looks fine now.

PR #6315, opened 2025-08-28
Adjust Engulfing Tutorial

### LINKED ISSUE #4916: Make sure engulfing tutorial is not too difficult by pointing to a small cell
Maybe either the engulfing tutorial shouldn't be able to point to a cell at all or only to cells above a given size (so that the speed difference between the player and the target isn't super high). That way it would ensure the tutorial doesn't point to a small super fast cell that is hard to eat. I had difficulty on the Thrive podcast 31 trying to complete the tutorial as I had like 5 organelles and I was trying to eat a single hex cell.

Or maybe it'd be enough to compare the speed of the player versus the other cells?

dylanlee082:

This PR adds an additional check to the engulfing tutorial to only select a closest object to engulf if it is bigger than the player object size * .4. 

Hi! As this is my first time contributing I had some questions, hopefully this is the right place for them.
Questions:
- Should the .4 value be in a CONSTANT?
- I took a bit of a guess with .4. I tried using half the player objects size and found that many of the other objects in the beginning of the game are size 2 and the player object is size 5, so with .5 all of the objects are smaller than 2.5 and therefore can't be engulfed. It's also hard that the CanEngulfObject size with the ratio is typically 3. Is there a more deterministic way to come up with a value for this?
- The issue description talks about maybe using the player and cell speed to determine which objects to use, but I can't find where in the codebase speed is calculated.

**Related Issues**

closes #4916 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/components/Engulfer.cs
+++ b/src/microbe_stage/components/Engulfer.cs
@@ -189,6 +189,13 @@ public static class EngulferHelpers
             if (!compounds.Compounds.Any(p => usefulCompoundSource.IsUseful(p.Key)))
                 continue;
 
+            // Skip entities that are too small to catch easily
+            // TODO: Does this .4 need to be in a CONSTANT?
+            // TODO: What is a deterministic way to come up with a better guess than .4 the size of the player object
+            // TODO: How is player object and the engulfable objects speed calculated to maybe compare a percentage of that instead?
+            if ((engulfer.EngulfingSize / .4) > engulfable.AdjustedEngulfSize)
+                continue;
+
             if (nearestPoint == null || distance < nearestDistanceSquared)
             {
                 nearestPoint = entityPosition.Position;
--- a/src/microbe_stage/components/Engulfer.cs
+++ b/src/microbe_stage/components/Engulfer.cs
@@ -190,10 +190,7 @@ public static class EngulferHelpers
                 continue;
 
             // Skip entities that are too small to catch easily
-            // TODO: Does this .4 need to be in a CONSTANT?
-            // TODO: What is a deterministic way to come up with a better guess than .4 the size of the player object
-            // TODO: How is player object and the engulfable objects speed calculated to maybe compare a percentage of that instead?
-            if ((engulfer.EngulfingSize / .4) > engulfable.AdjustedEngulfSize)
+            if (engulfer.EngulfingSize / .4 > engulfable.AdjustedEngulfSize)
                 continue;
 
             if (nearestPoint == null || distance < nearestDistanceSquared)


hhyyrylainen:
> Should the .4 value be in a CONSTANT?

Yes, it might need future tweaking so placing it in `Constants.cs` would be preferable.

> I took a bit of a guess with .4. I tried using half the player objects size and found that many of the other objects in the beginning of the game are size 2 and the player object is size 5, so with .5 all of the objects are smaller than 2.5 and therefore can't be engulfed. It's also hard that the CanEngulfObject size with the ratio is typically 3. Is there a more deterministic way to come up with a value for this?

Engulfing size is based on the adjusted hex size of an object. There's a constant that specifies how much bigger you need to be. This is the code that does that check:

```c#
            if (engulfer.EngulfingSize < targetSize * Constants.ENGULF_SIZE_RATIO_REQ)
                return EngulfCheckResult.TargetTooBig;
```

> The issue description talks about maybe using the player and cell speed to determine which objects to use, but I can't find where in the codebase speed is calculated.

Well that's quite complicated... we don't really have an accurate static calculation method for speed. So basing a speed estimate on size is likely the best option (I didn't look at the changes yet but that sounds like exactly what you did already).

There's a separate issue about fixing the speed calculation display in the editor, but this is quite a hard issue to tackle:

https://github.com/Revolutionary-Games/Thrive/issues/2095

---

I'll review the code changes in this PR sometime soon. Thanks for the interest in contributing to Thrive.

src/microbe_stage/components/Engulfer.cs
```diff
@@ -189,6 +189,10 @@ public static EngulfCheckResult CanEngulfObject(this ref Engulfer engulfer, uint
             if (!compounds.Compounds.Any(p => usefulCompoundSource.IsUseful(p.Key)))
                 continue;
 
+            // Skip entities that are too small to catch easily
+            if (engulfer.EngulfingSize / .4 > engulfable.AdjustedEngulfSize)
```
hhyyrylainen: Putting this here will impact AI behaviour as well. Perhaps it would be good to have a new parameter for this method to `skipLikelyTooFastTargets`? That way the info that is fed through the tutorial to the player could use that flag and the microbe AI, for now at least, could keep the old behaviour?

dylanlee082: Done, although I am unable to find where the AI is using this function. It looks like MicrobeAISystems has it's own function for doing this work called, GetNearestPreyItem.

dylanlee082:
For the tutorial it seems odd to have so much of this logic when it would potentially be easier to spawn new entities as needed for the tutorial. I looked through some of the other tutorials, like the initial glucose one, and it seems like that just isn't the pattern used so far in the game, but is that something you guys have already thought through?

src/microbe_stage/components/Engulfer.cs
hhyyrylainen: Yeah, that's probably due to efficiency. The microbe AI uses a shared cache of objects. Still, it is good here to add that parameter as the name of this method is pretty general so it would be surprising to by default filter by predicted speed.

hhyyrylainen:
> For the tutorial it seems odd to have so much of this logic when it would potentially be easier to spawn new entities as needed for the tutorial. I looked through some of the other tutorials, like the initial glucose one, and it seems like that just isn't the pattern used so far in the game, but is that something you guys have already thought through?

Well way back in the day like 2015, we had a separate tutorial environment the player was first spawned in. In that environment none of the usual gameplay things were running for absolute control of the tutorial. But since the engine changes the new tutorial was made from scratch has a reimplementation. And the main idea behind the tutorial was that it would provide context-dependent short tips the player needs to get started with Thrive. Since then due to people asking for it I've added some tutorials with way more text, but the primary way the tutorial is constructed still so that it just reacts to the game state and introduces parts of the game as they become relevant.

So yeah, our tutorial does not cheat at all, it just points out key game mechanics as they naturally happen. This also has the benefit that the tutorial is not detached from the actual game and doesn't require a separate gameplay environment to work.

simulation_parameters/Constants.cs
```diff
@@ -1455,6 +1455,13 @@ public static class Constants
     /// </summary>
     public const float MOVEMENT_MODE_SELECTION_DELAY = 1.55f;
 
+    /// <summary>
+    ///   The size ratio an entity has to be in relation to the player for the engulfable tutorial to point to it.
+    ///   This is to increase the chances of catching the target entity during the tutorial to reduce player
+    ///   frustration.
+    /// </summary>
+    public const float TUTORIAL_ENGULFABLE_SIZE_RATIO = .4f;
```
hhyyrylainen:
```suggestion
    public const float TUTORIAL_ENGULFABLE_SIZE_RATIO = 0.4f;
```

src/microbe_stage/MicrobeStage.cs
```diff
@@ -361,7 +361,7 @@ public override void _Process(double delta)
                     var position = engulfer.FindNearestEngulfableSlow(ref Player.Get<CellProperties>(),
                         ref Player.Get<OrganelleContainer>(), ref Player.Get<WorldPosition>(),
                         Player.Get<CompoundStorage>().Compounds, Player, Player.Get<SpeciesMember>().ID,
-                        WorldSimulation);
+                        WorldSimulation, skipLikelyTooFastTargets: true);
```
hhyyrylainen, line 364:
```suggestion
                        WorldSimulation, true);
```

src/microbe_stage/components/Engulfer.cs
```diff
@@ -189,6 +190,11 @@ public static EngulfCheckResult CanEngulfObject(this ref Engulfer engulfer, uint
             if (!compounds.Compounds.Any(p => usefulCompoundSource.IsUseful(p.Key)))
                 continue;
 
+            // Skip entities that are too small to catch easily
+            if (skipLikelyTooFastTargets &&
+                engulfer.EngulfingSize / Constants.TUTORIAL_ENGULFABLE_SIZE_RATIO > engulfable.AdjustedEngulfSize)
+                continue;
+
```
hhyyrylainen:
```suggestion
                engulfer.EngulfingSize / Constants.TUTORIAL_ENGULFABLE_SIZE_RATIO > engulfable.AdjustedEngulfSize)
                {
                    continue;
                }

```

src/microbe_stage/components/Engulfer.cs
```diff
@@ -189,6 +190,13 @@ public static EngulfCheckResult CanEngulfObject(this ref Engulfer engulfer, uint
             if (!compounds.Compounds.Any(p => usefulCompoundSource.IsUseful(p.Key)))
                 continue;
 
+            // Skip entities that are too small to catch easily
+            if (skipLikelyTooFastTargets &&
+                engulfer.EngulfingSize / Constants.TUTORIAL_ENGULFABLE_SIZE_RATIO > engulfable.AdjustedEngulfSize)
+                {
```
hhyyrylainen: Well I missed up here with the Github code editor.

I'll pull the changes locally to fix this and do some testing if the gameplay change is now good.

dylanlee082:
Turns out I had a bug. My original approach on this was to divide cell size by 2 and then compare and when I realized that wasn't going to work I switched the value to .4, but then never switched from dividing to multiplying. This now fixed. I also dropped the constant from .4 to .35 to ensure it will find more entities.

hhyyrylainen (approved):
I think this is pretty well tuned now. As long as there's some engulf candidates in the first two generations this seems to always trigger. Though if the player goes for full organelle amounts for 3 generations then this tutorial might not trigger until the player adds iron organelles, but I think this is about tuned well enough to trigger for most players, I think. 

Thanks for improving this tutorial.

PR #6317, opened 2025-08-29
Make hydrogen sulfide clouds deal damage

dligr:

Makes hydrogen sulfide clouds deal damage to cells that don't have chemosynthesizing proteins.

This doesn't mean much for sulfur-based metabolism, but sulfur itself is now a bit more unique.

**Related Issues**

Discussion in [this thread](https://forum.revolutionarygamesstudio.com/t/wrapping-up-the-microbe-stage/1170)

**Progress Checklist**

      break existing features:
      members reviewing this PR)


HyperbolicHadron:
I didn't test for long, but I did check that damage is taken without chemosynthesizing proteins and immunity is granted when adding them. The damage taken seems to be reasonable on Normal difficulty, though I wonder if it might be a bit too high on Easy difficulty. (I wasn't able to notice if Easy difficulty changed the damage taken by a noticeable amount, it seemed the same to me)

Deus-Codes:
Tested this for about 20 minutes; very good work, and it's cool to see a relatively simple tweak have a solid effect on gameplay, especially early on. Sulfur has been a pretty under-served metabolism, so this definitely adds a strong characteristic to it which can dramatically influence how we approach it. 

I think damage was fair; again, it doesn't need to be a gigantic hazard that you're likely to die to, just enough to deter the player. This is simple hazard to avoid just by placing one chemosynthesizing part, which is how it should be for a cloud. 

We do probably want to attach some sort of reason why a player shouldn't be incentivized to just keep one part in their organism with no consequences for the rest of their game. I think adding some oxygen tolerance debuff would be a good way of doing this. I can do that myself though when I generally tackle balancing metabolism, which is much needed. Balancing the availability of sulfur, how quickly it deteriorates, and environmental events adding sulfur as a hazard event, is also another control we can account for. 

Not saying the above to make you implement this all by the way. As is, this is a PU I think should be accepted.

HyperbolicHadron:
> We do probably want to attach some sort of reason why a player shouldn't be incentivized to just keep one part in their organism with no consequences for the rest of their game.

Having the first part cost a significant amount more than consecutive parts would also help out in this area, since it would not be a cheap investment to unlock the first chemosynthesizing part just to provide immunity.

Deus-Codes (approved):

simulation_parameters/microbe_stage/organelles.json
```diff
@@ -197,6 +197,9 @@
         "Capacity": 0.25
       }
     },
+    "FeatureTags": [
```
hhyyrylainen, line 200: Shouldn't the eukaryotic part variant also provide the protection?

src/microbe_stage/systems/CompoundAbsorptionSystem.cs
hhyyrylainen, line 1: I feel like it is wrong to put the code in here like this.

In my opinion a cleaner approach would be to mark hydrogen sulfide compound as always useful, and then something like osmoregulation and healing system could also deal damage if the cell is storing hydrogen sulfide and has no protection against it. To me it feels wrong to put this one special kind of damage type into the base absorption system.

hhyyrylainen:
Also one thing that might need considering is the AI. Taking in hydrogen sulfide and being damaged should make the AI move in a random roughly forwards direction at full speed for a few seconds. Because if the AI stops in a hydrogen sulfide cloud it will be really strange for the player to see the AI just want to die. Even fully sessile cells I think need to attempt this manoeuvre.

src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -734,19 +734,22 @@ public void AbsorbCompounds(int localX, int localY, CompoundBag storage,
                 break;
 
             // Skip if compound is non-useful or disallowed to be absorbed
-            if (!compoundDefinitions[i]!.IsAbsorbable || !storage.IsUseful(compound))
+            if (!compoundDefinitions[i]!.IsAbsorbable || (!storage.IsUseful(compound) && compound != Compound.Hydrogensulfide))
```
hhyyrylainen: I think this line can now be reverted?

src/microbe_stage/systems/CompoundAbsorptionSystem.cs
```diff
@@ -25,6 +25,11 @@ public CompoundAbsorptionSystem(CompoundCloudSystem compoundCloudSystem, World w
         this.compoundCloudSystem = compoundCloudSystem;
     }
 
+    protected override void PreUpdate(float delta)
```
hhyyrylainen: This looks like it can also be reverted.

src/microbe_stage/systems/CompoundAbsorptionSystem.cs
dligr: What I've noticed is that there are some side effects to marking hydrogen sulfide as always useful. For one thing, it makes the hydrogen sulfide storage bar always visible. It also seems like AI now always considers hydrogen sulfide to be useful and ends up taking damage trying to collect it.

While there are possible workarounds, would implementing them be a cleaner solution than detecting sulfide clouds in the absorption system?

hhyyrylainen: Yeah, I think that is probably preferable. There's been talk in the past about converting toxins into clouds as well, and for that a "forced absorption" mechanic would be useful. 

So I'd say that would be worth it to do now: adding a new flag for a Compound that forces it to be always absorbed. Maybe calling it something like `AlwaysAbsorbed` or something about forced absorption would be suitable?

src/microbe_stage/CompoundCloudPlane.cs
dligr: This is probably no longer relevant since the condition was changed (now it has to be present, because hydrogen sulfide isn't `AlwaysUseful`)

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -343,6 +343,15 @@ private void ChooseActions(in Entity entity, ref MicrobeAI ai, ref CompoundAbsor
             }
         }
 
+        // If motionless inside a hydrogen sulfide cloud, move away
+        if (!organelles.HydrogenSulfideProtection
+            && compounds.GetCompoundAmount(Compound.Hydrogensulfide) > MathUtils.EPSILON
+            && control.MovementDirection != Vector3.Zero)
```
hhyyrylainen: Isn't this last condition the wrong way around? Doesn't this now check that if the cell *is* moving?

src/microbe_stage/MicrobeHUD.cs
```diff
@@ -424,6 +424,18 @@ protected override void UpdateBarVisibility(Func<Compound, bool> isUseful)
...
 
         ingestedMatterBar.Visible = GetPlayerUsedIngestionCapacity() > 0;
+
+        foreach (var (compound, bar) in compoundBars)
+        {
+            if (bar.Visible)
+                continue;
+
+            if (SimulationParameters.GetCompound(compound).AlwaysAbsorbable
+                && GetPlayerStorage().GetCompoundAmount(compound) > MathUtils.EPSILON)
```
hhyyrylainen: `GetPlayerStorage()` should be outside the loop in my opinion (and stored in a variable)

src/microbe_stage/CompoundCloudPlane.cs
```diff
@@ -734,7 +734,8 @@ public void AbsorbCompounds(int localX, int localY, CompoundBag storage,
                 break;
 
             // Skip if compound is non-useful or disallowed to be absorbed
-            if (!compoundDefinitions[i]!.IsAbsorbable || !storage.IsUseful(compound))
+            if (!compoundDefinitions[i]!.IsAbsorbable
+                || (!storage.IsUseful(compound) && !compoundDefinitions[i]!.AlwaysAbsorbable))
                 continue;
```
hhyyrylainen, line 740:
```suggestion
            {
                continue;
            }
```

PR #6319, opened 2025-08-30
Avoid context specific songs playing in a row

### LINKED ISSUE #5615: Make context-specific selected music respect option to not play the same track twice in a row
as with the initial implementation there is a chance to play a context specific music track in a row again and again. It should be fixed so that if a context specific music was just played then when the track list is random or sequential mode that same track couldn't be selected at once.

dylanlee082:

Add a check for context specific songs to ensure they are only played once in the category rotation. Once all songs have been played in a category and played once is reset then the context specific song would be played again.

NOTE: It seems like the randomness of this should be removed and always have the context specific song play first and then allow other songs to be played. 

**Related Issues**

closes #5615 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/Jukebox.cs
+++ b/src/general/Jukebox.cs
@@ -558,7 +558,7 @@ public partial class Jukebox : Node
         if (random.NextFloat() <= Constants.CONTEXTUAL_ONLY_MUSIC_CHANCE)
         {
             var contextMusicOnly = list.GetTracksForContexts(activeContexts)
-                .Where(c => c is { ExclusiveToContexts: not null, PlayedOnce: false })
+                .Where(c => c.ExclusiveToContexts != null && !c.PlayedOnce)
                 .ToArray();
 
             if (contextMusicOnly.Length > 0)
--- a/src/general/Jukebox.cs
+++ b/src/general/Jukebox.cs
@@ -557,15 +557,13 @@ public partial class Jukebox : Node
 
         if (random.NextFloat() <= Constants.CONTEXTUAL_ONLY_MUSIC_CHANCE)
         {
-            var contextMusicOnly = list.GetTracksForContexts(activeContexts).Where(c => c.ExclusiveToContexts != null)
+            var contextMusicOnly = list.GetTracksForContexts(activeContexts)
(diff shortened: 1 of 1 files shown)


hhyyrylainen (changes requested):
This is slightly different than what I had in mind. Or at least I think so.
To me it looks like from the code that this suppresses context specific tracks after playing them once. However the issue calls for preventing a track from playing twice in a row. Meaning that after any other track has been played all the context specific tracks should again be eligible for playing.

dylanlee082:
That is correct, I originally went down that route, but ran into an issue trying to figure out the prior played song for comparison. The TrackList sets LastPlayedIndex and in other places uses that index on the tracks array to identify if the new song != old song, because the assumption is the list will always be in the same order. However, when the context random is hit a new array is created with new indexes, so I had different songs in the same index.

My change kept growing as I tried to work around that, so I was thinking of other ways to approach the issue and in my opinion this approach is simpler and more consistent with how I typically think of a playlist. If I were adding a new music set to music_tracks.json the only option I would think wouldn't respect PlayedOnce would be EntirelyRandom and the other options would go 1-10 or a random order, but still hit all songs.

However, I do recognize that since the contextual music is supposed to bring a consistent feel to an area of the game that playing Song A, Context Music, Song B, Context Music, Song C, Context Music does make sense. I can take another stab at the original changes. Would it be too expensive to store LastPlayedTrack instead of LastPlayedIndex for comparison, or I could compare against something unique per track like resourcePath or a generated id of some kind?

hhyyrylainen:
I think it would be a fine enough change for the jukebox class to remember what track it last played as I can foresee quite many situations where it is desirable to skip the immediately previously played track no matter what.

ChevyLevi:
@dylanlee082 Hi Dylan, looks like this PR has been stale for a while. I'd like to work on this issue right now, but I'm not sure if you want to continue your work here.

hhyyrylainen:
This was implemented by: https://github.com/Revolutionary-Games/Thrive/pull/6906

PR #6320, opened 2025-08-30
Metabolism & Playstyle Balancing

Deus-Codes:

Tweaking balance across multiple areas that I think are relevant to the various playstyles and metabolisms in game, including the various anaerobic and aerobic respiration methods, pacing of how quickly the atmosphere changes, and predation. 

- Slowed down the speed at which oxygen accumulates, and lowered how much oxygen is build up on average, while pacing carbon dioxide build up as well. This should make it take longer for aerobic respiration to become the clear dominant form of progression.
- Reduced the ATP output of metabolosomes and mitochondria, as it was quite extreme. 
- Slightly nerfed the ATP output of hydrogenase and hydrogenosomes to produce less ATP since they are unaffected by any atmospheric compounds. 
- Made iron respiration burn through iron quicker since it also is unaffected by any atmospheric compounds. This incentivizes players who utilize iron respiration to hang around iron chunks, and makes storage more important to these builds. 
Increased the mass of chloroplasts and thylakoids to make predation harder as a photoautotroph.
Reduced the mass of the nucleus so that predation is a bit easier for eukaryotes.
- Bumped up thermosynthesis ATP production, but increased amount of temperature needed. Should hopefully make it so that hot spots are more important, but rewarding. 
- Radiation should be more efficient, increasing how long you can go on without touching a radioactive zone.

Mitochondria and metabolosomes could arguably be bumped, but I do like the pacing of oxygen accumulation in the atmosphere currently. I also am not 100% how the densities work, so I'm going to investigate the .json file and might bump up density even more. 

Overall, this should slightly incentivize forms of respiration other than just burning glucose while maintaining the power of aerobic respiration. Iron and photo-autotrophy remain powerful but face more constraints altering strategy, thermosynthesis is made powerful but slightly fickle, and radiation is slightly buffed. Reducing the impact of the nucleus on speed should also hopefully make it a bit easier for predation to occur.

Sulfur is notably omitted from this pull request. I want dligr's pull request on sulfur damage to be implemented before tweaking balancing significantly there. 

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/microbe_stage/bio_processes.json
+++ b/simulation_parameters/microbe_stage/bio_processes.json
@@ -1,4 +1,4 @@
-﻿{
+{
   "respiration": {
     "Name": "RESPIRATION",
     "Inputs": {
@@ -6,7 +6,7 @@
       "glucose": 0.10
     },
     "Outputs": {
-      "atp": 110,
+      "atp": 40,
       "carbondioxide": 1
     },
     "IsMetabolismProcess": true
@@ -195,7 +195,7 @@
       "glucose": 0.09
     },
     "Outputs": {
-      "atp": 45,
+      "atp": 20,
       "carbondioxide": 1
     },
     "IsMetabolismProcess": true
@@ -206,7 +206,7 @@
       "glucose": 0.013
     },
     "Outputs": {
-      "atp": 5
+      "atp": 4
     },
     "IsMetabolismProcess": true
   },
@@ -216,7 +216,7 @@
       "glucose": 0.02
     },
     "Outputs": {
-      "atp": 13
+      "atp": 7
     },
     "IsMetabolismProcess": true
   },
@@ -234,7 +234,7 @@
   "iron_chemolithoautotrophy": {
     "Name": "IRON_OXIDATION",
     "Inputs": {
-      "iron": 0.02
+      "iron": 0.2
     },
     "Outputs": {
       "atp": 5.5
@@ -253,16 +253,16 @@
   "bacterial_thermosynthesis": {
     "Name": "THERMOSYNTHESIS",
     "Inputs": {
-      "temperature": 0.2
+      "temperature": 0.4
     },
     "Outputs": {
-      "atp": 4
+      "atp": 6
     }
   },
   "ferrosynthesis": {
     "Name": "IRON_OXIDATION",
     "Inputs": {
-      "iron": 0.033
+      "iron": 0.14
     },
     "Outputs": {
       "atp": 14
@@ -272,7 +272,7 @@
   "radiosynthesis": {
     "name": "RADIOSYNTHESIS",
     "inputs": {
-      "radiation": 0.4
+      "radiation": 0.2
     },
     "outputs": {
       "atp": 35
--- a/simulation_parameters/microbe_stage/organelles.json
+++ b/simulation_parameters/microbe_stage/organelles.json
@@ -1,4 +1,4 @@
-﻿{
+{
   "pilus": {
     "MPCost": 30,
     "InitialComposition": {
@@ -339,7 +339,7 @@
       "UV": 0.03,
       "Oxygen": 0.01
     },
-    "Density": 1200,
+    "Density": 1600,
     "Graphics": {
       "ScenePath": "res://assets/models/organelles/Chromatophore.tscn"
     },
@@ -937,7 +937,7 @@
       "UV": 0.08,
       "Oxygen": 0.02
     },
-    "Density": 1200,
+    "Density": 1800,
     "Graphics": {
       "ScenePath": "res://assets/models/organelles/Chloroplast.tscn"
     },
@@ -979,7 +979,7 @@
         "Capacity": 0.5
       }
     },
-    "Density": 1000,
+    "Density": 800,
     "CorpseChunkGraphics": {
       "ScenePath": "res://assets/models/organelles/Cytoplasm.tscn"
     },
@@ -1050,7 +1050,7 @@
       "Nucleus"
     ],
     "ShouldScale": false,
-    "Density": 1200,
+    "Density": 500,
     "RelativeDensityVolume": 1,
     "Graphics": {
       "ScenePath": "res://assets/models/organelles/Nucleus.tscn"
--- a/src/general/world_effects/PhotosynthesisProductionEffect.cs
+++ b/src/general/world_effects/PhotosynthesisProductionEffect.cs
@@ -29,11 +29,11 @@ public class PhotosynthesisProductionEffect : IWorldEffect
     private void ApplyCompoundsAddition()
     {
         // These affect the final balance
-        var outputModifier = 1.5f;
-        var inputModifier = 1.0f;
+        var outputModifier = 1.2f;
+        var inputModifier = 0.8f;
 
         // This affects how fast the conditions change, but also the final balance somewhat
-        var modifier = 0.00015f;
+        var modifier = 0.000010f;
 
         List<TweakedProcess> microbeProcesses = [];
 


Deus-Codes:
Writing a note to myself to check and make sure that hard difficulty isn't impossible because of the lower ATP generated, and to also bump cytoplasm storage potentially

simulation_parameters/microbe_stage/bio_processes.json
```diff
@@ -253,16 +253,16 @@
   "bacterial_thermosynthesis": {
     "Name": "THERMOSYNTHESIS",
     "Inputs": {
-      "temperature": 0.2
+      "temperature": 0.4
```
hhyyrylainen: I think this nerfs thermosynthesis harder than I buffed it for 0.8.3... and in 0.8.2 all the feedback we got that it was way too hard to use effectively. So while I'm worried about the other major debuffs to ATP, this seems way too much.

The same applies to radiosynthesis which I did a similar 20% buff (but now this would give a 50% debuff so making it even harder than it was in 0.8.2).

simulation_parameters/microbe_stage/organelles.json
```diff
@@ -1050,7 +1050,7 @@
       "Nucleus"
     ],
     "ShouldScale": false,
-    "Density": 1200,
+    "Density": 500,
```
hhyyrylainen: Why is nucleus now a very light organelle?

Deus-Codes: I'm using this as a way to potentially make predation a bit easier for players, since it can take a while for a eukaryote to catch a much more nimble bacterium without a significant number of flagellum

simulation_parameters/microbe_stage/bio_processes.json
Deus-Codes: I can bump it down to 0.3. With an even higher ATP output, an interpretation of thermosynthesis I was going for is a powerful but unreliable source of energy. 

I think there is a reason why we don't see this form of metabolism at all really with how theoretical it remains, and how unreliable heat gradients can be. Considering that, I think it's fair if thermosynthesis is seen as something players can make unique builds with, but not something that is very sustainable in the long-term. With balancing for these more exotic forms of metabolism, like radiosynthesis and thermosynthesis, a better balancing goal could be: accepting them as niche, but giving players looking for a different experience enough of an incentive to create unique playstyles. For radiosynthesis, that playstyle would revolve around dealing with the inherent fickleness of heat gradients for some powerful ATP bonuses. 

I can still bump it down to 0.3 however - it was pretty hard to scale up.

simulation_parameters/microbe_stage/organelles.json
hhyyrylainen: This seems like physically implausible change as the nucleus in real life is quite a concentration of mass in a cell. So making it a significantly less dense part of the cell than anything else seems unrealistic to me.

Edit: finished typing my last sentence.

simulation_parameters/microbe_stage/bio_processes.json
hhyyrylainen: I commented on the community forums a but, but I'll put my thoughts here as well. In 0.8.2 nobody commented that thermosynthesis felt like a good gameplay style. For 0.8.3 I have not heard about anything yet from the players about how it feels. So maybe we should try to reach out? But my hunch is that the buff in 0.8.3 was approximately correct and we *should not* debuff the gameplay style at all before we can confirm how it is perceived now.

Ultimately I think that if thermosynthesis or radiosynthesis is too useless *or* too hard for the average player to use, we might as well delete them entirely as they don't provide any gameplay variety as they cannot be relied on as a gameplay style. Thus debuffing them actually majorly reduces the available gameplay variety.

simulation_parameters/microbe_stage/organelles.json
Deus-Codes: Thing is that at that scale, size and mass doesn't really have a consistently negative relationship with the speed of a microbe. At a bacterial scale, movement can be actually be hampered and slow because at that small size, bacteria have to physically contend with water molecules exacting forces on them - it's as if they have to move through sludge, which is why bacteria under the microscope can look like they are really struggling. I think it's called Brownian motion. 

So even though the nucleus and a eukaryote have more mass and are bigger than a prokaryote, that actually can help them a lot when it comes to movement compared to bacteria. After a certain scale, fluids act like fluids as we are familiar with them and less like sludge, so eukaryotes can more seamlessly swim through water. That's part of the reason why predation is much more frequent in eukaryotes, as they are more readily able to reach potential prey items. 

An entirely realistic depiction in Thrive would probably have tiny bacteria struggling to move in a straight line, and eukaryotes more traditionally "swimming".

Accidental-Explorer: That sounds more like a reconsidering of how mass should affect speed in Thrive microbe stage than just something restricted to the nucleus?

Deus-Codes: I personally don't have the ability to tackle an overhaul of mass and speed's relationship, and don't want to task Hh or another programmer to tackle that challenge if we can approximate it with more simple .json tweaks such as this.

Deus-Codes:
At this point, I feel pretty satisfied with balancing for iron. Consumption is paced to the point that player's don't easily fall into a death loop if they don't spawn near an iron chunk, but still requires the player to be mindful of iron's more vapid nature. This gives iron a more distinct gameplay style that was missing.

What I'm going to investigate next is aerobic respiration and oxygen. Oxygen currently increases at a pretty slow pace, which I personally enjoy but could be too slow for the average player and could make the Microbe Stage too long. That would be beneficial currently regardless since it is our entire game at the moment, but it's still something to be mindful of. 

I'm more convinced that aerobic respiration needs a bump in efficiency because of how slowly oxygen builds now, so that's what I will nail down over this coming week. Currently, metabolosomes are pretty useless, and I'm not sure how mitochondria will perform.

simulation_parameters/microbe_stage/organelles.json
hhyyrylainen: Unless we wanted to make a custom water physics simulation, I don't think that is something that we could really change. We use a normal off-the-shelf physics engine and it has certain assumptions about mass baked in.

hhyyrylainen:
> I'm more convinced that aerobic respiration needs a bump in efficiency because of how slowly oxygen builds now, so that's what I will nail down over this coming week. 

Interesting. I've seen like the opposite on the community Discord lately with many people commenting how it is normal to reach 60% oxygen and kind of complaining about the unrealistic nature of that.

Deus-Codes:
> Interesting. I've seen like the opposite on the community Discord lately with many people commenting how it is normal to reach 60% oxygen and kind of complaining about the unrealistic nature of that.

Oh, I meant a buff relative to the balancing as it is in this PR, not relative to how it is in the base game currently. Because I made both aerobic respiration weaker and oxygen much slower to build up, the effect is that even 20 generations or so in, metabolosomes generate like 2 ATP. I think around that time, they should start being at least as efficient as hydrogenase. 

With the new baseline oxygen and carbon dioxide start to settle it and how long it takes to reach that, getting closer to prior balancing numbers isn't as excessive as it was in an extremely oxygenated world.

hhyyrylainen:
I agree, that sounds way too slow in comparison. As microbe stage is planned to be at most around 20 generations in length, I think any last gameplay effects to kick in should at the latest be at generation 15...

Deus-Codes:
I'm going to have another playtest session where my species doesn't evolve thylakoids/chloroplasts to validate, but I'm pretty satisfied where balancing currently is. In my playthrough, by turn 20, aerobic respiration became the most efficient form of ATP generation in the game. It's different than current game settings, where by turn 15, metabolosomes and mitochondria are absolute rockets - instead, by that time, aerobic respiration becomes a viable tool, and only continues to become more powerful as time goes on. Once again, I want to make sure that oxygen doesn't build catastrophically slow if the player doesn't choose to adapt photosynthesis. 

The result is a slower paced Thrive, where the two major halves of the Microbe Stage - an anaerobic and aerobic world - have more depth. And with changes to metabolism to make them more unique in playstyle, there is a bit more depth in my opinion. I think pace might be thought about more once more of the later stages are developed, but for our current game - entirely dependent on the Microbe Stage - this slower pace is pretty beneficial. 

This pull request isn't done yet: I still want to do some work on thermosynthesis, particularly exploring the output being glucose rather than ATP as Rathalos has suggested. But I do think this results in a more even experience. I will explore photosynthesis balancing next, but I want to focus on that in its own PR, as that might involve slight tweaks to the day and night cycle and a more comprehensive look at that metabolism.

Accidental-Explorer:
> I'm going to have another playtest session where my species doesn't evolve thylakoids/chloroplasts to validate

I also did this today:

- Reached surface after 600 myr (ice shelf)
- First photosynthesizer immediately emerged.
- First oxygen step 0.13
- First epipelagic patch reached 1% oxygen at 900 myr (As I had 7 hydrogenases at this point, this required me to raise oxygen resistance to 25% to avoid penalties)
- After 1300 Myr, oxygen seemed to suddenly jump up a lot. In the most extreme case, from 6,83 to 23.69, blowing past the metabolosome = hydrogenase point.
- At 1500 MYR, global glaciation hit. In some surface patches, metabolosome = hydrogenase, in others, metabolosomes provide double.
- Around this time, in the higest concentration patches, oxygen levels stall between 26 and 27, other surface patches slowly catching up.
- 1800 MYR, completed endosymbiosis for mitochondria. With two metabolosomes I have (and 5 hydrogenases), nucleus is viable in the high oxygen patches, now at about 30%. Up until this point, I was trying to not make edits that reduce total energy collection.
afbeelding
afbeelding

- At 2000 MYR, highest patch: 34.24, lowest surface patch: 12,88 (ice shelf):
afbeelding

- Starting volcanic vent still at 0.29%.

Deus-Codes:
> > I'm going to have another playtest session where my species doesn't evolve thylakoids/chloroplasts to validate
> 
> 
> 
> I also did this today:
> 
> 
> 
> - Reached surface after 600 myr (ice shelf)
> 
> - First photosynthesizer immediately emerged.
> 
> - First oxygen step 0.13
> 
> - First epipelagic patch reached 1% oxygen at 900 myr (As I had 7 hydrogenases at this point, this required me to raise oxygen resistance to 25% to avoid penalties)
> 
> - After 1300 Myr, oxygen seemed to suddenly jump up a lot. In the most extreme case, from 6,83 to 23.69, blowing past the metabolosome = hydrogenase point.
> 
> - At 1500 MYR, global glaciation hit. In some surface patches, metabolosome = hydrogenase, in others, metabolosomes provide double.
> 
> - Around this time, in the higest concentration patches, oxygen levels stall between 26 and 27, other surface patches slowly catching up.
> 
> - 1800 MYR, completed endosymbiosis for mitochondria. With two metabolosomes I have (and 5 hydrogenases), nucleus is viable in the high oxygen patches, now at about 30%. Up until this point, I was trying to not make edits that reduce total energy collection.
> 
> afbeelding
> 
> afbeelding
> 
> 
> 
> - At 2000 MYR, highest patch: 34.24, lowest surface patch: 12,88 (ice shelf):
> 
> afbeelding
> 
> 
> 
> - Starting volcanic vent still at 0.29%.
> 
> 

Wow, what a difference from my experience! I didn't think such a jump in oxygen content was still possible. My immediate impression is having this variety in experiences is beneficial - would you say your experience with these balancing numbers was generally positive?

Accidental-Explorer:
> Wow, what a difference from my experience! I didn't think such a jump in oxygen content was still possible. My immediate impression is having this variety in experiences is beneficial -

And I really have no clue what caused it! But I agree that a variety of different possible outcomes is a beneficial experience, as long as in either case you have enough of a variety of options to go forward.

>  would you say your experience with these balancing numbers was generally positive?

Generally positive yes. And perhaps equally important: it does not feel like a downgrade in any way. Technically it's slowed down by at least a few turns, but it did not feel boring to play through to me. And at least in this case, once I reached the inflection point, there was a rapid change in how I should build my organism.

I really want to see how this plays out in combination with the nucleus update and the auto-evo expansion I am looking at.

Accidental-Explorer:
First impression on the Thermosynthesis change is very positive, I think this can really work! (If you're within the right size frame,) There is still the gameplay pressure of quickly moving for heating up --> cooling down --> heating up again, but that pressure is now the very visibly glucose bar in combination with the already existing temperature bar.

I did notice two things:
- Getting enough Glucose-consuming ATP production to go with your glucose production (and ATP consumption) can feel a bit too limiting.
I would recommend giving Thermosynthase the Glycolysis process, like Chemosynthesizing proteins for example have.

- Glucose production can continue for quite some time after you leave a hot spot/are currently cooling down, not heating up. This gets longer the more storage(?) you have. Since the glucose itself is now the buffer, this is no longer necessary. It makes especially eukaryotic thermosynthesizers a bit too easy in my opinion. I think it also obscures a clear connection of "heating up = glucose production, not heating up = no glucose production" for players.
I would recommend increasing the temperature input of the processes by a lot (essentially trying to consume it close to as fast as it's coming in), so that they don't have any temperature compound to continue working with once you're no longer heating up.
(If this ends up producing too little glucose, that could be bumped up, but only if necessary)
Alternatively, the storage of the temperature compound could be made smaller, but I am not sure if that's a better or easier solution.

Of course there's also the remaining stuff, like adding the glucose node to the miche tree for thermosynthesis, and probably adding CO2 to the process.

hhyyrylainen:
> (If this ends up producing too little glucose, that could be bumped up, but only if necessary)
Alternatively, the storage of the temperature compound could be made smaller, but I am not sure if that's a better or easier solution.

The conversion factor of how much Temperature compound is given per unit of heating up can also be tweaked to make heating up give less "heat" for the bio process to work with.

>  and probably adding CO2 to the process.

This is the main reason why I didn't want to switch the thermosynthesis process to make glucose as this will break the symmetry of CO2 / Oxygen making balancing that then a whole mess again...

Accidental-Explorer:
> The conversion factor of how much Temperature compound is given per unit of heating up can also be tweaked to make heating up give less "heat" for the bio process to work with.

This could also work (or be part of the solution in combination with the other factors).

From a game design perspective, all I am looking for is not producing any glucose while cooling down, but producing enough glucose while heating up to cover the gap.

> This is the main reason why I didn't want to switch the thermosynthesis process to make glucose as this will break the symmetry of CO2 / Oxygen making balancing that then a whole mess again...

You mean a precise symmetry of consumption and production of CO2/O2, or just the rough balance where it is now? Because right now chemosynthesis seems to be consuming CO2 without producing O2? 
So assuming you meant the latter instead of the former: I think that's a good enough reason to not add real CO2 consumption for now, and I promise I will personally handle any questions that come in about it! (Should be easy enough, I think it was mainly me complaining about this before) At least until the time if/when I can add the CO2 consumption and balance the numbers myself to general satisfaction.
I think in this case the hopefully temporary inaccuracy is worth it for an (in my opinion) much better gameplay experience.

hhyyrylainen:
> You mean a precise symmetry of consumption and production of CO2/O2

I don't think we have precise symmetry... so I think I meant the latter that adding anything that produces CO2 as a new thing will need a full rebalance of the Oxygen production system to ensure it isn't thrown entirely out of whack. And also the CO2 generation sources because if the consumption goes up with producers not going up the natural events that try to pump enough CO2 to the world might not be able to keep up and the vents could go to very low CO2 levels which would have a very major impact on overall game balance (especially as vents is the new player biome).

hhyyrylainen:
I've merged in the other changes except the thermosynthesis changes. I expect that will cause some merge conflicts to be reported but those hopefully should be quite simple as the changes on master will almost exactly match what's already here.

Deus-Codes:
> First impression on the Thermosynthesis change is very positive, I think this can really work! (If you're within the right size frame,) There is still the gameplay pressure of quickly moving for heating up --> cooling down --> heating up again, but that pressure is now the very visibly glucose bar in combination with the already existing temperature bar.
> 
> I did notice two things:
> 
> * Getting enough Glucose-consuming ATP production to go with your glucose production (and ATP consumption) can feel a bit too limiting.
>   I would recommend giving Thermosynthase the Glycolysis process, like Chemosynthesizing proteins for example have.
> * Glucose production can continue for quite some time after you leave a hot spot/are currently cooling down, not heating up. This gets longer the more storage(?) you have. Since the glucose itself is now the buffer, this is no longer necessary. It makes especially eukaryotic thermosynthesizers a bit too easy in my opinion. I think it also obscures a clear connection of "heating up = glucose production, not heating up = no glucose production" for players.
>   I would recommend increasing the temperature input of the processes by a lot (essentially trying to consume it close to as fast as it's coming in), so that they don't have any temperature compound to continue working with once you're no longer heating up.
>   (If this ends up producing too little glucose, that could be bumped up, but only if necessary)
>   Alternatively, the storage of the temperature compound could be made smaller, but I am not sure if that's a better or easier solution.
> 
> Of course there's also the remaining stuff, like adding the glucose node to the miche tree for thermosynthesis, and probably adding CO2 to the process.

Bumped up how much temperature input is needed to produce glucose, and I think it has produced benefits.

For glycolysis - I am a bit hesitant to insert that, as I do think that because thermosynthesis is available basically immediately, it should be rather hard to lean into fully. Offering bonus ATP and glucose generation immediately could make thermosynthesis a no-brainer benefit as opposed to a playstyle with constraints you should commit to.

If you do feel strongly that it does need a bit of an ATP generation buff however, I can add it in.

Accidental-Explorer:
> Bumped up how much temperature input is needed to produce glucose, and I think it has produced benefits.

Nice! I will try that out later.
 
> For glycolysis - I am a bit hesitant to insert that, as I do think that because thermosynthesis is available basically immediately, it should be rather hard to lean into fully. Offering bonus ATP and glucose generation immediately could make thermosynthesis a no-brainer benefit as opposed to a playstyle with constraints you should commit to.

Well, thylakoids and chemosynthesizing proteins both can convert their own product into ATP and are available immediately (depending on starting location) and the rusticyanin obviously avoids that problem entirely. That means they can exist as one hex. Only new thermosynthesis cannot, which I guess feels a bit weird.

And I hope that by balancing the numbers, we can reach a point where your whole life as a thermosynthesiser is moving back and forth between a hot and cold spot to produce enough glucose to live. And if you actually want to get somewhere, plot a deliberate course between hot and cold spots to the target.

I think that's not necessarily more "no-brainer" than adding a thylakoids, chemosynthesis or rusticyanin. Especially since thermosynthesis is extremely limited in the patches where it can work.

> If you do feel strongly that it does need a bit of an ATP generation buff however, I can add it in.

I will see how it feels with the new change. I do feel like adding glycolysis is overall a good idea, and if necessary the glucose production itself can be lowered a bit to compensate (so that you still need around the same number of hexes in total dedicated to glucose + ATP).

Deus-Codes:
In my mind: with thylakoids, there atleast is the day-night constraint, and with chemosynthesizing, that is based more on the active pursuit of a cloud, so there's less of an issue in my eyes with ATP generation there.  I also want to tackle some photosynthesis and chemosynthesizing balancing in their own PRs after we finalize thermosynthase balancing, so we can look more at the glucose generation there.

With thermosynthesis, because there is much less of that constraint, I am more amendable to it not having glycolysis. Especially since with the new temperature input amount, players don't have to worry too much about the dance of going in and out of a hotspot; their temperature drops much more quickly, so the time before you can head into the hotspot and get even more glucose is really short. So having the build itself be more challenging to develop I think would be a good solution: it's autotrophic, sure, but you better makes sure you think your progression more thoroughly.

Accidental-Explorer:
Tested the temperature consumption increase now, and I think it works very well! With glucose production stopping as soon as you leave the hot spot (or stay there too long), you can practically manage your movement looking just at the glucose bar. So I think this is fairly clear.

However, it stops working when you get thermoplasts, because they consume *less* of the temperature compound while producing more than double the amount of glucose. I would just set that to 1.6 or even 2.0 now. Since we basically want all the "temperature compound" to be near-instantly consumed, I don't think we want the upgrade to go against that.

By the way, I was able to quite handily swing a nucleus with this:
afbeelding
Personally I think I would prefer the main unique metabolic organelle to be a bit more visually dominant. That's why I would prefer giving thermosynthase Glycolysis and nerfing the glucose output a bit.

> With thermosynthesis, because there is much less of that constraint, I am more amendable to it not having glycolysis. Especially since with the new temperature input amount, players don't have to worry too much about the dance of going in and out of a hotspot; their temperature drops much more quickly, so the time before you can head into the hotspot and get even more glucose is really short. So having the build itself be more challenging to develop I think would be a good solution: it's autotrophic, sure, but you better makes sure you think your progression more thoroughly.

I guess I see the different gameplay of thermosynthesis as "different constraints" rather than "less constraints", but I can see what you mean. And I suppose I would rather use the glucose production amounts as balancing lever rather than break the quite fundamental rule of "every protein part has Glycolysis (except the ones producing ATP directly)"

**I also discovered an unrelated potential balance problem while testing:**
The Hydrogenosome is now in some way inferior to two Hydrogenases. It provide less ATP per hex, even though the process efficiency itself is slightly better.

Other weirdness:
Ferroplast produces more than 2x more ATP than Rusticyanin (as it should), but also consumes *less* iron (it probably should not).

Looking at the Json now: Metabolosome does 0.09 glucose into 30 ATP while mitochondrion does 0.10 into 70. That's a bit much of an efficiency gain, isn't it?

Deus-Codes:
Added glycolysis to thermosynthase. Though I do prefer the prior version, this still performs well and is a step up from the current state, so it'll be solid regardless. If we get feedback that thermosynthase is too easy or something this next release cycle, we can look at removing glycolysis again. 

In regards to material/glucose input being lower for certain eukaryotic parts: I think about two years ago or so, a common complaint was that it was extremely easy to starve, especially as a eukaryote. There were also other complaints related to how inefficient predation was. So in order to create a greater feeling of progression when it comes to unlocking eukaryotic parts, I went with a general "more efficient output" balancing approach which didn't increase how much input was needed between enzymes/proteins and organelles, but still have organelles represent serious step ups. That way, eukaryotes are generally able to sustain higher rates of energy, but not have their food deplete very quickly.

Considering our more recent discussions related to difficulty, that is something I'm willing to discuss and change, but I'd rather do so in a subsequent PR rather than this one. 

With ferroplasts in particular - I was thinking of a progression to iron respiration which slightly increased how long a player can travel without having to stop at a iron chunk. So that input change is a bit more dramatic. 

With hydrogenosomes vs. hydrogenase progression: my approach there was to represent how difficult it can be to scale up anaerobic respiration, even in more complex structures. I think it's more acceptable for hydrogenosomes to not scale up as well, considering it's either meant to be an intermediate or an intentionally low-energy lifestyle in lieu of aerobic respiration; though if players find it much too unrewarding, we can alter that balancing.

Accidental-Explorer:
Tried it out again, I think this still works smoothly, and you especially need a lot more thermosynthases now.

Minimum Eukaryotic cell for comparison:
example eukaryote thermosynthesis
Interestingly, this also makes the transition more complicated, because thermoplasts don't produce ATP.

> Added glycolysis to thermosynthase. Though I do prefer the prior version, this still performs well and is a step up from the current state, so it'll be solid regardless. If we get feedback that thermosynthase is too easy or something this next release cycle, we can look at removing glycolysis again.

Thanks for giving it a try. I do think this version works well and is consistent with the other parts.

For what it's worth, I do think definitively separating ATP production from glucose production parts might also lead to more interesting cell design. But I think in that case we can look at cutting out secondary glycolysis across the board, not just for thermosynthesis.

> With hydrogenosomes vs. hydrogenase progression: my approach there was to represent how difficult it can be to scale up anaerobic respiration, even in more complex structures. I think it's more acceptable for hydrogenosomes to not scale up as well, considering it's either meant to be an intermediate or an intentionally low-energy lifestyle in lieu of aerobic respiration; though if players find it much too unrewarding, we can alter that balancing.

Thanks for explaining this and the other parts. I think we can leave this and the other parts for after the next release and round of player feedback.

I do suggest we pay extra attention to the hydrogenosome at that point, because right now I found myself wanting to *not* replace hydrogenases with hydrogenosomes.

> Thermosynthesis

As for the thermosynthesis, I think the current balance is good enough for the next release. We can also see how people feel about it at that point.

So that just leaves the the auto-evo adjustment. Which I think should just be adding  MinorGlucoseConversionEfficiencyPressure and maintainGlucose miches to the thermosynthesis chain in GenerateMiche.cs?

hhyyrylainen:
I noticed that this was not up to date with master and so showed quite many extra changes. I just updated this and this now shows only the thermosynthesis balance changes.

It was hopefully the same as before, but someone quickly checking that nothing changed with that master merge would be good.

I guess if you guys want to make thermosynthesis into a glucose producer I guess I'm fine with that change if we really want it. Though I would hazard a guess that either the wiki or the tooltips have to be reworded as they probably refer to thermosynthesis directly producing ATP and not glucose so there would be a massive disconnect between what happens in the game and what the text tells the player.

Deus-Codes:
Copying from the Discord:

In the metabolism balancing PR, it seems that oxygen still goes through some wild swings and can still pretty easily reach 35%+ oxygen. So I'm going to lower the balancing again, to generally result in less final oxygen within the world. Though overall levels of oxygen would be reduced, since aerobic respiration is still so powerful, players should still be able to find it be a useful metabolism. And it is much easier to balance for those parts with a smaller range (usually 25% or so in highly oxygenated patches) than it is with a much bigger range (up to 40% in some worlds).

Deus-Codes:
Also copying from Discord:

Turns out that I forgot that the auto-evo tool doesn't perfectly replicate actual gameplay experience, which in this case means that oxygen builds much more rapidly in the tool than it does in a playthrough. So I undid that change. I think we should just live with this current interpretation, and if we still see feedback that oxygen is super-charged in most playthroughs, take further actions. I also did bump up metabolosomes a bit to make it more powerful earlier. Not a big change, just making it more viable

hhyyrylainen:
I think this was done with a separate PR? So this doesn't need to be open anymore.

PR #6321, opened 2025-09-01
Convert Cases of Hardcoded Light Unit

### LINKED ISSUE #4895: Unify logic for displaying compound amounts with units
At the moment the logic for displaying compound amounts is spread over multiple locations - the patch details panel, compound equations in tooltips, etc. - and this sometimes results in inconsistent formatting. For example, light is formatted as '100% lx' in the patch details panel but '100%' in chemical equations.

We should unify this logic and probably put it in the compound class itself.

dylanlee082:

Convert all places of the hardcoded "lx" unit for light in the code to instead use a Unit set in a config file.

NOTE: This does not solve the actual problem of unifying all compound unit logic. I didn't go through and attempt to do that full fix for 2 reason .

1. The way localization is handled made it a little difficult for me to see where all the logic is for how these units work and feel confident I wouldn't completely break it. I'd like to eventually do a different PR for creating a percentage format and unit format function in a helper class somewhere, so that if/when we do eventually consolidate handling of Compound Display it's a little simpler.
2. The CompountAmount.cs class seems to have most of the logic I would have wanted to consolidate, but it seems to be very coupled with the MicrobeEditor display, particularly with the equations in the organelle descriptions. I didn't want to mess with this any more than I had to since I'm still learning my way around the project. 

**Related Issues**

closes #4895 

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/gui_common/CompoundAmount.cs
+++ b/src/gui_common/CompoundAmount.cs
@@ -304,7 +304,12 @@ public partial class CompoundAmount : HBoxContainer
         }
 
         string numberPart;
-        if (!string.IsNullOrEmpty(compoundDefinition!.Unit) && ShowUnit)
+        if (!string.IsNullOrEmpty(compoundDefinition!.Unit) && ShowUnit && UsePercentageDisplay)
+        {
+            numberPart = Localization.Translate("VALUE_WITH_UNIT")
+                .FormatSafe(Localization.Translate("PERCENTAGE_VALUE").FormatSafe(Math.Round(amount * 100, 1)), compoundDefinition.Unit);
+        }
+        else if (!string.IsNullOrEmpty(compoundDefinition!.Unit) && ShowUnit)
         {
             // TODO: implement also the small value display here?
 
--- a/src/microbe_stage/editor/MicrobeEditorReportComponent.cs
+++ b/src/microbe_stage/editor/MicrobeEditorReportComponent.cs
@@ -529,8 +529,8 @@ public partial class MicrobeEditorReportComponent : EditorComponentBase<IEditorR
         }
 
         var percentageFormat = Localization.Translate("PERCENTAGE_VALUE");
-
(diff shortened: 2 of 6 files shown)


hhyyrylainen:
I removed the mention of closing the issue that this only partially addresses.

src/microbe_stage/editor/MicrobeEditorReportComponent.cs
```diff
@@ -529,8 +529,8 @@ private void UpdateReportTabStatistics()
         }
 
         var percentageFormat = Localization.Translate("PERCENTAGE_VALUE");
-
-        sunlightChart.TooltipYAxisFormat = percentageFormat + " lx";
+        var sunlight = SimulationParameters.Instance.GetCompoundDefinition(Compound.Sunlight);
+        sunlightChart.TooltipYAxisFormat = percentageFormat + sunlight.Unit;
```
hhyyrylainen, line 533: This is now missing a space between the % sign and the unit, I think.

src/gui_common/CompoundAmount.cs
```diff
@@ -303,17 +303,32 @@ private void UpdateLabel()
...
+
         string numberPart;
-        if (!string.IsNullOrEmpty(compoundDefinition!.Unit) && ShowUnit)
+        if (!string.IsNullOrEmpty(compoundDefinition!.Unit) && ShowUnit && UsePercentageDisplay)
         {
-            // TODO: implement also the small value display here?
-
-            numberPart = Localization.Translate("VALUE_WITH_UNIT")
-                .FormatSafe(Math.Round(amount, decimals), compoundDefinition.Unit);
+            numberPart = unitFormat.FormatSafe(percentageFormat
```
hhyyrylainen, line 312: I think this should also have the "showEvenSmallValues" support and following the configured `decimals` value. Because right now if both ShowUnit and UsePercentageDisplay are defined, that stops many other features of this class from working, which seems very surprising from an API point of view to me.

PR #6323, opened 2025-09-02
Add new visuals and fixes to environmental tolerances

ender-titan1:

This PR reworks the environmental tolerances tab to be more player friendly and intuitive.

For this PR to be feature complete, the following tasks have to be finished:

- [x] New tolerances UI
- [x] Switch to preferred/tolerance system for pressure 
- [x] Make tooltips work (new ones if necessary) 
- [x] More obvious warning text (panel?)
- [x] Change tutorial text
- [x] Make sure MP costs work as intended
- [x] Make sure auto-evo's MP cost matches the player's
- [x] Code style and structure conforms to the styleguide and programming conventions

Additionally, the following tasks have been completed in the process of developing this:
- [x] Fix a bug that prevented auto-evo from making any changes to environmental tolerances
- [x] Add environmental tolerances to the auto-evo tool's CSV export
- [x] Add health icon to BBCode icons

**Related Issues**

N/A

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/EnvironmentalTolerances.cs
+++ b/src/general/EnvironmentalTolerances.cs
@@ -55,7 +55,7 @@ public class EnvironmentalTolerances
     }
 
     [JsonProperty]
-    public float PressureMaximum => MathF.Min(PressureMinimum + PressureTolerance, Constants.TOLERANCE_PRESSURE_MAX);
+    public float PressureMaximum => PressureMinimum + PressureTolerance;
 
     public static bool operator ==(EnvironmentalTolerances? left, EnvironmentalTolerances? right)
     {
@@ -85,7 +85,7 @@ public class EnvironmentalTolerances
 
     public bool SanityCheckNoThrow()
     {
-        if (PressureMinimum > PressureMaximum)
+        if (PressureMinimum > Math.Min(PressureMaximum, Constants.TOLERANCE_PRESSURE_MAX))
             return false;
 
         if (PressureMaximum < 0)
--- a/src/microbe_stage/MicrobeSpecies.cs
+++ b/src/microbe_stage/MicrobeSpecies.cs
@@ -372,7 +372,7 @@ public class MicrobeSpecies : Species, ICellDefinition
             Localization.Translate("TOLERANCE_DETAIL_TEXT").FormatSafe(Tolerances.PreferredTemperature,
                 Tolerances.TemperatureTolerance,
                 Tolerances.PressureMinimum,
-                Tolerances.PressureMaximum,
+                Math.Min(Tolerances.PressureMaximum, Constants.TOLERANCE_PRESSURE_MAX),
                 Tolerances.PressureTolerance,
                 Math.Round(Tolerances.OxygenResistance * 100, 2),
                 Math.Round(Tolerances.UVResistance * 100, 2));
--- a/src/microbe_stage/editor/TolerancesEditorSubComponent.cs
+++ b/src/microbe_stage/editor/TolerancesEditorSubComponent.cs
@@ -768,7 +768,7 @@ public partial class TolerancesEditorSubComponent : EditorComponentBase<ICellEdi
             pressureMaxLabel.LabelSettings = originalPressureFont;
             pressureRangeDisplay.SetColorsAndRedraw(optimalDisplayBadColor);
         }
-        else if (patchPressure > CurrentTolerances.PressureMaximum)
+        else if (patchPressure > Math.Min(CurrentTolerances.PressureMaximum, Constants.TOLERANCE_PRESSURE_MAX))
         {
             pressureMaxLabel.LabelSettings = badValueFontTiny;
             pressureMinLabel.LabelSettings = originalPressureFont;
--- a/src/gui_common/fonts/Body-Regular-AlmostTiny-Green.tres
+++ b/src/gui_common/fonts/Body-Regular-AlmostTiny-Green.tres
@@ -1,0 +1,8 @@
+[gd_resource type="LabelSettings" load_steps=2 format=3 uid="uid://bkbj6e1c0x8ac"]
+
+[ext_resource type="FontFile" uid="uid://s8m5c8resd50" path="res://assets/fonts/Lato-Regular.ttf" id="1_h1s5k"]
+
+[resource]
+font = ExtResource("1_h1s5k")
+font_size = 13
+font_color = Color(0, 1, 0.117647, 1)
--- a/src/microbe_stage/editor/TolerancesEditorSubComponent.tscn
+++ b/src/microbe_stage/editor/TolerancesEditorSubComponent.tscn
@@ -1,16 +1,17 @@
-[gd_scene load_steps=21 format=3 uid="uid://5kamud4sukbd"]
+[gd_scene load_steps=22 format=3 uid="uid://5kamud4sukbd"]
 
 [ext_resource type="Theme" uid="uid://b4cx0o110g4b6" path="res://src/gui_common/thrive_theme.tres" id="1_16mn0"]
 [ext_resource type="Script" uid="uid://dxl2pjshg67oa" path="res://src/microbe_stage/editor/TolerancesEditorSubComponent.cs" id="2_kqh2k"]
-[ext_resource type="LabelSettings" uid="uid://bl7dig4lq5ko" path="res://src/gui_common/fonts/Body-Regular-Small-Red.tres" id="3_q5r3n"]
+[ext_resource type="LabelSettings" uid="uid://58teykjxnrdt" path="res://src/gui_common/fonts/Body-Regular-AlmostTiny-Red.tres" id="3_n6u17"]
 [ext_resource type="Texture2D" uid="uid://bsg6xcrlspkdt" path="res://assets/textures/gui/bevel/hSeparatorLeftSide.png" id="3_ri2xt"]
-[ext_resource type="LabelSettings" uid="uid://cw40l6f6fo8ml" path="res://src/gui_common/fonts/Body-Regular-Small-Green.tres" id="4_15qmx"]
-[ext_resource type="LabelSettings" uid="uid://brgbuduvjby6a" path="res://src/gui_common/fonts/Title-SemiBold-Smaller.tres" id="5_6mela"]
+[ext_resource type="LabelSettings" uid="uid://bkbj6e1c0x8ac" path="res://src/gui_common/fonts/Body-Regular-AlmostTiny-Green.tres" id="4_evsrs"]
 [ext_resource type="LabelSettings" uid="uid://brnlfjye3ufic" path="res://src/gui_common/fonts/Body-Regular-Tiny-Red.tres" id="5_evsrs"]
 [ext_resource type="LabelSettings" uid="uid://cp2wdrb71h2qo" path="res://src/gui_common/fonts/Body-Regular-Tiny-Green.tres" id="6_jstjb"]
 [ext_resource type="Texture2D" uid="uid://ut73drrfbmk0" path="res://assets/textures/gui/bevel/Temperature.svg" id="6_ki8q3"]
+[ext_resource type="LabelSettings" uid="uid://csw4xjfy4y0dk" path="res://src/gui_common/fonts/Title-SemiBold-Small.tres" id="8_3r1ho"]
 [ext_resource type="LabelSettings" uid="uid://dvqx73nhtr0y2" path="res://src/gui_common/fonts/Body-Regular-Small.tres" id="8_ktuy7"]
 [ext_resource type="PackedScene" uid="uid://duf1lf5lp4v5p" path="res://src/microbe_stage/editor/ToleranceOptimalDisplay.tscn" id="9_xmghy"]
+[ext_resource type="LabelSettings" uid="uid://uxdd8vb4xedn" path="res://src/gui_common/fonts/Body-Regular-AlmostTiny.tres" id="10_jstjb"]
 [ext_resource type="Texture2D" uid="uid://d1tuehjbire76" path="res://assets/textures/gui/bevel/Pressure.svg" id="10_xmghy"]
 [ext_resource type="LabelSettings" uid="uid://c07qrffjvqfw" path="res://src/gui_common/fonts/Body-Regular-Tiny.tres" id="11_5y3m2"]
 [ext_resource type="Texture2D" uid="uid://djaidl3g28ibi" path="res://assets/textures/gui/bevel/Oxygen.svg" id="12_3r1ho"]
@@ -37,38 +38,38 @@ grow_horizontal = 2
 grow_vertical = 2
 theme = ExtResource("1_16mn0")
 script = ExtResource("2_kqh2k")
-temperatureToolTipContainer = NodePath("Temperature/ModifierAndTotal/HBoxContainer/Total")
-pressureToolTipContainer = NodePath("Pressure/ModifierAndTotal/HBoxContainer/Total")
-oxygenResistanceToolTipContainer = NodePath("Oxygen/ModifierAndTotal/HBoxContainer/Total")
-uvResistanceToolTipContainer = NodePath("UV/ModifierAndTotal/HBoxContainer/Total")
+temperatureToolTipContainer = NodePath("Temperature/HBoxContainer2/Total")
+pressureToolTipContainer = NodePath("Pressure/HBoxContainer2/Total")
+oxygenResistanceToolTipContainer = NodePath("Oxygen/HBoxContainer2/Total")
+uvResistanceToolTipContainer = NodePath("UV/HBoxContainer2/Total")
 temperatureSlider = NodePath("Temperature/RangeDisplayContainer/VBoxContainer/TemperatureSlider")
 temperatureToleranceRangeSlider = NodePath("Temperature/HBoxContainer3/TemperatureRange")
 pressureSlider = NodePath("Pressure/RangeDisplayContainer/VBoxContainer/PressureSlider")
 pressureToleranceRangeSlider = NodePath("Pressure/HBoxContainer4/PressureRange")
 oxygenResistanceSlider = NodePath("Oxygen/RangeDisplayContainer/VBoxContainer/OxygenResistanceSlider")
 uvResistanceSlider = NodePath("UV/RangeDisplayContainer/VBoxContainer/UVResistanceSlider")
-temperatureMinLabel = NodePath("Temperature/ModifierAndTotal/HBoxContainer/Total/TemperatureMinimum")
-temperatureMaxLabel = NodePath("Temperature/ModifierAndTotal/HBoxContainer/Total/TemperatureMaximum")
+temperatureMinLabel = NodePath("Temperature/HBoxContainer2/Total/TemperatureMinimum")
+temperatureMaxLabel = NodePath("Temperature/HBoxContainer2/Total/TemperatureMaximum")
 temperatureToleranceLabel = NodePath("Temperature/HBoxContainer3/MarginContainer/TemperatureRangeValue")
-temperatureModifierLabel = NodePath("Temperature/ModifierAndTotal/HBoxContainer/Modifier/ModifierValue")
+temperatureModifierLabel = NodePath("Temperature/HBoxContainer2/Modifier/ModifierValue")
 temperatureRangeDisplay = NodePath("Temperature/RangeDisplayContainer/VBoxContainer/TemperatureRangeDisplay")
-pressureMinLabel = NodePath("Pressure/ModifierAndTotal/HBoxContainer/Total/PressureMinimum")
-pressureMaxLabel = NodePath("Pressure/ModifierAndTotal/HBoxContainer/Total/PressureMaximum")
+pressureMinLabel = NodePath("Pressure/HBoxContainer2/Total/PressureMinimum")
+pressureMaxLabel = NodePath("Pressure/HBoxContainer2/Total/PressureMaximum")
 pressureToleranceLabel = NodePath("Pressure/HBoxContainer4/MarginContainer/PressureRangeValue")
-pressureModifierLabel = NodePath("Pressure/ModifierAndTotal/HBoxContainer/Modifier/ModifierValue")
+pressureModifierLabel = NodePath("Pressure/HBoxContainer2/Modifier/ModifierValue")
 pressureRangeDisplay = NodePath("Pressure/RangeDisplayContainer/VBoxContainer/PressureRangeDisplay")
-oxygenResistanceModifierLabel = NodePath("Oxygen/ModifierAndTotal/HBoxContainer/Modifier/ModifierValue")
-oxygenResistanceTotalLabel = NodePath("Oxygen/ModifierAndTotal/HBoxContainer/Total/TotalValue")
+oxygenResistanceModifierLabel = NodePath("Oxygen/HBoxContainer2/Modifier/ModifierValue")
+oxygenResistanceTotalLabel = NodePath("Oxygen/HBoxContainer2/Total/TotalValue")
 oxygenResistanceRangeDisplay = NodePath("Oxygen/RangeDisplayContainer/VBoxContainer/OxygenRangeDisplay")
-uvResistanceModifierLabel = NodePath("UV/ModifierAndTotal/HBoxContainer/Modifier/ModiferValue")
-uvResistanceTotalLabel = NodePath("UV/ModifierAndTotal/HBoxContainer/Total/TotalValue")
+uvResistanceModifierLabel = NodePath("UV/HBoxContainer2/Modifier/ModiferValue")
+uvResistanceTotalLabel = NodePath("UV/HBoxContainer2/Total/TotalValue")
 uvResistanceRangeDisplay = NodePath("UV/RangeDisplayContainer/VBoxContainer/UVRangeDisplay")
-badValueFont = ExtResource("3_q5r3n")
-perfectValueFont = ExtResource("4_15qmx")
+badValueFont = ExtResource("3_n6u17")
+perfectValueFont = ExtResource("4_evsrs")
 badValueFontTiny = ExtResource("5_evsrs")
 perfectValueFontTiny = ExtResource("6_jstjb")
-modifierBadFont = ExtResource("3_q5r3n")
-modifierGoodFont = ExtResource("4_15qmx")
+modifierBadFont = ExtResource("3_n6u17")
+modifierGoodFont = ExtResource("4_evsrs")
 optimalDisplayBadColor = Color(1, 0, 0, 1)
 optimalDisplayGoodColor = Color(0, 1, 0, 1)
 
@@ -80,7 +81,7 @@ mouse_filter = 0
 layout_mode = 2
 
 [node name="TextureRect" type="TextureRect" parent="Temperature/HBoxContainer"]
-custom_minimum_size = Vector2(20, 20)
+custom_minimum_size = Vector2(18, 18)
 layout_mode = 2
 size_flags_horizontal = 4
 size_flags_vertical = 4
@@ -91,7 +92,7 @@ stretch_mode = 4
 [node name="TemperatureTitle" type="Label" parent="Temperature/HBoxContainer"]
 layout_mode = 2
 text = "TEMPERATURE"
-label_settings = ExtResource("5_6mela")
+label_settings = ExtResource("8_3r1ho")
 
 [node name="HSeparator" type="HSeparator" parent="Temperature/HBoxContainer"]
 layout_mode = 2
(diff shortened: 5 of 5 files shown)


src/general/EnvironmentalTolerances.cs
```diff
@@ -16,23 +17,15 @@ public class EnvironmentalTolerances
...
-    ///   <para>
-    ///     The difference between the defaults may not be over Constants.TOLERANCE_PRESSURE_RANGE_MAX, otherwise the
-    ///     GUI will break when this data is fed in.
-    ///   </para>
-    /// </remarks>
-    public float PressureMinimum = 71325;
-
-    public float PressureMaximum = 301325;
+    public float PreferredPressure;
+    public float PressureTolerance;
```
hhyyrylainen, line 26: I don't think this is a good data model change whatsoever because most species, humans included, have very lopsided pressure tolerance range. For humans the range is approximately something like 0.06 atm to close to 50 atm. So the pressure range is extremely lopsided to higher pressure survivability. And I don't think the data model of having a fixed size tolerance around a preferred pressure can model that correctly. So I ask that this data model change is reverted.

ender-titan1: I believe I've introduced a change that fixes the issue with this implementation. Now the pressure range is always lopsided to higher pressures. Now the slider selects the minimum pressure, and the tolerance is applied only on the positive.

ender-titan1:
This PR is now feature complete and ready for testing and review.

src/general/EnvironmentalTolerances.cs
```diff
@@ -8,31 +9,42 @@ public class EnvironmentalTolerances
     /// <summary>
     ///   Temperature (in C) that this species likes to be in
     /// </summary>
-    public float PreferredTemperature = 15;
+    public float PreferredTemperature;
 
     /// <summary>
     ///   How wide a temperature range this species can stay in effectively. The range of temperatures is
     ///   <c>PreferredTemperature - TemperatureTolerance</c> to <c>PreferredTemperature + TemperatureTolerance</c>
```
Patryk26g, line 16: Now it is only PreferredTemperature + TemperatureTolerance, right?

Patryk26g (approved):

Patryk26g:
I have tested the functionality and did a code review, for me it looks fine :)

src/general/EnvironmentalTolerances.cs
ender-titan1: Temperature has not been changed, it is still Preferred +/- Tolerance, this is indicated in the UI by the line connecting the range to the grabber being drawn from the center of the range as opposed to pressure where it is drawn from the minimum.

src/auto-evo/AutoEvoExploringTool.Export.cs
```diff
@@ -56,12 +56,16 @@ private void ExportCurrentWorldSpeciesHistory(string basePath)
...
 
+        // Microbe specific headers
+
         header.AddRange(new[]
         {
             "Membrane type", "Membrane rigidity", "Base speed", "Base rotation speed", "Storage capacity", "Bacteria",
             "Organelle count",
         });
 
+        header.AddRange(typeof(EnvironmentalTolerances).GetFields().Select(f => f.Name));
```
hhyyrylainen, line 67: I think this should explicitly list out the field names because the code on line 110 is manually listing the names so this can get out of sync accidentally.

src/general/EnvironmentalTolerances.cs
```diff
@@ -8,31 +9,42 @@ public class EnvironmentalTolerances
...
-
-    public float PressureMaximum = 301325;
+    [JsonIgnore]
+    public float PressureTolerance;
 
     public float UVResistance;
     public float OxygenResistance;
 
+    public EnvironmentalTolerances()
+        : this(0, 0, 0, 0, 0, 0)
```
hhyyrylainen, line 32: Why are the default tolerances now basically unusable values?

src/general/EnvironmentalTolerances.cs
```diff
@@ -8,31 +9,42 @@ public class EnvironmentalTolerances
     /// <summary>
     ///   Temperature (in C) that this species likes to be in
     /// </summary>
-    public float PreferredTemperature = 15;
+    public float PreferredTemperature;
```
hhyyrylainen, line 12: So why was this default value removed?

src/microbe_stage/editor/CellEditorComponent.GUI.cs
```diff
@@ -791,15 +791,14 @@ void AddToleranceWarning(string text)
...
-                        AutowrapMode = TextServer.AutowrapMode.WordSmart,
-                        CustomMinimumSize = new Vector2(150, 0),
-                        LabelSettings = toleranceWarningsFont,
-                    };
+                    var warning = customRichTextLabelScene.Instantiate<CustomRichTextLabel>();
 
-                    warning.Text = text;
+                    warning.HorizontalAlignment = HorizontalAlignment.Center;
+                    warning.AutowrapMode = TextServer.AutowrapMode.WordSmart;
+                    warning.CustomMinimumSize = new Vector2(150, minimumY);
```
hhyyrylainen, line 798: Why is the Y value explicitly done like this? Is this to make sure the Y size is not 0? That *should* work fine when setting the right "fit to content" option on the custom rich text label, which is what all other rich text labels need to do.

src/microbe_stage/editor/CellEditorComponent.tscn
```diff
@@ -742,6 +742,7 @@ button_group = SubResource("ButtonGroup_xodoa")
 text = "ENVIRONMENT_TOLERANCE"
 
 [node name="Structure" type="PanelContainer" parent="LeftPanel/MainPanel/VBoxContainer/MarginContainer/EditingPanel"]
+visible = false
```
hhyyrylainen, line 745: I think you accidentally left the non-default tab visible in this scene...

src/microbe_stage/editor/tooltips/StatModifierToolTip.cs
```diff
@@ -39,9 +47,9 @@ public partial class StatModifierToolTip : Control, ICustomToolTip
 
     private bool formatAsPercentage;
     private string? valueSuffix;
+    private string? valuePrefix;
 
     [Export]
-    [ExportCategory("Configuration")]
```
hhyyrylainen, line 44: Why was this category removed?

src/microbe_stage/editor/tooltips/StatModifierToolTip.cs
```diff
@@ -149,7 +164,7 @@ public override void _Ready()
 
     /// <summary>
     ///   Shows a breakdown of values divided based on organelle types. Clears any old data that shouldn't be shown
-    ///   any more.
+    ///   anymore.
```
hhyyrylainen, line 167: I don't think this should be changed as my grammar checker in my IDE doesn't like "anymore" as it is not British English, and that's kind of the variant we are going for...

src/microbe_stage/editor/tooltips/StatModifierToolTip.cs
```diff
@@ -243,9 +265,6 @@ private void UpdateValueDisplay()
             text = value.ToString(CultureInfo.CurrentCulture);
         }
 
-        if (!string.IsNullOrEmpty(valueSuffix))
-            text = $"{text} {valueSuffix}";
-
-        valueLabel.Text = text;
+        valueLabel.Text = $"{valuePrefix ?? string.Empty} {text} {valueSuffix ?? string.Empty}";
```
hhyyrylainen, line 268: This now *always* adds preceding and trailing space. I don't think this is a good change, this should still use the approach where the spaces between parts are added if the suffix / prefix is not empty.

src/microbe_stage/editor/tooltips/StatModifierToolTip.cs
```diff
@@ -208,6 +223,13 @@ public void DisplayOrganelleBreakdown(Dictionary<OrganelleDefinition, float> ite
                         pair.Value);
             }
 
+            value.LabelSettings = valueToShow switch
+            {
+                > 0 => breakdownGoodFont,
```
hhyyrylainen, line 228: I think it would be a good idea for the fonts to fallback to the "breakdownFont" if these special case fonts are null.

src/microbe_stage/Patch.cs
```diff
@@ -519,45 +519,14 @@ public EnvironmentalTolerances GenerateTolerancesForMicrobe(IReadOnlyList<Organe
         // The multipliers cause things to be slightly higher than required so that there's no "rounding" errors with
         // some tolerances not being exactly right
         var pressure = Biome.Pressure - (organelleEffects.PressureMaximum - organelleEffects.PressureMinimum) * 0.51f;
```
hhyyrylainen, line 521: Isn't this logic now incorrect? As the minimum is always set at *exactly* the patch's pressure, meaning that this now must be *exactly* the patch's pressure whereas this is actually undershooting due to the `0.51` multiply.

In general I think this code should try to set the pressure minimum *slightly below* the patch's pressure so that the float inaccuracy is unlikely to cause a problem.

This is kind of the primary reason why I don't fully like the change to have the "minimum" pressure change to be the "preferred" value, as it causes math issues like this.

Also this math should have a `Math.Max` call to ensure minimum pressure does not go below 0.

src/microbe_stage/MicrobeSpecies.cs
```diff
@@ -372,7 +372,8 @@ public override string GetDetailString()
             Localization.Translate("TOLERANCE_DETAIL_TEXT").FormatSafe(Tolerances.PreferredTemperature,
                 Tolerances.TemperatureTolerance,
                 Tolerances.PressureMinimum,
-                Tolerances.PressureMaximum,
+                Math.Min(Tolerances.PressureMaximum, Constants.TOLERANCE_PRESSURE_MAX),
```
hhyyrylainen, line 375: I'm trying to wrap my head around this line, is this correct as to me this seems wrong?

src/gui_common/tooltip/ToolTipManager.tscn
```diff
@@ -50,23 +50,20 @@ offset_right = 40.0
 offset_bottom = 40.0
 mouse_filter = 2
 
-[node name="menuButton" parent="GroupHolder/default" node_paths=PackedStringArray("descriptionLabel") instance=ExtResource("1")]
+[node name="menuButton" parent="GroupHolder/default" instance=ExtResource("1")]
```
hhyyrylainen, line 53: Is this fine? To me it looks like this PR *might* be breaking a ton of unrelated tooltips...

src/gui_common/fonts/Body-Regular-Tiny-Green.tres
```diff
@@ -0,0 +1,8 @@
+[gd_resource type="LabelSettings" load_steps=2 format=3 uid="uid://cp2wdrb71h2qo"]
+
+[ext_resource type="FontFile" uid="uid://s8m5c8resd50" path="res://assets/fonts/Lato-Regular.ttf" id="1_jpyfw"]
+
+[resource]
+font = ExtResource("1_jpyfw")
+font_size = 12
+font_color = Color(0.2, 0.501961, 0, 1)
```
hhyyrylainen, line 8: Why is this green colour different? I've ensured that all of the green and red font variants use the same exact colours as the previous ones, so I'd really like you to double check that you are using existing colour values in these. If a special green variant is needed, the font name should have not just "green" but "LightGreen" or "DarkGreen" depending on if it is lighter / darker than the normal one.

src/general/EnvironmentalTolerances.cs
```diff
@@ -42,6 +54,9 @@ public enum ToleranceChangedStats
         OxygenResistance = 8,
     }
 
+    [JsonProperty]
```
hhyyrylainen, line 57: Computed properties should *never* be saved. As they can just be recalculated from the other properties whenever needed.

locale/en.po
```diff
@@ -3895,6 +3886,9 @@ msgstr "Models"
 msgid "MODE_CAN_BE_CHANGED_IN_OPTIONS"
 msgstr "2D movement mode can be later changed in the options menu at any time"
 
+msgid "MODIFIER"
```
hhyyrylainen, line 3889: I think these are a bit too general key names. I think this PR should use `STAT_MODIFIER` or `VALUE_MODIFIER` and `VALUE_OFFSET`.

locale/en.po
```diff
@@ -6175,13 +6172,14 @@ msgstr ""
 "  Range: ±{1}°C\n"
 "[b]Pressure[/b]\n"
 "  Range: {2}-{3} Pa\n"
+"  Tolerance: {4}\n"
```
hhyyrylainen, line 6175: What's the reasoning on adding this "tolerance" value? It can already be seen in the range above as it shows the min and max so the tolerance is just subtracting those values from each other?

Alternatively the text could be "Minimum: {2}" and "Tolerance: {3}" to just still show 2 numbers instead of 3 for the pressure.

locale/en.po
```diff
@@ -6175,13 +6172,14 @@ msgstr ""
...
 "[b]Oxygen Resistance[/b]\n"
-"  {4}%\n"
+"  {5}%\n"
 "[b]UV Resistance[/b]\n"
-"  {5}%"
+"  {6}%"
 
 msgid "TOLERANCE_FROM_ORGANELLES_TOOLTIP"
-msgstr "Tolerances are affected by some organelles, both positively and negatively. The effective value is shown here after the base value is added together with the change from organelles. Some organelles synergize with certain playstyles but harm others."
+msgstr "Tolerances are affected by some organelles, both positively and negatively. The total value is shown here after the base value is added together with the change from organelles. Some organelles synergize with certain playstyles but harm others."
```
hhyyrylainen, line 6182: To me this doesn't really clarify anything, and might even make the understandability worse... I tried to start a discussion on the dev discord about this, so let's see if anyone else agrees with me.

locale/en.po
```diff
@@ -6192,18 +6190,21 @@ msgstr "Tools"
...
 msgstr "Hand Axe"
 
-msgid "TOO_LARGE_PRESSURE_RANGE"
-msgstr "Pressure range too wide"
-
 msgid "TOO_MANY_RECENT_VERSIONS_TO_SHOW"
 msgstr "Not showing all recent versions as there are too many"
 
+msgid "TOTAL"
+msgstr "Total"
```
hhyyrylainen, line 6197: I think the key for clarity here should be "TOTAL_VALUE_SHORT"

locale/en.po
```diff
@@ -6479,15 +6480,19 @@ msgstr ""
 
 msgid "TUTORIAL_MICROBE_EDITOR_TOLERANCES_TAB"
 msgstr ""
-"The tolerances editor allows changing the [b]environmental conditions your species is adapted to[/b]. These include [thrive:compound type=\"temperature\"][/thrive:compound] and [b]pressure[/b] [thrive:icon]Pressure[/thrive:icon] of the patch you reside in.\n"
+"Your organism can only live in environmental conditions it is [b]adapted[/b] to.\n"
```
hhyyrylainen, line 6483:
```suggestion
"Your species can only live in environmental conditions it is [b]adapted[/b] to.\n"
```

I think that will be much more uniform wording with other text in the game.

locale/en.po
```diff
@@ -6479,15 +6480,19 @@ msgstr ""
...
 "\n"
-"To become [b]perfectly adapted[/b] and gain bonuses instead of negative effects, the flexibility ranges need to be reduced to minimal values while still staying adapted to the patch. The numbers will then turn green and grant positive effects.\n"
+"Here you can also increase your resistance to the harmful effects of [thrive:compound type=\"oxygen\"][/thrive:compound] and [thrive:compound type=\"sunlight\"][/thrive:compound].\n"
 "\n"
-"The trade-off with perfect adaptation is not being suitable for a wide range of patches. Planning before moving patches by increasing the tolerance ranges can be a good strategy if you have [b]some remaining MP[/b] [thrive:icon]MP[/thrive:icon] to use up before exiting the editor.\n"
+"[color=#ff4d4d]If the current conditions are out of range of your tolerances, you will suffer penalties[/color] to your [b]health[/b] [thrive:icon]HealthIcon[/thrive:icon], [b]osmoregulation[/b] [thrive:icon]OsmoIcon[/thrive:icon] and [b]bioprocess speed[/b].\n"
```
hhyyrylainen, line 6489: I think referring to the specific stats is much too specific as we may tweak the stats in the future and then the tutorial is outdated. This is why I don't want too much "specific" info in the tutorial.

locale/en.po
```diff
@@ -6479,15 +6480,19 @@ msgstr ""
...
 "\n"
-"Try changing one of the environmental tolerances to close this tutorial."
+"[color=#338000]Likewise, if your tolerance [b]range[/b] is small enough, those same attributes get buffed.[/color] The tradeoff here is that with a small range you are limited to a smaller number of patches.\n"
+"\n"
+"Beware that [u]organelles in your cell can affect your [b]resistances[/b] and [b]flexibility[/b][/u].\n"
```
hhyyrylainen, line 6493: Saying "resistances" here sounds wrong to me, I think it should say "tolerances".

locale/en.po
```diff
@@ -6479,15 +6480,19 @@ msgstr ""
...
+"Here you can also increase your resistance to the harmful effects of [thrive:compound type=\"oxygen\"][/thrive:compound] and [thrive:compound type=\"sunlight\"][/thrive:compound].\n"
 "\n"
-"The trade-off with perfect adaptation is not being suitable for a wide range of patches. Planning before moving patches by increasing the tolerance ranges can be a good strategy if you have [b]some remaining MP[/b] [thrive:icon]MP[/thrive:icon] to use up before exiting the editor.\n"
+"[color=#ff4d4d]If the current conditions are out of range of your tolerances, you will suffer penalties[/color] to your [b]health[/b] [thrive:icon]HealthIcon[/thrive:icon], [b]osmoregulation[/b] [thrive:icon]OsmoIcon[/thrive:icon] and [b]bioprocess speed[/b].\n"
 "\n"
-"Try changing one of the environmental tolerances to close this tutorial."
+"[color=#338000]Likewise, if your tolerance [b]range[/b] is small enough, those same attributes get buffed.[/color] The tradeoff here is that with a small range you are limited to a smaller number of patches.\n"
```
hhyyrylainen, line 6491: And also I think this is too specific because perfect tolerance of a single statistic doesn't buff everything. And also I think the word "perfect adaptation" should be used here as I think it is a pretty important term and I'd like to still be able to talk to players about perfect adaptation on the fan discord when some future player shows up who has played the new version of the tutorial.

src/auto-evo/mutation_strategy/ModifyEnvironmentalTolerance.cs
```diff
@@ -101,7 +101,6 @@ public class ModifyEnvironmentalTolerance : IMutationStrategy<MicrobeSpecies>
             {
                 var maxChange = mp / Constants.TOLERANCE_CHANGE_MP_PER_PRESSURE;
 
-                // Calculate in doubles as the pressure stuff needs many decimals
```
hhyyrylainen, line 104: Why was this comment removed?

src/auto-evo/mutation_strategy/ModifyEnvironmentalTolerance.cs
```diff
@@ -82,9 +82,9 @@ public class ModifyEnvironmentalTolerance : IMutationStrategy<MicrobeSpecies>
                 }
 #endif
 
-                var change = Math.Max(score.PerfectOxygenAdjustment, maxChange);
+                var change = Math.Max(score.PressureRangeSizeAdjustment, maxChange);
 
-                newTolerances.TemperatureTolerance -= change;
+                newTolerances.TemperatureTolerance += change;
```
hhyyrylainen, line 87: I assume this is a bug fix to temperature?

src/auto-evo/mutation_strategy/ModifyEnvironmentalTolerance.cs
```diff
@@ -113,45 +112,27 @@ public class ModifyEnvironmentalTolerance : IMutationStrategy<MicrobeSpecies>
                     change = Math.Min(score.PerfectPressureAdjustment, maxChange);
                 }
 
-                // These are adjusted in the same direction to keep the same range as before
-                newTolerances.PressureMinimum = Math.Max(newTolerances.PressureMinimum + (float)change, 0);
-                newTolerances.PressureMaximum = Math.Max(newTolerances.PressureMaximum + (float)change, 0);
+                newTolerances.PressureMinimum += (float)change;
```
hhyyrylainen, line 115: I think this should use a clamp to ensure the minimum doesn't go below 0 (and doesn't go above the max pressure).

src/auto-evo/mutation_strategy/ModifyEnvironmentalTolerance.cs
```diff
@@ -113,45 +112,27 @@ public class ModifyEnvironmentalTolerance : IMutationStrategy<MicrobeSpecies>
...
-                // Recalculate change potential for the other part of the calculation
-                mp -= (float)(maxChange * Constants.TOLERANCE_CHANGE_MP_PER_PRESSURE_TOLERANCE);
-                changePotential = mp / Constants.TOLERANCE_CHANGE_MP_PER_PRESSURE_TOLERANCE;
-
-                newTolerances.PressureMaximum = (float)(newTolerances.PressureMaximum - maxChange);
-
-                var minChange = Math.Min(changePotential, halfAdjustment);
-
-                newTolerances.PressureMinimum = (float)(newTolerances.PressureMinimum + minChange);
+                newTolerances.PressureTolerance += (float)change;
```
hhyyrylainen, line 133: I think this should have safety code to ensure that the pressure tolerance does not go below 0 as I assume your code changes are not setup to handle if something manages to get a negative pressure tolerance?

src/microbe_stage/editor/ToleranceRangeDisplay.cs
```diff
@@ -0,0 +1,199 @@
...
+
+/// <summary>
+///   Displays a tolerance with an optimal value in an intuitive way
+/// </summary>
+public partial class ToleranceRangeDisplay : HSlider
+{
+    private const float LINE_WIDTH = Constants.TOLERANCE_DISPLAY_MARKER_WIDTH;
+
+    private readonly Color mainColor = Color.FromHtml("#11FFD5");
+    private readonly Color mainColorTranslucent = Color.FromHtml("#11FFD5") with { A = 0.5f };
```
hhyyrylainen, line 13: `FromHtml` should not be used for color values that are know. Instead please use the direct float values in the Color constructor.

src/microbe_stage/editor/ToleranceRangeDisplay.cs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /// <summary>
+    ///   Sets the positions of the upper and lower bounds based on a middle value.
+    ///   All values have to be between this slider's max and min.
+    /// </summary>
+    /// <param name="preferred">The middle of the range</param>
+    /// <param name="flexibilityPositive">The offset of the upper bound from the middle</param>
+    /// <param name="flexibilityNegative">The offset of the lower bound from the middle</param>
+    public void SetBoundPositions(float preferred, float flexibilityPositive, float? flexibilityNegative = null)
```
hhyyrylainen, line 141: Nullable value types (`float?`) cause boxing, which allocates memory. So it should be avoided. I think this could instead be `float flexibilityNegative = float.NaN` to avoid memory allocations here.

src/microbe_stage/editor/ToleranceRangeDisplay.cs
```diff
@@ -0,0 +1,199 @@
...
+        upperBoundPos = Size.X * (float)upperBoundFraction - 1;
+
+        QueueRedraw();
+    }
+
+    /// <summary>
+    ///   Sets the color of the range between the upper and lower bounds and queues a redraw.
+    /// </summary>
+    /// <param name="color">New color. If null, will default to <see cref="mainColor"/></param>
+    public void SetColorsAndRedraw(Color? color)
```
hhyyrylainen, line 171: Wouldn't this make much more sense to be a property? It's kind of not C# style to have setter methods as we have properties.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -214,12 +211,26 @@ public override void OnFinishEditing()
...
+    public void OnDataTolerancesDependOnChanged(bool organellesChanged)
     {
+        if (organellesChanged)
+            CalculateOrganelleModifiers();
+
         UpdateCurrentValueDisplays();
         UpdateToolTipStats();
     }
 
+    public void CalculateOrganelleModifiers()
```
hhyyrylainen, line 223: Does this need to be public?

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -214,12 +211,26 @@ public override void OnFinishEditing()
...
+    {
+        if (Editor.EditedCellOrganelles == null)
+            return;
```
hhyyrylainen, line 226: This should print at least an error message.

src/microbe_stage/editor/tooltips/StatModifierToolTip.tscn
```diff
@@ -66,7 +70,7 @@ autowrap_mode = 2
 [node name="EffectiveValueTitle" type="Label" parent="MarginContainer/VBoxContainer/Top"]
 custom_minimum_size = Vector2(100, 0)
 layout_mode = 2
-text = "EFFECTIVE_VALUE"
+text = "TOTAL_VALUE"
```
hhyyrylainen, line 73: Doesn't this change impact other things as well as the tolerances because this is in `StatModifierToolTip` and I'm pretty sure this is used outside the tolerances GUI as well?

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -352,14 +361,19 @@ protected override void RegisterTooltips()
...
-        oxygenResistanceContainer.RegisterToolTipForControl("oxygenResistance", "tolerances");
-        oxygenResistanceModifierLabel.RegisterToolTipForControl("oxygenResistanceModifier", "tolerances");
-        uvResistanceContainer.RegisterToolTipForControl("uvResistance", "tolerances");
-        uvResistanceModifierLabel.RegisterToolTipForControl("uvResistanceModifier", "tolerances");
+        temperatureToolTipContainer.RegisterToolTipForControl("temperature", "tolerances");
+        pressureToolTipContainer.RegisterToolTipForControl("pressure", "tolerances");
+        oxygenResistanceToolTipContainer.RegisterToolTipForControl("oxygenResistance", "tolerances");
+        uvResistanceToolTipContainer.RegisterToolTipForControl("uvResistance", "tolerances");
+
+        temperatureModifierLabel.GetParent<Control>()
```
hhyyrylainen, line 369: Requiring to get the parent is kind of a code smell... I think it would be better to explicitly get the parent container references configured as `[Export]` variables like it is done elsewhere in the Thrive codebase.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -733,43 +644,57 @@ private void UpdateCurrentValueDisplays()
...
+            };
+        }
+        else
+        {
+            temperatureModifierLabel.GetParent<Control>().Visible = false;
+        }
+
         // Show in red the conditions that are not matching to make them easier to notice
-        if (Math.Abs(patchTemperature - preferredTemperatureWithOrganelles) >
+        if (Math.Abs(patchTemperature - CurrentTolerances.PreferredTemperature) >
```
hhyyrylainen, line 692: So now only conditions that are without modifiers bad are shown?

Wouldn't this make the GUI less helpful? Or does something compensate to show this? I haven't playtested yet so it might be the case that like the other modifier place turns red in this case and that helps?

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -779,189 +704,178 @@ private void UpdateCurrentValueDisplays()
...
             temperatureMinLabel.LabelSettings = perfectValueFont;
             temperatureMaxLabel.LabelSettings = perfectValueFont;
+
+            temperatureRangeDisplay.SetColorsAndRedraw(optimalDisplayGoodColor);
         }
         else
         {
             temperatureMinLabel.LabelSettings = originalTemperatureFont;
             temperatureMaxLabel.LabelSettings = originalTemperatureFont;
+            temperatureRangeDisplay.SetColorsAndRedraw(null);
```
hhyyrylainen, line 722: I think this is another situation with boxing, so I think a split would work where there's a property to set the colour and a method to clear the colour override? if that's what the null here means.

src/general/EnvironmentalTolerances.cs
ender-titan1: I might have missed something, but I don't think there was any point having default values here in the first place, as any time this constructor is used, it's followed by `CopyFrom`. 

The only case where that doesn't happen is in the unit tests, but in my opinion it would make more sense to assign test values there then have a bunch of default values in the class that almost always get immediately overwritten.

src/microbe_stage/editor/tooltips/StatModifierToolTip.cs
ender-titan1: Oh, to be honest I've never actually seen it spelt like that in my life...

src/microbe_stage/editor/tooltips/StatModifierToolTip.tscn
ender-titan1: Don't believe that it is. A quick search through the repo yielded only the one use.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
ender-titan1: I think this is because `PreferredTemperature` isn't actually affected by any modifiers, just the temperature range is, so there is no point in calculating `preferredTemperatureWithOrganelles`.

locale/en.po
ender-titan1: Are you referring to the sunlight and oxygen or to the penalties to health, osmoregulation and bioprocess speed?

In any case I feel like having a more specific and to-the-point tutorial that has to be updated when tolerances are updated is preferable to having a more general tutorial that doesn't require that change.

src/microbe_stage/Patch.cs
ender-titan1: I think that now with this implementation `organelleEffects.PressureMinimum` and `Maximum` should always just be zero. Same applies for `PreferredTemperature`. This is because organelles only modify the ranges, not preferred values.

src/general/EnvironmentalTolerances.cs
ender-titan1: So this was done for save continuity reasons, as old saves still use the min/max system. I don't know how else to convert the min/max system to min/range. I guess a save upgrader is an option but I can't help but feel like it's a bit overkill for something so simple.

Accidental-Explorer: Isn't the next release conveniently going to break save compatibility anyway?

src/general/EnvironmentalTolerances.cs
hhyyrylainen: Well, even if it is not currently used, I think it's still a useful thing to have. I know this is kind of a struct, but in OOP one key concept is that the constructor makes classes valid to use, so it should ensure that all values have sensible data in them after construction. So even if not currently used, I think the default values were useful at least in illustrating what kind of data the class might hold for future readers of the code.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
hhyyrylainen: I'm going to ask that you make it so. We should account for future flexibility if some organelle, for example, in the future wants to reduce the temperature a cell can be in. 

I just think that cutting out a part of the feature is going to make a potential future change so much harder, that we ought to do things fully now to support whatever tweaks someone doing balance changes might want to make.

src/microbe_stage/Patch.cs
hhyyrylainen: Again, I don't think we should cut out features that were already kind of supported. I think a tolerance GUI changes shouldn't remove features from the underlying system...

src/general/EnvironmentalTolerances.cs
hhyyrylainen: Yes, 0.8.4 is absolutely going to break all save compatibility as it will use a totally different saving format.

locale/en.po
hhyyrylainen: I mean at least this part: `to your [b]health[/b] [thrive:icon]HealthIcon[/thrive:icon], [b]osmoregulation[/b]`

> In any case I feel like having a more specific and to-the-point tutorial that has to be updated when tolerances are updated is preferable to having a more general tutorial that doesn't require that change.

No, because I will have to endure the maintenance burden, so I do not want this change. Or at least please do not bundle the change with the GUI changes that people already seem to want as it will be a pain for me to manually separate out just the GUI fixes and the required data model fixes for that from all of the other tweaks that are more for just "taste".

src/microbe_stage/editor/tooltips/StatModifierToolTip.tscn
hhyyrylainen: I see, so this whole tooltip type is only used there? I suppose that is fine, but the code still refers to it as effective" value, so that should be fixed for consistency.

src/auto-evo/mutation_strategy/ModifyEnvironmentalTolerance.cs
```diff
@@ -82,9 +82,9 @@ public class ModifyEnvironmentalTolerance : IMutationStrategy<MicrobeSpecies>
                 }
 #endif
 
-                var change = Math.Max(score.PerfectOxygenAdjustment, maxChange);
+                var change = Math.Max(score.PressureRangeSizeAdjustment, maxChange);
```
hhyyrylainen, line 85: I'm preparing a separate PR to get this quite important auto-evo fix in, and I noticed that even this fix still seems to have a typo, because this is used to modify temperature but the variable here refers to pressure range.

hhyyrylainen:
This feature was finished here: https://github.com/Revolutionary-Games/Thrive/pull/6725

PR #6327, opened 2025-09-08
Make disk cache saving and loading use background threads

hhyyrylainen:

Bit of an oops from me as I thought these were already setup to do that, but instead were running on the main thread. Hopefully this was the cause of the intermittent frequent lag spikes after playing for a while.

And hopefully this doesn't cause any stability problems. If this does then the image to texture conversion could be moved to be on the main thread.

**Related Issues**

Hopefully addresses lag spikes that when start persist for a while

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/caching/DiskCache.cs
+++ b/src/engine/caching/DiskCache.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Text;
 using System.Threading;
+using System.Threading.Tasks;
 using Godot;
 using Newtonsoft.Json;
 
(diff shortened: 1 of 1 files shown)


dligr (approved):
I tested this PR and didn't notice any lag spikes. Specifically, I added prints for when the image saving occurs and there was no lag when they triggered.

hhyyrylainen:
That's good. I assume you didn't notice any crashing or instability as you didn't mention any, which would have been my primary concern with these changes.

PR #6332, opened 2025-09-09
Ambient reflections

dligr:

Adds ambient reflections. Most noticable reflections are in the surface patches, especially costal (they are also affected by the day/night cycle in those patches).

Technically, reflections were already there, this PR just changes their source and adds variable colors, making them more noticable. Because of this the only performance impact is light level updating.

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/Biome.cs
+++ b/src/microbe_stage/Biome.cs
@@ -37,7 +37,7 @@ public class Biome : IRegistryType
     /// </summary>
     public LightDetails Sunlight = new();
 
-    public Color EnvironmentColour;
+    public Color EnvironmentColour = new(0, 0, 0, 1);
 
     /// <summary>
     ///   How much the temperature in this biome varies on a microscopic scale when moving around
@@ -104,6 +104,12 @@ public class Biome : IRegistryType
         {
             throw new InvalidRegistryDataException(name, GetType().Name,
                 "temperature variance scale needs to be over 0");
+        }
+
+        if (EnvironmentColour.A < 1.0f)
+        {
+            throw new InvalidRegistryDataException(name, GetType().Name,
+                "Environment colour alpha needs to be 1");
         }
 
         TranslationHelper.CopyTranslateTemplatesToTranslateSource(this);
--- a/src/microbe_stage/Biome.cs
+++ b/src/microbe_stage/Biome.cs
@@ -37,6 +37,8 @@ public class Biome : IRegistryType
     /// </summary>
     public LightDetails Sunlight = new();
 
+    public Color EnvironmentColour;
+
     /// <summary>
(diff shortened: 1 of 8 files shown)


third_party/JoltPhysics
hhyyrylainen, line 1: I think you accidentally changed the commit this submodule points to?

src/microbe_stage/MicrobeWorldEnvironment.cs
```diff
@@ -5,6 +5,8 @@
 /// </summary>
 public partial class MicrobeWorldEnvironment : WorldEnvironment
 {
+    private static StringName skyColorShaderParameter = new("colour");
```
hhyyrylainen: This should not be static, and this should be disposed in Dispose.

src/microbe_stage/Biome.cs
```diff
@@ -37,6 +37,8 @@ public class Biome : IRegistryType
     /// </summary>
     public LightDetails Sunlight = new();
 
+    public Color EnvironmentColour;
```
hhyyrylainen: As this doesn't look like it has a default value, the `Check` method should check if this value makes sense and if not throw an exception.

src/microbe_stage/MicrobeWorldEnvironment.cs
```diff
@@ -18,6 +20,11 @@ public override void _ExitTree()
         Settings.Instance.BloomStrength.OnChanged -= OnStrengthChanged;
     }
 
+    public void UpdateAmbientReflection(Color colour)
+    {
+        ((ShaderMaterial)Environment.Sky.SkyMaterial).SetShaderParameter(skyColorShaderParameter, colour);
```
hhyyrylainen: This should have an optimization where if this is called with the same `colour` as before this wouldn't call the shader update.

Also the cast to `(ShaderMaterial)` should be done in _Ready / _EnterTree so that it doesn't happen each frame.

src/microbe_stage/Biome.cs
dligr: Isn't the default value for `Color` black, which can be useful to disable reflections?

hhyyrylainen: Is it? I would have expected a 0-initialized struct to be the more likely case here, which would mean alpha is also 0, which then I would imagine could do some quite funky stuff. So actually I'd like to see the default value be explicitly `(0, 0, 0, 1)`.

dligr: Yeah, it seems to be the case that it's initialized with alpha as 0. Adding a check and a default value makes sense to me then.

PR #6333, opened 2025-09-12
Add a Miche for Endosymbiois-target Species

### LINKED ISSUE #5126: Make endosymbiosis target receive a population benefit
to help it stay around while being endosymsiosised

requires some change to auto-evo (something like bonus energy based on the energy gained by the species doing the endosymbiosing) or an external effect per endosymbiosis event giving some population increase.

Accidental-Explorer:

Adds a Miche to the Miche tree only in patches where the player species is present, while the player species has an ongoing Endosymbiosis process. 
Only the targeted endosymbiont species can fill this miche, and it receives energy based on the population of the host species in the patch.
This seems to practically prevent the endosymbiont species from going globally extinct until the process is complete.

* Local extinctions still happen. Intended in cases where the host is not present, unintended when caused by species splitting (and perhaps migration). But fixing the latter seemed out of scope here. (This mainly happens when the target endosymbiont happens to be doing quite well overall. So, silver lining?)
* The additional check for Player Species ID was necessary because descendent species apparently inherit the ongoing endosymbiosis process.
* Right now the endosymbiont receives energy based on the host's population and its own GetPredationTargetSizeFactor as a proxy for how much energy it takes. But this can of course be changed to a constant if preferred.
* I would like to eventually expand this to make the endosymbiont mutate in appropriate ways, it seemed worth pulling this first to make endosymbiosis more usable first.

**Related Issues**

Closes #5126 

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/auto-evo/steps/GenerateMiche.cs
```diff
@@ -166,6 +166,17 @@ public Miche GenerateMicheTree(AutoEvoGlobalCache globalCache)
 
         generatedMiche.AddChild(predationRoot);
 
+        // Endosymbiosis Miches
+        foreach (var possibleHost in patch.SpeciesInPatch.Keys)
```
adQuid: we could get some minor savings here by reusing the loop used for heterotrophic miches. I doubt that there will ever be enough species in the patch for big savings, but in a time-constrained area of the game it might help.

src/auto-evo/selection_pressure/EndosymbiosisPressure.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        return 0;
+    }
+
+    public override float GetEnergy(Patch patch)
+    {
+        if (!patch.SpeciesInPatch.TryGetValue(Host, out long population) || population <= 0)
+            return 0;
+
+        return population * Endosymbiont.GetPredationTargetSizeFactor();
```
adQuid, line 46: This does concern me since that could be weirdly big. I'm not sure what the energy ratio for a symbiont would be though. This might be a good thing for the theory team to chew on, but it might be smarter to just test this out in playtesting since we're already pushing the bounds of realism with this miche...

src/auto-evo/steps/GenerateMiche.cs
Accidental-Explorer: Should be easy to do. It would break the clean separation between Miche types there is in GenerateMiche.cs up until now though. Might make readability slightly worse? Is that an acceptable trade-off?

What I understand is that building the tree is not exactly the most expensive part of the auto-evo process, but every bit certainly helps.

src/auto-evo/selection_pressure/EndosymbiosisPressure.cs
Accidental-Explorer: The current solution essentially results in "one free endosymbiont population per host population". (Kind of represents them being fed by the host, and duplicating along with the host)

With a constant I would set it two the equivalent of a two or three hex size, so that the endosymbiont is only receiving "payment" for the final size of the target organelle. That will sometimes result in more population, usually the same or less, I think.

Of course, I can also use a constant on top of the current solution so it is not quite 1:1, but some amount less.

In any case it would be more realistic if it actually reduces the energy of the host I suppose. But I am not sure if the predation miche even does that/how that works.

src/auto-evo/steps/GenerateMiche.cs
adQuid: With good commenting, I think we can keep readability from dropping too low. Let's try it out!

src/auto-evo/selection_pressure/EndosymbiosisPressure.cs
adQuid: looks like the predation miche uses a flat energy multiplier, presumably to represent the energy loss from trophic levels: https://github.com/Revolutionary-Games/Thrive/blob/f532dcdbac3c3625811b8a08623e1c2f68ca7e63/src/auto-evo/selection_pressure/PredationEffectivenessPressure.cs#L73

That said, 1:1 population for a symbiote makes a lot of sense when I think about it. Meh, maybe just leave it.

src/auto-evo/steps/GenerateMiche.cs
Accidental-Explorer: Done, the predation loop has been re-labelled as a loop for any and all miches that need to be (potentially) generated per species in the patch.

So any similar hypothetical future miches like parasites or (non endo-)symbiosis should probably go into the same loop.

src/auto-evo/selection_pressure/EndosymbiosisPressure.cs
Accidental-Explorer: Okay, in that case I will just leave it like this for now and see if there is a need to change it later on.

hhyyrylainen:
The added file here needs to have UTF-8 BOM mark added at the start to fix one of the CI checks. And the other failing check wants a blank line between the fields of the new classes (for readability we want blank lines between properties that are more than 1 line long)

Accidental-Explorer:
> Code change looks good now.
> 
> Before this can be merged, the translation conflicts need resolving (I've found the quickest way to do that is to just reset all the translation files to master branch and then redo them from scratch after merge as you'll need to just copy-paste the one translation you did for en.po).

You mean, for now (before merge) within the branch just revert the changes?

hhyyrylainen:
> You mean, for now (before merge) within the branch just revert the changes?

That wouldn't result in the correct thing. How I personally do the translations is I update my local master branch version, then I merge that into my branch and for each merge conflict file I just accept the master branch version (as I have the local checks enabled the commit fails here with localizations out of date so my local git stops in the middle of a merge). Then I run the localization update and re-apply my English text changes, then I commit that whole thing as a merge.

src/auto-evo/selection_pressure/EndosymbiosisPressure.cs
GameDungeon: I'm slightly worried about a feedback loop between the two populations, as increasing one should end up increasing the other. This might be minor in the short timescales of symbioses though.

Accidental-Explorer: > I'm slightly worried about a feedback loop between the two populations, as increasing one should end up increasing the other. This might be minor in the short timescales of symbioses though.

I guess you mean because the host might also get energy from the endosymbiont through predation?

I guess I can write a check into the Predation Selection Pressure to disallow predating on your own endosymbiont. (would fit the in-game effect somewhat) Is that worth it?

But that might penalize the player population _a lot_ in auto-evo in some cases, so I would prefer to leave that for a later, more elaborate rework (allowing you to take energy from having a chloroplast in-progress, among other things).

GameDungeon: Endosymbiosis happens through predation, so I'm not sure about stopping the energy entirely, although I guess that makes some sense as it isn't being digested anymore. Sounds like a harsh punishment though.

I just want to make sure that the two populations don't randomly spike for no reason.

Accidental-Explorer: Understood. From my own observations, I saw nothing of the sort that stood out to me. But I suppose other people need to test it and see if they agree with me on that.

GameDungeon:
Outside of my comment there everything looks good to me.

hhyyrylainen:
I suppose this has been now been reviewed enough to merge.

PR #6340, opened 2025-09-16
Current visual improvement and trail-tearing fix

dligr:

1. Improves water current shader to enable lighting. Because trails are an imaginary representation of currents, it's not fully clear what they should look like, so I'm open to suggestions.
2. Prevents trails from being torn apart by shifting currents. Now all particles along a trail follow the exact same path as the lead particle does by using an offset time.

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Deus-Codes:
I do like this change, and do like your commitment towards working on current visuals. As you mention, it is a tricky thing to get right, so I think it's something that is naturally iterative. 

Some things to note...

1. I think one aspect that could be improved is the perception of waves not being at a uniform height and instead introducing depth, with some trails appearing closer to the camera and others looking further behind the cell. 

2. I think some of the "particle budget" being spent on more "dots" alongside these waves would also help break up their uniformity. I think even having a lot of the budget be spent on those dots to create a more hectic appearance, with less of that budget being on lines, would be something to consider.  The issue with lines is that they tend to have to be a lot more uniform, whereas dots can have more variable speeds and paths while still communication a direction (especially if mixed with lines). 

3. Related to point one - as Rathalos mentioned on Discord, one cool thing Spore does is have certain particles be out of focus and blurry. Not only does this add a layer of depth and complexity to the current graphics, but it also allows, for example, currents in the background that don't have a direct effect on the player. This "busy-ness" contributes to the feeling of being emersed in a fluid, but the lack of focus also makes sure that players don't get confused. 

I don't really have a direct recommendation here, but overall, I do think creating a sense of depth with current layers and utilizing more "dot" particles could be worth experimenting with. Overall, this PU is already an improvement in my eyes.

hhyyrylainen (approved):
From the technical side at least these changes seem to work fine.

PR #6342, opened 2025-09-20
Current particle depth

dligr:

Makes current particles (but not trails) spread over Y coordinate, with ones further away being bigger, slower, darker, and more transparent.

Also removes old environmental particles from volcanic vents and instead makes current particles orange to blend in better.

The next and final stage is probably to make textured particles, e.g. for the sea floor, and add bubble particles to surface patches.

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Patryk26g (approved):
The code and visuals look good to me

Deus-Codes:
Really like these changes. Surface patches look more empty without the particles being like those in the deeper patches - but like you mention, those being tweaked with surface textures. This in combination with chunks and - hopefully eventually - some sort of effect with floating large objects in the background layer will in a very pleasing representation of living as a tiny thing in a massive ocean.

One little suggestion I have: I do think particles in the background can move quicker, as they currently can appear somewhat still. They are more blurry and you did good work with dimming, so players can more easily understand them as not actually having an effect on movement.

hhyyrylainen:
> Also removes old environmental particles from volcanic vents and instead makes current particles orange to blend in better.

That was actually a thought when initially the current particles where added. I thought that why not make the particles match the overall colour of the particles in the patch?

hhyyrylainen:
I didn't playtest this yet, but doesn't this slightly reduce the verticality of the game as the "normal" old particles were able to fall down from above the camera?

hhyyrylainen (commented):
After testing this I feel like the particle colour and size looks good in the vents, but the seafloor patch looks worse than before as the big white blobs don't really look like anything, I'd much prefer if the old small size of currents was preserved in most other patches.

hhyyrylainen:
Sorry for spamming so many comments, but one more thing I noticed while playtesting is that in the epipelagic during the night the currents are basically invisible. I think the minimum light value for the currents needs increasing so that the currents aren't invisible for half of the game.

Accidental-Explorer:
First impression upon starting in the vents was: "wow, it's beautiful!"

But I agree with hhyrylainen that they look a bit more odd in the other patches. Maybe at least tone them down a bit in size and number until if/when we have textured particles that fit better there?

I do enjoy the overall business and depth they add to the screen.

dligr:
> But I agree with hhyrylainen that they look a bit more odd in the other patches. Maybe at least tone them down a bit in size and number until if/when we have textured particles that fit better there?

> After testing this I feel like the particle colour and size looks good in the vents, but the seafloor patch looks worse than before as the big white blobs don't really look like anything, I'd much prefer if the old small size of currents was preserved in most other patches.

It kind of does seem like this effect looks out of place in some patches. I think that we might want to experiment with their color to counter that. For example, this is how sea floor particles look with a more bluish color (they fit in a bit better in my opinion):

image

Should I do something similar for all non-surface patches or is it better to just disable depth variation for them?

Deus-Codes:
The depth variation is a net positive, I think where gripes can arise is just the particles themselves. The screenshot you have there is nice in my opinion. I think this is something that's a matter of the particle's texture if that makes sense, and having it vary in a way that is fitting of the different patches.

PR #6344, opened 2025-09-22
Stop species dropdown entries from colouring the selection radio buttons

### LINKED ISSUE #6307: Stop species dropdown entries from colouring the selection radio buttons
as these left buttons here are meant for normal selection, and are not meant to be affected by the species colour:

Image

This was likely unintentionally changed when we updated to Godot 4.

Some analysis by Athriq:

> Regarding that menu, the left buttons shouldn't have different hues (at least not what the original GUI design intended). Looking at the menu's code, maybe this line have something to do with:
CustomDropDown.cs
181 Popup.SetItemIconModulate(index, item.Color);

LMartin188:
closes #6307

### DIFF
--- a/src/gui_common/CustomDropDown.cs
+++ b/src/gui_common/CustomDropDown.cs
@@ -182,6 +182,7 @@ public partial class CustomDropDown : MenuButton
                         Popup.AddIconCheckItem(tintedIcon, item.Text, id);
                         var index = Popup.GetItemIndex(id);
                         Popup.SetItemIconMaxWidth(index, iconMaxWidth);
+
                         // Don't apply color modulation to checkable items as it affects the radio button icon
                         // The species color should only affect the custom icon, not the selection radio button
                         Popup.SetItemChecked(index, item.Checked);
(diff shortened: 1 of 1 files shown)


src/gui_common/CustomDropDown.cs
```diff
@@ -175,10 +175,16 @@ public void CreateElements()
...
                     {
-                        Popup.AddIconCheckItem(item.Icon, item.Text, id);
+                        // Use a pre-tinted icon so the species color is applied only to the custom icon
+                        // while the radio/checkbox indicator (selection control) keeps theme colors
+                        var tintedIcon = item.PreTintedIcon ??= CreateTintedIcon(item.Icon, item.Color);
+
+                        Popup.AddIconCheckItem(tintedIcon, item.Text, id);
                         var index = Popup.GetItemIndex(id);
                         Popup.SetItemIconMaxWidth(index, iconMaxWidth);
-                        Popup.SetItemIconModulate(index, item.Color);
```
hhyyrylainen, line 181: Upon reading that this SetItemIconModulate is from Godot, I think this might actually be an engine bug that should be reported that Godot 4 now applies icon modulation to the radiobutton as well rather than the icon only.

PR #6345, opened 2025-09-22
Make patches store event visuals history

### LINKED ISSUE #5672: Patch events data storage should be refactored to include historical events
so that they can be correctly displayed in the auto-evo exploring tool as swapping generations doesn't work correctly now

I did already request some changes to the code so the API interfaces to the rest of the code should be mostly done to support the patch event list to be persistent and storing the event times (instead of the current way that it is a list of just current events that gets cleared each time).

I think this would need to be done in time for 0.8.0 as otherwise the save data format for the events would end up changing.

dligr:

Adds a patch event history to make auto-evo exploring tool display past events correctly

**Related Issues**

Closes #5672

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/AutoEvoExploringTool.cs
+++ b/src/auto-evo/AutoEvoExploringTool.cs
@@ -668,7 +668,7 @@ public partial class AutoEvoExploringTool : NodeWithInput, ISpeciesDataProvider
         UpdateSpeciesList();
         SpeciesListMenuIndexChanged(0);
         UpdateCurrentWorldStatistics();
-        patchMapDrawer.UpdatePatchEvents();
+        patchMapDrawer.UpdatePatchEvents(world.GameProperties.GameWorld.TotalPassedTime);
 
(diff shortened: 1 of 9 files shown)


hhyyrylainen:
I'll note that as 0.8.4 is planning to break save compatibility anyway, this doesn't need to ensure the changes don't impact save compatibility.

Patryk26g:
I'm currently working on new patch events, in which I changed the way that events are stored and I also made so that the events are saved in the history so that will collide with my changes :/

hhyyrylainen:
Oh, I did consider it a bit strange to see this PR from dligr as I did recall seeing you doing some similar sounding changes on a branch.

dligr:
> I'm currently working on new patch events, in which I changed the way that events are stored and I also made so that the events are saved in the history so that will collide with my changes :/

Right, this PR seems unnecessary now that I see that `Events-refactor-and-env-volatility` also fixes this issue

PR #6349, opened 2025-09-26
Made auto-evo available enzyme calculation use actual enzymes

hhyyrylainen:
rather than the static organelle definition from the JSON which always had all enzymes defined leading to incorrect digestibility calculations

This PR does some stuff...

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/simulation/SimulationCache.cs
+++ b/src/auto-evo/simulation/SimulationCache.cs
@@ -58,6 +58,8 @@ public class SimulationCache
     private readonly Dictionary<(MicrobeSpecies, BiomeConditions), ResolvedMicrobeTolerances> cachedResolvedTolerances =
         new();
 
+    private readonly Dictionary<Enzyme, int> tempEnzymes = new();
+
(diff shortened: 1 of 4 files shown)


Accidental-Explorer:
At first glance using the name "tempEnzymes" in SimulationCache seems a bit weird, but I guess it is to make explicit it is not cached?

hhyyrylainen:
That's exactly the purpose (to be temporary). If the data was not "temporary" in that field, it would actually ruin everything. The same pattern is used elsewhere in the Thrive code to avoid short-lived memory allocations which cause garbage collector pressure (which is something you want to avoid in games as much as possible).

Accidental-Explorer:
Effect on auto-evo calculations confirmed to be working now. I tried to confirm if auto-evo mutations are still placing lysosomes, but the low occurrence of the Nucleus makes this difficult to verify.

Accidental-Explorer (approved):

PR #6361, opened 2025-10-01
Fix focus jumping after rebinding

### LINKED ISSUE #4638: Input rebinding conflict popup shouldn't scroll up the view on cancel
instead it should keep the GUI where it is so that the input that is still trying to rebind stays visible. Additionally it'd be pretty nice if active input rebinding controls would scroll themselves into view a few times per second to really ensure they are always visible.

### LINKED ISSUE #6206: After resolving a keybinding conflict the GUI focus jumps up a ton
which causes the player to lose sight of the keybinding they either just made or cancelled due to the conflict

I think this is caused as the popup correctly gets the focus but when it closes nothing is focused so the focus is grabbed by the focus grabber to the top of the current options menu tab. I think instead the rebinding control that is currently being interacted with should force grab keyboard focus once the popup closes.

dligr:

Fixes an issue where the input menu scrolled to the top after rebinding an action

**Related Issues**

Fixes #6206
Closes #4638

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/input/key_mapping/InputEventItem.cs
+++ b/src/engine/input/key_mapping/InputEventItem.cs
@@ -438,6 +438,11 @@ public partial class InputEventItem : MarginContainer
         return AssociatedEvent?.GetHashCode() ?? 13;
     }
 
+    public void GrabFocusDeferred()
+    {
+        button.CallDeferred(MethodName.GrabFocus);
+    }
+
     internal static InputEventItem BuildGUI(InputActionItem associatedAction, SpecifiedInputKey @event)
     {
         if (associatedAction.GroupList == null)
@@ -496,6 +501,7 @@ public partial class InputEventItem : MarginContainer
                 // because Equals treats it the same with the same AssociatedEvent.
                 AssociatedEvent = null;
                 Delete();
+                associatedAction.Inputs[i].GrabFocusDeferred();
                 return true;
             }
         }
--- a/src/engine/input/key_mapping/InputEventItem.cs
+++ b/src/engine/input/key_mapping/InputEventItem.cs
@@ -416,6 +416,11 @@ public partial class InputEventItem : MarginContainer
         OnKeybindingSuccessfullyChanged();
     }
 
+    public void MakeInputButtonGrabFocus()
+    {
+        button.GrabFocus();
+    }
+
     /// <summary>
     ///   Delete this event from the associated action and update the godot InputMap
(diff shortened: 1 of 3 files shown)


src/engine/input/key_mapping/InputGroupList.cs
```diff
@@ -21,6 +21,11 @@ public partial class InputGroupList : VBoxContainer
     private InputEventItem? latestDialogCaller;
     private InputEventItem? latestDialogConflict;
     private InputEvent? latestDialogNewEvent;
+
+    [Export]
+    private FocusGrabber focusGrabber = null!;
```
hhyyrylainen: This doesn't look like this is used for anything?

dligr:
I made deleting an `InputEventItem` automatically refocus to another node, which should fix both issues

hhyyrylainen:
I think it should default to focusing the node to the left (or right, and only up as a fallback), as focusing up is not possible on the upmost row of inputs, it causes this error: `WARNING: This control can't grab focus. Use set_focus_mode() and set_focus_behavior_recursive() to allow a control to get focus.`

and well most of the time when I'm deleting something, it is not the only input on the same row.

hhyyrylainen:
Also while rebinding a key I noticed that clicking the "x" button next to the "Press a key" text still somehow causes the focus to jump up. Hopefully that's not a very convoluted fix that would be required, but I'd be happy enough with the other fixes already if this turns out to be very hard.

src/engine/input/key_mapping/InputEventItem.cs
```diff
@@ -416,11 +417,32 @@ public void Rebind(InputEvent inputEvent)
...
     ///   Delete this event from the associated action and update the godot InputMap
     /// </summary>
+    /// <remarks>
+    ///   <para>
+    ///     Also changes GUI focus to another close button so that the focus doesn't default elsewhere
+    ///   </para>
+    /// </remarks>
     public void Delete()
     {
+        var nextFocusPath = new string?[]
```
hhyyrylainen: This allocates temporary memory... so in Thrive game context I think unfortunately the less readable code is preferred that doesn't allocate any extra memory. Also I think these focus things are NodePath type variables so comparing them against strings, again allocates a bunch of temporary memory... So this needs a bit of fiddling to get a variant of this code that doesn't allocate temporary memory.

PR #6362, opened 2025-10-02
Experimental auto evo chemoreceptor

### LINKED ISSUE #3034: Add some benefit in auto-evo calculation from chemoreceptor
right now only the player when playing gets something out of adding the organelle

Accidental-Explorer:

This PR adds auto-evo benefits to the chemoreceptor, so that auto-evo places them on AI cells, and the player is not punished as much for placing them.

I tried to keep the bonuses relatively small for now so that not _every single_ species contains them. I also used variable bonuses so that their usefulness isn't constant in every circumstance.
Right now, this consists of:
- For chunks and clouds, a base score addition, and a variable score addition which is inversely proportional to the amount of the cloud/chunk in the patch.
- For predation, a base score modifier on the active hunting scores, and a variable modifier to the same, which is proportional to the prey's individual energy cost. (not fully accurate to their actual population, but it's a number that is more stable)

PU was expanded to also check for the specific target selected, and allow auto-evo to control that selection.

**Related Issues**

closes #3034

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Accidental-Explorer:
Marking this as non-draft now because I need feedback on the number balance.

(And I need to be told where my code is bad)

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -300,6 +309,16 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
                 // You catch more preys if you are fast, and if they are slow.
                 // This incentivizes engulfment strategies in these cases.
                 catchScore += predatorSpeed / preySpeed;
+
+                // If you have a chemoreceptor, active hunting types are more effective
+                if (HasChemoreceptor(predator, biomeConditions))
```
hhyyrylainen: As below this code  HasChemoreceptor` is called anyway, I think this should be placed in a variable around line 296 or so in order to just call this method once.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -373,6 +403,75 @@ public bool GetUsesVaryingCompoundsForSpecies(MicrobeSpecies species, BiomeCondi
...
+    public bool HasChemoreceptor(MicrobeSpecies species, BiomeConditions biomeConditions)
+    {
+        var key = (species, biomeConditions);
+
+        if (cachedHasChemoreceptor.TryGetValue(key, out var cached))
+            return cached;
+
+        foreach (var organelle in species.Organelles.Organelles)
+        {
+            if (organelle.Definition.HasChemoreceptorComponent)
```
hhyyrylainen: This is such a simple check, that it might not be worth it to cache this separately. The other places that depend on this are already cached themselves, so this double layer of caching might just be adding a bunch of extra complexity that is already handled by the upper layer of caching well enough.

Or did you find this chemoreceptor check to add to the runtime of the other methods in a significant way?

Accidental-Explorer: It seemed to me that the processing time might escalate from running a loop too much, but I have no evidence for that. (basically just did it because GetPredationTools does, and it gets used in predation, cloud and chunk calculations + because of the plan for eventual expansion to also get the chemoreceptor targets)

I guess I have to figure out how to check the runtime to be sure.

hhyyrylainen: Just looping the organelles of a species is an operation that is done a bunch. So I would say that the operation here is so simple that the cost of calculating it and then storing it in a dictionary is likely more than what it would save. As the methods that would use the chemoreceptor check themselves are cached, the chemoreceptor check itself is done just a couple of times. So I think (without having profiled the code) that the tradeoff of caching is just way worse than not caching it. Thus I think this one specific piece of information should not be cached. As long as the loop does not allocate an enumerator, it should be pretty extremely cheap to just read a list of organelles. A dictionary will have a much higher constant cost than a linear scan of a small list.

There is a [dotbench](https://github.com/dotnet/BenchmarkDotNet) subproject setup for Thrive, but that's much harder to use than the JetBrains profiler available in Rider.

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: Okay, this part is done.

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: Okay, in that case I will definitely look into the profiler in Rider.

But for now just revised as suggested.

GameDungeon:
The Chemoreceptor needs to have a specific target, which this does not seem to handle at all. Currently, just having a chemoreceptor makes you way better at everything, which isn't really fair.

Edit: I reread your original post, and saw you addressed this, but I think this is a requirement for moving forward currently, as it is currently quite OP from an auto-evo perspective. Chunk and Compound species are pretty much guaranteed to evolve this, as there is no other way for a creature to match their score.

Accidental-Explorer:
> The Chemoreceptor needs to have a specific target, which this does not seem to handle at all. Currently, just having a chemoreceptor makes you way better at everything, which isn't really fair.
> 
> Edit: I reread your original post, and saw you addressed this, but I think this is a requirement for moving forward currently, as it is currently quite OP from an auto-evo perspective. Chunk and Compound species are pretty much guaranteed to evolve this, as there is no other way for a creature to match their score.

I only considered this a viable half-solution because as Hhyyrylainen previously stated, cell AI currently also does not consider chemoreceptor targets. But if you feel introducing this half-solution first is not a good idea, I will work on the final version of this first before moving on to other organelles.

> it is currently quite OP from an auto-evo perspective. Chunk and Compound species are pretty much guaranteed to evolve this, as there is no other way for a creature to match their score.

I think this is mainly a question of the design of the final bonus and number balance thereof. Because correctly implementing target reading will just move this from "one chemoreceptor is mandatory" to "chemoreceptor set to the target cloud/chunk is mandatory", as long as the numbers are still too strong.

For example, with the current numbers, hydrogen sulfide cloud usage at least part of the time disfavours chemoreceptors, because it is too abundant.

I think it's even an open question how common chemoreceptors _should_ be. I think as a player they're very widely useful.

Accidental-Explorer:
Well, I added in the checks for correct selected targets.

Auto-evo now seems very unlikely to first place a chemoreceptors, though they take over quite quickly afterwards.

Also the automated checks are complaining about something in OrganelleTemplate I didn't touch (and revising as suggested just raises a different warning).

hhyyrylainen:
Checks should be fixed after a merge of master. I fixed the field ordering: https://github.com/Revolutionary-Games/Thrive/pull/6366 which was probably a case of a newer version of some of the checks being installed since anyone last edited the organelle template file. So now with you editing it the checks re-checked the file and detected the problem only now.

Accidental-Explorer:
As far as I can tell, more different complaints about what I merged in from master.

hhyyrylainen:
Hmm. Seems like that's the fault of my PR for the environmental chunks... I really don't know why the CI isn't catching those.

As I'm in the middle of the massive save rework if have a bunch of in-progress changes making a context switch harder, could you edit the `src/microbe_stage/systems/MicrobeTerrainSystem.cs` file? There's just one line that needs changing:

```diff
--- a/src/microbe_stage/systems/MicrobeTerrainSystem.cs
+++ b/src/microbe_stage/systems/MicrobeTerrainSystem.cs
@@ -582,7 +582,7 @@ internal class SpawnedTerrainGroup(Vector3 position, float radius, uint groupId)
 
     public uint GroupId = groupId;
 
-    public bool MembersFetched = false;
+    public bool MembersFetched;
```

That should fix the only error left on this. I can eventually help but that might be still a few days away.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -373,6 +407,82 @@ public bool GetUsesVaryingCompoundsForSpecies(MicrobeSpecies species, BiomeCondi
...
+        }
+
+        cached = 0.0f;
+        var hasChemoreceptor = false;
+        foreach (var organelle in species.Organelles.Organelles)
+        {
+            if (!organelle.Definition.HasChemoreceptorComponent)
+                continue;
+
+            if (organelle.GetActiveTargetCompound() == Compound.Invalid)
```
hhyyrylainen: It would be slightly more efficient to store the result of this method call in a variable, so like:

```c#
var organelleTargetCompound = organelle.GetActiveTargetCompound();
if (organelleTargetCompound == Compound.Invalid)
     continue;

            if (organelleTargetCompound == compound.ID)
```

And that second check is also more efficient that way (as it avoids a dictionary lookup).

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -373,6 +407,82 @@ public bool GetUsesVaryingCompoundsForSpecies(MicrobeSpecies species, BiomeCondi
...
+                continue;
+
+            if (SimulationParameters.GetCompound(organelle.GetActiveTargetCompound()) == compound)
```
hhyyrylainen: Same optimization can be applied here as well.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -373,6 +407,82 @@ public bool GetUsesVaryingCompoundsForSpecies(MicrobeSpecies species, BiomeCondi
...
+                hasChemoreceptor = true;
+        }
+
+        if (hasChemoreceptor)
+        {
+            if (chunk.Compounds?.TryGetValue(compound.ID, out var compoundAmount) != true)
```
hhyyrylainen: I assume this method is meant to be called only on chunks that truly contain the given compound?

If so I would rewrite this line as `if (!chunk.Compounds!.TryGetValue(compound.ID, out var compoundAmount))`

because that will avoid allocating a boxed int per call. But I'd also add a comment here then that `// We use null suppression here as this method is only meant to be called on chunks that are known to contain the given compound`

GameDungeon:
From an auto-evo side, this looks good to me.

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: Okay, I'll do that. I basically copied this line from ChunkCompoundPressure.cs to make sure I called it without breaking anything.

And I think ChunkCompoundPressure.cs also does not run this check before it calls GetChemoreceptorChunkScore?

hhyyrylainen: I mean if you aren't sure the chunk actually has compounds (or the target compound). The method should *start* with that, before even checking the cache. As it would be totally useless to do a bunch of processing and then discover that the chunk didn't even have the given compound.

Accidental-Explorer: Well, this should basically only happen if someone messes up in AutoEvoGlobalCache.cs. 

Should I add a line to just throw an exception right at the start of ChunkCompoundPressure or Score in ChunkCompoundPressure.cs if that happens?

hhyyrylainen: Rather than checking for a situation that shouldn't happen, I think that null suppression approach would be better in this case. In C# if something is null it doesn't crash the program it just automatically makes a null reference exception.

Accidental-Explorer: Okay, then the last version I pushed should be as you suggested.

hhyyrylainen (approved):
I think the code is basically fine now. I did see one potential micro optimization where one square root calculation could be avoided, but that probably has minuscule impact on the overall auto-evo time.

PR #6375, opened 2025-10-13
Auto evo toxin upgrades

Accidental-Explorer:

This PR adds auto-evo support to toxin upgrades and the toxicity slider.

Macrolide and the channel inhibitor are set to decrease prey speed.

Toxic pilus upgrade was also added to auto-evo in the process.

Added cheap MP costs for "upgrading" the toxin organelles, pilus and slime jets back to their base version.

Adapted mutation strategies to add new species to check for each additional organelle upgraded, for upgrades that cost MP.

I added additional bonuses to the score for some upgrades (beyond my original logic) to try to increase their occurance, but this seems to not have much effect, raising the bonus even higher seems to actually have the opposite effect?

**Related Issues**

partially adresses #4095

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/Scripts/Scripts.csproj
+++ b/Scripts/Scripts.csproj
@@ -17,7 +17,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="AngleSharp" Version="1.3.0" />
+    <PackageReference Include="AngleSharp" Version="1.3.1" />
     <PackageReference Include="Karambolo.PO" Version="1.12.0" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.14.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
--- a/Thrive.csproj
+++ b/Thrive.csproj
@@ -50,7 +50,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.0.0" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="AngleSharp" Version="1.3.0" />
+    <PackageReference Include="AngleSharp" Version="1.3.1" />
     <PackageReference Include="GraphShape" Version="1.2.1" />
     <PackageReference Include="JetBrains.Annotations" Version="2025.2.2" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
--- a/doc/setup_instructions.md
+++ b/doc/setup_instructions.md
@@ -24,7 +24,7 @@ Prerequisites
 Godot .NET version
 ------------------
 
-The currently used Godot version is __4.4.1 .NET__. The regular version
+The currently used Godot version is __4.5 .NET__. The regular version
 will not work. You can download Godot here: https://godotengine.org/download/
 if it is still the latest stable version. If a new version of Godot has
 been released but Thrive has not been updated yet, you need to look
--- a/locale/de.po
+++ b/locale/de.po
@@ -8,8 +8,8 @@ msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
 "POT-Creation-Date: 2025-11-06 15:24+0200\n"
-"PO-Revision-Date: 2025-08-13 14:57+0000\n"
-"Last-Translator: Anonymous <noreply@weblate.org>\n"
+"PO-Revision-Date: 2025-11-10 08:33+0000\n"
+"Last-Translator: Mario Ebenhofer <mario.ebenhofer@ik.me>\n"
 "Language-Team: German <https://translate.revolutionarygamesstudio.com/projects/thrive/thrive-game/de/>\n"
 "Language: de\n"
 "MIME-Version: 1.0\n"
@@ -54,13 +54,13 @@ msgid "ACHIEVEMENTS"
 msgstr "Basisbewegung"
 
 msgid "ACHIEVEMENTS_TOTAL"
-msgstr ""
+msgstr "Errungenschaften {0} / {1}"
 
 msgid "ACHIEVEMENT_ACHIEVED"
-msgstr ""
+msgstr "Abgeschlossen"
 
 msgid "ACHIEVEMENT_LOCKED"
-msgstr ""
+msgstr "Ausstehend"
 
 msgid "ACTION_AWAKEN"
 msgstr "Erwachen ({0:F1}/{1:F1})"
@@ -163,9 +163,9 @@ msgstr ""
 "  Durchschnitt {1}; Standardabweichung {2}\n"
 "[b]Lebende Arten:[/b]\n"
 "  Durchschnitt {3}; Standardabweichung {4}\n"
-"[b]Artenzahl pro Bereich:[/b]\n"
+"[b]Artenzahl pro Gebiet:[/b]\n"
 "  Durchschnitt {5}; Standardabweichung {6}\n"
-"[b]Gesamtpopulation pro Bereich:[/b]\n"
+"[b]Gesamtpopulation pro Gebiet:[/b]\n"
 "  Durchschnitt {7}; Standardabweichung {8}\n"
 "[b]Durchschnittliche Größe der Mikrobenart (in hexagonalen Feldern):[/b]\n"
 "  Durchschnitt {9}; Standardabweichung {10}\n"
@@ -190,9 +190,8 @@ msgstr "Ammoniak"
 msgid "AMMONIA_COST"
 msgstr "Ammoniak-Kosten:"
 
-#, fuzzy
 msgid "AMMONIA_LEVELS"
-msgstr "Ammoniak-Kosten:"
+msgstr "Ammoniakvorkommen"
 
 msgid "AMOUNT_OF_AUTOSAVE_TO_KEEP"
 msgstr "Anzahl der Autospeicherungen:"
@@ -396,7 +395,7 @@ msgid "AUTO_EVO_RESULTS_GLOBAL_TITLE"
 msgstr "Globale Ergebnisse"
 
 msgid "AUTO_EVO_RESULTS_PATCH_TITLE"
-msgstr "Ergebnisse für den ausgewählten Bereich"
+msgstr "Ergebnisse für das ausgewählte Gebiet"
 
 msgid "AUTO_EVO_RUN_STATUS"
 msgstr "Laufstatus:"
@@ -417,19 +416,19 @@ msgid "AVAILABLE_MODS"
 msgstr "Verfügbare Mods"
 
 msgid "AVERAGE_SURFACE_CARBON_DIOXIDE_LEVEL"
-msgstr ""
+msgstr "Durchschnittliche CO2 Menge an der Oberfläche"
 
 msgid "AVERAGE_SURFACE_LIGHT_LEVEL"
-msgstr ""
+msgstr "Durchschnittliche Lichtmenge an der Oberfläche"
 
 msgid "AVERAGE_SURFACE_NITROGEN_LEVEL"
-msgstr ""
+msgstr "Durchschnittliche N2 Menge an der Oberfläche"
 
 msgid "AVERAGE_SURFACE_OTHER_LEVEL"
-msgstr ""
+msgstr "Durchschnittliches Vorkommen anderer Gase an der Oberfläche"
 
 msgid "AVERAGE_SURFACE_OXYGEN_LEVEL"
-msgstr ""
+msgstr "Durchschnittliche O2 Menge an der Oberfläche"
 
 msgid "AWAKENING_STAGE"
 msgstr "Phase des Erwachens"
@@ -450,7 +449,7 @@ msgid "BACKSPACE"
 msgstr "Rücktaste"
 
 msgid "BACK_TO_SETTINGS"
-msgstr ""
+msgstr "Zurück zu den Einstellungen"
 
 msgid "BACTERIAL_THERMOSYNTHESIS"
 msgstr "Bakterielle Thermosynthese"
@@ -853,22 +852,19 @@ msgid "CLICK_TO_SELECT"
 msgstr "Auswählen"
 
 msgid "CLIMATE_INSTABILITY"
-msgstr ""
+msgstr "Klimawandel"
 
-#, fuzzy
 msgid "CLIMATE_INSTABILITY_EXPLANATION"
-msgstr ""
-"Aktive Mikroben taumeln umher, wenn nichts Interessantes passiert.\n"
-"Sessile Mikroben sind stationär und warten darauf, dass sich die Umgebung verändert, bevor sie handeln."
+msgstr "(beeinflußt die Häufigkeit und Stärke von Gebietsereignissen wie z.B. einer globalen Eiszeit)"
 
 msgid "CLIMATE_STABILITY_AVERAGE"
-msgstr ""
+msgstr "normal"
 
 msgid "CLIMATE_STABILITY_STABLE"
-msgstr ""
+msgstr "stabil"
 
 msgid "CLIMATE_STABILITY_UNSTABLE"
-msgstr ""
+msgstr "instabil"
 
 msgid "CLOSE"
 msgstr "Schließen"
@@ -1034,26 +1030,22 @@ msgid "COMPOUND_HANDLE_TOP_UP"
 msgstr "Immer auf Ausgangsbestand auffüllen"
 
 msgid "COMPOUND_HANDLE_TOP_UP_ON_CHANGE"
-msgstr "Bei Bereichswechsel auffüllen"
+msgstr "Bei Gebietswechsel auffüllen"
 
-#, fuzzy
 msgid "COMPOUND_LEVEL_AVERAGE"
-msgstr "Durchschnitt"
+msgstr "gewöhnlich"
 
-#, fuzzy
 msgid "COMPOUND_LEVEL_HIGH"
-msgstr "Substanzhaushalt"
+msgstr "häufig"
 
-#, fuzzy
 msgid "COMPOUND_LEVEL_LOW"
-msgstr "Substanzwolken"
+msgstr "selten"
 
 msgid "COMPOUND_LEVEL_VERY_HIGH"
-msgstr ""
+msgstr "reichlich"
 
-#, fuzzy
 msgid "COMPOUND_LEVEL_VERY_LOW"
-msgstr "Alle behalten"
+msgstr "rar"
 
 msgid "COMPOUND_STORAGE_AMOUNT_DOES_NOT_LAST_NIGHT"
 msgstr "(nicht ausreichend, um die Nachtdauer von {0} s zu überstehen)"
@@ -1461,7 +1453,7 @@ msgid "CURRENT_WORLD_STATISTICS"
 msgstr ""
 "[b]Generationen:[/b]\n"
 "  {0}\n"
-"[b]Gesamtzahl Bereiche:[/b]\n"
+"[b]Gesamtzahl der Gebiete:[/b]\n"
 "  {1}\n"
 "[b]Gesamtzahl Auto-Evo-Zeit:[/b]\n"
 "  {2}\n"
@@ -1469,7 +1461,7 @@ msgstr ""
 "  {3}\n"
 "[b]Spezies noch am Leben:[/b]\n"
 "  {4}\n"
-"[b]Spezies Anzahl pro Bereich:[/b]\n"
+"[b]Spezies Anzahl pro Gebiet:[/b]\n"
 "  Durchschnitt {5}; Standardabweichung {6}\n"
 "[b]Gesamtpopulation:[/b]\n"
 "  {7}\n"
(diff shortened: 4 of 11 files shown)


Accidental-Explorer:
I am not going to mark this as non-draft yet because the correction to make upgrades cost MP has made them occur too infrequently to my taste and I have not fixed that yet. But if anyone wants to give suggestions on the code or try things out, please do.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -70,21 +93,50 @@ public UpgradeOrganelle(Func<OrganelleDefinition, bool> criteria, string upgrade
...
                 {
-                    // TODO: Once this is used with an upgrade that costs MP this will need to factor that in
+                    foreach (var availableUpgrade in organelle.Definition.AvailableUpgrades)
+                    {
+                        var availableUpgradeName = availableUpgrade.Key;
+                        if (availableUpgradeName == upgradeName)
+                        {
+                            mpcost = availableUpgrade.Value.MPCost;
+                        }
+                    }
```
GameDungeon: This should probably break out of the loop when an mpcost is found.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -70,21 +93,50 @@ public UpgradeOrganelle(Func<OrganelleDefinition, bool> criteria, string upgrade
...
+                            mutated.Add(new Tuple<MicrobeSpecies, double>(newSpecies, mp));
+                        }
+
+                        // temporary solution for chemoreceptors, which should not all be upgraded
+                        // should be removed once chemoreceptor target change costs MP
+                        if (organelle.Definition.HasChemoreceptorComponent)
+                        {
+                            mutated.Add(new Tuple<MicrobeSpecies, double>(newSpecies, mp));
+                            return mutated;
+                        }
```
GameDungeon: I'm going to advocate for sampling say 5 random organelles that support the upgrade and returning a species for each of them. I think that would remove most of the other logic and handle most of the edge cases.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
Accidental-Explorer: Right you are. I actually did that in UpgradeToxinOrganelle already, but missed also doing it here.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
Accidental-Explorer: Well specifically, the top one is so that the other customupgrades can still keep the current "mutate all organelles at once"  behaviour as soon as any of them get MP costs implemented. (currently, none of them do). Sampling just 5 would break that, is that acceptable?

The bottom is specifically because there is _definitely_ no point in setting more than one chemoreceptor to the same target. So random or no, it would be a complete waste of processing time to do more than one, no? Though I suppose that also depends on:

Which one do you mean:
- Select a random 5 fitting organelles and then do the same as current (species 1: organelle 1 mutated, species 2: organelle 1 _and_ 2 mutated, etc.)
- Create a species for each separate organelle mutated by itself (species 1: organelle 1 mutated, species 2: organelle 2 mutated, etc.)

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
Accidental-Explorer: Fixed now.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
GameDungeon: The latter. It would only end up with a single mp cost, so that is handled, and you would not need special handling for the chemoreceptor. I think AddOrganelle does something similar.

Accidental-Explorer: Okay, though in that case I think I will have to make sure that the method is set to repeatable, otherwise I think it will be overly limited in only being able to upgrade one organelle with this upgrade per generation.

Accidental-Explorer: Changed UpgradeOrganelle and UpgradeToxinOrganelle to use a random sample of eligible organelles. The method used in AddOrganelleAnywhere didn't really fit because I think that's taking a random sample from the list of organelles in the game? Meanwhile this had to take a random sample from placed organelles in the species.

I did find a solution that seems to work. Though I feel like my solution to select the organelle in each newSpecies to match the random sample from the baseSpecies feels a bit hacky. Seems like there should be a cleaner way to do that.

This seems to also make the ocurrance of some of the upgrades drop more? it's difficult for me to tell to be honest.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -52,7 +52,7 @@ public class SimulationCache
     private readonly Dictionary<(TweakedProcess, float, IBiomeConditions), ProcessSpeedInformation>
         cachedProcessSpeeds = new();
 
-    private readonly Dictionary<MicrobeSpecies, (float, float, float, float)>
+    private readonly Dictionary<MicrobeSpecies, (float, float, float, float, float, float, float, float, float, float)>
```
GameDungeon: I'm not sure if this is the PR to address this, but this is getting kinda ridiculous

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -525,30 +609,56 @@ public void Clear()
...
     }
 
-    public (float PilusScore, float OxytoxyScore, float SlimeJetScore, float MucocystsScore)
+    public (float PilusScore, float InjectisomeScore, float AverageToxicity, float OxytoxyScore, float CytotoxinScore,
+        float MacrolideScore, float ChannelInhibitorScore, float OxygenMetabolismInhibitorScore, float SlimeJetScore,
+        float MucocystsScore)
         GetPredationToolsRawScores(MicrobeSpecies microbeSpecies)
     {
         if (cachedPredationToolsRawScores.TryGetValue(microbeSpecies, out var cached))
             return cached;
```
GameDungeon: The toxin tools should probably be their own function at this point. I think the PredationTools function needed to be refactored anyway (possibly out of existence), but this is a little too far.

src/auto-evo/simulation/SimulationCache.cs
hhyyrylainen: Yeah... If this used at least like a record struct that would make this a lot nicer with quite few lines of code needed for that switch.

src/auto-evo/simulation/SimulationCache.cs
hhyyrylainen: The same record struct would help here a bit.

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: > I'm not sure if this is the PR to address this, but this is getting kinda ridiculous

Yeah, when I had to audibly count out the parts as I pasted them in I felt that was probably a bit off, but didn't know what else to do other than just expand the existing structure.

> Yeah... If this used at least like a record struct that would make this a lot nicer with quite few lines of code needed for that switch.

I guess I'll go look up what a record struct is.

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: > The toxin tools should probably be their own function at this point. I think the PredationTools function needed to be refactored anyway (possibly out of existence), but this is a little too far.

If for now it's just splitting up the function to make the toxin scores separate, I can manage that.

Accidental-Explorer: > The toxin tools should probably be their own function at this point. I think the PredationTools function needed to be refactored anyway (possibly out of existence), but this is a little too far.

> The same record struct would help here a bit.

Changed to record struct. Is splitting up the function into toxins and non-toxin tools still a desired measure here?

GameDungeon: It still bothers me a little bit how long and complex that function is, but I'm not sure if it can really be made much better, so I'm good with this as is. I'm not sure if splitting would make it a ton better. (Although maybe it should be split into like 5 functions really)

src/auto-evo/mutation_strategy/UpgradeToxinOrganelle.cs
```diff
@@ -0,0 +1,150 @@
...
+    private ToxinType toxinType;
+
+    /// <summary>
+    ///   Updates a toxin-launching organelle with a given toxin type, updating toxin type in custom data to match
+    ///   and randomly moves toxicity in a given direction.
+    /// </summary>
+    /// <param name="criteria">Organelle requirement to apply</param>
+    /// <param name="upgradeName">The name of the upgrade (from organelles.json) to apply</param>
+    /// <param name="direction">"increase", "decrease", or "both" to decide what to do with toxicity</param>
+    public UpgradeToxinOrganelle(Func<OrganelleDefinition, bool> criteria, string upgradeName, string direction)
```
hhyyrylainen: It sounds like this "direction" should be an enum. It's good programming practice to use an enum rather than a string with a few valid values in a situation like this to ensure that no one can pass in an invalid value that is not handled correctly.

Accidental-Explorer: Would it be better to use an enum, or two bools, one for enabling each direction?

hhyyrylainen: I think it would be better to use an enum that specifies all possible operation modes, which should be just 3 for now, right?

Accidental-Explorer: Yes it is just three, I'll use that solution then.

src/auto-evo/mutation_strategy/UpgradeToxinOrganelle.cs
```diff
@@ -0,0 +1,150 @@
...
+/// <summary>
+///   Specific mutationstrategy for toxin-launching organelles
+///   applies a fixed toxin type based on given upgradename
+///   and randomly adjusts toxicity upward, downward, or in both directions
+/// </summary>
+public class UpgradeToxinOrganelle : IMutationStrategy<MicrobeSpecies>
+{
+    private readonly Func<OrganelleDefinition, bool> criteria;
+    private readonly FrozenSet<OrganelleDefinition> allOrganelles;
+    private readonly string direction;
```
adQuid: not vital, but I would make this an enum, as it only works with a fixed set of values.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -285,20 +285,70 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
...
+        var preyOxygenUsingOrganellesCount = 0;
+        foreach (var organelle in prey.Organelles.Organelles)
+        {
+            if (organelle.Definition.IsOxygenMetabolism)
+                preyOxygenUsingOrganellesCount += 1;
+        }
+
+        // Calculating "hit chance" modifier from prey size and predator toxicity
+        var sizeHitFactor = Constants.AUTO_EVO_SIZE_AFFECTED_PROJECTILE_MISS_FACTOR / float.Sqrt(preyHexSize);
+        var toxicityHitFactor = toxicity / Constants.AUTO_EVO_TOXICITY_HIT_MODIFIER;
```
adQuid, line 328: I'm confused by this factor: more toxic toxins hit less often? This factor is being subtracted from the hit proportion.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -312,17 +362,24 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
...
+                // This incentivizes engulfment strategies in these cases.
+                catchScore += (predatorSpeed / slowedPreySpeed) * slowedProportion;
+            }
 
-                    // Uses crude estimate of population density assuming same energy capture
-                    catchScore *= 1 + Constants.AUTO_EVO_CHEMORECEPTOR_PREDATION_VARIABLE_MODIFIER
-                        * float.Sqrt(preyIndividualCost);
-                }
+            // If you have a chemoreceptor, active hunting types are more effective
+            if (hasChemoreceptor)
```
adQuid, line 376: was it too difficult to have this check if the chemorecepter looks for a specific species? I understand if so.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -340,38 +397,77 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
...
+
+        // Combine pili for further calculations
+        pilusScore += injectisomeScore;
+
         // Pili are much more useful if the microbe can close to melee
-        pilusScore *= predatorSpeed > preySpeed ? 1.0f : Constants.AUTO_EVO_ENGULF_LUCKY_CATCH_PROBABILITY;
+        if (predatorSpeed ! > preySpeed)
+        {
+            if (predatorSpeed > slowedPreySpeed)
+                pilusScore *= slowedProportion + Constants.AUTO_EVO_ENGULF_LUCKY_CATCH_PROBABILITY;
```
adQuid: This should probably be slowedProportion + (1-slowedProportion*lucky_catch)

src/auto-evo/simulation/SimulationCache.cs
hhyyrylainen: If I remember right, that's exactly it: higher toxicity means more damage per toxin hit but slower firerate.

Accidental-Explorer: hhyyrylainen is correct here. So the idea is that lower toxicity might be better for targets that are difficult to hit, while for bigger targets, high toxicity is more effective.

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: hasChemoreceptor might be a bit misnamed, but at line 315 you can see it actually _does_ check for chemoreceptor target species. (That was a previous PU)

src/auto-evo/mutation_strategy/UpgradeToxinOrganelle.cs
Accidental-Explorer: hhyyrylainen agrees with you, so I am already planning to do this!

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: You're quite right, fixed now!

src/auto-evo/simulation/SimulationCache.cs
adQuid: right, forgot those two were tied together.

adQuid (approved):

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -9,14 +9,30 @@ public class UpgradeOrganelle : IMutationStrategy<MicrobeSpecies>
...
-    public UpgradeOrganelle(Func<OrganelleDefinition, bool> criteria, string upgradeName)
+    /// <summary>
+    ///   Updates an organelle type with the given upgrade, creating a new species to test for every additional
+    ///   organelle upgraded.
+    /// </summary>
+    /// <param name="criteria">Organelle requirement to apply</param>
+    /// <param name="upgradeName">The name of the upgrade (from organelles.json) to apply</param>
+    /// <param name="shouldRepeat">
+    ///   Determines whether this mutation strategy can be used multiple times
+    /// should be false for any upgrade that does not cost MP</param>
```
hhyyrylainen:
```suggestion
    /// <param name="shouldRepeat">
    ///   Determines whether this mutation strategy can be used multiple times
    ///   should be false for any upgrade that does not cost MP
    /// </param>
```

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -340,38 +397,82 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
...
+
+        // But prey that resist toxin damage are less vulnerable to the injectisome
+        injectisomeScore /= prey.MembraneType.ToxinResistance;
+
+        // Combine pili for further calculations
+        pilusScore += injectisomeScore;
+
         // Pili are much more useful if the microbe can close to melee
-        pilusScore *= predatorSpeed > preySpeed ? 1.0f : Constants.AUTO_EVO_ENGULF_LUCKY_CATCH_PROBABILITY;
+        if (predatorSpeed ! > preySpeed)
```
hhyyrylainen: What is this syntax? I've not seen this before.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -571,21 +695,112 @@ public void Clear()
 
             foreach (var process in organelle.Definition.RunnableProcesses)
             {
-                if (process.Process.Outputs.TryGetValue(oxytoxy, out var oxytoxyAmount))
+                // Big branch to calculate scores for each toxin type
+                if (process.Process.Outputs.TryGetValue(oxytoxy, out var toxinAmount))
                 {
-                    oxytoxyScore += oxytoxyAmount * Constants.AUTO_EVO_TOXIN_PREDATION_SCORE;
+                    if (organelle.GetActiveToxin() == ToxinType.Oxytoxy && !hasOxytoxy)
```
hhyyrylainen: There should be just one call to GetActiveToxin and stored in a local variable for efficiency.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -747,4 +962,16 @@ private float CalculateAngleMultiplier(Hex pos)
         // If degrees is higher than 40 then return 0
         return angleCos >= 0.75 ? angleCos : 0;
     }
+
+    // helper for GetPredationToolsRawScores
+    public readonly record struct PredationToolsRawScores(float PilusScore,
```
hhyyrylainen, line 968: The record struct looks nice!

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -40,6 +57,12 @@ public UpgradeOrganelle(Func<OrganelleDefinition, bool> criteria, string upgrade
             return null;
         }
 
+        // If a cheaper organelle upgrade gets added, this will need to be updated
+        if (mp < 10)
+            return null;
+
```
hhyyrylainen: I think this code should be removed as this seems very unsafe as someone will add such an upgrade at some point, and this will be totally forgotten by then.

I'd allow this code *if* there's a check in SimulationParameters verifying on each game start that this is true.
I'd recommend just removing this so that such complex code doesn't need to be written.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
```diff
@@ -60,31 +83,60 @@ public UpgradeOrganelle(Func<OrganelleDefinition, bool> criteria, string upgrade
...
             {
-                var organelle = organelleList[i];
-
                 if (allOrganelles.Contains(organelle.Definition))
+                    eligibleOrganelles.Add(organelle);
+            }
+
+            var eligibleOrganelleSample = eligibleOrganelles
+                .OrderBy(_ => random.Next())
+                .Take(Constants.AUTO_EVO_ORGANELLE_UPGRADE_ATTEMPTS);
```
hhyyrylainen: Is this random sampling really necessary as this seems like the reason why a temporary list needs to be allocated here?

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: Oh, that should really be `if (predatorSpeed <= preySpeed)`, not sure how that slipped by me. Made a commit now.

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
Accidental-Explorer: Having a random sampling was a request from @GameDungeon.

There's a difference between applying a Macrolide upgrade to toxisome that currently has cytotoxin and applying it to one that currently has a metabolism inhibitor, so the different options do have to be tried. I think the alternative would be to apply it to *every* toxin organelle in the organism, but the "random limited sample" was done in case there's a very high number of toxin organelles.

I am not entirely sure of the greater structure of auto-evo yet, but I think if the mutation strategy was limited to only trying to apply the upgrade to *one* suitable organelle, the AI occurrence of the upgrades would drop even lower?

src/auto-evo/mutation_strategy/UpgradeOrganelle.cs
Accidental-Explorer: I can definitely remove this. I believe I just copied this from PlaceOrganelleAnywhere. It must simply be there to avoid unnecessarily spending time on the next steps, but you're right that it's likely to be missed.

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: Quite right, done now.

hhyyrylainen:
So I'll assume that with the doubled bonus the modifications happen often enough now? And I can merge this finally.

Accidental-Explorer:
> So I'll assume that with the doubled bonus the modifications happen often enough now? And I can merge this finally.

They happen at least part of the time. Occurrence is still lower than I would like, but in any case I have a general balance pass on auto-evo planned for well before 1.0, which would require this to be re-balanced again anyway.

PR #6377, opened 2025-10-14
Events refactor and environmental volatility events

### LINKED ISSUE #5672: Patch events data storage should be refactored to include historical events

### LINKED ISSUE #6275: Refactor environmental events ending to recalculate conditions from scratch
As we get regular reports about the sunlight ending at a value like 133% or 60% after an event ends. 

I think we need to refactor the condition changing so that we have one piece of code that is given the active events and then it calculates the effective values, rather than trying to apply a change and then revert it with math as that approach seems to cause problems relatively often.

Patryk26g:

Changes this PR implements:

- The history of the patch events is shown in previous generation in Auto-Evo Exploration Tool
- Current events are stored in the patch class
- Add PatchEventsManager class for management of patches' temperatures and light level, because those can get a little bit more complicated
- Event tooltip can have custom text if provided - for example to give more information about compounds that it affects
- Adds 3 new patch events:

1. Runoff - happens in continental patches, increases density of some compounds and corresponding chunks
2. Upwelling - happens in ocean patches (at once in one region), increases density of some compounds and corresponding chunks. The chunks densities varies between patches, smaller chunks are more common in top patches and less common in the bottom patches
3. Current dilution - works in opposition to the previous 2 patches, reducing compounds and chunks densities. Works in 2 modes -> in continental (like runoff) and oceanic (like upwelling)

Closes #5672
Closes #6275

TODO:

- [x] Extract common code
- [x] Create event history icons and save the info to history

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/microbe_stage/BiomeConditions.cs
```diff
@@ -157,6 +161,8 @@ public static BiomeConditions ReadFromArchive(ISArchiveReader reader, ushort ver
         {
             Chunks = reader.ReadObject<Dictionary<string, ChunkConfiguration>>(),
             Pressure = reader.ReadFloat(),
+            StartingSunlightValue = reader.ReadFloat(),
```
hhyyrylainen: Adding new properties requires updating the SERIALIZATION_VERSION and adding handling for when reading the old version. We can't break save compatibility with two releases in a row so this code needs to be made so that it still works even when loading an older save.

src/microbe_stage/PatchMapGenerator.cs
```diff
@@ -1002,6 +1002,26 @@ private static void ApplyPatchEnvironmentVariation(PatchMap map, WorldGeneration
...
+            }
+
+            if (!hasSunlight)
+            {
+                GD.PrintErr("Patch map generator encountered patch with unexpectedly no sunlight.");
+                return;
+            }
+
+            patch.Biome.StartingTemperatureValue = patchTemperature.Ambient;
+            patch.Biome.StartingSunlightValue = patchSunlight.Ambient;
```
hhyyrylainen, line 1024: Wouldn't these be available anyway through the biome template on the patch? I think that would be a better approach as it would allow seamlessly working old saves.

src/microbe_stage/Patch.cs
```diff
@@ -784,6 +785,7 @@ public static PatchSnapshot ReadFromArchive(ISArchiveReader reader, ushort versi
             SpeciesInPatch = reader.ReadObject<Dictionary<Species, long>>(),
             RecordedSpeciesInfo = reader.ReadObject<Dictionary<Species, SpeciesInfo>>(),
             EventsLog = reader.ReadObject<List<GameEventDescription>>(),
+            ActivePatchEvents = reader.ReadObject<Dictionary<PatchEventTypes, PatchEventProperties>>(),
```
hhyyrylainen: Old version needs handling here as well. I think it would be nice to try to construct the active patch events from the old data is possible? If we do lose active events list on loading an old save, I think that would be fine enough to still mark the saves as compatible.

src/microbe_stage/BiomeType.cs
```diff
@@ -9,16 +9,16 @@
...
 {
-    Coastal,
-    Estuary,
-    Tidepool,
-    Epipelagic,
-    Mesopelagic,
-    Bathypelagic,
-    Abyssopelagic,
-    Seafloor,
+    Coastal, // continental biome
```
hhyyrylainen: I think putting these comments like this doesn't look very nice...

Mainly as we do not use trailing comments in Thrive code style.

Something like

```
// Continental biomes
a,
b,
c,

// Ocean biomes
d,
...
etc.
```

Would be better, though slightly non-optimal as these values aren't all in order and trying to change the order would break saves probably.

src/general/GameWorld.cs
```diff
@@ -76,13 +76,21 @@ public GameWorld(WorldGenerationSettings settings, Species? startingSpecies = nu
...
-                new UnderwaterVentEruptionEffect(this, random.Next64()));
+                new UnderwaterVentEruptionEvent(this, random.Next64()));
+            TimedEffects.RegisterEffect("runoff_event",
+                new RunoffEvent(this, random.Next64()));
+            TimedEffects.RegisterEffect("upwelling_event",
+                new UpwellingEvent(this, random.Next64()));
+            TimedEffects.RegisterEffect("current_dilution_event",
+                new CurrentDilutionEvent(this, random.Next64()));
+            TimedEffects.RegisterEffect("patch_events_manager",
+                new PatchEventsManager(this, random.Next64()));
```
hhyyrylainen, line 93: To make old saves compatible again, I think this makes things slightly complicated. I think what should be done here is bumping up the serialization version and then on load checking if the version is < 2 and if it is then creating these new events and appending them to the data. That should make old saves work well with the new patch events manager. This is the power of the new saving system that should allow much easier save compatibility across versions.

locale/en.po
```diff
@@ -9491,6 +9500,16 @@ msgstr "Zoom in"
 msgid "ZOOM_OUT"
 msgstr "Zoom out"
 
+#~ msgid "EVENT_RUNOFF_TOOLTIP"
```
hhyyrylainen: Why is this tooltip in a deleted state? Was the tooltip removed for some reason from the new events.

assets/textures/gui/bevel/CurrentDilution.svg.import
```diff
@@ -0,0 +1,37 @@
...
+mipmaps/generate=true
+mipmaps/limit=-1
+roughness/mode=0
+roughness/src_normal=""
+process/fix_alpha_border=true
+process/premult_alpha=false
+process/normal_map_invert_y=false
+process/hdr_as_srgb=false
+process/hdr_clamp_exposure=false
+process/size_limit=0
```
hhyyrylainen: Are the SVGs naturally small or? I think the size limit ought to be set to 128 or 256 to ensure no unnecessary graphics memory use.

src/general/world_effects/PatchEventVisuals.cs
```diff
@@ -0,0 +1,69 @@
...
+        [PatchEventTypes.MeteorPlainImpact] = new LocalizedString("EVENT_METEOR_PLAIN"),
+        [PatchEventTypes.MeteorIronImpact] = new LocalizedString("EVENT_METEOR_IRON"),
+        [PatchEventTypes.MeteorPhosphatesImpact] = new LocalizedString("EVENT_METEOR_PHOSPHATES"),
+        [PatchEventTypes.MeteorRadioactiveImpact] = new LocalizedString("EVENT_METEOR_RADIOACTIVE"),
+        [PatchEventTypes.MeteorGlucoseImpact] = new LocalizedString("EVENT_METEOR_GLUCOSE"),
+        [PatchEventTypes.MeteorSulfurImpact] = new LocalizedString("EVENT_METEOR_SULFUR"),
+    };
+}
+
+public class PatchEventProperties
```
hhyyrylainen: I think this should be in its own file as this isn't really logically related to "patch event visuals"

src/microbe_stage/BiomeType.cs
Patryk26g: these changes were temporar just for me because sometimes I got confused. I will remove these

Patryk26g:
I'm not sure what I'm doing incorrectly, but I keep getting this error when I save and load the save:

Error Loading - An exception happened while loading the save data
exception: System.InvalidCastException: Unable to cast object of type 'HydrogenSulfideConsumptionEffect' to type 'System.Collections.Generic.Dictionary`2[System.Int32,System.Int32]'.
   at SharedBase.Archive.SArchiveReaderBase.ReadObjectOrNull[T]() in /home/patryk/Desktop/Thrive/RevolutionaryGamesCommon/SharedBase/Archive/SArchiveReaderBase.cs:line 325
   at SharedBase.Archive.ISArchiveReader.ReadObject[T]() in /home/patryk/Desktop/Thrive/RevolutionaryGamesCommon/SharedBase/Archive/ISArchiveReader.cs:line 99
   at CurrentDilutionEvent.ReadFromArchive(ISArchiveReader reader, UInt16 version, Int32 referenceId) in /home/patryk/Desktop/Thrive/src/general/world_effects/patch_events/CurrentDilutionEvent.cs:line 83
   
I guess it's because CurrentDilutionEvent is where HydrogenSulfideConsumptionEffect used to be in order for some reason?

src/microbe_stage/PatchMapGenerator.cs
Patryk26g: but these values change during the gameplay, so I guess I'm not able to acces the first ever value?

Patryk26g:
Btw the functionality of this pr is not ready to be tested as I first need to sort out the saving

locale/en.po
Patryk26g: Sorry, fixed now

hhyyrylainen:
> I'm not sure what I'm doing incorrectly, but I keep getting this error when I save and load the save:

> Error Loading - An exception happened while loading the save data
exception: System.InvalidCastException: Unable to cast object of type 'HydrogenSulfideConsumptionEffect' to type 'System.Collections.Generic.Dictionary`2[System.Int32,System.Int32]'.

There's probably a mismatch in the write and read methods leading to out of sync reading of data.

src/microbe_stage/Patch.cs
```diff
@@ -229,7 +228,6 @@ public void WriteToArchive(ISArchiveWriter writer)
         writer.WriteObject(BiomeTemplate);
         writer.WriteObject(currentSnapshot);
         writer.WriteObject(gameplayPopulations);
-        writer.WriteObject(activeWorldEffectVisuals);
```
hhyyrylainen, line 232: You've removed a write here, but you haven't modified the version number nor adjusted the read so that it does a dummy read of this data and then discards the thing it read.

src/microbe_stage/BiomeConditions.cs
hhyyrylainen: Also serialization version is not updated here either and read not adjusted.

Patryk26g:
I think I am finished. I have tested the event changes and:

- new and old events saving and loading works properly
- other events end properly

Old saves compatibility:

- active events are displayed
- active events end properly
- new events begin to occur

The only thing that is left now to do now  is create black-and-white event icons which I will probably tomorrow

assets/textures/gui/bevel/CurrentDilution.svg.import
Patryk26g: I have updated the sizes to be the same as in other events

src/microbe_stage/PatchMapGenerator.cs
hhyyrylainen: But the patch *template* should never be allowed to change, so it should stay the same, right?

hhyyrylainen: Or was this that case that this was dynamically generated and not matching the template exactly? (I just saw that this comment was still in unresolved state)

src/general/world_effects/PatchEventProperties.cs
```diff
@@ -0,0 +1,54 @@
...
+    public ushort CurrentArchiveVersion => SERIALIZATION_VERSION;
+    public ArchiveObjectType ArchiveObjectType => (ArchiveObjectType)ThriveArchiveObjectType.PatchEventProperties;
+    public bool CanBeReferencedInArchive => false;
+
+    public static void WriteToArchive(ISArchiveWriter writer, ArchiveObjectType type, object obj)
+    {
+        if (type != (ArchiveObjectType)ThriveArchiveObjectType.PatchEventProperties)
+            throw new NotSupportedException();
+
+        writer.WriteObject((PatchSnapshot)obj);
```
hhyyrylainen:
```suggestion
        writer.WriteObject((PatchEventProperties)obj);
```
Typo?

src/general/world_effects/PatchEventProperties.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public static PatchEventProperties ReadFromArchive(ISArchiveReader reader, ushort version, int referenceId)
+    {
+        if (version is > SERIALIZATION_VERSION or <= 0)
+            throw new InvalidArchiveVersionException(version, SERIALIZATION_VERSION);
+
+        return new PatchEventProperties
+        {
+            CustomTooltip = reader.ReadString() ?? string.Empty,
```
hhyyrylainen: This should actually throw NullArchiveObjectException as the string is not nullable to catch corrupt save data.

Edit: so `reader.ReadString() ?? throw new NullArchiveObjectException(),`

src/general/world_effects/PatchEventProperties.cs
```diff
@@ -0,0 +1,54 @@
...
+using SharedBase.Archive;
+
+public class PatchEventProperties : IArchivable
+{
+    public const ushort SERIALIZATION_VERSION = 1;
+
+    public string CustomTooltip = string.Empty;
+    public float SunlightAmbientMultiplier = 1.0f;
+    public float TemperatureAmbientChange;
+    public float? TemperatureAmbientFixedValue;
```
hhyyrylainen: This causes boxing so rather this should just have the value and if needed a separate bool marking if it is valid or not.

src/general/world_effects/PatchEventProperties.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        if (version is > SERIALIZATION_VERSION or <= 0)
+            throw new InvalidArchiveVersionException(version, SERIALIZATION_VERSION);
+
+        return new PatchEventProperties
+        {
+            CustomTooltip = reader.ReadString() ?? string.Empty,
+            SunlightAmbientMultiplier = reader.ReadFloat(),
+            TemperatureAmbientChange = reader.ReadFloat(),
+            TemperatureAmbientFixedValue = reader.ReadFloat(),
```
hhyyrylainen, line 35: This read already removes the nullability so you could use `float.Nan` as the indicator value for no value for that field.

src/general/world_effects/patch_events/PatchEventsManager.cs
```diff
@@ -0,0 +1,97 @@
...
+
+    private GameWorld targetWorld;
+
+    public PatchEventsManager(GameWorld targetWorld, long randomSeed)
+    {
+        this.targetWorld = targetWorld;
+        random = new XoShiRo256starstar(randomSeed);
+    }
+
+    [JsonConstructor]
```
hhyyrylainen: We no longer use Json serialization so properties related to it should not be used for any gameplay code (only registry stuff is still loaded on game startup).

src/general/world_effects/patch_events/PatchEventsManager.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+
+            bool hasTemperature =
+                patch.Biome.ChangeableCompounds.TryGetValue(Compound.Temperature, out var currentTemperature);
+            bool hasSunlight = patch.Biome.ChangeableCompounds.TryGetValue(Compound.Sunlight, out var currentSunlight);
+
+            if (!hasTemperature)
+            {
+                GD.PrintErr("Patch event manager encountered patch with unexpectedly no temperature.");
+                return;
```
hhyyrylainen: These returns break all patches on one invalid one so I'd say these should be `continue` instead.

src/microbe_stage/BiomeConditions.cs
```diff
@@ -149,15 +153,35 @@ public static BiomeConditions ReadFromArchive(ISArchiveReader reader, ushort ver
...
+            maximumCompoundAmounts,
+            minimumCompoundAmounts)
         {
-            Chunks = reader.ReadObject<Dictionary<string, ChunkConfiguration>>(),
-            Pressure = reader.ReadFloat(),
+            Chunks = chunks,
+            Pressure = pressure,
         };
+
+        // Values for version <= 1 are properly set in Patch after loading.
```
hhyyrylainen, line 178: Could you elaborate on this? I've read most of the PR now and I still cannot find what guarantees old versions initialize these values correctly.

src/microbe_stage/BiomeConditions.cs
```diff
@@ -72,6 +72,10 @@ public BiomeConditions(Dictionary<Compound, BiomeCompoundProperties> compounds,
     /// </remarks>
     public float Pressure { get; set; } = 101325;
 
+    public float StartingSunlightValue { get; set; }
```
hhyyrylainen, line 79: Also these values should be documented as these are quite unclear to someone who hasn't yet read the entire world effects system thoroughly as to what these are used for and when it is safe to modify them (I *assume* it is never safe to modify these).

src/microbe_stage/Patch.cs
```diff
@@ -198,15 +198,41 @@ public static Patch ReadFromArchive(ISArchiveReader reader, ushort version, int
...
+        if (version <= 1)
+        {
+            var patchEventTypes = reader.ReadObject<List<PatchEventTypes>>();
+            foreach (var eventType in patchEventTypes)
+            {
+                currentSnapshot.ActivePatchEvents.Add(eventType, new PatchEventProperties());
+            }
+
+            var biomeReference = SimulationParameters.Instance.GetBiome(biomeTemplate.InternalName);
+            currentSnapshot.Biome.StartingSunlightValue = biomeReference.Conditions.Compounds
```
hhyyrylainen, line 219: Ah, now I see where the default value is initialized. A hint on the other deserialize method as a comment would help locating this code.

src/microbe_stage/PatchMapGenerator.cs
hhyyrylainen: I read the code with a bit more care and yeah, I think this was something we discussed already so I'll mark this as resolved.

src/general/world_effects/patch_events/CurrentDilutionEvent.cs
```diff
@@ -0,0 +1,276 @@
...
+                    }
+                }
+
+                // iron, hydrogensulfide, phosphates
+                ReduceChunks(patch, compound);
+            }
+
+            if (compoundChanges.Count > 0)
+            {
+                patch.Biome.ApplyLongTermCompoundChanges(patch.BiomeTemplate, compoundChanges, cloudSizes);
```
hhyyrylainen, line 225: This doesn't again seem totally bulletproof, but maybe this should be fine even with overlapping events as you wrote the active events handling so maybe you used the lessons learned there to make these new events not conflict?

Patryk26g: Yes, I think it should be fine. The events manager was needed to handle sunlight and temperature levels as they are more complicated. Compounds are different as their values simply start from 0 and their densities values are only added/substracted. Overlapping with other events doesn't create conflicts

hhyyrylainen: That sounds good. There was a lot of code in each event so it was hard to see what the actual changes would be.

Patryk26g:
I fixed the code and added black and white icons for the events

assets/textures/gui/bevel/CurrentDilutionEvent.svg.import
```diff
@@ -0,0 +1,43 @@
...
+process/channel_remap/red=0
+process/channel_remap/green=1
+process/channel_remap/blue=2
+process/channel_remap/alpha=3
+process/fix_alpha_border=true
+process/premult_alpha=false
+process/normal_map_invert_y=false
+process/hdr_as_srgb=false
+process/hdr_clamp_exposure=false
+process/size_limit=0
```
hhyyrylainen: It looks like these new files don't have size limits set.

Patryk26g: Done :)

src/general/world_effects/patch_events/CurrentDilutionEvent.cs
```diff
@@ -0,0 +1,276 @@
...
+    private void ChangePatchProperties()
+    {
+        foreach (var patchId in eventDurationsInPatches.Keys.ToList())
+        {
+            var patch = targetWorld.Map.Patches[patchId];
+            eventDurationsInPatches[patchId] -= 1;
+
+            if (eventDurationsInPatches[patchId] <= 0)
+            {
+                eventDurationsInPatches.Remove(patchId);
```
hhyyrylainen, line 190: So I assume it is intentional that this event doesn't seem to restore normal levels after ending?

src/general/world_effects/patch_events/PatchEventsManager.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+    }
+
+    public void OnTimePassed(double elapsed, double totalTimePassed)
+    {
+        foreach (var patch in targetWorld.Map.Patches.Values)
+        {
+            var totalTemperatureChange = 0.0f;
+            var totalSunlightMultiplication = 1.0f;
+            float? fixedTemperatureValue = null;
```
hhyyrylainen: Shouldn't this also now use the default value of Nan rather than boxing?

src/general/world_effects/patch_events/RunoffEvent.cs
```diff
@@ -0,0 +1,220 @@
...
+    private bool CanTriggerEvent(Patch patch, double elapsed)
+    {
+        return patch.IsContinentalPatch() && !eventDurationsInPatches.ContainsKey(patch.ID) &&
+            random.NextFloat() < GetChanceOfTriggering(elapsed) &&
+            !patch.ActivePatchEvents.ContainsKey(PatchEventTypes.CurrentDilution);
+    }
+
+    private void ChangePatchProperties()
+    {
+        foreach (int patchId in eventDurationsInPatches.Keys.ToList())
```
hhyyrylainen: Why is ToList used here? This will allocate a temporary list each call.

Sorry I didn't catch this earlier but I think this is a waste to allocate temporary memory and should be refactored around.

hhyyrylainen: It's probably for removing stuff, right?

So the usual pattern I use is having a persistent temporary data list in the object that is given the stuff to delete and then that is processed after the loop finishes to avoid having to duplicate the entire list to iterate.

src/microbe_stage/BiomeConditions.cs
```diff
@@ -72,6 +72,14 @@ public BiomeConditions(Dictionary<Compound, BiomeCompoundProperties> compounds,
     /// </remarks>
     public float Pressure { get; set; } = 101325;
 
+    /// <summary>
+    ///   Starting sunlight and temperature values for this biome when a patch is created. They are used in
+    ///   <see cref="PatchEventsManager"/> to modify the patch's sunlight and temperature when patch events occur..
```
hhyyrylainen:
```suggestion
    ///   <see cref="PatchEventsManager"/> to modify the patch's sunlight and temperature when patch events occur.
```

src/general/world_effects/patch_events/UpwellingEvent.cs
```diff
@@ -0,0 +1,249 @@
...
+                }
+            }
+        }
+
+        return hasOceanicPatch;
+    }
+
+    private void ChangePatchProperties()
+    {
+        foreach (int patchId in eventDurationsInPatches.Keys.ToList())
```
hhyyrylainen: Now that I'm reading all of the code very carefully I'm noticing the same pattern would be useful to apply here as well.

src/general/world_effects/patch_events/CurrentDilutionEvent.cs
Patryk26g: Yes, it is intentional, the effects of all 3 event persist after they have finished

hhyyrylainen:
I playtested this and it seemed to function. I think final balancing will need some tweaks in the future, but I noticed that in 10 generations I tested there were only 3 generations when the ocean floor next to my start didn't have a patch event, so I think the event chances are a bit too strong now. So maybe about half of the current rate would be a more reasonable first go at a balance?

Other than that and the slight temporary memory inefficiencies, this is starting to look good.

src/general/world_effects/patch_events/CurrentDilutionEvent.cs
hhyyrylainen: Thanks for confirming.

Patryk26g:
I have applied memory fixes and reduced the chance of the events occuring

PR #6378, opened 2025-10-15
Fix GUI jiggle (e.g. in the evolutionary tree)

### LINKED ISSUE #6044: Evolutionary tree nodes or lines lag behind the other elements leading to a jiggly look
I think this is caused by the lines being drawn with Godot draw calls but the nodes are then actual GUI controls. So the positions versus when the rendering update happens aren't exact so when panning with the mouse the nodes and lines are slightly out of sync with each other and the screen looks jiggly / fluid / non-rigid. I think it would be nice to find a fix.

I tried really quickly making the node positions apply only after a graphics draw but that doesn't help with the problem (and introduced cases where sometimes the positions got really stuck and out of sync).

dligr:

Fixes GUI jiggling, which was apparently caused by Godot snapping controls to pixels. Other draggable UIs (e.g. the patch map) were actually also affected.

**Related Issues**

Fixes #6044

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
Is there no other way to fix this? As due to this we may end up with pixelated fonts all over the GUI (or at least that's my understanding why GUI stuff is quite often rounded to nearest integer positions)?

dligr:
Actually, this option doesn't seem to affect fonts in any way. This option snaps control positions to pixels, which I think is supposed to be used in pixel art games. In Thrive's case this option (it's enabled by default) causes UI to be unsmooth, though it's only visible when viewed at smaller scales

hhyyrylainen:
I tested this and this *does* have an effect on most in game fonts. It's quite subtle to tell, but when zooming in it is clear.

Here's how the things look with this change (open up this image in a new tab and zoom in), with the fonts being slightly soft:

Kuvakaappaus - 2025-10-20 18-33-04


Whereas in 0.8.3 the exact screen at the exact same size and screenshot position (notice how also the positions are slightly shifted which I think is from the rounding):

Kuvakaappaus - 2025-10-20 18-33-12


In my eyes the fonts are in this example definitely much more clear, and even in game when doing stuff like scrolling big lists fast, I think the font softness is noticeable as extra blur.

Edit: even when not zooming in the clarity difference is quite noticeable here on Github.

Edit 2: I asked for second opinions on the dev discord just so that it isn't just my eyes that are seeing a difference.

hhyyrylainen:
Okay so before I forget: there were other people confirming that what I thought I saw in the example screenshot was happening (edit: or I should say that they were able to see the same thing). And thus while at full size in the game the effect is less noticeable, I think the tradeoff to make all fonts slightly blurry to fix one pretty niche case in the GUI is not really worth it. So it would be really great to be able to turn off this setting selectively (or find other workaround for the draggable container).

dligr:
I've finally had some time to work on this, but based on my testing it seems like GUI snapping doesn't affect text blurring.

Here's a screenshot from the Steam version of the game:
image
And here's one from the version with this PR, which I think looks identical:
image

Could the blurring be an issue that was already present before this PR, or am I missing something?

hhyyrylainen (approved):
I repeated my test as well and I can't replicate the slight extra blurriness result. So I guess this doesn't cause issues after all?

I guess we can try merging and see if anyone notices any GUI weirdness that may be caused by this before the 1.0 release.
Note that I didn't test the evolutionary tree itself, but I trust that this actually does fix that.

PR #6379, opened 2025-10-16
Add an achievements button to the pause menu

sedigoj664:

Added an achievements button to the pause menu.

screenshot

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/gui_common/menus/AchievementsView.cs
```diff
@@ -0,0 +1,55 @@
...
+    public delegate void OnClosedEventHandler();
+
+    /// <summary>
+    ///   Opens a popup for achievements and refreshes them
+    /// </summary>
+    public void OpenPopup()
+    {
+        achievementsPopup.OpenCentered(false);
+
+        achievementsGallery.Refresh();
```
hhyyrylainen, line 26: As this is now called whenever the pause menu button is pressed, I think the TODO about a more efficient refresh in AchievementsGallery should be done.-

src/general/AchievementsView.cs
```diff
@@ -0,0 +1,55 @@
...
+    public delegate void OnClosedEventHandler();
+
+    /// <summary>
+    ///   Opens a popup for achievements and refreshes them
+    /// </summary>
+    public void OpenPopup()
+    {
+        achievementsPopup.OpenCentered(false);
+
+        achievementsGallery.Refresh();
```
sedigoj664: There's already a TODO in the `AchievementsGallery`'s `Refresh()` function:

```
// TODO: could do a lighter refresh than this, but for now the performance of this is sufficient with the
// achievement count we have currently
```

src/general/PauseMenu.tscn
```diff
@@ -181,9 +182,9 @@ size_flags_horizontal = 3
 size_flags_vertical = 3
 theme = ExtResource("2")
 script = ExtResource("1")
-LoadSaveListPath = NodePath("CenterContainer/LoadMenu/SaveList")
```
hhyyrylainen, line 184: I think you accidentally unset this variable.

src/general/achievements/AchievementsGallery.cs
```diff
@@ -60,4 +43,24 @@ public void Refresh()
...
+
+        foreach (var achievement in AchievementsManager.Instance.GetAchievements())
+        {
+            // TODO: should this show a card that just has the text "hidden" on it?
+            if (achievement.HideIfNotAchieved && !achievement.Achieved)
+                continue;
+
+            achievementCards.TryGetValue(achievement.Identifier, out var achievementCard);
+
+            if (achievementCard != null)
```
hhyyrylainen:
```suggestion
            if (achievementCards.TryGetValue(achievement.Identifier, out var achievementCard))
```

I think that should be the right simpler syntax for this

src/gui_common/menus/AchievementsView.cs
hhyyrylainen, line 1: One minor thing with this file (and the associated scene) is that I think this should be in the `src/gui_common/menus` folder as the `general` folder is purely for non-GUI stuff as we have the gui_common folder separately.

src/gui_common/menus/AchievementsView.tscn
```diff
@@ -0,0 +1,57 @@
...
+grow_vertical = 2
+size_flags_horizontal = 3
+size_flags_vertical = 3
+mouse_filter = 2
+theme = ExtResource("1_cnemp")
+script = ExtResource("2_kipvk")
+achievementsPopup = NodePath("AchievementsPopup")
+achievementsGallery = NodePath("AchievementsPopup/AchievementsGallery")
+
+[node name="AchievementsPopup" parent="." instance=ExtResource("2_pkbja")]
```
hhyyrylainen, line 23: One problem with this popup now as it can be overlaid on the gameplay, you need to go into the actual scene instantiated here and make sure the popup scroll container has the option to *not* forward scroll events on. Because right now when scrolled all the way to the or bottom the mousewheel controls the zoom level in the game, which is very unintended when scrolling on the GUI. That's a very subtle thing that is easy to miss in Godot, but I noticed when testing as I have experience forgetting to set it.

src/general/PauseMenu.tscn
sedigoj664: This variable is not used anywhere, so Godot removed it automatically.

hhyyrylainen: Huh, that's strange. I just verified that the string "LoadSaveListPath" does not appear anywhere else in the codebase. 

So this must be some leftover stuff and thus fine to remove.

sedigoj664:
I'm not sure what else I'm required to do for this PR to be merged.

hhyyrylainen:
I've been extremely busy the past two weeks trying to rewrite the entire saving system for Thrive, which is also holding up the next release. So unfortunately I haven't had time, but I should be able to squeeze in checking this again today.

It's also unfortunate that no one else on the team really does PR reviews, so when I don't have time to do them, they will take a long time.

hhyyrylainen (approved):
I briefly retested and didn't see the scroll problem again, and no further problems.

PR #6397, opened 2025-10-28
Reimplemented editor saving

### LINKED ISSUE #6391: Editor saving needs to be reimplemented
I'm opening this issue to remind that this is the highest priority thing, and to be able to point to people something if I need to (if someone mentions the inability to save in the editor before this gets done)

hhyyrylainen:

Reimplements a key saving part

**Related Issues**

closes #6391

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "jetbrains.resharper.globaltools": {
-      "version": "2025.2.3",
+      "version": "2025.2.4",
       "commands": [
         "jb"
       ]
--- a/shaders/FluidCurrentTrail.gdshader
+++ b/shaders/FluidCurrentTrail.gdshader
@@ -1,6 +1,8 @@
 shader_type spatial;
 render_mode blend_mix, depth_draw_opaque, particle_trails;
 
+const float alphaScaling = 0.25f;
+
 void fragment() {
     float uvR = (UV.r * 2.0f - 1.0f);
     vec3 normal = normalize(BINORMAL * vec3(-1.0f, -1.0f, 0.0f)) * uvR
@@ -8,7 +10,7 @@ void fragment() {
     NORMAL = normal;
 
     ALBEDO = COLOR.rgb;
-    ALPHA = (1.0f - abs(uvR)) * COLOR.a;
+    ALPHA = (1.0f - abs(uvR)) * COLOR.a * alphaScaling;
 
     SPECULAR = 1.0f;
     ROUGHNESS = 0.4f;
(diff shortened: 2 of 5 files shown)


hhyyrylainen:
Should work now but there's likely some combination of actions I haven't tested that causes a problem.

HyperbolicHadron:
If I go to the load game menu for the first time, save preview information loads correctly. If go back without doing anything, and then re enter the load game menu for a second time, then frequently the saves will be missing all information aside from their name, like this screenshot shows. It seems like it will eventually load in all of the data and preview images eventually, but is much slower than normal. (Over like a minute or so instead of the near instantly normal load time)
Screenshot from 2025-10-29 18-02-01

It seems so far to always load in the save preview data correctly/instantly the first time the menu is accessed. All further accesses are unreliable, until the game is restarted or going back to the main menu.

This occurs for the save menu from the main screen, and from in game.

And when actually loading a save, it seems like it takes significantly longer to load the save when the save preview data has not loaded correctly, but it will still end up with the save loaded correctly after the extra time taken. (This may or may not just be the same amount of additional time the previous issue takes, I am not sure if there is a separate issue going on here)

Further notes:
Loading editor saves seems to be much faster than in game saves. And they also don't seem to load any slower when the save preview data has not yet loaded, they are still fairly instant.

All of the actual saves seem to have saved and loaded to the correct state while testing the microbe editor and gameplay saves, aside from the aforementioned additional delay.

hhyyrylainen:
> If I go to the load game menu for the first time, save preview information loads correctly. If go back without doing anything, and then re enter the load game menu for a second time, then frequently the saves will be missing all information aside from their name, like this screenshot shows. It seems like it will eventually load in all of the data and preview images eventually, but is much slower than normal. (Over like a minute or so instead of the near instantly normal load time)

That shouldn't have really changed with this PR. It's always been like that: the save information load is queued on a background thread, but not canceled if you close the menu. So if you have a very large number of saves or a slow computer, the second opening of the menu is still reading the previous load and thus will take until the previous queue is empty. I just tested on my computer and while the second load of the menu is ever so slightly slower, it still takes less than a second.

Something could be done but as I don't think my PR affects that behaviour at all, I won't attempt to do anything about that in this PR.

> And when actually loading a save, it seems like it takes significantly longer to load the save when the save preview data has not loaded correctly, but it will still end up with the save loaded correctly after the extra time taken. (This may or may not just be the same amount of additional time the previous issue takes, I am not sure if there is a separate issue going on here)

I'm pretty sure that again this is just the task queue being filled with the previous load stuff and needing to wait until that finishes.

I think a potential solution would be to make our task system tasks cancellable if they haven't started yet, and using that if the load menu closes. 

> All of the actual saves seem to have saved and loaded to the correct state while testing the microbe editor and gameplay saves, aside from the aforementioned additional delay.

Nice.

Issue #5947, opened 2025-02-24
Auto-evo environmental conditions adaptation should also consider widening the range of adaptation


hhyyrylainen:
sometimes rather than always shrinking

PR #6421, opened 2025-11-05
Fixed major problems with auto-evo tolerance changes

hhyyrylainen:

mainly auto-evo not changing tolerances at all and nonsensical adjustments that it would have calculated. Added also safety checking code and switched tolerance scores to be doubles to avoid imprecision problems from slightly off stats

**Related Issues**

Auto-evo not working for tolerances at all

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/mutation_strategy/ModifyEnvironmentalTolerance.cs
+++ b/src/auto-evo/mutation_strategy/ModifyEnvironmentalTolerance.cs
@@ -217,6 +217,7 @@ public class ModifyEnvironmentalTolerance : IMutationStrategy<MicrobeSpecies>
         {
             // Didn't find anything to do after all. This condition should be ensured to be rare as we wasted some
             // processing time here
+            GD.Print("Wasted processing time trying to calculate auto-evo tolerance change but didn't find it");
             return null;
         }
 


Accidental-Explorer:
On a quick review, code seems sane to my eyes.
For testing:

- Tolerances are indeed adjusted.
- Temperature matches environment quite closely. Though the tolerance range seems to always be 10C.
- Oxygen tolerance also keeps up.
- UV resistance seems to sometimes stall quite low around ~ 16%, which seems low both for normal surfaces _and_ ice patches. Seems like species reach a "good enough" point and then don't improve it further. But in other tests it _did_ go up to near 100, so I am not sure.
- Pressure seems to match well on most species. Some even seem to have it fine-tuned rather than having a wide tolerance. Seems like it's adjusted infrequently, but accurately.
- Definitely no detrimental effect on the evolution of other auto-evo traits.

Could be that the weight of the environmental tolerance miche should be tuned up.
It could also be that the MP cost for adjusting tolerances to the point where it's worth it is too much for the AI in some cases. 

But this is already a large improvement from before.

Accidental-Explorer (approved):

hhyyrylainen:
It's "intentionally" missing feature that species could also widen their adaptations:

https://github.com/Revolutionary-Games/Thrive/issues/5947

I still didn't add that feature as I don't really know when it would make sense to do so, and I purely focused on making fixes.

---

> UV resistance seems to sometimes stall quite low around ~ 16%, which seems low both for normal surfaces and ice patches. Seems like species reach a "good enough" point and then don't improve it further. But in other tests it did go up to near 100, so I am not sure.

I checked the code again and I'm 95% sure that UV resistance calculations and required changes are now correct. UV resistance change is the last thing to be applied so if a species runs out of MP it can't fully adapt it, or adapt it at all if earlier tolerance changes already ate up most of the MP. From what I've seen species usually have around 16-35 MP left when they are considering environmental tolerances (so most species first spend MP on organelles etc. before checking tolerances).

The species actually do not consider *any* amount of environmental negative effects fine. Basically the code checks that if the score is below 0.999999998 (when 1 is good adaptation score) then it will try to apply further tolerance changes.

Edit: so species should eventually have enough MP to do all tolerance changes they want to have score of 1. And on top there's then a less than 1% chance per mutation check for the species to want to become perfectly adapted for tolerance in which case the tolerance ranges are shrunk.

PR #6426, opened 2025-11-05
Implement collision clipping on cell division

CI09:

Added special case for collision ignores for ignores that disappear when moving away from ignored entity.

It was supposed to be for cell division but I figured I can just send this in and *maybe* make shader for it some day in the future, if I'll ever want to and it would be somewhat requested.

Also I'm not optimistic about how well I performed when writing this code, and I guess having visual stuff in the same PR wouldn't improve the situation much.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
+++ b/src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
@@ -16,6 +16,8 @@ using World = Arch.Core.World;
 [ReadsComponent(typeof(CellProperties))]
 [ReadsComponent(typeof(WorldPosition))]
 [ReadsComponent(typeof(CellDivisionCollisionDisabler))]
+[ReadsComponent(typeof(MicrobeColony))]
+[ReadsComponent(typeof(AttachedToEntity))]
 [WritesToComponent(typeof(CollisionManagement))]
 [RunsBefore(typeof(PhysicsCollisionManagementSystem))]
 public partial class MicrobeDivisionClippingSystem : BaseSystem<World, float>
@@ -57,6 +59,21 @@ public partial class MicrobeDivisionClippingSystem : BaseSystem<World, float>
 
             // 2.2 = 2 (radii) * 1.1
             var clipOutDistanceSquared = cellProperties.Radius * 2.2f;
+
+            // A partial fix for multicellular division
+            if (otherEntity.Has<MicrobeColony>())
+            {
+                ref var colony = ref otherEntity.Get<MicrobeColony>();
+
+                clipOutDistanceSquared += colony.GetApproximateColonyRadius();
+            }
+
+            if (entity.Has<MicrobeColony>())
+            {
+                ref var colony = ref entity.Get<MicrobeColony>();
+
+                clipOutDistanceSquared += colony.GetApproximateColonyRadius();
+            }
 
             // Square
             clipOutDistanceSquared *= clipOutDistanceSquared;
--- a/src/microbe_stage/components/MicrobeColony.cs
+++ b/src/microbe_stage/components/MicrobeColony.cs
@@ -15,7 +15,7 @@ using SharedBase.Archive;
 /// </summary>
 public struct MicrobeColony : IArchivableComponent
 {
-    public const ushort SERIALIZATION_VERSION = 1;
+    public const ushort SERIALIZATION_VERSION = 2;
 
     /// <summary>
     ///   All colony members of this colony. The cell at index 0 has to be the <see cref="Leader"/>. Only modify
@@ -58,6 +58,13 @@ public struct MicrobeColony : IArchivableComponent
     // certainly created.
 
     public int HexCount;
+
+    /// <summary>
+    ///   Cached approximate radius of the colony. 0 or negative is a placeholder value. To access this use
+    ///   <see cref="MicrobeColonyHelpers.GetApproximateColonyRadius"/>
+    /// </summary>
+    public float CachedRadius;
+
     public bool CanEngulf;
 
     /// <summary>
@@ -74,7 +81,7 @@ public struct MicrobeColony : IArchivableComponent
     /// <remarks>
     ///   <para>
     ///     It is mandatory to call <see cref="MicrobeColonyHelpers.AddInitialColonyMember"/> on each of the
-    ///     members *after* the leader to setup the state for the entities correctly.
+    ///     members *after* the leader to set up the state for the entities correctly.
     ///   </para>
     /// </remarks>
     public MicrobeColony(in Entity leader, MicrobeState initialState, params Entity[] allMembers)
@@ -109,6 +116,7 @@ public struct MicrobeColony : IArchivableComponent
         HexCount = 0;
         CanEngulf = false;
         DerivedStatisticsCalculated = false;
+        CachedRadius = 0;
         EntityWeightApplied = false;
     }
 
@@ -141,6 +149,7 @@ public struct MicrobeColony : IArchivableComponent
         HexCount = 0;
         CanEngulf = false;
         DerivedStatisticsCalculated = false;
+        CachedRadius = 0;
         EntityWeightApplied = false;
     }
 
@@ -158,6 +167,7 @@ public struct MicrobeColony : IArchivableComponent
         writer.Write(CanEngulf);
         writer.Write(DerivedStatisticsCalculated);
         writer.Write(EntityWeightApplied);
+        writer.Write(CachedRadius);
     }
 }
 
@@ -185,6 +195,7 @@ public static class MicrobeColonyHelpers
             CanEngulf = reader.ReadBool(),
             DerivedStatisticsCalculated = reader.ReadBool(),
             EntityWeightApplied = reader.ReadBool(),
+            CachedRadius = version > 1 ? reader.ReadFloat() : 0,
         };
     }
 
@@ -354,6 +365,64 @@ public static class MicrobeColonyHelpers
         }
 
         return result;
+    }
+
+    /// <summary>
+    ///   Calculates a rough overall radius of the colony.
+    ///   Only possible once members are initialised with their membrane. The value is cached for future calls.
+    /// </summary>
+    /// <param name="colony">Colony to calculate a radius for</param>
+    /// <returns>The radius of the colony *or* 0 if data is not available yet</returns>
+    public static float GetApproximateColonyRadius(this ref MicrobeColony colony)
+    {
+        if (colony.CachedRadius > 0)
+            return colony.CachedRadius;
+
+        float maxRadiusSquared = 0;
+
+        foreach (var cell in colony.ColonyMembers)
+        {
+            // Make sure leader is not unnecessarily processed, but if there exists a single member colony for a while,
+            // it should be processed correctly
+            if (cell == colony.Leader && colony.ColonyMembers.Length > 1)
+                continue;
+
+            float attachDistance = 0;
+
+            if (cell.Has<AttachedToEntity>())
+            {
+                attachDistance = cell.Get<AttachedToEntity>().RelativePosition.LengthSquared();
+            }
+            else if (cell == colony.Leader)
+            {
+            }
+            else
+            {
+                GD.PrintErr("Colony member has no attached component in radius calculation");
+            }
+
+            if (!cell.Has<CellProperties>())
+            {
+                GD.PrintErr("Cell colony has something that isn't a cell (CellProperties missing)");
+                return 0;
+            }
+
+            // If data not ready yet, skip calculation
+            var membrane = cell.Get<CellProperties>().CreatedMembrane;
+            if (membrane == null)
+                return 0;
+
+            var radius = membrane.EncompassingCircleRadius;
+            var distanceSquared = attachDistance + radius * radius;
+
+            if (distanceSquared > maxRadiusSquared)
+                maxRadiusSquared = distanceSquared;
+        }
+
+        var maxRadius = MathF.Sqrt(maxRadiusSquared);
+
+        colony.CachedRadius = maxRadius;
+        return maxRadius;
     }
 
     /// <summary>
@@ -1296,6 +1365,7 @@ public static class MicrobeColonyHelpers
     private static void MarkMembersChanged(this ref MicrobeColony colony)
     {
         colony.DerivedStatisticsCalculated = false;
+        colony.CachedRadius = 0;
         colony.EntityWeightApplied = false;
 
         // TODO: maybe in some situations creating the compound bag could be entirely safely skipped here


src/microbe_stage/components/CellProperties.cs
```diff
@@ -303,11 +303,6 @@ public static void Divide(this ref CellProperties cellProperties, ref OrganelleC
             }
         }
```
CI09: Okay so code above can be theoretically removed but maybe it could be used somewhere or at least kept for multicellular stuff.
Also checks might see this and not let this through.

CI09:
It also turned turned out that collisions get disabled for all cells in multicellular which lets you easily clip out. Also pili collisions too turned out to get disabled. Cool I guess.

hhyyrylainen:
I think it would be clearer to use the existing mechanics: so that the ignored collisions with variable would be used, but then a new system (and component) would be used to ensure the collision is turned back on after sufficient distance. That same system could also apply the force for separation.

That way the code is much neater as it'll be contained in a new system rather than shoving microbe specific stuff in the base *common* collision management system. Because that system is a common system it should be kept clear of all stage-specific code.

CI09:
Can I keep CollisionManagement.ClipOutIgnoredCollisions though?

hhyyrylainen:
That was the main thing I think should be removed and switched over to use `IgnoredCollisionsWith`...

src/engine/common_components/CellDivisionCollisionDisablerComponent.cs
```diff
@@ -0,0 +1,31 @@
...
+
+    public ushort CurrentArchiveVersion => SERIALIZATION_VERSION;
+    public ThriveArchiveObjectType ArchiveObjectType => ThriveArchiveObjectType.ComponentCellDivisionCollisionDisabler;
+
+    public void WriteToArchive(ISArchiveWriter writer)
+    {
+        // Save only persistent state
+
+        // doesnt work
+        //writer.WriteObjectOrNull(IgnoredCollisionWith);
```
CI09: I assumed I need to have this archived (first time coding stuff with new ecs) and appearently Entity is not archivable (even though List<Entity>? is)

hhyyrylainen: Entity is a value type which means that you should never write `Entity?` as that causes boxing. Instead `Entity.Null` is used as a marker for no entity. 

Also you should write a constructor for this Edit: ~~class~~ struct that forces giving the entity to not collide with, because structs when 0-initialized actually mean that the value will be `default(Entity)` which unfortunately with Arch *does not equal Entity.Null*. So that causes bugs when there are all 0 entities rather than properly setup Entity.Null values, which has caused bugs for me quite many times already.

And yes it should be archived.

CI09:
Now it is using a seperate system and component for all of that. There is still a matter of keeping (or not) the potentially useful multicellular code in CellProperties that I didn't know if I should completely remove.

src/engine/common_components/CellDivisionCollisionDisablerComponent.cs.uid
hhyyrylainen, line 1: I think this is a left over file from some earlier version

src/microbe_stage/components/CellProperties.cs
hhyyrylainen: What does it look like in multicellular? If the system already works for multicellular as well, then I think the code can be removed for that as well.

src/saving/ThriveArchiveObjectType.cs
```diff
@@ -218,6 +218,7 @@ public enum ThriveArchiveObjectType : uint
     ComponentStaticBodyMarker = 4302,
     ComponentTimedLife = 4303,
     ComponentWorldPosition = 4304,
+    ComponentCellDivisionCollisionDisabler = 4305,
```
hhyyrylainen: You cannot add a duplicate value here, this will blow up saves. You need to add the new value at the end (and sadly not grouped with the other components).

src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+    }
+
+    [Query]
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private void Update(ref CellDivisionCollisionDisabler collisionDisabler,
+        ref CollisionManagement collisionManagement, ref CellProperties cellProperties,
+        ref WorldPosition worldPosition)
+    {
+        ref var otherEntity = ref collisionDisabler.IgnoredCollisionWith;
```
hhyyrylainen, line 41: You should add a bool flag to CellDivisionCollisionDisabler to mark when it is done, otherwise this system infinitely runs and eats performance unnecessarily.

src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        if (collisionManagement.IgnoredCollisionsWith != null)
+        {
+            if (!collisionManagement.IgnoredCollisionsWith.Contains(otherEntity))
```
hhyyrylainen, line 47: This check doesn't seem safe to me as the entity can be ignoring collisions first and then get this component added, thus it would just disable this from running. I think the beforementioned bool flag to keep track of when this system has processed an entity would be a better solution.

src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
```diff
@@ -0,0 +1,72 @@
...
+                return;
+        }
+
+        if (otherEntity.Has<WorldPosition>())
```
hhyyrylainen: This should use `IsAliveAndHas` for extra safety.

src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
```diff
@@ -0,0 +1,72 @@
...
+            clipOutDistanceSquared *= clipOutDistanceSquared;
+            if (otherEntity.Get<WorldPosition>().Position.DistanceSquaredTo(
+                    worldPosition.Position) >= clipOutDistanceSquared)
+            {
+                collisionManagement.RemoveIgnoredCollisions ??= new List<Entity>();
+
+                if (!collisionManagement.RemoveIgnoredCollisions.Contains(otherEntity))
+                {
+                    collisionManagement.RemoveIgnoredCollisions.Add(otherEntity);
+                    collisionManagement.StateApplied = false;
```
hhyyrylainen: I think this would be the place to mark the current entity as having fully separated and never needing to be touched by this clipping system again.

src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
```diff
@@ -0,0 +1,72 @@
...
+            else
+            {
+                collisionManagement.IgnoredCollisionsWith ??= new List<Entity>();
+
+                if (!collisionManagement.IgnoredCollisionsWith.Contains(otherEntity))
+                {
+                    collisionManagement.IgnoredCollisionsWith.Add(otherEntity);
+
+                    collisionManagement.StateApplied = false;
+                }
```
hhyyrylainen: I think that to ensure separation, there should be a small physics force applied to both entities away from each other. That force could also scale with time to ensure that if the player tries to intentionally stick inside daughter cell, they eventually get pushed out. The time variable should be put into the CellDivisionCollisionDisabler component.

src/engine/common_components/CollisionManagement.cs
```diff
@@ -31,6 +31,12 @@ public struct CollisionManagement : IArchivableComponent
 
     public List<Entity>? IgnoredCollisionsWith;
 
+    /// <summary>
+    ///   Due to how ignoring works, this list provides all to-be-removed ignores for management system. All elements
```
hhyyrylainen: I think this comment needs rewording, because it's pretty hard to read for me.

I'd word it like this:

```
Once a collision ignore is added, it will be active forever. That is not always desirable so adding an entity to this list will remove a collision ignore once it is no longer in the <see cref="IgnoredCollisionsWith"/> list.
When a collision is removed, it is automatically removed from this list to ensure this list doesn't grow indefinitely.
```

Also this component needs its serialization version updated to 2 and this list needs to be saved. Old saves should handle things so that if version < 2 then just leave this new field null.

src/engine/common_systems/PhysicsCollisionManagementSystem.cs
```diff
@@ -86,6 +86,40 @@ private void Update(ref Physics physics, ref CollisionManagement collisionManage
...
+            ref var removeIgnoreCollisions = ref collisionManagement.RemoveIgnoredCollisions;
+
+            // Check if any ignores can be removed.
+            if (removeIgnoreCollisions != null)
+            {
+                if (removeIgnoreCollisions.Count > 0)
+                {
+                    // Collision management must keep detecting for clip-out ignore collisions.
+                    // Maybe throw this before StateApplied check?
+                    collisionManagement.StateApplied = false;
```
hhyyrylainen: This seems pretty bad to constantly force re-checking. I think this needs re-thinking. When there is a change to either list that should mean that StateApplied is set to false, so that should be enough to detect when something needs to be removed.

src/engine/common_systems/PhysicsCollisionManagementSystem.cs
```diff
@@ -86,6 +86,40 @@ private void Update(ref Physics physics, ref CollisionManagement collisionManage
...
+
+                    for (int i = 0; i < removeIgnoreCollisions.Count; ++i)
+                    {
+                        var clipOutIgnore = removeIgnoreCollisions[i];
+                        removeIgnoreCollisions.Remove(clipOutIgnore);
```
hhyyrylainen: This should not remove items before it actually removes the collision. Also this seems to *not* check ignoreCollisions so I think this whole loop should be rethought.

src/engine/common_systems/PhysicsCollisionManagementSystem.cs
```diff
@@ -86,6 +86,40 @@ private void Update(ref Physics physics, ref CollisionManagement collisionManage
...
+                        var ignoreWith = GetPhysicsForEntity(clipOutIgnore, ref collisionManagement);
+
+                        if (ignoreWith != null)
+                            physicalWorld.BodyRemoveCollisionIgnoreWith(physicsBody, ignoreWith);
+
+                        if (ignoreCollisions != null)
+                        {
+                            if (ignoreCollisions.Contains(clipOutIgnore))
+                                ignoreCollisions.Remove(clipOutIgnore);
+                        }
```
hhyyrylainen: In my opinion this should be written the opposite way around.

src/engine/common_systems/PhysicsCollisionManagementSystem.cs
```diff
@@ -86,6 +86,40 @@ private void Update(ref Physics physics, ref CollisionManagement collisionManage
...
+
+                        // Loop again
+                        i = 0;
```
hhyyrylainen: Is this about removing items? If so this should just be `--i;` rather than restarting the entire loop.

hhyyrylainen (changes requested):
Again I wrote my review as individual comments, oops. But overall I like the direction, just quite many of the details I think require some more polishing.

src/engine/common_systems/PhysicsCollisionManagementSystem.cs
CI09: You mean to put everything inside `ignoreCollisions` null check? I didn't do that in case there would be collision ignore set without having it in the `ignoreCollisions` list

src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
CI09: How about removing the component entirely when it is done?

hhyyrylainen: I think that would also be a suitable solution, though you'll need to use the command buffer for recording it to make sure no component structure modification happens while data is being processed.

src/engine/common_systems/PhysicsCollisionManagementSystem.cs
hhyyrylainen: I mean the logic should be:

- foreach thing in remove ignores
- if the item is in ignored collisions continue loop from next item
- remove the collision ignore
- remove the item from the remove ignores list

This will make it so that you can just put an item on the remove ignores list and it will be removed from the collision ignores once the corresponding item is removed from the main ignore list.

Does that make sense? It's kind of difficult to explain, but I think that would result in an easier to use API.

For stuff you want to permanently ignore just 

```
ignoredCollisionsWith.Add(entity);
```

And for stuff you want to eventually resume collisions you do:

```
ignoredCollisionsWith.Add(entity);
removeIgnoreCollisions.Add(entity);
```

Then when you remove the item from ignoredCollisionsWith (sorry if I misremembered the variable name), then removeIgnoreCollisions would trigger automatically and disable the collision ignore *and* clean its entry from removeIgnoreCollisions.

That would ensure the removeIgnoreCollisions doesn't have a bunch of no longer necessary data. And the decision if you want to ignore something permanently or not can be made just once when setting up the ignore.

This approach would also not require the marked status to be kept up at all times so the performance impact of the extra checks should be minimal as it only would impact whenever either list is modified.

To make the API simpler the collision management helpers could also get 3 new methods:

```
AddPermanentCollisionIgnoreWith(ref this Collisionmanagement..., Entity entity);
AddTemporaryCollisionIgnoreWith(ref this Collisionmanagement..., Entity entity);
RemoveTemporyCollisionIgnoreWith(ref this Collisionmanagement..., Entity entity);
``` 

I think those methods would make things pretty easy to use, right?

CI09:
I changed it accordingly but there is still this archiving bug I mentioned on server.

src/engine/common_components/CollisionManagement.cs
```diff
@@ -100,6 +107,7 @@ public static CollisionManagement ReadFromArchive(ISArchiveReader reader, ushort
         {
             IgnoredCollisionsWith = reader.ReadObjectOrNull<List<Entity>>(),
             RecordActiveCollisions = reader.ReadInt32(),
+            RemoveIgnoredCollisions = reader.ReadObjectOrNull<List<Entity>>(),
```
hhyyrylainen: If version is less than two this cannot be read, so

`version > 1 ? reader.ReadObjectOrNull<List<Entity>>() : null,`

src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
```diff
@@ -0,0 +1,82 @@
...
+            // Square
+            clipOutDistanceSquared *= clipOutDistanceSquared;
+
+            var difference = worldPosition.Position - otherEntity.Get<WorldPosition>().Position;
+
+            if (difference.LengthSquared() >= clipOutDistanceSquared)
+            {
+                collisionManagement.RemoveTemporaryCollisionIgnoreWith(otherEntity);
+
+                entity.Remove<CellDivisionCollisionDisabler>();
```
hhyyrylainen: This is not safe, this needs to go through the command buffer recording interface.

CI09: Is there an example somewhere in the code?

src/microbe_stage/components/CellDivisionCollisionDisabler.cs
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public ushort CurrentArchiveVersion => SERIALIZATION_VERSION;
+    public ThriveArchiveObjectType ArchiveObjectType => ThriveArchiveObjectType.ComponentCellDivisionCollisionDisabler;
+
+    public void WriteToArchive(ISArchiveWriter writer)
+    {
+        // Save only persistent state
+
+        writer.WriteAnyRegisteredValueAsObject(IgnoredCollisionWith);
```
hhyyrylainen, line 29: You should add writing and reading for the SeparationForce so that it is not forgotten on save load.

src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
```diff
@@ -0,0 +1,82 @@
...
+            }
+
+            if (physics.Body != null)
+            {
+                physicalWorld.GiveImpulse(physics.Body, difference * 300f * collisionDisabler.SeparationForce, true);
```
hhyyrylainen: I'll need to double check, but I think this is probably safe...

Still I kind of feel like it would be better if this went through the Physics component but I won't require that change to be made now.

src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
hhyyrylainen: Look for usages of `StartRecordingEntityCommands`. One example where a new component is added is in `MicrobeShaderParametersHelpers` class.

src/microbe_stage/systems/MicrobeDivisionClippingSystem.cs
```diff
@@ -0,0 +1,91 @@
...
+            if (difference.LengthSquared() >= clipOutDistanceSquared)
+            {
+                collisionManagement.RemoveTemporaryCollisionIgnoreWith(otherEntity);
+
+                RemoveDivisionComponentFromEntity(entity);
+            }
+
+            if (physics.Body != null)
+            {
+                physicalWorld.GiveImpulse(physics.Body, difference * 300.0f * collisionDisabler.SeparationForce, true);
```
hhyyrylainen, line 99: Just in case it is *not* intentional, I'll note here that this *increases* the force as the distance increases. (that could be changed by normalizing the difference).

hhyyrylainen:
So I tested this and this works perfectly for the microbe stage, but in multicellular cells can get stuck.

I'll commit a few nitpicks I still had and a fix for that by basically getting the microbe colony radius to ensure that multicellular stuff doesn't get stuck as that would be quite a bad regression.

CI09:
Thanks. Though I'm thinking whether in the coming progress update should it be mentioned as cell division animation or just initial/basic implementation or something like that not to cause some kind of disappointment if people imagined something more complex like shaders

hhyyrylainen:
I'll try to remember to put it into the BOTD notes as a simple cell reproduction animation or something like that. I won't say it is just an initial implementation as a more complex animation is unlikely to be done (which is why I also closed the related issue after 5 years of no one working on it).

PR #6446, opened 2025-11-11
Macroscopic rotation

CI09:

Small addition, more decorative than anything else.

It's not big but it can wait even after 1.0.0 if there are bigger priorities.

https://github.com/user-attachments/assets/a81259aa-eab2-4c58-9727-8623c7a2e81c

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/macroscopic_stage/MacroscopicCreature.cs
+++ b/src/macroscopic_stage/MacroscopicCreature.cs
@@ -993,7 +993,9 @@ public partial class MacroscopicCreature : RigidBody3D, ICharacterInventory, IEn
         var transform = GlobalTransform;
         var rotation = transform.Basis.GetRotationQuaternion();
 
-        var worldTransform = new Transform3D(new Basis(rotation).Rotated(Vector3.Up, MathF.PI), transform.Origin + rotation * relative);
+        var worldTransform = new Transform3D(new Basis(rotation).Rotated(Vector3.Up, MathF.PI), transform.Origin +
+            rotation * relative);
+
         return worldTransform;
     }
 }


hhyyrylainen:
I'm kind of unsure about this as my original thoughts about macroscopic were that you could freely look with the mouse without causing your species to turn as it might be inconvenient to try to look at various things.

Of course we would need to have some actual macroscopic gameplay to really know how this feels. What does this feel like once on land? Does it cause issues like rotating into the items on the ground or something like that? That's the kind of thing I imagined could be more fiddly if the player creature turned immediately to follow the mouse direction.

CI09:
Now creature should rotate only when there is some horizontal input.
And also the code is using camera forward vector instead of mouse position so it shouldn't cause too much problems.

Accidental-Explorer:
> I'm kind of unsure about this as my original thoughts about macroscopic were that you could freely look with the mouse without causing your species to turn as it might be inconvenient to try to look at various things.
> 
> Of course we would need to have some actual macroscopic gameplay to really know how this feels. What does this feel like once on land? Does it cause issues like rotating into the items on the ground or something like that? That's the kind of thing I imagined could be more fiddly if the player creature turned immediately to follow the mouse direction.

> Now creature should rotate only when there is some horizontal input. And also the code is using camera forward vector instead of mouse position so it shouldn't cause too much problems.

Could also require for example right-click (or hold) to orient creature front towards camera/cursor. (or the opposite, a button press to de-couple them while held)

But I think this is a pretty fundamental game design question that requires its own discussion. (I don't see much trouble with a temporary solution though).

CI09:
I think yeah, right click would make it feel more controllable (doesn't require input anymore though)

hhyyrylainen:
I think that's too undiscoverable now. I'd be fine with just merging the alternative that rotates when moving in general. The input change would require bindings to the added to the input configuration etc. and probably a popup text in the HUD to explain that the feature exists, because otherwise no one will ue it.

CI09:
Only the structure was rotated the wrong way (Godot defines forward vector as negative Z, creatures are also implemented that way). Other stages are ok.

PR #6452, opened 2025-11-12
Auto evo reproduction cost

Accidental-Explorer:

This PR adds a new selection pressure type for the collection and usage of "reproduction compounds": compounds that need to be collected to reproduce, but do not provide energy themselves. Takes into account active and passive gathering, and the energy spent while doing those. 

Deliberately only added to the tree for "autotrophs", because it is assumed that heterotrophs get these from prey.

Clouds and chunks are deliberately combined into one, because right now we don't want to split the miche tree on these miches, creating more slots for species.

Also includes score calculation and mutation adding of organelles producing the compound, allowing nitrogenase and nitroplasts to be added.

**Related Issues**

The most important factors here adding value to nitrogenase/nitroplast, and chemoreceptors aimed at phosphate/ammonia.

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/general/GameWorld.cs
```diff
@@ -74,6 +74,7 @@ public GameWorld(WorldGenerationSettings settings, Species? startingSpecies = nu
             // the full effect to get balanced well enough
             TimedEffects.RegisterEffect("photosynthesis_production", new PhotosynthesisProductionEffect(this));
             TimedEffects.RegisterEffect("volcanism", new VolcanismEffect(this));
+            TimedEffects.RegisterEffect("ammonia_production", new AmmoniaProductionEffect(this));
```
hhyyrylainen, line 77: This will need a save version upgrade for the GameWorld and adding this for old saves that are loaded.
For example, such a thing was done in this PR: https://github.com/Revolutionary-Games/Thrive/pull/6377

src/saving/ThriveArchiveObjectType.cs
```diff
@@ -304,6 +304,8 @@ public enum ThriveArchiveObjectType : uint
     CurrentDilutionEvent = 4387,
     PatchEventsManager = 4388,
     PatchEventProperties = 4389,
+    ReproductionCompoundPressure = 4390,
+    AmmoniaProductionEffect = 4391,
 
     ComponentCellDivisionCollisionDisabler = 4385,
```
hhyyrylainen: Oh I just noticed that the IDs are messed up. Well at least we didn't release like this.

I'm doing again a big refactor so I don't want to switch to master branch just now to fix the numbering to be unique. But the new values added in this PR shouldn't be put here in the middle like this...

src/general/world_effects/AmmoniaProductionEffect.cs
```diff
@@ -0,0 +1,147 @@
...
+                if (species.Key is MicrobeSpecies microbeSpecies)
+                {
+                    resolvedTolerances = MicrobeEnvironmentalToleranceCalculations.ResolveToleranceValues(
+                        MicrobeEnvironmentalToleranceCalculations.CalculateTolerances(microbeSpecies, patch.Biome));
+                }
+
+                var balanceModifier = ProcessSystem.CalculateSpeciesActiveProcessListForEffect(species.Key,
+                    microbeProcesses, patch.Biome, resolvedTolerances, targetWorld.WorldSettings);
+
+                // Ammonia consumption of this speciesbased on reproduction cost of all individuals
```
hhyyrylainen:
```suggestion
                // Ammonia consumption of this species based on reproduction cost of all individuals
```

src/auto-evo/steps/GenerateMiche.cs
```diff
@@ -46,7 +51,7 @@ public Miche GenerateMicheTree(AutoEvoGlobalCache globalCache)
             var glucoseMiche = new Miche(globalCache.GlucoseConversionEfficiencyPressure);
             glucoseMiche.AddChild(new Miche(globalCache.GlucoseCloudPressure));
 
-            generatedMiche.AddChild(glucoseMiche);
+            ammoniaMiche.AddChild(glucoseMiche);
```
hhyyrylainen: Rather than renaming all of these specifically to be children of "ammonia", I think it would be better to introduce a new variable name so that inserting new miches at the top no longer requires renaming everything.

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    public override float GetEnergy(Patch patch)
+    {
+        return 0;
+    }
+
+    public override LocalizedString GetDescription()
+    {
+        return new LocalizedString("COMPOUND_FOOD_SOURCE",
```
hhyyrylainen: Is this the right translation key here?

src/general/GameWorld.cs
Accidental-Explorer: Okay, I will look into that.

src/saving/ThriveArchiveObjectType.cs
Accidental-Explorer: You're referring to `ComponentCellDivisionCollisionDisabler` not having a unique number, correct? It was weirdly separated from the rest of the list so I ignored it. I can indeed fix that in this PU.

> But the new values added in this PR shouldn't be put here in the middle like this...

So, move the new values under `ComponentCellDivisionCollisionDisabler`, give it a new ID number and move the other two number up one digit?

src/auto-evo/steps/GenerateMiche.cs
Accidental-Explorer: So, something like "lastGeneralAutotrophMiche" filled with whatever is currently designed to be the last one on that general branch?

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
Accidental-Explorer: Hmmm, I left that the same as in CompoundCloudPressure since it should be the exact same, but it might be cleaner to separate it to be sure?

src/saving/ThriveArchiveObjectType.cs
hhyyrylainen: Yes, a full fix would be that ComponentCellDivisionCollisionDisabler is placed on line 307 and its number updated to be what it should be, and then re-add the     ReproductionCompoundPressure = 4390, and
    AmmoniaProductionEffect = 4391, but update the numbers again to be unique.

src/auto-evo/steps/GenerateMiche.cs
hhyyrylainen: I think `lastGeneralMiche` would be fine as we don't need to go overboard with how long the variable name is.

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
hhyyrylainen: This is not really a food source, and I saw you added a new translation key anyway so I thought that that key would be meant to go here.

Accidental-Explorer: Done this, I just need to do the update from master and then re-translate trick when it's time to merge.

src/auto-evo/steps/GenerateMiche.cs
Accidental-Explorer: Alright, this adjustment was made.

src/general/GameWorld.cs
Accidental-Explorer: I think the changes I pushed should have resolved it.

src/general/GameWorld.cs
```diff
@@ -317,6 +318,11 @@ public static GameWorld ReadFromArchive(ISArchiveReader reader, ushort version,
                 new PatchEventsManager(instance, random.Next64()));
         }
 
+        if (version < 3)
+        {
+            instance.TimedEffects.RegisterEffect("ammonia_production", new AmmoniaProductionEffect(instance));
```
hhyyrylainen: This has a slight problem, @Patryk26g specified that the patch_events_manager must always be last to work correctly. So we need some way here to insert this new effect to be the second last effect and not the last.

src/general/world_effects/AmmoniaProductionEffect.cs
```diff
@@ -0,0 +1,147 @@
...
+        foreach (var patchKeyValue in targetWorld.Map.Patches)
+        {
+            var patch = patchKeyValue.Value;
+
+            // Skip empty patches as they don't need handling
+            if (patch.SpeciesInPatch.Count < 1)
+                continue;
+
+            // Skip patches that don't need handling
+            if (!patch.Biome.ChangeableCompounds.TryGetValue(Compound.Ammonia,
```
hhyyrylainen: I don't think this effect should skip any patches, ammonia is a key compound in every patch, and even if missing species should be able to bring it along them.

src/general/world_effects/AmmoniaProductionEffect.cs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    public void OnTimePassed(double elapsed, double totalTimePassed)
+    {
+        ApplyCompoundsAddition();
+    }
+
+    private void ApplyCompoundsAddition()
+    {
+        List<TweakedProcess> microbeProcesses = [];
```
hhyyrylainen: This is a good start on memory efficiency, however I think this should be a `private readonly` field to allocate even less instances.

src/general/world_effects/AmmoniaProductionEffect.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            ammoniaBalance *= Constants.AMMONIA_ENVIRONMENT_SPEED_MULTIPLIER;
+
+            ammonia.Density += ammoniaBalance;
+
+            patch.Biome.ModifyLongTermCondition(Compound.Ammonia, ammonia);
+        }
+    }
+
+    private bool IsProcessRelevant(BioProcess process)
```
hhyyrylainen, line 136: This should be marked for aggressive inlining.

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
```diff
@@ -0,0 +1,191 @@
...
+            new UpgradeOrganelle(organelle => organelle.HasChemoreceptorComponent,
+                new ChemoreceptorUpgrades(compound, null, Constants.CHEMORECEPTOR_RANGE_DEFAULT,
+                    Constants.CHEMORECEPTOR_AMOUNT_DEFAULT, SimulationParameters.GetCompound(compound).Colour)),
+            new ChangeMembraneType("single"),
+        ])
+    {
+        compoundDefinition = SimulationParameters.GetCompound(compound);
+
+        if (!compoundDefinition.IsCloud)
+            throw new ArgumentException("Given compound to cloud pressure is not of cloud type");
```
hhyyrylainen: This seems like a non-updated string for this new context.

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        this.compound = compound;
+        this.isDayNightCycleEnabled = isDayNightCycleEnabled;
+    }
+
+    public override LocalizedString Name => NameString;
+
+    public override ushort CurrentArchiveVersion => SERIALIZATION_VERSION;
+
+    // TODO: change this and add to ThriveArchiveObjectType (probably the new saving system?)
```
hhyyrylainen: There's a TODO here...

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
```diff
@@ -0,0 +1,191 @@
...
+            multiplier = Math.Max(multiplier, Constants.AUTO_EVO_MAX_NIGHT_SESSILITY_COLLECTING_PENALTY);
+
+            if (multiplier <= 1)
+                score *= multiplier;
+        }
+
+        var chemoreceptorScore = cache.GetChemoreceptorCloudScore(microbeSpecies, compoundDefinition, patch.Biome);
+        score += chemoreceptorScore;
+
+        // Combine with compound from all chunks
```
hhyyrylainen:
```suggestion
        // Combine with compounds from all chunks
```
?

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
```diff
@@ -0,0 +1,191 @@
...
+                if (!microbeSpecies.CanEngulf ||
+                    cache.GetBaseHexSizeForSpecies(microbeSpecies) < chunk.Value.Size * Constants.ENGULF_SIZE_RATIO_REQ)
+                {
+                    chunkScore *= Constants.AUTO_EVO_CHUNK_LEAK_MULTIPLIER;
+                }
+
+                chemoreceptorScore += chunkChemoreceptorScore;
+                score += chunkScore;
+
+                if (chunk.Value.Compounds?.TryGetValue(compoundDefinition.ID, out var chunkCompoundAmount) != true)
```
hhyyrylainen: Isn't the null suppression unnecessary here? There's a null check on line 107.

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
```diff
@@ -0,0 +1,191 @@
...
+                compoundAmount += ventedCompound;
+            }
+        }
+
+        // modify score by how much compound is available for collection
+        score *= compoundAmount;
+        chemoreceptorScore *= compoundAmount;
+
+        // Organelles that produce this compound
+        foreach (var organelle in microbeSpecies.Organelles.Organelles)
```
hhyyrylainen: This is not a really accurate way to check active processes, so this needs to be reworked to calculate the actual active processes. For example some upgrades override which processes an organelle provides.

I can't remember immediately if there's an example where this is done in auto-evo.

src/general/GameWorld.cs
Accidental-Explorer: I am guessing just adding it again at the end won't conveniently overwrite the previous one?

Having two of them seems especially unwise, so then we need a way delete the other one. 

Or maybe remove it from the starting list and always add it at the end? (That would break save compatibility again though, even if preventing it later)

src/general/world_effects/AmmoniaProductionEffect.cs
Accidental-Explorer: That was for future-proofing in case some worlds didn't handle Ammonia at all (for alternate forms of life), but thinking back, you're right that's unnecessary to consider.

src/general/GameWorld.cs
hhyyrylainen: > I am guessing just adding it again at the end won't conveniently overwrite the previous one?

It most likely does not.

My initial hunch is that adding a new bool parameter to the register effect and it's like "bool last = true" and then in this case it would be set to false and when done so the code in the register effect uses insert with length - 2 instead of Add for the item.

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
Accidental-Explorer: Ah, forgot to remove the TODO after I did what it was reminding me to do...

src/general/GameWorld.cs
Patryk26g: Well, my comment might be a little bit misleading, the Patch Event Manager has to be called after the patch events that change temperature or light levels. This here should be safe in my opinion. Although I didn't have time to look at this PR yet fully

src/general/world_effects/AmmoniaProductionEffect.cs
Accidental-Explorer: So, create a private readonly list outside the void, and just clear it in the place where it is now being created? I think I can make the same change in PhotosynthesisProductionEffect and HydrogenSulfideConsumptionEffect?

src/general/GameWorld.cs
Patryk26g: I would say that it might make difference wether it is called before or after volatility events but it shouldn't have to big impact on gameplay anyway

src/general/world_effects/AmmoniaProductionEffect.cs
hhyyrylainen: > outside the void

If you mean, outside the function, then yes that's what I meant. Storing the list instance in a field will mean that it can be reused across generations.

src/general/GameWorld.cs
hhyyrylainen: For clarity I think we should stick to it always being the last one. So that there's no need to think "is it safe for this effect type to be before the manager" in coding situations.

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
Accidental-Explorer: It's still the _one_ compound, just from different sources. But I hopefully clarified the comment.

hhyyrylainen: Ah, so it was supposed to be "with the compound"

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
Accidental-Explorer: Right now I only remember examples where it's done the wrong way, like in the calculation of toxin scores in SimulationCache.cs (the part that existed _before_ I touched it, that is).

I think that this has not been exposed in an easy way in OrganelleTemplate yet as RunnableProcesses is in OrganelleDefinition.

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
Accidental-Explorer: Indeed it is, rewrote it a bit.

But now I am wondering if it's worth not assuming that the reproduction compound has a cloud form and instead checking for that in the score method.

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
Accidental-Explorer: Indeed it is, removed now.

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
hhyyrylainen: For world effects there's already code to calculate the active processes list, and you've used that in this PR.
Might be worth it to add that to the microbe data cache so that it can be used without allocating totally temporary memory.

src/general/world_effects/AmmoniaProductionEffect.cs
Accidental-Explorer: Has been done (if I did it correctly).

src/saving/ThriveArchiveObjectType.cs
Accidental-Explorer: This part was resolved at least. Hopefully nothing else with new values needs to be merged before this is finished.

src/general/world_effects/AmmoniaProductionEffect.cs
hhyyrylainen: Yes, that is correct.

src/saving/ThriveArchiveObjectType.cs
hhyyrylainen: I really quickly wrote a test against this on my branch (as I was working on other unit tests anyway), and I basically had to fix this also there, but I think I fixed the one existing problematic value the same way as you so that should be pretty painless to merge, even if there is a merge conflict (for some reason).

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
Accidental-Explorer: I'll see if I can get this done today.

 I want to say in advance: looking only at the Organelle.Definition.RunnableProcesses is currently used everywhere from CompoundConversionPressure to MaintainCompoundPressure. So fixing this properly (rather than only for the new ReproductionCompoundPressure) is going to be a fairly wide-reaching change mechanically.

> Might be worth it to add that to the microbe data cache so that it can be used without allocating totally temporary memory.

By this do you just mean in SimulationCache per patch (including the environmental effects) or something else?

hhyyrylainen: > By this do you just mean in SimulationCache per patch (including the environmental effects) or something else?

No, that seems unnecessary. Just literally just passing in the microbe species to the cache should be enough to resolve the active processes, right?

And the change won't be that big by just needing a few places to call a new method and process the data it returns with the existing approach mostly.

Accidental-Explorer: > No, that seems unnecessary. Just literally just passing in the microbe species to the cache should be enough to resolve the active processes, right?

I guess I was also looking at correctly scaling processes based on evironmental equally-distributed compounds, but now that I look again I am not sure whether the world effects do that either (they do look at environmental tolerance effects though). 

Also that was my way of asking literally what you meant by "the microbe data cache", because I am not familiar with that.

Accidental-Explorer: Okay, I made GetActiveProcessList in SimulationCache, and changed this in ReproductionCompoundUsagePressure.

I'll see your opinion on this first, and if its fine I'll make the same change elsewhere.

hhyyrylainen: The usage of the new cache method seems good to me, but internally I think it could be optimized to not pointlessly use an extra list just to copy stuff into a new list.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -145,6 +147,7 @@ public float GetCompoundConversionScoreForSpecies(CompoundDefinition fromCompoun
         MicrobeSpecies species, BiomeConditions biomeConditions)
     {
         var key = (species, biomeConditions, fromCompound, toCompound);
+        var activeProcessList = GetActiveProcessList(species);
```
hhyyrylainen: This is unused when the thing is found in cache so you should move this variable initialization closer to where it is used.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -732,11 +744,18 @@ public PredationToolsRawScores GetPredationToolsRawScores(MicrobeSpecies microbe
 
                     totalToxicity += organelle.GetActiveToxicity();
                     totalToxinOrganellesCount += 1;
-                    totalToxinScore += toxinAmount * Constants.AUTO_EVO_TOXIN_PREDATION_SCORE;
                 }
             }
         }
 
+        foreach (var process in activeProcessList)
```
hhyyrylainen: This change seems a bit pointless to me as this decouples the toxin types from the amount, which could lead to very strange bugs. I'd say that either this whole function needs to use the new approach *or* the entire method should keep using the old approach.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -624,6 +617,23 @@ public void Clear()
...
+        if (cachedProcessLists.TryGetValue(microbeSpecies, out var cached))
+        {
+            return cached;
+        }
+
+        var microbeProcesses = new List<TweakedProcess>();
+        ProcessSystem.ComputeActiveProcessList(microbeSpecies.Organelles, ref microbeProcesses);
+
+        cached = new List<TweakedProcess>(microbeProcesses);
+        cachedProcessLists.Add(microbeSpecies, cached);
```
hhyyrylainen:
```suggestion
        var microbeProcesses = new List<TweakedProcess>();
        ProcessSystem.ComputeActiveProcessList(microbeSpecies.Organelles, ref microbeProcesses);

        cached = microbeProcesses;
        cachedProcessLists.Add(microbeSpecies, cached);
```
You are still making a double list, it should be like this

Or I think you could even just use

```
        ProcessSystem.ComputeActiveProcessList(microbeSpecies.Organelles, ref cached);

        cachedProcessLists.Add(microbeSpecies, cached);
```

Accidental-Explorer: Aha, I see now!

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: The toxin organelles are as far as I know the one place in the game where the upgrades _actually_ have different processes than base organelle, so this makes a functional difference. 
Meanwhile toxin type and especially toxicity are only stored in the organelle's ToxinUpgrades. (The former you _could_ get from reading the name of the process).

I have no particular problem with going back to the previous situation, but that would remove the "benefit" of auto-evo using the correct processes.

hhyyrylainen: Even if explained like that, I don't like how the logic can now get totally detached, like imagine a bug where some upgrade or something gets counted as a specific toxin type, but the upgrade's toxicity doesn't apply. Those are the nightmare bugs I want to avoid by ensuring there aren't two codepaths for basically doing the same thing.

Accidental-Explorer: Valid. I reverted the change for now. If I want to alter this again in the future I will find a better solution.

src/general/GameWorld.cs
Accidental-Explorer: This should be fixed now if I understood correctly.

PR #6461, opened 2025-11-14
Population visualization

Patryk26g:

Adds population indicators to patch maps based on @CI09 work

image

closes #5880

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/PatchMapDrawer.cs
+++ b/src/microbe_stage/editor/PatchMapDrawer.cs
@@ -1117,7 +1117,9 @@ public partial class PatchMapDrawer : Control
         {
             var playerPopulation = patch.GetSpeciesSimulationPopulation(playerSpecies);
 
-            for (var i = 0; i < MathF.Sqrt(playerPopulation) * Constants.INDICATORS_NUMBER_PER_POPULATION_SQUARED; ++i)
+            var numberOfIndicators = MathF.Sqrt(playerPopulation) *
+                Constants.INDICATORS_NUMBER_PER_POPULATION_SQUARED;
+            for (var i = 0; i < numberOfIndicators; ++i)
             {
                 var lifeIndicator = new TextureRect
                 {


Patryk26g:
Should I add something to the tutorial text about the population dots?

hhyyrylainen:
The two existing patch map tutorials already have so much text that I wouldn't really consider adding more. Hopefully the feature is self-explanatory enough. Or if you want you could add a new tutorial that triggers on like generation 5-6 on the patch map screen and explains the population indicators.

src/microbe_stage/editor/MicrobeEditor.cs
```diff
@@ -342,6 +342,7 @@ protected override void OnEditorReady()
         reportTab.UpdateEvents(CurrentGame.GameWorld.EventsLog, CurrentGame.GameWorld.TotalPassedTime);
 
         patchMapTab.UpdatePatchEvents();
+        patchMapTab.MarkDrawerDirty();
```
hhyyrylainen, line 345: Does this not need to trigger specially when loading a save? There's logic for triggering the event apply on loading a save, but it looks like you didn't add it specifically for this?

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1100,10 +1112,43 @@ private void AddPatchNode(Patch patch, Vector2 position)
 
         patch.ApplyPatchEventVisuals(node);
 
+        var playerSpecies = patch.FindSpeciesByID(PlayerSpeciesID);
+        if (playerSpecies != null)
+        {
+            var playerPopulation = patch.GetSpeciesSimulationPopulation(playerSpecies);
+
+            for (var i = 0; i < MathF.Sqrt(playerPopulation) * Constants.INDICATORS_NUMBER_PER_POPULATION_SQUARED; ++i)
```
hhyyrylainen: Unless I'm misremembering C# recalculates the loop check condition each time, so for each `i` value this is taking a square root which is about the slowest possible math primitive. So I think the loop end condition should be stored in a variable.

adQuid (approved):
Nothing concerns me here...

Patryk26g:
> The two existing patch map tutorials already have so much text that I wouldn't really consider adding more. Hopefully the feature is self-explanatory enough. Or if you want you could add a new tutorial that triggers on like generation 5-6 on the patch map screen and explains the population indicators.

I think I'll leave it as it is then

src/microbe_stage/editor/MicrobeEditor.cs
Patryk26g: I do not quite understand what you mean :(

hhyyrylainen: When loading a save `OnEditorReady` is not called. So whatever this new line of code is meant to do, gets totally skipped when loading a save made in the editor.

Patryk26g: Ahhh okay, but when a save is load then Init is called on the PatchMapDrawer, isn't it? So that solves the issue. This line of code is meant to update the map when going from editor -> game world -> editor, as without it the population wasn't updated - it was updated only when the patch was changed as that triggered the redraw

PR #6465, opened 2025-11-15
Make it possible to kill engulfer as a poisonous player cell

adQuid:

Quite a few players have been disappointed over the years to find out that current game mechanics make it impractical to survive being engulfed, even if you have a lot of toxin. This PR makes is so that you can get up to 80% engulfed before being auto-killed when released, and holds off the "mercy timer" for auto-killing an engulfed player until they don't have any more toxin.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/EngulfedHandlingSystem.cs
+++ b/src/microbe_stage/systems/EngulfedHandlingSystem.cs
@@ -93,7 +93,7 @@ public partial class EngulfedHandlingSystem : BaseSystem<World, float>
             if (engulfable.PhagocytosisStep == PhagocytosisPhase.Ingested && entity.Has<PlayerMarker>())
             {
                 // If the player still has toxin, there's a chance the player cell can escape
-                if (entity.Get<CompoundStorage>().Compounds.GetCompoundAmount(Compound.Oxytoxy) < 0.1)
+                if (entity.Get<CompoundStorage>().Compounds.GetCompoundAmount(Compound.Oxytoxy) < 0.1f)
                 {
                     playerEngulfedDeathTimer += delta;
                 }


simulation_parameters/Constants.cs
```diff
@@ -942,7 +942,7 @@ public static class Constants
     ///   The measure of which beyond this threshold an engulfable is considered partially digested.
     ///   Used to determine whether a cell should be able to heal after being expelled from engulfment.
     /// </summary>
-    public const float PARTIALLY_DIGESTED_THRESHOLD = 0.5f;
+    public const float PARTIALLY_DIGESTED_THRESHOLD = 0.8f;
```
hhyyrylainen, line 945: I'll note that this impacts *all* cells, not just the player, though I suppose nobody will really notice that this threshold was increased for everything.

src/microbe_stage/systems/EngulfedHandlingSystem.cs
```diff
@@ -91,7 +91,13 @@ private void Update([Data] in float delta, ref Engulfable engulfable, ref Microb
             // Species handling for the player microbe in case the process into partial digestion took too long
             // so here we want to limit how long the player should wait until they respawn
             if (engulfable.PhagocytosisStep == PhagocytosisPhase.Ingested && entity.Has<PlayerMarker>())
-                playerEngulfedDeathTimer += delta;
+            {
+                // If the player still has toxin, there's a chance the player cell can escape
+                if (entity.Get<CompoundStorage>().Compounds.GetCompoundAmount(Compound.Oxytoxy) == 0)
```
hhyyrylainen: Doesn't this completely get blocked if the player can produce toxin or has some leftover amount? For safety I would write this check as `>= 0.001f` or so.

adQuid: I didn't think about toxin production; I see the value in adding a small threshold value.

adQuid: I added a bigger threshold at 0.1, since that wouldn't be enough to kill almost anything.

PR #6466, opened 2025-11-15
Add Panspermia Intro Cutscene

### LINKED ISSUE #4170: Create variant of the intro video for panspermia start
As the vents *really* don't represent life arriving with a meteor. This is related to: https://github.com/Revolutionary-Games/Thrive/issues/4169

If the intros can't be made consistent also the vents intro video can be remade.

adQuid:

If the player chooses the "Panspermia" option for life origin, reward them for their good taste with a new cutscene.

**Related Issues**

closes #4170

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/Thrive.csproj
+++ b/Thrive.csproj
@@ -40,7 +40,7 @@
   </ItemGroup>
   <!-- Test resources -->
   <ItemGroup Condition="'$(Configuration)' != 'ExportRelease'">
-    <PackageReference Include="gdUnit4.api" Version="5.1.0-rc1" />
+    <PackageReference Include="gdUnit4.api" Version="5.1.0-rc3" />
     <PackageReference Include="gdUnit4.test.adapter" Version="3.0.0" />
     <!-- Re-enable this this once newer compiler versions are deployed everywhere
     <PackageReference Include="gdUnit4.analyzers" Version="1.0.0">
--- a/assets/sounds/soundeffects/engulfment.ogg
+++ b/assets/sounds/soundeffects/engulfment.ogg
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:5e39f629583d8855e4a5a17f45863d6920ef7727872ae9b6a1b6e5db79388617
-size 37752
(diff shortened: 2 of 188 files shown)


Deus-Codes:
It works!

Will note that the cutscenes being a bit low resolution might make certain parts a bit unclear for players who have no clue what is going on, but that isn't an issue with this specific PR. It could also be exaggerated by me having a larger monitor. Great work for everyone involved otherwise

hhyyrylainen:
I think the video could be made a bit better. It is 50 FPS, so it doesn't really smoothly map to vsync. I think it should either be 30 or 60 FPS, and not 50. Also it does look a bit blurry. Maybe doubling the filesize for it would be good? I noticed it is less than half of the initial intro video so I think it could be like up to 15 MB in size just fine.

If anyone has the original video I can try doing re-encoding myself as well.

adQuid:
> I think the video could be made a bit better. It is 50 FPS, so it doesn't really smoothly map to vsync. I think it should either be 30 or 60 FPS, and not 50. Also it does look a bit blurry. Maybe doubling the filesize for it would be good? I noticed it is less than half of the initial intro video so I think it could be like up to 15 MB in size just fine.
> 
> If anyone has the original video I can try doing re-encoding myself as well.

I will point you to the video in Discord, to avoid linking someone else's Google Doc on a public channel. If you're talking about taking the MP4 and reconverting it to OGV to change the framerate that makes sense to me, but as for the video quality I'm pretty sure nothing was dropped from the original we have.

hhyyrylainen (approved):
I did a re-encode that preserves detail a lot better in many parts of the video (still there are very blocky parts like the zoom to the ocean, but I think those are due to how noisy the original video is so no reasonable amount of bitrate can really save it). The video now takes up slightly more than the other intro video, which I think is fine to have the video not be super blurry at the start which would give a bad impression.

I think this can be merged now, but re-rendering the original video with less noise could make this even better but I guess we'll see if that happens or if this should be merged already.

hhyyrylainen:
I updated the video file again, it should now be a re-encode of the best quality intro video we can get for now.

PR #6468, opened 2025-11-16
Make Instant Kill Protection Toggleable as a Difficulty Option

adQuid:

Some people want to live dangerously. 

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)
(NOTE: I can't test save compatibility directly since I don't have a save with all three difficulty versions)

### DIFF
--- a/locale/messages.pot
+++ b/locale/messages.pot
@@ -85,15 +85,9 @@ msgid "ACTION_AWAKEN_TOOLTIP"
 msgstr ""
 
 #: ../src/general/base_stage/EditorBase.cs:910
-<<<<<<< HEAD
-#: ../src/macroscopic_stage/editor/MacroscopicEditor.cs:494
-#: ../src/microbe_stage/editor/CellEditorComponent.GUI.cs:62
-#: ../src/multicellular_stage/editor/MulticellularEditor.cs:468
-=======
 #: ../src/macroscopic_stage/editor/MacroscopicEditor.cs:495
 #: ../src/microbe_stage/editor/CellEditorComponent.GUI.cs:62
 #: ../src/multicellular_stage/editor/MulticellularEditor.cs:469
->>>>>>> master
 msgid "ACTION_BLOCKED_WHILE_ANOTHER_IN_PROGRESS"
 msgstr ""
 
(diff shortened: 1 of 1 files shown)


locale/en.po
```diff
@@ -3799,7 +3805,7 @@ msgstr ""
 
 msgid "MICROBE_STAGE_INITIAL_PANSPERMIA"
 msgstr ""
-"On a distant alien planet, a meteor has impacted after travelling vast distances, carrying within it the seeds of a new phenomenon on the planet.\n"
+"On a distant alien planet, meteor impacts are common, but this rock from far away happened to be carrying something very special.\n"
```
hhyyrylainen: I think you accidentally copied the text from your other PR here.

simulation_parameters/common/difficulty_presets.json
```diff
@@ -54,6 +56,7 @@
     "SwitchSpeciesOnExtinction": false,
     "LimitGrowthRate": true,
     "FogOfWarMode": "Regular",
+    "InstantKillProtection": false,
```
hhyyrylainen: I'd say that this is kind of a core gameplay experience thing so at least for now I think it should be on for all difficulties, and players can turn it off if they decide to go for custom difficulty.

src/general/CustomDifficulty.cs
```diff
@@ -82,6 +85,11 @@ public static CustomDifficulty ReadFromArchive(ISArchiveReader reader, ushort ve
             instance.limitGrowthRate = reader.ReadBool();
         }
 
+        if (version > 2)
+        {
+            instance.InstantKillProtection = reader.ReadBool();
+        }
```
hhyyrylainen, line 91: Else instant kill protection should default to *on* rather than off.

src/engine/common_components/Health.cs
```diff
@@ -86,6 +86,13 @@ public void WriteToArchive(ISArchiveWriter writer)
 
 public static class HealthHelpers
 {
+    private static IDifficulty? difficulty;
```
hhyyrylainen: I don't particularly like this being like this... It is the smallest change to make health be affected by current difficulty (I admit). A proper design would be quite a bit more complex and I can't really come up with it off the top of my head.

So here's my compromise: turn this into just a bool that is on by default, and modify the set and check so that they just store the boolean (that should default to true) and then just read it.

hhyyrylainen:
This is that follow up issue: https://github.com/Revolutionary-Games/Thrive/issues/6477

PR #6478, opened 2025-11-19
Auto evo turning speed and cilia

### LINKED ISSUE #3351: Auto-evo should take microbe turn rate into account
with the cilia microbes have different turn speeds so maybe the predation calculations should take it into account in maneuverability to make auto-evo actually add cilia

Accidental-Explorer:

This PR:

- Makes Predation score take into account rotation speed, approximating prey getting away before the predator can turn to catch them. 
- Also adds Cilia mutation adding to PredationSelectionPressure (and AvoidPredation) so that species can improve their turning speed.
- Pulling Cilia give a score modifier to engulfment, stronger for the "accidental" type of predation.
- Also adds upgrading of cilia to PredationSelectionPressure.

**Related Issues**

closes #3351

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -378,7 +396,9 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
             // ... but you may also catch them by luck (e.g. when they run into you),
             // and this is especially easy if you're huge.
             // This is also used to incentivize size in microbe species.
-            catchScore += Constants.AUTO_EVO_ENGULF_LUCKY_CATCH_PROBABILITY * predatorHexSize;
+            // Prey that can't turn away fast enough are more likely to get caught.
+            catchScore += Constants.AUTO_EVO_ENGULF_LUCKY_CATCH_PROBABILITY * predatorHexSize *
+                (1.0f + preyRotationSpeed);
```
Patryk26g: Doesn't this code mean that with bigger prey rotation speed the predator will have higher catchScore?

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -409,10 +429,13 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
             }
             else
             {
-                pilusScore *= Constants.AUTO_EVO_ENGULF_LUCKY_CATCH_PROBABILITY;
+                pilusScore *= Constants.AUTO_EVO_ENGULF_LUCKY_CATCH_PROBABILITY * (1.0f + preyRotationSpeed);
             }
         }
 
+        // Pili are also more useful if you can turn them towards the target in time
+        pilusScore *= 1.0f - predatorRotationSpeed / 2;
```
Patryk26g: Higher predatorRotationSpeed => lower pilusScore?
and higher preyRotationSpeed => higher pilusScore? Shouldn't it be the other way around?

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -458,6 +481,9 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
         // Prey that resist toxin are of course less vulnerable to being hunted with it
         damagingToxinScore /= prey.MembraneType.ToxinResistance;
 
+        // Toxins also require facing and tracking the target
+        damagingToxinScore *= 1.0f - predatorRotationSpeed / 2;
```
Patryk26g: Same case here

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: Rotation speed is an odd case, the lower the number, the faster you turn. The UI flips it around for the player to see.

The same of course applies to your other two comments.

Patryk26g: Ahhh, that explains a lot then! Some time ago i tried to make auto-evo be able to evolve cilia, the inverted rotation value is probably why I failed lmao

hhyyrylainen: The inverted rotation rate value kind of makes sense in context, but if you don't know the context it certainly is weird.

The context being that cell rotation (back in the day) used a slerp to smoothly rotate. And to do that it used a base rate, and then when different rotation speeds was added (to not blow up the slerp approach) the slerp factor was simply updated to be `constant / speed` so that there's a wide range of possible speed values that are usable. And that causes the "inverted" speed value (bigger means slower slerp).

Accidental-Explorer: I can rename the usage in this algorithm to "slowness" if that helps. Or just add a comment on it. I think I saw a comment like that closer to the root of where rotation speed originates in the code. But obviously you can't tell in this context.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -651,6 +688,7 @@ public PredationToolsRawScores GetPredationToolsRawScores(MicrobeSpecies microbe
         var injectisomeScore = Constants.AUTO_EVO_PILUS_PREDATION_SCORE;
         var slimeJetScore = Constants.AUTO_EVO_SLIME_JET_SCORE;
         var mucocystsScore = Constants.AUTO_EVO_MUCOCYST_SCORE;
+        var pullingCiliaModifier = 1.0f;
```
Patryk26g, line 699: I would initialize it with Constants.AUTO_EVO_PULL_CILIA_MODIFIER and then remove this constant from the calculations down below

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -780,9 +829,10 @@ public PredationToolsRawScores GetPredationToolsRawScores(MicrobeSpecies microbe
                     ToxinType.OxygenMetabolismInhibitor);
         }
 
-        // Having lots of extra pili, slime jets and mucocysts doesn't really help much
+        // Having lots of slime jets, mucocysts and pulling cilias doesn't really help much
         slimeJetScore *= MathF.Sqrt(slimeJetsCount);
         mucocystsScore *= MathF.Sqrt(mucocystsCount);
+        pullingCiliaModifier *= 1 + MathF.Sqrt(pullingCiliasCount) * Constants.AUTO_EVO_PULL_CILIA_MODIFIER;
```
Patryk26g, line 843: I would rename it to pullingCiliaScore and just do "*=MathF.Sqrt(pullingCiliasCount)" without the "1 +" for consistency. If you want bigger score than you can increase the Constants.AUTO_EVO_PULL_CILIA_MODIFIER value :)

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: That definitely would not be the same thing. The below section does 
1 **+** (count ***** constant).

Your proposed change would make it:
count * constant
It scales entirely differently, and the later code always uses this number because it is a multiplication modifier, not a score that is added.

src/auto-evo/simulation/SimulationCache.cs
Accidental-Explorer: Did you see the usage in PredationScore?

It's not a score that gets added in along with others, it's a modifier that multiplies other scores.

I would just have to reverse the operation and convert it into the intended 1.something modifier that it's supposed to be in GetPredationScore.

Patryk26g: oh okay must have missed that

hhyyrylainen:
Is this good now and ready for final reviews to put this into 0.9.2?

Accidental-Explorer:
> Is this good now and ready for final reviews to put this into 0.9.2?

Yes, I think it should be.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -296,12 +312,20 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
...
         var predatorSlimeJetScore = predatorToolScores.SlimeJetScore;
+        var pullingCiliaModifier = predatorToolScores.PullingCiliaModifier;
+        var strongPullingCiliaModifier = pullingCiliaModifier * pullingCiliaModifier;
 
         var preySlimeJetScore = preyToolScores.SlimeJetScore;
         var preyMucocystsScore = preyToolScores.MucocystsScore;
 
         var behaviourScore = predator.Behaviour.Aggression / Constants.MAX_SPECIES_AGGRESSION;
 
+        // This makes rotation "speed" not matter until the editor shows ~300,
```
hhyyrylainen, line 323: I think there needs to be one clarification that the editor GUI shows the rotation non-reversed so there higher values *do* mean faster rotation.

Other than this very minor thing I think this PR is ready to go. Of course with the caveat that I'm still no auto-evo wizard.

Accidental-Explorer: Added the clarification, I can call in GameDungeon to see if they have time.

hhyyrylainen: I think this is minor enough change to not need to bother them...

PR #6488, opened 2025-11-21
Moderately buff bacterial toxin generation

adQuid:

Based on my 11th-hour balance testing, I think the easiest change for making heterotrophs more viable is that toxin is just not economical in the early game. As a bacteria you just can't kill anything for less than the ATP that you would get by eating it. Therefore, this PR lowers the ATP cost of all bacterial toxin generation, and increases the output a bit for the toxins that do damage, since I found the direct damage toxins to be a little weaker than all the rest since you need multiple hits on the same target to land a kill in almost all cases. By not changing damage numbers, this wouldn't have much effect on players trying to not die to AI toxin cells (but it's still possible, of course :) )

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


Deus-Codes:
I quite like this change! An unfortunate aspect of predation via spikes and toxins is that cells bursting can inherently lead to some difficulty engulfing organelle parts. If it takes a while for you to kill another organism through either method, and then some time to actually engulf the goop, that's a lot of energy spent with a relatively uncertain reward. I think this does a good job to make that expenditure time a bit less taxing.

Deus-Codes (approved):

hhyyrylainen:
I guess we can put this into the quick patch to get more balance feedback.

Issue #1610, opened 2020-09-13
Make microbe movement sound scale by movement speed


hhyyrylainen:
to make slower cells quieter

PR #6489, opened 2025-11-22
Changes the volume of microbes based on their speed

ljm862:
I found that 30 is a pretty good step to scale the volume.

Uses the speed of a microbe to determine what its maximum volume should be.

Divide the speed by the sound scaler and clamp this value between 0 and the maximum volume scaler.
We then take the square root of that value and multiply it by the base volume.

$$\displaystyle scaler = \min(\frac{V}{SoundScaler},SoundScaler_{max})$$
$$\displaystyle volume_{max} = volume_{base} * \sqrt{scaler}$$

With the following constants:
$$SoundScaler = 30, SoundScaler_{max} = 4, volume_{base} = 0.4$$
**Related Issues**

Closes: #1610

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/MicrobeMovementSoundSystem.cs
+++ b/src/microbe_stage/systems/MicrobeMovementSoundSystem.cs
@@ -34,8 +34,8 @@ public partial class MicrobeMovementSoundSystem : BaseSystem<World, float>
             var acceleration = physics.Velocity - status.LastLinearVelocity;
             var deltaAcceleration = (acceleration - status.LastLinearAcceleration).LengthSquared();
 
-            var volumeScaler = Math.Clamp(physics.Velocity.Length() / Constants.MICROBE_MOVEMENT_SOUND_SPEED_SCALER,
-                0, Constants.MICROBE_MOVEMENT_SOUND_MAX_VOLUME_SCALER);
+            var volumeScaler = Math.Min(physics.Velocity.Length() / Constants.MICROBE_MOVEMENT_SOUND_SPEED_SCALER,
+                Constants.MICROBE_MOVEMENT_SOUND_MAX_VOLUME_SCALER);
 
             if (status.MovementSoundCooldownTimer > 0)
                 status.MovementSoundCooldownTimer -= delta;


hhyyrylainen:
Well there isn't really a max speed for cells. Whatever speed they can achieve is just how many flagella they have. You can probably try to make a cell with just one flagellum and metabolosomes (once oxygen comes in) and then with infinite compounds cheat test what's the max speed is. For safety I would put in a max loudness for the sound that cannot be exceeded.

ljm862:
> Well there isn't really a max speed for cells. Whatever speed they can achieve is just how many flagella they have. You can probably try to make a cell with just one flagellum and metabolosomes (once oxygen comes in) and then with infinite compounds cheat test what's the max speed is. For safety I would put in a max loudness for the sound that cannot be exceeded.

Thanks for the input! I've made the changes as suggested. 
I found that 30 is a good step to use as a scaler instead of trying to normalise everything. Otherwise all the early game cells will be super quiet and it doesn't feel right

src/microbe_stage/systems/MicrobeMovementSoundSystem.cs
```diff
@@ -33,6 +34,9 @@ private void Update([Data] in float delta, ref MicrobeStatus status, ref Microbe
             var acceleration = physics.Velocity - status.LastLinearVelocity;
             var deltaAcceleration = (acceleration - status.LastLinearAcceleration).LengthSquared();
 
+            var volumeScaler = Math.Clamp(physics.Velocity.Length() / Constants.MICROBE_MOVEMENT_SOUND_SPEED_SCALER,
```
Patryk26g: Shouldn't Math.Min just be enough? `physics.Velocity.Length() / Constants.MICROBE_MOVEMENT_SOUND_SPEED_SCALER` will never be negative

ljm862: Yeah I think it would be fine with just the Min. But I like the sign safety this brings! 
If the scaler is accidentally assigned to negative I expect the game would crash. Though maybe it's better that it does instead of just silently failing? Yes I think it probably is, I'll make the change!

Issue #5944, opened 2025-02-24
Make tolerance sliders move to the max amount they can if not enough MP for a full movement


hhyyrylainen:
as right now they just get blocked if trying to move too far at once

PR #6494, opened 2025-11-23
Fix sliders in editor at fast movements or far ranges

ljm862:

Improves all sliders with their handling of costly movements.

Now when dragging a slider if you drag too far it will clamp its value to as far as it can pay.
Scales down pressure values for UI use by a factor of 1000. Then rescales them back up for assigning for use in game.

**Related Issues**

Closes: #5944

**Progress Checklist**

      break existing features:
      members reviewing this PR)

I'm attaching a video demonstration to help explain a bit better!

https://github.com/user-attachments/assets/f6ac1b79-d241-4e81-8e0e-24e9798451a6

https://github.com/user-attachments/assets/41b4e3ab-dcff-4db2-bb18-e199b629fcbc

And a specific video demonstrating the pressure threshold changes

https://github.com/user-attachments/assets/6abc83e3-a8d7-442a-90bc-ff891d93fc02

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -1946,7 +1946,6 @@ public static class Constants
     public const float TOLERANCE_INITIAL_PRESSURE_MIN_FRACTION = 0.8f;
     public const float TOLERANCE_INITIAL_PRESSURE_MAX_FRACTION = 1.2f;
 
-    public const float TOLERANCE_PRESSURE_SCALE = 1000;
     public const float TOLERANCE_PRESSURE_RANGE_MAX = 2000000;
     public const float TOLERANCE_PERFECT_PRESSURE_SCORE = 0.1f;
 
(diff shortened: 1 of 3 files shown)


src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -709,12 +759,51 @@ private void OnUVResistanceSliderChanged(float value)
...
+    /// </summary>
+    /// <param name="cpa">Cost per action. How much MP does changing this value cost?</param>
+    /// <param name="sliderValue">What the slider value currently is at</param>
+    /// <param name="originalValue">What the slider value was originally at</param>
+    /// <param name="isPercentage">
+    ///   If the value being calculated is a percentage. This is used
+    ///   for determining the rounding for the amount we can change by.
+    /// </param>
+    /// <returns>The value the slider and parameter should now be at.</returns>
+    private float CalculateSliderExtremeValue(double cpa, float sliderValue, float originalValue,
```
hhyyrylainen: `cpa` should be fully written out as it is not on the approved list of abbreviations (check the style guide file).

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -709,12 +759,51 @@ private void OnUVResistanceSliderChanged(float value)
...
+        bool isPercentage = false)
+    {
+        var pointsLeft = Editor.MutationPoints;
+
+        // If a percentage, round to the nearest 0.05
+        var numActions = isPercentage ? Math.Floor(pointsLeft * 20 / cpa) / 20 : Math.Floor(pointsLeft / cpa);
```
hhyyrylainen: I think this is not a safe assumption that things must be rounded to nearest 0.05. Someone could edit the GUI. So instead I think what you need to do is pass in the step size to this method and have the callers read the values from the GUI slider step size.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
```diff
@@ -672,8 +692,28 @@ private void TryApplyPressureChange(float min, float max)
...
+                    min, CurrentTolerances.PressureMinimum);
+
+                var extremeMax = extremeMin + pressureRange;
+
+                reusableTolerances.PressureMinimum = extremeMin;
+                reusableTolerances.PressureMaximum = extremeMax;
+
+                pressureMinSlider.Value = extremeMin;
+                pressureMaxSlider.Value = extremeMax;
+            }
```
hhyyrylainen, line 713: Doesn't this need an else branch? Otherwise this PR is almost complete but not exactly.

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
ljm862: Noted!

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
ljm862: Yep, good point!

src/microbe_stage/editor/TolerancesEditorSubComponent.cs
ljm862: Yes, I put it off last night but today I had an idea for it and so I've completed it now! 
I thought it would require a much bigger rework, and potentially it still should, but the specific solution here works nicely now and addresses every slider.

simulation_parameters/Constants.cs
```diff
@@ -1946,6 +1946,7 @@ public static class Constants
     public const float TOLERANCE_INITIAL_PRESSURE_MIN_FRACTION = 0.8f;
     public const float TOLERANCE_INITIAL_PRESSURE_MAX_FRACTION = 1.2f;
 
+    public const float TOLERANCE_PRESSURE_SCALE = 1000;
```
hhyyrylainen: I see you've changed the GUI so that the values are not directly in there. Is there a reasoning for this change?

Also while in general fine, I do like to get the most performance out of the code we can (as we are working on a game) and as such I try to avoid *divisions* whenever possible. So I'll recommend adding a new constant like `TOLERANCE_PRESSURE_SCALE_INVERSE = 1 / TOLERANCE_PRESSURE_SCALE` and then swapping all the new divisions to be multiplies by that constant instead. I know it doesn't in this case make any difference for performance but I'd still like to keep the overall approach in the codebase about maximum performance when it doesn't hinder code clarity (much).

And sorry for the partial review only, I'm still fully focused on trying to improve the game stability and get a new patch out in the next few days.

ljm862: No worries for the partial review! I'm not in a rush :) 

I'd initially done this to help with some of the operations by bringing down the scale a bit more and to remove any doubt of large number precision being a factor anywhere in the code. I was having what seemed to be precision problems with these 2 sliders specifically due to their comparatively large scale. I did try to remove them once I'd finished the PR and found that the problems had come back. So I kept it in believing it was down to a precision error.
Though when removing them again just now I no longer have the problem at all, so I must have done something wrong previously! 

**I'll remove them as they're seemingly not needed!**

For the performance, I agree and with anything in the realtime parts of the game I'd agree. But as this wasn't a performance critical part of the code being a slider change in the editor, I didn't put too much thought behind it. I think it's fair to ask for keeping the overall approach though so in the future I'll keep that in mind!
With that I've changed the `CalculateSliderExtremeValue` in this new commit to do this by using precomputed inverses of the MP costs.

hhyyrylainen: Without fully checking the latest changes, I'd guess they were likely due to the step size as Godot disallows sliders to get a non-step multiple of their value.

Not related to this case specifically, but I remember there being some rigidity slider problems where a partial step was calculated and then when Godot snapped the value in the GUI to the closest step that then caused another slider change event and if I remember right that caused an infinite loop and a stackoverflow. So yeah, need to be careful with Godot sliders where the step size is either high or unexpected in the context of the code.

PR #6495, opened 2025-11-23
Fix cells spinning around

Patryk26g:

H2S concentration soesn't go all the way down to 0 and that causes spinning loops

Note: I didn't check the code for reduction of H2S concentraion in cell but that might need looking into, so that it goes all the way down to 0 anyway - sometimes the bar stays at 0.0/20.0 instead of disappearing

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/MicrobeAISystem.cs
+++ b/src/microbe_stage/systems/MicrobeAISystem.cs
@@ -322,7 +322,7 @@ public partial class MicrobeAISystem : BaseSystem<World, float>, ISpeciesMemberL
 
         // If motionless inside a hydrogen sulfide cloud, move away
         if (!organelles.HydrogenSulfideProtection
-            && compounds.GetCompoundAmount(Compound.Hydrogensulfide) > 0.1f)
+            && compounds.GetCompoundAmount(Compound.Hydrogensulfide) > Constants.HYDROGEN_SULFIDE_DAMAGE_THESHOLD)
         {
             // If you aren't moving, turn and start. If you are moving, keep going in that direction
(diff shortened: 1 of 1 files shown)


hhyyrylainen:
> Note: I didn't check the code for reduction of H2S concentraion in cell but that might need looking into, so that it goes all the way down to 0 anyway - sometimes the bar stays at 0.0/20.0 instead of disappearing

I tried this change on the original PR adding that damage system, however it can cause problems like extremely fast flashing of the sulfide bar appearing and disappearing and some other minor problems if I remember right. So I ended up reverting that experimental change.

src/microbe_stage/systems/MicrobeAISystem.cs
```diff
@@ -322,7 +322,7 @@ private void ChooseActions(in Entity entity, ref MicrobeAI ai, ref CompoundAbsor
 
         // If motionless inside a hydrogen sulfide cloud, move away
         if (!organelles.HydrogenSulfideProtection
-            && compounds.GetCompoundAmount(Compound.Hydrogensulfide) > MathUtils.EPSILON)
+            && compounds.GetCompoundAmount(Compound.Hydrogensulfide) > 0.1f)
```
hhyyrylainen: Shouldn't this be `> Constants.HYDROGEN_SULFIDE_DAMAGE_THESHOLD` for safety? I think that should correctly fix the bug and not accidentally leave some cell taking damage? Or some multiplier, your value now seems to be 2x that constant.

Patryk26g: True, that still works as intended

PR #6501, opened 2025-11-24
Pilus rebalance

Accidental-Explorer:

This PR changes the Pilus to do damage while in contact, even without breaking contact first before stabbing again,
like slowly driving a needle into a balloon until it pops.

Added a minimum damage to the pilus and reduced the minimum cooldown. 

Reduced max damage to compensate for other changes.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -1085,7 +1085,7 @@ public static class Constants
 
     public const float PILUS_MAX_COOLDOWN = 0.45f;
 
-    public const float PILUS_MIN_DAMAGE_TRIGGER_COOLDOWN = 1.0f;
+    public const float PILUS_MIN_DAMAGE_TRIGGER_COOLDOWN = PILUS_MIN_DAMAGE;
 
     /// <summary>
     ///   Osmoregulation ATP cost per second per hex


src/microbe_stage/systems/PilusDamageSystem.cs
```diff
@@ -111,9 +107,8 @@ private void DealPilusDamage(ref MicrobePhysicsExtraData ourExtraData, ref Physi
...
         float damage = Constants.PILUS_BASE_DAMAGE * collision.PenetrationAmount;
 
-        // Skip too small damage
-        if (damage < 0.01f)
-            return;
+        if (damage < Constants.PILUS_MIN_DAMAGE)
+            damage = Constants.PILUS_MIN_DAMAGE;
 
         if (damage > Constants.PILUS_MAX_DAMAGE)
             damage = Constants.PILUS_MAX_DAMAGE;
```
ljm862: A suggestion for this chunk, we could instead do
`damage = Math.Clamp(damage, Constants.PILUS_MIN_DAMAGE, Constants.PILUS_MAX_DAMAGE);`

To me it keeps the intention clearer in one line

Accidental-Explorer: I have no issue with that. Pushed this change.

hhyyrylainen:
Code change looks good to me now but I'll give some more time for more playtesting.

hhyyrylainen:
I did a quick playtest myself and I think the max damage per hit nerf is making hunting more difficult. As often my strategy is to turn my cell and quickly slam my pilus into a target cell. That causes the cell to fly away quite far so getting many hits in is difficult, and I think the max damage nerf here makes it take more hits.

What was the reasoning behind reducing the max damage per hit?

Edit: my suggestion would be to only slightly nerf the max damage and maybe increase the max invulnerability time if your concern is that you can kill things too fast?

Accidental-Explorer:
> I did a quick playtest myself and I think the max damage per hit nerf is making hunting more difficult. As often my strategy is to turn my cell and quickly slam my pilus into a target cell. That causes the cell to fly away quite far so getting many hits in is difficult, and I think the max damage nerf here makes it take more hits.
> 
> What was the reasoning behind reducing the max damage per hit?

With the more frequent hits, target cells seemed to be dying quite quickly, which made me feel like the max damage might have been to high (I am still not entirely sure if "penetration depth" only applies on instantaneous hits, or also affect "pushing" in practise). 

Since you notice a negative effect I will reverse it and see if the others agree that damage is not too high.

hhyyrylainen:
> With the more frequent hits, target cells seemed to be dying quite quickly, which made me feel like the max damage might have been to high (I am still not entirely sure if "penetration depth" only applies on instantaneous hits, or also affect "pushing" in practise).

Max damage applies when physics penetration amount is high. The max damage shouldn't be able to trigger on continuous hits. But if it does then tweaking `PILUS_BASE_DAMAGE` to be lower would be the right solution. That is what scales the base damage amount from the physics penetration. 

`float damage = Constants.PILUS_BASE_DAMAGE * collision.PenetrationAmount;`

As you added clamping to always deal at least 1 damage, lowering that shouldn't cause problems with continuous contacts not dealing damage.

simulation_parameters/Constants.cs
```diff
@@ -1079,11 +1081,11 @@ public static class Constants
     ///   How long the shortest pilus cooldown is after dealing damage. This is applied if the damage just barely
     ///   crosses <see cref="PILUS_MIN_DAMAGE_TRIGGER_COOLDOWN"/>
     /// </summary>
-    public const float PILUS_MIN_COOLDOWN = 0.2f;
+    public const float PILUS_MIN_COOLDOWN = 0.1f;
 
     public const float PILUS_MAX_COOLDOWN = 0.45f;
 
-    public const float PILUS_MIN_DAMAGE_TRIGGER_COOLDOWN = PILUS_MAX_DAMAGE * 0.6f;
+    public const float PILUS_MIN_DAMAGE_TRIGGER_COOLDOWN = 1.0f;
```
hhyyrylainen:
```suggestion
    public const float PILUS_MIN_DAMAGE_TRIGGER_COOLDOWN = PILUS_MIN_DAMAGE;
```
This won't change the behaviour but will make future balancing changes safe by not accidentally breaking the cooldown system totally.

PR #6502, opened 2025-11-24
Fix various cases of the new MP system not matching up actions to placed things

hhyyrylainen:

So far fixes a couple of problems in the microbe editor, still pending fixes for multicell editor

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/general/mutation_points/CellTypeFacadeHelper.cs
+++ b/src/general/mutation_points/CellTypeFacadeHelper.cs
@@ -22,6 +22,8 @@ public class CellTypeFacadeHelper
     {
         foreach (var addedCellType in activeCellTypes)
         {
+            // Reset previous changes so this is ready for new changes
+            addedCellType.Value.OnStartApplyChanges();
             unusedCellTypes.Push(addedCellType.Value);
         }
 


hhyyrylainen:
With the major fix to multicell editor, I think this is ready for testing now.

src/general/mutation_points/CellTypeEditsFacade.cs
```diff
@@ -315,6 +315,23 @@ internal override bool ApplyAction(EditorCombinableActionData actionData)
                 }
             }
 
+            if (original == null)
```
Patryk26g, line 318: Can't we put it in the "if" below?

Edit: nvm, now I see why it's like that

src/general/mutation_points/MulticellularEditsFacade.cs
```diff
@@ -290,18 +306,27 @@ public CellWithOriginalReference(IReadOnlyCellTemplate original, CellTypeEditsFa
...
 
-        public IReadOnlyCellTemplate OriginalFrom { get; private set; }
+        public Hex Position { get; set; }
+        public int Orientation { get; set; }
+        public IReadOnlyCellTemplate Data => this;
 
-        public override CellType ModifiableCellType
+        public IReadOnlyCellTypeDefinition CellType => usingCustomType;
+
+        public IReadOnlyHexWithData<IReadOnlyCellTemplate> OriginalFrom { get; private set; }
```
Patryk26g, line 315: Maybe better name would be OriginalHex / OriginalHexData? This name is kinda confusing

src/multicellular_stage/MulticellularLayoutHelpers.cs
```diff
@@ -0,0 +1,93 @@
...
+/// <summary>
+///   Conversion helpers between the full (gameplay) and editor layouts of a multicellular species
+/// </summary>
+public static class MulticellularLayoutHelpers
+{
+    public static void UpdateGameplayLayout(CellLayout<CellTemplate> target,
+        IndividualHexLayout<CellTemplate> copySourceTo, IndividualHexLayout<CellTemplate> source,
+        List<Hex> hexTemporaryMemory, List<Hex> hexTemporaryMemory2)
+    {
+        copySourceTo.Clear();
```
Patryk26g: maybe rename those to targetGameplayLayout and targetEditorLayout for more clarity (if that's what those are)

src/multicellular_stage/MulticellularSpecies.cs
```diff
@@ -63,19 +95,39 @@ public static MulticellularSpecies ReadFromArchive(ISArchiveReader reader, ushor
 
         instance.ReadNonConstructorBaseProperties(reader, 1);
 
-        instance.ModifiableCells = reader.ReadObject<CellLayout<CellTemplate>>();
-        instance.EditorCellLayout = reader.ReadObjectOrNull<IndividualHexLayout<CellTemplate>>();
+        instance.ModifiableGameplayCells = reader.ReadObject<CellLayout<CellTemplate>>();
+        instance.modifiableEditorCells = reader.ReadObjectOrNull<IndividualHexLayout<CellTemplate>>();
```
Patryk26g, line 99: modifiableEditorCells -> ModifiableEditorCells?

Patryk26g:
After 1 glance at the code, some needed name changes is all I have noticed

src/multicellular_stage/MulticellularSpecies.cs
hhyyrylainen: This intentionally uses the field directly as the saving *does not want any automatic actions to trigger*.

src/general/mutation_points/MulticellularEditsFacade.cs
hhyyrylainen: I could rename this but the single cell and macroscopic facades use the same exact name, so the names would no longer match and we would have:

```
OriginalOrganelle
OriginalCell
OriginalMetaball
```

I thought that it would help to have the code more uniform in all facades with the same name instead of the name depending on which facade the code needs to be in...

src/multicellular_stage/MulticellularLayoutHelpers.cs
hhyyrylainen: Makes sense, I'll rename those.

hhyyrylainen:
I'll merge this but immediately start a new PR to fix multicellular edits to different cell types which is totally unsupported with the current code and will immediately break.

PR #6507, opened 2025-11-25
Lighting element shader

ljm862:

Slightly reworks how lighting works, both for the background and the sunlight coming from the sky. 
Not only should this be more performant but also should prevent things slipping into HDR and causing everything to be white and pale. Clamping the cloud brightness as well to aid in this.

Changes the colour of the water depending on what compounds are available:
- Green - Iron
- Red - Iron & Oxygen
Can easily be extended to others that may have an impact in certain mixtures.
I'd like feedback before merging on these colours so I can adjust the colour formula as I've tried so many tweaks that colours just all look the same now.

Inspired by [this thread](https://forum.revolutionarygamesstudio.com/t/changing-environmental-colors-in-response-to-compounds/1195)
**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/shaders/CompoundCloudPlane.gdshader
+++ b/shaders/CompoundCloudPlane.gdshader
@@ -27,7 +27,8 @@ const float CLOUD_MAX_INTENSITY_SHOWN = 1000.f;
 
 float getIntensity(float value)
 {
-    return min(DENSITY_MULTIPLIER * atan(0.006f * CLOUD_MAX_INTENSITY_SHOWN * value), 1.0f) * BrightnessMultiplier;
+    float baseIntensity = DENSITY_MULTIPLIER * atan(0.006f * CLOUD_MAX_INTENSITY_SHOWN * value);
+    return min(baseIntensity * BrightnessMultiplier, 1.0f);
 }
 
 void fragment() {
@@ -60,7 +61,7 @@ void fragment() {
                   colour4 * cloud4;
 
     float adjustedAlpha = min(cloud1 + cloud2 + cloud3 + cloud4, 1.0) *
-        clamp(min(cloud1 + cloud2 + cloud3 + cloud4, 1.0) * 0.5, 0.5, 0.9);
+       clamp(min(cloud1 + cloud2 + cloud3 + cloud4, 1.0) * 0.5, 0.5, 0.9);
 
     ALPHA = adjustedAlpha;
     ALBEDO = colour.rgb;
--- a/shaders/CompoundCloudPlane.gdshader
+++ b/shaders/CompoundCloudPlane.gdshader
@@ -61,7 +61,7 @@ void fragment() {
                   colour4 * cloud4;
 
     float adjustedAlpha = min(cloud1 + cloud2 + cloud3 + cloud4, 1.0) *
-       clamp(min(cloud1 + cloud2 + cloud3 + cloud4, 1.0) * 0.5, 0.5, 0.9);
+        clamp(min(cloud1 + cloud2 + cloud3 + cloud4, 1.0) * 0.5, 0.5, 0.9);
 
     ALPHA = adjustedAlpha;
     ALBEDO = colour.rgb;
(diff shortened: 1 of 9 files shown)


shaders/CompoundCloudPlane.gdshader
```diff
@@ -60,7 +61,7 @@ void fragment() {
                   colour4 * cloud4;
 
     float adjustedAlpha = min(cloud1 + cloud2 + cloud3 + cloud4, 1.0) *
-        clamp(min(cloud1 + cloud2 + cloud3 + cloud4, 1.0) * 0.5, 0.5, 0.9);
+       clamp(min(cloud1 + cloud2 + cloud3 + cloud4, 1.0) * 0.5, 0.5, 0.9);
```
hhyyrylainen: If I counted right this indent is now 3 spaces which is incorrect?

src/microbe_stage/BackgroundPlane.cs
```diff
@@ -47,6 +47,8 @@ public partial class BackgroundPlane : Node3D
 
     private Vector2 previousWindowSize = new(1280, 720);
 
+    private Color previousColor;
```
hhyyrylainen, line 50: Shouldn't this be initialized to the same value as in the shader (1, 1, 1, 1)?

hhyyrylainen:
I don't have much time to review this yet, but I have one general question: how does this *compile*? The microbe editor also sets the light level but I saw no changes to the microbe editor in this PR. So how can it work when the previous light level variable was removed?

ljm862:
> I don't have much time to review this yet, but I have one general question: how does this _compile_? The microbe editor also sets the light level but I saw no changes to the microbe editor in this PR. So how can it work when the previous light level variable was removed?

So I'm on my phone so this is all from memory and conjecture because github on mobile is not a fun experience BUT,
light level got set in a few places and we still allow for light level to be set, but not the light level is taken into account for the sky shader only as the sunlight lights up the water. I'll sort out the editor one when I'm back! 
I also noticed that trying to set uniforms that don't exist just seems to not throw any sort of error or warning anywhere so if it tried to directly change the shader parameter then it just gets ignored (though it might actually throw an error somewhere which gets hidden and so would have a performance decrease, hmmm I'll have to check that too)

hhyyrylainen:
> if it tried to directly change the shader parameter then it just gets ignored (though it might actually throw an error somewhere which gets hidden and so would have a performance decrease, hmmm I'll have to check that too)

That's quite likely the case then... as you removed the light level parameter (in the shader file) and I'm pretty sure the microbe editor will still try to use that. So the time of day buttons in the microbe editor should not work correctly.

Overall I think you should move the background colour logic into the microbe camera class because the reason why the light level calculation was there originally is that so that the same logic can be shared by the stage and the editor as both want to use it. Without putting the logic there it would need a separate helper method that the editor could also use.

Deus-Codes:
Playtested this a while ago for only a few minutes I could spare, but great work visually!

One thing I will comment is that the tint effect is really strong, so it can look like an explicit filter applied to an image instead of a dilution or compounds. Perhaps if the tint can be made to look more "muddy" or "diluted", that could be a good balance - maybe retaining a bit of the original color of the background. I trust that you'll find a good balance.

I will also note that in my brief playtest, I didn't get to a point of seeing the oxygen-iron effect, or to see if the red filter disappears eventually to reveal the "actual" color of the waters. That's something we should make sure works well before merging this. 

Great start overall however!

ljm862:
I've restructured the lighting system a bit with regards to the whole atmosphere colouring and how lighting is affecting the shaders. I've also slightly tweaked the formula a bit for calculating the colour but still not sure if it's really good yet. If anyone has some colour suggestions of what they'd like, it would help a lot! Otherwise I'll keep trying until I get something we're all happy with

Deus-Codes:
image

I appear to get this error when testing this branch. The color filters also don't appear to be applied unless they are very subtle, so there might be an error here

Issue #5118, opened 2024-05-13
AI should know how to use the signaling agent to send commands


hhyyrylainen:
For example to warn members of their species when they are attacked / asking for backup.

hhyyrylainen:
Auto-evo can evolve this now so would be nice for the AI to finally be able to use it: https://github.com/Revolutionary-Games/Thrive/pull/6508

IdfbAn:
Would this be good behavior for the AI?

| Situation | Command |
| ---------- | ----------- |
| Precursor to becoming aggressive or self-defense | Move To Me |
| Multiple of the same species close to each other and Move To Me has been recently used | Become Aggressive |
| Hasn't moved much within 10 or so seconds, assumption is it's being blocked by something | Flee From Me |

I currently don't see why Follow Me would be used.

EDIT: Oh, I only just remembered this. Move To Me could also be used if the species has the binding agent, and this could be the start of other multicellular species besides the player's.

hhyyrylainen:
They should also depend on species behaviour values, sessile species or ones that aren't very aggressive I don't think should use the become aggressive commands.

Also the commands will need to be tweaked so that cells can ignore them because if there are too many cells sending signals then there's basically always going to be an active signal and that could make the AI appear really dumb by being forced to do the same action repeatedly.

Edit: but thinking up situations like those when to send signals does seem like a good start.

Accidental-Explorer:
- "Follow me" could be the default for some prey and pack hunters. It takes less time to call others to you if they're already there.
- Prey that are confronted by predators they want to flee from (this is controlled by behaviour scores), should probably also activate "Flee from me" at that time. (scatter in all directions, they can't catch all of us!)
- "Become Aggressive" seems like it's mainly appropriate for situations where species would normally run from predators, but can fight back if there's enough of them. (similarly, small predators attacking big ones)

> Also the commands will need to be tweaked so that cells can ignore them because if there are too many cells sending signals then there's basically always going to be an active signal and that could make the AI appear really dumb by being forced to do the same action repeatedly.

Maybe there could be a trigger here where no cell will activate a command if another cell of the same species is already using one? That way the mechanics can stay mostly similar to current, mainly focused on one cell at a time giving the orders.

hhyyrylainen:
> Maybe there could be a trigger here where no cell will activate a command if another cell of the same species is already using one? That way the mechanics can stay mostly similar to current, mainly focused on one cell at a time giving the orders.

[...]

PR #6508, opened 2025-11-25
Auto evo binding agent and signalling

Accidental-Explorer:

This PR adds auto-evo effects for binding and signalling agents.

The binding agent gives a fixed reduction in the energy expenditure in some places in auto-evo.

The signalling agent increases the previous effect (easier to make colonies) and some types of predation (easier to kill targets if you bring allies).

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
+++ b/src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
@@ -157,29 +157,12 @@ public class ReproductionCompoundPressure : SelectionPressure
 
         var finalScore = 0.1f;
 
-        // modifier to fit current mechanics of the Binding Agent. This should probably be removed or adjusted if
+        // Modifier to fit the current mechanics of the Binding Agent. This should probably be removed or adjusted if
         // being in a colony no longer reduces osmoregulation cost.
         var bindingModifier = 1.0f;
 
-        var hasBindingAgent = false;
-        foreach (var organelle in microbeSpecies.Organelles.Organelles)
-        {
-            if (organelle.Definition.HasBindingFeature)
-            {
-                hasBindingAgent = true;
-                break;
-            }
-        }
-
-        var hasSignallingAgent = false;
-        foreach (var organelle in microbeSpecies.Organelles.Organelles)
-        {
-            if (organelle.Definition.HasSignalingFeature)
-            {
-                hasSignallingAgent = true;
-                break;
-            }
-        }
+        MicrobeInternalCalculations.GetBindingAndSignalling(microbeSpecies.Organelles.Organelles,
+            out var hasBindingAgent, out var hasSignallingAgent);
 
         if (hasBindingAgent)
         {
--- a/src/auto-evo/simulation/SimulationCache.cs
+++ b/src/auto-evo/simulation/SimulationCache.cs
@@ -347,22 +347,25 @@ public class SimulationCache
         var preySlimeSpeed = preySpeed + preySlimeJetScore / (preyHexSize * 11);
 
         var hasChemoreceptor = false;
-        foreach (var organelle in predator.Organelles.Organelles)
+        var hasSignallingAgent = false;
+        var hasBindingAgent = false;
+
+        var organelles = predator.Organelles.Organelles;
+        int count = organelles.Count;
+        for (int i = 0; i < count; ++i)
         {
+            var organelle = organelles[i];
             if (organelle.Definition.HasChemoreceptorComponent && organelle.GetActiveTargetSpecies() == prey)
                 hasChemoreceptor = true;
-        }
 
-        var hasSignallingAgent = false;
-        foreach (var organelle in predator.Organelles.Organelles)
-        {
             if (organelle.Definition.HasSignalingFeature)
-            {
                 hasSignallingAgent = true;
-                break;
-            }
+
+            if (organelle.Definition.HasBindingFeature)
+                hasBindingAgent = true;
         }
 
+        // TODO: switch to a manual loop to avoid an allocation
         var preyOxygenUsingOrganellesCount = 0;
         foreach (var organelle in prey.Organelles.Organelles)
         {
@@ -576,19 +579,9 @@ public class SimulationCache
             scoreMultiplier *= Constants.AUTO_EVO_CHUNK_LEAK_MULTIPLIER;
         }
 
-        // modifier to fit current mechanics of the Binding Agent. This should probably be removed or adjusted if
+        // Modifier to fit the current mechanics of the Binding Agent. This should probably be removed or adjusted if
         // being in a colony no longer reduces osmoregulation cost.
         var bindingModifier = 1.0f;
-
-        var hasBindingAgent = false;
-        foreach (var organelle in predator.Organelles.Organelles)
-        {
-            if (organelle.Definition.HasBindingFeature)
-            {
-                hasBindingAgent = true;
-                break;
-            }
-        }
 
         if (hasBindingAgent)
         {
--- a/src/auto-evo/selection_pressure/CompoundConversionEfficiencyPressure.cs
+++ b/src/auto-evo/selection_pressure/CompoundConversionEfficiencyPressure.cs
@@ -67,29 +67,12 @@ public class CompoundConversionEfficiencyPressure : SelectionPressure
 
         var score = cache.GetCompoundConversionScoreForSpecies(FromCompound, ToCompound, microbeSpecies, patch.Biome);
 
-        // modifier to fit current mechanics of the Binding Agent. This should probably be removed or adjusted if
+        // Modifier to fit the current mechanics of the Binding Agent. This should probably be removed or adjusted if
         // being in a colony no longer reduces osmoregulation cost.
         var bindingModifier = 1.0f;
 
-        var hasBindingAgent = false;
-        foreach (var organelle in microbeSpecies.Organelles.Organelles)
-        {
-            if (organelle.Definition.HasBindingFeature)
-            {
-                hasBindingAgent = true;
-                break;
-            }
-        }
-
-        var hasSignallingAgent = false;
-        foreach (var organelle in microbeSpecies.Organelles.Organelles)
-        {
-            if (organelle.Definition.HasSignalingFeature)
-            {
-                hasSignallingAgent = true;
-                break;
-            }
-        }
+        MicrobeInternalCalculations.GetBindingAndSignalling(microbeSpecies.Organelles.Organelles,
+            out var hasBindingAgent, out var hasSignallingAgent);
 
         if (hasBindingAgent)
         {
--- a/src/microbe_stage/MicrobeInternalCalculations.cs
+++ b/src/microbe_stage/MicrobeInternalCalculations.cs
@@ -793,6 +793,35 @@ public static class MicrobeInternalCalculations
         return Constants.AI_ACTIVITY_NIGHT_MULTIPLIER;
     }
 
+    public static void GetBindingAndSignalling(IReadOnlyList<OrganelleTemplate> organelles, out bool hasBindingAgent,
+        out bool hasSignallingAgent)
+    {
+        int count = organelles.Count;
+        hasBindingAgent = false;
+        hasSignallingAgent = false;
+
+        for (int i = 0; i < count; ++i)
+        {
+            var organelleDefinition = organelles[i].Definition;
+
+            if (organelleDefinition.HasBindingFeature)
+            {
+                hasBindingAgent = true;
+
+                if (hasSignallingAgent)
+                    break;
+            }
+
+            if (organelleDefinition.HasSignalingFeature)
+            {
+                hasSignallingAgent = true;
+
+                if (hasBindingAgent)
+                    break;
+            }
+        }
+    }
+
     private static float MovementForce(float movementForce, float directionFactor)
     {
         if (directionFactor < 0)


adQuid:
does the AI use the binding agent?

Accidental-Explorer:
> does the AI use the binding agent?

As far as I am aware, no. I also didn't add mutation-adding for both these organelles in this draft yet.

From what I remember, previous discussion concluded "Put it in auto-evo first. And _then_ it might be worth giving AI effects to it."

hhyyrylainen:
> "Put it in auto-evo first. And then it might be worth giving AI effects to it."

I don't remember such a discussion.

It would be more helpful to make the AI logic first as the player species will be getting the organelles so the organelles functioning for AI members of the player species is a very important point.

Accidental-Explorer:
> > "Put it in auto-evo first. And then it might be worth giving AI effects to it."
> 
> I don't remember such a discussion.

I can't find it either now, so I am probably misremembering something.

> It would be more helpful to make the AI logic first as the player species will be getting the organelles so the organelles functioning for AI members of the player species is a very important point.

I don't really like having "the player is punished by auto-evo for placing Binding and Signalling Agents" as the interim situation (which is the current case). But I think for now this can then be considered "blocked" until there is satisfactory AI implementation.

hhyyrylainen:
But the player *already* is forced to place those organelles, and nobody has complained that it is unfair in terms of auto-evo numbers?

Accidental-Explorer:
Generally all cases of "this organelle is useful to the player, but useless for auto-evo" accumulate and contribute to the "why is my species population so low" question that at least used to come up quite often.

hhyyrylainen:
Okay, that's a fair point.

Accidental-Explorer:
Now I am not saying this is a _good_ solution we should go for, but there are some potential options if we want players to not be negatively affected by the lack of auto-evo integration without having AI species place parts that are useless to them:
1. Don't implement the mutation strategies to add them yet. This means only descendents of the player species might have them (the current situation) (Also, coincidental prevention of reduced auto-evo performance, yay!)
2. It would be very easy to add in a check that only applies the score bonus to player species. This is against general Thrive design ideas, but it would be very easy to simply remove the check once we have AI implementation. 

This would prevent any negative impression on players while not making AI species any less effective in-gameplay. *If* that is desired.
The alternative of simply delaying this is also fine in my opinion.

hhyyrylainen:
I don't think it is worth it to intentionally "sabotage" auto-evo here... if it wants to add organelles it shouldn't be prevented. Still I imagine it will be a much smaller harm of adding useless organelles and players reporting it that they never see the AI actually use their binding agents, than preventing auto-evo from placing those organelles in the first place. 

Though of course the auto-evo balancing might need to take some AI stuff into consideration, because the AI behaviour will need to be different if almost every species gets slapped with binding agents and signalling agents (like how commonly the AI tries to form colonies should be tweaked based on how many species get the binding agent). Granted also species diversity from auto-evo side *already* would have implications on how common the new organelles should be.

simulation_parameters/Constants.cs
```diff
@@ -1325,6 +1325,12 @@ public static class Constants
     public const float AUTO_EVO_ARTIFICIAL_UPGRADE_BONUS_SMALL = 1.5f;
     public const float AUTO_EVO_ARTIFICIAL_UPGRADE_BONUS = 20.0f;
 
+    // Auto-evo energy cost benefit to match the current (possibly temporary) gameplay effect of the binding agent
```
hhyyrylainen: These two new comments should be XML comments like this:

```
/// <summary>
///    A summary goes here
/// </summary>
```

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
```diff
@@ -157,15 +157,52 @@ public override float Score(Species species, Patch patch, SimulationCache cache)
...
+            if (hasSignallingAgent)
+            {
+                bindingModifier *= 1 -
+                    Constants.AUTO_EVO_COLONY_OSMOREGULATION_BONUS * Constants.AUTO_EVO_SIGNALLING_BONUS;
+            }
+            else
+            {
+                bindingModifier *= 1 - Constants.AUTO_EVO_COLONY_OSMOREGULATION_BONUS;
+            }
+        }
```
ljm862: I think this entire chunk can be pulled out of the for loop and so only evaluated once, not once per chunk as the logic here doesn't seem to depend on any chunk specific logic.

Accidental-Explorer: Very good point, I will do that.

Accidental-Explorer: Actually, looking at it again: Unless I am blind (which is entirely possible), I don't think it's *in* a for loop?

ljm862: Nope, you're fully right. I looked too many brackets across 🤦‍♂️ I should have actually looked at the code and not tried to figure out where it was from github!

We can instead extract it to a function like this to keep the method a bit smaller 
```
    private float GetBindingModifier(MicrobeSpecies microbeSpecies)
    {
        var hasBindingAgent = false;
        foreach (var organelle in microbeSpecies.Organelles.Organelles)
        {
            if (organelle.Definition.HasBindingFeature)
            {
                hasBindingAgent = true;
                break;
            }
        }

        var hasSignallingAgent = false;
        foreach (var organelle in microbeSpecies.Organelles.Organelles)
        {
            if (organelle.Definition.HasSignalingFeature)
            {
                hasSignallingAgent = true;
                break;
            }
        }

        // modifier to fit current mechanics of the Binding Agent. This should probably be removed or adjusted if
        // being in a colony no longer reduces osmoregulation cost.
        var bindingModifier = 1.0f;

        if (hasBindingAgent)
        {
            if (hasSignallingAgent)
            {
                bindingModifier *= 1 -
                    Constants.AUTO_EVO_COLONY_OSMOREGULATION_BONUS * Constants.AUTO_EVO_SIGNALLING_BONUS;
            }
            else
            {
                bindingModifier *= 1 - Constants.AUTO_EVO_COLONY_OSMOREGULATION_BONUS;
            }
        }

        return bindingModifier;
    }
```
And then just call `var bindingModifier = GetBindingModifier(microbeSpecies)`

Accidental-Explorer: > We can instead extract it to a function like this to keep the method a bit smaller

Since you seem to have more experience with programming structure than me: *where* do you think would be best in the file structure to put this? (since other selectionPressures could use the same function).

I could just put it in SimulationCache, but I am not sure if this calculation is expensive enough to be worth caching.

ljm862: I'm not sure on the best place as I'm still quite unfamiliar with the code overall.
Looking around quickly I would say MicrobeSpecies might be a good place for the checking, as this is likely the longest runtime part of the code.

I'd put a `public Boolean HasBindingAgent; public Boolean HasSignallingAgent;` as class variables for a MicrobeSpecies. Then in the `OnEdited` function, call to a function called `UpdateHasBindingAgent()` and `UpdateHasSignallingAgent()`. Then check the list of Organelles within those functions and set the values of the booleans whenever the species is actually edited (if it works how I'm assuming it does!).

Then back in `ReproductionCompoundPressure` we can keep the same `GetBindingModifier(MicrobeSpecies species)` method and then change what it does inside to just be
```
        var bindingModifier = 1.0f;

        if (species.HasBindingAgent)
        {
            if (species.HasSignallingAgent)
            {
                bindingModifier *= 1 -
                    Constants.AUTO_EVO_COLONY_OSMOREGULATION_BONUS * Constants.AUTO_EVO_SIGNALLING_BONUS;
            }
            else
            {
                bindingModifier *= 1 - Constants.AUTO_EVO_COLONY_OSMOREGULATION_BONUS;
            }
        }

        return bindingModifier;
```

That way, any selection pressure stuff doesn't need to recheck if the agents exist, they can then just decide what to do with it themselves. You could then store the bindingModifier value in the cache, but I can't really say if that's worth it 🤔

hhyyrylainen: That kind of code structure is not really going to fit that well... Due to all the caching and needing to be able to show information both in the editor and during gameplay we have `MicrobeInternalCalculations` helper class.
With my new MP system and `MicrobeEditsFacade` adding any new properties that need to be kept up to date to a species is going to be extra painful.

Edit: and that's basically the reason why we don't have derived stats of everything a species possibly has and instead code needs to loop the organelles when the information is required.

Accidental-Explorer: So... In that case what would you request I do?

ljm862: In this case, I think just extracting out the chunk into its own little `Private` method within the `ReproductionCompoundPressure` class is enough. It won't change anything performance wise but it will keep the code a bit neater! 
As all the selection pressures come from the `SelectionPressure` abstract class, we could instead put a `protected` version of the method in there instead. This will avoid the code duplication and then every selection pressure will read it the same way!

`SelectionPressure.cs`
```
    protected float GetBindingModifier(MicrobeSpecies microbeSpecies)
    {
        var hasBindingAgent = false;
        foreach (var organelle in microbeSpecies.Organelles.Organelles)
        {
            if (organelle.Definition.HasBindingFeature)
            {
                hasBindingAgent = true;
                break;
            }
        }

        var hasSignallingAgent = false;
        foreach (var organelle in microbeSpecies.Organelles.Organelles)
        {
            if (organelle.Definition.HasSignalingFeature)
            {
                hasSignallingAgent = true;
                break;
            }
        }

        // modifier to fit current mechanics of the Binding Agent. This should probably be removed or adjusted if
        // being in a colony no longer reduces osmoregulation cost.
        var bindingModifier = 1.0f;

        if (hasBindingAgent)
        {
            if (hasSignallingAgent)
            {
                bindingModifier *= 1 -
                    Constants.AUTO_EVO_COLONY_OSMOREGULATION_BONUS * Constants.AUTO_EVO_SIGNALLING_BONUS;
            }
            else
            {
                bindingModifier *= 1 - Constants.AUTO_EVO_COLONY_OSMOREGULATION_BONUS;
            }
        }

        return bindingModifier;
    }
```

`ReproductionCompoundPressure.cs`
```
...
var bindingModifier = GetBindingModifier(microbeSpecies);

// modify score by energy cost and activity
var activityFraction = ...
```

`CompoundConversionEfficiencyPressure.cs`
```
...
var bindingModifier = GetBindingModifier(microbeSpecies);

// we need to factor in both conversion from source to output, and energy expenditure time
if (usedForSurvival)
```

hhyyrylainen: > As all the selection pressures come from the SelectionPressure abstract class, we could instead put a protected version of the method in there instead. This will avoid the code duplication and then every selection pressure will read it the same way!

I would advice against this as many calculations end up needing access to it both inside and outside auto-evo so it makes more sense to put it into `MicrobeInternalCalculations` which is the class to share code between auto-evo and other uses for calculations and ensuring the same calculation is done the same way everywhere.

Though as this is literally just a small loop looking for an organelle with a feature flag, I would just leave the loop the way it is. (it is quite unlikely this logic ever can become out of sync between different implementations, and we have already a lot of code like this to everything would anyway break if anyone wanted to change the semantics of similar code)

hhyyrylainen: Actually maybe it would make sense to have a helper method to check if something has signalling agent and binding agent, because I just checked and this code *will allocate an enumerator*:

```
foreach (var organelle in microbeSpecies.ModifiableOrganelles.Organelles)
```

Meaning it should be always avoided in favour of a manual loop of the IReadOnlyList.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -559,14 +576,41 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
             scoreMultiplier *= Constants.AUTO_EVO_CHUNK_LEAK_MULTIPLIER;
         }
 
+        // modifier to fit current mechanics of the Binding Agent. This should probably be removed or adjusted if
+        // being in a colony no longer reduces osmoregulation cost.
+        var bindingModifier = 1.0f;
+
+        var hasBindingAgent = false;
+        foreach (var organelle in predator.Organelles.Organelles)
```
hhyyrylainen: There's already a loop above for the predator organelles so it would be probably better to do this there instead of here doing one more loop on the same data...

hhyyrylainen: I couldn't help myself so I had to write that helper method and fix some of these enumerator allocations.

src/auto-evo/selection_pressure/ReproductionCompoundPressure.cs
ljm862: Yes, that's another benefit I didn't think of, the enumerator! 

I thought about `MicrobeInternalCalculations` but I wasn't sure if that was the correct place for auto-evo and gameplay logic, thanks for the clarification!

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -551,6 +564,13 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
                 * float.Sqrt(preyIndividualCost);
         }
 
+        // Calling for allies helps with combative hunting.
+        if (hasSignallingAgent)
+        {
+            pilusScore *= Constants.AUTO_EVO_SIGNALLING_BONUS;
+            damagingToxinScore *= Constants.AUTO_EVO_SIGNALLING_BONUS;
+        }
```
adQuid, line 572: Does the AI actually use signaling agents to hunt (or at all)?

hhyyrylainen: No, the AI cannot trigger them, which is why I said this originally: https://github.com/Revolutionary-Games/Thrive/pull/6508#issuecomment-3579552052 and the ensuing discussion.

Relevant issue: https://github.com/Revolutionary-Games/Thrive/issues/5118

PR #6514, opened 2025-11-26
Change all other collision checks to use is alive check like the pilus 

hhyyrylainen:

I did one fix for pilus collision in time for 0.9.1.1, but that then encountered a very similar error in a different system. So I've now gone to all of the physics collision checkers and changed them to do a similar safety check

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/common_systems/DamageOnTouchSystem.cs
+++ b/src/engine/common_systems/DamageOnTouchSystem.cs
@@ -52,8 +52,11 @@ public partial class DamageOnTouchSystem : BaseSystem<World, float>
         {
             ref var collision = ref collisions![i];
 
+            if (collision.SecondEntity == default(Entity))
+                continue;
+
             // Skip collisions with things that can't be damaged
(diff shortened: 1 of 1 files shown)


src/engine/common_systems/DamageOnTouchSystem.cs
```diff
@@ -53,7 +53,7 @@ private void Update([Data] in float delta, ref DamageOnTouch damageTouch,
             ref var collision = ref collisions![i];
 
             // Skip collisions with things that can't be damaged
-            if (collision.SecondEntity == Entity.Null || !collision.SecondEntity.Has<Health>())
+            if (collision.SecondEntity == default(Entity) || !collision.SecondEntity.IsAliveAndHas<Health>())
```
ljm862: A minor point but: Sometimes an || is used, other times they're separate statements. I feel like these should all be unified

hhyyrylainen: I suppose I can split out the invalid entity check in this. I didn't feel like that was necessary when I was just modifying the few words on this line and not writing new statements.

PR #6519, opened 2025-11-28
Fixed two auto-evo NaN bugs and added more checking to the auto-evo exploring tool

hhyyrylainen:

Fixes some NaN problems and hopefully makes future NaNs slipping in more rare

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/simulation/SimulationCache.cs
+++ b/src/auto-evo/simulation/SimulationCache.cs
@@ -380,29 +380,14 @@ public class SimulationCache
             {
                 // You catch more preys if you are fast, and if they are slow.
                 // This incentivizes engulfment strategies in these cases.
-                // Avoid NaN values
-                if (preySpeed > 0)
-                {
-                    catchScore += (predatorSpeed / preySpeed) * (1 - slowedProportion);
-                }
-                else
-                {
-                    catchScore += predatorSpeed * (1 - slowedProportion);
-                }
+                // Sinusoid calculation to avoid divisions by zero
+                catchScore += (predatorSpeed + 0.001f) / (preySpeed + 0.0001f) * (1 - slowedProportion);
             }
 
             // If you can slow the target, some proportion of prey are easier to catch
             if (predatorSpeed > slowedPreySpeed)
             {
-                // Avoid NaN values
-                if (slowedPreySpeed > 0)
-                {
-                    catchScore += (predatorSpeed / slowedPreySpeed) * slowedProportion;
-                }
-                else
-                {
-                    catchScore += predatorSpeed * slowedProportion;
-                }
+                catchScore += (predatorSpeed + 0.001f) / (preySpeed + 0.0001f) * (1 - slowedProportion);
             }
 
             // But prey may escape if they move away before you can turn to chase them


src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -380,13 +380,29 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
...
-                catchScore += (predatorSpeed / preySpeed) * (1 - slowedProportion);
+                // Avoid NaN values
+                if (preySpeed > 0)
+                {
+                    catchScore += (predatorSpeed / preySpeed) * (1 - slowedProportion);
+                }
+                else
+                {
+                    catchScore += predatorSpeed * (1 - slowedProportion);
+                }
```
adQuid:
```suggestion
                catchScore += (predatorSpeed / (preySpeed + 0.1)) * (1 - slowedProprotion);
```

This if statement will discourage predators from being faster than prey. Let's put some sinusoidal logic in there instead.

src/general/utils/MathUtils.cs
```diff
@@ -245,4 +245,38 @@ public static double BreakOnNaN(this double value)
 
         return value;
     }
+
+    public static float BreakOnInfinity(this float value)
```
adQuid, line 249: It would be handy to modify these functions to have a string message, giving the dev a better hint at _which_ variable is infinity, but it's not necessary since you could always dig through the stack trace.

src/auto-evo/simulation/SimulationCache.cs
hhyyrylainen: 0.1 skews the numbers a lot as from my debugging the speed numbers are usually 0.027343 or so. So that would be basically more impactful than the actual speed values. I thought about adding something like 0.0001 but Rathalos pointed out pretty well on Discord that as the prey speed falls to close to zero it can cause very high catch scores.

src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -380,13 +380,29 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
...
-                catchScore += (predatorSpeed / slowedPreySpeed) * slowedProportion;
+                // Avoid NaN values
+                if (slowedPreySpeed > 0)
+                {
+                    catchScore += (predatorSpeed / slowedPreySpeed) * slowedProportion;
+                }
+                else
+                {
+                    catchScore += predatorSpeed * slowedProportion;
+                }
```
adQuid:
```suggestion
                catchScore += (predatorSpeed / (preySpeed + 0.1)) * (1 - slowedProprotion);
```

same logic as above

src/general/utils/MathUtils.cs
hhyyrylainen: It would be a massive pain to use these methods if you needed to specify the name. Right now the way I use these is to just paste `.BreakOnInfinity()` after a ton of numbers and see where my debugger hits. I can then immediately inspect the upper level stack frames to find all the relevant calculations. But did you have some automated way in mind for getting the name? Or how would you suggest adding that?

src/auto-evo/simulation/SimulationCache.cs
adQuid: an alternative would be to add 0.001 to the numerator and 0.0001 to the denominator. This would give an asymptote of 10, without skewing faster microbes too much. I think.

src/general/utils/MathUtils.cs
adQuid: I mean you could print the top line of the stack trace if you wanted, but other than that it might be better to just leave it if you're going to be using it to copy/paste a bunch.

src/auto-evo/simulation/SimulationCache.cs
hhyyrylainen: I don't know enough about math to say either way... there's just an hour left before the BOTD and I'd like to merge this before that. I guess I can use this approach as it sounds like it wouldn't change things too much and would also solve the NaN issue?

src/general/utils/MathUtils.cs
hhyyrylainen: Yeah. The exception anyway will have the stacktrace in it so people can look there if they didn't happen to be using a debugger.

src/auto-evo/simulation/SimulationCache.cs
hhyyrylainen: I made this change (with the slight numbers tweak).

hhyyrylainen:
The CI will probably pass so I'll merge in order to be able to put out the BOTD a bit earlier as it can only start building after this.

src/auto-evo/simulation/SimulationCache.cs
adQuid: Wait, I meant sigmoidal, not sinusoidal!

hhyyrylainen: Well too late now as it is in the committed code... and I'm too lazy just to make a PR to change the wording.

Accidental-Explorer: > Wait, I meant sigmoidal, not sinusoidal!

It's in the same chunk of code, so I went ahead and fixed this in https://github.com/Revolutionary-Games/Thrive/pull/6522

PR #6522, opened 2025-11-29
Auto evo slime jet rework and sprinting

Accidental-Explorer:

This PR reworks Slime Jets to boost engulfment and pilus predation instead of acting like its own offensive tool. (This fixes auto-evo producing "predators" completely incapable of hunting).

Also uses the same general technique to implement scoring for sprinting (giving a benefit to excess ATP production).

(There's some ancillary related bugfixes in here too that I noticed but weren't reported before)

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/simulation_parameters/Constants.cs
+++ b/simulation_parameters/Constants.cs
@@ -1058,6 +1058,8 @@ public static class Constants
     /// </summary>
     public const float PILUS_MAX_DAMAGE = 45;
 
+    public const float PILUS_MIN_DAMAGE = 1;
+
     public const float PILUS_PHYSICS_SIZE = 4.6f;
 
     public const float BACTERIA_PILUS_ATTACH_ADJUSTMENT_MULTIPLIER = 0.575f;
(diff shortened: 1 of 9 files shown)


hhyyrylainen (approved):
This runs without introducing any new errors, which is about the only thing I can confirm with these auto-evo PRs.

I'll note an unfortunate side effect of this that this basically undoes all of the performance improvement gained by: https://github.com/Revolutionary-Games/Thrive/pull/6527 which is unfortunate but without those performance changes this would get much slower.

Here's the 5 world 15 gen times I got with this PR:
- 2:56
- 4:05
- 3:41
- 3:27
- 1:32

And before these changes on master branch (with the performance improvement applied):
- 2:46
- 1:27
- 1:52
- 0:35
- 1:07

So it's not a totally a showstopper but I think we need to start either improving auto-evo performance at the same time as adding new features or limiting new features to keep runtimes under control.

PR #6523, opened 2025-11-29
Scale number of life indicators with species hex size

Patryk26g:
Also changes the distribution a little bit

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/PatchMapDrawer.cs
+++ b/src/microbe_stage/editor/PatchMapDrawer.cs
@@ -1121,7 +1121,7 @@ public partial class PatchMapDrawer : Control
 
             if (playerSpecies is MicrobeSpecies microbeSpecies)
             {
-                speciesSizeModifier = Math.Sqrt(microbeSpecies.BaseHexSize);
+                speciesSizeModifier = Math.Sqrt(microbeSpecies.Organelles.Organelles.Count);
             }
             else if (playerSpecies is MulticellularSpecies)
             {
@@ -1151,7 +1151,7 @@ public partial class PatchMapDrawer : Control
     private Vector2 GetIndicatorPosition(PatchMapNode node, int dotIndex, Texture2D texture)
     {
         var indexModifier = MathF.Sin(dotIndex) * 0.5f + 0.5f;
-        var nodeModifier = node.Patch.ID;
+        var nodeModifier = node.Position.LengthSquared();
         var nodeCenter = node.Position + new Vector2(Constants.PATCH_NODE_RECT_LENGTH / 2,
             Constants.PATCH_NODE_RECT_LENGTH / 2) - new Vector2(texture.GetWidth() / 2.0f, texture.GetHeight() / 2.0f);
 


src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1136,16 +1148,16 @@ private void AddPatchNode(Patch patch, Vector2 position)
         nodes.Add(node.Patch, node);
     }
 
-    private Vector2 GetIndicatorPosition(Control parent, int dotIndex, Texture2D texture)
+    private Vector2 GetIndicatorPosition(PatchMapNode node, int dotIndex, Texture2D texture)
     {
         var indexModifier = MathF.Sin(dotIndex) * 0.5f + 0.5f;
-        var nodeModifier = parent.Position.LengthSquared();
-        var nodeCenter = parent.Position + new Vector2(Constants.PATCH_NODE_RECT_LENGTH / 2,
+        var nodeModifier = node.Patch.ID;
```
hhyyrylainen: Why does this use ID of the patch to modify the positions? It seems kind of suspicious to me. If in the future we have patch IDs like 4000-100000 range, would that cause issues? I really would not like to let any feature through that may complicate things if patch ID range is in the future expanded or dynamically generated or anything like that.

src/microbe_stage/editor/PatchMapDrawer.cs
```diff
@@ -1117,8 +1117,20 @@ private void AddPatchNode(Patch patch, Vector2 position)
         {
             var playerPopulation = patch.GetSpeciesSimulationPopulation(playerSpecies);
 
+            var speciesSizeModifier = 1.0;
+
+            if (playerSpecies is MicrobeSpecies microbeSpecies)
+            {
+                speciesSizeModifier = Math.Sqrt(microbeSpecies.BaseHexSize);
```
hhyyrylainen: Base hex size is a very expensive thing to calculate, so I'd like to ask that you check the player population first and only do the hex size calculation when it would actually result in any dots on the map.

Patryk26g: I have changed it to organelles count - results are very simillar and calculation is much faster

src/microbe_stage/editor/PatchMapDrawer.cs
Patryk26g: I wanted to make sure that the distribution of the dots is 100% different in each patch, as there might have been a case where lenSquared for 2 patches would be the same, but I reverted these changes as I have not encountered that case

src/microbe_stage/editor/PatchMapDrawer.cs
hhyyrylainen: ~~The position could be used instead, right? That would also be guaranteed to be unique. Or adding the x component of the position to the length squared should again be a unique combination.~~

hhyyrylainen: Oops, commented on the wrong thing...

src/microbe_stage/editor/PatchMapDrawer.cs
hhyyrylainen: The position could be used instead, right? That would also be guaranteed to be unique. Or adding the x component of the position to the length squared should again be a unique combination.

src/microbe_stage/editor/PatchMapDrawer.cs
hhyyrylainen: I guess using the organelle count is a reasonable proxy for size.

PR #6527, opened 2025-12-01
Add a handful of small quick wins to auto-evo performance

ljm862:
Fixes a small number of performance issues I spotted whilst away. 

Here's the flamegraph via dottrace of how it looked before changes were made.
BeforeChanges

As can be seen the `ReproductionCompoundPressure.Score` is by far the largest factor in the auto-evo. Mostly just due to how many times it's called. 

With all the changes I've made here, I **managed to knock about 5-10 seconds off my average world of 15 generations**.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/auto-evo/mutation_strategy/AddOrganelleAnywhere.cs
+++ b/src/auto-evo/mutation_strategy/AddOrganelleAnywhere.cs
@@ -118,6 +118,10 @@ public class AddOrganelleAnywhere : IMutationStrategy<MicrobeSpecies>
             // Don't add duplicate unique organelles
             if (organelle.Unique)
             {
+                // TODO: determine if there are actually any cases where multiple unique organelles might be added at
+                // once, because right now this temporary memory allocation might not help. Because in most cases where
+                // the nucleus, for example, is tested that is the only thing in the organelles list so this temporary
+                // allocation is not reused so just wastes memory and doesn't save on computation.
                 existingDefs ??= baseSpecies.Organelles.Select(x => x.Definition).ToHashSet();
                 if (existingDefs.Contains(organelle))
                     continue;


src/auto-evo/mutation_strategy/AddOrganelleAnywhere.cs
```diff
@@ -100,6 +100,8 @@ public static AddOrganelleAnywhere ThatConvertBetweenCompounds(Compound fromComp
         var workMemory2 = new List<Hex>();
         var workMemory3 = new HashSet<Hex>();
 
+        var existingDefs = baseSpecies.Organelles.Select(x => x.Definition);
```
hhyyrylainen: Does this do anything? Or does this need to? Because to me this looks like it just saves the LINQ statement as a variable but doesn't evaluate it here so it anyway gets re-evaluated each time where this was.

hhyyrylainen: Looks like CI agrees with me: `src/auto-evo/mutation_strategy/AddOrganelleAnywhere.cs:117 Possible multiple enumeration type: PossibleMultipleEnumeration`

So yeah this just saves the LINQ statement as a variable but doesn't save on how many times it is evaluated.

ljm862: Yep! I was rushing to push this before I had to go to the shop so didn't check everything was on the right file version.
I think it's an almost negligible change to be honest most of the time. Only at later generations does it end up being a time save overall.

hhyyrylainen:
I do wonder if the extra ToHashSet creation for each organelle attempt is a good use of memory... it's probably fine and we aren't still yet fully focused on reducing auto-evo memory allocations (I've done that for the gameplay code but due to how much auto-evo allocates memory, I've not tried to reduce memory allocations in it yet). So I guess for now that's a good "temporary" change as long as it does make things faster overall.

hhyyrylainen (approved):
I tested this and this is quite significantly faster, like at least feels like 30% faster, so this is very nice change.

I further optimized the unique organelles list. I did not entirely remove it as I didn't do profiling, but my few debugger uses for that seem to indicate that the "organelles" list usually has just a single thing when it is a unique organelle. Meaning that the temporary unique organelles allocation is likely never reused thus it might mean that the temporary allocation just uses memory but doesn't save on time at all. I think that might also be why the original code was written the way it was as temporarily caching the uniques list doesn't really help as the code doesn't seem to check multiple unique organelles in a row. So yeah I added a code comment about future changes here, but did not remove as I couldn't be sure I'm not missing some case where it helps enough to balance out my concerns about the nucleus placement check now basically wasting temporary memory. Sorry that this paragraph got a bit long...

ljm862:
> I further optimized the unique organelles list. I did not entirely remove it as I didn't do profiling, but my few debugger uses for that seem to indicate that the "organelles" list usually has just a single thing when it is a unique organelle

I ran a quick test, just for one world at 15 generations and here's the breakdown! All boundaries are inclusive.

2140675 Instances between 0 and 1
525196 Instances between 2 and 5
513883 Instances between 6 and 10
634239 Instances between 11 and 20
0 Instances 21+

For a total of 3,813,993 operations.

The test is probably not the most conclusive and the methodology might not be perfect, but it shows that about 50% of the time, there is 0-1 values in the loop and so the hashset is truly pointless. I'd go so far as to say for anything with fewer than 6 entries that the hashset is pointless to be honest. Which brings it to **about 70% of the time the hashset is pointless**!
Looking up online, the consensus seems to be that fewer than 100 elements it's better to use an array, especially if we only want to check once.

So you're right, we should remove the hashset instantiation, possibly caching as an array

hhyyrylainen:
> The test is probably not the most conclusive and the methodology might not be perfect, but it shows that about 50% of the time, there is 0-1 values in the loop and so the hashset is truly pointless. I'd go so far as to say for anything with fewer than 6 entries that the hashset is pointless to be honest. Which brings it to about 70% of the time the hashset is pointless!
Looking up online, the consensus seems to be that fewer than 100 elements it's better to use an array, especially if we only want to check once.

Did you also check *uniqueness*? Because the cache value is *only used for unique organelles*. So any organelle that is not marked as unique (and there's only a couple of unique organelles like the nucleus and binding agent) will not trigger the code that would need to use the cache. So only the organelle list size is important *whenever it has more than 1 unique organelle in it*.

For example if the list of organelles to test is:

- Chemoplast
- Nucleus
- Ferroplast

Then the code will run the loop 3 times, which could in theory help, *however* only nucleus is a unique organelle so the cached value is only used once due to the if statement:

```c#
            // Don't add duplicate unique organelles
            if (organelle.Unique)
            {
                 // inside here is where the cached hash set is even used...
```

So the caching will *only* help if the list of organelles to test has multiple unique organelles, for example:
- Nucleus
- Binding agent

Such a list of tests would save one loop of the organelles. And that I'm unsure if it exists because in my debugger run it seemed that the "evolve a nucleus" mutation strategy only contained the single nucleus organelle, meaning that the loop to check placements would only run once, so the cache would never be used.

ljm862:
> Did you also check _uniqueness_? 
Nope, it was just raw numbers 😅 
Thanks for the explanation, that makes it clear now what's going on. I'd misunderstood the `.Unique` to be "is unique to the microbe" which, now that I think about it, doesn't really make any sense...

I redid the test including a uniqueness check and these are the numbers:
3032638 Between 0 and 1.
136670 Between 2 and 5.

That's not very clear, so let's breakdown even further now we know the upper limit seems to be 5!
3457068 With 0
348652 With 1
127150 With 2
39544 With 3

It's _definitely_ only worth an array at best and quite possibly not even worth caching at all. What do you think?

hhyyrylainen:
I'd personally not cache it, unless this had a measurable impact on the overall time, because trying to remove this temporary allocation would make further improvements harder. My dream is that auto-evo would only ever need to allocate a few temporary species objects + the results, as those can't be avoided, but that's currently extremely far off.

PR #6531, opened 2025-12-02
Add AO textures to terrain chunks

dligr:

Adds ambient occlusion textures to most terrain chunks (except for the polymetallic ones, which I think are too round for AO to have a visible effect). This imitates indirect lighting more realistically and makes shadowed parts of terrain chunks look more 3-dimensional.

Example:
image

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)


assets/models/microbe_terrain/GabbroTerrainChunk.material
hhyyrylainen, line 1: Unless I'm missing something, all of these material files are now unused and should be deleted.

It seems like each chunk now has a unique material due to the ambient occlusion map being so that seems to be the reason these shared material files are no longer used.

Patryk26g:
I have tested it in-game and the terrain looks so much better with it

PR #6532, opened 2025-12-02
Refactored editor auto-evo prediction to full wait for queued runs

hhyyrylainen:

this should fix the intermittent error about modifying an organelle list while auto-evo is running

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/CellEditorComponent.cs
+++ b/src/microbe_stage/editor/CellEditorComponent.cs
@@ -210,6 +210,8 @@ public partial class CellEditorComponent :
 
     private SelectionMenuTab selectedSelectionMenuTab = SelectionMenuTab.Structure;
 
+    private bool autoEvoPredictionDirty = true;
+    private double autoEvoPredictionStartTimer = 5;
     private bool? autoEvoPredictionRunSuccessful;
     private PendingAutoEvoPrediction? waitingForPrediction;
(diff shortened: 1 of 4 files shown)


src/microbe_stage/editor/CellEditorComponent.cs
```diff
@@ -687,6 +686,28 @@ public override void _Process(double delta)
...
+                {
+                    // Start a new run once safe to do so
+                    autoEvoPredictionStartTimer = 0;
+                    StartAutoEvoPrediction();
+                }
+            }
+        }
+        else
+        {
+            CheckRunningAutoEvoPrediction();
```
ljm862, line 708: Do we want to check this if the editor is not ready?

hhyyrylainen: `waitingForPrediction` should always be null as it cannot have been started before the editor is ready. So basically that method just compares a null against true and returns early. So it should not be able to do anything bad in this case. And even if the waitingForPrediction was not-null from a quick glance it looks like it should be safe to apply all of the results anyway even if the editor somehow started a prediction run before being ready. So I don't really feel like the if conditions here need more cases.

ljm862 (approved):
All makes sense to me

PR #6535, opened 2025-12-03
Added graphics presets and an option to turn off current particles

hhyyrylainen:

Adds an easier way to set the graphics options for lower end devices. Includes auto-detecting of setting low mode with a few conditions

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/engine/LogInterceptor.cs
+++ b/src/engine/LogInterceptor.cs
@@ -104,6 +104,14 @@ public partial class LogInterceptor : Logger
             return;
         }
 
+        // Ignore unsupported antialiasing modes as it would be very complex to hide the options in the GUI:
+        // https://github.com/Revolutionary-Games/Thrive/pull/6535#issuecomment-3611112455
+        if ((rationale.Contains("only available when using the") && rationale.Contains("renderer.")) ||
+            (rationale.Contains("currently unavailable on") && rationale.Contains("renderer.")))
+        {
+            return;
+        }
+
         // Release-mode-only bug: https://github.com/Revolutionary-Games/Thrive/issues/5082
         if (rationale.Contains("tempt to disconnect a nonexistent connection from 'root:<Window"))
             return;


HyperbolicHadron:
Initial testing of the presets with my minimum spec recommended machine: i5-7300HQ 4 core, IGPU: HD Graphics 630

Tested with a mid to late game save, 10-15 medium size organisms on screen:

Very High- 8-9fps
High- 10-11fps
Medium- 14-15fps
Low- 26-28fps
Very Low- 29-30fps

Initial impressions seem to be that the Low preset is ideal for the minimum spec recommended machine

On this system, it seems like enabling background particles and distortion to backgrounds does not affect the Low preset performance. It might still be good to keep those disabled though to be conservative, even though I haven't seen much fps effects from those presets on other machines, though that could be made an extra differentiation from the Very Low preset.

On the other side, for Medium preset it seems that disabling chromatic aberration and bloom effect brings the fps from 14-15fps to 19-20fps. Not sure if this is recommended, but it is a possibility to consider.

Also, I think that the circles only microbe current particles option saves about 1-2fps on the low preset, which seems worth it. I could see no difference between circles only and None options.

src/microbe_stage/FluidCurrentDisplay.cs
```diff
@@ -121,4 +168,70 @@ protected override void Dispose(bool disposing)
...
+        // Switching between the enabled modes
+        if (old == Settings.MicrobeCurrentParticlesMode.All &&
+            initializedMode == Settings.MicrobeCurrentParticlesMode.OnlyCircles)
+        {
+            if (TrailEnabled)
+            {
+                TrailEnabled = false;
+                DrawPass1 = normalParticleMesh;
+                material.SetShaderParameter(particleDepthVariationParameterName, 1.0f);
+            }
```
dligr: This code block (and the other one in this function) is very similar to the one that changes particle modes in `ApplyBiome`. So, in my view, it would be better to move the immediate trail/circle switching code to its own function: its logic might change if trail particles are combined with circle ones, as was the plan for surface patches.

Or, to remove code duplication entirely, maybe there could be a function like `ApplyTrailSettings(Biome? biome, Settings.MicrobeCurrentParticlesMode mode)`, which would also include turning particles on/off.

hhyyrylainen:
> Graphics presets work correctly, apart from one bug I've noticed: changing preset to High and then setting anti-aliasing to MSAA + TAA doesn't switch the preset to Custom, but reloading the game after that does for some reason.

I forgot to put in a method call to refresh the preset on changing anti-aliasing mode. Will fix that.

src/microbe_stage/FluidCurrentDisplay.cs
hhyyrylainen: I guess I'll just update this code to call `ApplyBiome(previousBiome);` and do nothing if previousBiome is not set.

I originally wanted to do the minimum needed changes when changing the modes but I needed to write more and more code in there to handle all the possible state changes. I suppose just reinitializing everything will be fine as the option will rarely be changed. And it seems previousBiome is basically always set.

hhyyrylainen:
I finished testing on Windows and fixing those few small things. I'll next suppress the unsupported anti-aliasing mode errors as it'd requite a very complex system to suppress the incorrect values (because the available modes will depends on current renderer and the selected upscaling mode).

PR #6536, opened 2025-12-03
Pilus damage-relative volume.

### LINKED ISSUE #6525: Tweak the pilus hit sound
I think it's a bit much how often and loud the pilus sound can now trigger with the pilus rebalancing. I think that the minimum sound at least should be made much quieter on the pilus.

Accidental-Explorer:

This PR makes Pilus puncture sound volume relative to damage inflicted, with a damage floor under which no sound is played.
(to prevent sound spam)

**Related Issues**

closes #6525

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/systems/DamageSoundSystem.cs
+++ b/src/microbe_stage/systems/DamageSoundSystem.cs
@@ -71,7 +71,7 @@ public partial class DamageSoundSystem : BaseSystem<World, float>
                 {
                     // Play the pilus sound (volume is relative to damage inflicted)
                     soundEffectPlayer.PlaySoundEffect("res://assets/sounds/soundeffects/pilus_puncture_stab.ogg",
-                        damageEventNotice.Amount / Constants.PILUS_MAX_DAMAGE * 4.0f);
+                        0.75f + damageEventNotice.Amount / Constants.PILUS_MAX_DAMAGE * 3.25f);
                 }
                 else if (damageSource == "chunk")
                 {


hhyyrylainen:
I think that the sound has to play when damage is being dealt as otherwise there is no feedback, but I think the minimum volume could be low enough that the pilus sound is barely noticeable under, for example, 2 damage like your current implementation.

Accidental-Explorer:
> I think that the sound has to play when damage is being dealt as otherwise there is no feedback, but I think the minimum volume could be low enough that the pilus sound is barely noticeable under, for example, 2 damage like your current implementation.

Right now as far as I can tell the only time it does not play a sound is when in constant contact i.e. post-initial impact. Where the initial impact that lead to that contact always _does_ play a sound. I first tried it without that limit, but I felt like even at low volume the repeated noise felt a bit off.

But that's just my personal opinion, I have no issue with removing the floor if that feels better.

Edit: an alternative could also be to go back to the pilus mechanics and make the minimum delay between damage ticks a bit larger again. (and adapt minimum damage accordingly)

hhyyrylainen:
> Edit: an alternative could also be to go back to the pilus mechanics and make the minimum delay between damage ticks a bit larger again. (and adapt minimum damage accordingly)

I had actually been considering that why couldn't the min damage cooldown stay at 0.2? If 5 damage per second is too low, maybe you could also change the min pilus damage from 1 to 2 so in effect the DPS would stay the same while triggering the sound at half the rate?

Accidental-Explorer:
> I had actually been considering that why couldn't the min damage cooldown stay at 0.2? If 5 damage per second is too low, maybe you could also change the min pilus damage from 1 to 2 so in effect the DPS would stay the same while triggering the sound at half the rate?

Alright, I made exactly this change, so that this PR does not become a balance change.

hhyyrylainen (approved):
I tweaked the minimum sound volume so that it is at least audible but not very loud. I think this scaling of the sound to the damage amount is a nice feature.

PR #6545, opened 2025-12-08
Increase population indicator texture resolution

dligr:

Increases the resolution of the .svg texture used for population indicator, while keeping the same dot size.

I think the dots look better with a higher resolution texture given that they are surrounded by high-res textures, but please let me know if there was a reason they were low-res before.

image

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
At least I'm not aware of why the circles were such low resolution originally...

The code change here looks good to me.

hhyyrylainen (approved):
Seems fine to me.

PR #6562, opened 2025-12-10
Make the food chain display moveable

dligr:

The display was technically already draggable before, but the tree was too small for that to be actually possible. This PR adds a 450x225 pixel margin to fix that.

**Related Issues**

--

**Progress Checklist**

      break existing features:
      members reviewing this PR)

### DIFF
--- a/src/microbe_stage/editor/FoodChainDisplay.cs
+++ b/src/microbe_stage/editor/FoodChainDisplay.cs
@@ -159,7 +159,6 @@ public partial class FoodChainDisplay : Control
             if (node.CreatedControl != null)
             {
                 position += node.CreatedControl.Position;
-                GD.Print(node.CreatedControl.Position);
             }
         }
 
--- a/src/microbe_stage/editor/MicrobeEditorReportComponent.cs
+++ b/src/microbe_stage/editor/MicrobeEditorReportComponent.cs
@@ -109,8 +109,6 @@ public partial class MicrobeEditorReportComponent : EditorComponentBase<IEditorR
     private RunResults? autoEvoResults;
     private Func<LocalizedStringBuilder>? textReportGenerator;
 
-    private bool centerFoodChainScrollNextFrame;
-
     public enum ReportSubtab
     {
         AutoEvo,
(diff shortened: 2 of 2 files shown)


hhyyrylainen:
I think the left margin is too much as the food chain is now offset quite a bit to the right and it doesn't center itself. So I think the left margin should be much lower so that the initial view of the food chain is actually visible.

With this change a big chain displays like this initially which I find quite annoying as you always have to drag to see it:



This is after dragging and I think this is much closer to what it was before the left margin change


So I think the left margin size needs changes.

dligr:
The problem with decreasing left margin is that it would make the scroll range assymetrical, which is wrong in my opinion.

I have devised a solution with centering the food chain, though it requires a delayed update, which I'm not sure is the best solution. Updating scroll values immediately after updating the tree however doesn't work, because the values get immediately reset to zero.

src/microbe_stage/editor/FoodChainDisplay.cs
```diff
@@ -147,6 +147,25 @@ public void DisplayFoodChainIfRequired(RunResults autoEvoResults, Patch forPatch
...
+            return Vector2.Zero;
+
+        var position = Vector2.Zero;
+
+        foreach (var node in graphNodes)
+        {
+            if (node.CreatedControl != null)
+            {
+                position += node.CreatedControl.Position;
+                GD.Print(node.CreatedControl.Position);
```
hhyyrylainen: There's a debug print here.

src/microbe_stage/editor/MicrobeEditorReportComponent.cs
```diff
@@ -389,6 +396,24 @@ private void UpdateReportTabPatchName()
...
     }
 
+    private void UpdateFoodChainIfNecessary(RunResults? autoEvoResults)
+    {
+        if (selectedReportSubtab != ReportSubtab.FoodChain || autoEvoResults == null)
+            return;
+
+        foodChainData.DisplayFoodChainIfRequired(autoEvoResults, PatchToShowInfoFor, PlayerSpecies);
+
+        centerFoodChainScrollNextFrame = true;
```
hhyyrylainen: Can't this use `Invoke.Instance.Queue`?

hhyyrylainen:
I have one more question: is the margin used in any other scroll containers? I think it is now causing one more isse with the mouse pan speed being extremely slow when near the center but fast at the edges. I think that doesn't happen on the map or other screens. So I think that instead of using margins to add the extra space, increasing the min size of the container might be the right solution?

hhyyrylainen:
Or actually I just tested on the map screen and, yeah it does exhibit the same behaviour, so the margin isn't actually at fault...

PR #6563, opened 2025-12-10
Auto evo general balance

Accidental-Explorer:

This PR is a general rebalance on Auto-evo that got slightly out of hand.

- consolidated negative effect of energy expenditure towards one selection pressure (removed from two others)
- Made distinction between offensive and defensive pili (no more butt-spikes on predators)
- Generally reduce predation score based on how effective the target's defences are against the predator.
- Made prey Fear behaviour affect prey's average speed and usage of defensive pili
- Added mutation of cell membrane types to the energy consumption reduction miche (greatly increased membrane variety)
- Added a miche node at the root to evaluate resistance to all predators, instead of creating temporary miches later in the process.
- Added many conditions to skip parts of the predation score system as they're often not necessary.
- decreased population impact of cells dying on screen since it seemed (from player observations, among other things) to be wiping out species a bit too often. (endosymbionts, for example).
- Lowered density of some predation-related organelles to make them a slightly better fit for predators than autotrophic parts.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


src/auto-evo/simulation/SimulationCache.cs
```diff
@@ -327,8 +329,18 @@ public float GetPredationScore(Species predatorSpecies, Species preySpecies, Bio
...
+        var defensivePilusScore = preyToolScores.DefensivePilusScore;
+        var defensiveInjectisomeScore = preyToolScores.DefensiveInjectisomeScore;
+
+        var agressionScore = predator.Behaviour.Aggression / Constants.MAX_SPECIES_AGGRESSION;
+        var activityScore = predator.Behaviour.Activity / Constants.MAX_SPECIES_ACTIVITY;
+        var preyFearScore = prey.Behaviour.Fear / Constants.MAX_SPECIES_FEAR;
 
-        var behaviourScore = predator.Behaviour.Aggression / Constants.MAX_SPECIES_AGGRESSION;
+        // prey effectiveness at running away depends on how quickly they choose to run away
+        // and fully sessile prey can't run at all
```
adQuid: fully sessile prey can still run if they have non-zero fear. Feel free to test that in game in case something was updated that I missed.

Accidental-Explorer: I remembered some community comments that suggested otherwise. But I checked, and you're correct.

hhyyrylainen:
Looks like this is still a draft so I assume I don't need to urgently look at this now that I'm back...

Accidental-Explorer:
Correct, I'm still trying finagle things to get the right kind of random results. And I'll see if I can get some feedback on the performance from other people before I bother you with it.

Accidental-Explorer:
I will mark this as non-draft so maybe more people can take a look at it.

Very likely that I want to find ways to optimise it before if/when it is merged.

Patryk26g:
I run 15 generataions x 10 on master and this branch on the same seed, results:

Master -> Mean: 160s, s. deviation: 39s
Branch -> Mean: 179s, s. deviation: 32s,

So quite unclear if it's faster or slower, would need more data on that

Patryk26g:
Also I run this without these change, cause I forgot to update the branch and got these results: 
image

Mean: 147s, SD: 31s

Accidental-Explorer:
Thanks for testing! As long as it's not running overly slower, I am pleasantly surprised and satisfied.

It seems like sometimes even simple scoring changes have big effects due to affecting species splits and migration (even without the algorithms themselves changing in expense)

hhyyrylainen:
Is this safe to merge today because this has been ongoing for a really long time?

simulation_parameters/Constants.cs
hhyyrylainen, line 1: So this has a bunch of general balance changes as well (organelles.json and bio_processes.json)?

Accidental-Explorer:
> Is this safe to merge today because this has been ongoing for a really long time?

I am fairly happy with it as a general improvement over the current live version. @GameDungeon said they would still look at this again later on if they have time, but their initial glance did not spot any problems.

When it comes to the full species results, I am going to have to rebalance for GameDungeon's performance update anyway, (assuming that works out).

simulation_parameters/Constants.cs
Accidental-Explorer: There's a change in bio_processes.json that is fairly gameplay balance-neutral but made for auto-evo purposes. (Actually had explicit approval from Deus)

Organelles.json has exclusively slight density reductions to cell parts useful for predators. (So that auto-evo prefers them a bit.)

hhyyrylainen:
I tested the performance (I started before the latest 2 commits so if those impacted it those aren't yet taken into account).

So here's the times for this PR:
3:01
1:51
2:27
1:38
2:30

And from latest master:
1:24
1:28
2:17
1:03
2:02

So as can be seen this has a relatively big increase in processing time. About maybe 30% to even 50% increased auto-evo runtime.

But with my cache improvements that actually kind of cancels out so we can merge this but then we are back to the auto-evo performance we got in 1.0...

src/auto-evo/steps/ModifyExistingSpecies.cs
```diff
@@ -306,8 +306,6 @@ private void GetMutationsForSpecies(MicrobeSpecies microbeSpecies)
                 temporaryPressures.Add(traversalMiche.Pressure);
             }
 
-            SpeciesDependentPressures(temporaryPressures, miche!, microbeSpecies);
```
hhyyrylainen, line 309: Removing these calls removes usage of `AvoidPredationSelectionPressure` and as a result causes `SpeciesDependentPressures` method to be unused. If that is intended I think the method itself should also be deleted.

Accidental-Explorer: Makes sense, I didn't want to remove any method that people might be planning to use elsewhere without asking.

hhyyrylainen: Unless the performance PR by @GameDungeon uses this, I don't think anyone else would be using it...

Accidental-Explorer:
> So as can be seen this has a relatively big increase in processing time. About maybe 30% to even 50% increased auto-evo runtime.
> 
> But with my cache improvements that actually kind of cancels out so we can merge this but then we are back to the auto-evo performance we got in 1.0...

That is unfortunate. I wonder what exactly has the biggest negative effect here. Overall, species are more frequently larger than two hexes, that might be part of it.

We'll have to weigh whether the increased times are worth the benefits. In particular, I've seen better predators (spikes actually on the front now). Species now also more effectively counteract predation, both by being bigger, and especially using the variety of membranes available. I know the latter in specific was a request, and there's now typically a fairly even split between them on average.

In any case my next project will be trying to get GameDungeon's optimisation working.

hhyyrylainen:
Here's the latest commit times:

2:20
2:31
1:33
2:36
3:42

So it doesn't look like the latest commits changed the times, though that one nearly 4 minutes is either an outlier (bad luck) or potentially some situation where the auto-evo times start to run away.

hhyyrylainen:
I fixed the merge conflicts so that I can merge this in a bit.

I'll then make a separate PR of my auto-evo cache improvements which should be enough to offset all slowdown from this PR.

PR #6571, opened 2025-12-12
Fix easy difficulty stalling auto-evo

Accidental-Explorer:

This PR sets osmoregulation cost modifier to 0.9 in easy difficulty to avoid auto-evo not making any species.

**Related Issues**

**Progress Checklist**

      break existing features:
      members reviewing this PR)


hhyyrylainen:
I want to wait at least until after 1.0 before merging this. And even then I don't really want to have the easy difficulty to become significantly harder to survive with (as I think osmoregulation cost is almost the most important factor in what makes the game easier).

Accidental-Explorer:
> I want to wait at least until after 1.0 before merging this. And even then I don't really want to have the easy difficulty to become significantly harder to survive with (as I think osmoregulation cost is almost the most important factor in what makes the game easier).

I agree, and I would actually rather further _decrease_ osmoregulation cost for players in easy mode. So I think the better option is to look at the species splitting algorithm and fix it from that side. But I am not sure if I will be able to manage that myself.

Accidental-Explorer:
This is definitely not needed anymore, since the problem is already fixed.

